PR #3822, opened 2019-06-12
Do logging with GRPC interceptors

msteffen:
Right now, only CreateRepo has been changed:
https://github.com/pachyderm/pachyderm/pull/3822/files#diff-013fc8aac894793fb95d86dc816adf28L80

But this should in theory let us get rid of all of our logging boilerplate.

Other things we might be able to do fairly easily with GRPC interceptors:
- Structured errors
- Auth (we could go back to passing around ctx without worrying about whether auth data had been copied into the outgoing GRPC metadata. In fact, any time we're passing around info in GRPC metadata, we could use GRPC interceptors to copy it from the incoming metadata to the outgoing metadata)
- Remove some of the tracing boilerplate. Tracing already uses interceptors, but they're 3rd party. We could fork that code and get more of what we want


msteffen:
Closing this PR as obsolete

PR #3829, opened 2019-06-14
Spoutservice

kevindelgado:
~WIP: wanted to get a working PR up by end of day.~

~This works, but a better solution (than just checking for spout existence before service existence) is to add a service field to `pps.Spout` (of type `*pps.Service`) and then in the `worker_rc` create a k8s user service based on the existence of `options.service` or `options.spout.Service`, but I couldn't quite get it to work yet.~

I've added the service field to pps.Spout and am checking for its existence within `master.go` instead of `worker_rc.go`.


src/server/pachyderm_test.go
```diff
@@ -9150,6 +9151,86 @@ func TestSpout(t *testing.T) {
...
+				ParallelismSpec: &pps.ParallelismSpec{
+					Constant: 1,
+				},
+				Input:  client.NewPFSInput(dataRepo, "/"),
+				Update: false,
+				Spout: &pps.Spout{
+					Service: &pps.Service{
+						InternalPort: 8000,
+						ExternalPort: 31800,
+						Type:         "NodePort",
```
kevindelgado: As asked in slack (https://pachyderm.slack.com/archives/CDWDPKPMY/p1560534040273200), I'd like the service within the spout field to default to type "NodePort" just as a regular pipeline service does, but I couldn't find where we actually do that, if any of you guys happen to know?

kevindelgado:
Ready for review now, one question left in line, PTAL

src/server/pachyderm_test.go
```diff
@@ -9150,6 +9151,94 @@ func TestSpout(t *testing.T) {
...
 	})
+	t.Run("ServiceSpout", func(t *testing.T) {
+		dataRepo := tu.UniqueString("TestServiceSpout_data")
+		require.NoError(t, c.CreateRepo(dataRepo))
+
+		annotations := map[string]string{"foo": "bar"}
+
+		// Create a pipeline that listens for tcp connections
+		// on internal port 8000 and dumps whatever it receives
+		// (should be in the form of a tar strea) to /pfs/out.
```
adelelopez: typo: strea -> stream

src/server/pachyderm_test.go
adelelopez: it's here: https://github.com/pachyderm/pachyderm/blob/master/src/server/pps/server/api_server.go#L1549

adelelopez: it's here: https://github.com/pachyderm/pachyderm/blob/master/src/server/pps/server/api_server.go#L2135

kevindelgado:
Fixed some typos and NPEs that travis caught and made it so the spout service type defaults to NodePort just like vanilla pipeline services. Will merge in once travis succeeds unless I hear otherwise.

src/server/pachyderm_test.go
kevindelgado: Dope, thanks!

PR #3833, opened 2019-06-14
Edit the Combining section

svekars:


doc/cookbook/combining.md
```diff
@@ -41,33 +86,32 @@ The first pipeline takes a union of `A` and `B` as inputs, each with glob patter
   B.json
 ```
 
-Note, that when performing this grouping:
+When you group your data, set the following parameters:
 
-- You should use `"lazy": true` to avoid unnecessary downloads of data.
+- Set `"lazy": true` to avoid unnecessary downloads of data.
```
gabrielgrant: I believe current best practice is actually to use `empty_files` here instead of `lazy`

gabrielgrant:
Thanks!

Issue #3837, opened 2019-06-17
put-header and put-footer are lost in 1.9


gabrielgrant:
`put-header` and `put-footer` commands are referenced in [the splitting docs](http://docs.pachyderm.io/en/latest/cookbook/splitting.html#specifying-a-header-or-footer), but no longer seem to exist. AFAICT there is no `pachctl put header` either, so it seems these got lost in the migration to the new verb-oriented CLI syntax?

This should be fixed in the CLI, and updated in the docs

Edit -- related: #3613, #3000

Tryneus:
It appears that `put-header` and `put-footer` were only available in `pachctl` versions `v1.7.10` - `v1.7.14`, so they weren't missed in the migration.  I think this is solely a docs issue, as these were removed a long time ago.

gabrielgrant:
Hrm...interesting... are you sure about that version range? I don't recall us ever making a conscious decision to remove that feature, and as of the release of 1.8 at the end of last year we were touting `put-header` as our official answer for how to handle CSV data: https://pachyderm.io/2018/11/15/structured-data.html

msteffen:
Basically, the way Header and Footer worked in 1.7 was:
- They're attached to a directory
- Anything underneath that directory (even nested several levels) inherits the Header and the Footer
- You can add/remove a Header and Footer from a dir with those CLI commands, and the data will be removed from the directory's children

The problem with this was that to read a file's contents, you had to read not just the file, but every directory above the file in case it head header/footer data. In 1.8, hashtrees (now FileSets) we redesigned to hold more data and be streamed. This meant that reading every parent of a file amounted to at least one extra S3 read (including streaming a potentially nontrivial amount of data), regardless of whether the feature was being used or not. All `GetFile` RPCs, and the process of downloading a worker's input datums in particular, would've gotten unacceptably slow and expensive if we'd naively implemented the feature the same way in 1.8

So instead the way we supported headers and footers in 1.8 (which we do still use for CSV and SQL data) is that headers and footers can only be set by `put-file --split`, they only affect the direct children of a directory, and files themselves note whether their parent dir contains header data, so that people not using the feature don't have to do extra S3 reads every time they get a file
Because files now note whether their parent dir contains a header, there's no way to remove a header from a dir without changing every file underneath the dir. Rather than try to support that, we simply prevent people from adding or removing headers, and require them to go through put-file --split

JoeyZwicker:
[...]

PR #3839, opened 2019-06-17
#3837 Update the header/footer content

svekars:


gabrielgrant:
As I mentioned in #3837, I'm not totally sure we want to remove this, since I think this is still needed for CSV. If that's the case, I think this is, in fact, something to be fixed on the eng side

doc/cookbook/splitting.md
```diff
@@ -133,116 +133,91 @@ the split files.
...
   ```
 
-## Specifying a Header or Footer
+## Specifying a Header
 
-Additionally, if your data has a common header or footer, you can specify them
-manually by using `pachctl put-header` or `pachctl put-footer`. This
-functionality is helpful for CSV data.
+If your data has a common header, you can specify it
+manually by using `pachctl put file` with the `--header-records` flag.
```
JoeyZwicker: is the --header-records flag used for SQL data as well? Is it still used to specify number of lines for the header?

svekars: Actually, I think it's `--split=(json|line|csv)` only according to help. Let me fix that.

PR #3842, opened 2019-06-18
Try to fix TestExtractRestore

msteffen:


msteffen:
Deleting this PR and branch b/c they're stale

PR #3871, opened 2019-06-27
[DOC]Add a spout and service description

svekars:


doc/fundamentals/spouts.md
```diff
@@ -282,3 +282,26 @@ setting it to `true` would be like having the `--overwrite` flag specified on ev
...
+
+You can create a pipeline that can act as both a spout and a service.
+If you want your spout pipeline to combine these two functionalities, add
+the following to the spout pipeline specification:
+
+```bash
+
+"spout": {
+    "overwrite": false,
+    "service": {
```
JoeyZwicker: Just to make sure you got validation from eng on this... Is the service info here supposed to be embedded within the spout json object? So you can service as a top-level object or under the spout?

If so, why aren't they just independent fields like service and spout already were?

kevindelgado: Yes it is supposed to be embedded. The reason being is that there are three mutually exclusive "types" of API servers (`spoutSpawner`, `serviceSpawner`, and `jobSpawner`). 

In order to have both a spout and a service, you are actually declaring that you want an API server of type `spoutSpawner` that also happens to expose a kubernetes service (which is a much narrower functionality that wanting a `serviceSpawner` type API server, which does a multitude of other things, such as waiting to create the kubernetes service until after the first commit is made, that are incompatible with the behavior of spouts).

Thus we figured the most straightforward approach was to reserve the top-level object for indicating the type of API server you want for the pipeline (and embed any configurations of that API server into fields of the top level object).

JoeyZwicker: OK that makes sense. PR LGTM

> (which is a much narrower functionality than wanting a serviceSpawner type API server, which does a multitude of other things, such as waiting to create the kubernetes service until after the first commit is made, that are incompatible with the behavior of spouts).

@kevindelgado Is this just because of the assumptions we built into `serviceSpawner` right now and in the future this might be something worth refactoring to not have this incompatibility, but this solution was easier in the short term? Or are these differences between `spoutSpawner` and `serviceSpawner` actually incompatible so this is the longterm solution? 

I don't remember why we wait have `serviceSpawner` wait for a commit -- I guess because what the heck would it be serving if we didnt? -- but that doesn't seem a hard rule that necessarily needs to be true. I imagine there are other incompatibilities too?

kevindelgado: The `spoutSpawner` and `serviceSpawner` are fairly divergent from each other https://github.com/pachyderm/pachyderm/blob/master/src/server/worker/master.go#L196-L322

I don't see how they would be reconciled without deeply rethinking how service pipelines work. @jdoliner has mentioned that the long term goal is to use kubernetes CRD's to enable exposing arbitrary kubernetes resources from pipelines. This seems like a more worthwhile  long-term solution than trying to reconcile these two API server types at the spwaner level.

JoeyZwicker: makes sense. ty for the detailed answer

doc/reference/pipeline_spec.md
```diff
@@ -151,6 +153,26 @@ create pipeline](../pachctl/pachctl_create_pipeline.html) doc.
...
+  "overwrite": bool
+  \\ Optionally, you can combine a spout with a service:
+  "service": {
+        "internal_port": int,
+        "external_port": int,
+        "annotations": {
+            "foo": "bar"
+        }
+    }
+}
```
svekars: @kevindelgado , can you please confirm that this is correct or needs changes?

kevindelgado: I think it is confusing to place it here with the list `input` options, because it makes it seem like a spout is an type of `input` (which it is not, it is just an optional field on the top-level object. So I would place it immediately after the service object (line 78).

I'm also not sure what the `env` object here is, can you clarify where this is from?

JoeyZwicker: I think he approved above

doc/reference/pipeline_spec.md
```diff
@@ -575,6 +597,17 @@ https://github.com/<your_org>/<your_repo>/settings/hooks/new
 ```
 Or navigate to webhooks under settings. Then you'll want to copy the `Githook URL` into the 'Payload URL' field.
 
+### Spout
```
kevindelgado: Similarly, I would move this description to immediately after the description of services (ie. after the paragraphed headed `Service (alpha feature, optional)`) 

Also be sure to mention that it is optional, and that either `Spout` OR `Service` can be set, but NOT both.

doc/fundamentals/spouts.md
adelelopez: spoutSpawners is basically just a stripped down version of serviceSpawner with a few spout-specific tweaks. Having `/pfs/out` be a named pipe seems like the main issue to me.

doc/reference/pipeline_spec.md
```diff
@@ -631,13 +644,28 @@ in the input repos.
...
+container, `"external_port"` is the port on which it is exposed through the
+`NodePorts` functionality of Kubernetes services. After a service has been
+created, you should be able to access it at
 `http://<kubernetes-host>:<external_port>`.
 
+### Spout (optional)
+
+`spout` is a type of pipeline that processes streaming data.
+Unlike a union or cross pipeline, a spout pipeline does not have
+a PFS input. Instead, it opens a Linux *named piped* into the source of the
```
adelelopez: piped -> pipe

PR #3875, opened 2019-06-29
Added GetOneTimePasswordCmd()

### LINKED ISSUE #3869: Expose auth.GetOneTimePassword() API endpoint in pachctl
Right now the CLI has the `auth.GetAuthToken()` API endpoint plumbed through, but, [as mentioned in the protos](https://github.com/pachyderm/pachyderm/blob/master/src/client/auth/auth.proto#L360), this endpoint is only really designed to be used by pipelines. Instead, most user-facing uses should instead be using [auth.GetOneTimePassword()](https://github.com/pachyderm/pachyderm/blob/master/src/client/auth/auth.proto#L428), since this is what generates a the one-time pass that is accepted by `pachctl auth login` (and is generally more secure than printing out/passing around long-lived auth tokens)

kevindelgado:
Fixes #3869

### DIFF
--- a/src/server/auth/cmds/cmds.go
+++ b/src/server/auth/cmds/cmds.go
@@ -513,7 +513,7 @@ func GetOneTimePasswordCmd() *cobra.Command {
 			return nil
 		}),
 	}
-	return cmdutil.CreateAlias(getOneTimePassword, "auth get-one-time-password")
+	return cmdutil.CreateAlias(getOneTimePassword, "auth get-otp")
 }
 
 // Cmds returns a list of cobra commands for authenticating and authorizing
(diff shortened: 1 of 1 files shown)


src/server/auth/cmds/cmds.go
```diff
@@ -490,6 +490,32 @@ func UseAuthTokenCmd() *cobra.Command {
...
+				Subject: subject,
+			})
+			if err != nil {
+				return grpcutil.ScrubGRPC(err)
+			}
+			fmt.Println(resp.Code)
+			return nil
+		}),
+	}
+	return cmdutil.CreateAlias(getOneTimePassword, "auth get-one-time-password")
```
msteffen: how would you feel about `get-otp`? Just because `get-one-time-password` seems kind of long?

kevindelgado: Yea, makes sense. I think my original logic for keeping it long was that the long form is how it is defined in the auth proto service and wanted to keep it consistent, but that reasoning feels kinda lame on second thought. Fixed.

Issue #3880, opened 2019-07-01
CLI examples break formatting on `pachctl` docs


ysimonson:
When examples of a command are available, they break the formatting of commands in the pachctl docs. See for example the docs for [`pachctl extract`](http://docs.pachyderm.io/en/latest/pachctl/pachctl_extract.html).

ysimonson:
Appears to be fixed by #3888

PR #3888, opened 2019-07-03
Remove shell syntax highlighting on doc examples as this breaks stuff

Tryneus:
`cobra`'s `GenMarkdownTree` already wraps examples in \`\`\`s, so adding our own broke the rendering.  See #3880.

### DIFF
--- a/src/server/cmd/pachctl-doc/main.go
+++ b/src/server/cmd/pachctl-doc/main.go
@@ -1,13 +1,11 @@
 package main
 
 import (
-	"fmt"
 	"os"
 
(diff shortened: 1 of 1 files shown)


JoeyZwicker:
@svekars ^

Tryneus:
@jdoliner - yes, but these were at the same time moved from the description field to the `Examples` field in `cobra.Command`.  When `cobra` generates docs, it automatically wraps examples in `\`\`\``, which results in this:
image

(see #3880 where this was reported)

PR #3936, opened 2019-07-23
fix dash compat file for pach 1.9.1

gabrielgrant:
for some reason this was pointing at an old version of the dash...
now points at 1.9.0 dash release


kevindelgado:
Actually @gabrielgrant I just tried building pachctl from this revision (`pachctl version 1.9.1-d4e7f3e9653313b87bcbbf885972b6b1d71f8be2`) When I try to spin up a pach cluster I am still seeing it use `pachderm/dash:1.8-preview-7` for the dash image. Am I just doing something dumb or can you confirm that you are able to see it default to using dash v1.9 when you spin up a pach cluster?

gabrielgrant:
sorry, should have clarified: this PR actually needs to get merged before it will have an effect. The updated file (like the specific one in the repo hosted at github.com/pachyderm/pachyderm) gets checked dynamically by pachctl at deploy time

PR #3938, opened 2019-07-24
Makefile enhancements

pappasilenus:
I added data download targets for gatk & object detection so they're built with one command.
Added a clean target so that you can redownload the data if needed.
Updated gatk so it creates the reference repo and adds the reference data.

Now you can just type "make hyperparameter" or "make gatk" and they'll build with that one command.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,4 +1,4 @@
-.PHONY: opencv object_detection delete minikube
+.PHONY: opencv object_detection delete minikube gatk hyperparameter
 
 GATK_GERMLINE_FILES	=	gatk/GATK_Germline/data/ref/ref.dict \
 				gatk/GATK_Germline/data/ref/ref.fasta \
@@ -40,13 +40,13 @@ ml/object-detection/frozen_inference_graph.pb:
 	wget -p http://download.tensorflow.org/models/object_detection/ssd_mobilenet_v1_coco_11_06_2017.tar.gz --directory-prefix /tmp
 	tar -C ml/object-detection --strip-components 1 -xvf /tmp/download.tensorflow.org/models/object_detection/ssd_mobilenet_v1_coco_11_06_2017.tar.gz ssd_mobilenet_v1_coco_11_06_2017/frozen_inference_graph.pb
 
-object_detection: ml/object-detection/frozen_inference_graph.pb
+object-detection: ml/object-detection/frozen_inference_graph.pb
 	pachctl create repo images
 	pachctl create repo training
-	pachctl create pipeline -f ml/object-detection/model.json
-	pachctl create pipeline -f ml/object-detection/detect.json
 	pachctl put file training@master:frozen_inference_graph.pb -f ml/object-detection/frozen_inference_graph.pb
 	pachctl put file images@master:dogs.jpg -f ml/object-detection/images/dogs.jpg
+	pachctl create pipeline -f ml/object-detection/model.json
+	pachctl create pipeline -f ml/object-detection/detect.json
 	pachctl put file images@master:kites.jpg -f ml/object-detection/images/kites.jpg
 
 hyperparameter:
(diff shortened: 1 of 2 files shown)


pappasilenus:
@JoeyZwicker I screwed up the commit somehow with data from the docs project.  I fixed it here by reverting that one file, so this just contains the makefile now.

examples/Makefile
```diff
@@ -1,25 +1,95 @@
...
+	wget -p http://download.tensorflow.org/models/object_detection/ssd_mobilenet_v1_coco_11_06_2017.tar.gz --directory-prefix /tmp
+	tar -C ml/object-detection --strip-components 1 -xvf /tmp/download.tensorflow.org/models/object_detection/ssd_mobilenet_v1_coco_11_06_2017.tar.gz ssd_mobilenet_v1_coco_11_06_2017/frozen_inference_graph.pb
+
+object_detection: ml/object-detection/frozen_inference_graph.pb
 	pachctl create repo images
 	pachctl create repo training
 	pachctl create pipeline -f ml/object-detection/model.json
 	pachctl create pipeline -f ml/object-detection/detect.json
-	pachctl put file training@master:frozen_inference_graph.pb -f ~/ssd_mobilenet_v1_coco_11_06_2017/frozen_inference_graph.pb
+	pachctl put file training@master:frozen_inference_graph.pb -f ml/object-detection/frozen_inference_graph.pb
```
JoeyZwicker: Minor nit:

This should put the frozen inference graph and the first image, dogs before it creates the pipelines. In the current order, you'll get a job with zero datums since there is just the frozen inference graph but no images yet. 

After creating the pipelines, then it puts the second image, kites and you'll get a new job for that.

examples/Makefile
```diff
@@ -1,25 +1,95 @@
...
 	pachctl put file images@master:dogs.jpg -f ml/object-detection/images/dogs.jpg
 	pachctl put file images@master:kites.jpg -f ml/object-detection/images/kites.jpg
```
JoeyZwicker: Broader note:

Is it easy enough to add comments or something like that to a makefile? It'd be nice to have the 3rd object detection images (airplane) or 3rd batch of images for opencv, etc available in a copy/paste instead of copying the previous put from the makefile and editing it

examples/Makefile
JoeyZwicker: Alternatively, you can use transactions to do the first two put files so no job triggers

examples/Makefile
```diff
@@ -1,25 +1,95 @@
...
+	pachctl put file raw_data@master:iris.csv -f ml/hyperparameter/data/noisy_iris.csv
+	pachctl put file parameters@master:c_parameters.txt -f ml/hyperparameter/data/parameters/c_parameters.txt --split line --target-file-datums 1
+	pachctl put file parameters@master:gamma_parameters.txt -f ml/hyperparameter/data/parameters/gamma_parameters.txt --split line --target-file-datums 1
+	pachctl create pipeline -f ml/hyperparameter/split.json
+	pachctl create pipeline -f ml/hyperparameter/model.json
+	pachctl create pipeline -f ml/hyperparameter/test.json
+	pachctl create pipeline -f ml/hyperparameter/select.json
+
+$(GATK_GERMLINE_FILES): 
+	mkdir -p gatk/GATK_Germline
```
JoeyZwicker: will this fail in an unfriendly way if we've already downloaded these files?

examples/Makefile
```diff
@@ -1,25 +1,95 @@
...
+	wget -p https://s3-us-west-1.amazonaws.com/pachyderm.io/Examples_Data_Repo/GATK_Germline.zip --directory-prefix=/tmp
+	unzip -o /tmp/s3-us-west-1.amazonaws.com/pachyderm.io/Examples_Data_Repo/GATK_Germline.zip  data/ref/ref.dict data/ref/ref.fasta data/ref/ref.fasta.fai data/ref/refSDF/* data/bams/*  -d gatk/GATK_Germline
+
+gatk_demo: $(GATK_GERMLINE_FILES)
+	pachctl create repo reference
+	pachctl put file reference@master:ref.dict  -f gatk/GATK_Germline/data/ref/ref.dict
```
JoeyZwicker: can't this be `put-file -r`  for the ref directory? isn't that what the readme does, or is there other crap in that dir that breaks things?

examples/Makefile
JoeyZwicker: I noticed the `clean` target below. Is there no way to make this friendlier so it doesn't barf if the files already exist?

examples/Makefile
pappasilenus: Ah, I was just using the Makefile as currently written here, but I can reconfigure it.  I won't do transactions, I'll just reorder things.

examples/Makefile
pappasilenus: If the files are already downloaded and copied in the places I specify, the dependency targets should not trigger ($(GATK_GERMLINE_FILES) for gatk_demo and  ml/object-detection/frozen_inference_graph.pb for object_detection).  That's makefile magic!

examples/Makefile
pappasilenus: I don't copy the other crap, just these files, so it could just be a putfile -r if this Makefile is  how you create the source of the data.  If you manually download and unpack it into that location, I don't know what would happen with a put -r.  I don't think this is worth testing, so I'm inclined to leave it as is.

examples/Makefile
pappasilenus: done.

PR #3945, opened 2019-07-26
[3463] Clarify the brew command usage

svekars:
- Editorial cleanup.

### DIFF
--- a/doc/managing_pachyderm/upgrades.md
+++ b/doc/managing_pachyderm/upgrades.md
@@ -1,5 +1,7 @@
 # Upgrades
 
+<!-- This section needs a thorough cleanup-->
+
 - [Introduction](#introduction)
 - [General upgrade procedure](#general-upgrade-procedure)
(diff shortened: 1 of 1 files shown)


doc/managing_pachyderm/upgrades.md
```diff
@@ -1,116 +1,102 @@
-# Upgrades
+# Upgrade Pachyderm to a Minor Version
```
JoeyZwicker:
```suggestion
# Upgrade Pachyderm Minor Versions
```
I don't think you're really upgrading "to" a minor version but "between minor versions"?

Also, I wonder if we shouldn't use the term "minor" or "major" version at all since we're not actually following the correct definitions of those in semantic versioning. I dont actually feel strongly about any of this, just a minor thought

svekars: I changed to "Upgrade Pachyderm".

Issue #3949, opened 2019-07-30
Default Auth Config


msteffen:
# Background
We want to change our `auth` configuration such that the semantics of an empty auth config (no ID providers enabled) are not the default semantics of our auth system (by default, GitHub is an enabled ID provider). This change of representation interacted poorly with the `live_config_version` field in our auth config, so this design describes how different states of the auth config should be represented internally so that both fields have sensible semantics.

# Core Semantics
These are the semantics that I believe auth's config API needs to have to be well-founded:

1. Calling `auth.GetConfiguration()` on a cluster where the auth config has never been set should yield the default config (something like the following):
```javascript
{
  "live_config_version": 1, // prevents RMW errors
  "id_providers": [
    {
      "name": "GitHub",
      "description": "oauth-based authentication with github.com",
      "github": {}
    }
  ]
}
```

2. Calling `auth.SetConfiguration` should cause the auth system's `live_config_version` to increase monotonically:
```shell
$ pachctl auth set-config -f foo.config
$ pachctl auth get-config | jq .live_config_version
1
$ pachctl auth set-config -f bar.config
$ pachctl auth get-config | jq .live_config_version
2
$ pachctl auth set-config -f baz.config
$ pachctl auth get-config | jq .live_config_version
3
```

3. Calling `auth.SetConfiguration` where the request's `live_config_version` doesn't match the `live_config_version` of pachd's config in etcd should yield an error (likely a R+M+W that was interrupted by another user):
```shell
$ pachctl auth get-config >foo.config
$ pachctl auth set-config -f foo.config # successful RMW
$ pachctl auth set-config -f foo.config # this config version was just replaced
error: expected config version 1 but live config has version 2
```

## Nice-to-have semantics
- Calling `auth.SetConfiguration` where the config's `live_config_version` isn't set should be a blind write (allows users to iterate on their config—the use-case is where users aren't editing a functional config, but rather iterating on establishing an initial functional config)

  Note that because the default config includes `live_config_version: 1`, there is no case where a user should call `auth.GetConfiguration` with `live_config_version` unset, and therefore there's no way that they would accidentally blind-write over another user's changes by accident. See below under "Internal Representation" for a note on users with an existing config where the `live_config_version == 1`.

- Calling `auth.SetConfiguration` where `req.config == nil` re-establishes the default config (though the config version should still increase, to avoid confusion over "version N" of the config).

[...]

PR #3954, opened 2019-07-31
Refactor auth config

msteffen:
Pull a lot of auth config code into `src/server/auth/server/config.go`. Also, change the internal representation of the auth config. Full design is at  https://github.com/pachyderm/pachyderm/issues/3949, but two highlights:
- no auth config in etcd => "default" auth config is the active config
- once an auth config is in etcd, there is no way to delete it. Sending an empty `SetConfiguration` RPC writes the default auth config into etcd (with the latest version)

(fix https://github.com/pachyderm/pachyderm/issues/3949)


src/server/auth/server/api_server.go
```diff
@@ -74,6 +73,8 @@ const (
 	SamlPort = 654
 )
 
+var defaultAuthConfig = authclient.AuthConfig{}
```
kevindelgado: It is here where I want to modify the default auth config to be 
```{
  "live_config_version": 1, // prevents RMW errors
  "id_providers": [
    {
      "name": "GitHub",
      "description": "oauth-based authentication with github.com",
      "github": {}
    }
  ]
}
``` 
right?

src/server/auth/server/api_server.go
```diff
@@ -2258,16 +2206,16 @@ func (a *apiServer) canonicalizeSubject(ctx context.Context, subject string) (st
...
 
 	// check prefix against config cache
-	if a.configCache != nil {
-		if prefix == a.configCache.IDP.Name {
-			return subject, nil
-		}
-		if prefix == path.Join("group", a.configCache.IDP.Name) {
-			return subject, nil // TODO(msteffen): check if this IdP supports groups
+	if config := a.getCacheConfig(); config != nil {
+		for _, idp := range config.IDPs {
```
kevindelgado: From our discussion, we are going to continue to canonicalize the subject based only on the IDP's name and not on anything that is in the various *Options fields, correct?

src/server/auth/server/api_server.go
msteffen: yup, exactly

src/server/auth/server/api_server.go
msteffen: I'm not 100% sure I undrestand the question, but I think that's right (this is also what we currently, do, if I understand the question).

To be clear, and maybe a little pendatic
1) I think the plan for now is that post-change, an IDP's `name` is `github` if and only if it has a `GitHubOptions`. It's just a matter of whether there's an IDP with `GitHubOptions`), or equivalently, an IDP whose `name` is `github`.
2) a SAML IDP can give you two subject types if it supports groups:
- `<idp.name>:<subject>` (e.g. `okta:matthew`)
- `group/<idp.name>:<subject>` (e.g. `group/okta:engineering`).
but we don't currently check the IDP to make sure `saml.group_attribute` is set before allowing `group/<idp_name>`-type subjects through `canonicalizeSubject`. Perhaps we should but I think it's fine to keep that outside the scope of this change.

PR #3956, opened 2019-07-31
kubeflow tfjob/s3 gateway example

pappasilenus:
Yes, it's true.  This is done.

Call me first gear, because I'm slow but I never stop.  (

Apologies to the real First Gear, quoted here: https://www.economist.com/united-states/2019/07/25/hotshots-and-global-warming-in-alaska)


examples/s3gateway/README.md
```diff
@@ -0,0 +1,152 @@
...
+If you want to test this in your own Kubeflow/Pachyderm deployment, 
+you can either build your own container using the `Dockerfile.tf_job_s3_gateway.yml` included with this example or
+use the prebuilt container with this code and sample data hosted in the Pachyderm docker registry.
+You create a TFJob using a TFJob manifest.
+We've included one with this example, 
+`tf_job_s3_gateway.yaml`, 
+which you would use to deploy the TFJob to your Kubeflow app
+after you've edited it for your environment.
+
+Here are the steps you need to take to run this in your environment using our prebuilt images.
```
JoeyZwicker: This is a lot of steps with a good amount of nuance. How can we simplify this? 

It seems like it can be broken down into more discreet categories:
1. Add your input data to pachyderm
     - create repo, create branch, add data, whatever.
2. We use the following default namespace, bucket, and repo names. If you have different names for any of these, make the BLAH BLAH changes in the manifest or wherever?
3. deploy the TFJob. You can track progress with...

examples/s3gateway/README.md
```diff
@@ -0,0 +1,152 @@
...
+6. Deploy the manifest to your Kubeflow using  `kubectl apply -f tf_job_s3_gateway.yaml`
+7. Using your Kubeflow TFJob dashboard or `kubectl`, monitor the TFJob pod created until it completes.
+   The pod will be named after the name of the TFJob in the manifest,
+   which is `s3-gateway-example` by default, 
+   and would be something like `s3-gateway-example-worker-0`.
+8. Using your Kubeflow TFJob dashboard or `kubectl`, look at the logs for that TFJob.
+   A sample log is included in the file `sample_tf_job_logs.txt` in this example.
+
+This demonstrates that anywhere that the `file_io` library is used, 
+you can put data into and get data out of Pachyderm versioned data repositories.
```
JoeyZwicker: Ah, so I think I just figured out a major confusion of mine. You're actually adding data to pachyderm the TFJOb script first, then reading it back out? Is that right?

If so, we should reverse those as it makes a much more straightforward demo. The whole point, I think, is that users have data in pachyderm and they want to process that data with a tfjob. So they should create the repo, add the data, and then read that in tfjob, do something with it (which might be do nothing) and then write it back to pachyderm in a different repo. 

Also, we should check with eng about how provenance info would or would not get tracked here? Don't need to cover this in this example, but we should understand that.

examples/s3gateway/tf_job_s3_gateway.py
```diff
@@ -0,0 +1,49 @@
...
+                      help="""The bucket where files will be put. This overrides the default and the environment variable S3_BUCKET.""",
+                      default=s3_bucket)
+  parser.add_argument('-i', '--inputpath', required=False,
+                      help="""The directories to walk for files to put in the bucket""",
+                      default=input_path)
+  
+  args, unparsed = parser.parse_known_args()
+  tf.app.run(main=main, argv=[sys.argv[0]] + unparsed)
+
+# todo
```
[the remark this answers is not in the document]
pappasilenus: I cleaned that up.  All those tasks were done.

examples/s3gateway/tf_job_s3_gateway.py
```diff
@@ -0,0 +1,49 @@
...
+from __future__ import print_function
+import os
+import argparse
+import sys
+import tensorflow as tf
+from tensorflow.python.lib.io import file_io
+
+# this is a directory with two directories in it
+# containing the first chapters of Moby Dick and A Tale of Two Cities
+# that's built into the image
```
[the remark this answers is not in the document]
pappasilenus: I elaborated on it in the  README in the commit I just pushed.  The idea here is to have an example that will let our customers verify that they've set up Pachyderm and Kubeflow to interoperate productively that's as simple as possible.

examples/s3gateway/README.md
```diff
@@ -0,0 +1,152 @@
...
+does simple reads and writes to a Pachyderm repo.
+
+The script `tf_job_s3_gateway.py` takes two flags `-b` or `--bucket`, 
+which specifies the `s3://` url to the bucket, and
+`-i` or `--inputpath`, 
+which specifies the local directory or mount point which will be the source of files to copy to the bucket.
+
+In `Dockerfile.tf_job_s3_gateway`, 
+that script is built into a Docker image
+along with the local folder `testdata`,
```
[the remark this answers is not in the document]
pappasilenus: See the edits I made to the README and the comment above.

examples/s3gateway/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+    for dirpath, dirs, files in file_io.walk(s3_path +  args.inputpath, True):
+        for file in files:
+            newpath = dirpath + "/" + file
+            print("printing {} in {}  as string: >>{}<<".format(file, dirpath, file_io.read_file_to_string(newpath, False)))
+```
+
+If you want to test this in your own Kubeflow/Pachyderm deployment, 
+you can either build your own container using the `Dockerfile.tf_job_s3_gateway.yml` included with this example or
+use the prebuilt container with this code and sample data hosted in the Pachyderm docker registry.
```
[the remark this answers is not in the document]
pappasilenus: Yeah, I'm hacking off the clauses after "prebuilt container".  Good call.

examples/s3gateway/README.md
```diff
@@ -0,0 +1,152 @@
...
+which specifies the `s3://` url to the bucket, and
+`-i` or `--inputpath`, 
+which specifies the local directory or mount point which will be the source of files to copy to the bucket.
+
+In `Dockerfile.tf_job_s3_gateway`, 
+that script is built into a Docker image
+along with the local folder `testdata`, 
+which contains the first chapters of the novels Moby Dick and A Tale of Two Cities.
+```
+$  ls -R testdata/
```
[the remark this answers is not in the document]
pappasilenus: I believe the latest cut of the README articulates this better.

examples/s3gateway/README.md
pappasilenus: See above.  I'm trying to keep this as simple as possible, so folks can run this as a test to see if their setup works., then look at the code to see how to use it themselves.  Copying to Pachyderm first allows everything to be self-contained and minimized.

examples/s3gateway/README.md
pappasilenus: The instructions for deploying Pachyderm with Kubeflow have us in the `pachyderm` namespace, so I want to keep it simple, but still provide information on customization.

examples/kubeflow/tfjob/README.md
```diff
@@ -136,14 +151,16 @@ Kubeflow installed with Pachyderm using the defaults for a joint deployment:
 Kubeflow in the `kubeflow` namespace and Pachyderm in the `pachyderm` namespace.
 
 1. Deploy Pachyderm and Kubeflow.
-2. Create a repo in Pachyderm using the command  `pachctl create repo testrepo`.
-3. Create a branch in that repo using the command `pachctl create branch testrepo@master`.
-4. Deploy the manifest to your Kubeflow using  `kubectl apply -f tf_job_s3_gateway.yaml`
-5. Using your Kubeflow TFJob dashboard or `kubectl`, monitor the TFJob pod created until it completes.
+2. Create a repo in Pachyderm using the command  `pachctl create repo inputrepo`.
+3. Add the mnist data to that repo using the command `pachctl put file inputrepo@master:/data -r -f mnist/`
```
JoeyZwicker: User should be told to either close this repo, grab the data from <URL where mnist data set is hosted by kubeflow>, or maybe we could put the data in a directory here and use the GitHub raw link with `put file -r`.

I think the 3rd option is the easiest if it works.

pappasilenus: I've clarified these instructions. I can't add the raw url until after the entire thing is committed. so going to merge this and then take a look.

PR #3975, opened 2019-08-06
Delete all should be paginated

adelelopez:
DeleteAll done in a single transaction will often have a transactions size exceeding etcd's limit. Since we're trying to delete everything, we're not as concerned about possible data races, and instead find it worth it to split it into separate transactions.


adelelopez:
This encountered a lot of CI failures on random tests. Since deleteAll is called from almost every test, it makes me concerned about actually merging it.

msteffen:

PR #3993, opened 2019-08-12
[WIP] Skaffold

chainlink:
Add skaffold config and new docker images for skaffold. Let me know if there's a better spot for those docs

### DIFF
--- a/Dockerfile-pachd-skaffold
+++ b/Dockerfile-pachd-skaffold
@@ -1,0 +1,25 @@
+# syntax = docker/dockerfile:experimental
+# Build the manager binary
+FROM golang:1.12.5 as builder
+
+WORKDIR /workspace
+
+COPY . .
+
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build \
+    -mod=vendor \
+    -installsuffix netgo \
+    -tags netgo \
+    -o pachd \
+    src/server/cmd/pachd/main.go
+
+FROM scratch
+LABEL maintainer="jdoliner@pachyderm.io"
+
+COPY --from=builder /workspace/pachd .
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+# ADD ca-certificates.crt /etc/ssl/certs/
+
+ENTRYPOINT ["/pachd"]
--- a/Dockerfile-worker-skaffold
+++ b/Dockerfile-worker-skaffold
@@ -1,0 +1,26 @@
+# syntax = docker/dockerfile:experimental
+# Build the manager binary
+FROM golang:1.12.5 as builder
+
+WORKDIR /workspace
+
+COPY . .
+
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build \
+    -mod=vendor \
+    -installsuffix netgo \
+    -tags netgo \
+    -o worker \
+    src/server/cmd/worker/main.go
+
+FROM ubuntu:16.04
+LABEL maintainer="jdoliner@pachyderm.io"
+
+WORKDIR /
+
+COPY --from=builder etc/worker/worker.sh /pach/
+COPY --from=builder /workspace/worker  /pach/
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+# ADD ca-certificates.crt /etc/ssl/certs/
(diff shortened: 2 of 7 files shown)


Dockerfile-pachd-skaffold
```diff
@@ -0,0 +1,25 @@
...
+    -tags netgo \
+    -o pachd \
+    src/server/cmd/pachd/main.go
+
+FROM scratch
+LABEL maintainer="jdoliner@pachyderm.io"
+
+COPY --from=builder /workspace/pachd .
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+# ADD ca-certificates.crt /etc/ssl/certs/
```
msteffen: is the commented line intentionally kept? Or is it left over?

Dockerfile-pachd-skaffold
```diff
@@ -0,0 +1,25 @@
+# syntax = docker/dockerfile:experimental
+# Build the manager binary
+FROM golang:1.12.5 as builder
+
+WORKDIR /workspace
+
+COPY . .
+
+RUN --mount=type=cache,target=/go/pkg/mod \
```
msteffen: This isn't a comment so much as a question—I've never seen `RUN` with no commands before. Does it use the `ENTRYPOINT` of the `golang:1.12.5` image or something? Is this effectively running `go build ... src/server/cmd/pachd/main.go`?

Finally, my slight preference would be to run `go build ... src/server/cmd/pachd` (i.e. build the whole package) if possible, just because I see no reason not to do it, but `go build ... src/.../pachd/main.go` only works if there are no other go files in the package. But then I'm not even sure that that's the command being run here...

Dockerfile-pachd-skaffold
```diff
@@ -0,0 +1,25 @@
+# syntax = docker/dockerfile:experimental
```
msteffen: Question on the file name: should this just replace the `Dockerfile.pachd` file that's already in our repo? Or is the goal to have both? Would it maybe make sense to put this in `etc/skaffold` or some such? I only ask because the name struck me as long, but maybe it makes sense.

Dockerfile-worker-skaffold
```diff
@@ -0,0 +1,26 @@
...
+
+FROM ubuntu:16.04
+LABEL maintainer="jdoliner@pachyderm.io"
+
+WORKDIR /
+
+COPY --from=builder etc/worker/worker.sh /pach/
+COPY --from=builder /workspace/worker  /pach/
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+# ADD ca-certificates.crt /etc/ssl/certs/
```
msteffen: Same question as above re: the commented out line. Is it intentional? Should we keep it?

Dockerfile-worker-skaffold
```diff
@@ -0,0 +1,26 @@
...
+COPY . .
+
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build \
+    -mod=vendor \
+    -installsuffix netgo \
+    -tags netgo \
+    -o worker \
+    src/server/cmd/worker/main.go
```
msteffen: Same suggestion as above re: `src/.../worker/main.go` vs `src/.../worker` (again, presuming the suggestion even makes sense)

k8s/local.yaml
```diff
@@ -0,0 +1,521 @@
+# pachctl deploy local --dry-run -o yaml
```
msteffen: Do you see any way we could generate this rather than checking it in (maybe generate it and store it in a tmp directory?). Maybe it's necessary, but I just ask because offhand this seems liable to get out of date.

skaffold-setup.md
```diff
@@ -0,0 +1,29 @@
...
+skaffold dev
+```
+
+## Notes / Gotchas
+- Skaffold will automatically detect the `docker-for-desktop` / `docker-desktop` contexts and not attempt to push an image. This can be overridden with:
+```
+build:
+  local:
+    push: false
+```
```
msteffen: where should this config be added (e.g. if this is being added to a file, what file?) to override the skaffold default?

chainlink: Can add this into the skaffold.yaml file directly

Dockerfile-pachd-skaffold
chainlink: Yeah, eventually it will replace the main docker file, I just want to scope out replacing our building container for the purpose of this pr (keep it scoped tightly to skaffold)

Dockerfile-pachd-skaffold
chainlink: This is actually all one long command (up to line 17), happy to change to build the package instead

Dockerfile-worker-skaffold
chainlink: Can remove

k8s/local.yaml
chainlink: yeah, I'm concerned about it getting out of date as well, in the interim, we could have people just generate manually

PR #4020, opened 2019-08-21
Overwrite context if it hasn't changed

### LINKED ISSUE #4017: Overwrite contexts
When `pachctl deploy` is done, a new pachyderm context is created, instead of overwriting the current active context. As a result, if a lot of deployments are done, a lot of contexts exist, which can be annoying. Instead, create a way to intelligently overwrite contexts, if no customizations to the context have been made.

ysimonson:
Closes #4017


ysimonson:
@Tryneus could you review? We discussed some alternative options, but they're not guaranteed and will, regardless, take a while. This would be nice in the short-term.

PR #4078, opened 2019-09-10
YAML Pipeline specs

msteffen:


src/server/pkg/ppsutil/util.go
```diff
@@ -259,20 +259,41 @@ func PipelineReqFromInfo(pipelineInfo *ppsclient.PipelineInfo) *ppsclient.Create
...
+//
+// Note that Go's json decoder is able to parse text that gopkg.in/yaml.v3
+// cannot (multiple json documents) so we currently attempt to parse a pipeline
+// with both parsers concurrently and keep whichever parse succeeds. Hopefully
+// multi-document support is added to our yaml parser and we can remove the json
+// parser.
 type PipelineManifestReader struct {
-	buf     bytes.Buffer
-	decoder *json.Decoder
+	ds []pipelineDecoder
```
adelelopez: I feel like `ds` should be something more descriptive, like `decoders`

src/server/pkg/ppsutil/util.go
```diff
@@ -283,31 +304,57 @@ func NewPipelineManifestReader(path string) (result *PipelineManifestReader, ret
...
 	var result ppsclient.CreatePipelineRequest
-	if err := jsonpb.UnmarshalNext(r.decoder, &result); err != nil {
-		if err == io.EOF {
-			return nil, err
+	retErr := NoParseErr // set to nil on 1st successful parse
+	dest := interface{}(&result)
+	// Call Decode() on every d.ds. After the first success, continue calling
+	// decode on other decoders (to keep them in sync) but throw other results
+	// away.
+	for i := 0; i < len(r.ds); {
```
adelelopez: this loop seems overly complicated for what it's trying to do... especially when we only ever expect two decoders

src/server/pkg/testutil/cmds.go
```diff
@@ -93,9 +87,23 @@ func BashCmd(cmd string, subs ...string) *TestCmd {
 	// Warn users that they must install 'match' if they want to run tests with
 	// this library, and enable 'pipefail' so that if any 'match' in a chain
 	// fails, the whole command fails.
+	// Note that to avoid SIGPIPE errors from derailing entire scripts, we
```
adelelopez: this seems to end abruptly

src/testing/match/match.go
```diff
@@ -35,40 +37,46 @@ func main() {
...
-			exit = 1
-		default: // Success, match -- copy input to output
-			exit = 0
+			os.Exit(1)
+		default:
+			matchFound = true // match found on this line
 		}
 	}
-	if exit >= 0 {
+	if matchFound || *invert {
```
adelelopez: not a big deal, but i feel like what `invert` is actually doing is very straightforward relative to the somewhat confusing if or else if not logic here

src/server/pkg/ppsutil/util.go
msteffen: I tried to shrink the code a bit by inlining comments and such, but I'm not quite sure what an overall shorter way to do this might be. Maybe the error semantics are too complicated: my goal was to return the error from the last parser to fail (so JSON specs get JSON errors and YAML specs get YAML errors, instead of something silly like "your YAML spec is not JSON" when the real problem is that the YAML spec has a tiny error".

When I just store e.g. `jsonDecoder` and `yamlDecoder`, it ended up being the same number of lines (after inlining comments) and IMO more confusing, but if you prefer it I'm happy to go with it:
```
func (r *PipelineManifestReader) NxtCreatePipelineRequest() (*ppsclient.CreatePipelineRequest, error) {
	if d.jsonErr != nil && d.yamlErr != nil {
		return errors.New("no decoders are able to parse pipeline spec")
	}
	var result ppsclient.CreatePipelineRequest
	dest := interface{}(&result)
	if d.jsonErr != nil {
		d.jsonErr = d.jsonDecoder.Decode(dest)
		if d.yamlErr == nil && d.jsonErr == nil {
			dest = &map[string]interface{}{} // deposit YAML parse into garbage map
		} else if d.yamlErr != nil && d.jsonErr != nil {
			// yaml parse failed earlier--json err is most recent
			if d.jsonErr == io.EOF {
				return io.EOF
			}
			return nil, fmt.Errorf("malformed pipeline spec: %s", d.jsonErr)
		}
	}
	if d.yamlErr != nil {
		d.yamlErr = d.yamlDecoder.Decode(dest)
		if d.jsonErr != nil && d.yamlErr != nil {
			if d.yamlErr == io.EOF {
				return io.EOF
			}
			return nil, fmt.Errorf("malformed pipeline spec: %s", d.yamlErr)
		}
	}

	return nil, fmt.Errorf("malformed pipeline spec: %s", retErr)
	return &result, nil
}
```

src/server/pkg/testutil/cmds.go
msteffen: Meant to remove this line along with the ones that used to be below it :-).

src/testing/match/match.go
msteffen: I agree that it's a little confusing. I tried to clarify w/ the comment in the if body, but if you see a better way I'm happy to go with it.

src/server/pkg/ppsutil/util.go
msteffen: Done

src/server/pkg/ppsutil/util.go
msteffen: Another idea I had which is pretty hacky but might simplify the code is to look for the first non-space character, and if it's `{` use the JSON parser, otherwise use the YAML parser?

src/testing/match/match.go
adelelopez: alright, i think this is good! -- i tried to adjust it but i couldn't think of a cleaner way

src/server/pkg/ppsutil/util.go
adelelopez: I think the new version is alright

JoeyZwicker:
@msteffen Are there docs to go with this? 

If not, can you open a docs issue describing the major changes so docs team can use that as a baseline for testing and writing up a first pass?

msteffen:
@JoeyZwicker Just filed https://github.com/pachyderm/pachyderm/issues/4112

PR #4084, opened 2019-09-13
Debug TestExtractRestore

msteffen:
Round N+1!

Creating this PR because I've only ever seen this error in CI


msteffen:
Test is fixed by https://github.com/pachyderm/pachyderm/pull/4132

PR #4089, opened 2019-09-13
Restructure combining merging

svekars:
- Edit combining data
- Edit ML workflows
- Edit example developer workflow
- Edit time windows
- Edit ingressing from diff cloud
- Added new diagrams


doc/cookbook/example_developer_workflow.md
```diff
@@ -1,49 +1,74 @@
...
+provenance and scalable processing to data
+scientists and engineers. You can make it even
+more powerful by integrating it with your existing
+continuous integration and continuous deployment (CI/CD)
+workflows and systems. This section walks you through the
+CI/CD processes that you can create to enable Pachyderm
+collaboration within your data science and engineering groups.
 
+As you write code, you test it in containers and
+notebooks against sample data that you store in Pachyderm repos.
```
JoeyZwicker:
```suggestion
notebooks against sample data that you store in Pachyderm repos or download/mount locally.
```

doc/cookbook/example_developer_workflow.md
```diff
@@ -1,49 +1,74 @@
...
+   Your CI process must automatically start the build of a Docker container
+   image based on your code and the Dockerfile.
 
-Your CI process should automatically kick off the build of an docker container image based on your code and the DOCKERFILE. That image will be used in the next step.
+3. Push the image tagged with commit ID to an image registry.
 
-### 3. Push to registry tagged with commit id
+   Your CI process pushes a Docker image created in Step 2 to your preferred
+   image registry. When you build your image by using  the `--built` flag with
+   the `pachctl update pipeline` command, Pachyderm tags the image with the
```
JoeyZwicker: > the `pachctl update pipeline` command, Pachyderm tags the image with the
   Git commit SHA.

This is incorrect. Best practices in CI is to use the commit sha. --build just gives it a random UUID tag to the image. 

**Also, your mixing up two concepts here -- you can't use --build AND use CI. Those are doing the same thing.** The `--build` flag is a flag to `pachctl update pipeline` so I'm not sure why that's being mentioned here. We can discuss in TOH or chat with @pappasilenus , but these are two distinct options. 

1. Users can develop locally and then use `update pipeline --build` which will build the image, tag it with a random UUID, push to a registry, AND update the pipeline to use that new image. This works great for the individual developer workflow because its one cmd that does it all. That said, it also requires your data scientists to have docker knowledge and have it all installed and running. It also is more limiting because it's all being done by pachyderm so there is less functionality/control than using CI.

OR

2. Use Ci to automate pipeline updates. Data scientist just pushes code to Git -- doesn't need to think about anything else. CI then does the above process here -- uses dockerfile to build, tag, and push container, and then update the json and makes an API call to pachyderm to update the pipeline. Generally speaking, using CI is much more heavy-weight process but it has a few advantages:
- data eng abstracts away lots fo details for data scientists who just want to push code and not think about all the docker/pachyderm details
- they can run other CI test on the code before pushing
- they have more control over naming/tagging schemes such as using the Git commit SHA
- probably others I'm not thinking of right now

doc/cookbook/example_developer_workflow.md
```diff
@@ -1,49 +1,74 @@
...
 
-### 4. 'update pipeline' using template with tagged image
+   In this step, your CI/CD infrastructure uses the `pipeline.json`
+   specification that you have checked in and fills in the Git commit
+   SHA for the version of the image that must be used in this pipeline.
+   Then, it runs the `pachctl update pipeline` command to push the
+   updated pipeline specification to Pachyderm.
 
-In this step, your CI/CD infrastructure would use the pipeline.json template that was checked in, and fill in the git commit SHA for the version of the image that should be used in this pipeline.  It will then use the ``pachctl update pipeline`` command to push it to pachyderm.
+5. Pull a tagged image from a registry
```
JoeyZwicker: This is not really a step. After CI calls update pipeline, this is jsut want pachyderm does. Probably should be combined with step 4 as just the "result" of that action.

PR #4093, opened 2019-09-16
Pachctl system certs

kevindelgado:


kevindelgado:
Passes CI now, PTAL @msteffen

kevindelgado:
friendly ping @msteffen

etc/deploy/gen_pachd_tls.sh
```diff
@@ -12,6 +12,7 @@ else
 fi
 
 eval "set -- $( getopt -l "dns:,ip:,port:" -o "o:" "--" "${0}" "${@:-}" )"
+echo "${@}"
```
msteffen: I think we can probably remove this echo (this was just because I didn't understand how getopt worked. IIRC this script still doesn't work on macs—at some point we should rewrite it in go or some such)

src/client/client.go
```diff
@@ -245,6 +260,9 @@ func WithAdditionalPachdCert() Option {
 func getCertOptionsFromEnv() ([]Option, error) {
 	var options []Option
 	if certPaths, ok := os.LookupEnv("PACH_CA_CERTS"); ok {
+		if pachdAddress, ok := os.LookupEnv("PACHD_ADDRESS"); !ok || !strings.HasPrefix(pachdAddress, "grpcs") {
```
msteffen: Nit: should we put "grpcs" in a constant at the top of the file?

src/client/client.go
```diff
@@ -245,6 +260,9 @@ func WithAdditionalPachdCert() Option {
...
+			return nil, errors.New("cannot set PACH_CA_CERTS without setting PACHD_ADDRESS to grpcs://... ")
```
msteffen: If we factor "grpcs" into a constant, I think we should reference it in the error so that the error stays up to date.

src/client/client.go
```diff
@@ -275,10 +293,16 @@ func getCertOptionsFromEnv() ([]Option, error) {
 // environment variables, config files, etc to figure out which address a user
 // running a command should connect to.
 func getUserMachineAddrAndOpts(context *config.Context) (string, []Option, error) {
+	var options []Option
+
 	// 1) PACHD_ADDRESS environment variable (shell-local) overrides global config
 	if envAddr, ok := os.LookupEnv("PACHD_ADDRESS"); ok {
+		if strings.HasPrefix(envAddr, "grpcs://") {
```
msteffen: As above, I think we should use the constant here (and below)

src/client/client.go
```diff
@@ -288,7 +312,16 @@ func getUserMachineAddrAndOpts(context *config.Context) (string, []Option, error
 	}
 
 	// 2) Get target address from global config if possible
-	if context != nil && context.PachdAddress != "" {
+	if context != nil && (context.ServerCAs != "" || context.PachdAddress != "") {
+		// Proactively return an error in this case, instead of falling back to the default address below
+		if context.ServerCAs != "" && !strings.HasPrefix(context.PachdAddress, "grpcs") {
```
msteffen: use constant here (and below)

src/client/client.go
kevindelgado: I decided to use "grpcs://" because even the instances when we just use "grpcs" should also be using "grpcs://

kevindelgado:
feedback addressed

PR #4110, opened 2019-09-19
Fix bug in pipeline controller

msteffen:


msteffen:
Deleting this PR because I changed my local branch. I'll recreate it if I want to submit

PR #4111, opened 2019-09-19
updating code and dockerfile to include pachyderm python client

Nick-Harvey:


ysimonson:
I just went ahead and reviewed this (only the changes, not the existing kubeflow stuff.) Looks good!

PR #4134, opened 2019-09-26
Rearrange tests and break up auth tests

msteffen:
- Auth tests are now broken up into two buckets, so that `BUCKET=MISC` doesn't take quite so long
- All of the `test-*-cmds` test targets, along with the auth cmds and pps cmds tests have been unified under a single `test-cmds` build target. This is so that that the PPS and auth cmds tests aren't run in every auth/pps bucket, and because the various ` cmds` tests all have the same dependencies
- The code in `travis.sh` that runs a subset of the PPS tests has been refactored into `test_bucket` (which is now used for both PPS and auth tests)


Makefile
```diff
@@ -455,31 +456,26 @@ enterprise-code-checkin-test:
...
-	  go test -v ./src/server/pps/cmds -count 1 -timeout $(TIMEOUT)
+	  go test -v ./src/server -parallel 1 -count 1 -timeout $(TIMEOUT) $(RUN)
+
+test-cmds:
+	go install -v ./src/testing/match
+	go test -v ./src/server/pkg/deploy/cmds -count 1 -timeout $(TIMEOUT)
+	go test -v ./src/server/pfs/cmds -count 1 -timeout $(TIMEOUT)
+	go test -v ./src/server/pps/cmds -count 1 -timeout $(TIMEOUT)
+	@# TODO(msteffen) does this test leave auth active? If so it must run last
+	go test -v ./src/server/auth/cmds -count 1 -timeout $(TIMEOUT)
```
ysimonson: This is my bad, both for not suffixing the package with `cmds`, and for naming the test incorrectly, but would you mind adding config cmd tests here? The current target is [`test-config`](https://github.com/pachyderm/pachyderm/blob/30fedffbaccbac64427eb4bee4a8a2c491b04d13/Makefile#L464-L465).

ysimonson:
One small issue, but otherwise lgtm

.travis.yml
```diff
@@ -14,11 +14,14 @@ env:
     - PATH=$HOME/cached-deps:$PATH
   matrix:
     - BUCKET=MISC
+    - BUCKET=AUTH1
+    - BUCKET=AUTH2
     # If you want to update the number of PPS workers, you'll also need to update
     # the value in ./etc/build/PPS_BUILD_BUCKET_COUNT
```
ysimonson: is this still the case?

msteffen: Ah, good catch. I updated the comment, and also set the total number of PPS and auth buckets in env vars in the same file (right above here) so it's easier to update

Makefile
msteffen: Done

PR #4138, opened 2019-09-26
Clarify local installation

### LINKED ISSUE #4137: Fix the documentation regarding steps to follow for local install
In the documentation, under Getting Started/Local Installation the steps outlined are as follows (my case, MacOS):

- Install Docker Desktop/Kubernetes (OK)

- Install pachctl via brew (OK)

- From the docs:

> Verify that installation was successful by running pachctl version:
```
$ pachctl version
COMPONENT           VERSION
pachctl             1.9.1
```

**This fails.** 
It times out trying to connect to the pachd server, which isn't running. Nowhere in the docs is the daemon even mentioned, not even in the troubleshooting section.

You've omitted a step. It should be : 

> Start the pachd server: 

```
$ pachctl deploy local
```

Then

>Verify that installation was successful by running pachctl version:

```
$ pachctl version
COMPONENT           VERSION
pachctl             1.9.5
pachd               1.9.5

```

Also, the troubleshooting guide has this section

> Cannot connect via pachctl - context deadline exceeded
> Symptom
> 
> You may be using the pachd address config value or environment variable to specify how pachctl talks to your Pachyderm cluster, or you may be forwarding the pachyderm port. 

...
> Recourse
> It’s possible that the connection is just taking a while. Occasionally this can happen if your cluster is far away (deployed in a region across the country). Check your internet connection.

This cannot connect error also pops up when pachd is simply not running, i.e. no deployment to the cluster. Probably it's no biggie, but easy to fix and might save some time trying the thing out.

svekars:
closes #4137

### DIFF
--- a/doc/getting_started/local_installation.md
+++ b/doc/getting_started/local_installation.md
@@ -105,6 +105,9 @@ to have pachctl installed on your machine by following these steps:
    pachctl             1.9.5
    ```
 
+   If you run `pachctl version` without `--client-only`, the command times
+   out. This is expected behavior because `pachd` is not yet running.
+
 ## Deploy Pachyderm
 
 After you configure all the [Prerequisites](#prerequisites),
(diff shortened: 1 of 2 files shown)


svekars:
@JoeyZwicker can you please review/approve?

PR #4147, opened 2019-09-28
Move to mkdocs

svekars:
Attempting to merge into a feature branch.


JoeyZwicker:
are there any content changes here or just doc infra? I've got no experience or ability to review your docs infra changes so you can just self-approve those into your side branch.

svekars:
There are some, but I suggest that we don't focus on them. I’ll self approve/merge.

PR #4165, opened 2019-10-03
[WIP] Push docker images after a successful test run

ysimonson:


ysimonson:
Closing this. It looks like TravisCI runs deploy targets _per job_ rather than _per build_. This is problematic because:
1. it'll push buggy pachd images that pass some but not all jobs (which is true for most buggy pachd images)
2. it'll redundantly push images to docker hub

PR #4167, opened 2019-10-04
Fix pachctl TLS bug and add basic TLS test

msteffen:


etc/testing/test_tls.sh
```diff
@@ -0,0 +1,33 @@
...
+
+pachctl version
+# Don't log our activation code when running this script in Travis
+pachctl enterprise activate "$(aws s3 cp s3://pachyderm-engineering/test_enterprise_activation_code.txt -)" && echo
+
+set -x
+
+# Split PACHD_ADDRESS into host and port
+IFS=: read -a addr_parts <<<"${PACHD_ADDRESS:-0.0.0.0:30650}"
+unset PACHD_ADDRESS # use config value set by gen_pachd_tls.sh
```
ysimonson: Where does this get set in the first place?

`PACHD_ADDRESS` will be removed at some point, so it might be worth just tackling that now.

src/server/config/cmds.go
```diff
@@ -189,35 +189,54 @@ func Cmds() []*cobra.Command {
...
+	var serverCAs string
 	var namespace string
-	updateContext := &cobra.Command{
+	var updateContext *cobra.Command // standalone declaration so Run() can refer
+	updateContext = &cobra.Command{
 		Short: "Updates a context.",
-		Long:  "Updates an existing context config from a given name.",
-		Run: cmdutil.RunCmdFixedArgs(1, func(cmd *cobra.Command, args []string) (retErr error) {
+		Long: "Updates an existing context config from a given name, or the " +
+			"currently-active context, if no name is given).",
```
ysimonson:
```suggestion
			"currently-active context, if no name is given.",
```

ysimonson: closing paren but no opening paren

src/server/config/cmds.go
```diff
@@ -226,7 +245,8 @@ func Cmds() []*cobra.Command {
 	}
 	updateContext.Flags().StringVar(&pachdAddress, "pachd-address", "", "Set a new name pachd address.")
 	updateContext.Flags().StringVar(&clusterName, "cluster-name", "", "Set a new cluster name.")
-	updateContext.Flags().StringVar(&authInfo, "auth-info", "", "Set a new auth info.")
+	updateContext.Flags().StringVar(&authInfo, "auth-info", "", "Set a new k8s auth info.")
+	updateContext.Flags().StringVar(&serverCAs, "server-cas", "", "Set new trusted CA certs")
```
ysimonson:
```suggestion
	updateContext.Flags().StringVar(&serverCAs, "server-cas", "", "Set new trusted CA certs.")
```

etc/testing/test_tls.sh
msteffen: I use it when testing locally :joy: . Travis uses the default value. Would you be okay with leaving it in until I'm forced to find a new way to work locally?

I way I could imagine seeing this work longer term is to add a pachctl command like `pachctl address` or some such that spits out the pachyderm address that it's connected to, and then having this script use that to avoid having to copy the address-determining logic that's already in `client.go`. How would you feel about taking that approach, but doing it in another PR?

src/server/config/cmds.go
msteffen: :+1: fixed

etc/testing/test_tls.sh
ysimonson: Yeah makes sense to wait.

I'm not sure if I see much utility to a `pachctl address`, but this is may very well be lack of familiarity with parts of the testing infrastructure.

ysimonson: Also, for posterity, I'm wrong about `PACHD_ADDRESS` being definitely removed at some point. It's not marked for deprecation yet, though it may be in the future. Given how disruptive its removal would be, though, it's safe to say that it's at least going to be around for the foreseeable future.

The reasoning is that contexts provide an alternative, and, while `PACHD_ADDRESS` is convenient for some usecases, it's also dangerous: it makes it trivial to send one cluster's credentials to another cluster.

etc/deploy/restart_with_tls.sh
```diff
@@ -18,44 +20,40 @@ while true; do
...
-# I'm not sure why this is necessary -- pachd should communicate with itself via an unencrypted, internal port
-# Empirically, though, the new pachd pod crashloops if I don't do this (2018/6/22)
-kubectl get deploy/pachd -o json | jq '.spec.replicas = 0' | kubectl apply -f -
+echo "Turning down any existing pachyderm cluster..."
+if pachctl version --timeout=5s; then
+  # Turn down old pachd deployment, so that new (TLS-enabled) pachd doesn't try
+  # to connect to old, non-TLS pods I'm not sure why this is necessary -- pachd
+  # should communicate with itself via an unencrypted, internal port.
+  #
+  # Empirically, though, new pachd pods crashloop if I don't do this (2018/6/22)
```
ysimonson: This sounds pretty fishy. Could you file a ticket for it, just so we don't lose track of this issue?

msteffen: Done: https://github.com/pachyderm/pachyderm/issues/4173

etc/testing/test_tls.sh
msteffen: That makes sense.

I don't feel any particular urgency to make this change, but I guess in case we decide to come back to it later, my idea with `pachctl address` was that there's a block of reasonably complex logic in `client.go` (in [getUserMachineAddrAndOps](https://github.com/pachyderm/pachyderm/blob/master/src/client/client.go#L300)) that determines which address `pachctl` is currently connected to. If the purpose of `gen_pachd_tls.sh` is to generate a self-signed TLS cert for your current pach cluster, then it could either duplicate the logic in `getUserMachineAddrAndOps` or ask `pachctl` to execute that logic and print the result. `pachctl address` would do the latter.

There are a lot of alternatives we could do, though—I don't feel at all wedded to this one. We could have `pachctl` generate the self-signed cert, or decide that we don't like self-signed certs and use some other PKI for setting up TLS

Issue #4166, opened 2019-10-03
`pachctl config set context -k` omits `"namespace"` field when it is empty


bpeabody:
It should be populated with `"default"`. When this field is missing `pachctl port-forward` does not work.

You can work around the problem by manually adding `"namespace": "default"` to `.pachyderm/config.json`.

pappasilenus:
@ysimonson Two choices, I think:

1. pachyderm contexts assume "default" when no namespace is specified in a config (the same as kubeconfig behavior)
2. `set context` automatically sets the "namespace" property in a context when importing from kubeconfig contexts with no namespace

I'm inclined to mimic the kubectl behavior, but it seems to me that there is an argument for making it easier to use  and precise by *requiring* a namespace spec in a context.

ysimonson:
Thanks @bpeabody! Fixed in #4170. In the mean-time, you should be able to work around this by doing `pachctl port-forward --namespace=default`.

PR #4202, opened 2019-10-09
[DOC] Add import k8s context

### LINKED ISSUE #4163: Create a doc that describes importing kubernetes contexts
based on https://github.com/pachyderm/pachyderm/pull/4152

svekars:
Closes #4163

### DIFF
--- a/doc/deployment/import-kubernetes-context.md
+++ b/doc/deployment/import-kubernetes-context.md
@@ -76,7 +76,8 @@ To import a Kubernetes context, complete the following steps:
    {
      "source": "IMPORTED",
      "cluster_name": "minikube",
-     "auth_info": "minikube"
+     "auth_info": "minikube",
+     "namespace": "default"
    }
    ```
 
--- a/doc/deployment/import-kubernetes-context.md
+++ b/doc/deployment/import-kubernetes-context.md
@@ -1,0 +1,114 @@
+# Import a Kubernetes Context
+
+**Note:** The steps in this section apply to your configuration only
+if you deployed Pachyderm from a manifest created by the `pachctl deploy`
+with the `--dry-run` flag. If you did not use the `--dry-run` flag, skip this
+section.
+
+When you run the `pachctl deploy` command with `--dry-run` flag for any
+underlying infrastructure, such as GKE, Amazon, local, or any other, you do
+not create an actual Pachyderm cluster, Instead you create a Kubernetes
+deployment manifest that you can later use to deploy a Pachyderm cluster.
+
+You can use that manifest with a standard `kubectl apply` command to deploy
+Pachyderm. For example, if you have created a manifest called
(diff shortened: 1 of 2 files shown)


doc/deployment/import-kubernetes-context.md
```diff
@@ -0,0 +1,114 @@
...
+1. Create a new Pachyderm context with the embedded Kubernetes context:
+
+   ```bash
+   $ pachctl config set context <new-pachyderm-context> -k `kubectl config current-context`
+   ```
+
+   **Example:**
+
+   ```bash
+   $ pachctl config set context test-context -k `kubectl config current-context`
```
JoeyZwicker: I dont think you need this example and the above. 

pachctl config set context <new-pachyderm-context> -k `kubectl config current-context` is pretty c;ear without needing to show an example where you replace one arugment

doc/deployment/import-kubernetes-context.md
```diff
@@ -0,0 +1,114 @@
...
+1. Verify that the new context has been activated:
+
+   ```bash
+   $ pachctl config get active-context
+   ```
+
+   **Example:**
+
+   ```bash
+   $ pachctl config get active-context
```
JoeyZwicker: same as above, we don't need this pattern of showing a command AND showing an example that is exactly or nearly the cmd but with output. just one is fine. We only need to include both when there are multiple arguments that might not be clear or it's a cmd that we expect the user to copy/paste and use as part of a tutorial

doc/deployment/import-kubernetes-context.md
svekars: I can remove them, but overall they don't hurt, IMO.

doc/deployment/import-kubernetes-context.md
svekars: removed

doc/deployment/import-kubernetes-context.md
```diff
@@ -0,0 +1,93 @@
...
+1. Verify that the context was successfully created and view the context parameters:
+
+   **Example:**
+
+   ```bash
+   $ pachctl config get context test-context
+   {
+     "source": "IMPORTED",
+     "cluster_name": "minikube",
+     "auth_info": "minikube"
```
pappasilenus: I believe this should also have this line:

"namespace": "default",

When https://github.com/pachyderm/pachyderm/pull/4170 is merged.

Until then, maybe the doc should use the workaround, and tell folks to specify namespaces?

https://github.com/pachyderm/pachyderm/issues/4166#issuecomment-538437301

svekars: It looks like the #4170 will be merged shortly. I'll wait for that to happen and will add the namespace field once that happens. I don't think it's worth it to add a workaround only to change it in a day or so.

svekars:
Since #4170 was merged, updating accordingly and it should be ready to merge.

PR #4208, opened 2019-10-10
More structured parsing of pachd address

### LINKED ISSUE #4181: PACHD_ADDRESS or context config can contain grpcs:// but not grpc://
**What happened?**:
I put grpc:// in the PACHD_ADDRESS
Connection as usual
put grpc://address_to_pachd:port in your PACHD_ADDRESS or in a context
```
could not connect to pachd at "grpc://***:650": address shouldn't contain protocol ("://"), but is: "grpc://***:650"

but with grpcs there are no complaints
```

Mostly, this just seems inconsistent. Minor bug.

ysimonson:
Add support for a couple of new corner cases with pachd address parsing:

* Addresses including `grpc://` (fixes #4181)
* Addresses where the hostname has a colon, but no port is specified (e.g. `[::1]`)

Generally, though, this change makes parsing more structured, with an eye towards bug prevention and maintainability:

* Pachd address parsing and error checking has been moved out of `client.go` into its own heavily tested struct
* Use of `net/url` for parsing when a scheme is included in the address (it is not used otherwise, as `net/url` does not handle addresses without a scheme well.)
* In contexts, `pachd_address` is set at config write time. This will allow us to simplify tests (see discussion [here](https://github.com/pachyderm/pachyderm/pull/4188#discussion_r333303849) and [here](https://github.com/pachyderm/pachyderm/pull/4188#pullrequestreview-300119573).)

Open questions:

* ~Should the pachd address parsing logic be in `grpcutil`? It seems like a somewhat awkward, but I didn't see a client package that was a more obvious fit, and this doesn't seem worthy of having its own package.~

~Blocked by #4225~


ysimonson:
@msteffen this is about ready for review. There are a lot of changes to the TLS scripts; part of these are simplifications facilitated by these changes, but most of the changes were made so that they are runnable on macOS.

etc/deploy/gen_pachd_tls.sh
```diff
@@ -66,17 +14,13 @@ distinguished_name = dn
 x509_extensions    = exn    # Since we're making self-signed certs. For CSRs, use req_extensions
 
 [ dn ]
-CN = ${dns:-localhost} # TODO(msteffen) better default domain name
+CN = localhost # TODO(msteffen) better default domain name
```
ysimonson: I still need to fix, will do soon

src/client/client.go
```diff
@@ -296,59 +297,62 @@ func getCertOptionsFromEnv() ([]Option, error) {
...
+		pachdAddress, err := grpcutil.ParsePachdAddress(context.PachdAddress)
+		if err != nil {
+			return nil, nil, fmt.Errorf("could not parse the active context's pachd address: %v", err)
+		}
+
 		// Proactively return an error in this case, instead of falling back to the default address below
-		if context.ServerCAs != "" && !strings.HasPrefix(context.PachdAddress, grpcs) {
-			return "", nil, fmt.Errorf("must set pachd_address to %s... if server_cas is set", grpcs)
+		if context.ServerCAs != "" && !pachdAddress.Secured {
+			return nil, nil, errors.New("must set pachd_address to grpc://... if server_cas is set")
```
msteffen: I think `!pachdAddress.Secured` implies that the prefix is probably `grpc://` already (in other words, I think the error really should be `must set pachd address to grpcs://...`)

src/client/pkg/grpcutil/addr.go
```diff
@@ -0,0 +1,127 @@
...
+	Secured bool
+	// Host specifies the pachd address host without the port
+	Host string
+	// Port specifies the pachd port
+	Port uint16
+}
+
+// ParsePachdAddress parses a string into a pachd address, or returns an error
+// if it's invalid
+func ParsePachdAddress(value string) (PachdAddress, error) {
```
msteffen: Should this return a pointer? I just ask because that's what `getUserMachineAddrAndOpts` returns.

src/client/pkg/grpcutil/addr.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+		if u.RawQuery != "" {
+			return PachdAddress{}, errors.New("pachd address should not include a query string")
+		}
+		if u.Fragment != "" {
+			return PachdAddress{}, errors.New("pachd address should not include a fragment")
+		}
+		if u.Path != "" {
+			return PachdAddress{}, errors.New("pachd address should not include a path")
+		}
```
msteffen: Would it make sense to compress these `if` blocks into a switch?

src/client/pkg/grpcutil/addr.go
```diff
@@ -0,0 +1,127 @@
...
+		secured = u.Scheme == "grpcs"
+	}
+
+	// port always starts after last colon, but net.SplitHostPort returns an
+	// error on a hostport without a colon, which this might be
+	colonIdx := strings.LastIndexByte(value, ':')
+	host := value
+	port := uint16(DefaultPachdNodePort)
+	if colonIdx >= 0 {
+		maybePort, err := strconv.ParseUint(value[colonIdx+1:], 10, 16)
```
msteffen: would it make sense to call net.SplitHostPort here? Just because it seems like it might be shorter

src/client/pkg/grpcutil/addr.go
```diff
@@ -0,0 +1,127 @@
...
+		// only parse the url if it contains a scheme, as net/url doesn't
+		// appropriately handle values without one
+
+		u, err := url.Parse(value)
+		if err != nil {
+			return PachdAddress{}, fmt.Errorf("could not parse pachd address: %v", err)
+		}
+
+		if u.Scheme != "grpc" && u.Scheme != "grpcs" {
+			return PachdAddress{}, fmt.Errorf("unrecognized scheme in pachd address: %s", u.Scheme)
```
msteffen: Do you think it would make sense to allow `http` and `https`? I think technically GRPC is layered on HTTP2. Also, if we don't allow `http`, perhaps the error should specify that the scheme must be `grpc` or `grpcs` (I can just see somebody using `http://whatever` and being like "what do you mean unrecognized scheme?")

src/client/client.go
```diff
@@ -423,42 +428,29 @@ func NewOnUserMachine(prefix string, options ...Option) (*APIClient, error) {
...
-					"seem to be host:port): %v", addr, err)
+					"%d or %d, but is currently set to %d - is this right?): %v", grpcutil.DefaultPachdNodePort, grpcutil.DefaultPachdPort, pachdAddress.Port, err)
+			} else if fw == nil && pachdAddress.IsLoopback() {
+				return nil, fmt.Errorf("could not connect (note: address %s looks "+
+					"like loopback, try unsetting it): %v",
+					pachdAddress.Qualified(), err)
 			}
 		}
-		return nil, fmt.Errorf("could not connect to pachd at %q: %v", addr, err)
+		return nil, fmt.Errorf("could not connect to pachd at %s: %v", pachdAddress.Qualified(), err)
```
msteffen: I'm not too bothered either way, but why change the formatting directive? Do the quotes look bad?

In the past I've favored `%q` over `%s` because it shows accidental leading and trailing spaces in user input, but I'd be happy to start doing the opposite

src/client/client.go
```diff
@@ -296,59 +297,62 @@ func getCertOptionsFromEnv() ([]Option, error) {
 // getUserMachineAddrAndOpts is a helper for NewOnUserMachine that uses
 // environment variables, config files, etc to figure out which address a user
 // running a command should connect to.
-func getUserMachineAddrAndOpts(context *config.Context) (string, []Option, error) {
+func getUserMachineAddrAndOpts(context *config.Context) (*grpcutil.PachdAddress, []Option, error) {
```
msteffen: Should this return a literal `PachdAddress`? (or should `ParsePachdAddress` return a pointer? See below)

src/client/pkg/grpcutil/addr.go
ysimonson: Sorry, I'm not following. How would I turn this into a switch, since I'm checking different fields in each case?

src/client/pkg/grpcutil/addr.go
ysimonson: It doesn't work, see the comment above this block

src/client/client.go
ysimonson: TBH, I think I just changed this from muscle memory. Happy to stick to `%q`.

src/client/client.go
ysimonson: Given it's just one value, I don't have an opinion either way, and just changed it to use pointers everywhere. I'm not sure if there's some sort of official go guidelines on when to use values vs pointers (if there are please do let me know), but I tend to lean on values when I can because:

1. It reduces the chances of panics from null pointer dereferencing.
2. It reduces the amount of pointer chasing and GC pressure.

src/client/pkg/grpcutil/addr.go
ysimonson: [Ah](https://github.com/golang/go/wiki/Switch#missing-expression). TBH, I didn't know this existed. That makes me slightly hesitant to use it (is it employed commonly enough for most golang developers to be familiar with it?), but I can give it a go.

ysimonson: I take it back, it's obvious enough what's going on. Will change this.

src/client/pkg/grpcutil/addr.go
msteffen: :sweat_smile:

src/client/client.go
msteffen: That makes sense. I found one doc in the golang repo on receiver type https://github.com/golang/go/wiki/CodeReviewComments#receiver-type but that involves slightly different considerations. Either way I think it's not a big deal.

ysimonson: Oh man, those heuristics are really helpful. Thanks!

src/client/pkg/grpcutil/addr.go
ysimonson: That relies on the URL always being parseable, but due to a separate issue, `url.Parse` doesn't handle values that don't include a protocol (which this is built to handle.)

msteffen: Ah, I'm suggesting adding a protocol at the beginning (the idea behind `if !strings.Contains(value, "://") { value = "grpc://" + value }`) precisely so that we *can* pass it to `url.Parse` in all cases, and then take advantage of its flexible port parsing, which handles cases like ipv6 and such.

Again, not trying to push this *too* hard, but I thought that adding the protocol might be easier than implementing a more flexible `net.SplitHostPort`, as the real implementation is fairly complex: https://golang.org/src/net/ipsock.go?s=4766:4832#L146

ysimonson: Oh, yeah I see that in your code now. OK I can make that change.

ysimonson: Done!

PR #4223, opened 2019-10-15
Add TTL to OTP request, + misc fixes

msteffen:


src/client/version/client.go
```diff
@@ -42,6 +42,15 @@ func PrettyPrintVersion(version *pb.Version) string {
 	return result
 }
 
+// This returns true if Pachyderm is at least at the given version. This allows
+// us to gate backwards-incompatible features on release boundaries.
+//
+// If version.IsAtLeast() is called anywhere in the code with an old version, it
+// can safely be assumed to be "true" and removed
```
ysimonson: I'm not sure what this means, is there a way to reword this?

src/client/version/client.go
```diff
@@ -42,6 +42,15 @@ func PrettyPrintVersion(version *pb.Version) string {
...
+func IsAtLeast(major, minor int) bool {
```
ysimonson: What about revision numbers? Do they not matter in this context?

src/server/auth/cmds/cmds.go
```diff
@@ -466,6 +478,10 @@ func GetAuthTokenCmd() *cobra.Command {
 	getAuthToken.PersistentFlags().BoolVarP(&quiet, "quiet", "q", false, "if "+
 		"set, only print the resulting token (if successful). This is useful for "+
 		"scripting, as the output can be piped to use-auth-token")
+	getAuthToken.PersistentFlags().StringVar(&ttl, "ttl", "", "if set, the "+
+		"resulting auth token will have the given lifetime (or the lifetime"+
+		"of the caller's current session, whichever is shorter). This flag shoul "+
```
ysimonson:
```suggestion
		"of the caller's current session, whichever is shorter). This flag should "+
```

src/server/auth/cmds/cmds.go
```diff
@@ -466,6 +478,10 @@ func GetAuthTokenCmd() *cobra.Command {
...
+		"be a golang duration (e.g. \"30s\" or \"1d2h3m4s\")")
```
ysimonson: what happens if the flag is not set?

src/server/auth/cmds/cmds.go
```diff
@@ -492,27 +508,46 @@ func UseAuthTokenCmd() *cobra.Command {
...
 				return grpcutil.ScrubGRPC(err)
 			}
 			fmt.Println(resp.Code)
 			return nil
 		}),
 	}
+	getOneTimePassword.PersistentFlags().StringVar(&ttl, "ttl", "", "if set, "+
+		"the resulting one-time password will have the given lifetime (or the "+
+		"lifetime of the caller's current session, whichever is shorter). This "+
+		"flag should be a golang duration (e.g. \"30s\" or \"1d2h3m4s\")")
```
ysimonson: likewise, what happens if this is not set?

src/client/version/client.go
msteffen: Sure, though I'm not sure what the best way to express the idea might be. Basically the idea is that if we're on Pachyderm 1.9.3, and you're a Pachyderm engineer and you see
```
if version.IsAtLeast(1, 9) {
  /* backwards-incompatible code */
}
```
then you can just replace this with
```
/* backwards-incompatible code */
```
This function is basically supposed to be a simple way to gate stuff (rather than what we've been doing, which is to file an issue for backwards-incompatible changes and hoping we remember to fix them between releasing the first RC for a new version and releasing the version itself).

Part of me thinks I should just get rid of the `If version.IsAtLeast()` paragraph, since the fact it's highlighting is pretty obvious, so it might be creating more confusion than it's clearing, but I'm curious what you think

src/client/version/client.go
msteffen: I came up with this function in the context of gating backwards-incompatible features. In that context I don't think revision numbers matter, just because revisions are our most granular releases, so there's no sense in gating a change on a revision vs just merging it.

I'm not opposed to adding a revision argument, though, if you think it could be useful

ysimonson: That makes sense

src/client/version/client.go
ysimonson: Yeah, I think the paragraph just adds confusion

src/server/auth/cmds/cmds.go
msteffen: :+1: Updated description

src/server/auth/cmds/cmds.go
msteffen: Updated description

src/server/auth/server/api_server.go
```diff
@@ -1694,25 +1743,30 @@ func (a *apiServer) GetAuthToken(ctx context.Context, req *auth.GetAuthTokenRequ
...
-		ttl, err = a.getCallerTTL(ctx)
+		ttl, err := a.getCallerTTL(ctx)
 		if err != nil {
 			return nil, err
 		}
-		if req.TTL == 0 || req.TTL > ttl {
+		if req.TTL <= 0 || req.TTL > ttl {
 			req.TTL = ttl
 		}
+	} else if version.IsAtLeast(1, 10) && req.TTL == 0 || req.TTL > defaultSessionTTLSecs {
```
ysimonson: I'm not sure what the precedence rules are here between `&&` and `||` - could you clarify with parens?

ysimonson:
One small thing left, otherwise lgtm

src/server/auth/server/api_server.go
msteffen: Done

src/server/auth/server/api_server.go
```diff
@@ -1646,19 +1697,50 @@ func (a *apiServer) SetACL(ctx context.Context, req *auth.SetACLRequest) (resp *
 	return response, nil
 }
 
+// authorizeNewToken is a helper for GetAuthToken and GetOTP. The code isn't too
+// long or complex, but centralizing it keeps the two API endpoints syncronized
```
ysimonson: nit: not clear what this function does

src/server/auth/server/api_server.go
```diff
@@ -1646,19 +1697,50 @@ func (a *apiServer) SetACL(ctx context.Context, req *auth.SetACLRequest) (resp *
...
+func (a *apiServer) authorizeNewToken(ctx context.Context, callerInfo *auth.TokenInfo, isAdmin bool, targetSubject string) (string, error) {
+	if targetSubject == "" {
+		// [Simple case] caller wants an implicit OTP for themselves
+		// Canonicalization: callerInfo.Subject is already canonical.
+		// Authorization: Getting a token/OTP for yourself is always authorized.
+		// NOTE: After this point, req.Subject is permitted to be ppsUser (even
+		// though we reject ppsUser when set explicitly
```
ysimonson:
```suggestion
		// though we reject ppsUser when set explicitly.)
```

src/server/auth/server/api_server.go
msteffen: Ah, whoops :-D. Fixed.

PR #4235, opened 2019-10-17
Fix port error message

msteffen:


msteffen:
There's some kind of bug in GitHub where it says that the travis build is in progress, but if you click through, it shows that all the tests passed. It doesn't make sense to me to worry about GH aesthetics, so I'm merging this PR.

PR #4253, opened 2019-10-24
More aggressive caching

msteffen:


Makefile
```diff
@@ -511,7 +511,7 @@ test-worker: launch-stats test-worker-helper
 
 test-worker-helper:
 	PROM_PORT=$$(kubectl --namespace=monitoring get svc/prometheus -o json | jq -r .spec.ports[0].nodePort) \
-	  go test -v ./src/server/worker/ -timeout $(TIMEOUT) -count 1
+	  go test -v -count=1 ./src/server/worker/ -timeout $(TIMEOUT)
```
ysimonson: just so I know, what's with moving all these args?

etc/testing/travis_cache.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Create all of our Travis cached dirs, or, if they exist, loosen their
+# permissions enough that our build process is guaranteed to be able to write
+# to them and Travis's caching process is guaranteed to be able to read them.
+# (Travis caching requires that all cached files be readable, and, I surmise,
+# that cached directores be executable, to be traversed).
```
ysimonson:
```suggestion
# that cached directories be executable, to be traversed).
```

Makefile
msteffen: Oh, it's just that I wrote a script to try and find all `go test` invocations that didn't have `-count=1`, and the way I did it, if it *did* have `-count 1` already it moved it to the front. I'd be happy to put these back to shrink the diff though.

msteffen: I'm not even sure we need all of this give that we also have `go clean -testcache` in `travis.sh` now, but I figured it'd be useful for running tests locally. I could just drop the commit that added all of these too

ysimonson: Actually the consistency is kinda nice. I do wonder if we should consider removing `-count=1` all-together, but that'd be a separate PR anyways.

PR #4257, opened 2019-10-24
Add robots txt

svekars:


ysimonson:
This is going to block all bots from crawling our site?

svekars:
@ysimonson, we will publish the docs prior to 1.9.8 to the docs-archive.pachyderm.com and we don't want them to be crawled. Everything after that will be crawled as we will have different settings in the master repo merged after that.

svekars:
@chainlink firendly ping.

chainlink:
So, it looks like robots.txt needs to be present on the root domain _and_ subdomains, so we'll need to put a bit more thought here.

svekars:
@chainlink I think with docs-archive.pachyderm.com, we will not really have subdomains but subfolders (i.e., /en/1.9.8), will we? Or did you mean the new website?

PR #4297, opened 2019-11-03
Update spf13/cobra

msteffen:
Replace removed `Out()` with the equivalent `OutOrStderr()`. To see that the
functions are the same, compare
old:
https://github.com/spf13/cobra/blob/1238ba19d24b0b9ceee2094e1cb31947d45c3e86/command.go#L145
new:
https://github.com/spf13/cobra/blob/77e4d5aecc4d34e58f72e5a1c4a5a13ef55e6f44/command.go#L294


src/server/cmd/pachctl/cmd/cmd.go
```diff
@@ -883,7 +883,7 @@ Use "{{.CommandPath}} [command] --help" for more information about a command.{{e
 		t := template.New("top")
 		t.Funcs(templateFuncs)
 		template.Must(t.Parse(text))
-		return t.Execute(cmd.Out(), cmd)
+		return t.Execute(cmd.OutOrStderr(), cmd)
 		return originalUsageFunc(cmd)
```
Tryneus: LGTM, although this line looks like dead code (probably my fault) - should be safe to remove, but I'm surprised this doesn't trigger lint.

PR #4312, opened 2019-11-07
Separate docs versions

msteffen:


doc/site/index.html
```diff
@@ -0,0 +1,9 @@
+<html>
+  <head>
+    <script type="text/javascript">
+      // redirect to /latest version of docs
+      window.location.pathname = "latest"
+    </script>
+  </head>
```
svekars: WIll this conflict with this: https://github.com/pachyderm/pachyderm/blob/master/netlify.toml#L6

msteffen: I don't know, but I now realize that it's at the very least unnecessary. I modified `serve.sh` to run `netlify dev`, which uses our `.netflify.toml` file instead of needing this.

msteffen:
Reference URL for looking at netlify logs:
https://app.netlify.com/sites/pachyderm-docs/deploys/5dc5c4b00d6dd40009757842

msteffen:
@Connor-Wallace I just wanted to put this PR on your radar because we're doing some janky HTML templating in [doc/build.sh](https://github.com/pachyderm/pachyderm/pull/4312/files#diff-0c979cac4288d9196f98279504c3c1cd), including some styling on our docs version selector that doesn't work on OS X (but does work on Linux).

I wanted to let you know because while I think this is probably fine for now, I think you might know the most about what a better process might look like, so I figured in case we do need to change this you'll have some context

PR #4322, opened 2019-11-08
Use vendor go mod flag on all calls to `go install`

ysimonson:
Otherwise pachyderm will build from the normal go mod directory, rather than our vendored directory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,6 @@ env:
     - PATH=${HOME}/cached-deps:${PATH}
     - PPS_BUCKETS=6
     - AUTH_BUCKETS=2
-    # Disable Go modules: we use go mod to track deps but keep everything in
-    # the top-level 'vendor' directory rather than 'go-get'ing all modules
-    # every time
-    - GO111MODULE=off
   matrix:
     - BUCKET=MISC
     # If you want to update the number of PPS or auth buckets, you'll neet to
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ cache:
     # etc/testing/travis_cache.sh
     - ${HOME}/.cache
     - ${HOME}/cached-deps
+    - ${HOME}/gopath/pkg/mod
 language: go
 go:
   - "1.13.1"
--- a/etc/testing/travis_cache.sh
+++ b/etc/testing/travis_cache.sh
@@ -11,6 +11,7 @@ set -ex
 dirs=(
   "${HOME}/.cache"
   "${HOME}/cached-deps"
+  "${HOME}/gopath/pkg/mod"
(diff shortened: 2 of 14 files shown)


ysimonson:
[`mod=vendor`](https://github.com/pachyderm/pachyderm/pull/4134/commits/31bf0fc8500d2d5ec51581fef744f9ad813c28da) was removed for a reason, so instead experimenting with entirely removing vendored libs: https://github.com/pachyderm/pachyderm/pull/4323

ysimonson:
@msteffen I'm still working on trying to remove vendoring all-together in a separate PR, but it's a big uphill battle. In the mean-time, this change has the merit of working. Maybe it makes sense to consider this change (at the very least as a stopgap to ensure CI and local dev act the same way), and get the vendoring removal merged when/if it's working. What do you think?

.travis.yml
```diff
@@ -8,6 +8,7 @@ cache:
     # etc/testing/travis_cache.sh
     - ${HOME}/.cache
     - ${HOME}/cached-deps
+    - ${HOME}/gopath/pkg/mod
```
msteffen: Would it make sense to use `$GOPATH` instead of `$HOME/gopath`?

etc/testing/travis_cache.sh
```diff
@@ -11,6 +11,7 @@ set -ex
 dirs=(
   "${HOME}/.cache"
   "${HOME}/cached-deps"
+  "${HOME}/gopath/pkg/mod"
```
msteffen: Similar to above: would it make sense to use `$GOPATH` instead of `$HOME/gopath`?

msteffen:
Sure! To be clear, my personal opinion is that I regard this to be equally good to `GO111MODULES=off` (I argued for switching b/c I was having trouble with `make lint`, but it looks like adding `$HOME/gopath/pkg/mod/cache` to `travis_cache.sh` fixed that), so if you think it's better, then I think we should do it. I just think if we *could* get rid of vendor, then that would be the strict improvement

One high-level question: do we want to add `-mod=vendor` to `go build` everywhere as well? My understanding is that otherwise we won't use the `vendor` dir for any `go build` commands (the PR where I switched it is here: https://github.com/pachyderm/pachyderm/pull/4134/files, so that could also be useful if you're trying to find all the places to switch back)

ysimonson:
Yeah that makes sense, made the change

ysimonson:
@msteffen I tried caching the entire gopath, but that leads to a problem: TravisCI clones the repo into $GOPATH, but also tries to pull the directory from cache. This leads to permissioning issues.

ysimonson:
Closing this out in favor of #4323

PR #4323, opened 2019-11-08
Remove vendor directory, in favor of non-vendored go modules

ysimonson:


pappasilenus:
There's this section of the main Makefile which seems to depend on the vendor directory, so this might break the build:

```
docker-build-kafka:
	cp -R vendor/github.com/segmentio/kafka-go etc/testing/kafka
	docker build -t kafka-demo etc/testing/kafka || { rm -r etc/testing/kafka/kafka-go; exit 1; }
	rm -r etc/testing/kafka/kafka-go
```
I also poached this for my kafka example.  How do we fix this for modules?

ysimonson:
@pappasilenus I'm not sure if we can remove vendoing in the first place, since it's pretty tied into our build infrastructure. This is still very much an open experiment. In the mean-time, I'm happy to walk through whatever issue you're trying to sort out for your example.

ysimonson:
@pappasilenus I think I got it working. See the difference in `./etc/testing/kafka/Dockerfile`.

etc/compile/wait.sh
```diff
@@ -2,8 +2,7 @@
 
 CONTAINER="${1}"
 
-docker wait "$CONTAINER" 2>/dev/null 1>/dev/null
-RET=$?
+RET=`docker wait "$CONTAINER"`
```
ysimonson: this was change because `docker wait` prints the exit code, rather than returning the exit code, of the container: https://docs.docker.com/engine/reference/commandline/wait/

ysimonson:
@msteffen I tried caching the entire gopath, but that leads to a problem: TravisCI clones the repo into `$GOPATH`, but also tries to pull the directory from cache. This leads to permissioning issues.

ysimonson:
@msteffen This is ready for review. Please be thorough with it; at a certain point, it became a trial and error process to get TravisCI to pass, so I'm not confident that I've ironed out all the issues with it.

Makefile
```diff
@@ -151,10 +151,10 @@ docker-clean-worker:
 
 docker-build-worker: docker-clean-worker
 	docker run \
-		-v $(PWD):/go/src/github.com/pachyderm/pachyderm \
+		-v $(PWD):/pachyderm \
```
msteffen: Not opposed to this path change, but just curious about it: do go modules not work under `$GOPATH` or something? Or is this just taking the opportunity to make the path a little nicer?

Makefile
```diff
@@ -185,13 +185,13 @@ docker-build-test: docker-clean-test docker-build-compile
 	  --attach stdout \
 	  --attach stderr \
 	  --rm \
-	  -w /go/src/github.com/pachyderm/pachyderm \
-	  -v $$GOPATH/src/github.com/pachyderm/pachyderm:/go/src/github.com/pachyderm/pachyderm \
+	  -w /pachyderm \
+	  -v $(PWD):/pachyderm \
```
msteffen: Just want to check that this make target has been tested or gets tested before merging, because I'm surprised this works. My expectation is that you'd need either `$$(PWD)` (make-escaped `$`) or `$(shell pwd)` or `$(CURDIR)` (but that would only work when running make from `$GOPATH/src/github.com/pachyderm/pachyderm`)

etc/proto/test_no_changes.sh
```diff
@@ -11,7 +11,6 @@ cd "${scriptdir}/../.."
 # hash our generated protobuf code, mostly to see if make proto changed anything
 orig_hash="$(
 find src -regex ".*\.pb\.go" \
-  | grep -v vendor \
```
msteffen: :tada:

etc/testing/kafka/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM golang:1.12.1
 RUN mkdir /app 
 ADD . /app/ 
-WORKDIR /app 
-ADD ./kafka-go /go/src/github.com/segmentio/kafka-go
+WORKDIR /app
+RUN go get github.com/segmentio/kafka-go
```
msteffen: Should we tie this to a release? (I *think* `go get <target>@<version>` works with vendoring on?)

src/server/pkg/testutil/cmds.go
```diff
@@ -105,7 +105,7 @@ export -f yes # use in subshells too
 
 which match >/dev/null || {
 	echo "You must have 'match' installed to run these tests. Please run:" >&2
-	echo "  cd ${GOPATH}/src/github.com/pachyderm/pachyderm/ && go install ./src/testing/match" >&2
+	echo "  go install ./src/testing/match" >&2
```
msteffen: How would you feel about `go install github.com/pachyderm/pachyderm/src/testing/match`? It's shorter and less side-effectful than `cd ...` but still works when you're not in the top-level `pachyderm/pachyderm` dir.

Makefile
ysimonson: IIRC, I was hitting some issue and wanted to rule out Go's sometimes semantically significant pathing (i.e., at least in Go < 1.13, module-enabled repos could not be in `$GOPATH`.) Even if Go 1.13 removed this limitation, I feel like putting the repo outside of `$GOPATH` reduces ambiguity.

Makefile
ysimonson: Huh, yeah good catch. It works on my system, but then I'm on macOS and the case insensitivty allows even `$(Pwd)` to work. Doesn't TravisCI call this target at some point?

src/server/pkg/testutil/cmds.go
ysimonson: I suppose the one minor risk with that is that we make changes to `match`, and someone trying to run older tests mistakenly installs the latest version of `match`. But I think it's minor enough that this makes sense.

etc/testing/kafka/Dockerfile
ysimonson: Actually, this didn't work due to the aforementioned semantic pathing issue. Even with go 1.13, `go get` using modules does not work in `$GOPATH`.

chainlink:
I'm assuming this has already been covered, but just want to make sure we have a good caching story in CI around this.

ysimonson:
Go mods are cached, however there does appear to be some overhead to this even though modules aren't re-pulled. I think it's overhead related to TravisCI caching (it's not as fast as cloning from github, and it has to run a diff when the tests finish.) Anecdotally, it seems to add 2-3 minutes to builds.

ysimonson:
Also, this is something that could be optimized on a more flexible CI; e.g. in CircleCI, we could bake the go modules into the docker image, and just re-build the image whenever the dependencies are updated.

PR #4324, opened 2019-11-08
Allow for retrying of `GetLogs` in test, in case k8s pods were evicted

ysimonson:
This may fix intermittent failures of `TestGetLogs`


ysimonson:
@msteffen this is ready for review. Two open questions:

1. Should we apply a backoff for the call to `GetLogs` when stats is enabled? Presently I do, but if I'm reading the code right, it shouldn't be necessary (but maybe it's a good idea anyways, for some other reason?)
2. Should we apply a backoff [to get the logs for a jobs/datums](https://github.com/pachyderm/pachyderm/pull/4324/files#diff-169b88a58cc85335a06d29c2e7678bc9R4090-R4118)? I'm pretty sure the answer is no, but wanted to double check.

ysimonson:
Given your intuition, adding a backoff for jobs/datums makes sense. My understanding of the code (which may be wrong) is that jobs/datums weren't prone to failure due to pod eviction since it wasn't pulling logs from the pods, but that may be wrong.

PR #4346, opened 2019-11-15
Add error and other info to obj traces

msteffen:


msteffen:
Travis claims that CI is in progress, but clicking through to the build page shows that all tests have passed

src/server/pfs/server/obj_block_api_server.go
```diff
@@ -534,7 +544,10 @@ func (s *objBlockAPIServer) InspectObject(ctx context.Context, request *pfsclien
 }
 
 func (s *objBlockAPIServer) CheckObject(ctx context.Context, request *pfsclient.CheckObjectRequest) (response *pfsclient.CheckObjectResponse, retErr error) {
-	func() { s.Log(request, nil, nil, 0) }()
+	func() {
+		tracing.TagAnySpan(ctx, "err", retErr, "object", s.prettyObjPath(request.Object))
+		s.Log(request, nil, nil, 0)
```
chainlink: Do we want to log out time here? (or take out time logging above if it should appear in the trace native?)

src/server/pkg/obj/amazon_client.go
```diff
@@ -291,9 +291,11 @@ func (c *amazonClient) Reader(ctx context.Context, name string, offset uint64, s
 		}
 		req.Header.Add("Range", byteRange)
 
-		backoff.RetryNotify(func() error {
-			span, _ := tracing.AddSpanToAnyExisting(ctx, "aws/cloudfront.Get")
-			defer tracing.FinishAnySpan(span)
+		backoff.RetryNotify(func() (retErr error) {
```
chainlink: Should we add these for all storage providers consistently? (ie, GCS, local, etc)

src/server/pfs/server/obj_block_api_server.go
msteffen: This was actually a mistake on my part, in where I put the tracing—though good catch noticing the discrepancy!

The way that `s.Log()` works is that when it's called it inspects its own stack, goes up the stack a certain number of frames, extracts the caller's name from that ancestral stack frame, and then uses the caller's name in its logging. Because the number of stack frames that it traverses is fixed, and we always call `defer func() { ... s.Log() }`, we need to wrap the logging that we do at the *beginning* of a call in a `func() { ... }` as well so that there are the same number of stack frames between `CheckObject` and `s.Log` in all cases.

So, basically, this function runs at the beginning of the call (note no `defer`). I meant to put `tracing.TagAnySpan` in the func right below that, that *is* deferred, so I've now moved it.

src/server/pkg/obj/amazon_client.go
msteffen: That's my goal (to try and handle all of the storage clients equivalently). The Amazon client is slightly unique in that it has the special handling for the case where a cloudfront distribution has been set up in front of S3, and that's what this trace is supposed to capture. Thus, this particular kind of tracing won't be in other clients, but other kinds of tracing should be:

- clients where `Exists` is implemented as some kind of `_, err := realclient.Stat(path); return err == nil` I tried to make sure to add the error to the trace
- clients that have their own implementation of `Reader` and `Writer`, I tried to make sure to instrument the custom calls to `Read` and `Write`

src/server/pfs/server/obj_block_api_server.go
msteffen: (I attempted to find any other places where I might have made the same mistake, and couldn't, but if you see any let me know!)

PR #4361, opened 2019-11-25
Greater resilience when following pipeline logs

### LINKED ISSUE #4358: pachctl logs -f  with -p flag should not exit if pipeline updated

If you follow the logs for a pipeline, and the pipeline is updated, the pachctl command following the logs exits.  Rerunning the command will follow the logs again.

Submitting this as a feature request because it's not rightly a bug

In the shell, when running the pipeline log, you can use a shell loop to force a reread when pachctl exits.  
```
while [ true ] ; do pachctl logs -p pipeline -f ; done
```

Since we are hiding the fact that the pipeline is running in a particular pod from the user with this bit of syntactic sugar, the command should not exit.  pachd and pachctl should handle the cutover to a new pod seamlessly, perhaps even putting a "pod updated" event into that log stream.

ysimonson:
Closes #4358 

Open questions/issues:
* Authorization check is only done up-front. If a pipeline is updated, in-flight `GetLogs` requests will continue to stream logs from the new pod(s). Is this OK?
* When following pipeline logs and stats is disabled, a couple of issues may arise:
  * Log lines may repeat as they're not de-duplicated between pod log fetch requests.
  * If the tail number is low, log lines may be skipped (because they happen between pod log fetch requests.)

While this PR contains some general improvements as well (avoidance of unstructured concurrency; support for cancelling stats logs requests, among others), I'm not sure if it's worthwhile due to the extra complexity and the above caveats. I'm leaning towards not merging it.


src/server/pps/server/api_server.go
```diff
@@ -1347,129 +1355,51 @@ func (a *apiServer) GetLogs(request *pps.GetLogsRequest, apiGetLogsServer pps.AP
...
+		statsLogs := false
 		if statsCommit != nil {
 			ci, err := pachClient.InspectCommit(statsCommit.Repo.Name, statsCommit.ID)
 			if err != nil {
 				return err
 			}
 			if ci.Finished != nil {
-				return a.getLogsFromStats(pachClient, request, apiGetLogsServer, statsCommit)
+				statsLogs = true
+				eg.Go(func() error {
```
msteffen: very very minor style nit, but how would you feel about having this be:
```
if ci.Finished != nil {
  statsLogs = true
}
...
switch {
  case statsLogs:
    eg.Go(func() error {
      return a.getLogsFromStats(pachClient, request, statsCommit, logCh)
    })
  case request.Follow:
    eg.Go(func() error {
      return a.followLogsForRC(pachClient, request, containerName, "", logCh)
    })
  default:
    eg.Go(func() error {
      return a.getLogsForRC(pachClient, request, containerName, "", logCh)
    })
}
```

src/server/pps/server/api_server.go
```diff
@@ -1519,19 +1448,197 @@ func (a *apiServer) getLogsFromStats(pachClient *client.APIClient, request *pps.
...
+			TailLines: tailLines,
+		}).Timeout(10 * time.Second).Stream()
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		// Close the stream if the context is done. This is needed because we
+		// otherwise risk blocking indefinitely on `scanner.Scan()`.
+		<-ctx.Done()
```
msteffen: I *think* we have to manually close `ctx` in the success case if we want to avoid leaking this goro. Just checking if that happens? (I couldn't figure out where right away)

src/server/pps/server/api_server.go
```diff
@@ -1519,19 +1448,197 @@ func (a *apiServer) getLogsFromStats(pachClient *client.APIClient, request *pps.
...
+		select {
+		case <-ctx.Done():
+			return err
+		default:
+			return nil
+		}
+	})
+
+	// Do not propagate EOF errors, as they're strictly used to signal
+	// shutdown for pod log goros
```
msteffen: I'm afraid I don't understand this part. Why return EOF from `getLogsFromPod` and then ignore it here?

src/server/pps/server/api_server.go
```diff
@@ -1519,19 +1448,197 @@ func (a *apiServer) getLogsFromStats(pachClient *client.APIClient, request *pps.
...
 		})
 	}
 	return eg.Wait()
 }
 
+func (a *apiServer) followLogsForRC(pachClient *client.APIClient, request *pps.GetLogsRequest, containerName, rcName string, logCh chan *pps.LogMessage) error {
+	defer close(logCh)
+	ctx := pachClient.Ctx()
+
+	err := backoff.RetryNotify(func() error {
```
msteffen: This may be a known drawback of this approach, but it occurs to me that a risk is that the `backoff` loop starts, sends a bunch of logs, and then has to retry and sends all the same logs again.

Do you think it would be worth it to inject a log message in the `Notify` fn at the bottom saying something to the effect of "retrying connection, logs after this line may be duplicates"?

src/server/pps/server/api_server.go
```diff
@@ -1303,18 +1303,26 @@ func (a *apiServer) GetLogs(request *pps.GetLogsRequest, apiGetLogsServer pps.AP
 	pachClient := a.env.GetPachClient(apiGetLogsServer.Context())
 	ctx := pachClient.Ctx() // pachClient will propagate auth info
 
-	// Authorize request and get list of pods containing logs we're interested in
-	// (based on pipeline and job filters)
-	var rcName, containerName string
+	logCh := make(chan *pps.LogMessage)
+	var eg errgroup.Group
```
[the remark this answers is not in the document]
ysimonson: That's correct, with a couple of notes:

1) There is no `followLogsForStats`, and no special logic within `getLogsForStats` for follow. This holds true for master as well. I think I verified that, despite this, stats w/ follow works as expected, but would want to double-check on that before merging.

2) The original idea behind this error group was to avoid the bug where one of those methods forgets to close `logCh`, but in retrospect that seems like overkill, so I can remove it.

src/server/pps/server/api_server.go
```diff
@@ -1347,129 +1355,51 @@ func (a *apiServer) GetLogs(request *pps.GetLogsRequest, apiGetLogsServer pps.AP
...
-	// Get pods managed by the RC we're scraping (either pipeline or pachd)
-	pods, err := a.rcPods(rcName)
-	if err != nil {
-		return fmt.Errorf("could not get pods in rc \"%s\" containing logs: %s", rcName, err.Error())
-	}
-	if len(pods) == 0 {
-		return fmt.Errorf("no pods belonging to the rc \"%s\" were found", rcName)
-	}
+		if !statsLogs {
+			containerName := client.PPSWorkerUserContainerName
```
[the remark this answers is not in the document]
ysimonson: `containerName` will be `pachd` instead of this value in the case of getting logs for pachd

src/server/pps/server/api_server.go
ysimonson: not minor at all! it definitely clarifies things, will add it.

src/server/pps/server/api_server.go
ysimonson: This context comes from an error group. In the success case:

1) This function will return `io.EOF`.
2) The error group will receive the error from one of its workers, and close `ctx`.
3) All of the other workers will receive the `ctx` close here, and close their streams.

src/server/pps/server/api_server.go
ysimonson: When using error groups with contexts, the context will close as soon as the first worker returns an error. I'm using an `io.EOF` here to signal that one of the workers finished successfully, so that the other workers get notified to close.

As an aside, I debated whether to create a custom error or use `io.EOF`. I think the latter is right because:

1) It's in keeping with go stdlib's use of EOF as a signal of a valid state rather than an error.
2) In a sense, an EOF did happen, as the underlying log stream was closed.

src/server/pps/server/api_server.go
ysimonson: That's right, and additionally some log lines might be missed (see the PR description.) It's for that reason that I'm hesitant to merge this PR.

ysimonson:
> I gather that the resiliance comes from the retry loop in `followLogsForRC`?

Correct

src/server/pps/server/api_server.go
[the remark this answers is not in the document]
msteffen: :+1: I this it's correct that there's no `followLogsForStats`. That code works by reading the logs from completed jobs from finished stats commit. Because it only works on jobs where the job is done and the stats commit is closed, you effectively can't use it to follow a job (which, for "follow" to be meaningful, must still be emitting logs and therefore *not* done)

src/server/pps/server/api_server.go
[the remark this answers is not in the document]
msteffen: :+1:

src/server/pps/server/api_server.go
msteffen: :+1: Thanks! I didn't realize that that was how `errgroup.WithContext()` worked so thanks for the explanation.

src/server/pps/server/api_server.go
msteffen: Got it, thanks! I guess my only remaining question is: do we want to close all per-pod goros as soon as the first goro finishes? It seems like that means we'll drop logs from all goros after the first—is that what happens/is that actually the desired behavior?

src/server/pps/server/api_server.go
msteffen: I see. I re-read the description, and I think I actually don't totally understand the description of when log lines get skipped. Is the idea:
```
----------+-------------
Server    | user session
----------+-------------
Line 1    |  
Line 2    |  
Line 3    |  
Line 4    |  $ pachctl get-logs --tail=3
Line 5    |  Line 2
Line 6    |  Line 3
Line 7    |  Line 4
Line 8    |
Line 9    |  $ pachctl get-logs --tail=3
Line 10   |  Line 7
Line 11   |  Line 8
Line 12   |  Line 9
```

ysimonson: `pachctl logs -p <pipeline> --follow` (i.e. follow w/o tail) will repeatedly fetch logs from pods, but may miss logs in the time between requests. However, this will be trivially fixable if I can switch over to getting service logs instead.

src/server/pps/server/api_server.go
ysimonson: Yeah, I learned about that from working on this, and there's some other places where it might be useful: I think you had a todo in either the pachd or worker `main.go` to address the default error group behavior of waiting until all goros finish.

src/server/pps/server/api_server.go
ysimonson: That's a good point, I'm not sure. I thought it was the right way because it protects against leaks (e.g. one log stream somehow staying open even when the others are winding down), but it likely introduces its own problems. Luckily, this'll be moot once I switch this over to getting service logs.

ysimonson:
It looks like k8s' API [does not offer a way to get service logs](https://godoc.org/k8s.io/client-go/kubernetes/typed/core/v1#ServiceInterface) :'(

ysimonson:
Closing. K8s' API doesn't offer a way to get service logs (`kubectl` does, but it must be syntactic sugar). As a result, I don't foresee a way to avoid all the extra complexity introduced by this. While there are some other benefits, I don't think they outweigh the risks.

PR #4367, opened 2019-11-27
[DOC]Clarify transaction conflicts

svekars:


svekars:
@Tryneus and @ysimonson let me know if this clarifies it or whether we should the wording further.

doc/docs/master/how-tos/use-transactions-to-run-multiple-commands.md
```diff
@@ -199,10 +199,14 @@ transactions will be executed as soon as you close those
 transactions. To resume the initial transaction, you need to run
 `pachctl resume transaction`.
 
-Because nested transactions are not supported, transactions cannot
+Because nested transactions are not supported, transactions typically cannot
```
ysimonson: I'd just remove this sentence, since support for transaction nesting is not relevant to conflict resolution (i.e., a system that supports transaction nesting is not more prone to conflict than one that doesn't.)

Tryneus: Agreed - I don't understand what this sentence is trying to communicate, and I don't think it adds much.  Other than that, everything else looks good.

PR #4376, opened 2019-12-03
Rewrite configs on reset

ysimonson:
This is just to address a small annoyance with the reset script. Presently, we create a new context on every reset (even though `pachctl` has logic to prevent this.) This fixes the issue.

### DIFF
--- a/etc/reset.py
+++ b/etc/reset.py
@@ -194,11 +194,8 @@ def rewrite_config():
     contexts = v2["contexts"]
 
     for k, v in contexts.items():
-        if not LOCAL_CONFIG_PATTERN.fullmatch(k):
-            continue
-        if len(v) == 0:
-            continue
-        keys.add(k)
+        if LOCAL_CONFIG_PATTERN.fullmatch(k) and len(v) > 0:
+            keys.add(k)
 
     for k in keys:
         del contexts[k]
--- a/etc/reset.py
+++ b/etc/reset.py
@@ -1,7 +1,9 @@
 #!/usr/bin/env python3
 
 import os
+import re
 import sys
+import json
 import time
 import select
 import logging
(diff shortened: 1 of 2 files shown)


etc/reset.py
```diff
@@ -172,9 +176,37 @@ def get_pachyderm(deploy_version):
...
+def rewrite_config():
+    log.info("Rewriting config")
+
+    keys = set([])
+
+    try:
+        with open(os.path.expanduser("~/.pachyderm/config.json"), "r") as f:
+            j = json.load(f)
+    except:
+        return
```
nitinjainsj: If open does fail, how would you know?

etc/reset.py
```diff
@@ -172,9 +176,37 @@ def get_pachyderm(deploy_version):
...
+        
+    v2 = j.get("v2")
+    if not v2:
+        return
```
nitinjainsj: How would you know its not v2?

etc/reset.py
ysimonson: Failure here or below most likely means the config does not exist, which means no rewriting is necessary. In the case where there is a failure, it's not a big deal since this is just simplifying the config, and the step can be skipped without negative effect. In the rare case where there is some sort of issue (e.g. the config file is invalid), an error will occur on the next invocation of `pachctl`.

nitinjainsj: Failures can be due to insufficient permissions. Do you want to add logs so it is clear why it failed?

ysimonson: In that case, the next invocation of `pachctl` will fail as well

etc/reset.py
```diff
@@ -172,9 +176,37 @@ def get_pachyderm(deploy_version):
...
+
+    contexts = v2["contexts"]
+
+    for k, v in contexts.items():
+        if LOCAL_CONFIG_PATTERN.fullmatch(k) and len(v) > 0:
```
nitinjainsj: This is for my info: Why only the local context? Why is this not a problem with a remote context?

ysimonson: That would remove the ability to access remote clusters, since their contexts would be gone

nitinjainsj: Isn't the intent here to reset everything -- be it remote or local? Or do we support reset only for local deployments which represents a dev/test kind of workflow?

ysimonson: It's only for resetting local dev deployments

PR #4393, opened 2019-12-06
Expose configurable object storage upload concurrency

brycemcanally:


brycemcanally:
Yeah, it probably makes more sense to have this happen in the PutObject request, but I don't think it would be that useful in the short term. This configuration is going to have different semantics in the new storage layer, and it is only being exposed for the current storage layer to solve a particular problem (high number of concurrent put file requests, particularly with recursive put files). I think the concern of high concurrency for putting hash trees will probably not be a real concern before the new storage layer is released, but I can make the changes if you feel they are necessary.

Tryneus:
That's a fair explanation as long as we intend to revisit this in the future.  Otherwise this feels like technical debt, but if we can implement this on top of the new storage later once it lands, that's fine.

PR #4394, opened 2019-12-06
[DOC] Added a note that stats cannot be disabled after being enabled.

### LINKED ISSUE #4332: Add a note that stats cannot be disabled after they're enabled
Once stats on a pipeline are enabled, they cannot be disabled. This should be noted in the pipeline spec docs: https://docs.pachyderm.com/latest/reference/pipeline_spec/#enable-stats-optional

svekars:
- Closes #4332
- Rephrased a bit


JoeyZwicker:
I think you mean they cannot be disabled? 

And it might be good to note that logs can take up a decent amount of space and since stats can't be disabled, there isn't a good way (yet) to reclaim that storage space

PR #4399, opened 2019-12-09
Remove enterprise check from S3 Gateway

msteffen:


ysimonson:
Needs the removal of enterprise enablement in the makefile target (`test-s3gateway-integration`); otherwise lgtm

ysimonson:
Also, this section needs to be removed: https://docs.pachyderm.com/latest/enterprise/overview/#s3gateway

PR #4408, opened 2019-12-11
Stop building as root in 'make docker-build'

msteffen:


Makefile
```diff
@@ -146,26 +146,34 @@ publish-compile: docker-build-compile
...
 
 docker-clean-worker:
+	rm -rf docker_build_worker.tmpdir
 	docker stop worker_compile || true
 	docker rm worker_compile || true
 
 docker-build-worker: docker-clean-worker
 	docker run \
+		--env=CALLING_USER_ID=$$(id -u $$USER) \
+		--env=DOCKER_GROUP_ID=$$(getent group docker | cut -d: -f3) \
```
ysimonson: this doesn't work because macOS doesn't have `getent`

msteffen: OK, thanks! I just changed this to avoid `getent`. The commands work on our office mac, though it doesn't have docker installed, so I can't test the full build on it.

I'll go down the path of setting up our office mac to build Pachyderm, but if you got the chance to pull the new commit and try it, that would be super helpful.

Makefile
```diff
@@ -146,26 +146,34 @@ publish-compile: docker-build-compile
 	docker push $(COMPILE_IMAGE)
 
 docker-clean-worker:
+	rm -rf docker_build_worker.tmpdir
```
ysimonson: Do we need this directory in the first place? It seems a bit atypical to me that we mount shared volume for builds.

ysimonson:
It looks like the new setup doesn't work on mac either, as `cat /etc/group | grep docker | cut -d: -f3` gives empty output

Makefile
msteffen: For posterity: there was a conversation about this in slack, and we all agreed that multi-stage builds would be better than bind mounts. @chainlink posted a solution that uses skaffold, which may be the best longer-term solution. For now, I don't think we could change this without changing our overall build process significantly.

On Slack, I believe the main reservations around bind mounts were whether they worked on docker-for-mac, and it seems that they do. Docker-for-mac seems to somehow mount host volumes into the docker-for-mac VM, and does so without changing the owner to root.

msteffen:
> It looks like the new setup doesn't work on mac either, as cat /etc/group | grep docker gives empty output. Maybe because it's running a VM?

**Update** I had to make quite a few changes to get this PR to work on mac, but I just tested it on our office mac and it seems to work.

Makefile
ysimonson: Yup it works on mac 👍

etc/compile/compile.sh
```diff
@@ -1,47 +1,46 @@
...
+# make targets. It doesn't actually run the go compiler (that's in
+# run_go_cmds.sh, called at the bottom) but it sets up an unprivileged user
+# with the host caller's user ID, under which the go compiler will run
 
 set -Eex
 
-DIR="$(cd "$(dirname "${0}")/../.." && pwd)"
-cd "${DIR}"
-pwd
+# On OS X, skip setting all of this user/group stuff and just build as root.
```
ysimonson: super nit, but it's called macOS now

etc/compile/compile.sh
```diff
@@ -1,47 +1,46 @@
...
+# The way bind mounts work in docker-for-mac are such that even files created
+# by root inside the docker container will be owned by the calling user and
+# group on the host filesystem
 
-BINARY="${1}"
-LD_FLAGS="${2}"
-PROFILE="${3}"
+env
```
ysimonson: do we still want this, or was it just for debugging?

etc/compile/run_go_cmds.sh
```diff
@@ -0,0 +1,56 @@
...
+#
+# This file actually builds Pachyderm binaries ('pachd' and 'worker') and is
+# called by etc/compile/compile.sh inside the Pachyderm build container, but it
+# run as an unprivileged user with the host caller's user ID, instead of as
+# root. This prevents 'go build' from littering the pachyderm git directory
+# with root-owned files
+
+set -Eex
+
+env
```
ysimonson: likewise, do we want to print out env here?

etc/compile/run_go_cmds.sh
```diff
@@ -0,0 +1,56 @@
...
+which go
+
+# Navigate to root of repo
+cd "$(dirname "${0}")/../.."
+
+BINARY="${1}"
+LD_FLAGS="${2}"
+PROFILE="${3}"
+
+mkdir -p _tmp
```
ysimonson: would it make sense to use `mktemp -d` instead? maybe also instead of `docker_build_worker.tmpdir` and `docker_build_worker.tmpdir`?

etc/compile/run_go_cmds.sh
msteffen: Fixed

etc/compile/compile.sh
msteffen: Fixed

etc/compile/compile.sh
msteffen: Jokes, it was for debugging. Now removed.

I stopped printing stdout from the build container without removing this and of course forgot about it.

etc/compile/run_go_cmds.sh
msteffen: I removed the `mkdir -p _tmp` line, which I should've removed a while ago (we don't use `_tmp` for anything in the script). I originally changed the directory used for building the `pachd` and `worker` images from `_tmp` to `docker_build_xyz.tmpdir` just so that when the build failed, I wouldn't be stuck with this `_tmp` directory in my repo, trying to remember why it was there (i.e. the new names are supposed to be clearer).

IMO the advantage of using a static name rather than `mktemp` is that if we're trying to fix compilation errors, and running `make docker-build` repeatedly (and it's failing, due to the compilation errors that we're trying to fix), we don't accumulate `docker_build_xyz` directories as we go. Instead the `docker_build_xyz.tmpdir` directory is reused. But I really don't have a strong opinion about this—either approach seems completely fine to me. If you'd prefer I change it in light of my rationale I'm happy to.

ysimonson: I think using `mktemp` makes more sense because on systems using tmpfs (including ubuntu and I imagine most linux distros):

1) The temporary directory is not persisted, so it's wiped out very regularly.
2) Because the contents are stored in-memory, it gives a big performance boost versus persisting to disk.

However, this isn't a blocker because really the long-term fix is multi-stage builds, which will obviate this anyways.

msteffen: This makes sense. I think on ubuntu, `/tmp` is actually a regular dir, while `/dev/shm` is a tmpfs mount. I couldn't actually build the docker image in `/dev/shm` because, by default, the ubuntu docker image limits that tmpfs mount to 64MB, which cause the build to fail with "out of space" errors.

But building in `/tmp` with regular `mktemp` worked, and, I realized, had the side benefit that we never leave temporary dirs on the host machine at all 🤦

PR #4447, opened 2019-12-20
Add additional logging to pachd setup

brycemcanally:
Adds logging to each grpc server setup in pachd and bumps up the timeout for setting up pachd, etcd, and kube clients to 5 mins.

### DIFF
--- a/src/server/cmd/pachd/main.go
+++ b/src/server/cmd/pachd/main.go
@@ -135,7 +135,6 @@ func doSidecarMode(config interface{}) (retErr error) {
 	if env.Metrics {
 		reporter = metrics.NewReporter(clusterID, env)
 	}
-
 	pfsCacheSize, err := strconv.Atoi(env.PFSCacheSize)
 	if err != nil {
 		return fmt.Errorf("atoi: %v", err)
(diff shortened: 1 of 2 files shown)


src/server/cmd/pachd/main.go
```diff
@@ -623,3 +739,15 @@ func getNamespace() string {
...
+	log.Printf("started setting up %v GRPC Server", name)
+	defer func() {
+		if retErr != nil {
+			log.Printf("errored setting up %v GRPC Server", name)
+		} else {
+			log.Printf("finished setting up %v GRPC Server", name)
+		}
+	}()
+	return f()
+}
```
ysimonson: I think this could be simplified to avoid named return arguments and deferred statements:

```go
func logGRPCServerSetup(name string, f func() error) error {
    log.Printf("started setting up %v GRPC Server", name)
    err := f()
    if err != nil {
    	log.Printf("errored setting up %v GRPC Server", name)
    } else {
    	log.Printf("finished setting up %v GRPC Server", name)
    }
    return err
}
```

src/server/pkg/serviceenv/service_env.go
```diff
@@ -153,7 +153,7 @@ func (env *ServiceEnv) initKubeClient() error {
 			return fmt.Errorf("could not initialize kube client: %v", err)
 		}
 		return nil
-	}, backoff.RetryEvery(time.Second).For(time.Minute))
+	}, backoff.RetryEvery(time.Second).For(5*time.Minute))
```
ysimonson: why these increases? 5min seems like a long time to wait for the servers to initialize.

ysimonson:
Going to merge this so we can get it in the new release, but will open a ticket to address PR feedback at a later point.

src/server/cmd/pachd/main.go
brycemcanally: I don't think this improves the readability. The common pattern in our codebase (and I think in general) for this type of basic logging is to check the returned error in a deferred function. The pattern you suggested only works when there is a single error source. I think it is better to be consistent rather than have a special pattern that works only in this case.

src/server/pkg/serviceenv/service_env.go
brycemcanally: Yeah, it should be in general. The main reason for the bump (as far as the pach client is concerned) was to ensure that an object storage client timeout would occur before the pach client timeout (if the object storage client would time out for whatever reason). I don't have a great idea where all the possible points of slowness for startup are now and where they will be in the future, so I figured being conservative at 5 mins would be fine considering this is not something that should happen frequently and is not recoverable. I am open to bumping this back down though if you feel strongly about it.

src/server/cmd/pachd/main.go
ysimonson: I won't block on this, but disagree. A discussion with @jdoliner has convinced me that named return arguments in go are a bad pattern, because they make it trivial to accidentally shadow the variable, causing unexpected behavior (go catches some, but not all of the ways this can happen.) That doesn't apply here, but I do think that overall, we should move more of the codebase away from named return arguments.

brycemcanally: I'm not sure I follow how named return arguments make it easier to accidentally shadow variables. A named return argument is in the outer scope of the function and a variable with the same name cannot be declared in that same scope. I think your concern might be more so with naked returns, which I think are atrocious. Having the returned values available in a deferred function though seems pretty useful in certain cases, such as this one.

ysimonson: https://github.com/pachyderm/pachyderm/pull/3906#discussion_r311259971

brycemcanally: Yeah, the naked return is the dangerous part there, not the named return arguments. Not sure if you are agreeing or disagreeing lol. But, I think that particular case has no use for named return arguments anyways. My main point is that I think they are useful in certain cases and do not increase the probability of accidental shadowing unless you use them with naked returns.

ysimonson: If instead of "avoid named return arguments", the heuristic is "avoid naked returns", that's still an improvement to me. I'd rather us just avoid named returns entirely, because I don't see much merit in named return arguments without naked returns.

brycemcanally: I think the merit is shown by the use case in this PR (and other logging code in the codebase). Being able to see the returned results of a function in a deferred function allows you to include some logic when a function returns that is based on the results of the function. Anyways, I think we agree that naked returns are dangerous and we should probably stay away from them.

PR #4490, opened 2020-01-10
Update deploy test to ignore local pach config

msteffen:


Makefile
```diff
@@ -399,7 +399,7 @@ staticcheck:
 test-proto-static:
 	./etc/proto/test_no_changes.sh || echo "Protos need to be recompiled; run make proto-no-cache."
 
-test-deploy-manifests:
+test-deploy-manifests: install
```
brycemcanally: I feel like the install should also apply when doing a regenerate. You could just add another make target for regenerating.

msteffen: :+1: Added a new make target

etc/testing/deploy-manifests/validate.sh
```diff
@@ -84,15 +89,21 @@ done
...
-    echo "If this deliberate, run:" >/dev/stderr
-    echo "  validate.sh --regenerate" >/dev/stderr
-    echo "  git add etc/testing/deploy-manifests/golden" >/dev/stderr
-    echo "  git commit" >/dev/stderr
-    echo "to replace golden deployment manifests with current output" >/dev/stderr
-    exit 1
-  fi
+echo ""
+echo "Diffing 'pachctl deploy' output with known-good golden manifests."
+echo "If this fails but should pass, run:"
```
brycemcanally: I don't think this is worded the way you intended. Also, it might make sense to only recommend using the make target rather than the shell script (to avoid regenerating against old versions of pachctl).

msteffen: Done :+1:

PR #4493, opened 2020-01-13
Fix python setup process

ysimonson:
The python-pachyderm library calls into the travis setup script in its pachyderm submodule. With the deprecation of python 2, pachyderm's travis setup script appears to fail in some environments. This PR is tracking the fixes that'll ensure the setup script works both here and in python-pachyderm.

### DIFF
--- a/etc/testing/travis_install.sh
+++ b/etc/testing/travis_install.sh
@@ -22,7 +22,6 @@ sudo cp etc/build/fuse.conf /etc/fuse.conf
 sudo chown root:root /etc/fuse.conf
 
 # Install aws CLI (for TLS test)
-pip3 install --upgrade --user pip
 pip3 install --upgrade --user wheel
 pip3 install --upgrade --user awscli
 


ysimonson:
It looks like this fixes the build process for python-pachyderm.

A little extra context. This line appears to install pip v8 (which uses python 2) rather than pip v19 (which uses python 3). Due to the EOL of python 2, that line appears to trigger an error in python-pachyderm's travis process. Strangely, that error doesn't happen in this repo's travis process, even though both repos use the same xenial-based distribution. It's possible that it's simply because this package is cached, and a future build of this repo without cache would trigger the same error.

ysimonson:
Tagging this for 1.9.11, because it's blocking the release of 1.9.10/1.9.11 support in python-pachyderm

ysimonson:
TBH, I'm not sure why it was attempting to install an older version of pip. My (completely speculative) guess is that `--user` made `pip3` attempt to install for a python 2.x environment. In any case, it looks like apt now relies on a relatively new version of pip, and this fixes the issue on both repos.

Issue #4465, opened 2020-01-03
Support pachctl commands to TLS-enabled pachd when port forwarding


jfhamlin:
**What is the goal / desired outcome?**

It would be great to be able to run `pachctl` commands against a pachyderm cluster with TLS enabled without having to make pachd directly addressable (i.e. reachable without kube port forwarding).

**If there is a way to accomplish this today via workaround, what does that require?**

As far as I can tell, there is no workaround. Either TLS has to be disabled, or `pachd` needs to be reachable without port forwarding.

**(Optional) What is your proposal for a feature to solve this?**

1. Support setting `ServerCAs` without setting `PachdAddress`. This should cause the client to assume a `Secured` address but fall back to port forwarding to connect to the daemon.
2. To allow certificate verification to succeed when connecting to localhost, support a new `ServerNameOverride` or `PachdServerNameOverride` context setting. This will be passed to [credentials.NewClientTLSFromCert ](https://github.com/pachyderm/pachyderm/blob/master/src/client/client.go#L559) to allow certificate verification to proceed despite a hostname mismatch.

**Environment?**:
Kubernetes version (use kubectl version): v1.14.7
Pachyderm CLI and pachd server version (use pachctl version): 1.9.10
Cloud provider (e.g. aws, azure, gke) or local deployment (e.g. minikube vs dockerized k8s): aws (eks)
OS (e.g. from /etc/os-release):
Others:

msteffen:
Hi James, thanks for filing this! This is definitely something we can add to our roadmap, but as a short-term workaround, you might be able to run `pachctl port-forward` and set `PachdAddress` to `grpcs://localhost:30650` (IIRC, that's the port that `port-forward` forwards) to get more or less this behavior.

Let us know if you try it, and if it works!

jfhamlin:
> Let us know if you try it, and if it works!

Yep, I tried this! Thinking about it again, I can probably get the port-forward workaround to work by adding a line to my `hosts` file mapping the hostname in the certificate to 127.0.1.1. I still think it'd be great for `pachd` to support this configuration, as it seems like it would be a common case for TLS deployments. I actually ended up hacking this in and running a patched pachctl locally, so if there's alignment on the proposal I can submit a PR sometime soon.

msteffen:
Oh interesting. I struck out my last response because I didn't think of modifying the hosts file.

[...]

PR #4509, opened 2020-01-20
Add new context value "server-name-override" to override TLS hostname

jfhamlin:
Setting pachd address is no longer required for TLS connections to pachd. If serverCAs is set, TLS is assumed, and a server name override will permit a host check to succeed against localhost when port forwarding.

Implements #4465


jfhamlin:
There are some conflicts in the proto definition and in some client connection logic I'll have to look through. A review on the general approach here would still be appreciated!

jrockway:
Hi @jfhamlin!  Are you still interested in merging this?  The architecture of Pachyderm has changed dramatically since you opened this PR, but I think this change could still apply cleanly with a few tweaks.

PR #4512, opened 2020-01-21
Configuration for only requiring critical Pachd servers to startup an…

brycemcanally:
…d run without error

Master PR #4536

### DIFF
--- a/src/server/pkg/deploy/assets/assets.go
+++ b/src/server/pkg/deploy/assets/assets.go
@@ -138,6 +138,18 @@ type StorageOpts struct {
 	UploadConcurrencyLimit int
 }
 
+const (
+	// RequireCriticalServersOnlyEnvVar is the environment variable for requiring critical servers only.
+	RequireCriticalServersOnlyEnvVar = "REQUIRE_CRITICAL_SERVERS_ONLY"
+)
+
+const (
+	// DefaultRequireCriticalServersOnly is the default for requiring critical servers only.
+	// (bryce) this default is set here and in the service env config, need to figure out how to refactor
+	// this to be in one place.
+	DefaultRequireCriticalServersOnly = false
+)
+
 // AssetOpts are options that are applicable to all the asset types.
 type AssetOpts struct {
 	FeatureFlags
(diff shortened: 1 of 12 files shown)


src/server/pkg/deploy/assets/assets.go
```diff
@@ -138,6 +138,18 @@ type StorageOpts struct {
 	UploadConcurrencyLimit int
 }
 
+const (
+	// RequireCriticalServersOnlyEnvVar is the environment variable for requiring critical servers only.
+	RequireCriticalServersOnlyEnvVar = "REQUIRE_CRITICAL_SERVERS_ONLY"
```
ysimonson: do we want this to be configurable?

src/server/pkg/deploy/assets/assets.go
```diff
@@ -138,6 +138,18 @@ type StorageOpts struct {
...
+)
+
+const (
+	// DefaultRequireCriticalServersOnly is the default for requiring critical servers only.
+	// (bryce) this default is set here and in the service env config, need to figure out how to refactor
+	// this to be in one place.
+	DefaultRequireCriticalServersOnly = false
```
ysimonson: I could see arguments both ways, but I'm curious: why did we decide to default this to false?

ysimonson:
Changes themselves look good, just wondering about configurability/defaults

src/server/pkg/deploy/assets/assets.go
brycemcanally: Yes, to allow users to be unblocked when a noncritical server errors during setup or while running.

src/server/pkg/deploy/assets/assets.go
brycemcanally: I made the default false because it made sense to me to fail fast if any server failed to come up. If the user is okay with the functionality of a noncritical server not being available, then they can set this flag. It basically ensures that a user is aware of a problem earlier rather than later. I can see the benefit of this defaulting to true for making starting out with Pachyderm easier though.

PR #4513, opened 2020-01-21
Change githook and prometheus ports

### LINKED ISSUE #4505: pachd won't start up because githook server port 999 open fails in non-root container
**What happened?**:
pachd 1.9.11 failed to start up on an OpenShift environment where pachd does not run as root, because the githook server is hard-coded to use port 999 and that port is privileged.

Attempted to configure the githook server using the api-git-port key in the spec for pachd to 1999, but that had no effect. 

````
               {
                "name": "api-git-port",
                "containerPort": 999,
                "protocol": "TCP"
              },
```
There appears to be a missing corresponding environment variable for setting the port?

1. Use a privilege-limited k8s environment, like OpenShift
2. Deploy Pachyderm using our recommendations: https://docs.pachyderm.com/latest/deploy-manage/deploy/openshift/, in particular setting `WORKER_USES_ROOT` to `false`.  Modify all ports to be above 1024.
3. Note that pachd crashes with a `permission denied` on port 999 for the githook server

**Environment?**:
- Kubernetes version (use `kubectl version`): OpenShift 
- Pachyderm CLI and pachd server version (use `pachctl version`): 1.9.11
- Cloud provider (e.g. aws, azure, gke) or local deployment (e.g. minikube vs dockerized k8s): on premisies
- Others:

brycemcanally:
Fixes #4505

Master PR #4537

### DIFF
--- a/etc/testing/deploy-manifests/golden/amazon-deploy-manifest.json
+++ b/etc/testing/deploy-manifests/golden/amazon-deploy-manifest.json
@@ -305,7 +305,7 @@
       "suite": "pachyderm"
     },
     "annotations": {
-      "prometheus.io/port": "9091",
+      "prometheus.io/port": "656",
       "prometheus.io/scrape": "true"
     }
   },
@@ -337,9 +337,9 @@
       },
       {
         "name": "api-git-port",
-        "port": 999,
+        "port": 655,
         "targetPort": 0,
-        "nodePort": 30999
+        "nodePort": 30655
       },
       {
         "name": "s3gateway-port",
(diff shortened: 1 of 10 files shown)


brycemcanally:
I need to make some changes here. I did not realize that the recommended way to configure ports was through manually editing the manifest (I thought they were not configurable based on the deploy commands). After reading over the openshift docs, it is clear to me that we need to follow the same pattern for the githook and prometheus ports.

brycemcanally:
We decided offline to kick full port configuration down the road until we have a template based deployment. We are still going to move the githook and prometheus ports into the same range as the other servers, but leave the configurability as is.

Issue #4118, opened 2019-09-21
pachctl delete pipeline -> etcdserver: request is too large


marcadella:
While running `pachctl delete pipeline` I got `etcdserver: request is too large`. This happened with `pachctl delete all` too.
Probably related to #3914.
Pachyderm version 1.9.3

ppaanngggg:
The same problem. 
Maybe there are too many commits in releatived pipelines or repos?

marcadella:
I didn't have more than 10 commits (x10 pipelines)

pappasilenus:
@marcadella @ppaanngggg Should have been fixed by https://github.com/pachyderm/pachyderm/pull/4121

Going to close this.  Please reopen if you hit it again.

marcadella:
Still having this issue with 1.9.10 with `pachctl start commit`...

echohack:
Hey @marcadella this is a bit of an old one, but it should be solved with the `--split-transaction` feature we added in 01-2021 for `pachctl delete pipeline`.

If you have any details around what you were running into with `pachctl start commit` could you open a new issue and attach logs of the repro, or add detailed steps to repro? Thank you!

PR #4525, opened 2020-01-24
Set grpc max message size for etcd client

brycemcanally:
This should fix https://github.com/pachyderm/pachyderm/issues/4118

### DIFF
--- a/src/server/pkg/serviceenv/service_env.go
+++ b/src/server/pkg/serviceenv/service_env.go
@@ -3,13 +3,13 @@ package serviceenv
 import (
 	"errors"
 	"fmt"
-	"math"
 	"net"
 	"os"
 	"time"
 
 	etcd "github.com/coreos/etcd/clientv3"
 	"github.com/pachyderm/pachyderm/src/client"
+	"github.com/pachyderm/pachyderm/src/client/pkg/grpcutil"
 	"github.com/pachyderm/pachyderm/src/server/pkg/backoff"
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


ysimonson:
I suspect that this change will cause more problems than it solves.

gRPC enforces message size limits on both the client- and server-side. The code removed in this PR removed the client-side enforcement of message size limits. As a result, the largest possible message size to etcd would be constrained by the server-side setting, which is [50mb by default.](https://github.com/pachyderm/pachyderm/blob/fa7cc8261d4aa62c4766b159f72f57bc1ca57774/src/server/pkg/deploy/assets/assets.go#L90)

If we merged this PR, the largest message an etcd client could send is [20mb.](https://github.com/pachyderm/pachyderm/blob/fa7cc8261d4aa62c4766b159f72f57bc1ca57774/src/client/pkg/grpcutil/stream.go#L13)

brycemcanally:
I'm not sure I follow this:

> The code removed in this PR removed the client-side enforcement of message size limits

I'm pretty sure this PR lowered the client side send message size limit (and the recv message size limit remains the same since what we were setting is the default). 

The goal with this was to reduce the max message size for grpc messages going from the etcd client to the etcd server. I do realize now there is a problem with this though (which I think you are getting at), which is that this configuration actually bounds etcd request sizes. I was thinking that this would control the splitting of requests across grpc messages rather than being the upper bound for requests.

ysimonson:
FWIW, this PR lowered _both_ the client-side send message size limit (from unlimited to 20mb) and client-side receive message size limit (from unlimited to gRPC's default of 2mb). I'm not sure the latter is a problem, but the former is.

brycemcanally:
The receive message size limit defaults to "unlimited".
```
// MaxCallRecvMsgSize is the client-side response receive limit.
// If 0, it defaults to "math.MaxInt32", because range response can
// easily exceed request send limits.
// Make sure that "MaxCallRecvMsgSize" >= server-side default send/recv limit.
// ("--max-request-bytes" flag to etcd or "embed.Config.MaxRequestBytes").
MaxCallRecvMsgSize int
```

I think the latter is a problem in the sense that it limits the request size to 20MB. I think leaving it "unlimited" allows us the option to handle large requests by bumping up the server side limit as well.

ysimonson:
I see - it looks like the etcd client itself overrides gRPC's default. Good to know.

PR #4542, opened 2020-01-29
update protobuf generation to use gogo library

Tryneus:
We currently have a problem which is that generated protobuf files use the `golang` protobuf library while our code uses the `gogo` protobuf library, which means that anything that needs to be registered with the library is missing from gogo.  This changes our protobuf generation script to use the `gogofast` compiler plugin (the equivalent of what we were previously using, `gofast`).


Tryneus:
Ref: see the "Protobuf Any types" section of [this blog post](https://jbrandhorst.com/post/gogoproto/).

PR #4571, opened 2020-02-05
Add a spout marker example

svekars:


examples/spouts/spout-marker/spout-marker-example.py
```diff
@@ -0,0 +1,48 @@
...
+        finally:
+            os.umask(umask_original)
+    return None
+try:
+    with open("/pfs/mymark", "r") as f:
+        lines = f.read().split("\n")
+except:
+    lines = []
+while True:
+    lines.append((lines[-1] if len(lines) > 0 else "") + "*")
```
pappasilenus: need to change this to
```
    lines.append((lines[-1] if len(lines) > 0 else "") + ".")
```

So the example starts off in the right state.

examples/spouts/spout-marker/README.md
```diff
@@ -0,0 +1,328 @@
...
+marker file that will track the progress of a pipeline.
+Then, we will modify the pipeline and observe
+how the spout continues to update records without interruption.
+
+## Prerequisites
+
+Before you begin, verify that you have the following components
+installed on your machine:
+
+* Pachyderm 1.9.10 or later
```
adelelopez: It's only fixed in 1.9.12

examples/spouts/spout-marker/README.md
```diff
@@ -0,0 +1,328 @@
...
+* Terminal
+
+## Pipeline Overview
+
+In this example, we will use a simple spout pipeline that
+will add dots into a spout marker file. Here is how the
+spout file will look like:
```
adelelopez: I think calling it the `marker file` would be more clear, since the spout will likely make several normal files as part of its normal behavior.

examples/spouts/spout-marker/README.md
```diff
@@ -0,0 +1,328 @@
...
+     "cmd": [ "python3", "/spout-marker-example.py" ],
+     "image": "spout-marker:v1"
+   },
+   "spout": {
+     "marker": "mymark"
+    }
+   }
+   ```
+
+   **Note:** In the `spout` section, you have a key-value pair
```
adelelopez: Maybe worth making it more explicit that `mymark` here will be the name of your marker file (or more accurately, will be make all files beginning with `mymark` into marker files, so you could have `mymark01` and `mymark02` if you needed two files)

examples/spouts/spout-marker/spout-marker-pipeline.json
```diff
@@ -0,0 +1,12 @@
+{
+  "pipeline": {
+    "name": "spoutmarker"
+  },
+  "transform": {
+    "cmd": [ "python3", "/spout-marker-example.py" ],
+    "image": "spout-marker:v10"
```
adelelopez: should this be `v1` instead?

examples/spouts/spout-marker/README.md
```diff
@@ -158,7 +158,10 @@ To create a spout pipeline, complete the following steps:
    ```
 
    **Note:** In the `spout` section, you have a key-value pair
-   `"marker": "mymark"`. This file will track your records.
+   `"marker": "mymark"`. `mymark` is the name of your marker file.
+   If you configure to use multiple marker files, `mymark` will be a
```
adelelopez: You don't have to configure anything, it automatically works either way!

PR #4578, opened 2020-02-07
[WIP] Refactor stopped pipeline logic

adelelopez:


msteffen:
Sorry for the drive-by comment but it just occurred to me: are we planning to merge this into 1.9? Could it cause issues if someone running 1.9 has a stopped pipeline (i.e. the PFS pipeline spec has `Stopped` set) and then upgrades to 1.9.13 with this fix? I'm not sure if they'll have any way to unset `Stopped` (`UpdatePipeline` preserves `Stopped` accross updates: https://github.com/pachyderm/pachyderm/blob/91019008652c7a41b9b8e234b87d343322c4ad7c/src/server/pps/server/api_server.go#L2065-L2067).

Maybe we should remove it from the proto completely? We could set the field number to `reserved` to ensure we can still read serialized protos where it's set, and then we'd be sure it's ignored everywhere.

adelelopez:
I was planning to remove it from the proto once I had the other issues resolved :)

adelelopez:
I think this approach is good, but it introduces another problem that I haven't had time to address yet. The new test I added now hangs on the last flush commit, possible due to an issue with the replication controller logic.

msteffen:

PR #4647, opened 2020-02-26
Split test envs out of the testutil package

Tryneus:
Under suggestion from @brycemcanally who is the lucky first user of the test env functions, this should make it easier to avoid circular dependencies when writing test code.  This adds two new packages, `testetcd` and `testpachd` which can be used to provide a temp directory and embedded (in-process) instance of `etcd` or the pachd API, respectively.  These are no longer a part of the `testutil` package.

Also fixed the `make test-transaction` target, which had the wrong path for tests.


src/server/pkg/testetcd/env.go
```diff
@@ -0,0 +1,151 @@
+package tetcd
```
brycemcanally: `tetcd` -> `testetcd`?

src/server/pkg/testpachd/mock_env.go
```diff
@@ -0,0 +1,85 @@
+package testpachd
+
+import (
+	"context"
+
+	"golang.org/x/sync/errgroup"
+
+	"github.com/pachyderm/pachyderm/src/client"
+	foo "github.com/pachyderm/pachyderm/src/server/pkg/testetcd"
```
brycemcanally: `foo`?

brycemcanally:
Also, why is this PR for `grey/updated-vscode-scripts`?

src/server/pkg/testpachd/mock_env.go
Tryneus: ugh - I ran into build problems (probably from the wrong package name that you spotted above), this was the workaround =(

fixed

Tryneus:
This is based off of `grey/updated-vscode-scripts` so that I could actually do builds in my environment.  I didn't want to pollute the diff by setting `master` as the base.

PR #4656, opened 2020-02-27
Fix fetching cached chunks in recycled workers

Tryneus:
If a pipeline picks up a job in the MERGING state, it fetches chunks from other workers by the address of the worker which originally processed the datums in that chunk.  There is the possibility for collisions in addresses, which can result in an empty hashtree for that chunk (because none of the workers have the chunks cached in this scenario).  This PR checks that the chunks are present rather than assuming they are and will fall back to fetching datum hashtrees from object storage when this fails.

### DIFF
--- a/src/server/worker/api_server.go
+++ b/src/server/worker/api_server.go
@@ -1125,10 +1125,15 @@ func (a *APIServer) Cancel(ctx context.Context, request *CancelRequest) (*Cancel
 // GetChunk returns the merged datum hashtrees of a particular chunk (if available)
 func (a *APIServer) GetChunk(request *GetChunkRequest, server Worker_GetChunkServer) error {
 	filter := hashtree.NewFilter(a.numShards, request.Shard)
+	cache := a.chunkCache
 	if request.Stats {
-		return a.chunkStatsCache.Get(request.Id, grpcutil.NewStreamingBytesWriter(server), filter)
+		cache = a.chunkStatsCache
 	}
-	return a.chunkCache.Get(request.Id, grpcutil.NewStreamingBytesWriter(server), filter)
+
+	if !cache.Has(request.Id) {
+		return fmt.Errorf("chunk (%d) missing from cache", request.Id)
+	}
+	return cache.Get(request.Id, grpcutil.NewStreamingBytesWriter(server), filter)
 }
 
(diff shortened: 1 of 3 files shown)


src/server/worker/api_server.go
```diff
@@ -1125,10 +1125,15 @@ func (a *APIServer) Cancel(ctx context.Context, request *CancelRequest) (*Cancel
...
 	filter := hashtree.NewFilter(a.numShards, request.Shard)
+	cache := a.chunkCache
 	if request.Stats {
-		return a.chunkStatsCache.Get(request.Id, grpcutil.NewStreamingBytesWriter(server), filter)
+		cache = a.chunkStatsCache
 	}
-	return a.chunkCache.Get(request.Id, grpcutil.NewStreamingBytesWriter(server), filter)
+
+	if !cache.Has(request.Id) {
+		return fmt.Errorf("chunk (%d) missing from cache", request.Id)
```
brycemcanally: Minor nit, but `cache` -> `chunk cache` to be consistent with the other error message.

src/server/worker/api_server.go
```diff
@@ -1374,8 +1379,11 @@ func (a *APIServer) mergeDatums(jobCtx context.Context, pachClient *client.APICl
 }
 
 func (a *APIServer) getChunk(ctx context.Context, id int64, address string, failed bool) error {
-	// If this worker processed the chunk, then it is already in the chunk cache
+	// If this worker processed the chunk, make sure it is already in the chunk cache
```
brycemcanally: Maybe this should be `If we think this worker processed this chunk...`. Also, I would be curious to know if the actual problem is recycled IPs.

src/server/worker/api_server.go
Tryneus: I didn't want to indicate it was the `chunkCache` over the `chunkStatsCache` - because either could hit this problem

src/server/worker/api_server.go
Tryneus: I think this situation could arise if a job errored and resumed in the `MERGING` state, even without recycling workers or IP addresses.  I haven't done a full audit of our error behavior to see exactly what's possible, though.

PR #4660, opened 2020-02-28
Add S3 inputs and outputs to pipelines

msteffen:
This makes it possible for jobs to read and write data via our S3 gateway. A few notes:
- Set `S3: true` in `Pfs` inputs to make them into s3 inputs
  - An input previously accessible at `/pfs/foo` is now accessible at `s3://foo`
- Set `S3Out: true` in the pipeline spec to write output via the S3 gateway (via `s3://out` instead of `/pfs/out` as you usually would)
  - Pipelines with `S3Out: true` cannot collect stats or skip datums because we have no idea what output files were written when—there's no way to know that datum i yielded output j and just merge j into the output like we do with regular pipeliens
- If a pipelines has any s3 inputs or s3 output, then the env var `S3_ENDPOINT` will be set in the user container, and you must use that to access the job's inputs and outputs, like so:
  ```
  echo foo_data | aws --endpoint=${S3_ENDPOINT} s3 cp - s3://out/foo_file
  ```


src/server/pps/server/api_server.go
```diff
@@ -195,14 +195,20 @@ func (a *apiServer) validateInput(pachClient *client.APIClient, pipelineName str
...
 						"second exposes repo contents via the filesystem")
-				case input.Pfs.S3:
-					return fmt.Errorf("'s3' inputs are not supported yet")
+				case input.Pfs.S3 && input.Pfs.Lazy:
+					return fmt.Errorf("input cannot specify both 's3' and 'lazy', as " +
+						"'s3' requires input data to be accessed via Pachyderm's S3 " +
+						"gateway rather than the file system")
+				case input.Pfs.S3 && input.Pfs.EmptyFiles:
+					return fmt.Errorf("input cannot specify both 's3' and " +
+						"'empty_files', as's3' requires input data to be accessed via " +
```
ysimonson:
```suggestion
						"'empty_files', as 's3' requires input data to be accessed via " +
```

src/server/pps/server/master.go
```diff
@@ -52,9 +52,8 @@ func (a *apiServer) master() {
 	backoff.RetryNotify(func() error {
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
-		// Use the PPS token to authenticate requests. Note that all requests
-		// performed in this function are performed as a cluster admin, so do not
-		// pass any unvalidated user input to any requests
+		// Note: 'pachClient' is unauthenticated. This will use the PPS token (via
+		// a.sudo())to authenticate requests.
```
ysimonson:
```suggestion
		// a.sudo()) to authenticate requests.
```

ysimonson:
@msteffen would you be able to review the s3g changes that I wrote?

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,359 @@
...
+				return fmt.Errorf("sidecar s3 gateway: couldn't initialize s3 gateway server: %v", err)
+			}
+			strport := strconv.FormatInt(int64(port), 10)
+			listener, err := net.Listen("tcp", ":"+strport)
+			if err != nil {
+				return fmt.Errorf("sidecar s3 gateway: could not serve on port %q: %v", port, err)
+			}
+			go func() {
+				server.Serve(listener)
+			}()
```
ysimonson: 1) This could be simplified with `server.ListenAndServe`.
2) This is presently ignore all errors from the HTTP server. Some errors will be for a legit state (i.e. because the server was closed), but we should probably just ignore those errors specifically.

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,359 @@
...
+			// server will forward requests based on the request hostname
+			port := s.apiServer.env.S3GatewayPort
+			server, err := s3.Server(port, driver, s)
+			if err != nil {
+				return fmt.Errorf("sidecar s3 gateway: couldn't initialize s3 gateway server: %v", err)
+			}
+			strport := strconv.FormatInt(int64(port), 10)
+			listener, err := net.Listen("tcp", ":"+strport)
+			if err != nil {
+				return fmt.Errorf("sidecar s3 gateway: could not serve on port %q: %v", port, err)
```
ysimonson: `%q` won't correctly format an int

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,359 @@
...
+			defer s.serversMu.Unlock()
+			// kill server
+			server, ok := s.servers[jobID]
+			if !ok {
+				// Note that because we call InspectJob after receiving a creation event, we
+				// may never create a service for a job that is created and then immediately
+				// deleted, so no error is returned if jobID is not in s.servers
+				return nil
+			}
+			if err := server.Close(); err != nil {
```
ysimonson: We probably want to use `server.Shutdown` instead

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,359 @@
...
+	masterLock := dlock.NewDLock(s.apiServer.env.GetEtcdClient(),
+		path.Join(s.apiServer.etcdPrefix,
+			s3gSidecarLockPath,
+			s.pipelineInfo.Pipeline.Name,
+			s.pipelineInfo.Salt))
+	ctx, err := masterLock.Lock(s.pachClient.Ctx())
+	if err != nil {
+		return err
+	}
+	// pachClient := s.pachClient.WithCtx(ctx)
```
ysimonson: Delete commented code

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,359 @@
...
+			if !ok {
+				// Note that because we call InspectJob after receiving a creation event, we
+				// may never create a service for a job that is created and then immediately
+				// deleted, so no error is returned if jobID is not in s.servers
+				return nil
+			}
+			if err := server.Close(); err != nil {
+				return fmt.Errorf("could not kill sidecar s3 gateway server for job %q: %v", jobID, err)
+			}
+			delete(s.servers, jobID)
```
ysimonson: if there is an error closing the server, it'll never be deleted from the map, resulting in a memory leak

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,359 @@
...
+						{
+							Port: int32(s.apiServer.env.S3GatewayPort),
+							Name: "s3-gateway-port",
+						},
+					},
+				},
+			}
+
+			if err := backoff.RetryNotify(func() error {
+				_, err := s.apiServer.env.GetKubeClient().CoreV1().Services(s.apiServer.namespace).Create(service)
```
ysimonson: Is it fair to assume that the kube account can create/delete services? I assume so since presumably the same account coordinates pipeline RC creation/deletion, but wanted to double-check.

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,359 @@
...
+	return nil
+}
+
+type JobHandler struct {
+	// Begin runs when a job is created
+	Begin func(ctx context.Context, jobInfo *pps.JobInfo) error
+
+	// End runs when a job ends
+	End func(jobID string) error
+}
```
ysimonson: It seems to me like it'd be more idiomatic to use an interface here

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,359 @@
...
+			}
+			if err := server.Close(); err != nil {
+				return fmt.Errorf("could not kill sidecar s3 gateway server for job %q: %v", jobID, err)
+			}
+			delete(s.servers, jobID)
+			return nil
+		},
+	})
+
+	return fmt.Errorf("sidecar s3 gateway: Serve() is exiting, which shouldn't happen")
```
ysimonson: would it make more sense to panic in this case?

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,359 @@
...
+				return err
+			}
+		}
+	}
+	return fmt.Errorf("job watcher loop was broken out of; this should not happen")
+}
+
+func (s *sidecarS3G) Client(authToken string) (*client.APIClient, error) {
+	newClient := s.apiServer.env.GetPachClient(s.pachClient.Ctx()) // clones s.pachClient
+	newClient.SetAuthToken(authToken)
```
ysimonson: This is presumably to match the client factory interface? A couple of things here:

1) I should remove the `authToken` parameter, and just call `SetAuthToken` on the client that's returned. It'll simplify the several implementations of client factory.
2) Got a recommended name instead of client factory? It's embarassingly java-esque :D

src/server/pps/server/worker_rc.go
```diff
@@ -466,10 +481,18 @@ func (a *apiServer) getWorkerOptions(ptr *pps.EtcdPipelineInfo, pipelineInfo *pp
 			}
 		}
 	}
+	var s3GatewayPort int32
+	if ppsutil.ContainsS3Inputs(pipelineInfo.Input) || pipelineInfo.S3Out {
+		// TODO(msteffen) do we need a separate option for the sidecar port?
+		// env.S3GatewayPort was originally added for main pachd.
```
ysimonson: I think not, we'll likely want to randomize the port the s3g sidecar instance

src/server/worker/api_server.go
```diff
@@ -551,8 +552,14 @@ func (a *APIServer) downloadData(pachClient *client.APIClient, logger *taggedLog
 			}
 		}
 	} else {
-		if err := os.MkdirAll(outPath, 0777); err != nil {
-			return "", err
+		if a.pipelineInfo.S3Out {
+			if err := os.RemoveAll(outPath); err != nil {
+				return "", fmt.Errorf("couldn't remove %q for s3-out pipeline: %v", outPath, err)
+			}
```
ysimonson: I'm assuming this is for `/pfs/out` -- if so, why the removal? Shouldn't it never be created in the first place?

src/server/worker/datum_iterator.go
```diff
@@ -40,9 +40,13 @@ func newPFSDatumIterator(pachClient *client.APIClient, input *pps.PFSInput) (Dat
...
 		return result, nil
 	}
+	impliedGlob := input.Glob
+	if input.S3 {
+		impliedGlob = "/"
+	}
 	fs, err := pachClient.GlobFileStream(pachClient.Ctx(), &pfs.GlobFileRequest{
 		Commit:  client.NewCommit(input.Repo, input.Commit),
-		Pattern: input.Glob,
+		Pattern: impliedGlob,
```
ysimonson: Instead of this glob fixing logic, should we just fail for any glob that's not `/` on s3 inputs? That may help users understand what's getting mounted in s3g.

src/server/pps/server/s3g_sidecar.go
msteffen: Fixed

src/server/pps/server/s3g_sidecar.go
msteffen: Done

src/server/pps/server/s3g_sidecar.go
msteffen: :sweat_smile: Done

src/server/pps/server/s3g_sidecar.go
msteffen: I updated this section of code. Let me know what you think. At this point, I think the only way that `delete` isn't reached is if the sidecar panics (which it does if it can't `Shutdown` the server correctly)

src/server/pps/server/s3g_sidecar.go
msteffen: Done

src/server/pps/server/s3g_sidecar.go
msteffen: Added a comment clarifying that yes, that is the goal :-).

I actually wonder if it might make sense for the s3 gateway to take `serviceenv.ServiceEnv` as a parameter, which implements `GetPachClient(ctx)`. Every pach client it returns is a clone, so there's no worry that calling `SetAuthToken` will somehow interfere with other code elsewhere.

src/server/pps/server/worker_rc.go
msteffen: Comment removed

src/server/worker/api_server.go
msteffen: This is a mistake—I was coding a bit defensively to try and get the PR out. In particular, I wasn't sure whether the `out` directory might be left behind from a previous datum but I now see that this is in a UUID subdirectory of the scratch directoy, so no, that can't happen, and I've gotten rid of the `RemoveAll` here and in `linkData`.

src/server/worker/datum_iterator.go
msteffen: That makes sense! Done.

src/server/pps/server/s3g_sidecar.go
ysimonson: Is there a way to trivially use `ServiceEnv` in tests? I'd need that for s3 integration testing.

src/server/pps/server/api_server.go
```diff
@@ -1549,8 +1566,11 @@ func (a *apiServer) validatePipelineRequest(request *pps.CreatePipelineRequest)
 	if request.TFJob != nil {
 		return goerr.New("embedding TFJobs in pipelines is not supported yet")
 	}
-	if request.S3Out {
-		return goerr.New("s3 gateway services for PPS pipelines are not supported yet")
+	if request.S3Out && ((request.Service != nil) || (request.Spout != nil)) {
+		return goerr.New("s3 output is not supported in spouts or services")
+	}
+	if request.S3Out && request.EnableStats {
+		return goerr.New("stats are not supported for pipelines that output via Pachyderm's S3 gateway")
```
ysimonson: how come?

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,475 @@
...
+
+type jobHandler struct {
+	s3g *sidecarS3G
+
+	// OnCreate runs when a job is created. Should be idempotent
+	OnCreate func(ctx context.Context, jobInfo *pps.JobInfo)
+
+	// OnTerminate runs when a job ends. Should be idempotent
+	OnTerminate func(ctx context.Context, jobID string)
+}
```
ysimonson: I still think this should be an interface, any reason for it not to be?

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,475 @@
...
+	finisher := <-done
+	panic(
+		fmt.Sprintf("sidecar s3 gateway: %s is exiting, which should never happen", finisher),
+	)
+}
+
+type jobHandler struct {
+	s3g *sidecarS3G
+
+	// OnCreate runs when a job is created. Should be idempotent
```
ysimonson:
```suggestion
	// OnCreate runs when a job is created. Should be idempotent.
```

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,475 @@
...
+	OnCreate func(ctx context.Context, jobInfo *pps.JobInfo)
+
+	// OnTerminate runs when a job ends. Should be idempotent
```
ysimonson:
```suggestion
	// OnTerminate runs when a job ends. Should be idempotent.
```

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,475 @@
...
+				s.pipelineInfo.Pipeline.Name,
+				s.pipelineInfo.Salt))
+		ctx, err := masterLock.Lock(s.pachClient.Ctx())
+		if err != nil {
+			// retry obtaining lock
+			return fmt.Errorf("error obtaining mastership: %v", err)
+		}
+
+		// Watch for new jobs & create kubernetes service for each new job
+		NewK8SServiceCreatingJobHandler(s).start()
```
ysimonson: is there any chance of this panicking, causing the dlock to be held?

src/server/pps/server/api_server.go
msteffen: We have no way of knowing which output files correspond to a given datum (in particular, in the case where you have a cross input and one of the cross elements is a regular PFS input, so that you have more than one datum but the workers are writing to `s3://out` throughout processing)

src/server/pps/server/s3g_sidecar.go
msteffen: Sorry, if that was an earlier comment I must have missed it; I apologize. I did it this way because there's no way to define interfaces inline in go. I'd have to define the interface, then define two types to implement the interface. And because this data structure is private (as an interface would also be) there is no way that code outside this package could implement an interface anyway. I therefore chose this approach because I estimated that it would be shorter without any downsides.

However, if you feel strongly I'm happy to change it.

src/server/pps/server/s3g_sidecar.go
msteffen: DLock expires 15 seconds after the last keepalive that it receives, so if the node dies another node will takeover mastership 15 seconds later: https://github.com/pachyderm/pachyderm/blob/master/src/server/pkg/dlock/dlock.go#L42

(if that weren't possible then master election would be quite brittle indeed :joy:)

src/server/pps/server/s3g_sidecar.go
msteffen: It's a great question, and fortunately the answer is yes: part of this PR is to create, in `pachctl deploy`, a service account with a service-creation-allowing role that worker pods are now assigned on creation and will use to do this.

It's significantly more restricted than the `pachyderm` service account that the `pachd` pods use, and is basically limited to being able to do this.

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+}
+
+func (h *jobHandler) start() {
+	defer func() {
+		logrus.Errorf("sidecar s3 gateway: start() is exiting; this should not happen")
+		panic(
+			fmt.Sprintf("!sidecar s3 gateway start() is exiting; this should never happen"),
+		)
+	}()
```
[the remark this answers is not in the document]
msteffen: This is a mistake on my part. I meant to replace the log with the panic. This should be fixed now

src/server/pps/server/s3g_sidecar.go
msteffen: I meant to comment on this, but `s3g_sidecar.go` uses `ListenAndServe` now, and handles errors from it

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,475 @@
...
+			}
+			return nil
+		}, backoff.NewInfiniteBackOff())
+		defer watcher.Close()
+
+		for e := range watcher.Watch() {
+			jobID := string(e.Key)
+			if e.Type == watch.EventError {
+				logrus.Errorf("sidecar s3 gateway watch error: %v", e.Err)
+				continue establish_watch
```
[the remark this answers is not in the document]
msteffen: I found this structure easy to follow visually when I was writing it (because if the `if` inside the `for`) but I really don't feel strongly about it so I changed it.

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,475 @@
...
+			}
+			return nil
+		}, backoff.NewInfiniteBackOff())
+		defer watcher.Close()
+
+		for e := range watcher.Watch() {
+			jobID := string(e.Key)
+			if e.Type == watch.EventError {
+				logrus.Errorf("sidecar s3 gateway watch job %q error: %v", jobID, e.Err)
+				continue establish_watch
```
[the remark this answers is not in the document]
msteffen: Same as above: I found this easier to follow visually when I was writing it because of the `if` inside the `for` but I don't feel strongly so I changed it.

src/server/pps/server/s3g_sidecar.go
msteffen: I'm marking this resolved because the code has changed a fair bit (it's now in `s3g_sidecar.go:254`) but rereading I feel confident that if the http server shuts down successfully, then it will be removed from `s.servers` (and if it doesn't shut down successfully then the sidecar will fully restart).

src/server/pps/server/s3g_sidecar.go
msteffen: Not yet, though I've discussed this with @Tryneus a bit. The goal is to create an interface that both `serviceenv.ServiceEnv` and `testing.RealEnv` satisfy, that can be passed in to all APIServers that need a client. We can punt on passing a `ServiceEnv` to s3 gateway until then.

It's a bit unfortunate that we have two data structures whose essential jobs are to pass a pachyderm client, but once we have a standard interface for both of them, I think there are many parts of our code that will be able to accept mock environments.

src/server/pps/server/s3g_sidecar.go
msteffen: Found it! Sorry, I think this comment was under at least two "show hidden" buttons. I did this refactoring, but it got a little bit strange because there are some methods of `jobHandler` that are meant to be different across the two implementations (`OnCreate` and `OnTerminate`) and some that aren't (`start`, `processJobEvent`, etc), and both access many of the same variables.

I just finished this refactoring and will push it so you can see what it looks like. If you think it's better we can keep it, but if not it's easy to drop. Just let me know.

src/server/pps/server/s3g_sidecar.go
ysimonson: That makes sense within the context of the current code. I'm still hesitant though because code changes, and there's a good chance this'll either be exported, or we'll need more implementations, or both. How about just leaving a comment on `jobHandler` saying this could be turned into an interface if more implementations are made?

etc/testing/images/s3testing/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+  apt-transport-https \
+  ca-certificates gnupg
+RUN python3 -m easy_install pip
+RUN pip3 install --upgrade pip
+RUN pip3 install awscli --upgrade
+RUN curl -L https://dl.min.io/client/mc/release/linux-amd64/mc >/usr/local/bin/mc \
+  && chmod ugo+x /usr/local/bin/mc
+RUN mkdir -p $HOME/.aws \
+  && printf "[default]\naws_access_key_id = \naws_secret_access_key = " >$HOME/.aws/credentials
+
```
ysimonson: Could you give this image another name? Without looking at the source code, I wouldn't be sure if this had to do with our testing of underlying PFS object stores, or s3g integration tests, or s3g conformance tests, or something else.

src/server/pps/server/s3g_sidecar.go
```diff
@@ -0,0 +1,482 @@
...
+	)
+}
+
+func (s *sidecarS3G) serveS3Instances() {
+	// Watch for new jobs & initialize s3g for each new job
+	(&handleJobsCtx{
+		s: s,
+		h: &s3InstanceCreatingJobHandler{s},
+	}).start()
+	panic("sidecar s3 gateway: serveS3Instances() is exiting, which shouldn't happen")
```
ysimonson: this panic is redundant, since exits are captured by the `done` channel

ysimonson:
LMK if you have any feedback on the s3g side

src/server/pps/server/s3g_sidecar.go
msteffen: Marking resolved, as this has been converted to an interface

etc/testing/images/s3testing/Dockerfile
msteffen: Done. This is now `ubuntu-with-s3-clients`

src/server/pps/server/s3g_sidecar.go
msteffen: Ah, thanks! Removed.

PR #4685, opened 2020-03-06
Fix formatting in Beginner Tutorial

svekars:


svekars:
slef-mreging as these are just formatting fixes.

Issue #4665, opened 2020-02-28
Audit & remove unused makefile targets


ysimonson:
#4663 will remove a bunch of makefile targets, but I only removed ones that appeared to be clearly unmaintained. There were other targets (e.g. `aws-test`) where it isn't clear whether they're still used and were kept.

At some point, we should audit & remove these unused makefile targets. There may also be some opportunities in trimming down the repo root `Makefile` by moving targets to `Makefile`s in subdirectories.

lukemarsden:
Relatedly, https://github.com/pachyderm/pachyderm/pull/5131 removes `integration-tests` target as it tripped me up when onboarding.

PR #4687, opened 2020-03-06
Shellcheck

ysimonson:
Applies [shellcheck](https://www.shellcheck.net/) to all of our scripts. The original impetus for this was the frequent employment of techniques that would work on linux but not macOS, or vice-versa, due to macOS' very old version of bash and general environment differences. This should help prevent those issues going forward, as well as catch general bugs in our scripts.

This is not completely comprehensive -- some of the trickier shellcheck warnings are punted on and are either entirely ignored, or have one-off instances ignored.

Builds off of #4663 because that had a number of changes to our shell scripts.


seslattery:
As a followup, we should ensure that most if not all scripts start with `set -euxo pipefail`. Rationale: https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/  I'm happy to open that in a separate issue if you don't think it belongs here.

ysimonson:
@seslattery that's a good idea - could you file a separate issue though? This PR has a high risk of not being merged in, because it's disruptive and it's not clear how much remaining work will be necessary to get it over the line.

chainlink:
Overall i think this is great, but this might be a good time to check in to see if we actually need/use some of these?

etc/proto/run.sh
```diff
@@ -17,14 +18,14 @@ fi
 
 go get github.com/gogo/protobuf/protoc-gen-gogofast
 
-cd ${GOPATH}/src/github.com/pachyderm/pachyderm
-for i in $(find src -name "*.proto"); do \
+cd "${GOPATH}/src/github.com/pachyderm/pachyderm"
+for i in src/**/*.proto; do \
     if ! grep -q 'go_package' "${i}"; then
         echo -e "\e[1;31mError:\e[0m missing \"go_package\" declaration in ${i}" >/dev/stderr
```
seslattery: We might want an `|| true` appended here, since we set -e and I am expecting we want to continue iterating through the loop on errors

etc/testing/deploy-manifests/validate.sh
```diff
@@ -21,27 +21,30 @@ fi
...
+# shellcheck disable=SC2034
 custom_args=(
 --cluster-deployment-id test
 --secure
 --dynamic-etcd-nodes 3
 --etcd-storage-class storage-class
 --namespace pachyderm
 --no-expose-docker-socket
---object-store=s3
+"--object-store=s3"
```
seslattery: Why the quotes here?

ysimonson: Shellcheck didn't like it for some reason, although I should be able to replace the `=` with a space and get rid of the quotes

etc/proto/run.sh
ysimonson: This is for the `grep` command, right? I actually assume the `if` would capture the exit code such that `set -e` wouldn't matter, but let me double-check.

ysimonson: Yeah, this is safe / will print `ok`:

```bash
#!/bin/bash
set -e
if ! which foobar; then
    echo "ok"
fi
```

ysimonson:
@chainlink agreed -- I wiped out a good number of them on the related PR (build streamlining), but it wasn't clear if some makefile targets (and the scripts that they call into) are used. I made an [open ticket](https://github.com/pachyderm/pachyderm/issues/4665) to address unused makefile targets (if any), and will update it to also note we should audit the scripts as well.

PR #4690, opened 2020-03-06
Check semvar

### LINKED ISSUE #4189: Incremental release names don't follow official sumver
Tags for incremental versions don't follow Semantic Versioning. In semver, pre-release use a dash `-` e.g. `1.0.0-alpha1` or `1.0.0-rc1` while incremental release should use the plus sign e.g. `1.0.1+d2f3a45`.

Currently these tags don't meet this requirement:

* `1.9.3-6221e9ed57a10b39d363e0e97cdd6e4e41796af2`
* `v1.9.4-6277238fa1e72d2adcbab36c2a829b0eb6309533`
* `v1.9.4-abbd633e9a33877386dafd891971d6abade0ad97`

The only obstacle to using semver is that docker doesn't accept plus sign in tag name. The most common way to address this is to replace the `+` with `_` in tag names.

So the version name would be `1.9.3+6221e9ed57a10b39d363e0e97cdd6e4e41796af2` while the docker tag would be `1.9.3_6221e9ed57a10b39d363e0e97cdd6e4e41796af2`.

nitinjainsj:
Enforce semvar release naming.

Fixes #4189

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ point-release:
 release-candidate:
 	@if [ "$(shell echo $(VERSION_ADDITIONAL) | head -c 1)" != "-" ]; \
 	then \
-	  echo "Specify VERSION_ADDITIONAL must start with a \"-\" character"; \
+	  echo "VERSION_ADDITIONAL must be specified and start with a \"-\" character. Eg. make VERSION_ADDITIONAL=-rc2 release-custom"; \
 	  exit 1; \
 	fi
 	@make release-helper
@@ -525,7 +525,7 @@ clean: clean-launch clean-launch-kube
 doc-custom: install-doc release-version
 	@if [ -n "$(VERSION_ADDITIONAL)" ] && [ "$(shell echo $(VERSION_ADDITIONAL) | head -c 1)" != "-" ]; \
 	then \
-	  echo "Specify VERSION_ADDITIONAL must start with a \"-\" character"; \
+	  echo "If VERSION_ADDITIONAL is specified, it must start with a \"-\" character. Eg make VERSION_ADDITIONAL=-rc2 doc-custom"; \
 	  exit 1; \
 	fi
 	./etc/build/doc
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -99,6 +99,11 @@ point-release:
 
 # Run via 'make VERSION_ADDITIONAL=rc2 release-custom' to specify a version string
 release-candidate:
+	@if [ "$(shell echo $(VERSION_ADDITIONAL) | head -c 1)" != "-" ]; \
+	then \
+	  echo "VERSION_ADDITIONAL must be specified and start with a \"-\" character. Eg. make VERSION_ADDITIONAL=-rc2 release-custom"; \
```
ysimonson: Super nitpick, but punctuation should be the same here vs the eg below

```suggestion
	  echo "VERSION_ADDITIONAL must be specified and start with a \"-\" character. Eg make VERSION_ADDITIONAL=-rc2 release-custom"; \
```

etc/build/doc
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-version="$($(GOPATH)/bin/pachctl version --client-only)"
+version="$($GOPATH/bin/pachctl version --client-only)"
```
ysimonson: Why this change? It would trigger a shellcheck warning.

nitinjainsj: I with the brackets, I get the following.

./etc/build/doc
./etc/build/doc: line 5: GOPATH: command not found
./etc/build/doc: line 5: /bin/pachctl: No such file or directory

Can you try it as well? (execute ./etc/build/doc or make VERSION_ADDITIONAL=-XYZ doc-custom) -- just to make sure something is not messed up on my side.

FWIW, this was a recent change I introduced with the last set of changes. I thought it had worked before I checked in. But since it did not work today, I am assuming I might have missed committing the right code version in the last merge.

ysimonson: Ahh that makes sense, LGTM

nitinjainsj:
I am going to hold off on this. We still need to address docker tag issue.

nitinjainsj:
Docker tags prevent us from being fully semvar compliant

PR #4704, opened 2020-03-11
Fix release doc to rebuild pachctl before regenerating manifests

msteffen:


doc/release_instructions.md
```diff
@@ -78,7 +78,7 @@ If you're doing a custom release (off a branch that isn't master), [skip to the
     > git push origin master
     ```
 
-8) Regenerate the golden deployment manifests: `./etc/testing/deploy-manifests/validate.sh --regenerate`
+8) Regenerate the golden deployment manifests: `make regenerate-test-deploy-manifests`
```
nitinjainsj: You have to pass VERSION_ADDITIONAL=-rc1 here?

msteffen: The reverse, we need to not set `VERSION_ADDITIONAL` at all. If `VERSION_ADDITIONAL` is set, then the golden manifests will contain whatever string is in `VERSION_ADDITIONAL`, but travis-generated manifests will not and the manifest tests will fail.

nitinjainsj: Oh wait I think you mean the exact opposite -- we should not pass VERSION_ADDITIONAL. That way the traget that get's built is only with 1.10.0 so travis-ci is happy. Is that's the case we should pass VERSION_ADDITIONAL=

Else the make file would take the commit it from the HEAD as VERSION_ADDITIONAL

msteffen: We actually don't need to set `VERSION_ADDITIONAL` to the empty string either (and travis doesn't when generating the test manifests—it doesn't set `VERSION_ADDITIONAL` at all), because `etc/testing/deploy-manifests/validate.sh` strips the commit hash from generated manifest files.

In general, my inclination would be to avoid setting `VERSION_ADDITIONAL` when generating the golden manifests, just because Travis doesn't set it for the test manifests

nitinjainsj: So I am not sure how would calling make target result in a different set of manifests than what we did with rc5, in which we directly called validate script.

msteffen: The make target has `install` as a make dependency, which causes it to rebuild `pachctl`. 

Because this line in the release_instructions doesn't set `VERSION_ADDITIONAL`, the dependency on `install` will rebuild `pachctl` with `VERSION_ADDITIONAL` unset (so it'll take the default value of the current commit hash). Then, `validate.sh` will generate new golden deploy manifests that include the commit hash, and strip out the commit hash with a regex.

Finally, when we run the deploy tests in travis, travis will similarly build pachctl with `VERSION_ADDITIONAL` unset, and then run `validate.sh`, which will generate test deploy manifests that include the commit hash, and strip the commit hash with a regex. If nothing changed in `pachctl deploy`, these should be identical to the golden deploy manifests because they were generated the same way

src/server/pkg/deploy/cmds/cmds.go
```diff
@@ -732,9 +732,10 @@ If <object store backend> is \"s3\", then the arguments are:
 
 			if namespace == "" {
 				kubeConfig := config.KubeConfig(nil)
+				var err error
 				namespace, _, err = kubeConfig.Namespace()
 				if err != nil {
-					log.Errorf("couldn't load kubernetes config (using \"default\"): %v", err)
+					fmt.Fprintf(os.Stderr, "WARNING: using namespace \"default\" (couldn't load namespace from kubernetes config: %v)\n", err)
```
ysimonson: Why this change? I'm trying to move us over to use logging on the client-side more

msteffen: There was some confusion in slack when Nitin was running the manifest test locally to understand the fix, and saw errors of the form `ERRO[0000] couldn't load kubernetes config (using "default"): invalid configuration: no configuration has been provided`. I had to go read the code to realize that the error was benign, so this was just me trying to make the error clearer. Specifcally:
1. I figured we shouldn't have an `ERRO[0000]` prefix, since that makes this look like a hard error, which it isn't
2. It now says `using namespace "default"`, which it didn't before ("default what?"). It also says it at the beginning, so it's clear what this error message is about
3. It prints to stderr so that it doesn't mess up this command's output.

I do think this is a clearer error that what we have (which I think is a low bar), but I cannot overstate my disinterest in dying on this hill, so if you want I can rebase this PR and drop this commit if you have a different way you want to fix this. Just let me know what you want. As well, if you overwrite this change in the next PR—great!

PR #4715, opened 2020-03-13
moved testing targets into Makefile.testing, created new ones, and cl…

pappasilenus:
…eaned up Makefile targets.  Please read Makefile.testing for instructions.

### DIFF
--- a/etc/examples/paths.sh
+++ b/etc/examples/paths.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+# these can be set by folks running the examples/Makefile.testing to use a different binaries than in the path
+ECHO=${ECHO:-echo}
+JQ=${JQ:-jq}
+KUBECTL=${KUBECTL:-kubectl}
+PACHCTL=${PACHCTL:-pachctl}
+RM=${RM:-rm}
+SLEEP=${SLEEP:-sleep}
+TAR=${TAR:-tar}
+WC=${WC:-WC}
+WGET=${WGET:-wget}
+
--- a/etc/examples/paths.sh
+++ b/etc/examples/paths.sh
@@ -1,12 +1,6 @@
 #!/bin/bash
 # these can be set by folks running the examples/Makefile.testing to use a different binaries than in the path
 ECHO=${ECHO:-echo}
-JQ=${JQ:-jq}
 KUBECTL=${KUBECTL:-kubectl}
 PACHCTL=${PACHCTL:-pachctl}
-RM=${RM:-rm}
-SLEEP=${SLEEP:-sleep}
-TAR=${TAR:-tar}
-WC=${WC:-WC}
-WGET=${WGET:-wget}
 
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,17 +1,9 @@
 .PHONY: opencv object_detection delete minikube gatk hyperparameter
 
-JQ := jq
-JQFLAGS :=
-DOCKER := docker
-ECHO := echo
-TOUCH := touch
-CP := cp
-RM := rm
-STATS := .enable_stats=true
-DETECTMODEL := .pipeline.name="object-detect"|.input.cross[1].pfs.repo="object-model"|.input.cross[1].pfs.name="model"
-HYPERPSPLIT := .input.cross[0].pfs.repo="hp-split"|.input.cross[0].pfs.name="split"
-HYPERPMODEL := .input.cross[1].pfs.repo="hp-model"|.input.cross[1].pfs.name="model"
-HYPERPTEST := .input.cross[0].pfs.repo="hp-test"|.input.cross[0].pfs.name="test"
+PACHCTL := /usr/local/bin/pachctl
+TAR := /usr/bin/tar
+WGET := /usr/local/bin/wget
+RM := /bin/rm
 
 
 GATK_GERMLINE_FILES	=	gatk/GATK_Germline/data/ref/ref.dict \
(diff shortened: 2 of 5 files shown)


examples/Makefile
```diff
@@ -43,155 +35,117 @@ GATK_GERMLINE_FILES	=	gatk/GATK_Germline/data/ref/ref.dict \
...
-	pachctl put file images@master:dogs.jpg -f ml/object-detection/images/dogs.jpg
+ml/object-detection/frozen_inference_graph.pb:
+	$(WGET) -p http://download.tensorflow.org/models/object_detection/ssd_mobilenet_v1_coco_11_06_2017.tar.gz --directory-prefix /tmp
+	$(TAR) -C ml/object-detection --strip-components 1 -xvf /tmp/download.tensorflow.org/models/object_detection/ssd_mobilenet_v1_coco_11_06_2017.tar.gz ssd_mobilenet_v1_coco_11_06_2017/frozen_inference_graph.pb
 
-object-detection:  object-detection-images-repo object-detection-demo
-	pachctl create pipeline -f ml/object-detection/model.json
-	pachctl create pipeline -f ml/object-detection/detect.json
+OBJECT_MODEL_PIPELINE = pachctl create pipeline -f ml/object-detection/model.json
+OBJECT_DETECT_PIPELINE = pachctl create pipeline -f ml/object-detection/detect.json
```
ysimonson: These are put in variables so you can override them in other targets, right? If so, I think it'd be more straight-forward to not use variables here, and just copy/paste as necessary.

examples/Makefile
```diff
@@ -1,17 +1,9 @@
...
-RM := rm
-STATS := .enable_stats=true
-DETECTMODEL := .pipeline.name="object-detect"|.input.cross[1].pfs.repo="object-model"|.input.cross[1].pfs.name="model"
-HYPERPSPLIT := .input.cross[0].pfs.repo="hp-split"|.input.cross[0].pfs.name="split"
-HYPERPMODEL := .input.cross[1].pfs.repo="hp-model"|.input.cross[1].pfs.name="model"
-HYPERPTEST := .input.cross[0].pfs.repo="hp-test"|.input.cross[0].pfs.name="test"
+PACHCTL := /usr/local/bin/pachctl
+TAR := /usr/bin/tar
+WGET := /usr/local/bin/wget
+RM := /bin/rm
```
ysimonson: Why set these to the specific executable paths? This won't work on, e.g. my linux desktop because:

1) `wget` is in `/usr/bin/wget`
2) `pachctl` is in `GOPATH`

examples/Makefile
```diff
@@ -43,155 +35,117 @@ GATK_GERMLINE_FILES	=	gatk/GATK_Germline/data/ref/ref.dict \
...
+	$(WGET) -p http://download.tensorflow.org/models/object_detection/ssd_mobilenet_v1_coco_11_06_2017.tar.gz --directory-prefix /tmp
+	$(TAR) -C ml/object-detection --strip-components 1 -xvf /tmp/download.tensorflow.org/models/object_detection/ssd_mobilenet_v1_coco_11_06_2017.tar.gz ssd_mobilenet_v1_coco_11_06_2017/frozen_inference_graph.pb
 
-object-detection:  object-detection-images-repo object-detection-demo
-	pachctl create pipeline -f ml/object-detection/model.json
-	pachctl create pipeline -f ml/object-detection/detect.json
+OBJECT_MODEL_PIPELINE = pachctl create pipeline -f ml/object-detection/model.json
+OBJECT_DETECT_PIPELINE = pachctl create pipeline -f ml/object-detection/detect.json
+object-detection-base: 
+	@$(OBJECT_IMAGES)
```
ysimonson: AFAICS there's no default set for `OBJECT_IMAGES`, so this won't work unless the variable is set

examples/Makefile.testing
```diff
@@ -0,0 +1,213 @@
+.PHONY:
+
+# This makefile is for creating clusters filled with repos & pipelines for end-to-end and functional testing.
```
svekars:
```suggestion
# This makefile is for creating clusters filled with repos and pipelines for end-to-end and functional testing.
```

examples/Makefile.testing
```diff
@@ -0,0 +1,213 @@
...
+# It builds on targets in the main Makefile.  You would invoke it like this
```
svekars:
```suggestion
# It builds on targets in the main Makefile.  You would invoke it by running the
```

examples/Makefile.testing
```diff
@@ -0,0 +1,213 @@
...
+# cd <source-location>/examples ;  make -s -f Makefile.testing dirty-testing-cluster
```
svekars:
```suggestion
# "cd <source-location>/examples  ; make -s -f Makefile.testing dirty-testing-cluster" commands
```

examples/Makefile.testing
```diff
@@ -0,0 +1,213 @@
...
+# That target, "dirty-testing-cluster" is the first workhorse testing target that's been developed.
+# It will create a cluster that has failed jobs, failed commmits, and deleted commits.
+# Invoking it with "-s" will produce terse, informative output on the steps it's taking to create the
+# cluster as it's working.  You might want to run a "watch -n 2 kubectl get pods" and a
+# "watch -n 2 pachctl list job" in other windows so you can understand what's happening as it runs.
```
svekars:
```suggestion
# "watch -n 2 pachctl list job" in other windows so that you can understand what's happening as it runs.
```

examples/Makefile.testing
```diff
@@ -0,0 +1,213 @@
...
+# At the end, it will print a message telling you the cluster is ready for your testing, and you
+# can use it to test extract/restore or for whatever other testing purposes you want.
+# There are some top-level functions defined, like pipeline-status and wait-for-jobs, that make it easier
```
svekars:
```suggestion
# There are some top-level functions defined, like "pipeline-status" and "wait-for-jobs", that make it easier
```

examples/Makefile.testing
```diff
@@ -0,0 +1,213 @@
...
+# to write targets that methodically create a cluster that's useful for testing.
+# The executable paths below are for macOS.  A future revision will set them correctly for other target OS's.
```
svekars:
```suggestion
# The executable paths below are for macOS.
```

Let's not talk about future revisions in the docs.

examples/Makefile.testing
```diff
@@ -0,0 +1,213 @@
...
+# Modify them for your system by setting environment variables, using a command like this
+# env PACHCTL=./pachctl  make -e  -f Makefile.testing dirty-testing-cluster
```
svekars:
```suggestion
# "env PACHCTL=./pachctl  make -e  -f Makefile.testing dirty-testing-cluster"
```

examples/Makefile.testing
```diff
@@ -0,0 +1,213 @@
...
+pipeline-status = $$(CURRENT_STATUS=`$(KUBECTL) get pod -l suite=pachyderm,pipelineName=$(1) \
+			-o jsonpath='{.items[*].status.conditions[?(@.type=="Ready")].status}'` ; \
+			if [ $$CURRENT_STATUS ] ; then $(ECHO) -n $$CURRENT_STATUS ; else $(ECHO) -n False ; fi)
+
+# wait-for-jobs will wait until there are no pachyderm jobs matching a jq filter condition
+# It takes 2 arguments.
+# $(1) the jq search filter: the filter '.state!="JOB_SUCCESS"' (without single quotes) will wait until all jobs are successful
+#                            the filter '.state!="JOB_SUCCESS" and .state!="JOB_FAILURE" will wait until all jobs have
+#                               either succeeded or failed.
+#                            You can put any jq filter in her that will work with the --raw output of list job to produce
```
svekars:
```suggestion
#                            You can put any jq filter that will work with the "--raw" output of "list job" to produce
```

examples/Makefile.testing
```diff
@@ -0,0 +1,213 @@
...
+#                               a list of job id's.
```
svekars:
```suggestion
#                               a list of job IDs.
```

examples/Makefile
pappasilenus: The problem is, for testing, PACHCTL may be in a variety of places, like when testing building a cluster in 1.9.12 and restoring it in 1.10.0, and confirming the same cluster builds properly in 1.10.0.  

I think I should set them to just the raw exectuable names, and I'll do that in an update.

I do want to add logic in a future version (outlined in the todo at top) to automatically set them to platform-specific versions.  I also  note, in Makefile.testing, how to override them when you invoke it.  I think I should add that here, as well.

examples/Makefile
pappasilenus: They're overridden in Makefile.testing.  I'm doing it as an exercise in DRY.  I note that I'm not using $(PACHCTL) here, either!

examples/Makefile
ysimonson: If you're looking to ensure a specific version of pachctl, and ensure you're always calling the correct path, you could:

1) first call `etc/fetch_release_pachctl.py`, e.g. `etc/fetch_release_pachctl.py 1.9.12`
2) reference the absolute path to `pachctl` in `GOPATH`

examples/Makefile
pappasilenus: Correct, it will be an empty operation, as intended.  The base- targets are intended to be built upon; I'm not sure how to make them not callable from outside.  That's currently beyond my make-fu. 

I'll make the fact that it's intended to be set by targets that *use* this one clearer.

examples/Makefile
ysimonson: Yeah DRY as a general principle makes sense, but I think it may be overdoing it here

examples/Makefile
pappasilenus: Oh, damn, that's awesome!  Yes, I'll add that.

examples/Makefile
pappasilenus: I really want to keep this, as it's easier to build variations on the pipelines by overriding the variables in new targets. Take a look at the variation I do in Makefile.testing, for example. I'd need to dupllicate lines, possibly introducing errors in the ordering in which pipelines are created, if I don't have this structure. It enforces uniformity in the testing structure.

ysimonson: This is what you have right now, between `Makefile` and `Makefile.testing`:

```
OBJECT_MODEL_PIPELINE = pachctl create pipeline -f ml/object-detection/model.json
OBJECT_DETECT_PIPELINE = pachctl create pipeline -f ml/object-detection/detect.json
object-detection-base: 
	@$(OBJECT_IMAGES)
	$(PACHCTL) create repo training
	@$(OBJECT_MODEL_PIPELINE)
	@$(OBJECT_DETECT_PIPELINE)

object-detection-testing-base: OBJECT_IMAGES = $(ECHO) "object-detection will be using opencv's images repo"
object-detection-testing-base: OBJECT_MODEL_PIPELINE = $(JQ) '$(STATS)|.pipeline.name="object-model"'  ml/object-detection/model.json | $(PACHCTL) create pipeline -f - >/dev/null
object-detection-testing-base: OBJECT_DETECT_PIPELINE = $(JQ) '$(STATS)|$(DETECTMODEL)'  ml/object-detection/detect.json | $(PACHCTL) create pipeline -f - >/dev/null
object-detection-testing-base:  object-detection-base
```

This is what it would look like without vars:

```
object-detection-base: 
    pachctl create repo images
    pachctl create repo training
    pachctl create pipeline -f ml/object-detection/model.json
    pachctl create pipeline -f ml/object-detection/detect.json

object-detection-testing-base: 
    echo "object-detection will be using opencv's images repo"
    pachctl create repo training
    jq '$(STATS)|.pipeline.name="object-model"'  ml/object-detection/model.json | pachctl create pipeline -f - >/dev/null
    jq '$(STATS)|$(DETECTMODEL)'  ml/object-detection/detect.json | pachctl create pipeline -f - >/dev/null
```

I think the latter is more straight-forward to understand, because there's less indirection, and it relies on less makefile magic. While DRY is generally a good principal to follow, it has been balanced with other principals.

Is there some other context I'm missing here? e.g. are you intending to create more `Makefile`s in the future that override these targets?

examples/Makefile
pappasilenus: OK, the paths are now neutral executables and I've added your instructions above.

examples/Makefile
pappasilenus: I've kept this in my latest commit for the reasons above.

examples/Makefile
pappasilenus: I've added the comments above.

examples/Makefile.testing
```diff
@@ -0,0 +1,211 @@
...
+JQ := jq
+JQFLAGS :=
+KUBECTL := kubectl
+PACHCTL := pachctl
+RM := rm
+SLEEP := sleep
+SORT := sort
+TOUCH := touch
+UNIQ  := uniq
+WC := wc
```
ysimonson: what's the utility of having override-able variables for standard unix tools?

examples/Makefile
pappasilenus: Yes, in the future there will be more targets and more makefiles.  So, for example, if I wanted to create a version where stats is turned  *off* for just one of the pipelines, it's two lines of code.

```
object-detection-testing-nostats: OBJECT_DETECT_PIPELINE = blah blah blah
object-detection-testing-nostats: object-detection-testing-base
```

examples/Makefile.testing
```diff
@@ -0,0 +1,211 @@
+.PHONY:
```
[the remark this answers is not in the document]
pappasilenus: I don't think I need it for any targets, so I can just remove it.

examples/Makefile
```diff
@@ -1,17 +1,14 @@
 .PHONY: opencv object_detection delete minikube gatk hyperparameter
```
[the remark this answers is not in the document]
pappasilenus: I should just remove this; it isn't needed,either.

examples/Makefile.testing
pappasilenus: There are sometimes subtle differences across platforms on the standard tools. For example, "echo" on macOS doesn't have the -e flag.  When I build these kind of files, I like the ability to call out the particular executables that'll be called, just in case.

examples/Makefile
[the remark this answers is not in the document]
pappasilenus: removed.

examples/Makefile.testing
pappasilenus: I've broken the complicated makefile functions out into what should be more understandable and reusable scripts in etc/examples.

I've removed most of these, leaving only those that have actual functional differences that could affect the running of the new scripts in etc/examples.

etc/examples/paths.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# these can be set by folks running the examples/Makefile.testing to use a different binaries than in the path
+ECHO=${ECHO:-echo}
```
ysimonson: What platform-specific behavior is being use with echo that requires this? All I see are the `-e` and `-n` flags being employed, both of which are supported by macOS.

etc/examples/paths.sh
```diff
@@ -0,0 +1,12 @@
...
+JQ=${JQ:-jq}
```
ysimonson: jq has no platform-specific behavior, and does not need to be overridden.

etc/examples/paths.sh
```diff
@@ -0,0 +1,12 @@
...
+KUBECTL=${KUBECTL:-kubectl}
```
ysimonson: I don't see any version-sensitive uses of kubectl that would require this.

etc/examples/paths.sh
```diff
@@ -0,0 +1,12 @@
...
+PACHCTL=${PACHCTL:-pachctl}
+RM=${RM:-rm}
```
ysimonson: AFAICS, this isn't being used

etc/examples/paths.sh
```diff
@@ -0,0 +1,12 @@
...
+SLEEP=${SLEEP:-sleep}
```
ysimonson: Definitely does not need to be overridden

etc/examples/paths.sh
```diff
@@ -0,0 +1,12 @@
...
+TAR=${TAR:-tar}
```
ysimonson: What platform-specific stuff is being used for tar?

etc/examples/paths.sh
```diff
@@ -0,0 +1,12 @@
...
+WC=${WC:-WC}
```
ysimonson: Not using anything platform-specific.

etc/examples/paths.sh
```diff
@@ -0,0 +1,12 @@
...
+WGET=${WGET:-wget}
```
ysimonson: I'm not aware of any platform differences for wget, could you elucidate?

examples/Makefile.testing
```diff
@@ -0,0 +1,178 @@
...
+		$(PACHCTL) put file samples@master:son/son.bam -f gatk/GATK_Germline/data/bams/son.bam ; \
+		$(PACHCTL) put file samples@master:son/son.bai -f gatk/GATK_Germline/data/bams/son.bai	; \
+		$(PACHCTL) finish commit samples@master > /dev/null ; \
+		$(ECHO) -e "Deleting father commit, samples@$${GATK_FATHER_COMMIT}..." ; \
+		$(PACHCTL) delete commit samples@$${GATK_FATHER_COMMIT} ; \
+	else \
+		$(ECHO) -e "Error: could not create commit in samples@master." ; \
+		exit 1 ; \
+	fi
+	../etc/examples/wait-for-jobs.sh succeeded-only "All jobs finished. Inspect output above to determine if cluster ready for testing."
```
ysimonson: I think once there's control structures, the makefile target should be broken out into a separate script

etc/examples/paths.sh
pappasilenus: -e isn't supported on macOS.  You need to support gnubin utils to get it.

etc/examples/paths.sh
pappasilenus: rm is used by "make clean"

etc/examples/paths.sh
pappasilenus: sleep has different behavior in the gnu utilities than in macOS.  It allows arguments in the form of different units, including minutes and hours, vs the seconds-only approach of the macOS version.

etc/examples/paths.sh
pappasilenus: This is to allow *different versions* of kubectl to be easily used by testing routines down the road, same as the specification of pachctl.

etc/examples/paths.sh
pappasilenus: Same armoring as above.  This makes it easier to switch versions in a tool that could have cross-platform subtleties.

examples/Makefile.testing
pappasilenus: That's an interesting approach.  I'm not sure I agree with it fully.  There are circumstances, like here and on lines 102-116, where a simple control structure in the makefile makes it easy to fully understand the flow of the test.   The creation of the father and son commits, and then the deletion of the father commit, are part of this  cluster setup, not a function that can be called again and again.

So, in this case (and at the beginning of the target on lines 102-116), breaking it out makes the target _less_ understandable.  The flow is an integral part of this particular test, so the logic belongs in the test.

etc/examples/paths.sh
pappasilenus: wget is a tool that would need to be installed, and this script may not run where it's in the PATH.  I don't want to rely on that, and I don't want to require that a PATH be changed to run this for tools that must be installed, for security reasons.  That's why I allow folks to specify the install location of tools that aren't part of a standard buid.

etc/examples/paths.sh
pappasilenus: See comment below: https://github.com/pachyderm/pachyderm/pull/4715#discussion_r397365325

etc/examples/paths.sh
pappasilenus: It's not installed by default on macOS, and can be installed by a user anywhere.  See https://github.com/pachyderm/pachyderm/pull/4715#discussion_r397365325

etc/examples/paths.sh
ysimonson: `man echo` on macOS implies otherwise; it gives the same description for `-e` as the ubuntu variant

etc/examples/paths.sh
ysimonson: That's called by the redundantly specified `RM` in the makefile

etc/examples/paths.sh
pappasilenus: Hmmm... Catalina shows just the -n flag supported, and using "-e" with ANSI codes doesn't do anything.  The gnu version works correctly.   What macOS are you on?

ysimonson: Catalina

ysimonson:
I'm having a lot of trouble getting behind the "vars as executable paths" concept going on here. It may make sense in the case of `pachctl`, where it's not uncommon for users to have multiple versions installed in multiple paths, but that's about it AFAICS. Do you have _concrete_ examples where this has come up for our users, or when demoing pachyderm?

Reasons for my reservations:
1. It adds a level of indirection that breaks away from _both_ shell built-ins (e.g., in the case of `echo`) and `PATH`. If I were looking at this without the context of the conversation, I would be _very_ confused about what was going on.
2. The variations between platforms for POSIX executables is minor enough that it's usually possible to write calls to work cross-platform. This is what we do everywhere else in the codebase to support both macOS and linux. It puts less onus on the consumer, because they don't need to understand this indirection, or know how to correctly configure the vars to work.
3. Frequently you're adding this indirection in places where it's not needed, in anticipation of possible future need. I think this is _too_ defensive, because we can always modify this in the future to add the indirection if necessary.

etc/examples/paths.sh
pappasilenus: Interesting.  Different results on my Catalina, where I'm using bash. The built-in echo on zsh understands the -e flag, but bash uses /bin/echo, which doesn't.  Which supports my case: even different shells will give different results, so giving some degree of control over this to the invoker of the makefile makes sense.

etc/examples/paths.sh
pappasilenus: It's not redundant; it's specified in there so it can be passed down to any scripts that are invoked, such as these.  Once it's specified, it's in the env.  The paths.sh will only set if not specified by the env, and the Makefile is asserting control over them so that the invoker of the script can control them.

pappasilenus:
> I'm having a lot of trouble getting behind the "vars as executable paths" concept going on here. It may make sense in the case of pachctl, where it's not uncommon for users to have multiple versions installed in multiple paths with non-overlapping functionality, but that's about it AFAICS. Do you have concrete examples where this has come up for our users, or when demoing pachyderm?

I understand what you’re saying. If I were building a system that’s designed to _overcome_ the differences between each platform to provide a base functionality, rather than _exploit_ them for testing purposes, I would agree with you.  But what I’m building here is a system that's designed to break things in a consistent way across inconsistent systems as well as provide a foundation for building on it.  Let me go through your reservations one by one.

> Reasons for my reservations:
>
> 1. It adds a level of indirection that breaks away from both shell built-ins (e.g., in the case of echo) and PATH. If I were looking at this without the context of the conversation, I would be very confused about what was going on, because this setup is fairly atypical.

I think this goes back to the use case for a system designed to *accommodate* those differences vs. one designed to _exploit_ them for testing purposes.

This system is designed with two goals in mind
1. To create the exact same cluster while easily varying the conditions under which it’s created: the platform *and* the tools
2. To easily and consistently create new variations on base testing clusters.

As you can see from the above, there are two target users here: The first is someone who will be asked to run these tests on a system they may have limited control over, with limited ability to install and update software to see what works.  The second is a Pachyderm QA resource, who will need to create a consistent cluster for testing while building on the clusters that have been created.

> 2. The variations between platforms for POSIX executables is minor enough that it's usually possible to write calls to work cross-platform. This is what we do everywhere else in the codebase to support both macOS and linux. It puts less onus on the consumer, because they don't need to understand this indirection, or know how to correctly configure the vars to work.

The purpose of the QA tester is to not only accommodate the differences between systems but to exploit them to test.  Giving it customization points that allow for a mix-and-match approach allows for one script to be easily customized for a variety of conditions, or to easily ask a user to customize it without having to modify the system they are running on. For example, a user may not be able to deploy new versions of kubectl even though we want to test this with a particular version.  Telling them "just put it in your home directory and run it like this: `env KUBECTL=./kubectl make -e blah blah blah`"  is a very easy instruction vs the alternative.

> 3. Frequently you're adding this indirection in places where it's not needed, in anticipation of possible future need. I think this is too defensive, because we can always modify this in the future to add the indirection if necessary.

I can understand that. I added them as I ran into issues, after some thought about "how would I ask a user to run this in their environment, if they had limited access to a system and limited expertise?"  I wanted to be able to sent them one command line that would make it work as desired, and test with different components. After deciding that I had run into enough problems that I wanted to isolate it completely,  I did take your advice pare them down to what I think is the minimum for running tests across a variety of platforms (I haven’t tested WSL yet, but I’m curious about it, and I think what I have will cover it.) 

To summarize: writing QA resources that allow for the consistent breaking of things requires slightly different patterns than writing cross-platform code that works everywhere.  It has to be written to both work _everywhere_, but allow for ways to consistently break it _anywhere_.

ysimonson:
I'm skeptical that this is the right way to isolate differences in systems, as it requires a great deal of knowledge on the part of the QA tester, but won't belay the point any further.

At the very least, please walk back indirection to `rm` and `sleep` - it's not necessary at the present time, and we can always add it in if necessary.

examples/Makefile
```diff
@@ -1,18 +1,12 @@
-.PHONY: opencv object_detection delete minikube gatk hyperparameter
```
ysimonson: 1) Entirely removing the phony targets could trigger a bug, if there's a file with the name of any of these targets. It's unlikely, but I think worth covering.
2) You should set the shell on makefiles as well, like so: https://github.com/pachyderm/python-pachyderm/blob/0d464d83b381dbe0248eedf0f22fe9703678bedd/Makefile#L1

pappasilenus: First, thought long about my arguments vs. yours, and I'll walk back every one but kubectl, pachctl & echo.  The others I put in there in _anticipation_ of conflicts, and, you're right, the added complexity is not worth the anticipated need.

1. I tested for that; there aren't files with those names, just directories.  It works as we expect.
2. Great tip, thank you!

pappasilenus: new commit should address these issues.  Thanks for all your help making this better!

PR #4716, opened 2020-03-16
kubeflow mnist demo

ysimonson:


examples/kubeflow/mnist/Makefile
```diff
@@ -0,0 +1,12 @@
+.PHONY: docker-build docker-push mnist
+
+docker-build:
+	docker build -t ysimonson/mnist_pachyderm_pipeline:v1.1.9 .
```
nitinjainsj: I am assuming "ysimonson" is only for testing right now.

ysimonson: Yes, will swap that over once changes to the image are done

ysimonson:
Which download are you referring to?

* It can't fail to download `mnist.npz`, because it's stored in the repo. I think it would be better to not store this in the repo since it's binary data, but it was already in there.
* If `boto3` fails to download, the kubeflow pipeline will fail, causing the pachyderm pipeline to fail.
* If there is an issue downloading or uploading to s3g, the kubeflow pipeline will fail, causing the pachyderm pipeline to fail.

PR #4721, opened 2020-03-18
Fix path name

ysimonson:
This should fix the netlify docs build errors

### DIFF
--- a/doc/build.sh
+++ b/doc/build.sh
@@ -25,6 +25,8 @@ cat <<EOF >material/partials/versions.html
         <option style="color:white;background-color:#4b2a5c;" value="latest">latest (${latest_version})</option>
 EOF
 for d in docs/*; do
+    d=$(basename "${d}")
+    
     # don't rebuild archive dir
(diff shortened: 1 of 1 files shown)


doc/build.sh
```diff
@@ -48,6 +48,8 @@ EOF
 
 # Rebuild all docs versions
 for d in docs/*; do
+    d=$(basename "${d}")
+
     # don't rebuild archive dir
```
svekars: lines 53 - 60 don't work now 
Screen Shot 2020-03-18 at 7 54 47 AM

It should exclude `docs/` and the `master` and `archive` dirs:

Screen Shot 2020-03-18 at 7 57 39 AM

ysimonson:
Do you know where that menu is generated? Because the build console is indicating that this change fixed at least the aforementioned issue: https://app.netlify.com/sites/pachyderm-docs/deploys/5e722e90518f790009610675

ysimonson:
Ah! I see, there's two loops - going to try another fix.

svekars:
Looking good now!

PR #4725, opened 2020-03-20
Removed stray vendored file

ysimonson:


ysimonson:
@adelelopez this is ready for review now; the netlify build status indicates it's still in progress, but that's a bug on their end, it built fine

PR #4729, opened 2020-03-23
Update certs on image build

### LINKED ISSUE #4722: Include `apt install ca-certificates` during docker build process
By not updating the certificate bundles, we are relying on the upstream golang container to have a build process that is providing us with the most recent certs.  Their is a potential for our certs in our containers to not be up to date causing errors going forwards.

ysimonson:
Closes #4722

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 # syntax=docker/dockerfile:1.0-experimental
 ARG GO_VERSION
 FROM golang:${GO_VERSION}
+RUN apt update && apt install ca-certificates
 RUN go get github.com/go-bindata/go-bindata/...
 WORKDIR /app
 COPY . .
--- a/etc/reset.py
+++ b/etc/reset.py
@@ -19,6 +19,8 @@ DELETABLE_RESOURCES = [
     "secrets",
     "clusterrole",
     "clusterrolebinding",
+    "roles.rbac.authorization.k8s.io",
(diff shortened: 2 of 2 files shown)


etc/reset.py
```diff
@@ -19,6 +19,8 @@
     "secrets",
     "clusterrole",
     "clusterrolebinding",
+    "roles.rbac.authorization.k8s.io",
```
seslattery: I don't have context on why this change is necessary, but at first glance it seems fine.

Dockerfile
```diff
@@ -1,6 +1,7 @@
 # syntax=docker/dockerfile:1.0-experimental
 ARG GO_VERSION
 FROM golang:${GO_VERSION}
+RUN apt update && apt install ca-certificates
```
seslattery: Might be overkill for single package, but `&& rm -rf /var/lib/apt/lists/*` is best practice.

ysimonson: I don't think that makes sense here, because this is a throwaway build container, i.e. the image isn't pushed, and only selective assets from the image are copied into other images.

etc/reset.py
ysimonson: You can use `./etc/reset.py` to fully reset a development deployment of pachyderm. It looks like there were some recent changes on docker for mac that now make it necessary to delete roles and rolebindings, and it was blocking testing of these changes.

PR #4768, opened 2020-04-03
fixed links to kops documentation

pappasilenus:


svekars:
LGTM, but there is no published version of these docs? Only GitHub ?

PR #4782, opened 2020-04-09
Fix pipeline restart

msteffen:


msteffen:
> @msteffen is this change complete and you want us to do a review or are did you want to add some tests?

Just added some tests (or rather, I extended `TestUpdatePipeline` to now check that pipeline resources are updated after `UpdatePipeline`), and this PR is now fully ready to review

PR #4786, opened 2020-04-10
Allow mnist pipeline to run standalone

msteffen:


examples/kubeflow/mnist/pipeline.py
```diff
@@ -82,21 +88,124 @@ def kubeflow_pipeline(s3_endpoint: str, input_bucket: str):
...
+                    'rather than uploading a new pipeline')
+parser.add_argument('--create_pipeline',
+                    default="", metavar="pipeline_name", type=str,
+                    help='If set, this script will create a kubeflow pipeline '+
+                    'with the given name, but will not create a run in that '+
+                    'pipeline. Used to create a pipeline from a local machine.')
+parser.add_argument('--force', action='store_true', default=False,
+                    help='If set, and this script tries to create a pipeline '+
+                    'where a kubeflow pipeline with the same name already '+
+                    'exists, the existing pipeline will be deleted')
```
ysimonson, line 110: 1) Put `parser` either in the `if __name__ == '__main__` block, and pass the arguments into the main method. This avoids globals and makes it easier in the future to hook into main outside of the CLI.
2) Arguments should use dashes instead of underscores. e.g. `--remote-host`. Python will automatically represent that as the `remote_host` attribute on args.
3) Use multiline strings for help `"""like this"""`

examples/kubeflow/mnist/pipeline.py
```diff
@@ -82,21 +88,124 @@ def kubeflow_pipeline(s3_endpoint: str, input_bucket: str):
...
+
+def pipeline_id(client : kfp.Client, name : str):
```
ysimonson, line 112: very nit, remove the spaces between the variable name and the the colon, e.g. `client: kfp.Client`. This is in keeping with how I've generally seen type signatures.

examples/kubeflow/mnist/pipeline.py
```diff
@@ -82,21 +88,124 @@ def kubeflow_pipeline(s3_endpoint: str, input_bucket: str):
...
+    if args.create_run_in != "":
+        logging.info('creating run in pipeline "{}"'.format(args.create_run_in))
+        pid = pipeline_id(client, args.create_run_in)
+        if pid == "":
+            logging.error('could not find pipeline "{}" to create job'.format(
+                args.create_run_in))
+            sys.exit(1)
+        # Create a run in the target pipeline using the new pipeline ID
+        run_info = client.run_pipeline(
+            job_name="pach-job-" + os.environ["PACH_JOB_ID"],
```
ysimonson, line 169: Use format strings, or `.format` here

examples/kubeflow/mnist/pipeline.py
```diff
@@ -82,21 +88,124 @@ def kubeflow_pipeline(s3_endpoint: str, input_bucket: str):
...
+            params = {
+                "s3_endpoint": os.environ["S3_ENDPOINT"],
+                "input_bucket": "input",
+            }
+        )
+        run_id = run_info.id
+    elif args.create_pipeline != "":
+        # Local machine is just creating the pipeline
+        try:
+            (_, pipeline_package_path) = tempfile.mkstemp(suffix='.zip')
```
ysimonson, line 181: Use `tempfile.NamedTemporaryFile` as a context manager to guarantee cleanup, e.g. see here https://docs.python.org/3/library/tempfile.html#examples

examples/kubeflow/mnist/pipeline.py
```diff
@@ -82,21 +88,124 @@ def kubeflow_pipeline(s3_endpoint: str, input_bucket: str):
...
+    if args.remote_host != "":
+        client = kfp.Client(host=args.remote_host)
+    else:
+        client = kfp.Client()
+
+    run_id = ""
+    if args.create_run_in != "" and args.create_pipeline != "":
+        logging.error('only one of --create_run_in and '+
+              '--create_pipeline may be set')
+        sys.exit(1)
```
ysimonson, line 159: Rather than calling sys.exit, have main return the process' rc, then below do this:

```
if __name__ == "__main__":
    sys.exit(main())
```

examples/kubeflow/mnist/pipeline.py
msteffen: Done.

examples/kubeflow/mnist/pipeline.py
msteffen: Done

examples/kubeflow/mnist/pipeline.py
msteffen: Done

examples/kubeflow/mnist/pipeline.py
msteffen: :+1: I also had to pass the `.name` field to the KF client, which I *think* was correct:
```
with tempfile.NamedTemporaryFile(suffix='.zip') as pipeline_file:
    compiler.Compiler().compile(kubeflow_pipeline, pipeline_file.name)
```

examples/kubeflow/mnist/pipeline.py
msteffen: Done

examples/kubeflow/mnist/pipeline.py
```diff
@@ -82,21 +88,124 @@ def kubeflow_pipeline(s3_endpoint: str, input_bucket: str):
...
+    while True:
+        p = client.list_pipelines(page_token=page_token, page_size=100)
+        if p.pipelines is None:
+            return ""
+        for p in p.pipelines:
+            if p.name == name:
+                return p.id
+        if p.next_page_token is None:
+            return ""
+        page_token = p.next_page_token
```
ysimonson, line 109: this could be simplified a little bit:

```python
    page_token = ""
    while page_token is not None:
        p = client.list_pipelines(page_token=page_token, page_size=100)
        if p.pipelines is None:
            return ""
        for p in p.pipelines:
            if p.name == name:
                return p.id
        page_token = p.next_page_token

    return ""
```

examples/kubeflow/mnist/pipeline.py
```diff
@@ -82,21 +88,124 @@ def kubeflow_pipeline(s3_endpoint: str, input_bucket: str):
...
+    """
+    page_token = ""
+    while True:
+        p = client.list_experiments(page_token=page_token, page_size=100)
+        for p in p.experiments:
+            if p.name == name:
+                return p.id
+        if p.next_page_token is None:
+            return ""
+        page_token = p.next_page_token
```
ysimonson, line 126: similar simplification could be made on this loop

examples/kubeflow/mnist/pipeline.py
```diff
@@ -82,21 +88,124 @@ def kubeflow_pipeline(s3_endpoint: str, input_bucket: str):
...
+
+def main(host: str, create_pipeline: str, create_run_in: str, force: bool):
+    if create_pipeline != "" and create_run_in != "":
+        logging.error('only one of --create_run_in and --create_pipeline may be set')
```
ysimonson, line 130:
```suggestion
        logging.error('only one of --create-run-in and --create-pipeline may be set')
```

examples/kubeflow/mnist/pipeline.py
msteffen: Done

examples/kubeflow/mnist/pipeline.py
msteffen: Done

PR #4797, opened 2020-04-16
Fix small NPE in metrics.go

msteffen:
cc @actgardner


msteffen:
Merging without waiting for CI—Nitin and I decided we were okay with the technical risk because this is a two-line PR, and I tested it manually on my local machine (probably not a decision we should make habitually, but we want to get the release out)

PR #4798, opened 2020-04-16
Fix NPE in metrics.go

msteffen:


Tryneus:
I just merged an identical PR in #4844 - closing this one.  It looks like this fix was merged to `1.10.x` but not `master`.

PR #4799, opened 2020-04-16
Modernize CI environment

ysimonson:
Pulling some inspiration from work being done in https://github.com/pachyderm/jupyterhub-pachyderm/pull/49, it looks like we should be able to update the CI environment to something newer relatively painlessly

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: xenial
+dist: bionic
 sudo: required
 cache:
   directories:


etc/kube/start-minikube.sh
```diff
@@ -3,22 +3,10 @@
...
-  case "${opt}" in
-    v)
-      VERSION="v${OPTARG}"
-      ;;
-    \?)
-      echo "Invalid argument: ${opt}"
-      exit 1
-      ;;
-  esac
-done
```
ysimonson, line 21: This was removed because we're ignoring the flag anyways

etc/testing/travis_install.sh
```diff
@@ -21,7 +21,9 @@ sudo apt-get install -y -qq \
   python3-pip \
   python3-setuptools \
   pkg-config \
-  fuse
+  fuse \
+  conntrack \
+  pv
```
ysimonson, line 26: Dependencies were added because apparently they're not built-in in bionic

ysimonson:
This is turning out to be more work than I anticipated, so closing out

PR #4806, opened 2020-04-17
Add long description for exit command

actgardner:
Fix CI on master by adding a long description for the exit command.


actgardner:
CI is passing so I'm going to merge this to fix master, but I'm happy to do a follow-up to remove the exitr command instead.

PR #4817, opened 2020-04-22
Update the doc theme

svekars:
- Import the doc theme in the config file instead of keeping in the repo.
- Update the copyright year.
- Fix a few UI bugs.


ysimonson:
You should be able to remove the assets directory now

PR #4839, opened 2020-04-30
Set commit Started/Finished during Extract and Restore

msteffen:


src/server/pfs/server/driver.go
```diff
@@ -1173,6 +1175,37 @@ func (d *driver) makeCommit(
...
+	// 2) 'started' <= 'finished'
+	switch {
+	case started.IsZero() && finished.IsZero():
+		// set 'started' to Now() && leave 'finished' unset (open commit)
+		started = time.Now()
+	case started.IsZero() && !finished.IsZero():
+		// set 'started' to 'finished'
+		started = finished
+	case !started.IsZero() && finished.IsZero():
+		if now := time.Now(); now.Before(started) {
```
kevindelgado, line 1189: What conditions would this ever happen where started is before now? If it's only in errors, would we want to fail loudly instead of silently modifying started?

src/server/pfs/server/driver.go
```diff
@@ -1173,6 +1175,37 @@ func (d *driver) makeCommit(
...
+			started = now // prevent finished < started (if user finishes commit soon)
+		}
+	case !started.IsZero() && !finished.IsZero():
+		if finished.Before(started) {
+			started = finished // prevent finished < started
```
kevindelgado, line 1194: same as above, would we rather fail loudly here?

src/server/pfs/server/driver.go
msteffen: The main case where I could see this happening is time skew, but I see your point. My feeling is that this shouldn't derail a `Restore()` operation (which returning an error would do), but I did change this to log a warning if encountered

src/server/pfs/server/driver.go
msteffen: As above, log a warning if this block is encountered

src/server/admin/server/api_server.go
```diff
@@ -170,6 +170,11 @@ func (a *apiServer) Extract(request *admin.ExtractRequest, extractServer admin.A
 			if ci.ParentCommit == nil {
 				ci.ParentCommit = client.NewCommit(ci.Commit.Repo.Name, "")
 			}
+			// Restore must not create any open commits (which can interfere with
+			// restoring other commits), so started and finished are always set
+			if ci.Finished == nil {
```
kevindelgado, line 175: What happens if you have call extract on a cluster that has legitimately open commits that the user just hasn't closed yet. Is this possible, and if so would it be good to warn the user like "hey, you tried to extract a cluster with open commits, we closed them for you." or something?

msteffen: Per face-to-face discussion, I'm happy to add a warning. My current feeling is that it's fine—I think this would come up if extract/restore ran while a user was loading data, which would be difficult to make work (we'd have to extract data as they were loading it)

PR #3939, opened 2019-07-24
[#2550][DOC] Add instruction for EKS deployment

svekars:
* Edited the AWS deployment section.
* Added the EKS deployment instructions
.

PR #4840, opened 2020-04-30
Set commit Started/Finished during Extract and Restore

msteffen:
Like #4839 but applied to 1.10.x


msteffen:
As this PR is almost identical to #3939 (with the exception of not using the `errors` library, which only exists in master), I'm merging it on the approval of that PR

PR #4861, opened 2020-05-06
Allow standby pipelines to be stopped (like #4860 for 1.10.x)

msteffen:


msteffen:
Accidentally created this PR in the wrong git branch. Deleting and recreating in the correct branch...

PR #4864, opened 2020-05-06
Add retry loop to compactWorker

### LINKED ISSUE #4855: Compaction worker code should be wrapped in a retry with an exponential backoff
The compaction worker code (specifically the run function call here https://github.com/pachyderm/pachyderm/blob/03f22b5ebec7f56813fced4f7215de46c50ccf20/src/server/pfs/server/driver_new_storage.go#L348) should be wrapped in a retry with an exponential backoff. The compaction worker runs in a goroutine in Pachd and is responsible for claiming and compacting compaction shards. We have functions that implement retries with backoffs here https://github.com/pachyderm/pachyderm/blob/master/src/server/pkg/backoff/retry.go. The function that should be used is https://github.com/pachyderm/pachyderm/blob/03f22b5ebec7f56813fced4f7215de46c50ccf20/src/server/pkg/backoff/retry.go#L29, and the notify should log the error (this can just be a print for the time being).

brendoncarroll:
If compactWorker encounters an error while talking to etcd, it needs to
retry with backoff and never exit.
If the Retry function ever returns, we panic because compaction would no longer happen, and we should never be in that state.

Closes #4855

### DIFF
--- a/src/server/pfs/server/driver_new_storage.go
+++ b/src/server/pfs/server/driver_new_storage.go
@@ -8,6 +8,7 @@ import (
 	"path"
 	"regexp"
 	"strconv"
+	"time"
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/gogo/protobuf/types"
@@ -340,12 +341,19 @@ func deserializeShard(shardAny *types.Any) (*pfs.Shard, error) {
 }
 
 func (d *driver) compactionWorker() {
+	ctx := context.Background()
 	w := work.NewWorker(d.etcdClient, d.prefix, storageTaskNamespace)
+
 	backoffStrat := backoff.NewExponentialBackOff()
 	backoffStrat.MaxElapsedTime = 0 // don't give up, compactionWorker should not exit
 
-	err := backoff.Retry(func() error {
-		err := w.Run(context.Background(), func(ctx context.Context, subtask *work.Task) error {
+	notifier := func(err error, t time.Duration) error {
+		log.Printf("error in compaction worker: %v", err)
+		return nil // non-nil shuts down retry loop
+	}
+
+	err := backoff.RetryNotify(func() error {
+		return w.Run(ctx, func(ctx context.Context, subtask *work.Task) error {
 			shard, err := deserializeShard(subtask.Data)
 			if err != nil {
 				return err
@@ -356,11 +364,7 @@ func (d *driver) compactionWorker() {
 			}
 			return d.storage.Compact(ctx, shard.OutputPath, shard.Compaction.InputPrefixes, index.WithRange(pathRange))
 		})
-		if err != nil {
-			log.Printf("error in compaction worker: %v", err)
-		}
-		return err
-	}, backoffStrat)
+	}, backoffStrat, notifier)
 
 	panic(err) // never ending backoff should prevent us from getting here.
 }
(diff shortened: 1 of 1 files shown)


src/server/pfs/server/driver_new_storage.go
```diff
@@ -339,23 +339,28 @@ func deserializeShard(shardAny *types.Any) (*pfs.Shard, error) {
...
-// notification for the user (outside of the log message).
-// (bryce) ^ maybe just a retry would be good enough.
 func (d *driver) compactionWorker() {
 	w := work.NewWorker(d.etcdClient, d.prefix, storageTaskNamespace)
-	if err := w.Run(context.Background(), func(ctx context.Context, subtask *work.Task) error {
-		shard, err := deserializeShard(subtask.Data)
+	backoffStrat := backoff.NewExponentialBackOff()
+	backoffStrat.MaxElapsedTime = 0 // don't give up, compactionWorker should not exit
+
+	err := backoff.Retry(func() error {
```
brycemcanally, line 347: We should probably use the `RetryNotify` function and have the logging be done in the notify callback. This is the pattern we follow in the codebase because it is a bit cleaner and it more explicitly separates the logging of the function from the actual function (this would allow you to just return the results of the run function call in the operation callback).

src/server/pfs/server/driver_new_storage.go
```diff
@@ -339,23 +339,28 @@ func deserializeShard(shardAny *types.Any) (*pfs.Shard, error) {
...
-// (bryce) it might potentially make sense to exit if an error occurs in this function
-// because each pachd instance that errors here will lose its compaction worker without an obvious
-// notification for the user (outside of the log message).
-// (bryce) ^ maybe just a retry would be good enough.
 func (d *driver) compactionWorker() {
 	w := work.NewWorker(d.etcdClient, d.prefix, storageTaskNamespace)
-	if err := w.Run(context.Background(), func(ctx context.Context, subtask *work.Task) error {
-		shard, err := deserializeShard(subtask.Data)
+	backoffStrat := backoff.NewExponentialBackOff()
+	backoffStrat.MaxElapsedTime = 0 // don't give up, compactionWorker should not exit
```
brycemcanally, line 345: Just a stylistic note, I tend to prefer not adding spaces between lines of code and instead use comments to separate logical groups (or to just add notes). Also, I don't really comment on the same line as code. In the interest of consistency, we should probably try and follow the same style.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -343,15 +343,14 @@ func deserializeShard(shardAny *types.Any) (*pfs.Shard, error) {
 func (d *driver) compactionWorker() {
 	ctx := context.Background()
 	w := work.NewWorker(d.etcdClient, d.prefix, storageTaskNamespace)
-
+	// Configure backoff so we retry indefinitely
 	backoffStrat := backoff.NewExponentialBackOff()
-	backoffStrat.MaxElapsedTime = 0 // don't give up, compactionWorker should not exit
-
+	backoffStrat.MaxElapsedTime = 0
 	notifier := func(err error, t time.Duration) error {
```
brycemcanally, line 349: We tend to define this inline with the `RetryNotify` call (similar to how the operation is defined). It is not going to get used anywhere else.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -366,5 +365,6 @@ func (d *driver) compactionWorker() {
 		})
 	}, backoffStrat, notifier)
```
brycemcanally, line 367: Space.

PR #4870, opened 2020-05-08
Change PutTar to operate on a stream of tagged tar streams.

brycemcanally:
This PR implements functionality for sending a batch of tagged tar streams to pfs through a single grpc stream. This should be useful for pps because it would enable efficient upload of a datum set. The put tar client would be setup before processing a datum set, then each datum's output would be sent as a tagged tar stream where the tag would be the datum identifier, then the put tar client would be closed after the datum set is completed.


src/client/pfs_new_storage.go
```diff
@@ -10,25 +10,60 @@ import (
...
+		return nil, err
+	}
+	return &PutTarClient{client: client}, nil
+}
+
+// PutTar puts a tar stream into PFS.
+// The files are not persisted until the PutTarClient is closed.
+func (ptc *PutTarClient) PutTar(r io.Reader, tag ...string) error {
+	if len(tag) > 0 {
+		if err := ptc.client.Send(&pfs.PutTarRequest{Tag: tag[0]}); err != nil {
```
Tryneus, line 52: This discards all tags after the first - maybe this should error if someone provides multiple tags?

src/client/pfs/pfs.proto
```diff
@@ -498,9 +498,22 @@ message FileInfoNewStorage {
...
+//       Client sends a request with the bytes stored in the data field.
+//     Client sends a request with the eof field set to true.
+//   Client connection is closed.
+
 message PutTarRequest {
   Commit commit = 1;
-  bytes data = 2;
+  string tag = 2;
+  bytes data = 3;
+  bool eof = 4;
```
Tryneus, line 516: should this be `EOF` in the go code for style consistency?

src/server/pfs/server/api_server_new_storage.go
```diff
@@ -10,35 +10,57 @@ import (
 	"github.com/pachyderm/pachyderm/src/client/pfs"
 	"github.com/pachyderm/pachyderm/src/client/pkg/errors"
 	"github.com/pachyderm/pachyderm/src/client/pkg/grpcutil"
+	"github.com/pachyderm/pachyderm/src/server/pkg/storage/fileset"
 	"github.com/pachyderm/pachyderm/src/server/pkg/storage/metrics"
 )
 
 func (a *apiServer) PutTar(server pfs.API_PutTarServer) (retErr error) {
```
Tryneus, line 17: It would be nice if this function enforced the protocol a bit more strongly (to help catch misuse earlier).  E.g. error if subsequent requests set the `Commit` field, error if a request has both `EOF` and `Data` set.

src/server/pfs/server/api_server_new_storage.go
```diff
@@ -10,35 +10,57 @@ import (
...
+					r:      bytes.NewReader(req.Data),
+				}
+				defer func() {
+					bytesRead += ptr.bytesRead
+				}()
+				if req.Tag != "" {
+					if err := fs.Put(ptr, req.Tag); err != nil {
+						return err
+					}
+					continue
```
Tryneus, line 51: maybe drop the `continue` and make the below `if` an `else if`?

src/server/pkg/storage/fileset/fileset.go
```diff
@@ -52,7 +53,12 @@ func newFileSet(ctx context.Context, storage *Storage, name string, memThreshold
 }
 
 // Put reads files from a tar stream and adds them to the fileset.
-func (f *FileSet) Put(r io.Reader) error {
+// (bryce) probably should prevent / clean files that end with "/", since that will indicate a directory.
+func (f *FileSet) Put(r io.Reader, customTag ...string) error {
+	tag := f.defaultTag
+	if len(customTag) > 0 {
+		tag = customTag[0]
+	}
```
Tryneus, line 61: similarly here, consider erroring if more than one `customTag` is passed in

src/server/pfs/server/api_server_new_storage.go
```diff
@@ -10,35 +10,57 @@ import (
...
+					}
+					return err
+				}
+				ptr := &putTarReader{
+					server: server,
+					r:      bytes.NewReader(req.Data),
+				}
+				defer func() {
+					bytesRead += ptr.bytesRead
+				}()
```
Tryneus, line 46: There might be ~thousands of these defers - does that have any implications on golang?

src/client/pfs_new_storage.go
brycemcanally: I guess I could add that.

src/client/pfs/pfs.proto
brycemcanally: Sure.

src/server/pfs/server/api_server_new_storage.go
brycemcanally: Sure, but that is something that is not too high priority now, particularly because I am still trying to figure out what will be needed from the API. We are not even in alpha yet lol.

src/server/pfs/server/api_server_new_storage.go
brycemcanally: Yeah, I started with that but I don't like else statements. I think I might tweak this code a bit to just avoid having to check for the empty string.

src/server/pkg/storage/fileset/fileset.go
brycemcanally: Yeah, none of the optional parameter code I have seen checks if it is too long, and none of mine does. I'm not too sure if it is really worth the extra code.

src/server/pfs/server/api_server_new_storage.go
brycemcanally: We should not be computing the bytes read like this. I for the most part pulled the existing put files code over. I'm pretty sure I would have noticed this when reviewing / testing this. My main goal with the current state of this PR was to give you a chance to take a look at the API and see if it is going to work for you.

PR #4875, opened 2020-05-12
Adds ListFileV2 rpc call

### LINKED ISSUE #4856: Implement list file with the new storage layer
We need to implement list file with the new storage layer. This should be pretty straightforward since most of the heavy lifting will be done by get files conditional (https://github.com/pachyderm/pachyderm/blob/03f22b5ebec7f56813fced4f7215de46c50ccf20/src/server/pfs/server/driver_new_storage.go#L145). The way this should be implemented is by adding a new function to https://github.com/pachyderm/pachyderm/blob/master/src/server/pfs/server/driver_new_storage.go for listing files. This function should be implemented by calling get files conditional with a callback that just uses the file info (these should get returned through the grpc stream). A new grpc endpoint will need to be setup for this, and the corresponding code that handles the client side will need to be implemented as well. A good example to follow would be the end-to-end of get tar (the general code path from client to server is: https://github.com/pachyderm/pachyderm/blob/master/src/client/pfs_new_storage.go, https://github.com/pachyderm/pachyderm/blob/master/src/server/pfs/server/api_server_new_storage.go, https://github.com/pachyderm/pachyderm/blob/master/src/server/pfs/server/driver_new_storage.go). The pfs proto file is https://github.com/pachyderm/pachyderm/blob/master/src/client/pfs/pfs.proto and it is compiled by `make proto`. We will probably need to put a little bit of thought into the naming scheme for this because the new storage layer endpoints are not namespaced with respect to the current system, so list file would not be the name of the grpc endpoint.

brendoncarroll:
ListFileNS implements ListFile using the new storage layer.
- Changes the pfs.proto interface and regenerates stubs.
- Adds methods on the storage driver.

Closes #4856


src/client/pfs/pfs.proto
```diff
@@ -613,6 +613,7 @@ service API {
   rpc GetTar(GetTarRequest) returns (stream google.protobuf.BytesValue) {}
   // Refer to the GetTarConditionalRequest / GetTarConditionalResponse message definitions for the protocol.
   rpc GetTarConditional(stream GetTarConditionalRequest) returns (stream GetTarConditionalResponse) {}
+  rpc ListFileNS(ListFileRequest) returns (stream FileInfoNewStorage) {}
```
brycemcanally, line 616: I think it would be better to name this `ListFileNewStorage` because I think it will be unclear to others what `NS` means.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -132,6 +132,28 @@ func (d *driver) getFilesNewStorageLayer(ctx context.Context, repo, commit, glob
 	return tar.NewWriter(w).Close()
 }
 
+func (d *driver) listFileNS(pachClient *client.APIClient, file *pfs.File, full bool, history int64, f func(*pfs.FileInfoNewStorage) error) (retErr error) {
+	ctx := pachClient.Ctx()
+	// Validate arguments
+	if file == nil {
```
brycemcanally, line 138: This validation should probably be in a helper function.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -132,6 +132,28 @@ func (d *driver) getFilesNewStorageLayer(ctx context.Context, repo, commit, glob
...
+		return errors.New("file commit cannot be nil")
+	}
+	if file.Commit.Repo == nil {
+		return errors.New("file commit repo cannot be nil")
+	}
+	// auth check
+	if err := d.checkIsAuthorized(pachClient, file.Commit.Repo, auth.Scope_READER); err != nil {
+		return err
+	}
+	err := d.getFilesConditional(ctx, file.Commit.Repo.Name, file.Commit.ID, file.Path, func(fr *FileReader) error {
```
brycemcanally, line 151: You can just return the results of the function call, rather than storing the error and returning it.

src/server/pfs/server/api_server_new_storage.go
```diff
@@ -152,3 +152,14 @@ func (w *getTarConditionalWriter) Write(data []byte) (int, error) {
 	w.bytesWritten += int64(len(data))
 	return len(data), nil
 }
+
+func (a *apiServer) ListFileNS(req *pfs.ListFileRequest, server pfs.API_ListFileNSServer) error {
+	pachClient := a.env.GetPachClient(server.Context())
+	err := a.driver.listFileNS(pachClient, req.File, req.Full, req.History, func(finfo *pfs.FileInfoNewStorage) error {
```
brycemcanally, line 158: Similar to below, you can just return the result of the function call.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -132,6 +132,28 @@ func (d *driver) getFilesNewStorageLayer(ctx context.Context, repo, commit, glob
...
+	if file == nil {
+		return errors.New("file cannot be nil")
+	}
+	if file.Commit == nil {
+		return errors.New("file commit cannot be nil")
+	}
+	if file.Commit.Repo == nil {
+		return errors.New("file commit repo cannot be nil")
+	}
+	// auth check
```
brycemcanally, line 147: This comment seems kind of unnecessary considering how descriptive the function name.

src/client/pfs_new_storage.go
```diff
@@ -89,7 +90,33 @@ func (c APIClient) GetTarConditional(repoName string, commitID string, path stri
 		}
 	}
 	return nil
+}
 
+// ListFileNS returns info about all files in a Commit under path.
+func (c APIClient) ListFileNS(repoName string, commitID string, path string) (finfos []*pfs.FileInfoNewStorage, retErr error) {
```
brycemcanally, line 96: This should be callback based similar to `GetTarConditional`. This will use too much memory for a large number of files (plus you have to wait for the full set of file infos before returning any). The callback pattern allows us to take advantage of the fact that list file rpc streams the file infos.

src/client/pfs/pfs.proto
brendoncarroll: What do you think about standardizing around the suffix "2" or something less verbose? If this is expected by default in 2.0, then that seems like a reasonable naming convention.

brycemcanally: Yeah, it is a bit verbose. I think moving over to a suffix such as `V2` would work. I had some concerns about the fact that we are probably going to have three parallel configurations (1.x, 1.x with new storage, and 2.x), but the new endpoint interfaces should be the same in 2.x. Two feature flags should be enough to navigate the 1.x with new storage and 2.x branches.

src/client/pfs_new_storage.go
```diff
@@ -128,7 +129,35 @@ func (c APIClient) GetTarConditional(repoName string, commitID string, path stri
 		}
 	}
 	return nil
+}
 
+// ListFileV2 streams FileInfoNewStorage similar to ListFileF
```
brycemcanally, line 134: We probably shouldn't reference functions that are going to be removed. `ListFileV2` will subsume the `ListFileF` functionality.

src/client/pfs_new_storage.go
```diff
@@ -128,7 +129,35 @@ func (c APIClient) GetTarConditional(repoName string, commitID string, path stri
...
+		File: NewFile(repoName, commitID, path),
+		Full: true,
+	}
+	client, err := c.PfsAPIClient.ListFileV2(ctx, req)
+	if err != nil {
+		return err
+	}
+	for {
+		finfo, err := client.Recv()
+		if err == io.EOF {
```
brycemcanally, line 151: The way I tend to structure things like this is:
```
if err != nil {
  if err == io.EOF {
    break
  }
  return err
}
```
I think it is a little bit nicer from a readability standpoint because it groups the non-nil error path.

src/client/pfs_new_storage.go
```diff
@@ -128,7 +129,35 @@ func (c APIClient) GetTarConditional(repoName string, commitID string, path stri
...
 	}
 	return nil
+}
 
+// ListFileV2 streams FileInfoNewStorage similar to ListFileF
+func (c APIClient) ListFileV2(repoName string, commitID string, path string, f func(fileInfo *pfs.FileInfoNewStorage) error) (retErr error) {
+	defer func() {
+		retErr = grpcutil.ScrubGRPC(retErr)
+	}()
+	ctx, cf := context.WithCancel(c.Ctx())
```
brycemcanally, line 139: This would probably be a good thing to add in the other client side implementations that have code other than the grpc stuff that can error (`GetTarConditional` seems like the obvious one). We don't need it in this PR, I just thought it was noteworthy.

brycemcanally: Also, just to nitpick while I am here lol, can we name the cancel function `cancel` because that is the name most widely used in the codebase.

src/server/pfs/server/api_server_new_storage.go
```diff
@@ -173,3 +173,10 @@ func (w *getTarConditionalWriter) Write(data []byte) (int, error) {
 	w.bytesWritten += int64(len(data))
 	return len(data), nil
 }
+
+func (a *apiServer) ListFileV2(req *pfs.ListFileRequest, server pfs.API_ListFileV2Server) error {
+	pachClient := a.env.GetPachClient(server.Context())
+	return a.driver.listFileNS(pachClient, req.File, req.Full, req.History, func(finfo *pfs.FileInfoNewStorage) error {
```
brycemcanally, line 179: `NS` -> `V2`

src/server/pfs/server/driver.go
```diff
@@ -3960,6 +3945,19 @@ func (d *driver) listFile(pachClient *client.APIClient, file *pfs.File, full boo
 	})
 }
 
+func validateListFileArgs(file *pfs.File) error {
```
brycemcanally, line 3948: This should probably be `validateFile`, especially since you added it to inspect file.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -131,6 +131,19 @@ func (d *driver) getFilesNewStorageLayer(ctx context.Context, repo, commit, glob
 	return tar.NewWriter(w).Close()
 }
 
+func (d *driver) listFileNS(pachClient *client.APIClient, file *pfs.File, full bool, history int64, f func(*pfs.FileInfoNewStorage) error) (retErr error) {
```
brycemcanally, line 134: `NS` -> `V2`

src/server/pkg/testpachd/mock_pachd.go
```diff
@@ -399,6 +399,7 @@ type fsckFunc func(*pfs.FsckRequest, pfs.API_FsckServer) error
 type putTarFunc func(pfs.API_PutTarServer) error
 type getTarFunc func(*pfs.GetTarRequest, pfs.API_GetTarServer) error
 type getTarConditionalFunc func(pfs.API_GetTarConditionalServer) error
+type listFileNSFunc func(*pfs.ListFileRequest, pfs.API_ListFileV2Server) error
```
brycemcanally, line 402: `NS` -> `V2`

PR #4876, opened 2020-05-13
Use host native DNS resolver for hostnames in grpc connections

### LINKED ISSUE #4807: pachctl does not respect OSX dns over vpn
**What happened?**:
Because pachctl is not linked against CGO (CGO_ENABLED=0) on osx, it does not honor the DNS resolver setting son OSX when using a vpn.

DNS resolution should work with pachctl commands.

Easiest way to test:
Edit your /etc/resolv.conf on osx and set the nameserver to something nonsensical like 127.0.0.1.
Observe that your osx applications, and ping on the command line, all still have dns resolution.
Observe that pachctl does not. 

Similar kubectl issue:
https://github.com/kubernetes/release/issues/469

**Environment?**:
Any osx client using pachctl of any version.

I could have sworn I opened this issue before but I couldn't find it on the issue tracker, so here it is again.<br><br><i>(related to <a href='https://pachyderm.zendesk.com/agent/tickets/42'>Zendesk ticket #42</a>)<br>gz#42</i><br><br><i>(related to <a href='https://pachyderm.zendesk.com/agent/tickets/44'>Zendesk ticket #44</a>)<br>gz#44</i>

nitinjainsj:
Fixes #4807 

Testing...

- Created a hub cluster so that access is over the network.
- Set the pachctl context to hub cluster.
- Run pachctl version -- it should give you the client and server version
- Edit the /etc/resolve.conf and change the nameserver to 127.0.0.1
- pachctl version hangs to get the server version
- made changes to add the resolver to grpc
- mypachctl version completes with expected info


ysimonson:
Have you verified that this fixes the issue? I'm not sure from the docs:

* The docs for [`grpc.dial`](https://godoc.org/google.golang.org/grpc#Dial) say "The target name syntax is defined in https://github.com/grpc/grpc/blob/master/doc/naming.md. e.g. to use dns resolver, a "dns:///" prefix should be applied to the target."
* However, the [naming docs](https://github.com/grpc/grpc/blob/master/doc/naming.md) imply that it's defaulting to the DNS URI anyways: "The URI scheme indicates what resolver plugin to use. If no scheme prefix is specified or the scheme is unknown, the dns scheme is used by default."

nitinjainsj:
> Have you verified that this fixes the issue? I'm not sure from the docs:
The user ticket had repo steps. It's easy to test on Mac. I am planning to boot my VM and test it on linux.

Testing...
* I created a hub cluster and set the pachctl context to hub.
* Run pachctl version -- it should give you the client and server version
* Edit the /etc/resolve.conf and change the nameserver to 127.0.0.1
* pachctl version hangs to get the server version
* make change to add the resolver
* mypachctl version completes with expected info

> 
> * The docs for [`grpc.dial`](https://godoc.org/google.golang.org/grpc#Dial) say "The target name syntax is defined in https://github.com/grpc/grpc/blob/master/doc/naming.md. e.g. to use dns resolver, a "dns:///" prefix should be applied to the target."
> * However, the [naming docs](https://github.com/grpc/grpc/blob/master/doc/naming.md) imply that it's defaulting to the DNS URI anyways: "The URI scheme indicates what resolver plugin to use. If no scheme prefix is specified or the scheme is unknown, the dns scheme is used by default."

Yeah. That's the odd part. The doc says dns resolver is default, but clearly it did not work until I explicitly added to the name. Maybe there is dialer option somewhere that is the right way to solve this.

I'll add the dns:/// to grpc.dial as well -- though some of the code is probably only communicating within the cluster. So it might not even need it.

src/client/pkg/grpcutil/dialer.go
```diff
@@ -35,13 +35,17 @@ func newDialer(opts ...grpc.DialOption) *dialer {
...
 	defer d.lock.Unlock()
+
 	if conn, ok := d.connMap[addr]; ok {
 		return conn, nil
 	}
 	opts := append(d.opts,
 		grpc.WithUnaryInterceptor(tracing.UnaryClientInterceptor()),
 		grpc.WithStreamInterceptor(tracing.StreamClientInterceptor()),
 	)
+	if (strings.Index(addr, "dns:///") == -1) {
```
ysimonson, line 47: It'd be slightly faster to [just check for the prefix](https://golang.org/pkg/strings/#HasPrefix). If `dns:///` is in the string and it's not the prefix, it's going to be broke anyways.

src/server/pkg/pool/pool.go
```diff
@@ -109,7 +109,7 @@ func (p *Pool) Do(ctx context.Context, f func(cc *grpc.ClientConn) error) error
 		for {
 			for addr, mapConn := range p.conns {
 				if mapConn.cc == nil {
-					cc, err := grpc.DialContext(ctx, addr, p.opts...)
+					cc, err := grpc.DialContext(ctx, "dns:///"+addr, p.opts...)
```
ysimonson, line 112: I don't think we actually want it here, since the new DNS resolution behavior should be client-side only

src/server/worker/client.go
```diff
@@ -126,7 +126,7 @@ func NewClient(address string) (Client, error) {
 	if err != nil {
 		return Client{}, err
 	}
-	conn, err := grpc.Dial(fmt.Sprintf("%s:%d", address, port),
+	conn, err := grpc.Dial(fmt.Sprintf("dns:///%s:%d", address, port),
```
ysimonson, line 129: Likewise, i don't think we want this here

src/client/pkg/grpcutil/dialer.go
nitinjainsj: Yeah. The updated code has changed it to HasPrefix

src/server/pkg/pool/pool.go
nitinjainsj: I don't have a strong opinion on this. But the reason to go change all the grpc.Dial call is so that we have the same grpc options used everywhere. I don't think we have any specific DNS configuration dependency.

If we think this is can have unintended side effects, I'll remove it.

src/server/worker/client.go
nitinjainsj: Agree. The same argument above. Use the underlying system DNS by default.

src/server/pkg/pool/pool.go
ysimonson: It's not clear to me yet how `dns:///` changes behavior. gRPC apparently [has a system for plugging in resolvers](https://godoc.org/google.golang.org/grpc#WithResolvers), so my off-hand guess (without having dug too much into the code) is that using the dns uri causes gRPC to use one of its own resolvers, which follows proper system behavior, rather than go stdlib's, which [hand-rolled its own janky solution when cgo is disabled.](https://golang.org/pkg/net/#hdr-Name_Resolution)

My concern with bringing this change to the server-side is that we don't really know what happens when you use the `dns` uri though. What we do know is that:

1) DNS resolution is broken for us client-side on macOS.
2) We haven't had any reports of issues server-side. 
3) Even with these changes, server-side DNS is going to be radically different because it's running on linux, and because it's on kubernetes (which messes around with DNS.)

By restricting these changes to the client-side, we run the risk of "just" breaking `pachctl`.

nitinjainsj: @ysimonson I made the change to be client-side only. It's going through CI. Please take a look.

PR #4884, opened 2020-05-13
s3g examples

ysimonson:


ysimonson:
Instead of going this route, I'll wire in a more fully-fledged integration test suite from s2: https://github.com/pachyderm/s2/pull/25

PR #4901, opened 2020-05-19
Limit fan-In during compaction.

### LINKED ISSUE #4857: Make the compaction process with the new storage layer hierarchical
We need to make the compaction process hierarchical to prevent excessive object storage request fan-out and memory usage. The current compaction process is susceptible to object storage throttling and/or getting OOM killed when a large number of filesets are getting compacted (this can occur because we will buffer a chunk from each fileset being compacted). To prevent this, we need some configurable bound on the fan-out during the compaction process that will be enforced by a hierarchical compaction scheme. The compaction code starts here https://github.com/pachyderm/pachyderm/blob/03f22b5ebec7f56813fced4f7215de46c50ccf20/src/server/pfs/server/driver_new_storage.go#L285. This should be pretty straightforward because we just need to perform the work in batches as we work up the hierarchy until we are left with the final compacted fileset. The fan-out should be exposed as a storage configuration here https://github.com/pachyderm/pachyderm/blob/03f22b5ebec7f56813fced4f7215de46c50ccf20/src/server/pkg/serviceenv/config.go#L69. With this change, we should also think a bit more about how we should be setting up the compaction shards, and what configuration should be available for this. Figuring out a good default for these things will be important for good performance out of the box.

brendoncarroll:
Limits the number of objects that are open at a time during compact.  Adds a configurable `StorageCompactFanIn` parameter.

The workers will never have to open more than the fan-in value.  The master may need to compact more than fan-in filesets at once, but they will be well clustered because of a previous sharded compaction, and they can be concatenated while maintaining a correct ordering of keys in the final output. During concatenation only one fileset needs to be read from at a time.

Closes #4857

### DIFF
--- a/src/server/pfs/server/driver.go
+++ b/src/server/pfs/server/driver.go
@@ -156,6 +156,7 @@ func newDriver(
 		memoryLimiter:    semaphore.NewWeighted(memoryRequest / 3),
 		putObjectLimiter: limit.New(env.StorageUploadConcurrencyLimit),
 		putFileLimiter:   limit.New(env.StoragePutFileConcurrencyLimit),
+		// TODO: set maxFanIn based on downward API.
 	}
 
 	// Create spec repo (default repo)
--- a/src/server/pfs/server/driver_v2.go
+++ b/src/server/pfs/server/driver_v2.go
@@ -297,29 +297,102 @@ func (fr *FileReader) drain() error {
 	return nil
 }
 
-func (d *driver) compact(master *work.Master, outputPath string, inputPrefixes []string) (retErr error) {
+func (d *driver) compact(master *work.Master, outputPath string, inputPrefixes []string) error {
+	ctx := master.Ctx()
+	// resolve prefixes into paths
+	inputPaths := []string{}
+	for _, inputPrefix := range inputPrefixes {
+		if err := d.storage.WalkFileSet(ctx, inputPrefix, func(inputPath string) error {
+			inputPaths = append(inputPaths, inputPath)
+			return nil
+		}); err != nil {
+			return err
+		}
+	}
+	return d.compactIter(ctx, compactSpec{
+		master:     master,
+		inputPaths: inputPaths,
+		outputPath: outputPath,
+		maxFanIn:   d.env.StorageCompactionMaxFanIn,
+	})
+}
+
+type compactSpec struct {
+	master     *work.Master
+	outputPath string
+	inputPaths []string
+	maxFanIn   int
+}
+
+// compactIter is one level of compaction.  It will only perform compaction
+// if len(inputPaths) <= params.maxFanIn otherwise it will split inputPaths recursively.
+func (d *driver) compactIter(ctx context.Context, params compactSpec) (retErr error) {
+	if len(params.inputPaths) <= params.maxFanIn {
+		return d.shardedCompact(ctx, params.master, params.outputPath, params.inputPaths)
+	}
 	scratch := path.Join(tmpPrefix, uuid.NewWithoutDashes())
 	defer func() {
-		if err := d.storage.Delete(context.Background(), scratch); retErr == nil {
+		if err := d.storage.Delete(ctx, scratch); retErr == nil {
 			retErr = err
 		}
 	}()
-	compaction := &pfs.Compaction{InputPrefixes: inputPrefixes}
+	childOutputPaths := []string{}
+	childSize := len(params.inputPaths) / params.maxFanIn
+	if len(params.inputPaths)%params.maxFanIn != 0 {
+		childSize++
+	}
+	// TODO: use an errgroup to make the recursion concurrecnt.
+	// this requires changing the master to allow multiple calls to RunSubtasks
+	// don't forget to pass the errgroups childCtx to compactIter instead of ctx.
+	for i := 0; i < params.maxFanIn; i++ {
+		start := i * childSize
+		end := (i + 1) * childSize
+		if end > len(params.inputPaths) {
+			end = len(params.inputPaths)
+		}
+		childOutputPath := path.Join(scratch, strconv.Itoa(i))
+		childOutputPaths = append(childOutputPaths, childOutputPath)
+		if err := d.compactIter(ctx, compactSpec{
+			master:     params.master,
+			inputPaths: params.inputPaths[start:end],
+			outputPath: childOutputPath,
+			maxFanIn:   params.maxFanIn,
+		}); err != nil {
+			return err
+		}
+	}
+	return d.shardedCompact(ctx, params.master, params.outputPath, childOutputPaths)
+}
+
+// shardedCompact generates shards for the fileset(s) in inputPaths,
+// gives those shards to workers, and waits for them to complete.
+// Fan in is bound by len(inputPaths), concatenating shards have
+// fan in of one because they are concatenated sequentially.
+func (d *driver) shardedCompact(ctx context.Context, master *work.Master, outputPath string, inputPaths []string) (retErr error) {
+	scratch := path.Join(tmpPrefix, uuid.NewWithoutDashes())
+	defer func() {
+		if err := d.storage.Delete(ctx, scratch); retErr == nil {
+			retErr = err
+		}
+	}()
+	compaction := &pfs.Compaction{InputPrefixes: inputPaths}
 	var subtasks []*work.Task
-	if err := d.storage.Shard(master.Ctx(), inputPrefixes, func(pathRange *index.PathRange) error {
-		outputPath := path.Join(scratch, strconv.Itoa(len(subtasks)))
+	var shardOutputs []string
+	if err := d.storage.Shard(ctx, inputPaths, func(pathRange *index.PathRange) error {
+		shardOutputPath := path.Join(scratch, strconv.Itoa(len(subtasks)))
 		shard, err := serializeShard(&pfs.Shard{
 			Compaction: compaction,
 			Range: &pfs.PathRange{
 				Lower: pathRange.Lower,
 				Upper: pathRange.Upper,
 			},
-			OutputPath: outputPath,
+			OutputPath: shardOutputPath,
 		})
 		if err != nil {
 			return err
 		}
 		subtasks = append(subtasks, &work.Task{Data: shard})
+		shardOutputs = append(shardOutputs, shardOutputPath)
 		return nil
 	}); err != nil {
 		return err
@@ -332,7 +405,22 @@ func (d *driver) compact(master *work.Master, outputPath string, inputPrefixes [
 	}); err != nil {
 		return err
 	}
-	return d.storage.Compact(master.Ctx(), outputPath, []string{scratch})
+	return d.concatFileSets(ctx, outputPath, shardOutputs)
+}
+
+// concatFileSets concatenates the filesets in inputPaths and writes the result to outputPath
+// TODO: move this to the fileset package, and error if the entries are not sorted.
+func (d *driver) concatFileSets(ctx context.Context, outputPath string, inputPaths []string) error {
+	fsw := d.storage.NewWriter(ctx, outputPath)
+	for _, inputPath := range inputPaths {
+		fsr := d.storage.NewReader(ctx, inputPath)
+		if err := fsr.Iterate(func(fr *fileset.FileReader) error {
+			return fsw.CopyFile(fr)
+		}); err != nil {
+			return err
+		}
+	}
+	return fsw.Close()
 }
 
 func serializeShard(shard *pfs.Shard) (*types.Any, error) {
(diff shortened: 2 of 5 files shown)


src/server/pfs/server/driver_new_storage.go
```diff
@@ -298,24 +299,105 @@ func (fr *FileReader) drain() error {
 	return nil
 }
 
-func (d *driver) compact(master *work.Master, outputPath string, inputPrefixes []string) (retErr error) {
+// compact is the entrypoint to compaction.
+func (d *driver) compact(ctx context.Context, master *work.Master, outputPath string, inputPrefixes []string) error {
+	// resolve prefixes into paths
+	inputPaths := []string{}
+	for _, inputPrefix := range inputPrefixes {
+		err := d.storage.WalkFileSets(ctx, inputPrefix, func(inputPath string) error {
```
brycemcanally, line 307: This should be pulling from the master context (`m.Ctx()`), not the background. The `compact` function should not take a separate context as a parameter.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -298,24 +299,105 @@ func (fr *FileReader) drain() error {
 	return nil
 }
 
-func (d *driver) compact(master *work.Master, outputPath string, inputPrefixes []string) (retErr error) {
+// compact is the entrypoint to compaction.
```
brycemcanally, line 302: Kind of unnecessary comment.

src/server/pfs/server/driver_new_storage.go
brycemcanally: Also, `WalkFileSets` -> `WalkFileSet`

brycemcanally: Also, the if statement can just operate on the results of the function call, you don't need it to be below.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -298,24 +299,105 @@ func (fr *FileReader) drain() error {
...
+	for _, inputPrefix := range inputPrefixes {
+		err := d.storage.WalkFileSets(ctx, inputPrefix, func(inputPath string) error {
+			inputPaths = append(inputPaths, inputPath)
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+	}
+	if d.env.StorageCompactionMaxFanIn < 2 {
```
brycemcanally, line 315: We probably should not get to this point if this is the case, we probably want this grouped with other validation (I think we can punt on this tbh).

src/server/pfs/server/driver_new_storage.go
```diff
@@ -298,24 +299,105 @@ func (fr *FileReader) drain() error {
...
+	}
+	return d.compactIter(ctx, compactParams{
+		master:     master,
+		inputPaths: inputPaths,
+		outputPath: outputPath,
+		maxFanIn:   d.env.StorageCompactionMaxFanIn,
+	})
+}
+
+type compactParams struct {
```
brycemcanally, line 326: I think `compactSpec` might be a better name.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -298,24 +299,105 @@ func (fr *FileReader) drain() error {
...
+// compactIter is one level of compaction.  It will only perform compaction
+// if len(inputPaths) < params.maxFanIn otherwise it will split inputPaths recursively.
+func (d *driver) compactIter(ctx context.Context, params compactParams) (retErr error) {
+	scratch := path.Join(tmpPrefix, uuid.NewWithoutDashes())
+	defer func() {
+		if err := d.storage.Delete(ctx, scratch); retErr == nil {
+			retErr = err
+		}
+	}()
+	// exceeds maxFanIn need to branch out
```
brycemcanally, line 342: Might want to reword a bit.

src/server/pfs/server/driver_new_storage.go
brendoncarroll: Isn't there a method `CompactSpec` on the storage type that would get confused with this?

src/server/pfs/server/driver_new_storage.go
```diff
@@ -298,24 +299,105 @@ func (fr *FileReader) drain() error {
...
+	inputPaths := params.inputPaths
+	if len(inputPaths) > params.maxFanIn {
+		childOutputPaths := []string{}
+		childSize := len(params.inputPaths) / params.maxFanIn
+		if len(params.inputPaths)%params.maxFanIn != 0 {
+			childSize++
+		}
+		group, ctx := errgroup.WithContext(ctx)
```
brycemcanally, line 350: Just for consistency, `group` -> `eg`.

src/server/pfs/server/driver_new_storage.go
brendoncarroll: I always find it hard to read the compound if statements when they are broken up by a function literal.  What do you think?

src/server/pfs/server/driver_new_storage.go
brendoncarroll: I guess I can remove it. It will probably OOM pretty quickly anways.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -298,24 +299,105 @@ func (fr *FileReader) drain() error {
...
+func (d *driver) compactIter(ctx context.Context, params compactParams) (retErr error) {
+	scratch := path.Join(tmpPrefix, uuid.NewWithoutDashes())
+	defer func() {
+		if err := d.storage.Delete(ctx, scratch); retErr == nil {
+			retErr = err
+		}
+	}()
+	// exceeds maxFanIn need to branch out
+	inputPaths := params.inputPaths
+	if len(inputPaths) > params.maxFanIn {
```
brycemcanally, line 344: Something that I think could be beneficial from a readability standpoint would be to structure this how most recursive functions are where the base case is checked first, then the recursive step is handled after the check. So, you would basically invert what you have now, so the check would be less than or equal to fan in, then run sharded compact. If not, then kick off the recursion and run sharded compact on the results. Just an idea, what are your thoughts?

src/server/pfs/server/driver_new_storage.go
```diff
@@ -298,24 +299,105 @@ func (fr *FileReader) drain() error {
...
+		}
+		inputPaths = childOutputPaths
+	}
+	// within maxFanIn, generate compaction tasks for workers
+	return d.shardedCompact(ctx, params.master, params.outputPath, inputPaths)
+}
+
+// shardedCompact generates shards for the fileset(s) in inputPaths,
+// gives those shards to workers, and waits for them to complete
+// fan-in: spatially, and temporally to each shard will be len(inputPaths)
```
brycemcanally, line 379: Not sure I understand what these two lines are saying? Maybe reword? Fan in is bound by len(inputPaths), concatenating shards have fan in of one because they are concatenated sequentially.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -333,7 +415,22 @@ func (d *driver) compact(master *work.Master, outputPath string, inputPrefixes [
 	}); err != nil {
 		return err
 	}
-	return d.storage.Compact(master.Ctx(), outputPath, []string{scratch})
+	return d.concatFileSets(ctx, outputPath, shardOutputs)
+}
+
+// concatFileSets concatenates the filesets in inputPaths and writes the result to outputPath
+// TODO: move this to the fileset package, and error if the entries are not sorted.
```
brycemcanally, line 422: :+1:

src/server/pfs/server/driver_new_storage.go
brycemcanally: Yeah, I think it is worth making a todo (and you could even have the todo somewhere around here), but I think this type of validation should be handled at setup, not while we are in the middle of a compaction.

src/server/pfs/server/driver_new_storage.go
brycemcanally: Yeah, that is actually why I prefer it. There is a notion of a fileset compact spec and a pfs compact spec. The fileset compact spec shards across a set of filesets, and the pfs compact spec manages fan out.

src/server/pfs/server/driver_new_storage.go
brendoncarroll: Yeah that could work.  I think it prevents the reassign to inputPaths, which is cleaner.  It also lets us skip the scratch space for the base case.  I'll give it a try.

src/server/pfs/server/driver_new_storage.go
brycemcanally: I kind of prefer this pattern for the simple case where I am just returning the error if it is not nil, but for more complex logic it could make sense to split it up. The general pattern in the codebase for the simple case is to just have the function call in the if statement.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -298,24 +299,99 @@ func (fr *FileReader) drain() error {
...
+	master     *work.Master
+	outputPath string
+	inputPaths []string
+	maxFanIn   int
+}
+
+// compactIter is one level of compaction.  It will only perform compaction
+// if len(inputPaths) < params.maxFanIn otherwise it will split inputPaths recursively.
+func (d *driver) compactIter(ctx context.Context, params compactSpec) (retErr error) {
+	// within maxFanIn
```
brycemcanally, line 333: Kind of unnecessary.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -298,24 +299,99 @@ func (fr *FileReader) drain() error {
...
+
+type compactSpec struct {
+	master     *work.Master
+	outputPath string
+	inputPaths []string
+	maxFanIn   int
+}
+
+// compactIter is one level of compaction.  It will only perform compaction
+// if len(inputPaths) < params.maxFanIn otherwise it will split inputPaths recursively.
```
brycemcanally, line 331: <=

src/server/pfs/server/testing/server_new_storage_test.go
```diff
@@ -35,6 +37,7 @@ type loadConfig struct {
 func newLoadConfig(opts ...loadConfigOption) *loadConfig {
 	config := &loadConfig{}
 	config.pachdConfig = newPachdConfig()
+	config.pachdConfig.StorageCompactionMaxFanIn = 2
```
brycemcanally, line 40: It seems like if this should be the default it would be in the `newPachdConfig` function, but this seems like a pretty low default.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -333,7 +409,30 @@ func (d *driver) compact(master *work.Master, outputPath string, inputPrefixes [
 	}); err != nil {
 		return err
 	}
-	return d.storage.Compact(master.Ctx(), outputPath, []string{scratch})
+
+	var shardOutputs []string
```
brycemcanally, line 413: I think we should follow the same pattern you used for recording the `childOutputPaths` here. This does not actually preserve the shard order because walk will return the paths lexicographically.

src/server/pfs/server/testing/server_new_storage_test.go
brendoncarroll: It should be 2 here because its the best number for the low memory testing, and least likely to explode CI.  The best default depends on the amount of memory we want to use. I think larger will almost always be better as far as time is concerned.

src/server/pfs/server/driver_new_storage.go
brendoncarroll: I think this can produce an error if no shard is written out.

src/server/pfs/server/testing/server_new_storage_test.go
brycemcanally: I think it makes sense for the default to be a bit higher, and for the low memory testing you should configure it as two with a functional option for that field. Also, the default should be set with the other defaults in `newPachdConfig`.

src/server/pfs/server/driver_new_storage.go
brycemcanally: If a shard is created, then a fileset, empty or not, should get written to the semantic path for a shard (if it is not, then there is a bug somewhere).

src/server/pfs/server/driver_new_storage.go
```diff
@@ -429,6 +418,7 @@ func (d *driver) concatFileSets(ctx context.Context, outputPath string, inputPat
 		if err := fsr.Iterate(func(fr *fileset.FileReader) error {
 			return fsw.CopyFile(fr)
 		}); err != nil {
+			panic(err)
```
brycemcanally, line 421: Remove panic debug.

PR #4902, opened 2020-05-19
Improved s3gateway compatibility

### LINKED ISSUE #4187: list_buckets() and list_objects() are returning empty lists from pachyderm using minio client
Dear all,

when using list_buckets() or list_objects() on an account/bucket in pachyderm, the result is an empty list, allthough there are buckets/objects in that account.
```
s3_client = Minio(endpoint=s3_endpoint,
                  access_key=s3_accesskey,
                  secret_key=s3_secretkey,
                  region=s3_region,
                  secure=s3_secure)
the_bucket = s3_bucket
list_buckets = s3_client.list_buckets()  #Returns empty list
is_bucket = s3_client.bucket_exists(the_bucket)  #Returns True
list_objects = s3_client.list_objects(the_bucket)  #Returns empty generator object
[print(obj.object_name) for obj in list_objects]  #Throws the error below
```

**Environment?**:
kubernetes version v12
pachctl version         
COMPONENT           VERSION             
pachctl             1.9.2               
pachd               1.9.4  

- Cloud provider (e.g. aws, azure, gke) or local deployment 
custom kubernetes on azure 
Mac OS Sierra

Error on trying to access the objects from existing bucket

Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/minio/parsers.py", line 104, in get_child_text
    return self.element.find('s3:{}'.format(name), _S3_NS).text
AttributeError: 'NoneType' object has no attribute 'text'

During handling of the above exception, another exception occurred:
```
Traceback (most recent call last):
  File "pachyderm_s3.py", line 75, in <module>
    main()
  File "pachyderm_s3.py", line 60, in main
    for obj in objects:
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/minio/api.py", line 906, in list_objects
    bucket_name=bucket_name)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/minio/parsers.py", line 238, in parse_list_objects
    is_truncated = root.get_child_text('IsTruncated').lower() == 'true'
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/minio/parsers.py", line 108, in get_child_text
    'Message: {}').format(self.root_name, name, error.message)
AttributeError: 'AttributeError' object has no attribute 'message'```

### PR DESCRIPTION
**Pull request by ysimonson**
### s3gateway

* Re-synced conformance test ignores
* Handle errors when attempting to put to a file path that includes an existing, non-directory parent path
* Fixed support for arbitrary credentials with AWS v4 auth. Now, when pach auth is disabled, you can still pass an arbitrary access key, but the secret key must match the access key. This is necessary for signature checking in AWS v4.
* Fixed incorrect error when completing a multipart with a valid upload ID, but an invalid part number
* Support for bucket names of just the form `repo` instead of `branch.repo`. This should maybe be discouraged, because it can lead to unexpected results; e.g. because the list buckets operation will always return buckets of the form `branch.repo`. But it made wiring in the new s2 integration tests easier.
* Removed the method for deleting a multipart chunk, as s2 no longer employs it.

### s2

See associated ticket: https://github.com/pachyderm/s2/pull/25

* Added integration tests, which checks that basic functionality works with minio's cli, aws' cli, the boto3 python library, minio's python library, and minio's go library. While there were extensive conformance tests already, they ran strictly against boto3. Because s3's spec is fairly loosely defined, disparate clis/libraries tend to implement non-overlapping subsets. With these integration tests, we'll better be able to catch and support these differences.
* Added support for chunked uploads (note these are not the same as multipart uploads.) minio's go library and `mc` needs it for non-TLS endpoints.
* Added xmlns to all returned XML documents, except for errors. minio-python requires them.
* Adjusted datetimes to be rounded to the nearest second, and use only UTC. minio-python apparently parses only a subset of RFC3339/ISO8601-formatted datetimes.
* Fixed missing quotes around etags for complete multipart requests.
* Allow for no spaces between values in the AWS auth v4 header. minio's go library will _sometimes_ issue these requests.

## Secondary changes

* Fixed the path reported when attempting to put a file in a path that includes an existing, non-directory parent path
* More logging to s3g to help diagnose future issues
* Have `reset.py` deploy pachd clusters with debug logs enabled
* Fixed various issues with pachd log formatting:
  * Only do special formatted output if both the service and method fields are set
  * Fixed a bug where fields were sometimes not printed
  * Removed frivolous type coercion of field json

Closes #4187

### CODE REVIEW

File: src/server/pfs/s3/auth.go [hunk 1]
```diff
@@ -3,32 +3,45 @@ package s3
...
 	"net/http"
 
+	"github.com/gorilla/mux"
 	"github.com/pachyderm/pachyderm/src/client/auth"
 	"github.com/pachyderm/pachyderm/src/client/pkg/errors"
 )
 
 func (c *controller) SecretKey(r *http.Request, accessKey string, region *string) (*string, error) {
-	pc, err := c.clientFactory.Client(accessKey)
+	c.logger.Debugf("SecretKey: accessKey=%+v, region=%+v", accessKey, region)
```
msteffen, line 12: Just checking: is `accessKey` secret? Are there risks associated with logging it?

src/server/pfs/s3/bucket.go
```diff
@@ -192,12 +190,13 @@ func (c *controller) CreateBucket(r *http.Request, bucketName string) error {
 }
 
 func (c *controller) DeleteBucket(r *http.Request, bucketName string) error {
+	c.logger.Debugf("DeleteBucket: %+v", bucketName)
```
msteffen, line 193: Would it make sense to use Pachyderm's internal log library to log GRPC fn call names more automatically?https://github.com/pachyderm/pachyderm/blob/b94abf8d1cbaccd8539c5bac85fdd90d43a6464e/src/server/pkg/log/log.go#L89

Example: https://github.com/pachyderm/pachyderm/blob/master/src/server/pps/server/api_server.go#L887-L889

src/server/pfs/s3/auth.go
ysimonson: Yeah it is, good point. I'll remove this.

src/server/pfs/s3/bucket.go
ysimonson: The `Log` function is specific to gRPC calls, and these aren't. I could make them log something syntactically more similar to gRPC log statements though, if you think that would be better?

ysimonson: Let me know if you want to address this at some point and I can open a separate ticket to fix.

PR #4910, opened 2020-05-21
Make pipelines more robust to potential missing objects

adelelopez:
Should mitigate most of https://github.com/pachyderm/pachyderm/issues/4889

Although we could not find a repro, we expect that the change made here will give us more information about what went wrong while also leaving the cluster in an easier state to recover from.


nitinjainsj:
@adelelopez Is this going in 1.11.0?

PR #4920, opened 2020-05-27
Wrap errors from vendor code with stack trace

brendoncarroll:


src/client/pkg/pbutil/pbutil.go
```diff
@@ -65,7 +66,7 @@ func (r *readWriter) WriteBytes(bytes []byte) (int64, error) {
 	}
 	lenByteSize := unsafe.Sizeof(int64(len(bytes)))
 	n, err := r.w.Write(bytes)
-	return int64(lenByteSize) + int64(n), err
+	return int64(lenByteSize) + int64(n), errors.WithStack(err)
```
brycemcanally, line 69: I don't think we want this here. This should not be the line between a vendor's code and ours. Some of the places above that use the binary / proto package seem like better candidates.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -339,7 +339,7 @@ func (d *driver) compact(master *work.Master, outputPath string, inputPrefixes [
 func serializeShard(shard *pfs.Shard) (*types.Any, error) {
 	serializedShard, err := proto.Marshal(shard)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error serializing shard")
```
brycemcanally, line 342: I think the extra message does not provide any value here. I think we would be better off with a `WithStack`.

src/server/pfs/server/driver_new_storage.go
```diff
@@ -350,7 +350,7 @@ func serializeShard(shard *pfs.Shard) (*types.Any, error) {
 func deserializeShard(shardAny *types.Any) (*pfs.Shard, error) {
 	shard := &pfs.Shard{}
 	if err := types.UnmarshalAny(shardAny, shard); err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error deserializing shard")
```
brycemcanally, line 353: Same as above.

src/server/pkg/collection/collection.go
```diff
@@ -115,7 +115,7 @@ func (c *collection) Claim(ctx context.Context, key string, val proto.Message, f
 		claimed = false
 		return nil
 	}); err != nil {
-		return err
+		return errors.Wrap(err, "error during claim")
```
brycemcanally, line 118: This isn't the line between the etcd code and ours. It is a few levels deeper.

brycemcanally:
Several of the places where stack collection was added are not actually at the boundary of vendor code and ours. Make another pass over these places to confirm that they are actually at the boundaries between our code and vendor code. Also, in general I think we are better suited by using `WithStack` by default and only using `Wrapf` when there is some clear value that can be added through an additional message.

PR #4921, opened 2020-05-27
Expose / implement deletion of tagged file content

brycemcanally:
This PR implements and exposes an rpc for file deletion and deletion of tagged content in a file. It also wraps the PFS rpcs that operate on files into one rpc (`FileOperation`) and fixes some issues encountered through deletion testing with the PFS fuzzer.


src/server/pkg/storage/fileset/priority_queue.go
```diff
@@ -7,6 +7,7 @@ import (
 type stream interface {
 	next() error
 	key() string
+	streamPriority() int
```
brendoncarroll, line 10: Isn't this implied by the ordering of `ss` in `newPriorityQueue` can this change after the fact?

src/server/pkg/storage/fileset/merge.go
```diff
@@ -381,26 +496,21 @@ func shard(mr *MergeReader, shardThreshold int64, f ShardFunc) error {
...
-		size += fmr.Index().SizeBytes
 		// A shard is created when we have encountered more than shardThreshold content bytes.
 		if size >= shardThreshold {
 			pathRange.Upper = fmr.Index().Path
 			if err := f(pathRange); err != nil {
 				return err
 			}
 			size = 0
-			pathRange = &index.PathRange{}
+			pathRange = &index.PathRange{
```
brendoncarroll, line 506: This is just a cleaner way of doing this, right? Not meaningful for this change?

brycemcanally: It is cleaner, but it also was necessary for handling deletions correctly. The merge reader does not return deleted files, but we need to apply the deletions during compaction. An easy case to see where the prior algorithm does not work is if the last shard should just be deletions. In this case, the shard would not get created because the callback would not be executed for that shard (since the merge reader won't return anything for deleted files). This makes sure that the full path range of the input filesets is accounted for.

PR #4928, opened 2020-05-28
Adds obj.Client factory for StorageV2

brendoncarroll:


src/server/pfs/server/server_v2.go
```diff
@@ -0,0 +1,40 @@
+package server
+
+import (
+	"github.com/pachyderm/pachyderm/src/server/pkg/obj"
+	"github.com/pachyderm/pachyderm/src/server/pkg/serviceenv"
+)
+
+func NewObjClient(conf *serviceenv.Configuration) (objClient obj.Client, err error) {
```
brycemcanally, line 8: We could probably clean this up a bit by just returning the client creation, rather than using named return values. Also, lets set dir using `dir := config.StorageRoot`. That is a more concise initialization and is more consistent with the codebase.

src/server/pfs/server/driver.go
```diff
@@ -170,14 +170,12 @@ func newDriver(
...
-		// (bryce) local client for testing.
-		// need to figure out obj_block_api_server before this
-		// can be changed.
-		objClient, err := obj.NewLocalClient(storageRoot)
+	if env.StorageV2 {
+		objClient, err := NewObjClient(env.Configuration)
 		if err != nil {
 			return nil, err
 		}
+
```
brycemcanally, line 178: Space.

PR #4934, opened 2020-05-29
[DOC] Add "Monitor Job Progress"

### LINKED ISSUE #4932: Clarify what progress bar colors mean
Explain how to interpret the progress bar in the `pachctl list commit <repo>` command output. 
It might be confusing to users whether the progress bar represents datums or jobs. It represents the jobs that run for downstream pipelines. For example, you have 3 pipelines that depend on each other. When the first pipeline starts, the progress bar will change to yellow. Then, when the job for the first pipeline completes, the progress bar changes to 1/3 green and 2/3 yellow. And so on. 

Related discussion: https://pachyderm-users.slack.com/archives/G011C8BGCJ1/p1590684076272500

svekars:
- Closes #4932


doc/docs/master/how-tos/monitor-job-progress.md
```diff
@@ -0,0 +1,52 @@
...
+
+After a pipeline starts a job, you can run one of the following commands
+to monitor its status:
+
+* `pachctl list pipeline`
+
+  This command shows all pipelines that run in your cluster
+  and their current status. Jobs for each pipeline will finish with either
+  `success` or `fail` status.
+
```
JoeyZwicker, line 11: I think we want to expand on more detail here. particularly, the STATE / LAST JOB field is important. 

STATE is the state of the pipeline. A pipeline state is "running" if the worker pods are spun up. That does _not_ necessarily mean that the pipeline is actively processing a job. A "failed" state indicates that the pipeline workers were not able to be created. This is mostly often something at the kubernetes or infrastructure level such as an incorrect image name or networking failures. 

LAST JOB indicates exactly that -- the status of the most recent job for that pipeline.

doc/docs/master/how-tos/monitor-job-progress.md
```diff
@@ -0,0 +1,52 @@
...
+  ```bash
+  NAME    VERSION INPUT                 CREATED       STATE / LAST JOB    DESCRIPTION
+  montage 1       (edges:/ ⨯ images:/)  2 seconds ago starting / starting A montage pipeline
+  edges   1       images:/*             2 seconds ago running / starting  An edge detection pipeline.
+  ```
+
+* `pachctl list job`
+
+  This command shows jobs that were run for each pipeline. For each job,
+  Pachyderm shows the number of datums in the **PROGRESS** section, the amount
```
JoeyZwicker, line 23: Might as well re-explain progress here: 

Format of the progress bar is DATUMS PROCESSED + DATUM SKIPPED / TOTAL DATUMS. See (LINK) for more details on datums and when they are skipped versus processed.

doc/docs/master/how-tos/monitor-job-progress.md
```diff
@@ -0,0 +1,52 @@
...
+  ID                               PIPELINE STARTED       DURATION           RESTART PROGRESS    DL       UL       STATE
+  7321952b9a214d3dbb64cc4369cc67da montage  6 minutes ago 1 second           0       1 + 0 / 1   371.9KiB 1.283MiB success
+  95adc138e82e48949909364e8b9dbb53 edges    6 minutes ago 1 second           0       2 + 1 / 3   181.1KiB 111.4KiB success
+  84fe22432f22492c9fd4f23036c3c8b5 montage  6 minutes ago Less than a second 0       1 + 0 / 1   79.49KiB 378.6KiB success
+  2fbbc54ab3514d8a94d1b7a75bab96a7 edges    6 minutes ago Less than a second 0       1 + 0 / 1   57.27KiB 22.22KiB success
+  ```
+
+* `pachctl list commit <first-repo-in-dag>`
+
+  This command, when ran against the first repo in your DAG,
```
JoeyZwicker, line 39: Does this only work in the first repo in your DAG? or does it also show in other repos -- of course only showing things downstream from them?

doc/docs/master/how-tos/monitor-job-progress.md
```diff
@@ -0,0 +1,52 @@
...
+  shows the status of the downstream commits further in the DAG.
```
JoeyZwicker, line 40:
```suggestion
  shows the status of the downstream jobs further in the DAG that result from this commit.
```

doc/docs/master/how-tos/monitor-job-progress.md
```diff
@@ -0,0 +1,52 @@
...
+  In the [Hyperparameter Tuning example](https://github.com/pachyderm/pachyderm/tree/master/examples/ml/hyperparameter), we have four pipelines,
+  or a four-stage pipeline. Every consequent pipeline takes the results
+  in the output repository of the previous pipeline and performs a
+  computation. Therefore, each step is executed one after another.
+  The **PROGRESS** bar in the output of the `pachctl list commit <first-repo-in-dag>`
+  command reflects these changes.
```
JoeyZwicker, line 46:
```suggestion
Specifically, it shows successful downstream jobs as green, running/waiting jobs as yellow, and failed jobs as red. One you hit a failed job, all jobs after that will also fail. 

For example, in the [Hyperparameter Tuning example](https://github.com/pachyderm/pachyderm/tree/master/examples/ml/hyperparameter), we have four pipelines. Every subsequent pipeline takes the results
  in the output repository of the previous pipeline(s) and performs a
  computation. `
```
:) 
Consequent - following as a result or causal outcome of
Subsequent - Coming after

doc/docs/master/how-tos/monitor-job-progress.md
svekars: added.

doc/docs/master/how-tos/monitor-job-progress.md
svekars: added

doc/docs/master/how-tos/monitor-job-progress.md
svekars: I've tried to run it against the hyperparameter tuning example and I could only see the bar updating as the jobs were running for the `raw_data` repo. For others, it just turned green from yellow. Although, it might be just this particular example and it would be different in another use case. Still, I think it probably makes more sense to run it against the first repo in the DAG to see the whole picture.

doc/docs/master/how-tos/monitor-job-progress.md
svekars: corrected.

doc/docs/master/how-tos/monitor-job-progress.md
JoeyZwicker: > STATE is the state of the pipeline. A pipeline state is "running" if the worker pods are spun up. That does not necessarily mean that the pipeline is actively processing a job. A "failed" state indicates that the pipeline workers were not able to be created. This is mostly often something at the kubernetes or infrastructure level such as an incorrect image name or networking failures.

I think your version misses a few important items that I have in mine. Specifically:

> The running state does not necessarily mean
  that the pipeline is actively running

Yes it does. It does mean the pipeline is running, but it does _not_ mean it is actively processing a job. I think your phrasing should be more precise. 

Second, I think we should note that "failed" is a terminal state for a pipeline -- meaning the user will need to update it in order to get it out of that state. 

Also note, as JD demoed last week, in 1.11 (including the current beta1) there is a new state for pipelines called `crashing`. `crashing` indicates that the worker that is failing, but for potentially transient reasons. The two most common reasons for `crashing` are image pull failures (e.g. incorrect image name or registry credentials) or scheduling failures (e.g. not enough resources on your k8s cluster).

svekars: Added. Will also need to update the troubleshooting section with this.

PR #4941, opened 2020-06-02
Add limitedObjClient for controlling upload and download concurrency

brendoncarroll:


src/server/pkg/storage/chunk/limit_objclient.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"context"
+	io "io"
+	"math"
+
+	"github.com/pachyderm/pachyderm/src/server/pkg/obj"
+	"golang.org/x/sync/semaphore"
+)
+
+var _ obj.Client = limitedObjClient{}
```
brycemcanally, line 12: Is this checking if `limitedObjClient` satisfies the `obj.Client` interface?

src/server/pkg/storage/chunk/limit_objclient.go
```diff
@@ -0,0 +1,74 @@
...
+
+type limitedObjClient struct {
+	obj.Client
+	writersSem *semaphore.Weighted
+	readersSem *semaphore.Weighted
+}
+
+func newLimitedObjClient(client obj.Client, maxReaders, maxWriters int) limitedObjClient {
```
brycemcanally, line 20: I think it would be better to operate on a pointer to a `limitedObjClient`.

src/server/pkg/storage/chunk/option.go
```diff
@@ -19,10 +19,21 @@ func WithGarbageCollection(gcClient gc.Client) StorageOption {
 	}
 }
 
+// WithMaxConcurrentObjects sets the maximum number of object writers (upload)
+// and readers (download) that can be open at a time.
+func WithMaxConcurrentObjects(maxUpload, maxDownload int) StorageOption {
```
brycemcanally, line 24: The fact that the parameters for this function are upload, download and the internal call's are download, upload is bothering me lol. Would you mind making those consistent?

src/server/pkg/storage/chunk/limit_objclient.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/pachyderm/pachyderm/src/server/pkg/obj"
+	"golang.org/x/sync/semaphore"
+)
+
+var _ obj.Client = limitedObjClient{}
+
+type limitedObjClient struct {
+	obj.Client
+	writersSem *semaphore.Weighted
```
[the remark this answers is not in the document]
brendoncarroll: Those don't allow us to respect the context passed in to the Reader and Writer constructors.  It can't handle "I want a writer but I'm only willing to wait 1 minute".

src/server/pkg/storage/chunk/limit_objclient.go
brendoncarroll: yeah

src/server/pkg/storage/chunk/limit_objclient.go
brendoncarroll: okay can change it. I don't really have any strong intuitions about which is more performant.  Essentially following one pointer to get to 3 rather than copying 3 around.

src/server/pkg/storage/chunk/limit_objclient.go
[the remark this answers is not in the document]
brycemcanally: Fair point.

src/server/pkg/storage/chunk/limit_objclient.go
brycemcanally: I think the performance difference doesn't matter in this case, I just prefer it for stylistic consistency.

PR #4974, opened 2020-06-09
Split StorageV2 functionality into a separate driver and apiServer

brendoncarroll:
- Creates driverV2
- Creates apiServerV2
- APIServer interface contains old and new methods.
- apiServer implementation returns "not implemented" error for new methods.


src/client/pfs/pfs.proto
```diff
@@ -624,12 +624,12 @@ service API {
   rpc DeleteAll(google.protobuf.Empty) returns (google.protobuf.Empty) {}
   // Fsck does a file system consistency check for pfs
   rpc Fsck(FsckRequest) returns (stream FsckResponse) {}
-
-  // RPCs specific to Pachyderm 2.
-  rpc PutTarV2(stream PutTarRequestV2) returns (google.protobuf.Empty) {}
-  rpc GetTarV2(GetTarRequestV2) returns (stream google.protobuf.BytesValue) {}
+  
+  // V2 methods
```
brycemcanally, line 628: Would you change the comment back to the original. It is more descriptive and is a pattern I am following elsewhere (* specific to Pachyderm 2).

src/client/pfs/pfs.proto
```diff
@@ -513,14 +513,14 @@ message FileInfoV2 {
 //     Client sends a request with the EOF field set to true.
 //   Client connection is closed.
 
-message PutTarRequestV2 {
+message PutTarRequest {
```
brycemcanally, line 516: It seems like it would be better to keep the V2 suffix for all messages / rpcs specific to Pachyderm 2. Although we don't need it for name collisions, it seems like it would be a good design to have clear lines between the V1 and V2 functionality.

src/server/pfs/server/api_server.go
```diff
@@ -599,6 +595,27 @@ func (a *apiServer) DeleteAll(ctx context.Context, request *types.Empty) (respon
 	return &types.Empty{}, nil
 }
 
+// V2 Methods
+// GetTar not implemented
+func (a *apiServer) GetTar(request *pfs.GetTarRequest, server pfs.API_GetTarServer) (retErr error) {
+	return errors.Errorf("method not implemented")
```
brycemcanally, line 601: Probably could use a more descriptive message that mentions that this is a V2 rpc.

src/server/pfs/server/api_server_v2.go
```diff
@@ -148,34 +183,45 @@ func (a *apiServer) GetTarConditionalV2(server pfs.API_GetTarConditionalV2Server
...
+func (a *apiServerV2) ListFileV2(req *pfs.ListFileRequest, server pfs.API_ListFileV2Server) error {
 	pachClient := a.env.GetPachClient(server.Context())
 	return a.driver.listFileV2(pachClient, req.File, req.Full, req.History, func(finfo *pfs.FileInfoV2) error {
 		return server.Send(finfo)
 	})
 }
+
+// FinishCommitInTransaction is identical to FinishCommit except that it can run
+// inside an existing etcd STM transaction.  This is not an RPC.
+func (a *apiServerV2) FinishCommitInTransaction(
```
brycemcanally, line 220: Does this work? It seems like a FinishCommit call on *apiServerV2 would get dispatched to the embedded *apiServer and the subsequent calls would have *apiServer as the receiver, not the "parent" *apiServerV2.

brycemcanally: I think any V2 code changes will require the rpc function to be attached to *apiServerV2, otherwise you would be making changes to *apiServer which I think is what you want to avoid with this.

src/server/pfs/server/driver_v2.go
```diff
@@ -35,7 +39,49 @@ const (
...
+	etcdPrefix string,
+	treeCache *hashtree.Cache,
+	storageRoot string,
+	memoryRequest int64,
+) (*driverV2, error) {
+	d1, err := newDriver(env, txnEnv, etcdPrefix, treeCache, storageRoot, memoryRequest)
+	if err != nil {
+		return nil, err
+	}
+	d := &driverV2{driver: d1}
```
brycemcanally, line 59: d -> d2

src/server/pfs/server/api_server_v2.go
brendoncarroll: Replying here for visibility: Because this calls another rpc method it starts from the external interface (in this case APIServer).  The RealEnv setup passes an APIServer (the interface not the implementation) to txnEnv

PR #4975, opened 2020-06-09
Add Cluster-wide FS Admin role

actgardner:
Pachyderm currently has two types of permissions:
- ACLs, which are scoped to repos
- Admin status, which is binary and scoped to the entire cluster

To permit more flexible cluster-scoped permissions, this introduces the concept of admin roles which replace the current binary admin status. Users can have zero or more admin roles including `SUPER`, which is equivalent to the current `admin` role and permits all operations on the cluster.

This also adds the `FS` role, which gives the user `OWNER` on all repos so they can read/write/modify ACLs, but they can't do things like impersonate other users by generating OTPs, or configure auth on the cluster, or manage group membership (when we wire up groups). This is immediately useful in Hub so we can give org owners limited admin control over their own clusters without exposing features that aren't supported.

The CLI/RPCs model `FS` and `OWNER` as two types of admin roles, but on the backend we store them in different etcd collections - to make it possible to upgrade from 1.10. In the future if we introduce more roles it would make sense to store them as a single collection keyed on the principal.


src/server/auth/server/api_server.go
```diff
@@ -374,12 +374,12 @@ func (a *apiServer) watchAdmins(fullAdminPrefix string) {
 
 				// Parse event data and potentially update adminCache
 				var key string
-				var boolProto types.BoolValue
-				ev.Unmarshal(&key, &boolProto)
+				var grant auth.AdminGrant
+				ev.Unmarshal(&key, &grant)
```
actgardner, line 378: This probably needs an error-guard for actually deserialzing correctly.

src/server/pfs/server/driver.go
```diff
@@ -333,9 +333,15 @@ func (d *driver) getAccessLevel(pachClient *client.APIClient, repo *pfs.Repo) (a
 	if err != nil {
 		return auth.Scope_NONE, err
 	}
-	if who.IsAdmin {
-		return auth.Scope_OWNER, nil
+
+	if who.AdminGrant != nil {
```
actgardner, line 337: This method lives in the PFS server but it seems like a re-implementation of Authorize in the auth server - especially if we wire up group membership it probably makes sense to consolidate this logic in the auth server.

src/client/auth/auth.proto
```diff
@@ -157,17 +157,30 @@ message SetConfigurationRequest {
...
+  repeated Scope scopes = 1;
+}
+
 // Get the current list of cluster admins
 message GetAdminsRequest{}
 message GetAdminsResponse{
-  // admins contains the list of cluster admins
-  repeated string admins = 1;
+  // admins contains a mapping of principals to admin grants 
+  map<string, AdminGrant> admins = 1;
```
msteffen, line 177: I think if we want to change the field, we should mark field 1 reserved and add a new field (it's a protobuf convention to avoid serialization errors across versions—https://developers.google.com/protocol-buffers/docs/proto3#reserved explains the rationale).

Also, I think we might need to update our other language clients (JS, Python, Ruby, etc). We maybe don't need to do it in this PR, but I think we'll need to do it before 1.11, so we should make sure to track it. Also, I think the only thing to do on our end is tell Customer Team, but users using those libraries will have to update their clients along with the update to 1.11.

src/client/auth/auth.proto
```diff
@@ -157,17 +157,30 @@ message SetConfigurationRequest {
...
 }
 
-// Add or remove cluster admins
+// Set admin grants for the specified admin.
```
msteffen, line 180: Similar to above, I think we should mark fields 1 and 2 reserved and make these 3 and 4.

Also, we should make sure there's a way to revoke a user's admin privileges (If we add a `NONE` scope, it could be by setting `grant` to `NONE`, which is how SetScope works)

src/client/auth/auth.proto
```diff
@@ -232,8 +245,8 @@ message WhoAmIRequest {}
 
 message WhoAmIResponse {
   string username = 1;
-  bool is_admin = 2;
```
msteffen, line 235: We should mark field 2 reserved, so it doesn't get re-used.

src/server/auth/cmds/cmds.go
```diff
@@ -401,22 +403,35 @@ func ListAdminsCmd() *cobra.Command {
...
-		Short: "Modify the current cluster admins",
-		Long: "Modify the current cluster admins. --add accepts a comma-" +
-			"separated list of users to grant admin status, and --remove accepts a " +
-			"comma-separated list of users to revoke admin status",
+		Short: "Set the admin scopes for a given user",
+		Long: "Set the cluster admin scopes for a given user." +
+			"--user is the user to modify, and --scopes is a comma-separated list of scopes." +
+			"An empty list removes all admin scopes for the user.",
 		Run: cmdutil.Run(func([]string) error {
 			c, err := client.NewOnUserMachine("user")
```
msteffen, line 414: IMO this should validate that `user` isn't empty (and perhaps that `scopes` isn't empty, though it would be reasonable in that case to simply do nothing)

Given that IIUC, both `user` and `scopes` would typically be set, it might be ergonomic to make them positional arguments and change the command to `modify-admin` (singular), so that invocation looks like `pachctl auth modify-admin alice super`, analogous to `pachctl auth set alice reader myrepo`.

(It also occurred to me that we could fold this into `pachctl auth set` and just have `cluster` take the place of the repo name, so `pachctl auth set alice super cluster`, but then we'd have to ban repos named `cluster`)

src/server/auth/server/api_server.go
```diff
@@ -750,57 +747,30 @@ func (a *apiServer) ModifyAdmins(ctx context.Context, req *auth.ModifyAdminsRequ
...
+
+	err = a.validateModifyAdminsRequest(canonicalizedUser, *req.Scopes)
 	if err != nil {
 		return nil, err
 	}
 
-	// Update "admins" list (watchAdmins() will update admins cache)
-	for _, user := range canonicalizedToAdd {
+	// Update "admins" list (watchAdmins() will update admins cache).
+	if len(req.Scopes.Scopes) > 0 {
```
msteffen, line 761: Probably obvious, but here (and in `validateModifyAdminsRequest` we should make sure to change logic a bit so that if a principal is being given the `NONE` admin scope, then they're deleted from the admin collection (I think currently here and in `validate...` is uses the old logic, which just writes the `NONE` scope to etcd)

actgardner:
- Renamed Scopes to Roles for the admin case
- Changed `modify-admin` to be singular and take two positional arguments, `pachctl auth modify-admin <user> <roles>`

"Specify an empty string for roles to remove all roles" isn't the best UX, but I don't think it makes sense to make the 0 value for AdminRoles_Role mean "delete all the user's admin scopes". For one, that violates the idea that the zero value for an enum is unspecified and indicates incompatibility. Also, it would permit a list of roles like "NONE,SUPER" which is nonsense we'd need to validate out. We could add a placeholder in the CLI to indicate "remove all scopes" like special-casing the string "NONE".

Since this is primarily for Hub 1.0, if we don't think this can land in 1.11 because of upgrade concerns we should chat about alternatives. My goal here was mostly to spike out the feasibility of getting this feature into 1.11.

src/server/auth/server/api_server.go
actgardner: This code currently detects if there's no admin with `SUPER`, and removes the entry in etcd if the list of roles is empty. I think this misunderstanding is because `UNAVAILABLE` doesn't mean "remove all roles" in this PR, it's an invalid value. An empty list of roles means "remove all roles".

msteffen: Ah, sorry, I honestly just forgot that the enum was structured this way—this makes sense!

actgardner:
Based on the discussion with @nitinjainsj and @msteffen we're going to implement this with two separate etcd collections, one for existing admins (now "super admins") and one for the new set of FS admins. This avoids the need to extract/restore the auth info which isn't implemented, and allows us to ship this change in 1.11.

From the user perspective we decided to start the move towards cluster roles, so `modify-admin` still supports `SUPER` and `FS`, `whoami` returns a set of cluster roles instead of just `isAdmin`, etc. The plan is to refactor the backend once we have the capacity to migrate existing etcd collections.

actgardner:
I've fixed all the existing tests to use the SUPER role and added FS admin tests for Repo RWO and sensitive RPCs (deactivate, configuring auth, generating tokens and OTPs).

src/server/auth/cmds/cmds.go
```diff
@@ -389,34 +391,52 @@ func ListAdminsCmd() *cobra.Command {
...
-	var remove []string
-	modifyAdmins := &cobra.Command{
-		Short: "Modify the current cluster admins",
-		Long: "Modify the current cluster admins. --add accepts a comma-" +
-			"separated list of users to grant admin status, and --remove accepts a " +
-			"comma-separated list of users to revoke admin status",
-		Run: cmdutil.Run(func([]string) error {
+func ModifyAdminCmd() *cobra.Command {
+	modifyAdmin := &cobra.Command{
+		Use:   "{{alias}} <user> <scopes>",
```
actgardner, line 407: scopes -> roles

src/server/auth/server/api_server.go
```diff
@@ -684,33 +701,44 @@ func (a *apiServer) GetAdmins(ctx context.Context, req *auth.GetAdminsRequest) (
...
+	// Check to make sure that req doesn't remove all cluster SUPER admins
+	m := make(map[string]interface{})
+
 	// copy existing admins into m
 	func() {
 		a.adminMu.Lock()
 		defer a.adminMu.Unlock()
-		for u := range a.adminCache {
-			m[u] = struct{}{}
+		for u, roles := range a.adminCache {
```
actgardner, line 730: roles -> _

src/server/auth/server/api_server.go
```diff
@@ -990,6 +1038,8 @@ func (a *apiServer) GetOneTimePassword(ctx context.Context, req *auth.GetOneTime
 		return nil, err
 	}
 
+	fmt.Printf("Req subject: %q, caller: %q, isadmin: %v\n", req.Subject, callerInfo.Subject, isAdmin)
```
actgardner, line 1041: delete

src/server/auth/server/testing/admin_test.go
```diff
@@ -172,14 +201,14 @@ func TestAdminRWO(t *testing.T) {
 
 	// 'admin' revokes bob's admin status
 	_, err = adminClient.ModifyAdmins(adminClient.Ctx(),
-		&auth.ModifyAdminsRequest{Remove: []string{bob}})
+		&auth.ModifyAdminsRequest{Principal: gh(bob), Roles: &auth.AdminRoles{Roles: []auth.AdminRoles_Role{}}})
```
actgardner, line 204: rewrite with helper

actgardner:
@msteffen Should be ready for another pass.

src/server/auth/server/api_server.go
```diff
@@ -347,23 +359,48 @@ func (a *apiServer) retrieveOrGeneratePPSToken() {
 	}
 }
 
-func (a *apiServer) watchAdmins(fullAdminPrefix string) {
+func (a *apiServer) watchAdmins(superAdminPrefix, fsAdminPrefix string) {
```
msteffen, line 362: The way this whole function works now is quite nice

src/server/auth/server/testing/auth_test.go
```diff
@@ -253,7 +253,12 @@ func getPachClientP(tb testing.TB, subject string, checkConfig bool) *client.API
 		if len(getAdminsResp.Admins) == 0 {
 			panic("it should not be possible to leave a cluster with no admins")
 		}
-		adminClient = getPachClientInternal(tb, getAdminsResp.Admins[0])
+		var currentAdmin string
+		for a := range getAdminsResp.Admins {
+			currentAdmin = a
+			break
+		}
+		adminClient = getPachClientInternal(tb, currentAdmin)
```
msteffen, line 261: Just trying to understand this block of code: I think if `getAdminsResp` is empty, then currentAdmin is `""` and `adminClient` is an anonymous client (vs `Admins[0]`, which crashes). Does that sound right? Is that expected behavior?

src/server/auth/server/api_server.go
```diff
@@ -750,57 +814,79 @@ func (a *apiServer) ModifyAdmins(ctx context.Context, req *auth.ModifyAdminsRequ
...
+
+	err = a.validateModifyAdminsRequest(canonicalizedUser, grantSuper)
 	if err != nil {
 		return nil, err
 	}
 
+	// Get the user's existing grants
+	var hasSuper bool
+	var hasFS bool
+	func() {
```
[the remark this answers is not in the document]
actgardner: Rewrote this particular instance to do everything based on etcd state in a single STM. I think this line of thinking also extends to all of the admin interface - we currently use the cache to check whether an operation will remove all SUPER admins, but we read from the cache so it's entirely possible to interleave operations like:

- 2 admins, A and B
- RPC to remove A checks the cache, sees B
- RPC to remove B checks the cache, sees A
- A is removed from etcd
- B is removed from etcd
- no admins exist

Do we have a sense of how expensive it would be to just eliminate the cache altogether? I'm thinking along the lines of "ListAdmins should just go to the source, but then we have no good way to interrogate the state of the cache".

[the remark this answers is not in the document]
actgardner: I looked into it briefly, listing keys doesn't seem to be supported in STM so I'm not sure how we could remove the race in checking for at least one admin. One possibility would be to keep a separate integer key we increment and decrement as part of the transaction. Another would be to reorganize the data into a single, large key with all admins. Either of those seems beyond the scope of this PR.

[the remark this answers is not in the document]
msteffen: Yeah this is a really good point on both counts (that we shouldn't use the cache to validate modifyAdmins and that fixing this will be hard).

I haven't finished review the new commits yet, but I think what you already did (fixing this instance, since it can be done in an STM easily, and ignoring admin validation for now) makes complete sense to me.

(For 1.12, it seems like it's definitely time to implement auth extract/restore, which would let us change how admins are represented. I like the idea of having a single large key for all admins—it would amount to a cluster-level ACL, which is consistent with the GCP-ization of the auth model, where the cluster is a resource and SUPER and FS are the roles with respect to that resource.

Finally, if I'm being honest, part of me thinks the cache might still have value for `Authorize()` and `WhoAmI` since they get called so much—every RPC, and a single top-level user RPC might trigger a lot of internal RPCs—but a) I haven't tested it, and b) there are definitely safer ways to implement that. E.g. making sure the cache is only used in Authorize and WhoAmI, and nowhere else, or caching it but keying the cache with an ID tied to the top-level request, or embeddingsome of the caller's resource roles, including admin status, in the RPC metadata and propagating that for internal requests).

src/server/auth/server/testing/admin_test.go
```diff
@@ -262,116 +436,89 @@ func TestAdminFixBrokenRepo(t *testing.T) {
 	require.Equal(t, 1, CommitCnt(t, adminClient, repo)) // check that a new commit was created
 }
 
-// TestModifyAdminsErrorMissingAdmin tests that trying to remove a nonexistant
-// admin returns an error
-func TestModifyAdminsErrorMissingAdmin(t *testing.T) {
```
msteffen, line 267: Just noting here for posterity that this (removing all roles from an admin that has none to begin with) is no longer an error

src/server/auth/server/testing/admin_test.go
```diff
@@ -262,116 +436,89 @@ func TestAdminFixBrokenRepo(t *testing.T) {
...
-		&auth.ModifyAdminsRequest{Remove: []string{alice}})
+		&auth.ModifyAdminsRequest{Principal: alice})
 	require.YesError(t, err)
 	require.NoError(t, backoff.Retry(func() error {
 		resp, err = aliceClient.GetAdmins(aliceClient.Ctx(), &auth.GetAdminsRequest{})
 		require.NoError(t, err)
-		return require.ElementsEqualOrErr([]string{gh(alice)}, resp.Admins)
-	}, backoff.NewTestingBackOff()))
-
-	// alice *can* swap herself and "admin"
```
msteffen, line 364: Just noting for posterity, because `modify-admins` is now `modify-admin`, swapping one admin for another isn't something that can be done transactionally, so there's nothing to test.

actgardner:
Since this changes existing RPCs (as noted above) do we need to coordinate changes in other language libraries or dash before this can merge?

actgardner:
@msteffen @jdoliner @nitinjainsj Updated to be backwards compatible at the API layer, although I'm not super happy about the naming of `GetAdminsV2`. I think we're going to revisit this more comprehensively in 2.0 so it might not matter?

The CLI isn't backwards compatible, if that's also a concern.

nitinjainsj:
> I'll defer to @nitinjainsj regarding whether we're OK changing the CLI. Also @Tryneus worked on our last big CLI change and might have some ideas for how to set up an alias to the existing `modify-admins` command.
> 
> One thought that occurred to me is that if we need to leave `modify-admins` in place, instead of `modify-admin` and `modify-admins` existing concurrently, we could add a `--cluster` flag to `pachctl auth set`, so users would run something like `pachctl auth set oidc:alysha@pachyderm.io super-admin --cluster`, analogous to the existing `pachctl auth set oidc:alysha@pachyderm.io reader my-repo`?

The expectation we set with our users is that a minor & point release does not break backward compatibility across upgrade/downgrade. We don't make a promise on the CLI, but in the spirit of SemVar releases, we want to avoid breaking any existing automation.

Unless there is a specific use case, having multiple CLI that performs the same (sub)task makes is very confusing - I am not a big fan of it.

Another option is to leave the CLI unchanged and only support fs-admin via the APIs. This will work for what we need on Hub but not introduce a new role for general core pach users yet. As we re-work core pach auth, we introduce a new set of roles (including fs-admin) and permissions. @msteffen and I talked about the model he has in mind, I'll post a recording of it on slack.

actgardner:
- terminology is now "cluster role" and "binding"
- reverted any changes to the CLI

PR #4991, opened 2020-06-11
[DOC] Clarify err_cmd

### LINKED ISSUE #4980: Explain how to skip a datum with err_cmd without failing a job
Based on https://pachyderm-users.slack.com/archives/C3UN0S0Q0/p1591807888396700

Add a note to the https://docs.pachyderm.com/latest/how-tos/err_cmd/ section that if you want to skip all errors and not fail the job even if a datum fails, you need to set `"err_cmd" : ["true"]`.

svekars:
Closes #4980

### DIFF
--- a/doc/docs/1.10.x/how-tos/err_cmd.md
+++ b/doc/docs/1.10.x/how-tos/err_cmd.md
@@ -65,5 +65,8 @@ If one of the datums is marked as *recovered* and two others are
 successfully processed, only these two successful datums are used in
 the next pipeline.
 
+If you want to skip all errors and let the job proceed with the
+successful datums only, set `"err_cmd" : ["true"]`.
+
 !!! note "See Also:"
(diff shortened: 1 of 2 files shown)


doc/docs/1.10.x/how-tos/err_cmd.md
```diff
@@ -65,5 +65,8 @@ If one of the datums is marked as *recovered* and two others are
 successfully processed, only these two successful datums are used in
 the next pipeline.
 
+If you want to skip all errors and let the job proceed with the
+successful datums only, set `"err_cmd" : ["true"]`.
```
JoeyZwicker, line 69:
```suggestion
If you want to let the job proceed with only the
successful datums being written to the output, set `"err_cmd" : ["true"]`. Note that failed datums (which are "recovered" by your err_cmd in this way, will be retried on the next job, just as failed datums. 
```

JoeyZwicker: Same change to below

PR #4998, opened 2020-06-15
Run auth tests in CI and fix auth tokens with no expiration

actgardner:
Fixes:
- `make test-auth` was calling `go test` on the wrong package so no tests were running
- the regex in the travis bucket script was too permissive and filtered legit tests
- the test for admins getting auth tokens with infinite TTLs was failing for several reasons

### DIFF
--- a/etc/testing/travis.sh
+++ b/etc/testing/travis.sh
@@ -68,7 +68,7 @@ function test_bucket {
 
     echo "Running bucket $bucket_num of $num_buckets"
     # shellcheck disable=SC2207
-    tests=( $(go test -v  "${package}" -list ".*" | grep -v ok | grep -v Benchmark) )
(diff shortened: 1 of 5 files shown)


etc/testing/travis.sh
```diff
@@ -68,7 +68,7 @@ function test_bucket {
 
     echo "Running bucket $bucket_num of $num_buckets"
     # shellcheck disable=SC2207
-    tests=( $(go test -v  "${package}" -list ".*" | grep -v ok | grep -v Benchmark) )
+    tests=( $(go test -v  "${package}" -list ".*" | grep -v '^ok' | grep -v '^Benchmark') )
```
actgardner, line 71: This was filtering any test with `ok` in it, like say "Token". Looks like it affected 14/89 tests in the auth package.

PR #5004, opened 2020-06-16
Run auth tests in CI (1.10.x)

msteffen:


msteffen:
> Does this mean the indefinite TTL admin token isn't being used anywhere? Would it make sense to just eliminate it?

There's one customer in particular that we added this feature for, and I think we've managed to avoid trouble with them b/c they're still on 1.9. We could maybe check in with them, though

PR #5007, opened 2020-06-16
CI Improvements

### LINKED ISSUE #4983: Send builds from CI to a Docker Repository
It would be good if we could start sending the builds we create in CI to dockerhub or GCR (For eventual consumption in hub) I'd rather not clog up our main docker account, but maybe create a public pachyderm-builds account in dockerhub and send them there?

### LINKED ISSUE #4985: CI: Experiment with switching the examples bucket to hub

ysimonson:
Some improvements to our CI infra:

* Removed build container caching. In practice, this has slowed things down more than it's helped, because (1) build containers are invalidated frequently, (2) even though we `rm -rf` all the old images on every run, I _suspect_ travisci does not recognize this, causing the cache to bloat. If this is true, it's likely what's causing the frequent timeouts in TravisCI.
* Added a separate build step run before any tests. When the build completes, the images are pushed to docker hub, and then each test runner pulls the images. As a result, we only build once for all tests, rather than for each test. This will substantially reduce our TravisCI node utilization, and should reduce failure rate further. And we need to push images to docker hub anyway to do #4985.

Closes #4983


.travis.yml
```diff
@@ -33,21 +32,26 @@ env:
...
   - etc/testing/travis_cache.sh
-  - etc/testing/travis_before_install.sh
 install:
   - etc/testing/travis_install.sh
-before_script:
-  - docker version
-script:
-  - etc/testing/travis.sh
+script: etc/testing/travis.sh
 before_cache:
```
nitinjainsj, line 51: Same thing here. Do we need before_cache step?

ysimonson: Yes, we need both because they both need to run `travis_cache.sh`. While we're no longer caching the docker build container, other directories are still being cached - and due to a TravisCI bug, we need to fix permissions on them for caching to work.

ysimonson:
You can see which directories are still cached under the `cache/directories` list, toward the top of `.travis.yml`.

PR #5008, opened 2020-06-16
GlobFileV2

brendoncarroll:
Adds GlobFileV2.  Branches from the storage v2 driver split.


src/server/pfs/server/api_server.go
```diff
@@ -616,6 +616,11 @@ func (a *apiServer) ListFileV2(req *pfs.ListFileRequest, server pfs.API_ListFile
 	return errors.Errorf("v2 method not implemented")
 }
 
+// GlobFileV2 not implemented
+func (a *apiServer) GlobFileV2(req *pfs.GlobFileRequest, server pfs.API_GlobFileV2Server) error {
+	return errors.Errorf("method not implemented")
```
brycemcanally, line 621: This error message should be consistent with the others. Maybe this should be a constant that is returned by each of the V2 methods.

src/server/pfs/server/master_v2.go
```diff
@@ -21,7 +21,7 @@ const (
 func (d *driverV2) master(env *serviceenv.ServiceEnv, objClient obj.Client, db *gorm.DB) {
 	ctx := context.Background()
 	masterLock := dlock.NewDLock(d.etcdClient, path.Join(d.prefix, masterLockPath))
-	err := backoff.RetryNotify(func() error {
+	err := backoff.RetryUntilCancel(ctx, func() error {
```
brycemcanally, line 24: I don't think this is going to solve the problem. If the master context gets canceled, then the run function is going to return nil and we will exit the retry. We want the master goroutine to try and re-claim the master lock after it loses it. The retry until cancel should be in the polling function in the garbage collector.

src/server/pfs/server/master_v2.go
```diff
@@ -21,21 +21,24 @@ const (
 func (d *driverV2) master(env *serviceenv.ServiceEnv, objClient obj.Client, db *gorm.DB) {
 	ctx := context.Background()
 	masterLock := dlock.NewDLock(d.etcdClient, path.Join(d.prefix, masterLockPath))
-	err := backoff.RetryNotify(func() error {
-		masterCtx, err := masterLock.Lock(ctx)
-		if err != nil {
+	for {
```
brycemcanally, line 24: This outer for loop should be unnecessary. If we get in this situation, then there is a bug elsewhere that is causing us to return nil in the retry. I think we just want the panic if we ever exit the retry.

src/server/pkg/testpachd/mock_pachd.go
```diff
@@ -717,6 +721,12 @@ func (api *pfsServerAPI) ListFileV2(req *pfs.ListFileRequest, serv pfs.API_ListF
 	}
 	return errors.Errorf("unhandled pachd mock pfs.ListFileV2")
 }
+func (api *pfsServerAPI) GlobFileV2(req *pfs.GlobFileRequest, serv pfs.API_GlobFileV2Server) error {
+	if api.mock.ListFileV2.handler != nil {
```
brycemcanally, line 725: ListFileV2 -> GlobFileV2

src/server/pkg/testpachd/mock_pachd.go
```diff
@@ -717,6 +721,12 @@ func (api *pfsServerAPI) ListFileV2(req *pfs.ListFileRequest, serv pfs.API_ListF
...
+		return api.mock.GlobFileV2.handler(req, serv)
+	}
+	return errors.Errorf("unhandled pachd mock pfs.ListFileV2")
```
brycemcanally, line 728: Same as above.

PR #5010, opened 2020-06-16
Fixed overwriting old transaction id with new one

farhaanbukhsh:
This change fixed overwritting old transaction id with new one, so now no more transactions
are created unless and untill the previous transaction gets completed.

Signed-off-by: Farhaan Bukhsh <farhaan.bukhsh@gmail.com>

### DIFF
--- a/src/server/transaction/cmds/cmds.go
+++ b/src/server/transaction/cmds/cmds.go
@@ -92,6 +92,14 @@ transaction' or cancelled with 'delete transaction'.`,
 				return err
 			}
 			defer c.Close()
+			txn, err := getActiveTransaction()
+			if err != nil {
+				return err
+			}
+			if txn != nil {
+				return errors.Errorf("cannot start a new transaction, since transaction with ID %q already exists", txn.ID)
+			}
+
 			transaction, err := c.StartTransaction()
 			if err != nil {
 				return grpcutil.ScrubGRPC(err)
@@ -102,7 +110,7 @@ transaction' or cancelled with 'delete transaction'.`,
 			if err != nil {
 				return err
 			}
-			fmt.Printf("Started new transaction: %s\n", transaction.ID)
+			fmt.Printf("started new transaction: %q\n", transaction.ID)
 			return nil
 		}),
 	}


src/server/transaction/cmds/cmds.go
```diff
@@ -92,18 +92,28 @@ transaction' or cancelled with 'delete transaction'.`,
...
-			if err != nil {
-				return grpcutil.ScrubGRPC(err)
-			}
-			// TODO: use advisory locks on config so we don't have a race condition if
-			// two commands are run simultaneously
-			err = setActiveTransaction(transaction)
+			txn, err := getActiveTransaction()
 			if err != nil {
 				return err
 			}
```
ysimonson, line 98: nit: it's a bit more conventional in go to minimize nesting and avoid else blocks, i.e. do this instead:

```go
if txn != nil {
    return errors.Errorf("cannot start a new transaction, since transaction with ID %q already exists", txn.ID)
}
```

Also note a couple of other changes in that code:

1) Errors messages are lowercased, I'm not sure if this is a go convention or just a convention we follow though
2) Use `%q` to format `txn.ID`, so it'll be surrounded by quotes

src/server/transaction/cmds/cmds.go
```diff
@@ -92,18 +92,28 @@ transaction' or cancelled with 'delete transaction'.`,
...
+				if err != nil {
+					return grpcutil.ScrubGRPC(err)
+				}
+				// TODO: use advisory locks on config so we don't have a race condition if
+				// two commands are run simultaneously
+				err = setActiveTransaction(transaction)
+				if err != nil {
+					return err
+				}
+				fmt.Printf("Started new transaction: %s\n", transaction.ID)
```
ysimonson, line 111: use lowercase here as well

src/server/transaction/cmds/cmds.go
farhaanbukhsh: hey thanks, I didn't know about this convention.

farhaanbukhsh: hey thanks, I didn't know about this convention. Will keep in mind :smile:

src/server/transaction/cmds/cmds.go
farhaanbukhsh: done

farhaanbukhsh:
@ysimonson CI seems to be stuck here as well

ysimonson:
This one's good! Just the docs CI acting up, but since this isn't a docs change, it's OK.

PR #5011, opened 2020-06-17
Fix error comparisons

Tryneus:
Since we've moved to using stackful errors, many of our error comparisons (e.g. `err == io.EOF`) are incorrect.  This PR seeks to migrate all such code to use the correct way of doing this in golang1.13.


Tryneus:
Added a new function `errors.As`, wrapping the underlying `errors.As` from `github.com/pkg/errors`, which wraps the base `errors.As` from golang1.13.  The new version will attempt multiple times to do the cast, adding and removing pointer-indirection if necessary.  This is due to an inconsistency in how `errors.As` performs the cast, which is rooted in golang pointer-receiver and non-pointer receiver structs both being usable as errors but with slightly different behavior.

This should be ready to go @brycemcanally

src/client/pkg/require/require.go
```diff
@@ -511,6 +511,7 @@ func logMessage(tb testing.TB, msgAndArgs []interface{}) {
 	if len(msgAndArgs) == 1 {
 		tb.Logf(msgAndArgs[0].(string))
 	}
+
```
brycemcanally, line 514: Was there a change here at some point, or did you just really feel that there needed to be a space here lol?

src/server/auth/server/testing/auth_test.go
```diff
@@ -1873,7 +1873,6 @@ func TestListDatum(t *testing.T) {
 		t.Skip("Skipping integration tests in short mode")
 	}
 	deleteAll(t)
-	defer deleteAll(t)
```
brycemcanally, line 1876: What was the reason for removing this?

src/client/pkg/errors/as.go
```diff
@@ -0,0 +1,46 @@
...
+	return false
+}
+
+// As finds the first error in err's chain that matches the target's type, and
+// if so, sets target to that error value and returns true.
+// As is a wrapper for the underlying errors.As function, which may panic or
+// return unexpected results based on how err was constructed (with or without
+// a pointer).  This works by inspecting the type of target and attempting
+// multiple errors.As calls if necessary.
+func As(err error, target interface{}) bool {
```
brycemcanally, line 25: I'm not really understanding some of the decisions here. Could you explain a bit more why non-pointer concrete error types are just going to be false here (I see that the Go implementation of `As` panics, but this does not seem to handle the non-pointer concrete error type either outside of avoiding the panic), and why we seem to support nested pointers and use nested pointers for a lot, or maybe all, of the `As` checks?

src/server/auth/server/testing/auth_test.go
Tryneus: Ah, yes, the test was failing but since it cleaned up after itself, I couldn't inspect the cluster state.  I see the problem here, though, which is that the auth credentials can't be accessed outside of the test and then you can't clean up your own cluster.  I'll revert this.

src/client/pkg/errors/as.go
Tryneus: You can observe the behavior yourself by running the `src/client/pkg/errors/testing` tests against the built-in `errors.As` - there will be several panics and failed conversions due to how go converts structs to the error interface.  Here is the diff if the tests were to pass against the base `errors` package:

```diff
diff --git a/src/client/pkg/errors/testing/errors_test.go b/src/client/pkg/errors/testing/errors_test.go
index 98f9fe393..19fac2143 100644
--- a/src/client/pkg/errors/testing/errors_test.go
+++ b/src/client/pkg/errors/testing/errors_test.go
@@ -1,10 +1,10 @@
 package testing

 import (
+  "errors"
   "fmt"
   "testing"

-  "github.com/pachyderm/pachyderm/src/client/pkg/errors"
   "github.com/pachyderm/pachyderm/src/client/pkg/require"
 )

@@ -80,18 +80,16 @@ func TestAsFooErr(t *testing.T) {
   require.Equal(t, fooerr, &FooErr{2})

   err = FooErr{3}
-  require.True(t, errors.As(err, &fooerr))
+  require.False(t, errors.As(err, &fooerr))
   require.False(t, errors.As(err, &otherErr))
-  require.Equal(t, fooerr, &FooErr{3})

   err = &FooErr{4}
-  require.True(t, errors.As(err, &FooErr{}))
+  require.False(t, errors.As(err, &FooErr{}))
   require.False(t, errors.As(err, &OtherErr{}))

   err = &FooErr{5}
-  require.True(t, errors.As(err, fooerr))
+  require.False(t, errors.As(err, fooerr))
   require.False(t, errors.As(err, otherErr))
-  require.Equal(t, fooerr, &FooErr{5})

   err = &FooErr{6}
   require.True(t, errors.As(err, &fooerr))
@@ -120,13 +118,12 @@ func TestAsPtrFooErr(t *testing.T) {
   // require.Equal(t, fooptrerr, &PtrFooErr{3})

   require.True(t, errors.As(err, &fooerr))
@@ -120,13 +118,12 @@ func TestAsPtrFooErr(t *testing.T) {
   // require.Equal(t, fooptrerr, &PtrFooErr{3})

   err = &PtrFooErr{4}
-  require.True(t, errors.As(err, &PtrFooErr{}))
+  // require.True(t, errors.As(err, &PtrFooErr{})) // does not compile
   require.False(t, errors.As(err, &OtherErr{}))

   err = &PtrFooErr{5}
-  require.True(t, errors.As(err, fooptrerr))
+  // require.True(t, errors.As(err, fooptrerr)) // does not compile
   require.False(t, errors.As(err, otherErr))
-  require.Equal(t, fooptrerr, &PtrFooErr{5})

   err = &PtrFooErr{6}
   require.True(t, errors.As(err, &fooptrerr))
@@ -146,10 +143,9 @@ func TestAsConcreteIntErr(t *testing.T) {
   require.False(t, errors.As(err, &OtherErr{}))

   err = ConcreteIntErr{2}
-  require.True(t, errors.As(err, interr))
-  require.False(t, errors.As(err, interrFace))
+  // require.True(t, errors.As(err, interr)) // does not compile
+  // require.False(t, errors.As(err, interrFace)) // does not compile
   require.False(t, errors.As(err, otherErr))
-  require.Equal(t, &ConcreteIntErr{2}, interr)

   err = ConcreteIntErr{3}
   require.True(t, errors.As(err, &interr))
@@ -165,10 +161,9 @@ func TestAsConcreteIntErr(t *testing.T) {

   err = &ConcreteIntErr{5}
   interr = &ConcreteIntErr{}
-  require.True(t, errors.As(err, interr))
-  require.False(t, errors.As(err, interrFace))
+  // require.True(t, errors.As(err, interr)) // does not compile
+  // require.False(t, errors.As(err, interrFace)) // does not compile
   require.False(t, errors.As(err, otherErr))
-  require.Equal(t, &ConcreteIntErr{5}, interr)

   err = &ConcreteIntErr{6}
   require.True(t, errors.As(err, &interr))
@@ -205,10 +200,9 @@ func TestAsConcretePtrIntErr(t *testing.T) {
   require.False(t, errors.As(err, &OtherErr{}))

   err = &ConcretePtrIntErr{5}
-  require.True(t, errors.As(err, interr))
-  require.False(t, errors.As(err, interrFace))
+  // require.True(t, errors.As(err, interr)) // does not compile
+  // require.False(t, errors.As(err, interrFace)) // does not compile
   require.False(t, errors.As(err, otherErr))
-  require.Equal(t, &ConcretePtrIntErr{5}, interr)

   err = &ConcretePtrIntErr{6}
   require.True(t, errors.As(err, &interr))
```

Tryneus: The point is that when we are using `errors.As`, we shouldn't have to know how the error struct was defined, or how it was constructed - our new `errors.As` takes the guesswork out.

brycemcanally: Correct me if I am wrong, but a non-pointer `target` will not work correctly with this change and will not panic. It will work if you pass a pointer to the concrete type even if the receiver for the error interface functions are non-pointer type. If I am understanding correctly, it seems like we could silently get incorrect behavior when the target is non-pointer type. Also, I don't understand why we would need support for double pointer `targets`? Are there any errors that come from external packages that are double pointer concrete types?

Tryneus: Most examples of `errors.As` use double-pointer `targets`, so I'm pretty sure that's the intended behavior.  As for non-pointer targets, I'd be ok with moving that to a panic, although I thought we were biasing towards not panicking in our code.  Keep in mind that a lot of code paths using `errors.As` are not very well exercised, and it is very likely that we will only find out the code is wrong when it happens in someone's production cluster.

brycemcanally: Ahh, I misunderstood how the Golang `As` works. It is expecting a pointer to a type that implements the error interface, not a type that implements the error interface. This is why a non-pointer `target` is not allowed. If you have a non-pointer type that implements the error interface then you are supposed to pass in a pointer to it. If the type is a pointer type, then you end up passing a double pointer as the `target`. So, I think the Golang `As` should be correct for our use case?

Tryneus: The problem is that if the error type is implemented with a non-pointer-receiver method, some `errors.As` calls will return true, and some will return false, depending on if the error was created with `err := &ErrorType{}` or `err := ErrorType{}` due to how golang interface casting works.  In addition, the set of `errors.As` calls that work in this situation depends on if you pass `&ErrorType{}` or `&&ErrorType{}` as the `target` field.  There is no parameter you can use that covers all cases.

Tryneus:
Discussed this offline with @brycemcanally and got his approval with the change from `return false` to a panic.

PR #5022, opened 2020-06-18
Wrap 'go tool' cmd in'pachctl debug pprof' with --interactive

msteffen:
This means that users collecting profiles for us won't get an error if they don't have 'go' installed


msteffen:
Tests had passed—the only remaining checks were netlify-related, so I merged the PR

PR #5023, opened 2020-06-18
PFS-over-HTTP supports TLS

actgardner:
To supporting downloading files from the dash in Hub we should support terminating TLS in pachd for PFS-over-HTTPS. We'll need to make a corresponding change in the dash to generate appropriate links (always TLS?).

We should also probably do this for the Githook server but that seems like it's going to have a lot of knock-on effects.

### DIFF
--- a/src/server/cmd/pachd/main.go
+++ b/src/server/cmd/pachd/main.go
@@ -684,7 +684,26 @@ func doFullMode(config interface{}) (retErr error) {
 		if err != nil {
 			return err
 		}
-		return http.ListenAndServe(fmt.Sprintf(":%v", env.HTTPPort), httpServer)
+		server := http.Server{
+			Addr:    fmt.Sprintf(":%v", env.HTTPPort),
+			Handler: httpServer,
+		}
+
+		certPath, keyPath, err := tls.GetCertPaths()
+		if err != nil {
+			log.Warnf("pfs-over-HTTP - TLS disabled: %v", err)
+			return server.ListenAndServe()
+		}
+
+		cLoader := tls.NewCertLoader(certPath, keyPath, tls.CertCheckFrequency)
+		err = cLoader.LoadAndStart()
+		if err != nil {
+			return errors.Wrapf(err, "couldn't load TLS cert for pfs-over-http: %v", err)
+		}
+
+		server.TLSConfig = &gotls.Config{GetCertificate: cLoader.GetCertificate}
+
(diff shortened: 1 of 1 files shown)


src/server/cmd/pachd/main.go
```diff
@@ -684,7 +684,26 @@ func doFullMode(config interface{}) (retErr error) {
...
 		}
-		return http.ListenAndServe(fmt.Sprintf(":%v", env.HTTPPort), httpServer)
+		server := http.Server{
+			Addr:    fmt.Sprintf(":%v", env.HTTPPort),
+			Handler: httpServer,
+		}
+
+		certPath, keyPath, err := tls.GetCertPaths()
+		if err != nil {
+			log.Warnf("pfs-over-HTTP - TLS disabled: %v", err)
```
ysimonson, line 694:
```suggestion
			log.Warnf("pfs-over-http TLS disabled: %v", err)
```

nit, slightly adjusting the error message:

1) use consistent casing for pfs and http (either both uppercase or both lowercase.)
2) remove the hyphen to be aligned with the [s3g warning](https://github.com/pachyderm/pachyderm/blob/7fc8aa9b56bab2fc7e91d9713bd9b35f6f3c5d1d/src/server/cmd/pachd/main.go#L701)

nitinjainsj:
@actgardner Is this ready to merge?

actgardner:
We'll need to coordinate the dash change, I'm not sure what the process is? If we merge this as-is it's just going to break the existing dash.

nitinjainsj:
@Connor-Wallace To see what change might be needed here.

actgardner:
Talked to JK, this is only used with port-forwarding currently. The options would be:
- make the dash download link configurable with an env var
- proxy the file download through the dash
- tell people to just use pachctl

nitinjainsj:
> Talked to JK, this is only used with port-forwarding currently. The options would be:
> 
> * make the dash download link configurable with an env var
> * proxy the file download through the dash
> * tell people to just use pachctl

I'd say 1. It gives us what we need for hub and also does not require changes to any existing users.

JoeyZwicker:
Very few users use the dash for download to 3 would be fine too

Joey Zwicker
Founder, Pachyderm

nitinjainsj:
@actgardner Another path forward. We make the change in master after 1.11.0 is shipped and also make the corresponding change in dash. We will have to release a hub specific pach release with 1.0 GA.

PR #5033, opened 2020-06-19
Storage V2 Copy File

brendoncarroll:
- Adds validation server for validating requests and checking permissions
- Adds CopyFile implementation for storage v2, the signature is that same as v1.
- Adds InspectFile
- Adds WalkFile

### DIFF
--- a/src/server/pfs/server/api_server_v2.go
+++ b/src/server/pfs/server/api_server_v2.go
@@ -245,6 +245,16 @@ func (a *apiServerV2) GlobFileV2(request *pfs.GlobFileRequest, server pfs.API_Gl
 	})
 }
 
+// CopyFile implements the protobuf pfs.CopyFile RPC
+func (a *apiServerV2) CopyFile(ctx context.Context, request *pfs.CopyFileRequest) (response *types.Empty, retErr error) {
+	func() { a.Log(request, nil, nil, 0) }()
+	defer func(start time.Time) { a.Log(request, response, retErr, time.Since(start)) }(time.Now())
+	if err := a.driver.copyFile(a.env.GetPachClient(ctx), request.Src, request.Dst, request.Overwrite); err != nil {
+		return nil, err
+	}
+	return &types.Empty{}, nil
+}
+
 // InspectFileV2 returns info about a file.
 func (a *apiServerV2) InspectFileV2(ctx context.Context, req *pfs.InspectFileRequest) (*pfs.FileInfoV2, error) {
 	return a.driver.inspectFile(a.env.GetPachClient(ctx), req.File)
--- a/src/server/pfs/server/driver.go
+++ b/src/server/pfs/server/driver.go
@@ -2836,14 +2836,6 @@ func (d *driver) scratchCommitPrefix(commit *pfs.Commit) string {
 	return path.Join(commit.Repo.Name, commit.ID)
 }
 
-func checkFilePath(path string) error {
-	path = filepath.Clean(path)
-	if strings.HasPrefix(path, "../") {
-		return errors.Errorf("path (%s) invalid: traverses above root", path)
-	}
-	return nil
-}
-
 // scratchFilePrefix returns an etcd prefix that's used to temporarily
 // store the state of a file in an open commit.  Once the commit is finished,
 // the scratch space is removed.
--- a/src/server/pfs/server/driver_v2.go
+++ b/src/server/pfs/server/driver_v2.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"log"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -21,6 +22,7 @@ import (
 	"github.com/pachyderm/pachyderm/src/server/pkg/storage/chunk"
 	"github.com/pachyderm/pachyderm/src/server/pkg/storage/fileset"
 	"github.com/pachyderm/pachyderm/src/server/pkg/storage/fileset/index"
+	"github.com/pachyderm/pachyderm/src/server/pkg/storage/fileset/tar"
 	"github.com/pachyderm/pachyderm/src/server/pkg/storage/gc"
 	txnenv "github.com/pachyderm/pachyderm/src/server/pkg/transactionenv"
 	"github.com/pachyderm/pachyderm/src/server/pkg/uuid"
@@ -165,6 +167,28 @@ func (d *driverV2) withFileSet(ctx context.Context, repo, commit string, cb func
 		return err
 	}
 	if err := fs.Close(); err != nil {
+		return err
+	}
+	return d.compactionQueue.RunTaskBlock(ctx, func(m *work.Master) error {
+		_, err := d.compact(m, subFileSetPath, []string{path.Join(tmpPrefix, subFileSetPath)})
+		return err
+	})
+}
+
+func (d *driverV2) withWriter(ctx context.Context, commit *pfs.Commit, cb func(int64, *fileset.Writer) error) (retErr error) {
+	n := d.getSubFileSet()
+	subFileSetStr := fileset.SubFileSetStr(n)
+	subFileSetPath := path.Join(commit.Repo.Name, commit.ID, subFileSetStr)
+	fsw := d.storage.NewWriter(ctx, path.Join(tmpPrefix, subFileSetPath))
+	defer func() {
+		if err := d.storage.Delete(ctx, path.Join(tmpPrefix, subFileSetPath)); retErr == nil {
+			retErr = err
+		}
+	}()
+	if err := cb(n, fsw); err != nil {
+		return err
+	}
+	if err := fsw.Close(); err != nil {
 		return err
 	}
 	return d.compactionQueue.RunTaskBlock(ctx, func(m *work.Master) error {
@@ -470,6 +494,49 @@ func (d *driverV2) globFileV2(pachClient *client.APIClient, commit *pfs.Commit, 
 	})
 	return s.Iterate(ctx, func(fi *pfs.FileInfoV2, f fileset.File) error {
 		return cb(fi)
+	})
+}
+
+func (d *driverV2) copyFile(pachClient *client.APIClient, src *pfs.File, dst *pfs.File, overwrite bool) (retErr error) {
+	ctx := pachClient.Ctx()
+	_, err := d.inspectCommit(pachClient, src.Commit, pfs.CommitState_FINISHED)
+	if err != nil {
+		return err
+	}
+	if overwrite {
+		// TODO: after delete merging is sorted out add overwrite support
+		return errors.New("overwrite not yet supported")
+	}
+	srcPath := cleanPath(src.Path)
+	dstPath := cleanPath(dst.Path)
+	pathTransform := func(x string) string {
+		relPath, err := filepath.Rel(srcPath, x)
+		if err != nil {
+			panic("cannot apply path transform")
+		}
+		return path.Join(dstPath, relPath)
+	}
+	s := d.storage.NewSource(ctx, compactedCommitPath(src.Commit), index.WithPrefix(srcPath))
+	s = fileset.NewIndexFilter(s, func(idx *index.Index) bool {
+		return idx.Path == srcPath || strings.HasPrefix(idx.Path, srcPath+"/")
+	})
+	s = fileset.NewHeaderMapper(s, func(th *tar.Header) *tar.Header {
+		th.Name = pathTransform(th.Name)
+		return th
+	})
+	s = fileset.NewDirInserter(s)
+	return d.withWriter(ctx, dst.Commit, func(n int64, dst *fileset.Writer) error {
+		return s.Iterate(ctx, func(f fileset.File) error {
+			hdr, err := f.Header()
+			if err != nil {
+				return err
+			}
+			if err := dst.WriteHeader(hdr); err != nil {
+				return err
+			}
+			dst.Tag(fileset.SubFileSetStr(n))
+			return f.Content(dst)
+		})
 	})
 }
 
--- a/src/server/pfs/server/driver_v2.go
+++ b/src/server/pfs/server/driver_v2.go
@@ -191,10 +191,10 @@ func (d *driverV2) withWriter(ctx context.Context, commit *pfs.Commit, cb func(i
 	if err := fsw.Close(); err != nil {
 		return err
 	}
-	return d.compactionQueue.RunTaskBlock(ctx, func(m *work.Master) error {
-		_, err := d.compact(m, subFileSetPath, []string{path.Join(tmpPrefix, subFileSetPath)})
-		return err
-	})
+	// There is no need to queue this because we just wrote to one place.  We expect the storage layer
+	// to handle the single file case efficiently.
+	_, err := d.storage.Compact(ctx, subFileSetPath, []string{path.Join(tmpPrefix, subFileSetPath)})
+	return err
 }
 
 func (d *driverV2) getTar(ctx context.Context, commit *pfs.Commit, glob string, w io.Writer) error {
--- a/src/server/pfs/server/testing/server_v2_test.go
+++ b/src/server/pfs/server/testing/server_v2_test.go
@@ -644,7 +644,6 @@ func (v *validator) validate(r io.Reader) error {
 	for _, file := range filesSorted {
 		buf := &bytes.Buffer{}
 		if _, err := io.CopyN(buf, r, int64(len(v.files[file]))); err != nil {
-			fmt.Println("missing", file)
 			return err
 		}
 		if !bytes.Equal(v.files[file], buf.Bytes()) {
--- a/src/server/pfs/server/testing/server_v2_test.go
+++ b/src/server/pfs/server/testing/server_v2_test.go
@@ -1037,6 +1037,13 @@ func TestCopyFile2(t *testing.T) {
 			})
 			require.NoError(t, err)
 		}
+		_, err = env.PachClient.InspectFileV2(ctx, &pfs.InspectFileRequest{
+			File: &pfs.File{
+				Commit: otherCommit,
+				Path:   "file0",
+			},
+		})
+		require.NoError(t, err)
 		return nil
 	}, conf)
 	require.NoError(t, err)
--- a/src/server/pfs/server/testing/server_v2_test.go
+++ b/src/server/pfs/server/testing/server_v2_test.go
@@ -644,6 +644,7 @@ func (v *validator) validate(r io.Reader) error {
 	for _, file := range filesSorted {
 		buf := &bytes.Buffer{}
 		if _, err := io.CopyN(buf, r, int64(len(v.files[file]))); err != nil {
+			fmt.Println("missing", file)
 			return err
 		}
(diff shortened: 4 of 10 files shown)


src/server/pfs/server/api_server_v2.go
```diff
@@ -221,7 +226,31 @@ func (a *apiServerV2) FinishCommitInTransaction(
 	txnCtx *txnenv.TransactionContext,
 	request *pfs.FinishCommitRequest,
 ) error {
+	fmt.Println("called finish commit")
```
brycemcanally, line 229: remove

src/client/pfs/pfs.proto
```diff
@@ -631,6 +631,10 @@ service API {
   // Refer to the GetTarConditionalRequest / GetTarConditionalResponse message definitions for the protocol.
   rpc GetTarConditionalV2(stream GetTarConditionalRequestV2) returns (stream GetTarConditionalResponseV2) {}
   rpc ListFileV2(ListFileRequest) returns (stream FileInfoV2) {}
+  // InspectFileV2 returns info about a file.
```
brycemcanally, line 634: Where is copy file?

src/server/pfs/server/driver_v2.go
```diff
@@ -513,3 +515,55 @@ func (d *driverV2) compactionWorker() {
...
 	panic(err)
 }
+
+func (d *driverV2) copyFile(pachClient *client.APIClient, src *pfs.File, dst *pfs.File, overwrite bool) (retErr error) {
+	if overwrite {
+		// TODO: after delete merging is sorted out add overwrite support
+		return errors.New("overwrite not yet supported")
+	}
+	return d.withFileSet(pachClient.Ctx(), dst.Commit.Repo.Name, dst.Commit.ID, func(dstFs *fileset.FileSet) error {
+		return d.getTarConditional(pachClient.Ctx(), src.Commit.Repo.Name, src.Commit.ID, src.Path, func(rf *FileReader) error {
```
brycemcanally, line 525: This is not the way we want to do this. This does not take advantage of the cheap copying. This should be doing essentially what the storage compact function does (setup a merge reader from the compaction levels and writing it to an output fileset). I think we might want support for filtering based on globs, but a simple single file copy is probably fine for the time being. So, this could just make a compact call on the src compaction levels with the index option being the src file path. I think you might hit some trickiness with handling the renaming since the content stream has file metadata (the tar header). This will probably require a bit more thought, but I think the basic approach would be to read the tar header, apply the change, then cheap copy the content.

src/server/pfs/server/driver_v2.go
```diff
@@ -513,3 +515,55 @@ func (d *driverV2) compactionWorker() {
...
+
+func (d *driverV2) inspectFile(pachClient *client.APIClient, file *pfs.File) (fi *pfs.FileInfoV2, retErr error) {
+	commitInfo, err := d.inspectCommit(pachClient, file.Commit, pfs.CommitState_STARTED)
+	if err != nil {
+		return nil, err
+	}
+	commit := commitInfo.Commit
+	stopIter := errors.New("stop iteration")
+	path := "/" + strings.Trim(file.Path, "/")
+	err = d.getTarConditional(pachClient.Ctx(), commit.Repo.Name, commit.ID, path, func(fr *FileReader) error {
```
brycemcanally, line 547: It might make sense to just have a full path match option, rather than this stop iteration stuff.

src/server/pfs/server/driver_v2.go
```diff
@@ -513,3 +515,55 @@ func (d *driverV2) compactionWorker() {
...
+	return fi, nil
+}
+
+func (d *driverV2) walkFile(pachClient *client.APIClient, file *pfs.File, f func(*pfs.FileInfoV2) error) (retErr error) {
+	commitInfo, err := d.inspectCommit(pachClient, file.Commit, pfs.CommitState_STARTED)
+	if err != nil {
+		return err
+	}
+	commit := commitInfo.Commit
+	return d.getTarConditional(pachClient.Ctx(), commit.Repo.Name, commit.ID, file.Path+"/*", func(fr *FileReader) error {
```
brycemcanally, line 566: You should be using `path.Join` rather than string concatenation. Also, I think you will be wanting to use `/**` to match multiple levels.

src/server/pfs/server/driver_v2.go
```diff
@@ -513,3 +515,55 @@ func (d *driverV2) compactionWorker() {
...
+		return nil, err
+	}
+	if fi == nil {
+		return nil, pfsserver.ErrFileNotFound{File: file}
+	}
+	return fi, nil
+}
+
+func (d *driverV2) walkFile(pachClient *client.APIClient, file *pfs.File, f func(*pfs.FileInfoV2) error) (retErr error) {
+	commitInfo, err := d.inspectCommit(pachClient, file.Commit, pfs.CommitState_STARTED)
```
brycemcanally, line 561: Probably should be checking finished commit. In general, we should be validating that commits are finished before attempting to read.

src/server/pfs/server/driver_v2.go
brendoncarroll: Maybe we should return an error if the user passes a glob? Passing a single path without `*` to getTarConditional should only match one path.  Maybe have a separate getTarSingle function which wraps getTarConditional and panics if the callback is called more than once.

brendoncarroll: Looking at this some more. I think the right thing to do is error if the path/glob is ambiguous.

src/server/pfs/server/api_server.go
```diff
@@ -623,6 +623,14 @@ func (a *apiServer) GlobFileV2(req *pfs.GlobFileRequest, server pfs.API_GlobFile
 	return errV2NotImplemented
 }
 
+func (a *apiServer) InspectFileV2(ctx context.Context, req *pfs.InspectFileRequest) (*pfs.FileInfoV2, error) {
+	return nil, errors.Errorf("v2 method not implemented")
```
brycemcanally, line 627: `InspectFileV2` and `WalkFileV2` should be using the error variable you setup (`errV2NotImplemented`).

src/server/pfs/server/driver.go
```diff
@@ -2831,6 +2831,10 @@ func (d *driver) scratchCommitPrefix(commit *pfs.Commit) string {
 }
 
 func (d *driver) checkFilePath(path string) error {
+	return checkFilePath(path)
```
brycemcanally, line 2834: Why is this attached to the driver?

src/server/pfs/server/driver_v2.go
```diff
@@ -144,14 +145,17 @@ func (d *driverV2) finishCommitV2(txnCtx *txnenv.TransactionContext, commit *pfs
 func (d *driverV2) withFileSet(ctx context.Context, repo, commit string, f func(*fileset.FileSet) error) (retErr error) {
 	// TODO subFileSet will need to be incremented through postgres or etcd.
```
brycemcanally, line 146: It would probably be a good idea to refactor this into a helper function. The function would lock / defer unlock, then read and increment subfileset. It would return the subfileset value.

src/server/pfs/server/driver_v2.go
```diff
@@ -168,6 +172,32 @@ func (d *driverV2) withFileSet(ctx context.Context, repo, commit string, f func(
 	})
 }
 
+func (d *driverV2) withWriter(ctx context.Context, repo, commit string, f func(int64, *fileset.Writer) error) (retErr error) {
+	// TODO subFileSet will need to be incremented through postgres or etcd.
```
brycemcanally, line 176: This would use the helper function mentioned above.

src/server/pfs/server/driver_v2.go
```diff
@@ -168,6 +172,32 @@ func (d *driverV2) withFileSet(ctx context.Context, repo, commit string, f func(
...
+
+	subFileSetStr := fileset.SubFileSetStr(n)
+	subFileSetPath := path.Join(repo, commit, subFileSetStr)
+	fsw := d.storage.NewWriter(ctx, path.Join(tmpPrefix, subFileSetPath))
+	defer func() {
+		if err := d.storage.Delete(ctx, path.Join(tmpPrefix, subFileSetPath)); retErr == nil {
+			retErr = err
+		}
+	}()
+	if err := f(n, fsw); err != nil {
```
brycemcanally, line 190: `f` -> `cb`. Also, if you see any callback code that uses the name `f`, go ahead and move it over to `cb`.

src/server/pfs/server/driver_v2.go
```diff
@@ -258,6 +289,20 @@ func (d *driverV2) getTarConditional(ctx context.Context, repo, commit, glob str
 	return nil
 }
 
+func (d *driverV2) getFile(ctx context.Context, repoName, commitID, p string, f func(fr *FileReader) error) error {
```
brycemcanally, line 292: This should be removed right?

src/server/pfs/server/driver_v2.go
```diff
@@ -536,3 +581,81 @@ func (d *driverV2) globFileV2(pachClient *client.APIClient, commit *pfs.Commit,
...
+	if err != nil {
+		return err
+	}
+	return d.withWriter(ctx, dst.Commit.Repo.Name, dst.Commit.ID, func(n int64, dstFs *fileset.Writer) error {
+		return mr.Iterate(func(mfr *fileset.FileMergeReader) error {
+			h, err := mfr.Header()
+			if err != nil {
+				return err
+			}
+			h.Name = cleanPath(h.Name)
```
brycemcanally, line 610: This shouldn't be necessary right?

src/server/pfs/server/driver_v2.go
```diff
@@ -536,3 +581,81 @@ func (d *driverV2) globFileV2(pachClient *client.APIClient, commit *pfs.Commit,
...
+	srcPrefix := cleanPath(src.Path)
+	dstPrefix := cleanPath(dst.Path)
+	pathTransform := func(x string) string {
+		if !strings.HasPrefix(x, srcPrefix) {
+			panic("cannot apply path transform")
+		}
+		return path.Join(dstPrefix, x[len(srcPrefix):])
+	}
+
+	mr, err := d.storage.NewMergeReader(ctx, []string{commitKey(src.Commit)})
```
brycemcanally, line 600: This is using a helper function from v1 for creating the commit prefix. It is correct because the etcd prefix in v1 is the same as the object storage prefix in v2, but I think it would be better if we were to use this in the other places where we compute a commit prefix for consistency.

src/server/pfs/server/driver_v2.go
```diff
@@ -536,3 +581,81 @@ func (d *driverV2) globFileV2(pachClient *client.APIClient, commit *pfs.Commit,
 		return f(fr.Info())
 	})
 }
+
+func (d *driverV2) copyFile(pachClient *client.APIClient, src *pfs.File, dst *pfs.File, overwrite bool) (retErr error) {
```
brycemcanally, line 585: We need some validation for checking if the dst commit is finished.

src/server/pfs/server/driver_v2.go
```diff
@@ -536,3 +581,81 @@ func (d *driverV2) globFileV2(pachClient *client.APIClient, commit *pfs.Commit,
...
+	ctx := pachClient.Ctx()
+	if overwrite {
+		// TODO: after delete merging is sorted out add overwrite support
+		return errors.New("overwrite not yet supported")
+	}
+	srcPrefix := cleanPath(src.Path)
+	dstPrefix := cleanPath(dst.Path)
+	pathTransform := func(x string) string {
+		if !strings.HasPrefix(x, srcPrefix) {
```
brycemcanally, line 594: I think this would probably be cleaner if it used a relative path function (such as `filepath.Rel`) rather than checking a prefix then appending everything after it. Since we don't support Windows, I think this might be the better approach from a readability standpoint.

src/server/pfs/server/driver_v2.go
```diff
@@ -536,3 +581,81 @@ func (d *driverV2) globFileV2(pachClient *client.APIClient, commit *pfs.Commit,
...
+	})
+}
+
+func (d *driverV2) inspectFile(pachClient *client.APIClient, file *pfs.File) (fi *pfs.FileInfoV2, retErr error) {
+	commitInfo, err := d.inspectCommit(pachClient, file.Commit, pfs.CommitState_FINISHED)
+	if err != nil {
+		return nil, err
+	}
+	commit := commitInfo.Commit
+	called := false
```
brycemcanally, line 629: I think I mentioned this before, but I would prefer the functionality of this called check to be handled by an exact match functional option on the indexing layer. Also, no globs for inspect file.

src/server/pfs/server/testing/server_v2_test.go
```diff
@@ -619,31 +624,84 @@ func (v *validator) recordFileSet(files fileSetSpec) {
...
+	// 		require.NoError(err)
+	// 	}
+	// 	// if !bytes.Equal(v.files[file], buf.Bytes()) {
+	// 	// 	fmt.Println("VALIDATE", len(v.files[file]), buf.Len())
+	// 	// 	return errors.Errorf("file %v's header and/or content is incorrect", file)
+	// 	// }
+	// }
+}
+
+func diffTarStreams(t *testing.T, a, b io.Reader) error {
```
brycemcanally, line 657: This doesn't seem to check the content. I think the approach we want to take is what originally existed, which basically just compares the serialized content. We also will want to add some configuration for the larger workloads to check hashes rather than storing and checking actual content.

src/server/pkg/storage/fileset/storage.go
```diff
@@ -178,52 +198,44 @@ func (s *Storage) CompactSpec(ctx context.Context, fileSet string, compactedFile
...
 	}
-	// Handle commits with a parent commit.
-	compactedFileSet[0] = applyPrefix(compactedFileSet[0])
-	if err := s.objC.Walk(ctx, path.Join(compactedFileSet[0], Compacted), func(name string) error {
-		nextLevel, err := strconv.Atoi(path.Base(name))
+	// While we can't fit it all in the current level
+	for {
+		levelPath := path.Join(compactedFileSet[0], Compacted, levelName(level))
+		idx, err := index.GetTopLevelIndex(ctx, s.objC, levelPath)
+		if err != nil && !s.objC.IsNotExist(err) {
```
brycemcanally, line 211: Include the slight refactor I mentioned (not exist error checked within an err != nil block, which means subsequent code is implicitly for a nil error).

src/server/pkg/storage/fileset/storage.go
```diff
@@ -178,52 +198,44 @@ func (s *Storage) CompactSpec(ctx context.Context, fileSet string, compactedFile
...
+		lName := path.Base(src)
+		l, err := parseLevel(lName)
 		if err != nil {
 			return err
 		}
-		// Handle levels that are non-empty.
-		if nextLevel == level {
-			idx, err := index.GetTopLevelIndex(ctx, s.objC, name)
-			if err != nil {
+		dst := path.Join(fileSet, Compacted, levelName(l))
```
brycemcanally, line 233: This should probably be setup in the below if block (it would be unused otherwise).

src/server/pkg/obj/local_client.go
```diff
@@ -135,3 +138,31 @@ func newSectionReadCloser(f *os.File, offset uint64, size uint64) *sectionReadCl
 func (s *sectionReadCloser) Close() error {
 	return errors.EnsureStack(s.f.Close())
 }
+
+// type localWriter struct {
```
brendoncarroll, line 142: Need to remove this

src/server/pfs/server/driver_v2.go
brendoncarroll: We would have to rewrite the getTarConditional functionality on top of a MergeReader with that index option. I can make that change as part of the FilterHeader Reader I am working on.

src/server/pfs/server/driver_v2.go
```diff
@@ -173,6 +175,28 @@ func (d *driverV2) withFileSet(ctx context.Context, repo, commit string, cb func
...
+			retErr = err
+		}
+	}()
+	if err := cb(n, fsw); err != nil {
+		return err
+	}
+	if err := fsw.Close(); err != nil {
+		return err
+	}
+	return d.compactionQueue.RunTaskBlock(ctx, func(m *work.Master) error {
```
brycemcanally, line 194: I don't think this needs to be compacted. I think you could just do a copy object. Also, we should probably add special casing for a compaction that only involves one subfileset to just copy the object.

src/server/pkg/storage/fileset/writer.go
```diff
@@ -222,3 +224,16 @@ func (w *Writer) Close() error {
 	// Close the index writer.
 	return w.iw.Close()
 }
+
+func (w *Writer) checkPath(p string) error {
```
brycemcanally, line 228: :+1:

src/server/pkg/storage/fileset/dir_inserter.go
```diff
@@ -0,0 +1,79 @@
...
+
+type dirInserter struct {
+	x FileSource
+}
+
+func NewDirInserter(x FileSource) FileSource {
+	return &dirInserter{x: x}
+}
+
+func (s *dirInserter) Iterate(ctx context.Context, cb func(File) error, stopBefore ...string) error {
```
brycemcanally, line 21: This seems fine for the time being, but I think we may want this functionality to be on the write side. Maybe a wrapper for the fileset writer. This would make it so that both the copy file process and the serialization of the in memory filesets can have one implementation that handles creating directories implicitly.

brendoncarroll: We limit the usefulness of a transformation if it is on the write side.  Writing goes to somewhere we can't get at.  Transformations on the read side we can reuse.  The best example right now is GetTar. If we want to return directory entries in that, we would need to wrap this around the `FileSource` that is currently returned.

src/server/pfs/server/driver_v2.go
brendoncarroll: I changed this to a call to `d.storage.Compact` which we can change later to handle that case well.

src/server/pkg/storage/fileset/dir_inserter.go
brendoncarroll: Thinking more about this.  If the "UnorderedWriter" was able to wrap its internal map with something that looked like a `FileSource` we could read from it and apply the directory inserter before writing it to the "OrderedWriter".

brendoncarroll: I think this is more or less what we want
```go
package fileset

import "sort"

type Buffer map[string]File

func (b Buffer) Iterate(ctx context.Context, cb func(f File) error) error {
        keys := []string{}
        for k := range b {
                keys = append(keys, key)
        }
        sort.Strings(keys)
        for k := range keys {
                f = b[k]
                if err := cb(f); err != nil {
                        return err
                }
        }
        return nil
}
```

brycemcanally: This makes sense to me. The main thing is that I want that functionality to be captured in one place, and having it be a read abstraction could potentially be useful if we come up with another use case for inserting directories that is not doing so just to satisfy the invariant we have on ordered writes.

PR #5038, opened 2020-06-22
Fixes CI for outside contributors

### LINKED ISSUE #5037: CI: Breakages with outside contributors
**What happened?**: Contributors who are not members of the pachyderm org will presently see a couple of CI issues when opening a PR. Both are triggered by not having access to secret environment variables:

1. We push images to docker hub on the build phase.
2. Auth tests are run.

ysimonson:
Changes:

* Disable the build step from outside PRs, as we won't be able to push the docker images anyway. Instead, we build in each step, like the way we did it before #5007.
* Added a check in go test code to skip tests that require enterprise mode when secrets aren't available. This allows us to selectively skip tests rather than skip entire buckets.
* Disable `test-tls` on outside PRs since it needs secrets.
* Explicitly enable enterprise mode in `TestActivateAsRobotUser` and `TestActivateMismatchedUsernames`, as before they implicitly assumed that other tests that were run before had already enabled enterprise.

As a test, this PR is opened from my fork of pachyderm, which should reproduce any issues that exist.

Closes #5037

### DIFF
--- a/src/server/worker/stats/stats_test.go
+++ b/src/server/worker/stats/stats_test.go
@@ -24,9 +24,9 @@ import (
 	prom_model "github.com/prometheus/common/model"
 )
 
-func activateEnterprise(c *client.APIClient) error {
+func activateEnterprise(t testing.TB, c *client.APIClient) error {
 	_, err := c.Enterprise.Activate(context.Background(),
-		&enterprise.ActivateRequest{ActivationCode: tu.GetTestEnterpriseCode()})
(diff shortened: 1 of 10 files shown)


src/server/worker/stats/stats_test.go
```diff
@@ -24,9 +24,9 @@ import (
 	prom_model "github.com/prometheus/common/model"
 )
 
-func activateEnterprise(c *client.APIClient) error {
```
Tryneus, line 27: Nit: this looks basically identical to the new code in `src/server/auth/server/testing/admin_test.go` - it might be worth moving this into a common area and reusing it.

PR #5062, opened 2020-07-01
Deprecation notice for S3V2 flag (minio client)

nitinjainsj:


src/server/pkg/deploy/cmds/cmds.go
```diff
@@ -629,6 +629,9 @@ If <object store backend> is \"s3\", then the arguments are:
 		"(required) Backend providing an object-storage API to pachyderm. One of: "+
 			"s3, gcs, or azure-blob.")
 	deployCustom.Flags().BoolVar(&isS3V2, "isS3V2", false, "Enable S3V2 client")
```
Tryneus, line 631: Should probably indicate that the flag is deprecated in the help as well
```suggestion
	deployCustom.Flags().BoolVar(&isS3V2, "isS3V2", false, "Enable S3V2 client (DEPRECATED)")
```

src/server/pkg/deploy/cmds/cmds.go
```diff
@@ -629,6 +629,9 @@ If <object store backend> is \"s3\", then the arguments are:
...
+	if isS3V2 {
```
Tryneus, line 632: So this check has to go up in the `deployCustom` `Run` callback above - the code at this level is run every time `pachctl` runs, but before arg-parsing, so this variable will always be `false`.  Just cutting-pasting this to the start of the `Run` callback should be good.

src/server/pkg/deploy/cmds/cmds.go
```diff
@@ -629,6 +629,9 @@ If <object store backend> is \"s3\", then the arguments are:
...
+		fmt.Printf("DEPRECATED: The use of S3V2 option is being deprecated and will be removed in future version\n\n")
```
Tryneus, line 633: wording nit
```suggestion
		fmt.Printf("DEPRECATED: The use of the S3V2 option is being deprecated and will be removed in a future version\n\n")
```

nitinjainsj: Changed this to be the same as the warning string.

src/server/pkg/deploy/cmds/cmds.go
nitinjainsj: Interesting. I thought this was under the standardDeployCmds, so it would only be run when someone uses the deploy command. But I guess any time sone types `pachctl deploy --help` this will also run to display all the help.

Fixed it.

PR #5063, opened 2020-07-01
Fix dash compatibility check

nitinjainsj:


src/server/pkg/deploy/cmds/cmds.go
```diff
@@ -1319,7 +1319,7 @@ func getDefaultOrLatestDashImage(dashImage string, dryRun bool) string {
 		return dashImage
 	}
 	dashImage = defaultDashImage
-	compatibleDashVersionsURL := fmt.Sprintf("https://raw.githubusercontent.com/pachyderm/pachyderm/master/etc/compatibility/%v", version)
```
Tryneus, line 1322: Not that this is up for review yet, but I don't think we typically generate dash compatibility files for custom releases, so the first `version` arg should be the version without `VersionAdditional` if (and only if) `VersionAdditional` looks like a git commit hash.

doc/release_instructions.md
```diff
@@ -79,20 +94,20 @@ nothing added to commit but untracked files present (use "git add" to track)
...
 
-9) Commit the changes:
+10) Commit the changes:
 
 ```
 > git commit -a -m"Regenerate golden deployment manifests for $(pachctl version --client-only)"
 > git push origin master
 ```
 
-10) Update the
```
Tryneus, line 91: If we're reconsidering the order of these steps - I'd propose we move the changelog up above the `make point-release` step.  Obviously we can't edit the release notes, but it's rather nice to have the `CHANGELOG.md` contain the relevant data under the release's git tag.

doc/release_instructions.md
```diff
@@ -36,33 +39,45 @@ If you're doing a custom release (off a branch that isn't master), [skip to the
...
 2) Make sure that you have no uncommitted files in the current branch. Note that `make doc` (next step) will fail if there are any uncommitted changes in the current branch
 
-3) Update `src/client/version/client.go` version values, build a new local version of pachctl, and **commit the change** (locally—you'll push it to GitHub in the next step, but this allows `make doc` to run):
-
+3) Update client version. Commit these changes locally. You will push to GitHub in the next step.
+    - [Required for Major, Minor, and Patch releases]
+      - `src/client/version/client.go` version values
+        - for a major release increment the MajorVersion and set the MinorVersion and MicroVersion to 0 --> eg. 2.0.0
+        - for a minor release leave the MajorVersion unchanged, increment the MinorVersion, and set the MicroVersion to 0 --> eg. 1.10.0
+        - for a patch release leave the MajorVersion and MinorVersion unchanged and increment the MicorVersion --> eg. 1.9.8
```
Tryneus, line 47: typo here
```suggestion
        - for a patch release, leave the MajorVersion and MinorVersion unchanged and increment the MicroVersion --> eg. 1.9.8
```

doc/release_instructions.md
```diff
@@ -36,33 +39,45 @@ If you're doing a custom release (off a branch that isn't master), [skip to the
 
 2) Make sure that you have no uncommitted files in the current branch. Note that `make doc` (next step) will fail if there are any uncommitted changes in the current branch
 
-3) Update `src/client/version/client.go` version values, build a new local version of pachctl, and **commit the change** (locally—you'll push it to GitHub in the next step, but this allows `make doc` to run):
-
+3) Update client version. Commit these changes locally. You will push to GitHub in the next step.
+    - [Required for Major, Minor, and Patch releases]
+      - `src/client/version/client.go` version values
+        - for a major release increment the MajorVersion and set the MinorVersion and MicroVersion to 0 --> eg. 2.0.0
```
Tryneus, line 45: punctuation nit
```suggestion
        - for a major release, increment the MajorVersion and set the MinorVersion and MicroVersion to 0 --> eg. 2.0.0
```

doc/release_instructions.md
```diff
@@ -36,33 +39,45 @@ If you're doing a custom release (off a branch that isn't master), [skip to the
...
+        - for a minor release leave the MajorVersion unchanged, increment the MinorVersion, and set the MicroVersion to 0 --> eg. 1.10.0
```
Tryneus, line 46: punctuation nit
```suggestion
        - for a minor release, leave the MajorVersion unchanged, increment the MinorVersion, and set the MicroVersion to 0 --> eg. 1.10.0
```

src/server/pkg/deploy/cmds/cmds.go
nitinjainsj: Technically we are creating the compatibility file so we should commit it. But the instructions don't call for it, so we have not been doing it for a long time.

I'll make this change.

nitinjainsj:
@msteffen I made changes to the release process. Most of it is moving all the commits we do as part of the release process before the git tag is created. You may want to take a quick pass at it as well.

etc/build/dash_compatibility.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+VERSION=$("$GOPATH/bin/pachctl" version --client-only)
```
msteffen, line 3: Nit: I think at one point we had an issue calling `$GOPATH/bin/pachctl` vs just `pachctl`—I don't remember which was the problematic one, but one wasn't getting updated for some engineers and was causing issues.

msteffen: Update, this is fine—`$GOPATH/bin/pachctl` was the preferred call path

PR #5067, opened 2020-07-01
Log half of OIDC state token

msteffen:
This allows k8s cluster admins to correlate login errors with pachd
logs, without giving them a full authenticator that they can exercise

### DIFF
--- a/src/server/auth/cmds/cmds.go
+++ b/src/server/auth/cmds/cmds.go
@@ -204,7 +204,9 @@ func LoginCmd() *cobra.Command {
 				if authErr != nil && !auth.IsErrPartiallyActivated(authErr) {
 					return errors.Wrapf(grpcutil.ScrubGRPC(authErr),
 						"authorization failed (OIDC state token: %q; Pachyderm logs may "+
-							"contain more information)", state)
+							"contain more information)",
+						// Print state token as it's logged, for easy searching
+						fmt.Sprintf("%s.../%d", state[:len(state)/2], len(state)))
 				}
 			} else {
 				// Exchange GitHub token for Pachyderm token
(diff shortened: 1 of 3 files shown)


src/server/auth/cmds/cmds.go
```diff
@@ -204,7 +204,9 @@ func LoginCmd() *cobra.Command {
 				if authErr != nil && !auth.IsErrPartiallyActivated(authErr) {
 					return errors.Wrapf(grpcutil.ScrubGRPC(authErr),
 						"authorization failed (OIDC state token: %q; Pachyderm logs may "+
-							"contain more information)", state)
+							"contain more information)",
+						// Print state token as it's logged, for easy searching
+						fmt.Sprintf("%s.../%d", state[:len(state)/2], len(state)))
```
adelelopez, line 209: use `half` here

msteffen: Unfortunately this is in a different package, so `half` isn't visible here. Some options that occur to me, though:
- I could move it into a shared package (e.g. `client/auth.go`) but TBH I don't love this idea because it'll make all the calls in `oidc.go` longer—they'll all be `client.Half(state)` instead of `half(state)`. Also the `client/auth` package will have a helper function in it that it itself never uses. (These are, I admit, purely code-aesthetics-based objections). I can go this route if you think it's better, though.
- I could create another function called `half` in `auth/cmds.go` and just add a comment that it's the same as the one in `oidc.go`
- I could just log the whole state token from `pachctl`. The string that users see won't match the string in the logs, but we're not revealing anything.

Let me know what you think makes sense.

adelelopez: ah okay, that makes sense then

PR #5079, opened 2020-07-08
Reduce size of pachctl images

### LINKED ISSUE #5078: Trim size of dockerized `pachctl`
**What is the goal / desired outcome?** The newly-released [dockerized pachctl images](https://github.com/pachyderm/pachyderm/pull/5073) are very large - around 1GB. This could be trimmed significantly, most likely with multi-stage builds that don't copy over the module cache.

ysimonson:
Reduced size of pachctl images from ~1gb to ~155mb~ <200mb by:

* Using multi-stage builds to avoid storage of intermediate caches
* Using buster-slim rather than buster
* ~Not updating apt~

~One tradeoff of this is that ca certs aren't updated. My assumption here is that debian regularly pushes updates to their ca certs in their images, so apt update isn't necessary, but I'm not sure. It does look like like the buster-slim image [was last updated a month ago.](https://hub.docker.com/layers/debian/library/debian/buster-slim/images/sha256-5bc04978162be23ebb822cb8cfad22619ef8069f33b1d7b433191270a4f578ea?context=explore)~

Closes #5078

### DIFF
--- a/Dockerfile.pachctl
+++ b/Dockerfile.pachctl
@@ -1,6 +1,7 @@
 # syntax=docker/dockerfile:1.0-experimental
 ARG GO_VERSION
 FROM golang:${GO_VERSION}-buster AS pachctl_build
+RUN apt update && apt install ca-certificates
 WORKDIR /app
 COPY . .
 ARG LD_FLAGS
@@ -11,4 +12,6 @@ RUN --mount=type=cache,target=/go/pkg/mod \
     go install -ldflags "${LD_FLAGS}" -gcflags "${GC_FLAGS}" ./src/server/cmd/pachctl
 
 FROM debian:buster-slim
+COPY --from=pachctl_build /etc/ssl/certs /etc/ssl/certs
+COPY --from=pachctl_build /usr/share/ca-certificates /usr/share/ca-certificates
 COPY --from=pachctl_build /go/bin/pachctl /usr/local/bin/pachctl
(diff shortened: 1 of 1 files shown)


ysimonson:
@seslattery could you take a look? In particular I think you might know the answer to the ca certs stuff outlined in the PR description.

Dockerfile.pachctl
```diff
@@ -1,7 +1,6 @@
 # syntax=docker/dockerfile:1.0-experimental
 ARG GO_VERSION
-FROM golang:${GO_VERSION}
-RUN apt update && apt install ca-certificates
```
seslattery, line 4: I'd actually recommend leaving this line in, and copying the certificates to the final image.  

`COPY --from=pachctl_build /etc/ssl/certs /etc/ssl/certs`
`COPY --from=pachctl_build /usr/share/ca-certificates /usr/share/ca-certificates`

I think that's sufficient on debian to copy over the certs.  By default debian:buster-slim actually doesn't have any ca-certs installed.

Dockerfile.pachctl
```diff
@@ -10,3 +9,6 @@ ARG LD_FLAGS
 RUN --mount=type=cache,target=/go/pkg/mod \
```
seslattery, line 9: nit:  can we start run statements with the command instead of flags? I think it's a more standard usage

ysimonson: These are flags for BuildKit. AFAIK it's the only way to specify them; otherwise, docker would think the flags are for the `go install` command.

PR #5080, opened 2020-07-08
WIP: Autoformat docs

jrockway:
This PR adds a Prettier configuration to allow auto-formatting of docs, either in your editor or with `make format-docs`.

Marked as a work in progress because I can't get the docs to serve locally yet, to see if the formatting breaks any cosmetics.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,8 @@
 name: Bug report
 about: Create a report to help us improve
 ---
-<!-- This form is for bug reports ONLY! 
+
+<!-- This form is for bug reports ONLY!
 
 If you're looking for help, please check:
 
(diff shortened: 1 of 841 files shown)


jrockway:
Indeed, the auto-formatting breaks things; at least "!!! tip" sections are broken.  Will investigate.

svekars:
@jrockway we do use quite a lot of markdown extensions on the official docs website which formatting will likely be broken by this change. Here is the full list of extensions: https://github.com/pachyderm/pachyderm/blob/master/doc/mkdocs-master.yml#L63.

jrockway:
I think my plan is to go through and add `<!-- prettier ignore -->` directives in the affected areas.  It's a known issue: https://github.com/prettier/prettier/issues/3362

svekars:
Sounds good. Just need to make sure we add all needed exceptions.

PR #5081, opened 2020-07-08
Don't use S3 for ent key

chainlink:


chainlink:
@ysimonson how should we handle reset.py?

ysimonson:
In reset.py, instead of:

```
            enterprise_token = await capture(
                "aws", "s3", "cp",
                "s3://pachyderm-engineering/test_enterprise_activation_code.txt",
                "-"
            )
```

You can put:

```
enterprise_token = os.environ["ENT_ACT_KEY"]
```

as well as a guard toward the top of main, after the args are parsed:

```
    if "ENT_ACT_KEY" in os.environ:
        raise Exception("Must unset ENT_ACT_KEY")
```

ysimonson:
We'll probably want to get rid of `enterprise-code-checkin-test`, since we're not dealing with files containing the code, and even if we were, it wouldn't catch anything once we have per-user enterprise keys

ysimonson:
One nit, we tend to prefix env vars to make it clear that they're for pachyderm. Maybe `PACH_`?

chainlink:
Will ship this to unblock tests and will fix up reset.py in a future PR

PR #5094, opened 2020-07-13
Docker image for running example e2e tests

ysimonson:


ysimonson:
Closing. As per separate discussion, we're going to instead go with an image that's tightly coupled to hub, for ease of ops management on that end.

PR #5098, opened 2020-07-14
Filtering for fileset readers.

brendoncarroll:


src/server/pfs/server/driver_v2.go
```diff
@@ -168,39 +165,87 @@ func (d *driverV2) withFileSet(ctx context.Context, repo, commit string, f func(
...
-
-var globRegex = regexp.MustCompile(`[*?[\]{}!()@+^]`)
-
-func globLiteralPrefix(glob string) string {
-	idx := globRegex.FindStringIndex(glob)
-	if idx == nil {
-		return glob
+	name := strings.TrimRight(file.Path, "/")
+	// exact match
+	getExact := func() (fileset.FileSource, error) {
```
brycemcanally, line 206: I still think the two passes approach is unnecessary and it does not seem cleaner. I think you could get the same behavior with an index prefix of `name` and the following filter:
```
if idx.Path == name {
  return true
}
if strings.HasPrefix(idx.Path, name+"/") && idx.Path != name+"/" {
  return true
}
return false
```

Also, it seems like it could make sense to just have the index filter be a functional option in the index package that would get passed in the same way we pass in index options to the merge reader.

src/server/pfs/server/driver_v2.go
```diff
@@ -168,39 +165,87 @@ func (d *driverV2) withFileSet(ctx context.Context, repo, commit string, f func(
...
+	children := NewReader(file.Commit, getChildren)
+	// iterate
+	if err := exact.Iterate(ctx, func(finfo *pfs.FileInfoV2, _ fileset.File) error {
+		return cb(finfo)
+	}); err != nil {
+		return err
 	}
-	return glob[:idx[0]]
+	if err := children.Iterate(ctx, func(finfo *pfs.FileInfoV2, _ fileset.File) error {
+		if pathIsChild(name, finfo.File.Path) {
```
brycemcanally, line 241: Why is this necessary? Shouldn't the index filter already handle this?

src/server/pfs/server/driver_v2.go
```diff
@@ -168,39 +165,87 @@ func (d *driverV2) withFileSet(ctx context.Context, repo, commit string, f func(
...
+	// 	F: func(th *tar.Header) *tar.Header {
+	// 		th.Name = "." + th.Name
+	// 		return th
+	// 	},
+	// }
+	return fileset.WriteTarStream(w, filter)
 }
 
-func (d *driverV2) listFileV2(pachClient *client.APIClient, file *pfs.File, full bool, history int64, f func(*pfs.FileInfoV2) error) error {
+func (d *driverV2) listFileV2(pachClient *client.APIClient, file *pfs.File, full bool, history int64, cb func(*pfs.FileInfoV2) error) error {
```
brycemcanally, line 196: Few minor nits for this function:
- In general, if you want to add comments, then I think they need to add more information than can be gleaned from the code (variable / function names). The comments in this function don't seem like they add any value and they take up space.
- `finfo` -> `fi`. Similarly to `cb`, having a common naming convention for even simple local variable names really adds up in terms of removing mental overhead when reading code.
- When the last statement in a function that returns an error is just a function call that returns an error, then you should just return the result of the function call. It is a pretty trivial detail, but I am convinced that simplifying wherever you can really adds up.

src/server/pfs/server/driver_v2.go
brendoncarroll: This doesn't really do 2 passes. The exact match is hardly a pass. It's either 1 or none.  Using a prefix of `name` means we will also have to iterate and filter over things with prefix `name + x`.  Using a prefix of `name + "/"` means we only iterate over things in a hypothetical directory `name`, instead of all directories starting with `name`.

src/server/pfs/server/driver_v2.go
brendoncarroll: ListFile only returns immediate children, but we have to include recursive children to compute the hashes.

src/server/pkg/storage/fileset/transforms.go
```diff
@@ -0,0 +1,95 @@
...
+var _ File = headerMap{}
+
+type headerMap struct {
+	header *tar.Header
+	inner  File
+}
+
+func (hm headerMap) Index() *index.Index {
+	// TODO: should be able to transform the path?
+	return hm.Index()
```
brendoncarroll, line 75: Just caught this now.  Should not be infinite recursion regardless...

src/server/pfs/server/driver_v2.go
brycemcanally: Ahh, I see. This handles the skipping of children of children.

src/server/pfs/server/driver_v2.go
brendoncarroll: I can add a Concat reader in the fileset package if you would rather only create one `server.Reader`

brycemcanally: Yeah, I don't think the performance gain in that very specific situation is worth the extra code. Also, I tweaked the filter a bit to account for the children of children stuff. My issue with the two passes approach is not so much about performance, since they will probably result with cache hits, I am more concerned with all of the extra code that has to be understood and maintained.

src/server/pfs/server/paths.go
```diff
@@ -0,0 +1,51 @@
...
+		return nil, err
+	}
+	return func(s string) bool {
+		s = path.Clean(s)
+		return g.Match(s) && (parentG == nil || !parentG.Match(s))
+	}, nil
+}
+
+func pathIsChild(parent, child string) bool {
+	parent = strings.TrimLeft(parent, "/")
```
brycemcanally, line 43: I think we should split canonicalization from the check for children of children.

src/server/pkg/storage/fileset/api.go
```diff
@@ -0,0 +1,29 @@
...
+	// Header returns a *tar.Header with metadata about the file, or an error
+	Header() (*tar.Header, error)
+	// Content writes the content of the file to w
+	Content(w io.Writer) error
+}
+
+var _ File = &FileMergeReader{}
+var _ File = &FileReader{}
+
+type FileSource interface {
```
brycemcanally, line 23: Maybe this should just be `FileIterator`, since that is what it does?

brendoncarroll: `Iterator` and `Reader` both seem stateful to me.  Iterator in most other contexts, java, python etc. is a stateful object with a next or hasNext method. I think the eventual behavior here should be that Iterate can be called as many times as you want.  The MergeReader and Reader do not support that yet.

src/server/pfs/server/reader.go
```diff
@@ -0,0 +1,263 @@
...
+				h.Write([]byte(dataRef.Hash))
+			}
+		}
+	}
+	hashBytes := h.Sum(nil)
+	cache[x] = hashBytes
+	return hashBytes, nil
+}
+
+type stream struct {
```
brycemcanally, line 191: I think the following would be a cleaner design for the stream (and is closer to the stream I implemented for the datum stream). We could add a context that would be used for breaking out of the iteration so we don't need an explicit close that forces the full iteration. Also, `io.EOF` is a better way to indicate the end of  the stream rather than a nil file returned.

```
type stream struct {
	peek     fileset.File
	fileChan chan fileset.File
	errChan  chan error
}

func newStream(source fileset.Source) *stream {
	fileChan := make(chan fileset.File)
	errChan := make(chan error, 1)
	go func() {
		if err := source.Iterate(func(file fileset.File) error {
                        fileChan <- file
			return nil
		}); err != nil {
			errChan <- err
			return
		}
		close(fileChan)
	}()
	return &stream{
		fileChan: fileChan,
		errChan:  errChan,
	}
}

func (s *stream) Peek() (fileset.File, error) {
        if s.peek != nil {
            return s.peek, nil
        }
        var err error
        s.peek, err = s.Next()
        return s.peek, err
}

func (s *stream) Next() (fileset.File, error) {
        if s.peek != nil {
            tmp := s.peek
            s.peek = nil
            return tmp, nil
        }
	select {
	case file, more := <-s.fileChan:
		if !more {
			return nil, io.EOF
		}
		return file, nil
	case err := <-s.errChan:
		return nil, err
	}
}
```

src/server/pkg/storage/fileset/transforms.go
```diff
@@ -0,0 +1,95 @@
...
+	"io"
+
+	"github.com/pachyderm/pachyderm/src/server/pkg/storage/fileset/index"
+	"github.com/pachyderm/pachyderm/src/server/pkg/storage/fileset/tar"
+)
+
+var _ FileSource = &HeaderFilter{}
+
+type HeaderFilter struct {
+	F func(th *tar.Header) bool
```
brycemcanally, line 13: I think I prefer descriptive field names. The one letter variable names should be limited to variables that have a small scope.

src/server/pfs/server/reader.go
```diff
@@ -0,0 +1,263 @@
...
+	if hashBytes, exists := cache[x]; exists {
+		return hashBytes, nil
+	}
+	// check that the next item in the stream is x
+	idx := s.Peek()
+	if idx == nil {
+		return nil, errors.Errorf("stream is done, can't compute hash for %s", x)
+	}
+	if idx.Path != x {
+		return nil, errors.Errorf("stream is is wrong place to compute hash for %s", x)
```
brycemcanally, line 153: is is

src/server/pfs/server/reader.go
```diff
@@ -0,0 +1,263 @@
...
+		return err
+	}
+	r2, err := r.getReader()
+	if err != nil {
+		return err
+	}
+	s1, s2 := newStream(r1), newStream(r2)
+	defer s1.Close()
+	defer s2.Close()
+	cache := make(map[string][]byte)
```
brycemcanally, line 118: We want to cache the full FileInfoV2.

src/server/pfs/server/source.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+}
+
+func (s *Source) Iterate(ctx context.Context, cb func(*pfs.FileInfoV2, fileset.File) error) error {
+	ctx, cf := context.WithCancel(ctx)
+	defer cf()
+	fs1 := s.getReader()
+	fs2 := s.getReader()
+	s2 := newStream(ctx, fs2)
+	cache := make(map[string][]byte)
```
brendoncarroll, line 114: Will change this to `FileInfoV2`s with updates to the protobuf for size and whatever else.  I made a note with that item in the doc.

src/server/pfs/server/source.go
```diff
@@ -0,0 +1,255 @@
...
+	return fs1.Iterate(ctx, func(fr fileset.File) error {
+		idx := fr.Index()
+		finfo := &pfs.FileInfoV2{
```
brycemcanally, line 117: finfo -> fi

src/server/pfs/server/paths.go
```diff
@@ -0,0 +1,76 @@
...
+// cleanPath converts paths to a canonical form used in the driver
+// "" -> "/"
+// "abc" -> "/abc"
+// "/abc" -> "/abc"
+// "/" -> "/"
+func cleanPath(x string) string {
+	if x == "" {
+		x = "/"
+	}
+	if len(x) > 1 && x[0] != '/' {
```
brycemcanally, line 72: I think this is going to miss a length 1 path? I don't think you need the length check. I think this could just be:
```
if strings.HasPrefix(str, "/") {
  return str
}
return "/" + str
```

src/server/pfs/server/source.go
```diff
@@ -0,0 +1,255 @@
...
+	// check that the next item in the stream is x
+	fr, err := s.Peek()
+	if err != nil {
+		if err == io.EOF {
+			return nil, errors.Errorf("stream is done, can't compute hash for %s", x)
+		}
+		return nil, err
+	}
+	idx := fr.Index()
+	if idx.Path != x {
```
brycemcanally, line 154: Not a big fan of the use of the variable name `x` here and in other places. Even though it is locally scoped, nothing about the letter `x` is relevant to the path we are computing the hash of. I think some options for better names would be `pathStr` or `idxPath`. In general lets try to have the variable names be meaningful, even the locally scoped ones. I'm not against single letter local variables when the purpose of the variable is clear. This is just a string with the variable name `x` which tells me nothing about the meaning of this variable in this function.

src/server/pfs/server/source.go
```diff
@@ -0,0 +1,255 @@
...
+		delete(cache, idx.Path)
+		return nil
+	})
+}
+
+func computeHash(cache map[string][]byte, s *stream, x string) ([]byte, error) {
+	if hashBytes, exists := cache[x]; exists {
+		return hashBytes, nil
+	}
+	// check that the next item in the stream is x
```
brycemcanally, line 145: I don't think we need to peek. I think we can just progress the stream then error if it is not in the correct place. I don't think having the streams out of sync will be recoverable.

src/server/pfs/server/source.go
```diff
@@ -0,0 +1,255 @@
...
+		h.Write(childHash)
+	}
+	hashBytes := h.Sum(nil)
+	cache[x] = hashBytes
+	return hashBytes, nil
+}
+
+func computeFileHash(idx *index.Index) []byte {
+	h := pfs.NewHash()
+	if idx.DataOp != nil {
```
brycemcanally, line 194: Lets go ahead and get the fix in for full chunk references. `dataRef.Hash` will be `""` if it is a full chunk reference, so the hash should be pulled from `dataRef.ChunkInfo.Chunk.Hash` in that case.

src/server/pkg/storage/fileset/reader.go
```diff
@@ -86,6 +94,10 @@ func (fr *FileReader) Header() (*tar.Header, error) {
 		}
 		var err error
 		fr.hdr, err = tar.NewReader(buf).Next()
+		if err == io.EOF {
+			fmt.Println(fr.idx)
+			//panic(err)
```
brycemcanally, line 99: remove

src/server/pkg/storage/fileset/reader.go
```diff
@@ -86,6 +94,10 @@ func (fr *FileReader) Header() (*tar.Header, error) {
 		}
 		var err error
 		fr.hdr, err = tar.NewReader(buf).Next()
+		if err == io.EOF {
```
brycemcanally, line 97: remove

src/server/pkg/storage/fileset/index_resolver.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return err
+	}
+	w := mr.s.newWriter(ctx, "", WithNoUpload(func(idx *index.Index) error {
+		fmr, err := mr2.Next()
+		if err != nil {
+			return err
+		}
+		if fmr.Index().Path != idx.Path {
+			panic("different merge readers from getReader")
```
brycemcanally, line 43: I'm not sure how I feel about a panic here. I don't think we want to kill the pachd instance because we got into this state when doing a merge.

src/server/pkg/storage/fileset/util.go
```diff
@@ -12,3 +17,43 @@ func WithLocalStorage(f func(*Storage) error) error {
...
+func WriteTarEntry(w io.Writer, f File) error {
+	h, err := f.Header()
+	if err != nil {
+		return err
+	}
+	tw := tar.NewWriter(w)
+	if err := tw.WriteHeader(h); err != nil {
+		return err
+	}
+	if (h.Mode == tar.TypeReg || h.Mode == tar.TypeRegA) && h.Size > 0 {
```
brycemcanally, line 42: You shouldn't need this check. `Content` should write zero bytes when the size is 0.

src/server/pfs/server/source.go
brendoncarroll: I'll change it here.  For simple pure functions, like `cleanPath` for example, there is no better argument name than `in` or `input` or `x`, which has a pretty deep tradition of being used as an input variable in mathematics. It's also one character.

src/server/pkg/storage/fileset/index_resolver.go
brendoncarroll: I'm usually okay with panicking on logic errors.  But I think there are probably cases here where the semantic path could change in between calls, or other out of process things happen.

src/server/pfs/server/source.go
brycemcanally: I mean `pathStr` would be a pretty clear name for the input to a `cleanPath` function. Conventions for expressing mathematical functions are not always clear when trying to apply them to programming. I don't really care too much about the `cleanPath` case though because that is simple enough to intuit the purpose of the variable.

src/server/pfs/server/paths.go
```diff
@@ -0,0 +1,71 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	g, err := globlib.Compile(glob, '/')
+	if err != nil {
+		return nil, err
+	}
+	return func(s string) bool {
+		s = path.Clean(s)
```
brycemcanally, line 38: Why is this here?

brendoncarroll: I think that was to get the match function working on the tar names before path cleaning. In the latest commit the match function is only used in GetTarConditional.

brycemcanally: Ohh, I see. We will need to unify this functinality, but not in this PR.

PR #5109, opened 2020-07-16
Rework & fix formatting of release notes

ysimonson:


ysimonson:
@nitinjainsj I wrote up the docs a more detailed extract/restore, but I really think it adds way more confusion than it elucidates. Instead, maybe we should just [link of to a diff that demonstrates what's involved?](https://github.com/pachyderm/pachyderm/pull/5111/commits/d96b82725d5683925d730f1554b67152aa450970)

PR #5111, opened 2020-07-16
v1.12

ysimonson:


src/client/admin.go
```diff
@@ -84,10 +84,10 @@ func (c APIClient) ExtractPipeline(pipelineName string) (*pps.CreatePipelineRequ
 	if err != nil {
 		return nil, grpcutil.ScrubGRPC(err)
 	}
-	if op.Op1_11 == nil || op.Op1_11.Pipeline == nil {
+	if op.Op1_12 == nil || op.Op1_12.Pipeline == nil {
```
msteffen, line 87: This is probably a question for Nitin, but when we change the version of proto that we extract, should we change the version in src/client/version/client.go at the same time?

ysimonson: The version is changed in this PR: https://github.com/pachyderm/pachyderm/pull/5111/commits/959dbb0642583399758e4b881dfed227838cb108#diff-0b1649da9b588637da6a5311c4725f9f

PR #5132, opened 2020-07-22
Push the image that was built

lukemarsden:
This change fixes `(cd etc/pipeline-build; make push-to-minikube)` for me, by making it try to push the image it just built, rather than one named e.g. 'go' or 'python'.


lukemarsden:
Thanks! Ignoring test flakes for this change which doesn't affect CI (since CI doesn't use minikube), and merging.

PR #5137, opened 2020-07-23
Workaround Travis/GitHub flakiness and also retry our tests

lukemarsden:
This PR addresses two worries:

# Travis/GitHub flakiness
As documented [here](https://www.dropbox.com/scl/fi/361jzngrcfhb6yic4es9o/Notes-on-CI-issues.paper?dl=0&rlkey=uf6bz9pxpgfgfsohrsc5vrjwb), Travis/GitHub sometimes give you the wrong commit: sometimes a too-old one, sometimes a too-new one. This PR implements the design proposed in the document to workaround this.

# Our test flakiness

The change also introduces the use of the `travis_retry` operator in `.travis.yml`, which will retry our tests 3 times before giving up. Various fixes were made to support retrying a test.

This will reduce the appearance of flakiness and reduce the manual overhead of retrying flakes.

This is not a replacement for root-cauisng and fixing flaky tests -- this should still be addressed separately.


etc/testing/travis_install.sh
```diff
@@ -2,6 +2,10 @@
 
 set -ex
 
+# TODO(lukemarsden): try bumping dist: to bionic in .travis.yml and setting
+# services: - docker to see if docker 18.06 is sufficient, rather than
+# downloading docker every time. Also, why do we use devicemapper?
```
ysimonson, line 7: There was an attempt to upgrade dist to bionic, along with upgrading minikube [here.](https://github.com/pachyderm/pachyderm/pull/4799) It ended up being less straight-forward that I thought, so I scrapped it. Just upgrading the dist to bionic (w/o upgrading minikube) may work though.

.travis.yml
```diff
@@ -8,6 +8,8 @@ cache:
     - ${HOME}/cached-deps
     - ${HOME}/gopath/pkg/mod
     - "$(python3 -c 'import site; print(site.USER_BASE)')"
+    - ${HOME}/docker_images
```
ysimonson, line 11: Docker image caching was scrapped [in this PR](https://github.com/pachyderm/pachyderm/pull/5007) for two reasons:
1) The cache would slowly leak and increase in size, eventually creating cache assets that are hundreds of gigabytes in size and triggering timeouts.
2) Caches aren't shared between the build step and test steps.

.travis.yml
```diff
@@ -47,11 +49,12 @@ before_install:
   - etc/testing/travis_cache.sh
 install:
   - etc/testing/travis_install.sh
-script: etc/testing/travis.sh
+script: travis_retry timeout 9m etc/testing/travis.sh; status=$?; if [ $status -ne 0 ]; then etc/testing/kube_debug.sh; fi; exit $status
```
ysimonson, line 52: A bunch of debug info is already printed, see the `after_failure` section of this file

lukemarsden: Ah yes! And, as I'm trying to debug test failures in this `travis_retry` block, I'm also interested in seeing debug info before the end of the retries. I shall try to unify these, though.

.travis.yml
lukemarsden: OK, useful context thanks! I've moved to trying to only cache specific named docker images to improve build times, but that may be futile, and I'm close to ripping it back out :) The main thrust of this PR will be retries (reliability) rather than speed, but it's always tempting to work on improving the speed of CI while waiting for it to run!

etc/testing/travis_install.sh
lukemarsden: Thanks! I'll probably just delete this comment then. I was thinking of ways to speed up the setup, and not having to download and install docker every time would have helped a bit, but it's probably not the overriding cost.

.travis.yml
lukemarsden: Unified in https://github.com/pachyderm/pachyderm/pull/5137/commits/390ff3589634c8c6fccff17ff27afae92b9a4f3a

etc/testing/travis_build_check_and_stash.sh
```diff
@@ -0,0 +1,85 @@
...
+#
+# We know that we are trying to test code in a merge preview with
+# $TRAVIS_PULL_REQUEST_SHA on the RHS of the merge though, that information is
+# reliable at least.
+#
+# So, try a few times to fetch and identify a valid version of the code. If we
+# time out, give up. If we succeed, stash that code in a docker image which
+# later tests can use to get it back.
+
+cd /home/travis/gopath/src/github.com/pachyderm/pachyderm
```
ysimonson, line 21: why the cd here?

etc/testing/travis_build_check_and_stash.sh
```diff
@@ -0,0 +1,85 @@
...
+# commit ID at HEAD, which will break trying to pull docker images tagged with
+# the output of pachctl version).
+
+if [[ ! "$TRAVIS_SECURE_ENV_VARS" == "true" ]]; then
+    echo "Need travis env vars so we can auth to docker hub."
+    exit 1
+fi
+docker login -u pachydermbuildbot -p "${DOCKER_PWD}"
+
+cd /home/travis/gopath/src/github.com/pachyderm
```
ysimonson, line 74: why do we need to cd to the parent directory here?

etc/testing/travis_build_check_and_stash.sh
```diff
@@ -0,0 +1,85 @@
...
+tar cf /tmp/save_git_tarball/pachyderm.tar pachyderm
+cd /tmp/save_git_tarball
+
+cat <<EOT >Dockerfile
+FROM ubuntu:xenial
+COPY pachyderm.tar /
+EOT
+
+docker build -t pachyderm/ci_code_bundle:${TRAVIS_BUILD_NUMBER} .
+docker push pachyderm/ci_code_bundle:${TRAVIS_BUILD_NUMBER}
```
ysimonson, line 85: This would be a little more work, but I feel like it might be cleaner for us to just setup a dedicated bucket for storing CI build assets. We could then store:

* the pachd and worker images
* `pachctl` so it doesn't have to be re-compiled
* `pachyderm.tar`

etc/testing/travis_test_check_and_pull.sh
```diff
@@ -0,0 +1,43 @@
...
+
+if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
+    # These shenannigans not needed for release and branch builds, hopefully.
+    exit 0
+fi
+
+if [[ ! "$TRAVIS_SECURE_ENV_VARS" == "true" ]]; then
+    echo "Need travis env vars so we can auth to docker hub."
+    exit 1
+fi
```
ysimonson, line 20: Note that PRs from outside contributors will not have access to secure env vars. Will this break their PRs?

etc/testing/travis_build_check_and_stash.sh
lukemarsden: Because we're about to tar up the code, and where you tar from matters as it determines the directory structure inside the tar file

etc/testing/travis_build_check_and_stash.sh
lukemarsden: Mostly as documentation of where we are (context for the git commands to operate in) and to make sure we are where we think we are

etc/testing/travis_build_check_and_stash.sh
lukemarsden: Images are probably better in a container registry (a local one would be slightly better than docker hub for intra-job asset passing), but I agree about the binaries and the tarball. Ok if we make this a follow-up issue?

etc/testing/travis_test_check_and_pull.sh
lukemarsden: Good catch - I'll test this and fix any breakages

lukemarsden: Is this not already broken if build stage needs docker hub access?

etc/testing/travis_build_check_and_stash.sh
ysimonson: The repo is stored in a directory directly under `/home/travis` though, right? (Maybe `/home/travis/pachyderm`?)

lukemarsden: It seems to have ended up in the gopath. Not sure how it got there :)

etc/testing/travis_test_check_and_pull.sh
lukemarsden: (and if that's the case, let's move it to a followup issue but definitely fix it soon)

etc/testing/travis_build_check_and_stash.sh
lukemarsden: Looks like this is magic from `language: go` https://docs.travis-ci.com/user/languages/go/#go-import-path

lukemarsden:
CI is doing its job and failed on some linting of the shell that I wrote. Hoping that now that's fixed we should see some green test runs. I hope to see a string of green test runs (with no manual retries, or at least far fewer than we had to do before) before we merge this.

etc/testing/travis_build_check_and_stash.sh
ysimonson: Sure thing

etc/testing/travis_test_check_and_pull.sh
ysimonson: In the case of outside PRs, we don't push images, we just quickly exit the build step and run the build process in each test bucket

lukemarsden:
Looks like we still have issues with retries in the case of the auth tests.

etc/testing/travis_build_check_and_stash.sh
ysimonson: circling back to the images, I feel like they'd be better in a bucket because then we wouldn't pollute our docker hub account with a bunch of releases of `pachyderm/pachd` and `pachyderm/worker` -- though it's not a huge deal. It's straight-forward to serialize images via `docker save` and `docker load`, too.

lukemarsden: sure, makes sense

lukemarsden: https://github.com/pachyderm/pachyderm/issues/5144

etc/testing/travis_test_check_and_pull.sh
lukemarsden: got it - https://github.com/pachyderm/pachyderm/pull/5137/commits/9bf5f669cad5cd79735a2bae9e9f5a8ac23168e8 should fix this

lukemarsden:
Per detailed notes in [https://www.dropbox.com/scl/fi/361jzngrcfhb6yic4es9o/Notes-on-CI-issues.paper?dl=0&rlkey=uf6bz9pxpgfgfsohrsc5vrjwb](the CI issues doc), I now recommend that we merge this as-is.

The current state of this PR is certainly a major improvement on what we have already:
* for some pipelines [retries are helping](https://travis-ci.com/github/pachyderm/pachyderm/jobs/365287566) :smile: 
* And we’re [refusing to proceed when we can’t get the correct version of the code to test](https://travis-ci.com/github/pachyderm/pachyderm/jobs/365165570) :smile: (I later made this failure actually fail the build, so that it doesn’t bother trying to run the tests)
* Our tests seem to be much more likely to pass when the Travis build number starts with a 9 than a 1 :open_mouth:  I reckon the 9 builds are on some newer infrastructure.

Followup issues can be raised to further deep dive into specific flakes (see doc above) and for followup improvements.

etc/testing/kube_debug.sh
```diff
@@ -0,0 +1,18 @@
...
+# probably show why things are hanging.
+
+kubectl version
+kubectl get all --namespace kafka
+kubectl describe pod -l app=pachd
+kubectl describe pod -l suite=pachyderm,app=etcd
+kubectl logs -l app=pachd | tail -n 100
+sudo dmesg |tail -n 20
+top -b -n 1|head -n 20
+df -h
```
ysimonson, line 18: do these last 3 lines provide helpful output?

etc/testing/travis.sh
```diff
@@ -2,14 +2,34 @@
...
 set -ex
 
+(
+    # Stop Travis from timing us out in 10m when we want to get retried after 20m
+    set +x
+    while true; do
+        echo "liveness ping $(date)"
+        sleep 60
+    done
+) &
```
ysimonson, line 12: This seems risky to me - won't this stop all timeouts?

etc/testing/kube_debug.sh
lukemarsden: I suspect they may do in the future, in case we hit OOM, or memory or disk limits.

etc/testing/travis.sh
lukemarsden: Execution of `travis.sh` is wrapped in a `timeout` command, see https://github.com/pachyderm/pachyderm/pull/5137/files#diff-354f30a63fb0907d4ad57269548329e3R59

PR #5138, opened 2020-07-24
Example/housing prices

JimmyWhitaker:
I finished the housing prices example. It has a lot in there, so let me know if something isn't explained well or is missing.


examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+  --target-col TARGET_COL
+                        column with target values
+  --output DIR          output directory
+  ```
+
+The regression code performs three functions:
+
+1. Data analysis
+2. Train a regressor
+3. Evaluate the model
```
JoeyZwicker, line 90: is there a reason we have just one pipeline step that does all 3 of these? is it common for this type of model/process to do that all together or would these typically be broken up into 2 or 3 different pipeline steps?

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Data Analysis
+The first step in the pipeeline creates a pairplot showing the relationship between features. By seeing what features are positively or negatively correlated to the target value (or each other), it can helps us understand what features may be valuable to the model.
```
JoeyZwicker, line 93:
```suggestion
The first step in the pipeline creates a pairplot showing the relationship between features. By seeing what features are positively or negatively correlated to the target value (or each other), it can helps us understand what features may be valuable to the model.
```

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+        "pfs": {
+            "glob": "/*",
+            "repo": "housing_data"
+        }
+    },
+    "transform": {
+        "cmd": [
+            "python",
+            "regression.py",
+            "--input",
```
JoeyZwicker, line 166: If you agree with argument layout in other comment, update here too

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+When we inspect the learning curve, we can see that there is a large gap between the training score and the validation score. This typically indicates that our model could benefit from the addition of more data. 
+
+<p align="center">
+  <img width="400" height="350" src="images/learning_curve-1.png">
+</p>
+
+So let's update our dataset with additional examples.
+
+### Step 5: Update Dataset
+Here's where pachyderm truely starts to shine. To update our dataset we can run the following command (note that we could also append new examples to the existing file):
```
JoeyZwicker, line 236:
```suggestion
Here's where Pachyderm truly starts to shine. To update our dataset we can run the following command (note that we could also append new examples to the existing file, but in this example we're simply overwriting our previous file to one with more data):
```

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+```bash
+$ pachctl put file housing_data@master:housing-simplified.csv -f data/housing-simplified-2.csv --overwrite
+```
+
+The new commit of data to the `housing_data` repository automatically kicks off a job on the `regression` pipeline without us having to do anything. 
+
+When the job is complete we can download the new files and see that our model has improved, given the new learning curve.
```
JoeyZwicker, line 244: I think we can add a good callout/summary here to data lineage & versioning for experiment tracking. Dont want to make this too much long, but something along the lines of:

Note that because versions all of our input and output data automatically, we can continue to iterate on our data and code and Pachyderm will track all of our experiments. For any given output commit, Pachyderm will tell us exactly which input commit of data was run. In our simple example we only have 2 experiments run so far, but this becomes incredibly important and valuable when we do many more iterations.

Show list commit on the input and output repo. Show inspect commit on the earlier (less good) output commit and experiment and highlight that it includes a reference to the first input commit.

examples/ml/housing-prices/regression.json
```diff
@@ -0,0 +1,25 @@
+{
+    "pipeline": {
+        "name": "regression"
+    },
+    "description": "A pipeline that trains produces a regression model for housing prices.",
+    "input": {
+        "pfs": {
+            "glob": "/*",
```
JoeyZwicker, line 8: Just to double check ourselves, do we want `/*` here or `/`?

How would your code handle multiple files added instead of the just 1 that is overwritten? Looks like it would create a new output for each input file, right?

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+For the pipeline **input** we can define what data repo(s) our pipeline is connected to and what [glob pattern](https://docs.pachyderm.com/latest/concepts/pipeline-concepts/datum/glob-pattern/) to use. A glob pattern tells the pipeline how to map data into a job, here we have it create a new job for each datum in the `housing_data` repository.
+
+The **image** defines what Docker image will be used for the pipeline, and the **transform** is the command run once a pipeline job starts.
+
+Once this pipeline is created, it will be looking for any changes to its input, if detected it will start a new job to train given the new dataset.
+
+```bash
+$ pachctl create pipeline -f regression.json
```
svekars, line 185: we don't use the $ sign in the command blocks.

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+The pipeline will write the output to a PFS repo (`/pfs/out/` in the pipeline json) created with the same name as the pipeline.
```
svekars, line 188:
```suggestion
The pipeline writes the output to a PFS repo (`/pfs/out/` in the pipeline json) created with the same name as the pipeline.
```

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Step 3: Add the housing dataset to the repo
+Now we can add the data which will kick off the processing automatically. If we update the data with a new commit, then the pipeline will automatically re-run.
```
svekars, line 191:
```suggestion
Now we can add the data, which will kick off the processing automatically. If we update the data with a new commit, then the pipeline will automatically re-run. 
```

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+$ pachctl get file regression@master:/ --recursive --output .
+```
+
+When we inspect the learning curve, we can see that there is a large gap between the training score and the validation score. This typically indicates that our model could benefit from the addition of more data. 
+
+<p align="center">
+  <img width="400" height="350" src="images/learning_curve-1.png">
+</p>
+
+So let's update our dataset with additional examples.
```
svekars, line 233:
```suggestion
Now let's update our dataset with additional examples.
```

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+</p>
+
+The Pachyderm Pipeline will perform the following:
+
+1. Import the structured dataset (`.csv`) with pandas
+2. Perform some light data analysis with scikit-learn
+3. Train a regression model to predict housing prices
+4. Generate a learning curve and performance metrics to estimate the quality of the model
+
+Outline of this Example:
```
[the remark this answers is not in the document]
JimmyWhitaker: Agreed -changing.

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+    - [Evaluate the model](#evaluate-the-model)
+  - [Pachyderm Pipeline](#pachyderm-pipeline)
+    - [TLDR; Just give me the code](#tldr-just-give-me-the-code)
+    - [Step 1: Create input data repository](#step-1-create-input-data-repository)
+    - [Step 2: Create the regression pipeline](#step-2-create-the-regression-pipeline)
+    - [Step 3: Add the housing dataset to the repo](#step-3-add-the-housing-dataset-to-the-repo)
+    - [Step 4: Download files once the pipeline has finished](#step-4-download-files-once-the-pipeline-has-finished)
+    - [Step 5: Update Dataset](#step-5-update-dataset)
+
+## Housing prices datasets
```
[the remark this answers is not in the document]
JimmyWhitaker: Corrected.

examples/ml/housing-prices/regression.json
JimmyWhitaker: It's a good question. I had originally thought that each new file would be a different regression dataset (i.e. Boston Housing and separately California Housing). We could bake this in as a separate branch for each dataset as well though. If this example grows, the we would want to consider the dataset structure in more detail, because we would potentially want a dedicated testing set that doesn't change with new data (so models could be comparable). Do you this example would go more towards multiple datasets or growing a single dataset?

examples/ml/housing-prices/regression.json
```diff
@@ -0,0 +1,25 @@
...
+    "transform": {
+        "cmd": [
+            "python",
+            "regression.py",
+            "--input",
+            "/pfs/housing_data/",
+            "--target-col",
+            "MEDV",
+            "--output",
+            "/pfs/out/"
```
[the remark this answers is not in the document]
JimmyWhitaker: That would be invalid `json` with the space in there, sadly. I can put it all on the same line like this if you like? 
```
"cmd": [
            "python", "regression.py", "--input", "/pfs/housing_data/", "--target-col", "MEDV", "--output", "/pfs/out/"
        ],

File: examples/ml/housing-prices/regression.json [hunk 14]
```diff
@@ -0,0 +1,25 @@
...
+        ],
+        "image": "jimmywhitaker/housing-prices"
```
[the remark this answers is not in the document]
JimmyWhitaker: Agreed. Any opinions on tag naming conventions? I've typically gone the route of having tags like `v0.1`, but I think I've seen Pachyderm do `v1` for tags.

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+            "MEDV",
+            "--output",
+            "/pfs/out/"
+        ],
+        "image": "jimmywhitaker/housing-prices"
+    }
+}
+```
+
+For the pipeline **input** we can define what data repo(s) our pipeline is connected to and what [glob pattern](https://docs.pachyderm.com/latest/concepts/pipeline-concepts/datum/glob-pattern/) to use. A glob pattern tells the pipeline how to map data into a job, here we have it create a new job for each datum in the `housing_data` repository.
```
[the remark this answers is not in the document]
JimmyWhitaker: Edited.

examples/ml/housing-prices/README.md
JimmyWhitaker: Corrected.

examples/ml/housing-prices/README.md
JimmyWhitaker: I like it. Will add basically exactly what you have here with some examples.

examples/ml/housing-prices/README.md
JimmyWhitaker: I think the lack of a new command indicator is a little difficult to read in code blocks sometimes. For example, [here](https://github.com/pachyderm/pachyderm/blob/example/housing-prices/examples/ml/housing-prices/README.md#step-6-inspect-the-pipeline-lineage). What do you recommend?

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Getting Started
+
+This example requires a running Pachyderm deployment. You can deploy a cluster on [PacHub](hub.pachyderm.com) or deploy locally as described here:
+
+- [Pachyderm Getting Started](https://docs.pachyderm.com/latest/getting_started/)
+
+Once everything is up, we can check the setup by running:
+
+1. `kubectl get all` to ensure all the pods are up.
```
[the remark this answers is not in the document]
JimmyWhitaker: Corrected.

examples/ml/housing-prices/README.md
```diff
@@ -0,0 +1,247 @@
...
+    - [Step 5: Update Dataset](#step-5-update-dataset)
+
+## Housing prices datasets
+
+The housing prices dataset used for this example is a reduced version of the original, focusing on a subset of the features. It contains 3 input features (RM, LSTST, and PTRATIO) and the output, or target (MEDV) that we are learning to predict.
+|Feature| Description|
+|---|---|
+|RM |       Average number of rooms per dwelling|
+|LSTAT |    Percent lower status of the population|
+|PTRATIO |  Pupil-teacher ratio by town|
```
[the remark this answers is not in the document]
JimmyWhitaker: I added some more information to the description of the features and what they roughly indicate. I can't say exactly why some of the features were originally selected, in the general case, but we selected the subset because they had correlation to the output class and made the problem run faster.

examples/ml/housing-prices/README.md
JimmyWhitaker: Added an additional step to inspect the lineage.

examples/ml/housing-prices/README.md
JimmyWhitaker: I kept them all in one pipeline to keep the example simple, but also was setting up for potentially doing multiple pipelines in the future. Any of these functions would need access to the data, so it would make sense to execute some of them in parallel pipelines, such as data analysis and training. But we would want training and eval to be in the same pipeline unless we separated out the test data into another repo.

PR #5139, opened 2020-07-24
Fix stdin in reset.py

ysimonson:
Fix a minor bug where stdin wasn't properly fed when the timeout was set


lukemarsden:
I see that `stdin` wasn't being correctly `utf-8` encoded in the timeout case, and this simplification LGTM.

PR #5146, opened 2020-07-27
Adds compactResult type to return information about compaction

brendoncarroll:
This adds a `compactResult` type to the `./src/server/pfs/server` package with a field `OutputSize` for returning the size of the compacted data after compaction.
Also has changes to `fileset.Writer` which allow a callback to be called with index data without disabling chunk uploading.

### DIFF
--- a/src/server/pfs/server/driver_v2.go
+++ b/src/server/pfs/server/driver_v2.go
@@ -129,7 +129,7 @@ func (d *driverV2) finishCommitV2(txnCtx *txnenv.TransactionContext, commit *pfs
 		if err != nil {
 			return err
 		}
-		commitInfo.SizeBytes = compactRes.OutputSize
+		commitInfo.SizeBytes = uint64(compactRes.OutputSize)
 		commitInfo.Finished = types.TimestampNow()
 		return d.writeFinishedCommit(txnCtx.Stm, commit, commitInfo)
 	})
@@ -353,11 +353,11 @@ func (fr *FileReader) drain() error {
 	return nil
 }
 
-type compactResult struct {
-	OutputSize uint64
+type compactStats struct {
+	OutputSize int64
 }
 
-func (d *driverV2) compact(master *work.Master, outputPath string, inputPrefixes []string) (*compactResult, error) {
+func (d *driverV2) compact(master *work.Master, outputPath string, inputPrefixes []string) (*compactStats, error) {
 	ctx := master.Ctx()
 	// resolve prefixes into paths
 	inputPaths := []string{}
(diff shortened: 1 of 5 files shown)


src/server/pfs/server/driver_v2.go
```diff
@@ -352,7 +353,11 @@ func (fr *FileReader) drain() error {
 	return nil
 }
 
-func (d *driverV2) compact(master *work.Master, outputPath string, inputPrefixes []string) error {
+type compactResult struct {
```
brycemcanally, line 356: I think `compactStats` might be a better name for this. Also, lets stick with `int64` for the size type across the board.

src/server/pfs/server/driver_v2.go
```diff
@@ -381,7 +386,7 @@ type compactSpec struct {
 
 // compactIter is one level of compaction.  It will only perform compaction
 // if len(inputPaths) <= params.maxFanIn otherwise it will split inputPaths recursively.
-func (d *driverV2) compactIter(ctx context.Context, params compactSpec) (retErr error) {
+func (d *driverV2) compactIter(ctx context.Context, params compactSpec) (ret *compactResult, retErr error) {
```
brycemcanally, line 389: If a named return variable is unused, just use `_` instead.

src/server/pfs/server/driver_v2.go
```diff
@@ -423,7 +428,7 @@ func (d *driverV2) compactIter(ctx context.Context, params compactSpec) (retErr
 // gives those shards to workers, and waits for them to complete.
 // Fan in is bound by len(inputPaths), concatenating shards have
 // fan in of one because they are concatenated sequentially.
-func (d *driverV2) shardedCompact(ctx context.Context, master *work.Master, outputPath string, inputPaths []string) (retErr error) {
+func (d *driverV2) shardedCompact(ctx context.Context, master *work.Master, outputPath string, inputPaths []string) (ret *compactResult, retErr error) {
```
brycemcanally, line 431: Same as above.

src/server/pkg/storage/fileset/option.go
```diff
@@ -65,10 +65,17 @@ type WriterOption func(w *Writer)
 // WithNoUpload sets the writer to no upload (will not upload chunks).
```
brycemcanally, line 65: I think it might be better to just nest the `WithIndexCallback` option here. Also, `f` -> `cb` if you would.

src/server/pkg/storage/fileset/writer.go
```diff
@@ -198,15 +201,19 @@ func (w *Writer) Close() error {
 	}
 	// Write out the last index.
 	if w.lastIdx != nil {
+		idx := w.lastIdx
+		if !w.noUpload {
```
brycemcanally, line 205: Just for consistency, lets execute the idx callback before we write the index here. If you prefer the opposite, then switch the order in the callback.

brendoncarroll: I have a weak preference for always after.  It seems like having an un-uploaded index in hand is less useful than an uploaded index.  I will change to match.

src/server/pkg/storage/fileset/option.go
brendoncarroll: What do you think about removing the callback from `WithNoUpload` and have the 2 options be totally orthogonal?

brycemcanally: I considered that approach, but I cannot come up with a use case for `NoUpload` without the index callback. As is, there would not be a way to collect or store any information with `NoUpload` without the index callback, which does not seem useful to me. I don't have a strong opinion either way, but I think having `NoUpload` include an index callback makes sense.

PR #5148, opened 2020-07-28
Lint first; disable Travis cache

lukemarsden:
## Lint first
Lint as basically the first thing we do, and fail the build if it fails. This catches a lot of errors that are otherwise buried deep in the Travis logs and gives developers faster feedback than waiting ~20 minutes and then digging through the logs.

Examples of where this would have helped:
* https://github.com/pachyderm/pachyderm/pull/5147
* https://github.com/pachyderm/pachyderm/pull/5146

## Stop using the Travis cache
The travis cache is unreliable in what it gives you, but worse, loading the cache takes 3+ minutes on every single build. Just downloading the binaries each time (and compiling the code from scratch) is almost certainly faster.


lukemarsden:
Closing this after discussion with Yusuf.

* The cache helps protect against services like GCS, hashicorp releases server, and github being down. So let's not sacrifice speed for reliability.
* Also, let's not slow down all the tests by linting (~2 minutes) in the build. Instead, let's make a separate lint bucket that doesn't get retried (so that the signal from it is clearer). Follow-up issue: https://github.com/pachyderm/pachyderm/issues/5149

PR #5156, opened 2020-07-29
Setup basic tar utilities

brycemcanally:
This PR moves our fork of the standard tar package to `src/server/pkg` and adds a `tarutil` package in the same place with some basic utilities for writing, reading, and validating tar streams.


etc/testing/lint.sh
```diff
@@ -6,7 +6,7 @@ GIT_REPO_DIR=$(cd "$( dirname "${BASH_SOURCE[0]}" )" 2>&1 > /dev/null && git rev
 
 go get -u golang.org/x/lint/golint
 for file in $(find "${GIT_REPO_DIR}/src" -name '*.go' | grep -v '\.pb\.go'); do
-    if [[ $file == *fileset/tar* ]]; then
+    if [[ $file == *pkg/tar* ]]; then
```
brendoncarroll, line 9: Our linter doesn't like a package straight from the standard library?

brycemcanally: Yep, lol.

src/server/pkg/tarutil/util.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func newFile(hdr *tar.Header, data []byte) File {
+	return &file{
+		hdr:  hdr,
+		data: data,
+	}
+}
+
+func NewHeader(name string, size int64) *tar.Header {
```
brendoncarroll, line 32: This should set the TypeFlag based on trailing slash like in the package.  And then also zero out size.

src/server/pkg/tarutil/util.go
```diff
@@ -0,0 +1,106 @@
...
+	}()
+	return cb(tw)
+}
+
+func Iterate(r io.Reader, cb func(File) error) error {
+	tr := tar.NewReader(r)
+	for {
+		hdr, err := tr.Next()
+		if err != nil {
+			if errors.Is(err, io.EOF) {
```
brendoncarroll, line 77: I assume this is like a root cause thing? Idk if it's appropriate for a stop iteration sentinel like `io.EOF`.

brendoncarroll: Put another way: the protocol is "io.EOF signals the end of the tar stream". Not "something happening which has a root, intermediate, or proximate cause of io.EOF" signals the end of the tar stream.

src/server/pkg/tarutil/util.go
brycemcanally: Not sure I understand what you are suggesting here? The TypeFlag is set at write time (`WriteHeader`) and the size should be set in accordance with the size of the data (so a directory would be size zero).

src/server/pkg/tarutil/util.go
brycemcanally: There is another error type called `io.ErrUnexpectedEOF` that is intended for actual errors, and it will be emitted by the tar reader when unexpected `io.EOF`s occur. `io.EOF` is a pretty standard sentinel for the successful end of a stream / file.

src/server/pkg/tarutil/util.go
brendoncarroll: My concern is that this is a public constructor. The normal usecase is creating a header right before writing it.  This seems like it could be used to create and pass around an invalid header.

src/server/pkg/tarutil/util.go
brendoncarroll: I agree that it is a standard sentinel.  It would be nonstandard to signal the end of iteration with a wrapped error. I don't know of anything that does that. And it seems weird for this code to expect or at least handle that case.

brycemcanally: Ahh, I see. Yeah, I'm not too sure. I think I would prefer to play it safe for the time being.

src/server/pkg/tarutil/util.go
brycemcanally: My thinking was that it would expose a way for a client to create a default header. We either don't have a way to create a default header for a client, or we do and we run the risk that they could incorrectly make modifications to the header. I think the latter is preferable.

brendoncarroll: Okay. This is fine.  I'm not even sure how much use it will get, since a header literal is about as simple.

src/server/pkg/tarutil/util.go
```diff
@@ -0,0 +1,251 @@
...
+	Header() (*tar.Header, error)
+	Content(io.Writer) error
+}
+
+type file struct {
+	hdr  *tar.Header
+	data []byte
+}
+
+func NewFile(name string, data []byte) File {
```
brendoncarroll, line 24: Maybe call this `NewMemFile` or `NewInMemoryFile`.  This shouldn't be used for large files.

src/server/pkg/tarutil/util.go
```diff
@@ -0,0 +1,251 @@
...
+	_, err := w.Write(f.data)
+	return err
+}
+
+type streamFile struct {
+	hdr *tar.Header
+	r   io.Reader
+}
+
+func NewStreamFile(name string, size int64, r io.Reader) File {
```
brendoncarroll, line 56: `NewFileFromReader` would probably be more discoverable.

src/server/pkg/tarutil/util.go
```diff
@@ -0,0 +1,251 @@
...
+		hdr: hdr,
+		r:   r,
+	}
+}
+
+func (sf *streamFile) Header() (*tar.Header, error) {
+	return sf.hdr, nil
+}
+
+func (sf *streamFile) Content(w io.Writer) error {
```
brendoncarroll, line 71: This would violate the idempotence contract of the similar method in the `fileset` package. I think we need to pick the behavior, document it in the interface, and have the stream file return an error if the behavior is at most 1 call.

src/server/pkg/tarutil/util.go
```diff
@@ -0,0 +1,251 @@
...
+			return err
+		}
+		if err := cb(newFile(hdr, buf.Bytes())); err != nil {
+			return err
+		}
+	}
+}
+
+func Equal(file1, file2 File, full ...bool) (bool, error) {
+	buf1, buf2 := &bytes.Buffer{}, &bytes.Buffer{}
```
brendoncarroll, line 128: use a `hash.Hash` for this.  It implements `io.Writer`.

src/server/pkg/tarutil/util.go
```diff
@@ -0,0 +1,251 @@
...
+	if err := file1.Content(buf1); err != nil {
+		return false, err
+	}
+	if err := file2.Content(buf2); err != nil {
+		return false, err
+	}
+	return bytes.Equal(buf1.Bytes(), buf2.Bytes()), nil
+}
+
+func TarToLocal(storageRoot string, r io.Reader, cb ...func(*tar.Header) error) error {
```
brendoncarroll, line 161: Maybe call this `ExportTar` or `UnpackTar`.

src/server/pkg/tarutil/util.go
```diff
@@ -0,0 +1,251 @@
...
+	defer func() {
+		if err := f.Close(); retErr == nil {
+			retErr = err
+		}
+	}()
+	_, err = io.Copy(f, r)
+	return err
+}
+
+func LocalToTar(storageRoot string, w io.Writer, cb ...func(*tar.Header) error) error {
```
brendoncarroll, line 207: Maybe `ImportTar` `PackTar`

brendoncarroll: Since it's the tarutil package, maybe just `Pack` and `Unpack` ?

src/server/pkg/tarutil/util.go
brycemcanally: I'll go with `NewMemFile`.

src/server/pkg/tarutil/util.go
brycemcanally: I would prefer to keep this consistent with the other constructor.

src/server/pkg/tarutil/util.go
brycemcanally: This was discussed offline, we are going to revisit this post alpha.

src/server/pkg/tarutil/util.go
brycemcanally: I would prefer to leave it as is for now. We can revisit this later if we want to use this for checking if large files are equal.

src/server/pkg/tarutil/util.go
brycemcanally: I agree with the import / export name change. This function seems like it should be named import though, so `tarutil.Import`.

src/server/pkg/tarutil/util.go
brycemcanally: Addressed above.

PR #5160, opened 2020-07-30
Fixes a problem where CompactSpec would return an empty output path

brendoncarroll:


src/server/pkg/storage/fileset/storage.go
```diff
@@ -166,7 +166,22 @@ type CompactSpec struct {
 // CompactSpec returns a compaction specification that determines the input filesets (the diff file set and potentially
 // compacted filesets) and output fileset.
 func (s *Storage) CompactSpec(ctx context.Context, fileSet string, compactedFileSet ...string) (*CompactSpec, error) {
+	if len(compactedFileSet) > 1 {
+		return nil, errors.WithStack(errors.Errorf("multiple compacted FileSets"))
```
brycemcanally, line 170: It looks like `Errorf` grabs the stack, so wrapping it with `WithStack` should be unnecessary.

src/server/pkg/storage/fileset/storage.go
```diff
@@ -272,7 +283,43 @@ func removePrefix(fileSet string) string {
 	return fileSet[len(prefix):]
 }
 
+func removePrefixes(xs []string) (ys []string) {
```
brycemcanally, line 286: Would you mind switching this to not using a named return. Although this is a simple case, unnecessarily using named returns can lead to subtle bugs. I think in general we should only be operating on named return variables when handling errors.

PR #5162, opened 2020-07-31
Get enterprise key from stdin

### LINKED ISSUE #5159: Enterprise activation should take key from stdin
**What is the goal / desired outcome?** We currently pass in enterprise tokens as an argument to `pachctl enterprise activate`. Since it's a secret, it should, similarly to passwords, be passed in via stdin. This will ensure it doesn't show up in, e.g. the shell history.

ysimonson:
On `pachctl enterprise activate`, capture the key via stdin rather than as an CLI argument. This will prevent the key from being leaked, e.g. in bash history.

Technically it's a breaking change since `pachctl enterprise activate`'s interface is changing. In practice, I don't think anyone automates `pachctl enterprise activate` though (especially since you only need to call it once every ~year), so it seems pretty innocuous to change.

Closes #5159


ysimonson:
@nitinjainsj Could you review? In particular see the PR description.

ysimonson:
It's failing due to an issue on master

PR #5164, opened 2020-07-31
Warn users when they use the `latest` tag

lukemarsden:
Attempts to solve https://github.com/pachyderm/pachyderm/issues/5114

Demo:

## latest tag case
```
luke@light:~/Projects/Pachyderm/pachyderm$ pachctl create pipeline <<EOF
            {
              "pipeline": {
                "name": "first"
              },
              "transform": {
                "image": "ubuntu:latest"
              },
              "input": {
                "pfs": {
                  "repo": "in",
                  "glob": "/*"
                }
              }
            }
EOF
Reading from stdin.
WARNING: please do not specify the ':latest' tag for the docker image in your
transform.image spec. For example, change 'python:latest' to 'python:3' or
'bash:latest' to 'bash:5'. This improves reproducibility of your pipelines.
```

## no tag case
```
luke@light:~/Projects/Pachyderm/pachyderm$ pachctl create pipeline <<EOF
            {
              "pipeline": {
                "name": "first"
              },
              "transform": {
                "image": "ubuntu"
              },
              "input": {
                "pfs": {
                  "repo": "in",
                  "glob": "/*"
                }
              }
            }
EOF
Reading from stdin.
WARNING: please specify a tag for the docker image in your transform.image spec.
For example, change 'python' to 'python:3' or 'bash' to 'bash:5'. This improves
reproducibility of your pipelines.
```

### DIFF
--- a/src/server/pps/cmds/cmds_test.go
+++ b/src/server/pps/cmds/cmds_test.go
@@ -835,17 +835,22 @@ func TestPipelineBuildMissingPath(t *testing.T) {
 // 	require.NoError(t, rootCmd().Execute())
 // }
 
-func TestWarningLatestTag(t *testing.T) {
-	if testing.Short() {
-		t.Skip("Skipping integration tests in short mode")
-	}
-	// should emit a warning because user specified latest tag on docker image
+func runPipelineWithImageGetStderr(t *testing.T, image string) string {
+	// reset and create some test input
+	require.NoError(t, tu.BashCmd(`
+		yes | pachctl delete all
+		pachctl create repo in
+		pachctl put file -r in@master:/ -f ../../../../etc/testing/pipeline-build/input
+	`).Run())
 
-	cmd := exec.Command("/bin/bash", "-c", `
+	cmd := exec.Command("/bin/bash", "-c", fmt.Sprintf(`
 		pachctl create pipeline <<EOF
 			{
+			  "pipeline": {
+				"name": "first"
+			  },
 			  "transform": {
-				"image": "ubuntu:latest"
+				"image": "%s"
 			  },
 			  "input": {
 			    "pfs": {
@@ -855,13 +860,22 @@ func TestWarningLatestTag(t *testing.T) {
 			  }
 			}
 EOF
-	`)
+	`, image))
 	buf := &bytes.Buffer{}
 	cmd.Stderr = buf
+	// cmd.Stdout = os.Stdout // uncomment for debugging
 	cmd.Env = os.Environ()
 	cmd.Run()
 	stderr := buf.String()
+	return stderr
+}
 
+func TestWarningLatestTag(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration tests in short mode")
+	}
+	// should emit a warning because user specified latest tag on docker image
+	stderr := runPipelineWithImageGetStderr(t, "ubuntu:latest")
 	if !strings.Contains(stderr, "WARNING") {
 		t.Logf(
 			"pachctl create pipeline failed to emit a WARNING to stderr, got %s",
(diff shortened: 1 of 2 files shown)


msteffen:
One small question regarding testing

src/server/pps/cmds/cmds_test.go
```diff
@@ -830,3 +834,85 @@ func TestPipelineBuildMissingPath(t *testing.T) {
...
+
+func runPipelineWithImageGetStderr(t *testing.T, image string) string {
+	// reset and create some test input
+	require.NoError(t, tu.BashCmd(`
+		yes | pachctl delete all
+		pachctl create repo in
+		pachctl put file -r in@master:/ -f ../../../../etc/testing/pipeline-build/input
+	`).Run())
+
+	cmd := exec.Command("/bin/bash", "-c", fmt.Sprintf(`
```
msteffen, line 846: My opinion on this is mixed, but I think the way to do this that's most consistent with the other command tests would be to use the `match` binary; something like:
```
tu.BashCmd(`
pachctl create pipeline <<EOF 2>&1 >/dev/null \
  | match "expected error"
...pipeline text...
EOF
`)
```

I think for consistency's sake it might be worth changing, but let me know what you think.

lukemarsden: Hmm.

I think I prefer the current approach for two reasons:

1. It asserts that the warning comes out of stderr, rather than "stdout or stderr". Warnings should come out of stderr ([ref](https://stackoverflow.com/questions/1430956/should-i-output-warnings-to-stderr-or-stdout#:~:text=For%20those%20reasons%2C%20in%20general%2C%20you%20output%20warnings%20to%20stderr.&text=The%20warning%20should%20go%20to,is%20an%20issue%20of%20flexibility.)), and it's nice that the tests assert that.
2. One of the tests wants to assert a negative - that there is no warning. That's (a) tricky to do with `match` and (b) would require pushing the logic up into the `exec.Command` command body which is currently factored out (and unfactoring this would make the tests overall have more duplication in them).

I'd be happy to move the "run this command and capture stderr" code into `testutil` though, as that would be more idiomatic than calling `exec.Command` directly from the test suite. What do you think?

I pushed a separate commit which makes `create pipeline` error fail the test by just checking for the error.

I don't feel super strongly about it, and if you'd prefer to have more consistency between tests by using the `match` tool I'm happy to update the tests to do that. Let me know 😄

lukemarsden:
CI is red due to an unrelated issue. Will retry once https://github.com/pachyderm/pachyderm/pull/5170 is merged.

src/server/pps/cmds/cmds_test.go
msteffen: I think it's fine to leave this as-is, your argument makes sense.

> I'd be happy to move the "run this command and capture stderr" code into testutil though, as that would be more idiomatic than calling exec.Command directly from the test suite. What do you think?

I think this is probably the right long-term answer, though I also don't think it makes sense to block this PR on it. I filed an issue for it here: https://github.com/pachyderm/pachyderm/issues/5174, in case we're motivated to work on it later.

src/server/pps/cmds/cmds_test.go
```diff
@@ -830,3 +834,97 @@ func TestPipelineBuildMissingPath(t *testing.T) {
...
+	return stderr, err
+}
+
+func TestWarningLatestTag(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration tests in short mode")
+	}
+	// should emit a warning because user specified latest tag on docker image
+	stderr, err := runPipelineWithImageGetStderr(t, "ubuntu:latest")
+	if err != nil {
```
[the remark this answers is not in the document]
lukemarsden: Nice, updated in https://github.com/pachyderm/pachyderm/pull/5164/commits/c5f4e240cad7f7fa8356942ff7bd7ef344775c70, makes it much more concise 👍

PR #5166, opened 2020-07-31
DiffFileV2

brendoncarroll:
- Adds DiffFileV2 rpc.
- DiffFileV2 uses same request as v1.
- DiffFileV2 returns a stream of DiffFileV2Responses which will be entries from either the new or the old commit or both. Each response will correspond to one path.  And responses will be streamed lexicographically.


src/server/pfs/server/diff.go
```diff
@@ -0,0 +1,93 @@
...
+	return &Differ{a: a, b: b}
+}
+
+func (d *Differ) IterateDiff(ctx context.Context, cb func(aFi, bFi *pfs.FileInfoV2) error) error {
+	ctx, cf := context.WithCancel(ctx)
+	defer cf()
+	aInfos := make(chan *pfs.FileInfoV2)
+	bInfos := make(chan *pfs.FileInfoV2)
+	eg, ctx := errgroup.WithContext(ctx)
+	// iterate over a
```
brycemcanally, line 25: These comments don't really add any value.

src/server/pkg/storage/fileset/reader.go
```diff
@@ -93,6 +94,10 @@ func (fr *FileReader) Header() (*tar.Header, error) {
 		}
 		var err error
 		fr.hdr, err = tar.NewReader(buf).Next()
+		if err == io.EOF {
```
brycemcanally, line 97: remove

src/server/pfs/server/diff.go
```diff
@@ -0,0 +1,93 @@
...
+// Differ compares two sources and iterates over the items that are not equal.
+type Differ struct {
+	a, b Source
+}
+
+func NewDiffer(a, b Source) *Differ {
+	return &Differ{a: a, b: b}
+}
+
+func (d *Differ) IterateDiff(ctx context.Context, cb func(aFi, bFi *pfs.FileInfoV2) error) error {
```
brycemcanally, line 19: Seems like calling this function `Iterate` would be more appropriate considering the abstraction is specifically for diffing.

PR #5168, opened 2020-08-01
Test spout ci trigger

adelelopez:


src/server/worker/pipeline/common.go
```diff
@@ -63,100 +220,43 @@ func ReceiveSpout(
...
+				continue
 			}
+			logger.Logf("Got data: ", n, buff.Len())
+			// send the contents to the tar reader
+			TarReader(buff, logger)
+			// and process any new files received
+			logger.Logf("Total files", fileCount)
+			processFilesReceived(ctx, pachClient, pipelineInfo, logger)
+		case notify.Remove:
+			logger.Logf("Fatal: pipe /pfs/out removed")
```
nitinjainsj, line 257: We should change this handling to reinitialize the pipe and get back into this loop. If we fail 3 consecutive times, that's when we should abort.

src/server/worker/pipeline/common.go
```diff
@@ -63,100 +220,43 @@ func ReceiveSpout(
...
-			}(); err != nil {
-				return err
+			if n == 0 {
+				continue
 			}
+			logger.Logf("Got data: ", n, buff.Len())
+			// send the contents to the tar reader
+			TarReader(buff, logger)
+			// and process any new files received
+			logger.Logf("Total files", fileCount)
```
nitinjainsj, line 254: We should change the log messages to be more meaningful. I had them to debug the loop.

src/server/worker/pipeline/common.go
```diff
@@ -54,6 +122,95 @@ func RunUserCode(
...
 
+func processFiles(ctx context.Context,
+	pachClient *client.APIClient,
+	pipelineInfo *pps.PipelineInfo,
+	logger logs.TaggedLogger) (retErr1 error) {
+	repo := pipelineInfo.Pipeline.Name
+	var commit *pfs.Commit
+	var err error
+	for _, f := range fileList {
+		if commit == nil {
```
nitinjainsj, line 133: I think this loop creates one commit for all the files received in on batch. Is that correct? IIRC, we want to create a commit for each tar file we receive.

src/server/worker/pipeline/common.go
```diff
@@ -19,20 +20,87 @@ import (
...
+	hdr  *tar.Header
+	body []byte
+}
+
+var fileList []TarFile
+var curr TarFile
+var fileCount int
+
+// SkipTarPadding return number of bytes skipped from a tar stream and an EOF
+func SkipTarPadding(r *bytes.Reader) (int64, error) {
```
nitinjainsj, line 38: We have our own forked version of tar file. We should check how we handle the padding bytes. It would be good to leverage our implementation.

src/server/worker/pipeline/common.go
```diff
@@ -19,20 +20,87 @@ import (
...
-	openAndWait, err := os.Open("/pfs/out")
-	if err != nil {
-		return err
+// TarFile holder structure
+type TarFile struct {
+	hdr  *tar.Header
+	body []byte
+}
+
+var fileList []TarFile
```
nitinjainsj, line 33: I had added the global variables for a quick implementation. As the implementation evolved, I realized we don't need the global state. This should be easy to fix.

src/server/pachyderm_test.go
```diff
@@ -10622,18 +10622,20 @@ func TestSpout(t *testing.T) {
...
 		require.NoError(t, err)
 
-		for i := 0; i < 10; i++ {
+		var files []*pfs.FileInfo
+		for i := 0; len(files) < 1000; i++ {
 			commitInfo, err := iter.Next()
 			require.NoError(t, err)
-			files, err := c.ListFile(pipeline, commitInfo.Commit.ID, "")
+			files, err = c.ListFile(pipeline, commitInfo.Commit.ID, "")
+			t.Logf("List files: [%v] [%d] %d", err, i+1, len(files))
```
nitinjainsj, line 10630: Don't need the log message here.

src/server/pachyderm_test.go
```diff
@@ -10622,18 +10622,20 @@ func TestSpout(t *testing.T) {
 		iter, err := c.SubscribeCommit(pipeline, "master", nil, "", pfs.CommitState_FINISHED)
 		require.NoError(t, err)
 
-		for i := 0; i < 10; i++ {
+		var files []*pfs.FileInfo
+		for i := 0; len(files) < 1000; i++ {
```
nitinjainsj, line 10626: Previously, the loop was going through 10 times, now it goes for as many commits as necessary. What changed in the behavior? In both cases, we should be checking for 1000 files.

src/server/worker/pipeline/common.go
adelelopez: yeah, this looks right

src/server/worker/pipeline/common.go
adelelopez: i checked, and it doesn't make any changes to how padding bytes are handled

also, that version is specialized for use in PFS storage internals, while here we need to be using a standard tar, so i don't think it's appropriate to use our fork here

src/server/worker/pipeline/common.go
adelelopez: ah yeah, that makes sense: i thought it looked weird but figured you had had a reason to do it that way

src/server/pachyderm_test.go
adelelopez: I assumed you increased this just to make the test more stringent. Not sure what you mean by 'changed in the behavior'.

PR #5178, opened 2020-08-05
InspectFileV2 and WalkFileV2

brendoncarroll:


src/server/pfs/server/api_server_v2.go
```diff
@@ -250,3 +244,16 @@ func (a *apiServerV2) GlobFileV2(request *pfs.GlobFileRequest, server pfs.API_Gl
...
+
+// InspectFileV2 returns info about a file.
+func (a *apiServerV2) InspectFileV2(ctx context.Context, req *pfs.InspectFileRequest) (*pfs.FileInfoV2, error) {
+	return a.driver.inspectFile(a.env.GetPachClient(ctx), req.File)
+}
+
+// WalkFileV2 walks over all the files under a directory, including children of children.
+func (a *apiServerV2) WalkFileV2(req *pfs.WalkFileRequest, server pfs.API_WalkFileV2Server) error {
+	pachClient := a.env.GetPachClient(server.Context())
+	return a.driver.walkFile(pachClient, req.File, func(finfo *pfs.FileInfoV2) error {
```
brycemcanally, line 256: nitpick, `finfo` -> `fi`.

src/server/pfs/server/driver.go
```diff
@@ -2837,6 +2837,10 @@ func (d *driver) scratchCommitPrefix(commit *pfs.Commit) string {
 }
 
 func (d *driver) checkFilePath(path string) error {
+	return checkFilePath(path)
```
brycemcanally, line 2840: It would probably be better to just change the call site to use the function without the driver receiver.

src/server/pfs/server/driver_v2.go
```diff
@@ -137,24 +137,31 @@ func (d *driverV2) finishCommitV2(txnCtx *txnenv.TransactionContext, commit *pfs
...
 
-// TODO Need commit validation and handling of branch names.
-func (d *driverV2) withFileSet(ctx context.Context, repo, commit string, f func(*fileset.FileSet) error) (retErr error) {
+func (d *driverV2) getSubFileSet() int64 {
 	// TODO subFileSet will need to be incremented through postgres or etcd.
 	d.mu.Lock()
-	subFileSetStr := fileset.SubFileSetStr(d.subFileSet)
+	n := d.subFileSet
+	d.subFileSet++
+	d.mu.Unlock()
```
brycemcanally, line 145: I would prefer this to be a defer for when we make changes to this later.

src/server/pfs/server/api_server_v2.go
```diff
@@ -250,3 +244,16 @@ func (a *apiServerV2) GlobFileV2(request *pfs.GlobFileRequest, server pfs.API_Gl
...
+
+// InspectFileV2 returns info about a file.
+func (a *apiServerV2) InspectFileV2(ctx context.Context, req *pfs.InspectFileRequest) (*pfs.FileInfoV2, error) {
+	return a.driver.inspectFile(a.env.GetPachClient(ctx), req.File)
+}
+
+// WalkFileV2 walks over all the files under a directory, including children of children.
+func (a *apiServerV2) WalkFileV2(req *pfs.WalkFileRequest, server pfs.API_WalkFileV2Server) error {
+	pachClient := a.env.GetPachClient(server.Context())
+	return a.driver.walkFile(pachClient, req.File, func(fin *pfs.FileInfoV2) error {
```
brycemcanally, line 256: `fin` -> `fi`

src/server/pfs/server/driver_v2.go
```diff
@@ -478,6 +473,53 @@ func (d *driverV2) globFileV2(pachClient *client.APIClient, commit *pfs.Commit,
 	})
 }
 
-func compactedCommitPath(commit *pfs.Commit) string {
-	return path.Join(commitKey(commit), fileset.Compacted)
+func (d *driverV2) inspectFile(pachClient *client.APIClient, file *pfs.File) (ret *pfs.FileInfoV2, retErr error) {
```
brycemcanally, line 476: The named returns are unnecessary here.

src/server/pfs/server/driver_v2.go
```diff
@@ -478,6 +473,53 @@ func (d *driverV2) globFileV2(pachClient *client.APIClient, commit *pfs.Commit,
...
+	p := cleanPath(file.Path)
+	s := NewSource(file.Commit, true, func() fileset.FileSource {
+		x := d.storage.NewSource(ctx, compactedCommitPath(file.Commit), index.WithPrefix(p))
+		x = fileset.NewIndexResolver(x)
+		x = fileset.NewIndexFilter(x, func(idx *index.Index) bool {
+			return idx.Path == p || strings.HasPrefix(idx.Path, p+"/")
+		})
+		return x
+	})
+	s = InsertNotFound(s, file)
```
brycemcanally, line 491: This seems like it would be better as a helper function, but naming it seems a bit weird (something like `RequireFileIterate` or `RequireNonEmptyIterate` might make sense?). Right now it just checks if nothing was iterated over, then returns the passed in file in an error if so. I think we need to pick either handling just an empty iteration or a particular file and checking for the file in the iteration (we don't check that the particular file was the one we iterated over).

src/server/pfs/server/testing/server_v2_test.go
```diff
@@ -640,6 +644,7 @@ func (v *validator) validate(r io.Reader) error {
 	for _, file := range filesSorted {
 		buf := &bytes.Buffer{}
 		if _, err := io.CopyN(buf, r, int64(len(v.files[file]))); err != nil {
+			fmt.Println("missing", file)
```
brycemcanally, line 647: remove

src/server/pfs/server/val_server.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/pachyderm/pachyderm/src/client/pfs"
+	"github.com/pachyderm/pachyderm/src/client/pkg/errors"
+	"github.com/pachyderm/pachyderm/src/client/pkg/grpcutil"
+	"github.com/pachyderm/pachyderm/src/server/pkg/serviceenv"
+	"golang.org/x/net/context"
+)
+
+var _ APIServer = &valAPIServer{}
+
+type valAPIServer struct {
```
brycemcanally, line 15: I think this should be `validateAPIServer` or `validatedAPIServer`, then the constructor should be named appropriately (`new<full abstraction name>`).

src/server/pkg/storage/fileset/index_resolver.go
```diff
@@ -49,5 +49,8 @@ func (mr *mergeResolver) Iterate(ctx context.Context, cb func(File) error, stopB
 	if err := mr1.WriteTo(w); err != nil {
 		return err
 	}
-	return w.Close()
```
brycemcanally, line 52: Why was this change made?

src/server/pkg/storage/fileset/fileset_test.go
```diff
@@ -123,7 +123,7 @@ func TestWriteThenRead(t *testing.T) {
 		for _, fileName := range fileNames {
 			data := chunk.RandSeq(rand.Intn(max))
 			files = append(files, &testFile{
-				name: fileName,
+				name: "/" + fileName,
```
brycemcanally, line 126: We shouldn't need this anymore right? Since we are canonicalizing and applying cleaning to requests.

src/server/pkg/storage/fileset/reader.go
```diff
@@ -91,9 +91,14 @@ func (fr *FileReader) Header() (*tar.Header, error) {
...
 			return nil, err
 		}
-		var err error
-		fr.hdr, err = tar.NewReader(buf).Next()
-		return fr.hdr, err
+		hdr, err := tar.NewReader(buf).Next()
+		if err != nil {
+			return nil, err
+		}
+		if !IsCleanTarPath(hdr.Name, hdr.FileInfo().IsDir()) {
```
brycemcanally, line 98: :+1:

src/server/pkg/storage/fileset/util.go
```diff
@@ -56,3 +65,20 @@ func WriteTarStream(ctx context.Context, w io.Writer, fs FileSource) error {
 	}
 	return tar.NewWriter(w).Close()
 }
+
+// CleanTarPath ensures that the path is in the canonical format for tar header names.
+// This includes ensuring a prepending /'s and ensure directory paths
+// have a trailing slash.
+func CleanTarPath(x string, isDir bool) string {
+	y := "/" + strings.Trim(x, "/")
```
brycemcanally, line 73: This is what `cleanPath` does right? Maybe `cleanPath` should be in the fileset package and exported, then used by this function?

brendoncarroll: Paths in the driver have to be the same whether they are directories or files, so that matching works.  We cannot call this function without `isDir` which we don't have at the usual call sites of `cleanPath`

src/server/pkg/storage/fileset/fileset_test.go
brendoncarroll: It validates based on what's in files.

src/server/pkg/storage/fileset/util.go
brycemcanally: I meant line 73 specifically. I think we could have a `CleanPath` function exposed by the fileset package that would get called at line 73.

brendoncarroll: I'm not sure that would get reuse in the driver. I think we may find that it makes the path matching in the driver simpler to work on paths without the leading slash.  I don't think there is a good reason to create a dependency here yet.

src/server/pfs/server/driver_v2.go
brycemcanally: Thoughts on this?

src/server/pfs/server/source.go
```diff
@@ -199,6 +205,32 @@ func computeFileHash(idx *index.Index) []byte {
 	return h.Sum(nil)
 }
 
+type errOnEmpty struct {
+	source Source
+	err    error
+}
+
+// InsertNotFound causes iterate to return a not found error if there are no items to iterate over
+func ErrOnEmpty(s Source, err error) Source {
```
brycemcanally, line 214: `NewErrOnEmpty` if you are sticking with an abstraction rather than a helper function. Also, change the comment or the linter won't let you pass.

Issue #4713, opened 2020-03-13
`pachctl port-forward` refuses to start after previous run was killed


Tryneus:
It appears that `pachctl port-forward` determines if port forwarding is already running by checking the `~/.pachyderm/config.json` file.  If the previous run was unceremoniously killed, the data may remain in the config file and a new run of `pachctl port-forward` will result in: 

```
$ pachctl port-forward
port forwarding appears to already be running for this context
```

The workaround is to delete the port-forwarder entry from the config file.

I think the usual way to solve this is using pidfiles.  We could probably do this without too much trouble using cooperatively-locked files, but there's the complication that we can port-forward any set of ports as long as it doesn't overlap with an existing port-forwarder.  We might need to support a set of port-forwarder pidfiles and refer to which ones are valid for which ports in the config file.  That or have a pidfile per forwarded port.  You figure it out.

ysimonson:
Agreed that pidfiles would be good. Do not try to lock the config file itself, because we move the file into place on save to ensure the operation is atomic, but that surrenders the lock.

lukemarsden:
Here is a terrible workaround in case it helps anyone else:
```
python3 -c '
import json
a = json.loads(open(".pachyderm/config.json").read())
if len(list(a.get("v2", {}).get("contexts", {}).items())) > 0 and "port_forwarders" in list(a.get("v2", {}).get("contexts", {}).items())[0][1]:
    del(list(a.get("v2", {}).get("contexts", {}).items())[0][1]["port_forwarders"])
    open(".pachyderm/config.json", "w").write(json.dumps(a, indent=4))
'
```

reergymerej:
Thanks @lukemarsden for the path.

PR #5192, opened 2020-08-13
Add cooperative lockfile as port-forward mutex

### LINKED ISSUE #5182: Stale port-forward config causes errant error
I was trying to start port-forwarding to my pachub cluster, and got: `port forwarding appears to already be running for this context`

Looking in `src/server/cmd/pachctl/cmd/cmd.go` I saw this error is given if any PortForwarders are configured in the config. However, I had no port forwarders running on my system, and yet they were still configured in the config. Runtime state != config state on disk.

Manually deleting that section from my `~/.pachyderm/config` file (`port-forwards` stanza I think? not sure, I deleted it) made `pachctl port-forward` work again. But maybe we should check to see if the system is _actually_ forwarding the ports in the error checking there, rather than if it's just configured to? How is it currently supposed to work?

Tryneus:
Puts a lockfile in the config directory that prevents multiple `pachctl port-forward`s from running for the same context concurrently.  This is more reliable than keeping track of `port-forward` in the config file, as we can maintain a file-lock that is scoped to the lifetime of the locking process rather than relying on an orderly shutdown to clean up the config file.  There is a separate lockfile per-context.

This fixes #5182 and #4713.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -51,6 +51,7 @@ require (
 	github.com/jehiah/go-strftime v0.0.0-20171201141054-1d33003b3869 // indirect
 	github.com/jinzhu/gorm v1.9.12
 	github.com/juju/ansiterm v0.0.0-20180109212912-720a0952cc2a
+	github.com/juju/fslock v0.0.0-20160525022230-4d5c94c67b4b
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/lib/pq v1.3.0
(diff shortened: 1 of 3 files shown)


ysimonson:
This makes sense as a strategy. `src/client/client.go` will also need updating (grep for `PortForwarders` there.)

msteffen:

Issue #5141, opened 2020-07-24
Build-enabled pipelines do not work in hub


ysimonson:
(I'm not sure if this is a core issue or hub issue.)

**What happened?**: Build-step enabled pipelines do not work in hub. This is because they first create a repo, then create a pipeline on that existing repo, and this triggers a permissions error - even though this works fine on clusters deployed elsewhere.

**How to reproduce it (as minimally and precisely as possible)?**: Run `examples/word_count` on hub. It will trigger this error when trying to create the map pipeline:

```
failed to create build pipeline for build step-enabled pipeline: cannot overwrite repo "map_build" with new output repo
```

msteffen:
Sorry, this issue was robo-closed by GitHub, but there's another change we need to make in CreatePipeline before this is fixed

ysimonson:
That's my bad! I attached the issue to the PR thinking it was definitely going to fix it.

PFedak:
(Should have been closed by #5594, whoops)

PR #5193, opened 2020-08-13
CreateRepo with update=True no longer resets repo ACL

msteffen:


msteffen:
@nitinjainsj this is the `create_python_pipeline` issue from slack
@brycemcanally I wasn't sure if you or Marc were the right person to review this, so feel free to add yourself or just start commenting if you have thoughts

ysimonson:
This will also fix #5141

ysimonson:
And it might fix #5018!

msteffen:
@ysimonson :+1: I'm hopeful about #5018. For #5141, I think there *might* be a change in CreatePipeline that's necessary as well, but I'm not sure—I'll test it separately and see

src/server/auth/server/testing/auth_test.go
```diff
@@ -742,6 +742,90 @@ func TestGetSetReverse(t *testing.T) {
...
+	// bob can read
+	buf.Reset()
+	require.NoError(t, bobClient.GetFile(dataRepo, "master", "/file", 0, 0, buf))
+	require.Equal(t, "1", buf.String())
+
+	/// alice updates the repo
+	_, err = aliceClient.PfsAPIClient.CreateRepo(aliceClient.Ctx(), &pfs.CreateRepoRequest{
+		Repo:   client.NewRepo(dataRepo),
+		Update: true,
+	})
```
Tryneus, line 787: This takes the shortcut out of the update because the description hasn't changed - I think you should set a non-trivial description here just to be a bit more complete.

msteffen: :+1:

PR #5197, opened 2020-08-13
[1.11] Fix python build pipelines when there are no dependencies

ysimonson:


ysimonson:
Closing and re-opening because it appears switching the base branch broke TravisCI triggers

PR #5199, opened 2020-08-13
client: allow the client to use interceptors

jrockway:
This adds two new client options; `WithAdditionalUnaryClientInterceptor` and `WithAdditionalStreamClientInterceptor`.  gRPC client interceptors can be cleanly chained, so users of the library can add their own interceptors, and everything we do involving Jaeger tracing still works.

I considered letting callers add any arbitrary dial option, but that removes our ability to chain Jaeger on to any interceptor-related options that they explicitly pass in.  (Once they add an interceptor-related dial option, we lose the ability to see the interceptor they're adding and chain it onto interceptors that we want to add.)  And, arbitrary dial options can make problems with the client harder to debug -- they can override behaviors that we add with our own dial options.  So because of that, and because I don't need it, I decided to just stick with interceptors.  It's the minimum change required, at the cost of it being confusing in the future when people can do `WithAdditionalUnaryClientInterceptor(i)` or `WithAdditionalDialOptions(grpc.UnaryClientInterceptor(i))`.  But because of the issues I mentioned, that future seems unlikely :)

### DIFF
--- a/src/client/client.go
+++ b/src/client/client.go
@@ -146,6 +146,8 @@ type clientSettings struct {
 	dialTimeout          time.Duration
 	caCerts              *x509.CertPool
 	storageV2            bool
+	unaryInterceptors    []grpc.UnaryClientInterceptor
+	streamInterceptors   []grpc.StreamClientInterceptor
 }
 
 // NewFromAddress constructs a new APIClient for the server at addr.
@@ -155,7 +157,7 @@ func NewFromAddress(addr string, options ...Option) (*APIClient, error) {
 		return nil, errors.Errorf("address shouldn't contain protocol (\"://\"), but is: %q", addr)
 	}
 	// Apply creation options
-	settings := clientSettings{
+	settings := &clientSettings{
 		maxConcurrentStreams: DefaultMaxConcurrentStreams,
 		dialTimeout:          DefaultDialTimeout,
 	}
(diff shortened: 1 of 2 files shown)


jrockway:
I am still unsure about how one is supposed to run this test.  I used `go test -run=TestInterceptors` to develop it.  The Makefile seems to suggest that CI tests with `go test -count=1 -cover $(go list ./src/client/...)`, and that does select this new test.  But before it can run, the not-supposed-to-work `godoc_test.go` is selected and fails, preventing this test from running.  I must be missing something or maybe there's a better location for this new test file.

jrockway:
@msteffen As discussed on Thursday, I've "unrefactored" `connect`, and now the diff is much easier to read.

msteffen (approved):
iloveyouiknowlgtm

Using the UnimplementedServer to test the client is brilliant, btw. It's a little embarrassing that we had no tests in that package, but this is great, and we can use it for other stuff too.

jrockway:
Are we OK with the Travis failures?  I don't think they're related to this PR, but I'm always hesitant to hit the merge button when CI is failing.

msteffen:
> Are we OK with the Travis failures? I don't think they're related to this PR, but I'm always hesitant to hit the merge button when CI is failing.
Probably not, though I think as a matter of principle we try to get it green before merging. I think one of the extract/restore tests hung for some reason, which we should look into separately, but for now I just restarted the test shard and hopefully it works this time

jrockway:
Looks like things are passing, so I'm going to merge.

Do I need to update the changelog or anything like that for this?

PR #5200, opened 2020-08-14
Use ReadPassword in auth commands

msteffen:


msteffen:
Background question: why does ReadPassword print the prompt to stderr? Is it a security thing, and if so, should the auth cmds be doing it elsewhere?

ysimonson:
That was just an oversight when refactoring code. There's no security implications that I'm aware of, though I could see the argument that interactive prompts should go to stderr so as to not mess with formatted output to stdout.

PR #5203, opened 2020-08-14
Adds DeleteCommit, DeleteRepo, and DeleteAll

brendoncarroll:
Also removes BuildCommit

### DIFF
--- a/src/server/pfs/server/driver_v2.go
+++ b/src/server/pfs/server/driver_v2.go
@@ -16,6 +16,7 @@ import (
 	"github.com/pachyderm/pachyderm/src/client/pkg/errors"
 	pfsserver "github.com/pachyderm/pachyderm/src/server/pfs"
 	"github.com/pachyderm/pachyderm/src/server/pkg/backoff"
+	col "github.com/pachyderm/pachyderm/src/server/pkg/collection"
 	"github.com/pachyderm/pachyderm/src/server/pkg/hashtree"
 	"github.com/pachyderm/pachyderm/src/server/pkg/serviceenv"
 	"github.com/pachyderm/pachyderm/src/server/pkg/storage/chunk"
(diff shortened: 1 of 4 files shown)


src/server/pfs/server/driver_v2.go
```diff
@@ -524,3 +520,310 @@ func (d *driverV2) walkFile(pachClient *client.APIClient, file *pfs.File, cb fun
...
+	ctx := txnCtx.ClientContext
+	if err := d.storage.WalkFileSet(ctx, repo.Name, func(p string) error {
+		return d.storage.Delete(ctx, p)
+	}); err != nil {
+		return err
+	}
+	return d.driver.deleteRepo(txnCtx, repo, force)
+}
+
+func (d *driverV2) deleteAll(txnCtx *txnenv.TransactionContext) error {
```
brycemcanally, line 534: What is going on with the auth stuff in this? Is `includeAuth` a global variable, and why do we filter out auth errors on the `deleteRepo` call?

brendoncarroll: I'm not sure this is all just copied from the existing driver, with calls to the new storage layer added in.

src/server/pkg/storage/fileset/storage.go
```diff
@@ -269,10 +269,13 @@ func (s *Storage) compactSpec(ctx context.Context, fileSet string, compactedFile
 func (s *Storage) Delete(ctx context.Context, fileSet string) error {
 	fileSet = applyPrefix(fileSet)
 	return s.objC.Walk(ctx, fileSet, func(name string) error {
+		if err := s.objC.Delete(ctx, name); err != nil {
```
brycemcanally, line 272: I think we are eventually going to want this to be handled by a tombstone so we have some way to handle failing to delete the semantic reference. Also, you can just return the result of the `DeleteSemanticReference` call.

PR #5204, opened 2020-08-14
V2 client and deploy changes

brycemcanally:
This PR adds functionality for creating a V2 client using the V1 interface and makes some changes to the V2 deployment process. Some V1 functionality is not supported by this conversion, but in most cases we should fail fast when a client attempts to use the unsupported functionality.


src/server/pfs/server/source.go
```diff
@@ -1,93 +1,18 @@
...
 	"github.com/pachyderm/pachyderm/src/server/pkg/storage/fileset"
 	"github.com/pachyderm/pachyderm/src/server/pkg/storage/fileset/index"
-	"github.com/pachyderm/pachyderm/src/server/pkg/tar"
 	"golang.org/x/net/context"
 )
 
-// FileReader is a PFS wrapper for a fileset.MergeReader.
-// The primary purpose of this abstraction is to convert from index.Index to
-// pfs.FileInfo and to convert a set of index hashes to a file hash.
-type FileReader struct {
```
brendoncarroll, line 20: 👍

src/server/pfs/server/source.go
```diff
@@ -119,38 +44,50 @@ func (s *source) Iterate(ctx context.Context, cb func(*pfs.FileInfo, fileset.Fil
...
+				fi.FileType = computedFi.FileType
+				fi.SizeBytes = computedFi.SizeBytes
+				fi.Hash = computedFi.Hash
 			}
-			fi.Hash = hashBytes
-		}
-		if err := cb(fi, fr); err != nil {
-			return err
 		}
-		delete(cache, idx.Path)
```
brendoncarroll, line 144: Doesn't lexicographical order ensure that a parent is emitted before it's child?  By the time anything is emitted it's FileInfo is no longer needed by any path coming later.
Why can't we delete here?

src/server/pfs/server/val_server.go
```diff
@@ -19,16 +19,61 @@ type validatedAPIServer struct {
...
-func newValidatedAPIServer(inner APIServer, env *serviceenv.ServiceEnv) *validatedAPIServer {
+func newValidatedAPIServer(embeddedServer APIServer, env *serviceenv.ServiceEnv) *validatedAPIServer {
 	return &validatedAPIServer{
-		APIServer: inner,
+		APIServer: embeddedServer,
 		env:       env,
 	}
 }
 
+// DeleteRepoInTransaction is identical to DeleteRepo except that it can run
```
brendoncarroll, line 29: 👍

src/server/pfs/server/source.go
brycemcanally: Yes, but for the subsequent child files we need to check if the parent directory info is cached. This allows us to correctly avoid progressing the second stream when it has already been progressed past a file. When we compute a directory info we progress the second stream past all of the files / directories in that directory.

brycemcanally: We can delete a directory when we are past all of its files / directories, I just need to put more thought into how we handle that. I'm not too concerned about this caching for now, which is why I left it as a TODO.

PR #5212, opened 2020-08-18
Minor typo and wording changing

tantanchen:
I understand that changing this phrase is debatable, so I won't be offended if you don't accept this change. I'm proposing this change because I misunderstood the wording. All throughout this tutorial, I am getting started, so my mental state is one where I am installing the Pachyderm code. So to me, the phrase "your Pachyderm directory" was very easily misunderstood to mean the folder where Pachyderm resides. I think this can be easily avoided by using the term _your working directory_. Some other options are _your project directory_ and _your working directory/project directory_.

### DIFF
--- a/doc/docs/1.10.x/getting_started/local_installation.md
+++ b/doc/docs/1.10.x/getting_started/local_installation.md
@@ -155,7 +155,7 @@ deploy Pachyderm by following these steps:
   1. From the same directory, run:
 
      ```bash
-     kubectl create -f .\pachyderm.json
+     kubectl create -f ./pachyderm.json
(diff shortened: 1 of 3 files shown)


JoeyZwicker:
Thanks for the PR @tantanchen. I like the phrase change and it resolves a minor confusion that we've seen users have on rare occasion. I thought I saw your CLA in my inbox, but for some reason our CLA bot isn't detecting it. hmm. 

If it's not too much trouble, would you mind try signing it again with this link? https://cla-assistant.io/pachyderm/pachyderm?pullRequest=5212

PR #5222, opened 2020-08-24
Add "everyone" subject to ACLs

actgardner:
Similar to the `allAuthorizedUsers` IAM subject, this would allow a user to express an ACL like "everyone in the company can read this repo" without actively maintaining the set of users in the ACL, or group membership.

I think this + default ACLs for new repos would solve a significant pain point.

### DIFF
--- a/src/server/auth/server/api_server.go
+++ b/src/server/auth/server/api_server.go
@@ -44,7 +44,7 @@ const (
 	// pachyderm token for any username in the AuthenticateRequest.GitHubToken field
 	DisableAuthenticationEnvVar = "PACHYDERM_AUTHENTICATION_DISABLED_FOR_TESTING"
 
-	everyoneSubject = "everyone"
+	allClusterUsersSubject = "allClusterUsers"
 
 	tokensPrefix           = "/tokens"
 	oneTimePasswordsPrefix = "/auth-codes"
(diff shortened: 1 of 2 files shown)


src/server/auth/server/api_server.go
```diff
@@ -44,6 +44,8 @@ const (
 	// pachyderm token for any username in the AuthenticateRequest.GitHubToken field
 	DisableAuthenticationEnvVar = "PACHYDERM_AUTHENTICATION_DISABLED_FOR_TESTING"
 
+	everyoneSubject = "everyone"
```
msteffen, line 47: Re: the name `everyone`, do you think it would make sense to use a narrower name, in case we want a principal similar to GCP's `allUsers`? (for totally public resources, e.g. publicly-hosted buckets, where you don't have to be logged in to see it). Like `allUsers` vs `allClusterUsers`?

I also wonder if there might be demand for a version of `allAuthenticatedUsers` that means "users with an account in the IdP, but not an account in Pachyderm"? Like "people using one of our enterprise key seats can be individually on ACLs and do stuff with private data, but the set of general employees who don't have a Pachyderm seat can still look at the usage reports hosted in Pachyderm".

actgardner: `allClusterUsers` seems like it leaves room for future, more broadly-scoped groups.

PR #5228, opened 2020-08-26
Remove 1.8 command compatibility layer

Tryneus:
This has been around with deprecation warnings for several releases now, we can probably remove it soon.  We may want to hold off on this for the new storage layer so we can do all the incompatible changes at once.


msteffen:

PR #5239, opened 2020-08-31
[WIP] Use slashes in tags for worker artifacts

Tryneus:
Slow `ListTags` performance can cause jobs to block in the worker while cleaning up job artifacts.  Some object stores may have better performance when the tag prefix contains slashes, so this PR changes the prefix for worker job artifacts, and also moves the cleanup to happen after succeeding the job in the job chain to avoid blocking parallel jobs.

In local testing, using tags with slashes only caused issues in the `localcache` package, which uses the tag name as a filename (and there was no logic to provision directory trees).  Avoided this problem by url-encoding the tag name.

This should not be merged as-is, as the impacts on all object stores is not well-understood.  Therefore, this should be behind a feature-flag that must be added, and enabling or disabling this flag may cause spurious failures in any currently-in-progress jobs due to the tag naming scheme changing.

### DIFF
--- a/src/server/cmd/worker/main.go
+++ b/src/server/cmd/worker/main.go
@@ -83,7 +83,7 @@ func do(config interface{}) error {
 
 	// Construct worker API server.
 	workerRcName := ppsutil.PipelineRcName(pipelineInfo.Pipeline.Name, pipelineInfo.Version)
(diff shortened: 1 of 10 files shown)


msteffen:

PR #5253, opened 2020-09-08
Adds endpoints for temporary filesets

brendoncarroll:


src/client/pfs/pfs.proto
```diff
@@ -561,6 +561,16 @@ message DiffFileResponseV2 {
   FileInfo new_file = 2;
 }
 
+message CreateTempFileSetResponse {
```
brycemcanally, line 564: Just a nitpicky thing, but I have been using `tmp` as a shorthand for temporary and it would be nice to be consistent about this in PFS / PPS.

src/server/pfs/server/api_server_v2.go
```diff
@@ -314,3 +318,26 @@ func (a *apiServerV2) DeleteAll(ctx context.Context, request *types.Empty) (resp
...
+	})
+}
+
+func (a *apiServerV2) RenewTempFileSet(ctx context.Context, req *pfs.RenewTempFileSetRequest) (*pfs.CreateTempFileSetResponse, error) {
+	ttl := time.Duration(req.ExtendSeconds) * time.Second
+	t, err := a.driver.renewTempFileSet(ctx, req.FilesetId, ttl)
+	if err != nil {
+		return nil, err
+	}
+	return &pfs.CreateTempFileSetResponse{
```
brycemcanally, line 339: Is this information useful to return? If so, it seems like it should be in a different message type because this is not the `CreateTempFileSet` rpc.

src/server/pfs/server/driver_v2.go
```diff
@@ -70,7 +72,11 @@ func newDriverV2(
 	if err != nil {
 		return nil, err
 	}
-	chunkStorageOpts := append([]chunk.StorageOption{chunk.WithGarbageCollection(gcClient)}, chunk.ServiceEnvToOptions(env)...)
+	chunkOptsSrv, err := chunk.ServiceEnvToOptions(env)
```
brycemcanally, line 75: `chunkOptsSrv` -> `chunkStorageOpts` (just reassign below).

src/server/pkg/storage/fileset/option.go
```diff
@@ -52,6 +54,12 @@ func WithMaxOpenFileSets(max int) StorageOption {
 // UWriterOption configures an UnorderedWriter.
```
brycemcanally, line 54: I would prefer that we spell this out here, so `UWriterOption` -> `UnorderedWriterOption`. I think we will probably want to minimize the abbreviation of configuration.

src/client/pfs_v2.go
```diff
@@ -40,8 +41,10 @@ func (c APIClient) DeleteFilesV2(repo, commit string, files []string, tag ...str
 // FileOperationClient is not thread safe. Multiple FileOperationClients
 // should be used for concurrent upload.
 type FileOperationClient struct {
```
brycemcanally, line 43: So I think we should probably tweak this design a bit. I think the basic idea of reusing the `FileOperationV2` functionality makes sense, and I think there is probably more we can refactor on the server side to really make this clean (essentially just have the interpreting of the operations / setting up of the unordered writer in helper functions that both rpcs use). The server side changes can come later, but I think we want to go ahead and tweak this client stuff a bit, because the closing stuff is pretty messy. I think we could set this up such that the `FileOperationV2` and `CreateTempFileSet` client wrappers each have their own abstraction which embeds something that functions similarly to the `FileOperationClient`. The key change though would be that the embedded abstraction would just take a client and be able to send `FileOperationRequestV2`s, but it would have no notion of closing the client. The outer abstraction would have a close method attached to it that would handle the closing differences between the two rpcs. Therefore, we could hand back a client for each rpc that could perform all of the operations as well as have the correct closing semantics.

src/client/pfs/pfs.proto
```diff
@@ -561,6 +561,16 @@ message DiffFileResponseV2 {
...
+  string fileset_id = 1;
+  int64 expires_at = 2;
```
brycemcanally, line 566: After looking over this a bit more, I'm not sure if returning the expire time is that useful and it requires quite a bit of threading through the storage layers. I guess you could use this information to handle renewal for the rare case where a response is really delayed, but I think most use cases will probably just have a fixed renewal based on preexisting knowledge of the default ttl.

src/client/pfs_v2.go
```diff
@@ -267,3 +292,31 @@ func (r *getTarConditionalReader) drain() error {
...
+		return "", err
+	}
+	res, err := foc.closeRecvTempFileSet()
+	if err != nil {
+		return "", err
+	}
+	return res.FilesetId, nil
+}
+
+const TempRepoName = "__tmp__"
```
brycemcanally, line 315: Since we have a separate endpoint, I don't think this is necessary on the client side. Also, it looks like the below function is unused.

src/client/pfs/pfs.proto
```diff
@@ -561,6 +561,16 @@ message DiffFileResponseV2 {
...
+}
+
+message RenewTempFileSetRequest {
+  string fileset_id = 1;
+  int64 extend_seconds = 2;
```
brycemcanally, line 571: I think you may have mentioned this when we discussed this, but why is this not just interpreted as a duration as compared to number of seconds?

src/server/pkg/testpachd/mock_pachd.go
```diff
@@ -788,6 +796,18 @@ func (api *pfsServerAPI) WalkFileV2(req *pfs.WalkFileRequest, serv pfs.API_WalkF
 	}
 	return errors.Errorf("unhandled pachd mock pfs.WalkFileV2")
 }
+func (api *pfsServerAPI) CreateTempFileSet(srv pfs.API_CreateTempFileSetServer) error {
+	if api.mock.CreateTempFileSet.handler != nil {
+		return api.mock.CreateTempFileSet.handler(srv)
+	}
+	return errors.Errorf("unhandled pachd mock pfs.WalkFileV2")
```
brycemcanally, line 803: Wrong rpc.

src/server/pfs/server/driver_v2.go
```diff
@@ -961,3 +970,45 @@ func (d *driverV2) deleteCommit(txnCtx *txnenv.TransactionContext, userCommit *p
...
+	// if err := d.txnEnv.WithWriteContext(ctx, func(txnCtx *transactionenv.TransactionContext) error {
+	// 	return d.finishCommitV2(txnCtx, commit, "")
+	// }); err != nil {
+	// 	return "", nil, err
+	// }
+	return commit.ID, expiresAt, nil
+}
+
+func (d *driverV2) renewTempFileSet(ctx context.Context, id string, ttl time.Duration) (*time.Time, error) {
+	if ttl > maxTTL {
```
brycemcanally, line 1009: It might be better to surface this as an error.

src/server/pkg/storage/chunk/storage.go
```diff
@@ -64,15 +69,42 @@ func (s *Storage) CreateSemanticReference(ctx context.Context, name string, chun
 	return s.gcClient.CreateReference(ctx, semanticReference(name, chunk.Hash))
 }
 
+// CreateSemanticReference creates a semantic reference to a chunk.
+func (s *Storage) CreateTemporaryReference(ctx context.Context, name string, chunk *Chunk, ttl time.Duration) (*time.Time, error) {
```
brycemcanally, line 73: Why are we passing around pointers to a `time.Time`? The Go recommendation is to pass by value for this type: https://golang.org/pkg/time/#Time

src/server/pkg/storage/chunk/storage.go
```diff
@@ -39,7 +44,7 @@ func (s *Storage) NewReader(ctx context.Context, dataRefs ...*DataRef) *Reader {
 // Chunks are created based on the content, then hashed and deduplicated/uploaded to
 // object storage.
 func (s *Storage) NewWriter(ctx context.Context, tmpID string, f WriterFunc, opts ...WriterOption) *Writer {
-	return newWriter(ctx, s.objClient, s.gcClient, tmpID, f, opts...)
+	return newWriter(ctx, s.objClient, s.gcClient, tmpID, f, append(opts, WithChunkTTL(defaultChunkTTL))...)
```
brycemcanally, line 47: I think the order is wrong here. You want the default option first, then the client supplied options afterwards.

src/server/pfs/server/driver_v2.go
```diff
@@ -961,3 +970,45 @@ func (d *driverV2) deleteCommit(txnCtx *txnenv.TransactionContext, userCommit *p
 
 	return nil
 }
+
+func (d *driverV2) createTempFileSet(server pfs.API_CreateTempFileSetServer) (string, *time.Time, error) {
+	ctx := server.Context()
+	opts := []fileset.UWriterOption{fileset.WithWriterOption(fileset.WithTTL(30 * time.Minute))}
+	id := uuid.NewWithoutDashes()
+	var commit *pfs.Commit
+	// TODO: within commit?
```
brycemcanally, line 979: I think we wanted it in a commit so that the rpcs that access files would work out of the box. Is the concern around the cleanup of the commit? I guess that would not be captured in this model. I think we are eventually going to have a background process that cleans up commits for data retention purposes and maybe it could be what catches the commits associated with cleaned up temporary filesets?

src/client/pfs/pfs.proto
brendoncarroll: I like the lease renewal algorithm that works as follows: Get a lease, check when it expires, schedule a renewal halfway to the expire time, repeat.  This allows the client to know if it has a lease that it can't work with (because it expires too quickly).

src/client/pfs/pfs.proto
brendoncarroll: Do you mean like using the well-known-type for duration? Depending on how protoc is setup those can be a pain, sometimes requiring another conversion in the code.  We don't really need sub-second resolution for leases, since they will typically be on the order of minutes or at lease tens of seconds.  If we use the duration types elsewhere, I can switch to those.

src/client/pfs_v2.go
brendoncarroll: This is used to construct the commit for the File methods. Instead of having a separate set of functions for commits, we have a sentinel repo that causes the commit id to be interpreted as a temp fileset id.

src/client/pfs/pfs.proto
```diff
@@ -652,6 +661,10 @@ service API {
   rpc InspectFileV2(InspectFileRequest) returns (FileInfo) {}
   // WalkFileV2 walks over all the files under a directory, including children of children.
   rpc WalkFileV2(WalkFileRequest) returns (stream FileInfo) {}
+  // CreateTempFileSet creates a new temp fileset
```
brycemcanally, line 664: These comments need to be updated.

src/client/pfs_v2.go
```diff
@@ -58,11 +58,69 @@ func (c APIClient) NewFileOperationClientV2(repo, commit string) (_ *FileOperati
...
+	})
+}
+
+// CreateTmpFileSetClient is used to create a temporary fileset.
+type CreateTmpFileSetClient struct {
+	client pfs.API_CreateTmpFileSetClient
+	fileOperationCore
+}
+
+func (c APIClient) NewCreateTmpFileSetClient() (ret *CreateTmpFileSetClient, retErr error) {
```
brycemcanally, line 83: `ret` -> `_`

src/client/pfs_v2.go
```diff
@@ -58,11 +58,69 @@ func (c APIClient) NewFileOperationClientV2(repo, commit string) (_ *FileOperati
...
+	return &CreateTmpFileSetClient{
+		client: client,
+		fileOperationCore: fileOperationCore{
+			client: client,
+		},
+	}, nil
+}
+
+// Close closes the FileOperationClient.
+func (foc *CreateTmpFileSetClient) CloseAndRecv() (*pfs.CreateTmpFileSetResponse, error) {
```
brycemcanally, line 100: I think `Close` is actually a better name for this method. The gRPC aspect should stay encapsulated in my opinion. Also, update the comment appropriately.

brycemcanally: Also `foc` is not really the appropriate variable name. The name for this abstraction might be a bit too wordy.

src/server/pfs/server/api_server.go
```diff
@@ -638,6 +638,16 @@ func (a *apiServer) WalkFileV2(req *pfs.WalkFileRequest, server pfs.API_WalkFile
 	return errV2NotImplemented
 }
 
+// CreateTempFileSet not implemented
```
brycemcanally, line 641: Update comments.

src/client/pfs_v2.go
```diff
@@ -267,3 +317,13 @@ func (r *getTarConditionalReader) drain() error {
 		}
 	}
 }
+
+const TempRepoName = "__tmp__"
+
+// TmpFileSetCommit
```
brycemcanally, line 323: Comment?

src/server/pfs/server/driver_v2.go
```diff
@@ -165,10 +171,13 @@ func (d *driverV2) withUnorderedWriter(ctx context.Context, repo, commit string,
 	if err := fs.Close(); err != nil {
 		return err
 	}
-	return d.compactionQueue.RunTaskBlock(ctx, func(m *work.Master) error {
+	if err := d.compactionQueue.RunTaskBlock(ctx, func(m *work.Master) error {
```
brycemcanally, line 174: Just return here.

src/client/pfs_v2.go
brycemcanally: `TempRepoName` -> `TmpRepoName`

src/server/pfs/server/driver_v2.go
brycemcanally: Bump

src/server/pfs/server/driver_v2.go
```diff
@@ -961,3 +970,45 @@ func (d *driverV2) deleteCommit(txnCtx *txnenv.TransactionContext, userCommit *p
...
+	if len(id) != 32 {
+		return errors.Errorf("invalid id (%s)", id)
+	}
+	p := path.Join(tmpRepo, id)
+	_, err := d.storage.RenewFileSet(ctx, p, ttl)
+	return err
+}
+
+func (d *driverV2) inspectCommit(pachClient *client.APIClient, commit *pfs.Commit, blockState pfs.CommitState) (*pfs.CommitInfo, error) {
+	if commit.GetRepo().GetName() == tmpRepo {
```
brycemcanally, line 1005: What is going on here? Why do we return a commit info, shouldn't we just error?

src/server/pkg/storage/gc/client.go
```diff
@@ -24,10 +33,12 @@ var (
 //  * Source - the source of the reference, this may be a temporary id, chunk id, or a
 //    semantic name.
 //  * Chunk - the target chunk being referenced.
+//  * ExpiresAt - the time after which a temporary chunk will be destroyed
 type Reference struct {
-	Sourcetype string
+	Sourcetype SourceType
 	Source     string
 	Chunk      string
+	ExpiresAt  *time.Time
```
brycemcanally, line 41: Why is this a pointer?

brycemcanally: It looks like gorm is operating on a pointer in some queries and a struct in others.

src/server/pfs/server/driver_v2.go
brendoncarroll: We want to call inspect commit on all the accessors before proceeding.  This creates a phony commit info so the accessors can proceed as usually without any changes.

src/client/pfs_v2.go
```diff
@@ -58,11 +58,69 @@ func (c APIClient) NewFileOperationClientV2(repo, commit string) (_ *FileOperati
...
+	}
+	return &CreateTmpFileSetClient{
+		client: client,
+		fileOperationCore: fileOperationCore{
+			client: client,
+		},
+	}, nil
+}
+
+// Close closes the FileOperationClient.
```
brycemcanally, line 99: `FileOperationClient` -> `CreateTmpFileSetClient`

src/server/pkg/storage/fileset/storage.go
```diff
@@ -287,6 +318,92 @@ func (s *Storage) WalkFileSet(ctx context.Context, prefix string, f func(string)
...
+		defer cancel()
+		return cb(ctx2)
+	})
+	if err := eg.Wait(); err != nil {
+		return err
+	}
+	return s.Delete(ctx, prefix)
+}
+
+func (s *Storage) GetExpiresAt(ctx context.Context, prefix string) (time.Time, error) {
```
brycemcanally, line 380: Is this dead code? I cannot find a call site?

src/server/pkg/storage/fileset/storage.go
```diff
@@ -287,6 +318,92 @@ func (s *Storage) WalkFileSet(ctx context.Context, prefix string, f func(string)
...
+	if expiresAt == nil {
+		return time.Time{}, ErrNoFileSetFound
+	}
+	return *expiresAt, nil
+}
+
+// RenewDuring ensures that none of the filesets under prefix expire by repeatedly calling SetTTL in a separate goroutine
+// until cb returns.
+// RenewDuring attempts to delete the contents of prefix after cb has returned
+func (s *Storage) RenewDuring(ctx context.Context, prefix string, cb func(context.Context) error) error {
```
brycemcanally, line 348: I don't think this model is going to work for us. I think we need to be able to associate one `expiresAt` with the prefix rather than one for each individual path under that prefix. When we move to postgres, we will be cleaning up the semantic path, so I can imagine a scenario where a path under the prefix expires and is deleted, but we don't catch that it was deleted during the renewal since the renewal just looks at the existing paths. Or, we could just be explicit about which paths to renew, but I kind of like the model where we are renewing on a prefix.

src/server/pkg/storage/fileset/storage.go
```diff
@@ -287,6 +318,92 @@ func (s *Storage) WalkFileSet(ctx context.Context, prefix string, f func(string)
 	})
 }
 
+// SetTTL sets the time-to-live for the prefix. If the prefix contains multiple filesets
+// the ttl will be set on all of them and the soonest expiration time will be returned.
+// if no fileset is found SetTTL returns ErrNoFileSetFound
+func (s *Storage) SetTTL(ctx context.Context, prefix string, ttl time.Duration) (time.Time, error) {
```
brycemcanally, line 324: I think we should be a bit more consistent with the naming conventions for this stuff. So, if we need a function like this (I think we may not if we implement renewal on a prefix), then something like `RenewFileSet` seems more consistent and the last parameter would be `expiresAt`. Also, the scoped function below could be `WithRenewFileSet`.

src/server/pkg/storage/fileset/storage.go
```diff
@@ -287,6 +318,92 @@ func (s *Storage) WalkFileSet(ctx context.Context, prefix string, f func(string)
...
+	}
+	return *expiresAt, nil
+}
+
+// RenewDuring ensures that none of the filesets under prefix expire by repeatedly calling SetTTL in a separate goroutine
+// until cb returns.
+// RenewDuring attempts to delete the contents of prefix after cb has returned
+func (s *Storage) RenewDuring(ctx context.Context, prefix string, cb func(context.Context) error) error {
+	const defaultTTL = 10 * time.Minute
+	eg, ctx2 := errgroup.WithContext(ctx)
```
brycemcanally, line 350: I tend to name derived contexts with this pattern `*Ctx`. So, this could be `errCtx`. I think that is a bit more informative.

src/server/pkg/storage/fileset/storage.go
```diff
@@ -287,6 +318,92 @@ func (s *Storage) WalkFileSet(ctx context.Context, prefix string, f func(string)
...
+	ctx2, cancel := context.WithCancel(ctx2)
+	eg.Go(func() error {
+		var expiresAt time.Time
+		var err error
+		for {
+			expiresAt, err = s.SetTTL(ctx2, prefix, defaultTTL)
```
brycemcanally, line 356: I don't think this is the algorithm we want for two reasons: it is not as simple as renewing at a fixed period, and it will actually result with more requests when high latency is encountered which is probably not the appropriate behavior for the system as a whole. I think we should look at this as more of a cooperative process, and error out appropriately when high latency is encountered (so a renewal will eventually fail in this case).

src/server/pfs/server/driver_v2.go
```diff
@@ -144,53 +143,52 @@ func (d *driverV2) getSubFileSet() int64 {
...
+}
 
-	defer func() {
-		if err := d.storage.Delete(ctx, path.Join(tmpPrefix, subFileSetPath)); retErr == nil {
-			retErr = err
-		}
-	}()
-	if err := cb(fs); err != nil {
-		return err
+func (d *driverV2) withTempUnorderedWriter(ctx context.Context, compact bool, cb func(*fileset.UnorderedWriter) error) (string, error) {
```
brycemcanally, line 159: `Temp` -> `Tmp`

src/server/pfs/server/driver_v2.go
```diff
@@ -144,53 +143,52 @@ func (d *driverV2) getSubFileSet() int64 {
...
 	n := d.getSubFileSet()
 	subFileSetStr := fileset.SubFileSetStr(n)
 	subFileSetPath := path.Join(repo, commit, subFileSetStr)
-	fs, err := d.storage.New(ctx, path.Join(tmpPrefix, subFileSetPath), subFileSetStr)
+	id, err := d.withTempUnorderedWriter(ctx, false, cb)
 	if err != nil {
 		return err
 	}
+	tmpPath := path.Join(tmpRepo, id)
+	return d.storage.Copy(ctx, tmpPath, subFileSetPath, defaultTTL)
```
brycemcanally, line 156: Why do the output filesets have a ttl? The output filesets should be permanent.

brycemcanally:
Take another pass over the expires at code, particularly the code for propagating it back up the layers. I think some of it is dead code after the removal of the expires at from the grpc response.

src/server/pkg/storage/gc/client.go
brendoncarroll: In our model it's optional. When interacting with the database we use `sql.NullTime`

src/server/pkg/storage/fileset/storage.go
brendoncarroll: Not used, but I think it's useful to expose this.  The API is pretty straightforward. I think it may be a good idea to have tests to ensure commits do not have a TTL associated with them.  Maybe even do that as part of a set of live integrity checks.

src/server/pkg/storage/fileset/storage.go
brendoncarroll: This has changed to take a list of paths, each of which will be passed to `SetTTL`

src/server/pkg/storage/fileset/storage.go
brendoncarroll: We discussed this offline. `RenewDuring` -> `WithRenewal` to be consistent using `With*` for callback style APIs.  `SetTTL` is more literally what's going on, and could be used to set a TTL on a key without one.  After postgres this will no longer talk to the garbage collector client.

src/server/pkg/storage/fileset/storage.go
brendoncarroll: This is changed to a simple ticker. And due to the new semantics of `WithRenewal` we now detect slowness with not exist errors.

src/server/pfs/server/driver_v2.go
brendoncarroll: Oops, that would be bad.

src/server/pfs/server/driver_v2.go
```diff
@@ -28,15 +27,16 @@ import (
...
 const (
-	// tmpPrefix is for temporary storage paths.
-	// TODO Temporary prefix cleanup needs some work.
-	// Paths should get cleaned up in the background.
-	tmpPrefix            = "tmp"
 	storageTaskNamespace = "storage"
+	tmpRepo              = client.TmpRepoName
+	maxTTL               = 30 * time.Minute
+	defaultTTL           = 10 * time.Minute
+	driftMargin          = defaultTTL / 4
```
brycemcanally, line 39: This seems to be unused.

src/server/pkg/storage/fileset/storage.go
brycemcanally: I think we should remove it for the time being, because it would need to be changed anyway to accurately reflect the new model. I think most people would interpret this function as returning the expires at for a prefix, which is not how the garbage collection expires at works. Since it operates on independent paths, some paths may be removed much later than the returned time (and some paths might not even be removed at all if they don't have an expires at). I agree that integrity checks will be useful, but I don't think we should keep this around as is just because it could technically work to check if a commit has an expires at in any subfileset.

src/server/pfs/server/driver_v2.go
```diff
@@ -144,53 +144,60 @@ func (d *driverV2) getSubFileSet() int64 {
...
 	subFileSetStr := fileset.SubFileSetStr(n)
 	subFileSetPath := path.Join(repo, commit, subFileSetStr)
-	fs, err := d.storage.New(ctx, path.Join(tmpPrefix, subFileSetPath), subFileSetStr)
+	id, pathsWritten, err := d.withTmpUnorderedWriter(ctx, false, cb)
 	if err != nil {
 		return err
 	}
+	// collect paths for renewal
+	tmpPath := path.Join(tmpRepo, id)
+	return d.storage.WithRenewal(ctx, defaultTTL, pathsWritten, func(ctx context.Context) error {
```
brycemcanally, line 158: I don't think this is going to accomplish what we want. A long upload could still result with temporary subfilesets getting deleted since we wouldn't renew until the upload is finished. I was thinking about this a bit, and I think a model we could use that would work across the board would be to have a `Renewer` abstraction that could be passed around and you would be able to make calls on it to kick off renewals for new paths. Then you could have a `Close` function on it that would cancel all of the renewals that have been setup. This means you could setup the `Renewer` in the outer scope and defer the close to be sure that all the renewal goroutines are cleaned up (it might even make sense to just have a `WithRenewer` function). So, how this would work in the unordered writer case is that a `Renewer` would be setup in the outer scope (`withCommitWriter` in this case) and the unordered writer would take the `Renewer` as a parameter and add new paths to renew as it serializes subfilesets. Then when the outer scope is finished, all the renewals that were setup are canceled. For compaction the parent node would setup a `Renewer` for the child nodes and pass it in. Then when the children finish and the parent compacts them, the `Renewer` would be closed which would cancel all of the renewals.

brycemcanally (approved):
lgtm

PR #5256, opened 2020-09-09
Chunk cache for the storage layer

brendoncarroll:


src/server/pkg/obj/cache_client.go
```diff
@@ -0,0 +1,192 @@
...
+		// lru.NewWithEvict only errors for size < 1
+		panic(err)
+	}
+	client.cache = cache
+	return client
+}
+
+func (c *cacheClient) Reader(ctx context.Context, p string, offset, size uint64) (io.ReadCloser, error) {
+	c.doPopulateOnce(ctx)
+	if offset > 0 || size > 0 {
```
brycemcanally, line 45: Why does this need to go through the slow client? We cannot get part of a chunk from the fast client? If we end up leaning fully on the cache client for lexicographical reads (rather than caching a chunk in the chunk reader), then we are going to want these types of reads to go to the fast client.

brendoncarroll: Right now the storage layer only gets full chunks so that path is not used.  We don't want to cache pieces of chunks, that greatly complicates things, so in the case of a cache miss we want to get the whole thing from slow, and then get the fragment from fast. I don't think that would be too difficult to add. I can look into it if we pick this vs. the object caching.

brycemcanally: Yeah, it would not be used now, but I could see it being used later. Also, I'm not suggesting caching pieces of chunks. I am suggesting being able to read parts of full chunks that are cached. Our access patterns are lexicographical, so caching parts of chunks would probably be less performant even if we implemented it well.

brendoncarroll: Actually if we always gzip the chunks, it might never be used because you can't seek into a gzip stream.

src/server/pkg/obj/util.go
```diff
@@ -28,3 +30,21 @@ func WithLocalClient(f func(objC Client) error) (retErr error) {
 	}
 	return f(objC)
 }
+
+// Copy copys an object from src at srcPath to dst at dstPath
+func Copy(ctx context.Context, src, dst Client, srcPath, dstPath string) error {
```
brycemcanally, line 35: :+1:

src/server/pkg/obj/cache_client.go
brendoncarroll: This should work now.

src/server/pkg/obj/cache_client.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (c *cacheClient) IsNotExist(err error) bool {
+	return c.fast.IsNotExist(err) || c.slow.IsNotExist(err)
+}
+
+func (c *cacheClient) IsRetryable(err error) bool {
+	return c.fast.IsRetryable(err) || c.slow.IsRetryable(err)
+}
+
+// deleteFromCache should only be called with c.mu
```
brycemcanally, line 94: Why not just lock / unlock the mutex in this function?

src/server/pkg/obj/util.go
```diff
@@ -28,3 +30,21 @@ func WithLocalClient(f func(objC Client) error) (retErr error) {
...
+	rc, err := src.Reader(ctx, srcPath, 0, 0)
+	if err != nil {
+		return err
+	}
+	defer rc.Close()
```
brycemcanally, line 40: I think we should capture error state from the reader / writer close. So, they would look like this:
```
defer func() {
  if err := rc.Close(); retErr == nil {
    retErr = err
  }
}()
```
Also, this would allow you to just return the err from the `io.Copy`, rather than close. I am a bit unsure about whether closing the writer makes sense when an error occurs. I think there was a similar concern when I was working on closing the index writer, and what I went with was always closing with the expectation that the client writer would handle errors that occur while copying (so closing shouldn't store an incomplete state). Here is the code for reference: https://github.com/pachyderm/pachyderm/blob/e5b5944fdc533b6b98516e38d33d17f38193b2a2/src/server/pkg/storage/fileset/index/writer.go#L157

src/server/pkg/obj/cache_client.go
```diff
@@ -0,0 +1,149 @@
...
+		return err
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.deleteFromCache(ctx, p)
+}
+
+func (c *cacheClient) Exists(ctx context.Context, p string) bool {
+	c.doPopulateOnce(ctx)
+	return c.fast.Exists(ctx, p) || c.slow.Exists(ctx, p)
```
brycemcanally, line 75: I think we should only be checking the slow client here. We depend on this to determine if we should upload a chunk or not. (We reserve a chunk with gc to prevent deletion, then use the existence check to determine if we need to upload, so a stale cache entry could be problematic here). Really, I think we should primarily be leaning on the fast client for read operations. The error type checks probably make sense as is though.

src/server/pkg/obj/cache_client.go
```diff
@@ -0,0 +1,149 @@
...
+	c.cache.Add(p, struct{}{})
+	return c.fast.Reader(ctx, p, offset, size)
+}
+
+func (c *cacheClient) Writer(ctx context.Context, p string) (io.WriteCloser, error) {
+	wc, err := c.slow.Writer(ctx, p)
+	if err != nil {
+		return nil, err
+	}
+	return &cacheWriter{WriteCloser: wc, path: p, client: c}, nil
```
brycemcanally, line 61: I don't think the cache invalidation here is going to get us much because writes / deletes can be performed by other Pachd instances and we will not be overwriting an existing chunk. I think we should probably only use this caching client for chunks because stale reads are much less / not problematic in that narrow use case. If the other parts of our system are working correctly, then stale reads should not occur since the chunks should exist if we are trying to read them. In the case where they don't (in the slow storage), we should expect to eventually error when reading a deleted chunk in the slow storage or just succeed as if the data was not deleted yet (I think this would primarily come into play with temporary file sets).

brendoncarroll: I agree that this is terrible general purpose object cache.  But this at least allows Read-Your-Writes behavior.  If this is living in the `obj` package I think we should do the best we can.

src/server/pkg/obj/util.go
brendoncarroll: This makes sense. Notably wc does not have an "abort" alternative to `Close`. So if the reader errors during the copy, we still call `Close` even though we haven't copied all the data.  The client will get an error and can retry, I suppose. We do have to close the writer though, it can't be left open. We opened it, and if it is backed by the `LimitedClient` then it is consuming one of the allowed active writer slots.

src/server/pkg/obj/cache_client.go
brendoncarroll: We always write directly to the slow layer (write through). Anything that makes it into the fast layer, had to have come from the slow layer.  We call Exists a lot, and it also benefits from reduced latency. I think we should keep it as is.

brycemcanally: A different Pachd instance could have uploaded the chunk being checked. We cannot depend on the cache for this check. This is operating on full chunks, so the reduced latency does not get us much and could result with an additional write if the chunk exists. The real gain from the cache will be doing many small reads from a chunk (multiple small files in a chunk).

src/server/pkg/obj/cache_client.go
brycemcanally: I think we should just appropriately document the intended use of this caching client to make clear the limitations. I don't think cache invalidation should be supported in any fashion by this client to keep it simple. I don't think we want to get in  the territory of kind of supporting cache invalidation, but not in the general sense.

src/server/pkg/obj/cache_client.go
brendoncarroll: If a different pachd uploaded the chunk then `fast` returns false and `slow` is evaluated.  So if we don't have the chunk in fast we check both.  There is no reason not to do this; It occasionally saves a `slow.Exists` and it isn't even an extra line of code.

brycemcanally: Whoops, my bad. The problem is with deletes not uploads. If a Pachd instance deletes a chunk and it is cached in a different Pachd instance, then the Pachd instance with it cached will think it exists in the slow storage when it does not.

brycemcanally (approved):
lgtm

PR #5257, opened 2020-09-09
Adds Key-Value package `kv` with interface and cache

brendoncarroll:


src/server/pkg/storage/kv/cache.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (c *WriteThroughCache) Delete(ctx context.Context, key string) error {
+	c.mu.Lock()
+	if err := c.deleteFromCache(ctx, key); err != nil {
+		c.mu.Unlock()
+		return err
+	}
+	c.mu.Unlock()
+	return c.slow.Delete(ctx, key)
```
brendoncarroll, line 78: Order here is wrong, should be like Put

src/server/pkg/storage/kv/cache.go
```diff
@@ -0,0 +1,103 @@
...
+		panic(err)
+	}
+	return c
+}
+
+func (c *WriteThroughCache) GetF(ctx context.Context, p string, cb func([]byte) error) error {
+	c.mu.Lock()
+	if _, exists := c.cache.Get(p); exists {
+		c.mu.Unlock()
+		if err := c.fast.GetF(ctx, p, cb); err != nil {
```
brycemcanally, line 38: Is this a safe operation? It would require that the backing store either blocks a delete on a full read, or appropriately errors if the full read cannot be performed. If that is not the case, then we probably need a path lock similarly to the other implementation. If it is safe, then the other implementation probably does not need the path lock.

brendoncarroll: I think you are correct. It does have the same problem as the other implementation.  But in this case "full read" is guaranteed to be as quick as possible because it's into memory.  Whereas a full read in the other implementation can take arbitrarily long, if there is a slow reader. Improper use (failure to close the reader) would cause the fast layer to grow unbounded.  In this implementation the only advantage is that the backing store doesn't have to track open readers.  Once the content is in a buffer you can forget about it.

I looked more into how files work on Linux: deleting a file does not seem to cause readers with pre-existing file descriptors to error.
https://stackoverflow.com/questions/2028874/what-happens-to-an-open-file-handle-on-linux-if-the-pointed-file-gets-moved-or-d
So that should work for our use case.  I'm not sure if all the object stores also work this way. It's difficult to find documentation about this specific case.

PR #5260, opened 2020-09-10
[WIP] Add default ACL for new repos

actgardner:
Add the concept of a cluster-wide "default ACL" which is applied to new repos at creation time. This allows for use cases like "all users can read all repos by default" or "the engineering group has write access to everything". It would be even better with namespaces or some sort of hierarchy ;)


src/server/auth/server/api_server.go
```diff
@@ -253,6 +257,14 @@ func NewAuthServer(
 			nil,
 			nil,
 		),
+		defaultAcl: col.NewCollection(
```
actgardner, line 260: I considered making this a special-case in the existing ACLs collection (using some name that's not a valid repo for example) but decided it's better to be explicit and more verbose.

src/server/auth/server/api_server.go
```diff
@@ -1812,7 +1920,10 @@ func (a *apiServer) SetACLInTransaction(
 			}
 			aclMu.Lock()
 			defer aclMu.Unlock()
-			newACL.Entries[principal] = scope
+			// If the list includes multiple scopes for one principal, choose the highest
+			if scope > newACL.Entries[principal] {
```
actgardner, line 1924: This makes SetACL deterministic if the same user is supplied multiple times with different scopes. This allows us to append an entry to the default ACL to grant the repo creator OWNER.

src/server/auth/server/api_server.go
```diff
@@ -1866,8 +1977,7 @@ func (a *apiServer) SetACLInTransaction(
 		} else if !strings.HasSuffix(err.Error(), "not found") {
 			// Unclear if repo exists -- return error
 			return false, errors.Wrapf(err, "could not inspect \"%s\"", req.Repo)
-		} else if len(newACL.Entries) == 1 &&
-			newACL.Entries[callerInfo.Subject] == auth.Scope_OWNER {
+		} else if newACL.Entries[callerInfo.Subject] == auth.Scope_OWNER {
```
actgardner, line 1980: AFAICT this is belt-and-suspenders and it's safe to remove the check that the new ACL length is 1 when creating an ACL for a new repo.

src/server/auth/server/api_server.go
msteffen: This makes sense to me—maybe it would be helpful to note that this is for the new repo case (i.e. so that appending `<creater, OWNER>` works) in the comment?

Also, maybe this would be a good opportunity to get rid of the errgroup here and just do all of this serially? The thinking in making it parallel was that `canonicalizeSubject` can send an RPC to GitHub, but I know Go maps aren't thread-safe (using one here was always a mistake) and between that and GitHub not being popular anyway, it seems like the parallelism here might be net-negative value

src/server/auth/server/api_server.go
msteffen: Basically yes. The thinking was that ideally only PFS itself would be able to create an ACL for a new repo, and `SetACL` would just require the repo to exist in all cases. This was basically trying to limit what's possible in the weird case that an end-user (possibly malicious) is directly calling `SetACL` on a repo that doesn't exist yet.

Right now, setting an ACL for a nonexistant repo mostly doesn't do anything because CreateRepo calls `SetACL` with an ACL that it constructs from scratch, but this check was a backup for CreateRepo's behavior, basically.

I guess an attack would be possible if someone changed PFS in the future to use `SetScope(<creator>, OWNER)` instead of `SetACL` in CreateRepo. Then you could imagine an attack where:
1. Mallory calls `SetACL(future, {Mallory: OWNER, Alice: OWNER})`, which succeeds b/c there's no repo.
2. Alice calls `CreatePipeline(future) -> CreateRepo(future) -> SetScope(future, Alice, OWNER)`
3. Even though the ACL exists, Alice passes the permission check. Then Mallory can see the pipeline output even though Alice never added them to the ACL, and stays on it basically until someone calls `GetACL` and notices.

src/server/auth/server/api_server.go
```diff
@@ -1707,6 +1720,101 @@ func (a *apiServer) GetScope(ctx context.Context, req *auth.GetScopeRequest) (re
...
+				return err
+			}
+		}
+		if existing.Entries == nil {
+			existing.Entries = make(map[string]auth.Scope)
+		}
+		if req.Scope == auth.Scope_NONE {
+			delete(existing.Entries, principal)
+		} else {
+			existing.Entries[principal] = req.Scope
```
msteffen, line 1767: I know I always nitpick over names (sorry!) but FWIW the semantics here look more like the `SetScope` API than the `SetACL` API. Do you think it would be worth naming it something like `SetDefaultScope`?

msteffen (commented):
Thanks for being patient, and sorry for the slow reply! I didn't notice the PR, but I did just get a chrome extension that will hopefully make it easier for me to stay on top of these.

I had a few questions, along with one high-level design thought: if we proceed with trying to copy the GCP model for Pachyderm, then in the future it would be possible to give principals `READER`/`WRITER`/`OWNER` roles on the cluster, which would transitively apply to repos under the cluster. The thought is: 1) do you think it would make sense to provide a version of this feature by implementing that (i.e. extending `{Get,Modify}ClusterRoleBinding` to allow `READER`/`WRITER`/`OWNER` roles), and 2) if not, is there anything that it would make sense to start doing so that the features don't collide? E.g. if someone is removed from the admins, should we warn if they're still on the default ACL?

actgardner:
On reflection I like the concept of cluster-wide ACLs better than this, shelving it.

PR #5283, opened 2020-09-21
Setup Postgres in CI for Storage V2 Tests

brendoncarroll:


src/server/pfs/server/master_v2.go
```diff
@@ -21,20 +21,24 @@ const (
 func (d *driverV2) master(env *serviceenv.ServiceEnv, objClient obj.Client, db *gorm.DB) {
 	ctx := context.Background()
 	masterLock := dlock.NewDLock(d.etcdClient, path.Join(d.prefix, masterLockPath))
-	err := backoff.RetryNotify(func() error {
-		masterCtx, err := masterLock.Lock(ctx)
-		if err != nil {
+	for {
```
brycemcanally, line 24: This shouldn't need a for loop wrapping it. If `Run` is returning nil, then that is a problem that needs to be fixed (I vaguely remember this being an issue that had to do with the context cancel not being propagated).

src/server/pkg/testpachd/real_env.go
```diff
@@ -40,6 +41,18 @@ const (
 // environment in order to spin up pipelines, which is not yet supported by this
 // package, but the other API servers work.
 func WithRealEnv(cb func(*RealEnv) error, customConfig ...*serviceenv.PachdFullConfiguration) error {
+	// TODO: This means StorageV2 tests can not run in parallel with V1 test.
```
brycemcanally, line 44: Should we be unsetting the environment variable? I think this would prevent us from running V2 tests in parallel with other V2 tests. If we setup the V2 tests to run on their own machine, then we would just need to set it.

src/server/pkg/serviceenv/service_env.go
```diff
@@ -107,7 +107,11 @@ func (env *ServiceEnv) initPachClient() error {
 	// Initialize pach client
 	return backoff.Retry(func() error {
 		var err error
-		env.pachClient, err = client.NewFromAddress(env.pachAddress)
+		var opts []client.Option
```
brycemcanally, line 110: This is unnecessary for the time being, and will just be removed after 1.12. Although I love the functional option pattern, I don't think we should bother setting it up for the V2 config since the configuration will be removed after 1.12. The current approach for enabling V2 is by propagating the `STORAGE_V2` environment variable to each container and having the client check for this at creation time. I think this pattern will accomplish what we need in the short term, and will be easier to remove when V2 is no longer a configuration.

src/server/pkg/testpachd/real_env.go
brycemcanally: I see, PFS V1 and V2 tests are running on the same machine. Not running them in parallel is probably fine since they are the faster tests.

etc/contributing/run_postgres.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
```
brycemcanally, line 1: What is this used for? Just debugging?

etc/testing/start_postgres.sh
```diff
@@ -0,0 +1,14 @@
...
+
+if ! docker ps | grep -q postgres
+then
+    echo "starting postgres..."
+    docker run -d \
+    -e POSTGRES_USER=pachyderm \
+    -e POSTGRES_DB=pgc \
+    -e POSTGRES_HOST_AUTH_METHOD=trust \
+    -p 32228:5432 \
+    postgres:latest
```
brycemcanally, line 11: Using the latest might get us in trouble at some point. Maybe we should be explicit about a version.

src/server/pfs/server/testing/server_v2_test.go
```diff
@@ -694,14 +690,10 @@ func finfosToPaths(finfos []*pfs.FileInfo) (paths []string) {
...
-	// TODO: remove once postgres runs in CI
-	if os.Getenv("CI") == "true" {
-		t.SkipNow()
-	}
 	config := newPachdConfig()
-	require.NoError(t, os.Setenv("STORAGE_V2", "true"))
+	config.StorageV2 = true
 	require.NoError(t, testpachd.WithRealEnv(func(env *testpachd.RealEnv) error {
-		repo := "test"
+		repo := "TestListFileV2"
```
brycemcanally, line 696: It probably makes sense to go ahead and switch these to tests to create a unique string for the repo name.

src/server/pfs/server/master_v2.go
brendoncarroll: Yeah I thought this was fixed.  `gc.Run` should never return right?

src/server/pkg/serviceenv/service_env.go
brendoncarroll: We need to be able to pass this programmatically, and not rely on a global variable, if we want the tests to run concurrently. That is not an issue now.

src/server/pfs/server/testing/server_v2_test.go
brendoncarroll: The repo name changes away from just `test` ended up not being necessary.  We get a new environment for every test.

etc/contributing/run_postgres.sh
brendoncarroll: It's nice to have postgres without minikube being spun up.  It also logs invalid queries/statements, which is more useful than the errors you get in Go from the driver.

etc/testing/start_postgres.sh
brendoncarroll: They just released 13. Want to commit to that?

src/server/pfs/server/master_v2.go
brendoncarroll: Discussed offline. Will add a TODO here. This may be related to the environment being torn down.

brycemcanally (approved):
lgtm

PR #5290, opened 2020-09-24
Added a note about data persistence between datums

RaananHadar:
This addresses the docs issue in #4988.


JoeyZwicker:
Thanks for another great PR @RaananHadar. I think the datum leakage note is great. 

The stats stuff covered is a good start, but we'll ultimately want to outline quite a bit more information there to fully satisfy #5262. I'm happy to merge this PR for now, but will keep that issue open as something we'll follow on more later.

JoeyZwicker (approved):

PR #5291, opened 2020-09-24
Run Pipeline fixes

adelelopez:
This fixes a bug in Run Pipeline in which the provenance isn't properly deduplicated in certain cases. It also makes it easier to use by allowing the commit provenance to be input in a more flexible way. It adds tests for both the CLI change and the bug fix.


jdoliner (approved):
LGTM

PR #5292, opened 2020-09-24
Support OIDC ID tokens in Authenticate

actgardner:
This PR allows users who have OIDC configured to exchange an ID token issued by their OIDC server for a pachyderm token using the Authenticate API. To ensure this is secure the ID token:
- must be signed with a valid key from the OIDC server's JWKS endpoint
- must not be expired or used before the not-before date
- must include the pachd client_id in the audience claim

In particular the audience claim is important to [prevent replay attacks of bearer tokens](https://www.keycloak.org/docs/4.8/server_admin/#_audience). The resulting pachyderm token's lifespan is the minimum of the default TTL or the ID token expiration. 

The use case for authenticating with an ID token is: when a user wants to authenticate in the browser to an app, and in turn have that app make calls to Pachyderm on their behalf. Examples include dash and notebooks. The flow looks like:
- the user does a normal OAuth flow to a trusted app (an app which is allowed, in the IDP config, to request tokens with the pachyderm audience)
- the app recieves an access token and id token which include pachyderm in the audience claim
- the app then makes an RPC to exchange the id token for a pachyderm token to make calls on behalf of the user
- the app may cache the id token and resulting pachyderm token in a secure cookie until they expires
- at expiration the user re-authenticates (this may be entirely transparent) and the app recieves a new set of tokens which can be exchanged for a new pachyderm token

This flow is supported by both dex and keycloak, I haven't looked at other OIDC providers.


src/server/auth/server/api_server.go
```diff
@@ -1115,7 +1115,55 @@ func (a *apiServer) Authenticate(ctx context.Context, req *auth.AuthenticateRequ
...
 		}
+	case req.IdToken != "":
+		// confirm OIDC has been configured and get OIDC prefix
+		oidcSP := a.getOIDCSP()
+		if oidcSP == nil {
+			return nil, errors.Errorf("error authorizing OIDC id token: no OIDC ID provider is configured")
+		}
+
+		// Determine caller's Pachyderm/OIDC user info (email)
+		claims, err := oidcSP.ValidateJWT(req.IdToken)
```
msteffen, line 1126: I think our OIDC library has a built-in function for ID token validation, though it could be that it doesn't work for this case. If you haven't seen it, though, I think we use it in `src/server/auth/server/oidc.go:374`:
```
var verifier = sp.Provider.Verifier(&oidc.Config{ClientID: conf.ClientID})
// Extract the ID Token from OAuth2 token.
rawIDToken, ok := tok.Extra("id_token").(string)
if !ok {
	return "", "", errors.New("missing id token")
}

// Parse and verify ID Token payload.
idToken, err := verifier.Verify(ctx, rawIDToken)
if err != nil {
	return "", "", errors.Wrapf(err, "could not verify token")
}
```

Would it make sense to use this (or some version of it) here?

src/server/auth/server/oidc.go
```diff
@@ -151,6 +172,78 @@ func (a *apiServer) NewOIDCSP(name, issuer, clientID, clientSecret, redirectURI
...
+			return k.Key, nil
+		}
+	}
+	return nil, fmt.Errorf("no matching key for kid %q alg %q", kid, alg)
+}
+
+// ValidateJWT verifies a provided JWT using the JWKS endpoint in the OIDC config.
+// It also checks that the ClientID is in the audience and the token is not
+// expired or being used before it's valid.
+func (o *InternalOIDCProvider) ValidateJWT(token string) (*JWTClaims, error) {
```
msteffen, line 207: See comment in api_server.go about `go-oidc` having a version of this

msteffen (commented):
Just a few questions, with the main one being about whether we could use `go-oidc` for ID token validation.

Also, one high-level question: what's the context where a user would use this flow? E.g. in the step:
> the app then makes an RPC to exchange the id token for a pachyderm token to make calls on behalf of the user

Is the app here Dex? Does Dex make Pachyderm calls on behalf of the user? Or is it Jupyter or something? Or Dash?

src/server/auth/server/api_server.go
actgardner: Yeah this was an oversight on my part. The oidc-go library looks like it has the correct behaviour, I'll test it out today.

actgardner:
> what's the context where a user would use this flow? E.g. in the step:

> > the app then makes an RPC to exchange the id token for a pachyderm token to make calls on behalf of the user

> Is the app here Dex? Does Dex make Pachyderm calls on behalf of the user? Or is it Jupyter or something? Or Dash?

Dex is the OIDC provider, you can imagine Keycloak in this role as well. The provider just issues the ID token, it doesn't make any calls to Pachyderm. 

Jupyter, Dash, etc. (the app) does a normal OAuth flow with the provider and gets back an ID token. The app can then use this token to make authenticated calls to Pachyderm on the user's behalf.

You can think of this like when you OAuth to a web app and it uses Google OAuth to get access to your Google Drive.

actgardner:
This should be good for a second pass. Happy to answer any more questions about the motivation as well.

src/server/auth/server/oidc.go
```diff
@@ -464,30 +395,19 @@ func (a *apiServer) handleOIDCExchangeInternal(ctx context.Context, sp *Internal
...
-		&oauth2.Token{
-			AccessToken: tok.AccessToken,
-		},
-	)
-	userInfo, err := sp.Provider.UserInfo(ctx, ts)
-	if err != nil {
-		return "", "", errors.Wrapf(err, "could not get user info")
-	}
-	return idToken.Nonce, userInfo.Email, nil
+	return idToken.Nonce, claims.Email, nil
```
msteffen: This is just me not understanding the ID Token claims, but do we need to check EmailVerified here too?

msteffen (approved):
Overall LGTM. 2 remaining questions (that are just informational):

1. Why undo the changes to the OIDC flow (commit 605af9d). Does it not work when you hold it like that?
2. I think this question only made sense before the OIDC flow was reverted, but would we need to check the EmailVerified claim in both codepaths or just the ID Token codepath? (restated inline as well)

Otherwise LGTM

actgardner:
>Why undo the changes to the OIDC flow (commit 605af9d). Does it not work when you hold it like that?

It seemed like bad form to rework existing code we've tested extensively in a drive-by way. When I tested with Dex the id token from the auth code flow did contain the email claim. In my experience that's always the case. However it was difficult to find a citation in the spec to justify this. So I decided to keep it as-is.

>I think this question only made sense before the OIDC flow was reverted, but would we need to check the EmailVerified claim in both codepaths or just the ID Token codepath? (restated inline as well)

We probably should, my only concern is breaking existing integrations that customers may be using. 

I'd be happy to make these changes in a follow-up if we know what set of oidc providers we need to test against.

msteffen (approved):
Got it, thanks! LGTM

adelelopez:
Yeah, this LGTM!

adelelopez (approved):

PR #5300, opened 2020-09-28
[WIP] Internal groups

actgardner:
This isn't complete but there's enough sketched out that it seemed like a good point to get feedback versus investing more heads-down time. I'll try and explicitly note the TODOs below.

This PR implements the proposal in https://github.com/pachyderm/rfcs/issues/19:

- Remove the `SetGroupsForUser` RPC. This wasn't exposed in the CLI but it is a breaking API change that probably needs to wait for 2.0. The semantics of this command are pretty confusing in a world where admins can only change _some_ groups - does it ignore groups synced from SAML/OIDC? Does it error out if the user would be removed from those groups? It's also not clear if this would be applicable to both groups and users - if you can bulk edit the group membership of a group, the logic to update our secondary indexes becomes much more complicated.

- Add a `ListGroups` RPC. Lets admins list all groups on the cluster. The naming is slightly confusing when contrasted to `GetGroups`, which returns the groups a given user is in - if we're going all-in on breaking changed we could rename `GetGroups` to `GetGroupsForUser`, etc.

- Rename `ModifyMembers` to `ModifyGroupMembers`. Because users can only manage internal groups, the semantics of this call changed - `group` is assumed to be the name of an internal pachyderm group. Instead of leaving in a superficially similar RPC with different behaviour, I thought it was safer to rename this one to indicate a breaking change in behaviour.

- Change the name of the datatype we store in `groups` from `Users` to `GroupMembership`. We could modify the `Users` type but the name is misleading. We could also add a second etcd key that only stores the secondary index data (parents and transitive children) - that might help us avoid size limits in etcd, if we're concerned about those.

- `getScope` implements the "group membership interview question" - check for intersection between the transitive children of every group in the ACL and the group membership of the principal being evaluated. We try to bail out if we find OWNER early, but this would be much more efficient in most cases if we knew the precise level of access the user had requested.

- `ModifyGroupMembers` attempts to implement updating group membership without re-evaluating the transitive children of every parent by maintaining the number of unique routes between two groups. There's a comment that hopefully lays out the theory. 

- `setGroupsForUserInternal` now only operates on external groups from the IDP being synced - so SAML logins don't clobber internal group membership. It's worth noting because external groups can't contain nested groups we don't maintain their set of transitive child groups.

- Add the appropriate CRUD commands in pachctl - not a lot to say here.

Open Questions:
- Using maps for sets? We could define our own set operations over a sorted list - this should make evaluating intersection faster, at the cost of slower inserts/deletions/single point lookups. It's also extra complexity and we could always switch if we discover the performance of maps at scale is poor.

- Requiring fully-qualified names? I think for 2.0 we want to seize the opportunity to require fully-qualified user/group names for role bindings and group membership? Right now this PR punts on that decision and attempts to guess fully-qualified names for group members.

- Passing the desired level of access to `getScope`? I think we can punt on this as well, in 2.0 I'd love to have generic role bindings instead of reader/writer/owner, in which case this will need to be reconsidered.

TODO:
- `DeleteGroup` RPC and CLI operation.
- Unit and integration tests.


msteffen:
Just capturing notes from our pairing session just now:

> Open Questions:
>
> Using maps for sets? We could define our own set operations over a sorted list - this should make evaluating intersection faster, at the cost of slower inserts/deletions/single point lookups. It's also extra complexity and we could always switch if we discover the performance of maps at scale is poor.

- Per discussion, this makes sense, with the one caveat that it might make sense to design the collection in etcd differently (so that group membership collections are long & narrow vs short & wide). The next step was just to see if etcd has documented limits on row sizes & check if the largest groups we expect will fit or not

> Requiring fully-qualified names? I think for 2.0 we want to seize the opportunity to require fully-qualified user/group names for role bindings and group membership? Right now this PR punts on that decision and attempts to guess fully-qualified names for group members.

This makes complete sense to me. IIRC, the discussion here was mostly about how 1) the kind of canonicalization we do for GitHub won't work for other ID providers, and 2) We could provide `ListUsers()` and `ListGroup()` APIs for the dash, so that users editing ACLs there can rely on autocomplete. This means that a user must sign in before they can be added to an ACL or group, which seems like a reasonable restriction (and on reflection, internal groups could be introduced via an explicit `CreateGroup()` API, while external groups shouldn't normally be edited)

> Passing the desired level of access to getScope? I think we can punt on this as well, in 2.0 I'd love to have generic role bindings instead of reader/writer/owner, in which case this will need to be reconsidered.

This makes sense to me as well. The discussion was about how most uses of `getScope` are internal authorization inside of the auth APIServer, and the one external use is ListRepo. `getScope` could accept a list of roles and just pick the meet, which would work for all cases.

Other questions that came up:
1. Does tracking the number of paths from a node N to a downstream node M allow the transitive closure of N to be updated when an edge is added/removed? Note that in general, there can be at most one path between a pair of nodes (A, B).

2. Longer term, how should we deal with ACLs on groups? Maybe we shouldn't try to distinguish between ACLs and groups, and just make everything a set? What's the right internal model?

PR #5303, opened 2020-09-29
Fixing pachctl hanging on connecting on windows

Tryneus:


nitinjainsj (approved):
This change just generally makes me nervous, because we don't fully understand how this is suppose to work. Even at the time, we made the earlier change, it was not entirely clear why the name was not getting resolved (without the dns).

Worth testing it out on mac and linux.

nitinjainsj:
@Tryneus This works on Mac as well. You should be good to merge this in.

PR #5304, opened 2020-09-29
Fix error printing in list pipeline

Tryneus:
This error statement was printed multiple times if multiple pipelines had errors, which seems unintended.  Also, it was printed to stdout, which can hurt things like parsing of a json-formatted output, so I moved it to stderr.


jdoliner (approved):
LGTM

PR #5307, opened 2020-09-30
Update minikube version to try & fix config issue

msteffen:
Recently, CI runs have been failing with:
```
[kubeconfig] Writing "kubelet.conf" kubeconfig file
error execution phase kubeconfig/controller-manager: failed to load kubeconfig file /etc/kubernetes/controller-manager.conf that already exists on disk: read /etc/kubernetes/controller-manager.conf: is a directory
: running command: 
sudo /usr/bin/kubeadm init --config /var/lib/kubeadm.yaml --ignore-preflight-errors=DirAvailable--etc-kubernetes-manifests --ignore-preflight-errors=DirAvailable--data-minikube --ignore-preflight-errors=Port-10250 --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-scheduler.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-apiserver.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-kube-controller-manager.yaml --ignore-preflight-errors=FileAvailable--etc-kubernetes-manifests-etcd.yaml --ignore-preflight-errors=Swap --ignore-preflight-errors=CRI 
```
I haven't been able to find any relevant reports on GitHub, but maybe this will help?


lukemarsden:
eh, I kinda feel like and suspect that the original issue is flakiness to do with retrying - and failing to properly teardown the cluster during the retry. Did it always happen after at least one retry?

That said bumping the versions is _probably_ not a bad idea. CI had a failure in the ADMIN suite for this PR, I've retried it so let's see if that ADMIN failure was a flake or a 

If my theory in the first sentence is correct, the CI upgrade I have planned should fundamentally solve that problem by giving each retry a brand new test VM, pre-warmed with Kube running in it. I'm hoping to dedicate serious time and focus to that once I've shipped the first version of KFData.

lukemarsden (approved):
I'm ok with this if CI is no worse than before, but haven't seen it fully green on this branch yet.

msteffen:
> eh, I kinda feel like and suspect that the original issue is flakiness to do with retrying - and failing to properly teardown the cluster during the retry. Did it always happen after at least one retry?
> 
> That said bumping the versions is _probably_ not a bad idea. CI had a failure in the ADMIN suite for this PR, I've retried it so let's see if that ADMIN failure was a flake or a permafail. If this doesn't make CI worse, we can merge it.
> 
> If my theory in the first sentence is correct, the CI upgrade I have planned should fundamentally solve that problem by giving each retry a brand new test VM, pre-warmed with Kube running in it. I'm hoping to dedicate serious time and focus to that once I've shipped the first version of KFData.

Makes sense, and I'm looking forward to it! Given that theory, I'll exend the retry code to remove whatever's at this path and add that to this PR. I'll ping the PR again once it's green

lukemarsden:
Merge at will once it's green! Thank you for working on this!

lukemarsden:
BTW things that are files unexpectedly being directories is almost always due to the behaviour that `docker run -v foo:/bar` makes `/bar` into a directory if it doesn't exist, which is bonkers behavior which should never have been allowed in the first place, but hey Docker was a wildly successful prototype.

lukemarsden:
@msteffen it's not completely obvious what the motivation is behind the diff, especially moving the bits of `travis.sh` around. Could you write a few words explaining that here? Alternatively, merge if you are confident with the change. Thanks!

msteffen:
Ah, so all of these changes are to the end of resolving the `failing to properly teardown the cluster during the retry` issue. In particular, I guessed (given the error text) that the particular piece of cleanup that was failing was deleting `/etc/kubernetes/controller-manager.conf`.

However, after trying to add code to clean up that file before retrying CI, I noticed that that cleanup code did actually exist in `etc/testing/travis.sh` (line 25, `rm -rf /etc/kubernetes...`).

After looking through the surriounding code a bit more, I noticed that despite the cleanup code being present, it wasn't being run: `etc/testing/travis.sh` calls `make launch-kube` (line 30), which has its own retry loop. I concluded that we could be seeing this failure when `etc/kube/start-minikube.sh` (called by `make launch-kube`) is going through its own retry process. I could take the cleanup code in `etc/testing/travis.sh` and move it into `etc/kube/start-minikube.sh` but I decided that would be a mistake: the former is designed to run in Travis while the latter is not, and those directories only need to be cleaned up in the case where we're running minikube with `--vm-driver=None` (a practice that is only recommended in transient VMs, and which our code only does in Travis. My operating model here is that `--vm-driver=None` is set iff running in Travis). Therefore I removed the retry loop from `etc/kube/start-minikube.sh` as I don't think it does much for the general usefulness of that script and hinders our ability to cleanup in Travis). Now, there is only one retry loop (instead of two nested loops, with the inner loop deficient of cleanup) and that one remaining retry loop includes the cleanup code that I think will prevent this error.

While I was at it, both `etc/kube/start-minikube.sh` and `etc/testing/travis.sh` attempt to connect to minikube to confirm that it's been brought up, and that code I left in `etc/testing/travis.sh` and removed from `etc/kube/start-minikube.sh` (so that that's also only done once)

Those two changes I've made since the last review, all mostly an attempted fix for the cleanup error you pointed out.

msteffen:
CI looks green so I'm merging. See comment above describing recent changes, and if there's more I should clarify just let me know!

PR #5308, opened 2020-09-30
OIDC integration supports additional claims and groups

actgardner:
This has a few OIDC changes that kind of go together:
- Remove the existing call to the userinfo endpoint after authenticating. According to the specification, when the `openid` scope is included the response must include an ID token. In testing with Auth0, Google and Dex, this is true. This way we can reuse the same logic for ID token parsing everywhere.
- Add the ability to request arbitrary additional scopes. This allows users to add the `groups` scope to the request, if their provider supports it.
- Sync the groups that are returned, if applicable. I tested this by wiring up Dex to Github and syncing our org's team membership. This should also work for syncing group membership from LDAP, SAML, and Google (via Dex).
- Add the option to disable checking `email_verified`. Even though it's in the spec some providers apparently don't support it.


src/server/auth/server/oidc.go
```diff
@@ -34,16 +34,17 @@ var (
...
+	Groups        []string `json:"groups"`
 }
 
 // InternalOIDCProvider contains information about the configured OIDC ID
 // provider, as well as auth information identifying Pachyderm in the ID
 // provider (ClientID and ClientSecret), which Pachyderm needs to perform
 // authorization with it.
 type InternalOIDCProvider struct {
-	// a points back to the owning auth/server.apiServer, currently just so that
+	// a points back to the owning auth/json:"groups"server.apiServer, currently just so that
```
adelelopez, line 47: did you mean to change this line?

adelelopez (approved):
Sweet! It will be nice to have support for groups

msteffen (approved):
LGTM

Outside the scope of this PR, but just out of curiosity: given the experience you've developed with Dex, do you think it would be possible to deploy it in minikube in CI and emulate an authentication flow (for automated testing)?

IIUC, SAML makes this a bit easier because there's no auth code <-> ID/access token exchange (I think it's basically similar to the OIDC implicit flow), so you can just generate a signed SAML assertion in the test and send it to pachd's SAML handler. But for OIDC you have to have an endpoint that pachd can do the exchange with.

src/server/auth/server/oidc.go
actgardner: Good catch 😬

actgardner:
> Outside the scope of this PR, but just out of curiosity: given the experience you've developed with Dex, do you think it would be possible to deploy it in minikube in CI and emulate an authentication flow (for automated testing)?

I can spike this out next week - dex supports configuring a user with a fixed password, so it may be possible to cURL our way through the auth code flow.

PR #5311, opened 2020-10-01
Several pfs 2.0 changes / fixes

brycemcanally:


src/client/pfs_v2.go
```diff
@@ -65,22 +78,22 @@ func (c APIClient) NewFileOperationClientV2(repo, commit string) (_ *FileOperati
 }
 
 // PutTar puts a tar stream into PFS.
-func (foc *FileOperationClient) PutTar(r io.Reader, tag ...string) error {
+func (foc *FileOperationClient) PutTar(r io.Reader, overwrite bool, tag ...string) error {
 	return foc.maybeError(func() error {
+		ptr := &pfs.PutTarRequestV2{Overwrite: overwrite}
 		if len(tag) > 0 {
```
brendoncarroll, line 84: This might be cleaner as a switch.

brendoncarroll, line 84: This might be cleaner as a switch.

src/server/pfs/server/driver_v2.go
```diff
@@ -334,13 +370,18 @@ func (d *driverV2) compactIter(ctx context.Context, params compactSpec) (_ *comp
...
 		start := i * childSize
+		if start >= len(params.inputPaths) {
+			break
+		}
 		end := (i + 1) * childSize
 		if end > len(params.inputPaths) {
 			end = len(params.inputPaths)
 		}
-		childOutputPath := path.Join(scratch, strconv.Itoa(i))
+		childOutputPath := path.Join(scratch, fileset.SubFileSetStr(int64(i)))
```
brendoncarroll, line 384: This wasn't a bug before, the list ensures ordering, we don't have to rely on lexicographical ordering.

src/server/pkg/storage/fileset/merge.go
```diff
@@ -31,64 +31,80 @@ func newMergeReader(rs []*Reader) *MergeReader {
...
+			next := strings.TrimRight(fr.Index().Path, "/") + "0"
+			if err := fss[i].r.iterate(func(_ *FileReader) error { return nil }, next); err != nil {
+				return err
+			}
 		}
 	}
-	return result
+	contentTags, deleteTags := computeTags(frs)
+	if len(contentTags) == 0 {
+		if len(deleter) > 0 && len(deleteTags) > 0 {
```
brendoncarroll, line 76: Is deleter just a predicate? I think this would be more clear as a non variadic argument.   You could set it to always return true if it was nil.

src/server/pkg/storage/fileset/merge.go
```diff
@@ -31,64 +31,80 @@ func newMergeReader(rs []*Reader) *MergeReader {
...
+		}
+	}
+	// Progress each lower priority stream past the deleted file(s).
+	for i += 1; i < len(fss); i++ {
+		fr, err := fss[i].r.Next()
+		if err != nil {
+			return err
+		}
+		// If the full delete is for a directory, progress all lower priority streams past the directory.
+		if strings.HasSuffix(fr.Index().Path, "/") {
```
brendoncarroll, line 67: Do we have an `IsDir` utility function somewhere? We do this a lot.

src/server/pkg/storage/fileset/merge.go
```diff
@@ -31,64 +31,80 @@ func newMergeReader(rs []*Reader) *MergeReader {
...
+			next := strings.TrimRight(fr.Index().Path, "/") + "0"
+			if err := fss[i].r.iterate(func(_ *FileReader) error { return nil }, next); err != nil {
```
brendoncarroll, line 69: You shouldn't need that underscore just the type is allowed.

src/server/pkg/storage/fileset/merge.go
```diff
@@ -224,7 +215,7 @@ func (fmr *FileMergeReader) Index() *index.Index {
 }
 
 // Header returns the tar header for the merged file.
-func (fmr *FileMergeReader) Header() (*tar.Header, error) {
+func (fmr *FileMergeReader) Header() (_ *tar.Header, retErr error) {
```
brendoncarroll, line 218: retErr appears unused. Revert this line

src/server/pkg/storage/fileset/merge.go
```diff
@@ -254,7 +245,7 @@ func (fmr *FileMergeReader) Header() (*tar.Header, error) {
 }
 
 // WriteTo writes the merged file to the passed in fileset writer.
-func (fmr *FileMergeReader) WriteTo(w *Writer) error {
```
brendoncarroll, line 257: Same here as above

src/server/pkg/storage/fileset/merge.go
```diff
@@ -341,86 +317,66 @@ func newTagSetMergeReader(frs []*FileReader) *TagSetMergeReader {
...
+}
+
+func (tsmr *TagSetMergeReader) mergeTag(ss []stream, cb func(*TagMergeReader) error, deleter ...func(string)) (retErr error) {
+	// Convert generic streams to tag streams.
+	var i int
+	var tss []*tagStream
+	for ; i < len(ss); i++ {
+		if dts, ok := ss[i].(*deleteTagStream); ok {
+			if len(deleter) > 0 {
+				deleter[0](dts.tag.Id)
```
brendoncarroll, line 333: Same feedback as earlier. deleter should not be variadic. If it is nil, assign a function that always returned false.

src/server/pkg/storage/fileset/merge.go
```diff
@@ -341,86 +317,66 @@ func newTagSetMergeReader(frs []*FileReader) *TagSetMergeReader {
...
-		}
-		// Create tag merge reader and execute callback.
-		var trs []*chunk.TagReader
-		for _, ts := range tss {
-			trs = append(trs, ts.fr.NextTagReader())
-		}
-		return f(newTagMergeReader(trs))
-	})
+	}
+	if len(tss) == 0 {
```
brendoncarroll, line 348: Shouldn't we do this check before the loop above?

src/server/pkg/storage/fileset/util.go
```diff
@@ -82,3 +83,12 @@ func IsCleanTarPath(x string, isDir bool) bool {
 	y := CleanTarPath(x, isDir)
 	return y == x
 }
+
+func getSortedKeys(set map[string]struct{}) []string {
+	var keys []string
```
brendoncarroll, line 88: nit: I like to pre-allocate with make here since you can get the since of the map.

src/server/pkg/storage/fileset/merge.go
```diff
@@ -31,64 +31,80 @@ func newMergeReader(rs []*Reader) *MergeReader {
...
+	}
+	// Progress each lower priority stream past the deleted file(s).
+	for i += 1; i < len(fss); i++ {
+		fr, err := fss[i].r.Next()
+		if err != nil {
+			return err
+		}
+		// If the full delete is for a directory, progress all lower priority streams past the directory.
+		if strings.HasSuffix(fr.Index().Path, "/") {
+			next := strings.TrimRight(fr.Index().Path, "/") + "0"
```
brendoncarroll, line 68: This is a little sketchy. Unless we check all the paths to make sure that 0 is the first allowed character we should probably find a better way of doing this.  Maybe have a utility function that does this so the intent is more obvious.  A NULL character instead of 0 might make this slightly less sketchy.

src/client/pfs_v2.go
brycemcanally: I prefer the current pattern for consistency reasons. Most variadic functions in our codebase capture the handling of the optional arguments in a single block with the check len(args) > 0.

src/server/pfs/server/driver_v2.go
brycemcanally: The issue is the common prefix, not the ordering. So, when setting up the merge reader we would end up grabbing some filesets more than once (i.e. greater than 10 paths would result with duplicate data because the 10th fileset would be grabbed twice, once by the walk of 1 and once by the walk of 10). This was causing data to be duplicated.

src/server/pkg/storage/fileset/merge.go
brycemcanally: No, deleter is a callback that performs a deletion. It is optional functionality provided by `mergeFile` and `mergeTag`.

src/server/pkg/storage/fileset/merge.go
brycemcanally: Good point. I'll refactor this into a `IsDir` utility function.

src/server/pkg/storage/fileset/merge.go
brycemcanally: Same as above. It is an optional callback that performs a deletion, not a predicate.

src/server/pkg/storage/fileset/merge.go
brycemcanally: This is operating on the tag streams, not the generic streams. If it was before (and operated on the generic streams), then deletions would not be handled properly since we would never get to the point where the deletion callback is executed and we would not progress the other streams past the deleted tag.

src/server/pkg/storage/fileset/merge.go
brycemcanally: The immediate next character after "/" in a lexicograpical sort is "0", so the lowest next path (in terms of the sort) after "/dir/" would be "/dir0". The next in this case defines an exclusive upper bound for the iteration. I can put this in a utility function.

src/server/pkg/storage/fileset/merge.go
brendoncarroll: Regardless of how you want to encode the parameter I still think it makes the core algorithm more clear to deal with the peculiarities of whatever parameter encoding once at the top, and turn those branches into calls to a no-op.

brendoncarroll (approved):
LGTM. Disagree and commit on the optional parameter.

PR #5314, opened 2020-10-01
Fix update pipeline with reprocess and stats

brycemcanally:
Update pipeline removes the branch provenance from the output branch when a pipeline is being updated. This prevents new commits from getting created during the update (and therefore jobs). After the update, the branch provenance is re-enabled which will create a new commit that corresponds with the new provenant spec commit. Based on the code, it looks like this same pattern was intended to apply to the stats branch, but the part where the provenance is removed before the update is missing. So, this ended up working fine when doing an update without a reprocess because we would set the head commit for each branch to the current head when re-enabling provenance. When the output branch provenance was re-enabled, it would generate a new commit which would then result with a new stats commit due the stats branch provenance still being enabled. Then "re-enabling" the stats branch would result with no new commits because the head commit would have the correct provenance. This is not the case with reprocess because the commit chain gets broken in both branches. So, re-enabling the output branch provenance would still create an output commit with no parent and a stats commit with a parent in the same chain. Then, we would set the stats branch head to nil (to break the stats commit chain), which would result with propagate commit creating another stats commit. This is why we would see two stats commits after an update pipeline with reprocess and stats enabled. So, the fix is just to disable the stats branch provenance and re-enable after the output branch provenance is restored. This should guarantee that our stats commits are setup correctly after update pipeline.

### DIFF
--- a/src/server/pps/server/api_server.go
+++ b/src/server/pps/server/api_server.go
@@ -1833,6 +1833,16 @@ func (a *apiServer) hardStopPipeline(pachClient *client.APIClient, pipelineInfo 
 	); err != nil && !isNotFoundErr(err) {
 		return errors.Wrapf(err, "could not recreate original output branch")
 	}
+	if pipelineInfo.EnableStats {
+		if err := pachClient.CreateBranch(
+			pipelineInfo.Pipeline.Name,
+			"stats",
+			"stats",
+			nil,
+		); err != nil && !isNotFoundErr(err) {
+			return errors.Wrapf(err, "could not recreate original stats branch")
+		}
+	}
 
 	// Now that new commits won't be created on the master branch, enumerate
 	// existing commits and close any open ones.
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -2880,20 +2880,24 @@ func TestUpdatePipeline(t *testing.T) {
 	dataRepo := tu.UniqueString("TestUpdatePipeline_data")
 	require.NoError(t, c.CreateRepo(dataRepo))
 	pipelineName := tu.UniqueString("pipeline")
-	require.NoError(t, c.CreatePipeline(
-		pipelineName,
-		"",
-		[]string{"bash"},
-		[]string{"echo foo >/pfs/out/file"},
-		&pps.ParallelismSpec{
-			Constant: 1,
-		},
-		client.NewPFSInput(dataRepo, "/*"),
-		"",
-		true,
-	))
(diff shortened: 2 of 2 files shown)


src/server/pps/server/api_server.go
```diff
@@ -1833,6 +1833,16 @@ func (a *apiServer) hardStopPipeline(pachClient *client.APIClient, pipelineInfo
...
+			"stats",
+			"stats",
+			nil,
+		); err != nil && !isNotFoundErr(err) {
+			return errors.Wrapf(err, "could not recreate original stats branch")
+		}
+	}
 
 	// Now that new commits won't be created on the master branch, enumerate
 	// existing commits and close any open ones.
```
msteffen, line 1848: Does this logic (closing any existing open commits) need to be applied to the stats branch as well?

msteffen (approved):
LGTM, with one question about hardStopPipeline. Good catch!

src/server/pps/server/api_server.go
brycemcanally: Yeah, I thought about adding some code to do that, but I was thinking we could lean on the recovery code for the time being since there are a few other places where we only finish the output commit to terminate a job. But, I guess that won't work in the reprocess case since the commit chain is broken. So, we wouldn't be blocked anywhere, but some stats commits could be left open after we break the commit chain. Making this change could still result with job state not being updated correctly though. I've discussed this a few times with @jdoliner and @nitinjainsj , but I really think we need to go ahead and make the appropriate changes such that we finish all relevant job state / commits transactionally rather than having this indirection through pfs. So, it basically plays out like: finish commit -> notice commit is finished -> finish job / stats commit. A better model that would guarantee consistent state would just be a single finish / stop job that handles it all transactionally. It is pretty easy to map an output commit to a job, but since jobs may not always exist for an output commit, that may make things a bit trickier.

msteffen: Sure—as an intermediate fix, we could change FinishCommit so that if `empty = true` it also finishes all subvenant commits (also setting `empty = true`), all transactionally?

brycemcanally: After some thought, I think it makes sense to keep this fix simple for the time being to keep the risk of new issues low. In a 2.0 release I will add a more complete fix to this problem, then backport it when we are confident the problem is solved.

msteffen: Sure, that sounds good. LGTM still LGTM

PR #5320, opened 2020-10-05
Add OIDC integration test

actgardner:
Stand up a Dex server with user/password authentication to test our OIDC support.

Open Q: do we need to do more to make this easy to use locally? Right now to run `make test-auth` locally you need to have run `make launch-dex` first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -392,6 +392,15 @@ launch-loki:
 clean-launch-loki:
 	helm uninstall loki
 
+launch-dex:
+	helm repo add stable https://kubernetes-charts.storage.googleapis.com/
+	helm repo update
+	helm upgrade --install dex stable/dex -f etc/testing/auth/dex.yaml
+	until timeout 1s ./etc/kube/check_ready.sh release=dex; do sleep 1; done
+
+clean-launch-dex:
+	helm uninstall dex
+
 logs: check-kubectl
 	kubectl $(KUBECTLFLAGS) get pod -l app=pachd | sed '1d' | cut -f1 -d ' ' | xargs -n 1 -I pod sh -c 'echo pod && kubectl $(KUBECTLFLAGS) logs pod'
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -392,6 +392,15 @@ launch-loki:
 clean-launch-loki:
 	helm uninstall loki
 
+launch-dex:
+	helm repo add stable https://kubernetes-charts.storage.googleapis.com/
```
msteffen, line 396: Using --force-update (as in `helm repo add --force-update stable ...`) might be good here—in a few other places (e.g. our loki tests) it's prevented CI flakes when Minikube chokes and travis retries the make target

src/server/auth/server/testing/oidc_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	loginInfo, err := adminClient.GetOIDCLogin(adminClient.Ctx(), &auth.GetOIDCLoginRequest{})
+	require.NoError(t, err)
+
+	// Create an HTTP client that doesn't follow redirects.
+	// We rewrite the host names for each redirect to avoid issues because
+	// pachd is configured to reach dex with kube dns, but the tests might be
+	// outside the cluster.
+	c := &http.Client{}
+	c.CheckRedirect = func(_ *http.Request, via []*http.Request) error {
```
msteffen, line 48: Brilliant

msteffen: This was one of the issues @adelelopez and I ran into the first time we worked on OIDC, and I remember trying *so hard* to get the same URL to resolve the same way inside and outside the cluster by overwriting `/etc/hosts` and such. This is way simpler.

msteffen (approved):
It'll be really great to have tests for this. One small suggestion re:helm, but overall, LGTM in the extreme.

Cookie monster LGTM

(re: running this locally, I think expecting engineers to run `make launch-dex` seems reasonable. We can always make `test-auth` run `launch-dex` later, if needed)

PR #5322, opened 2020-10-06
Better handling of incomplete get pipeline info when PFS objects are unavailable

Tryneus:
`ppsutil.GetPipelineInfo` was changed a while back to not error if the `PipelineInfo` object could not be loaded from the spec repo (to work around problems talking to object storage or a missing object).  Unfortunately, this means that many of the `PipelineInfo` fields were seemingly valid but wrong (due to golang's zero-initialization).  This PR adds a separate call `ppsutil.GetPipelineInfoAllowIncomplete` that will preserve this behavior, while `ppsutil.GetPipelineInfo` will once again error in this case.

Ideally, a pipeline corrupted like this will not completely block most PPS operations on the cluster, and the following calls should still work:
 * `list pipeline`
 * `update pipeline`
 * `delete pipeline`

For `list pipeline` to still work, the request was given a new field `allow_incomplete`, which will ignore errors loading the pipeline spec and return an incomplete struct.  The pretty-printing code has been updated to indicate which fields could not be determined and why.

For `update pipeline`, `pps.CreatePipeline` was updated to allow incomplete information on the pipeline.  This was used for the old pipeline's `Salt`, `Version`, and `Stopped`:
 * For `Stopped`, I think we are ok with this state being lost in the case of recovering from a missing pipeline spec
 * For `Version`, the latest version of the pipeline has been added to the `EtcdPipelineInfo`, which is not missing
 * For `Salt`, this is not recovered, so updating a pipeline in such a state requires `Reprocess=true`

For `delete pipeline`, this already appears to work if the spec could not be loaded.


src/server/pkg/ppsutil/util.go
```diff
@@ -146,6 +147,16 @@ func GetPipelineInfo(pachClient *client.APIClient, name string, ptr *pps.EtcdPip
 	return result, nil
 }
 
+// GetPipelineInfo retrieves and returns a valid PipelineInfo from PFS. It does
+// the PFS read/unmarshalling of bytes as well as filling in missing fields
+func GetPipelineInfo(pachClient *client.APIClient, name string, ptr *pps.EtcdPipelineInfo) (*pps.PipelineInfo, error) {
```
adelelopez, line 152: Why not just add a new bool field `allowIncomplete` to this?

src/server/pachyderm_test.go
```diff
@@ -12311,6 +12401,7 @@ func getUsablePachClient(t *testing.T) *client.APIClient {
 		if err == nil {
 			return client
 		}
+		time.Sleep(5 * time.Second)
```
adelelopez, line 12404: do we still need this?

src/server/pachyderm_test.go
```diff
@@ -11221,6 +11221,95 @@ func TestPipelineHistory(t *testing.T) {
...
+
+	// Should be able to list pipelines with AllowIncomplete=true
+	response, err := c.PpsAPIClient.ListPipeline(c.Ctx(), &pps.ListPipelineRequest{AllowIncomplete: true})
+	require.NoError(t, err)
+	require.Equal(t, 1, len(response.PipelineInfo))
+
+	// Pipeline should have the correct version
+	require.Equal(t, uint64(1), response.PipelineInfo[0].Version)
+
+	// TODO: test updating the pipeline
```
adelelopez, line 11301: Seems like we should be testing this and the reprocess since the code paths for those have been changed.

adelelopez (changes requested):

src/server/pkg/ppsutil/util.go
Tryneus: Because a bool field isn't as easy to parse as something unsafe.  If go had named parameters that would be great, but this makes it obvious if the function will return something unexpected.

src/server/pachyderm_test.go
Tryneus: I think this is still useful - on my machine the reconnect would error immediately and it wouldn't wait long enough for the new pods to come up (even with trying 10x).

src/server/pachyderm_test.go
Tryneus: Yep, I'll add more test stuff!

src/server/pkg/ppsutil/util.go
adelelopez: Makes sense!

src/server/pachyderm_test.go
adelelopez: Alright, would one of our Backoff functions be a better fit here then?

src/server/pachyderm_test.go
Tryneus: Turns out we can't actually do an update because that still requires the previous commit to be accessible in the spec repo, so  I've changed that code path to error as well - the only recovery course is to delete and recreate.

src/server/pachyderm_test.go
Tryneus: Done

adelelopez (approved):
Awesome, LGTM 😄

PR #5323, opened 2020-10-06
converted credentials to secrets, fixed spout to not close /pfs/out

pappasilenus:


JimmyWhitaker:
Overall everything looks pretty good. I wonder if we should look at using `.env` files to manage the config locally and create secrets from there? Just a thought. Here's a quick [reference](https://medium.com/codait/environment-variables-or-keeping-your-secrets-secret-in-a-node-js-app-99019dfff716) on doing it that way.

JimmyWhitaker (approved):
Everything looks good to me. I added a couple of comments, but they're more food for thought.

PR #5327, opened 2020-10-07
Storage Paths in Postgres

brendoncarroll:


src/server/pkg/storage/fileset/index/writer.go
```diff
@@ -149,38 +140,11 @@ func (w *Writer) Close() (retErr error) {
...
-			retErr = err
-		}
-	}()
-	// Handles the empty file set case.
-	if w.root == nil {
-		_, err = pbutil.NewWriter(objW).Write(&Index{})
-		return err
-	}
-	chunk := w.root.DataOp.DataRefs[0].ChunkInfo.Chunk
-	if w.rootTTL > 0 {
```
brendoncarroll, line 174: I think we may need to add this back for the time being.  It will be changed with the chunk/gc changes.

Issue #5298, opened 2020-09-28
Pipeline stuck in "Starting" after period of master unavailability


actgardner:
**What happened?**:

When running workloads automatically on a new Hub cluster in the examples runner, pipelines occasionally stick in a state where they are "starting" but can never make progress. My suspicion is that the GKE cluster begins to automatically resize (adding a nodepool) which makes the k8s master unavailable for several minutes. During this time we exhaust retries setting up the RCs.

In this case the examples runner got stuck running the openCV example: https://github.com/pachyderm/pachyderm/tree/master/examples/opencv

```
root@example-runner-1601056800-qqkt5:/app# pachctl list pipeline
NAME    VERSION INPUT                CREATED    STATE / LAST JOB    DESCRIPTION
montage 1       (edges:/ ⨯ images:/) 2 days ago starting / starting A pipeline that combines images from the `images` and `edges` repositories into a montage.
edges   1       images:/*            2 days ago starting / starting A pipeline that performs image edge detection by using the OpenCV library.
```

Logs:
```
2020-09-25T18:10:14Z ERROR PPS master: error retrieving RC for "montage": Get https://10.4.0.1:443/api/v1/namespaces/default/replicationcontrollers?labelSelector=pipelineName%3Dmontage: dial tcp 10.4.0.1:443: i/o timeout; retrying in 544.806308ms
2020-09-25T18:10:21Z ERROR PPS master: error retrieving RC for "montage": Get https://10.4.0.1:443/api/v1/namespaces/default/replicationcontrollers?labelSelector=pipelineName%3Dmontage: dial tcp 10.4.0.1:443: connect: connection refused; retrying in 758.900801ms
2020-09-25T18:10:22Z ERROR PPS master: Get https://10.4.0.1:443/api/v1/namespaces/default/replicationcontrollers?labelSelector=pipelineName%3Dmontage: dial tcp 10.4.0.1:443: connect: connection refused
```

No replication controllers have been created:
```
k get rc
No resources found in default namespace.
```

**What you expected to happen?**:

Pachd should be able to tolerate arbitrary outages of the k8s master and restart pipelines without intervention.

**How to reproduce it (as minimally and precisely as possible)?**:

Add a new nodepool to a GKE cluster and try to create a pipeline while the cluster is in maintenance mode. You can try to `kubectl` to confirm the master is inaccesible during the operation.

**Anything else we need to know?**:

**Environment?**:
- Kubernetes version (use `kubectl version`): N/A
[...]

PR #5330, opened 2020-10-08
More changes to allow CRASHING -> STANDBY

msteffen:
This PR overall addresses several bugs in PPS discovered by @brycemcanally:
- If the kubernetes API server is unavailable, `step()` called by the PPS master will log an error and exit, dropping the pipline event (https://github.com/pachyderm/pachyderm/issues/5298)
  - **Fix**: add a new function, `pollPipelines` that overwrites every pipeline in etcd with itself (one pipeline every 2 seconds), ensuring that every pipeline is eventually brought into the correct state. It also deletes dangling RCs and monitorPipeline goros for pipelines that have been removed from etcd
- DeletePipeline doesn't cancel any running `monitorPipeline` goroutines
  - **Fix**: add a handler for etcd Delete events to `master()` in `pps/server/master.go`. Also, `pollPipelines` scans for orphaned `monitorPipeline` goroutines and cancels them
- pipelines in CRASHING never re-enter standby
  - Two fixes here:
  - `monitorPipeline` is no longer cancelled when a pipeline enters CRASHING, and `monitorPipeline` can now transition pipelines into STANDBY from either CRASHING or RUNNING (pipelines may complete even when in CRASHING if some but not all pods are able start, and the ones that start are able to finish the pipeline)
  - `pipeline_controller.go` now calls `scaleUpPipeline()` after a pipeline enters CRASHING, ensuring that kubernetes is trying to bring up pods, and that if the pipeline is able to finish the job despite some workers being unable to schedule (which puts the pipeline in CRASHING), it will finish the job and it'll go back into standby
- `stopPipelineMonitor` blocks until the `monitorPipeline` goroutine being stopped exits, preventing orphaned `monitorPipeline` goroutines from causing problems.


src/server/pkg/ppsutil/util.go
```diff
@@ -172,24 +172,31 @@ func CrashingPipeline(ctx context.Context, etcdClient *etcd.Client, pipelinesCol
...
+		if i > 0 {
+			froms.WriteString(", ")
+		}
+		froms.WriteString(state.String())
+	}
+	return fmt.Sprintf("could not transition %q from any of [%s] -> %s, as it is in %s",
+		p.Pipeline, froms.String(), p.Target, p.Current)
 }
 
 // SetPipelineState is a helper that moves the state of 'pipeline' from 'from'
```
adelelopez, line 192: update comment to reflect change in type of 'from'

src/server/pps/server/pipeline_controller.go
```diff
@@ -135,8 +137,12 @@ func (a *apiServer) step(pachClient *client.APIClient, pipeline string, keyVer,
 		if op.pipelineInfo.Stopped {
 			return op.setPipelineState(pps.PipelineState_PIPELINE_PAUSED, "")
 		}
-		// default: scale down if standby hasn't propagated to kube RC yet
+		// pipelines in standby shouldn't be able to crash (CRASHING is most like
+		// RUNNING)
```
adelelopez, line 141: nit: line break here looks weird

adelelopez (approved):
overall looks good, just a few minor nits

src/server/pps/server/pipeline_controller.go
msteffen: I actually ended up removing this comment as it's incorrect—crashingPipelineMonitor takes pipelines out of CRASHING, but it doesn't put them into CRASHING. The reason to cancel it here is because if the pipeline is in PAUSED, then a goro to take it out of CRASHING doesn't make sense.

Marking this resolved.

adelelopez (approved):
LGTM!

PR #5331, opened 2020-10-08
Tracing

msteffen:


src/server/cmd/pachctl/cmd/cmd.go
```diff
@@ -321,8 +321,11 @@ func PachctlCmd() *cobra.Command {
 Environment variables:
   PACH_CONFIG=<path>, the path where pachctl will attempt to load your pach config.
   JAEGER_ENDPOINT=<host>:<port>, the Jaeger server to connect to, if PACH_TRACE is set
-  PACH_TRACE={true,false}, If true, and JAEGER_ENDPOINT is set, attach a
-    Jaeger trace to any outgoing RPCs
+  PACH_TRACE={true,false}, if true, and JAEGER_ENDPOINT is
+    set, attach a Jaeger trace to any outgoing RPCs.
```
jrockway, line 325: Somewhat short line-wrapping going on here.  We should pick a column limit and try to fill it out.

src/client/pkg/tracing/extended/extended_trace.go
```diff
@@ -2,141 +2,137 @@ package extended
...
-	// traceCtxKey is the grpc metadata key whose value is a ExtendedTrace
-	// identifying the current RPC/commit
-	traceCtxKey = "commit-trace"
+	// traceCtxKey is the grpc metadata key whose value is a serialized
+	// ExtendedTrace proto tied to the current CreatePipeline request. In a grpc
+	// quirk, this key must end in '-bin' so that the value (a serialized
+	// timestamp) is treated as arbitrary bytes and base-64 encoded before being
+	// transmitted (see
+	// https://github.com/grpc/grpc-go/blob/b2c5f4a808fd5de543c4e987cd85d356140ed681/Documentation/grpc-metadata.md)
+	traceCtxKey = "pipeline-trace-duration-bin"
```
jrockway, line 25: I'd be careful about calling this a "context key", because those are something different; see `context.WithValue` and friends.

src/client/pkg/tracing/extended/extended_trace.go
```diff
@@ -2,141 +2,137 @@ package extended
...
 	return col.NewCollection(c,
-		TracesCollectionPrefix,
-		[]*col.Index{CommitIDIndex, PipelineIndex},
+		tracesCollectionPrefix,
+		nil, // no indexes
 		&TraceProto{},
-		nil,
-		nil)
+		nil, // no key check (keys are pipeline names)
+		nil) // no val check
```
jrockway, line 49: I like these comments a lot!

jrockway (approved):
This looks good to me.  Something that has bugged me about Jaeger and the Opentracing API is that you can't really have a certain trace apply to another trace (there's a long-running operation that was traced, and an RPC that affects that long-running operation; and the RPC spans can't be parented to both the RPC and the long-running operation).  This seems like a good workaround for that, if I'm understanding it correctly -- we just store the traces related to the long-running pipeline in etcd, and have a tool that can show you all those traces.  Then you can see your RPC in the context of long-running operations.

Something related to this is golang.org/x/net/trace.  That is unrelated to opentracing (or runtime/trace! confusing!) but acts as a garbage-collected event log for RPCs.  You add each RPC, and any related information, and it will generate a webpage that shows you the last 10.  I think this custom implementation is better than that, so no need to switch, but it's something that exists that could be interesting to integrate in the future.

src/client/pkg/tracing/extended/extended_trace.go
msteffen: Done. I went with traceMDKey here—hopefully that's clearer. We can also easily change this later since it's private.

src/server/cmd/pachctl/cmd/cmd.go
msteffen: Yeah, I think must've been trying to lay out the lines manually and made a mistake. There was certainly no deliberate effort to keep this line super short. I reflowed these lines and they should be fixed (all ~80 characters)

PR #5334, opened 2020-10-09
Fix lint error & shorten failure output

msteffen:
Travis is whining in https://github.com/pachyderm/pachyderm/pull/5331, but that's a bigger PR so I factored this out


actgardner (approved):
👍

msteffen:
Just noticed @actgardner is out for the day, so I'm reassigning

PR #5335, opened 2020-10-09
Add test for OIDC ID token in Authorize

actgardner:
Adds an integration test where we use the Resource Owner Password Flow (https://auth0.com/docs/flows/resource-owner-password-flow) to get an ID token using one client ID, but with Pachyderm's client id in the audience. We then use this token to authenticate to pachyderm on behalf of the user. 

NB: IRL the ROP Flow is not good and we shouldn't use it, but it avoids a lot of redirects here. To make it work I had to disable `staticPasswords` in Dex and instead use their mockPassword connector.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -393,9 +393,9 @@ clean-launch-loki:
 	helm uninstall loki
 
 launch-dex:
-	helm repo add stable https://charts.helm.sh/stable
+	helm repo add --force-update stable https://charts.helm.sh/stable
 	helm repo update
-	helm upgrade --install --force dex stable/dex -f etc/testing/auth/dex.yaml
+	helm upgrade --install dex stable/dex -f etc/testing/auth/dex.yaml
 	until timeout 1s bash -x ./etc/kube/check_ready.sh 'app.kubernetes.io/name=dex'; do sleep 1; done
 
 clean-launch-dex:
--- a/etc/testing/auth/dex.yaml
+++ b/etc/testing/auth/dex.yaml
@@ -18,18 +18,27 @@ config:
     tlsKey: /etc/dex/tls/https/server/tls.key
     allowedOrigins: []
   staticClients:
+  - id: testapp
+    name: "Test App"
+    redirectURIs:
+    - 'http://test.example.com:657/authorization-code/callback'
+    secret: test
   - id: pachyderm
     name: "Pachyderm"
     redirectURIs:
     - 'http://pachd:657/authorization-code/callback'
-    secret: blahblahblah
-  enablePasswordDB: true
-  staticPasswords:
-  - email: "admin@example.com"
-  # bcrypt hash of the string "password"
-    hash: "$2a$10$2b2cU8CPhOTaGrs1HRQuAueS7JTT5ZHsHSzYiFPm1leZck7Mc8T4W"
-    username: "admin"
-    userID: "08a8684b-db88-4b73-90a9-3cd1661f5466" 
+    secret: notsecret
+    trustedPeers:
+    - testapp
+  connectors:
+  - id: test
+    name: test
+    type: mockPassword
+    config:
+      username: admin
+      password: password
+  enablePasswordDB: false
   oauth2:
     alwaysShowLoginScreen: false
     skipApprovalScreen: true
+    passwordConnector: test
--- a/src/server/auth/server/testing/oidc_test.go
+++ b/src/server/auth/server/testing/oidc_test.go
@@ -13,6 +13,12 @@ import (
 	"github.com/pachyderm/pachyderm/src/client/pkg/require"
 )
 
+// Dex's mock connector always returns the same identity for all authentication
+// requests, and this is that identity's email (see
+// https://github.com/dexidp/dex/blob/c113df2730052e20881dd68561289f8ae121300b/connector/mock/connectortest.go#L21)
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -393,9 +393,9 @@ clean-launch-loki:
 	helm uninstall loki
 
 launch-dex:
-	helm repo add stable https://kubernetes-charts.storage.googleapis.com/
+	helm repo add --force-update stable https://kubernetes-charts.storage.googleapis.com/
```
actgardner, line 396: I badly misunderstood the feedback on my last PR, this should be better :/

etc/testing/auth/dex.yaml
```diff
@@ -18,18 +18,27 @@ config:
     tlsKey: /etc/dex/tls/https/server/tls.key
     allowedOrigins: []
   staticClients:
+  - id: testapp
+    name: "Test App"
+    redirectURIs:
+    - 'http://test.example.com:657/authorization-code/callback'
+    secret: test
```
actgardner, line 25: Add the new app that requests the token. IRL this would be Notebooks or something similar, and the user would do a normal auth code flow.

src/server/auth/server/testing/oidc_test.go
```diff
@@ -57,7 +60,7 @@ func TestOIDCAuthCodeFlow(t *testing.T) {
 	// to the login page. The params have the session state. POST our hard-coded
 	// credentials to the login page.
 	vals := make(url.Values)
-	vals.Add("login", "admin@example.com")
+	vals.Add("login", "admin")
```
actgardner, line 63: The new username for the mockPassword connector is slightly different from our old, fixed email for no particular reason.

src/server/auth/server/testing/oidc_test.go
```diff
@@ -72,15 +75,51 @@ func TestOIDCAuthCodeFlow(t *testing.T) {
 	require.NoError(t, err)
 
 	// Check that pachd recorded the response from the redirect
-	adminClient.Authenticate(adminClient.Ctx(), &auth.AuthenticateRequest{OIDCState: loginInfo.State})
+	_, err = adminClient.Authenticate(adminClient.Ctx(), &auth.AuthenticateRequest{OIDCState: loginInfo.State})
+	require.NoError(t, err)
```
actgardner, line 79: This used to be a useless check because we didn't assign the return value from Authenticate to anything. Whoops! We should consider adding a linter that would catch unchecked error returns.

src/server/auth/server/testing/oidc_test.go
```diff
@@ -72,15 +75,51 @@ func TestOIDCAuthCodeFlow(t *testing.T) {
...
+		"client_id":     []string{"testapp"},
+		"client_secret": []string{"test"},
+		"grant_type":    []string{"password"},
+		"scope":         []string{"openid email profile audience:server:client_id:pachyderm"},
+		"username":      []string{"admin"},
+		"password":      []string{"password"},
+	}
+
+	c := &http.Client{}
+	resp, err := c.PostForm(fmt.Sprintf("http://%v/token", dexHost(adminClient)), vals)
```
actgardner, line 106: The ROP Flow request (because `grant_type=password`). Don't try this at home.

adelelopez (approved):
LGTM

msteffen:
(I just rebased the two commits in this PR onto master to fix the merge conflict in `Makefile` and force-pushed to this branch to make this PR mergeable. The issue is just that at some point `helm repo add ... https://kubernetes-charts.storage.googleapis.com/` started giving us errors saying we should switch to `https://charts.helm.sh/stable`, so we did. I just had to use the new URL in this PR)

src/server/auth/server/testing/oidc_test.go
msteffen: My only feedback re:these tests was that 1) ideally a client with no credentials would go through the the auth flow, just to be sure that works, and b) ideally we'd check the token that's returned by using it for `Authorize()` or `WhoAmI()` or something like that.

I pushed a small commit to this branch to make that change

msteffen (approved):
Two comments. I had one request (the other comment is just curiosity), but I just pushed a commit to the branch to make the requested change—hopefully that's OK. I figured with that commit, the PR LGTM and that way I could still merge it.

Obviously I'd be happy to merge a follow-up PR to make any changes necessitated by the extra commit.

etc/testing/auth/dex.yaml
```diff
@@ -18,18 +18,27 @@ config:
...
+    trustedPeers:
+    - testapp
+  connectors:
+  - id: test
+    name: test
+    type: mockPassword
+    config:
+      username: admin
+      password: password
+  enablePasswordDB: false
```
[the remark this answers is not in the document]
actgardner: IIRC I was looking at how dex writes their own unit tests, which led to the mock callback connector (https://github.com/dexidp/dex/blob/bca77245df40601d09f48394135dec8390d59e18/cmd/dex/config_test.go#L94). They're both implemented in the same place: https://github.com/dexidp/dex/blob/c113df2730052e20881dd68561289f8ae121300b/connector/mock/connectortest.go

actgardner:
Looks good, thanks for the improvements. I'll merge it once CI is green.

PR #5339, opened 2020-10-12
Adds Chunk Tracker for storing chunk metadata

brendoncarroll:
Adds a Chunk Tracker for storing chunk metadata, and mediating chunk operations during creation and deletion.
Also creates a Client object in the chunk package which knows how to interact with a Tracker and object store to ensure consistency.


src/server/pkg/storage/chunk/metadata.go
```diff
@@ -0,0 +1,89 @@
...
+	SetChunkMetadata(ctx context.Context, chunkID ChunkID, md ChunkMetadata) error
+	// GetChunkInfo returns info about the chunk if it exists
+	GetChunkMetadata(ctx context.Context, chunkID ChunkID) (*ChunkMetadata, error)
+	// DeleteChunkInfo removes chunk metadata from the tracker
+	DeleteChunkMetadata(ctx context.Context, chunkID ChunkID) error
+}
+
+var _ MetadataStore = &PGStore{}
+
+type PGStore struct {
```
brycemcanally, line 43: I prefer this to be spelled out as `PostgresStore`.

src/server/pkg/storage/chunk/chunk.proto
```diff
@@ -6,7 +6,7 @@ option go_package = "github.com/pachyderm/pachyderm/src/server/pkg/storage/chunk
 // DataRef is a reference to data within a chunk.
 message DataRef {
   // The chunk the referenced data is located in.
-  ChunkInfo chunk_info = 1;
+  ChunkRef chunk_ref = 1;
```
brycemcanally, line 9: I know I said this change is fine, but I am not sure about it now. It seems like it could be confusing since we call data refs chunk refs when they refer to the full chunk. Why did you want to change this? Something to note is that the ChunkInfo / Chunk pattern lines up pretty well with the general proto naming convention we follow in pfs / pps (i.e. CommitInfo / Commit).

src/server/pkg/storage/chunk/client.go
```diff
@@ -0,0 +1,173 @@
...
+		return chunkID, nil
+	}
+	if err := c.mdstore.SetChunkMetadata(ctx, chunkID, md); err != nil {
+		return nil, err
+	}
+	objW, err := c.objc.Writer(ctx, p)
+	if err != nil {
+		return nil, err
+	}
+	defer objW.Close()
```
brycemcanally, line 93: We should capture any errors that occur here. Also, why are we closing twice?

src/server/pkg/storage/tracker/gc.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (dm DeleterMux) Delete(ctx context.Context, id string) error {
+	deleter := dm(id)
+	if deleter == nil {
+		return errors.Errorf("deleter mux does not have deleter for (%s)", id)
+	}
+	return deleter.Delete(ctx, id)
+}
+
+type GC struct {
```
brycemcanally, line 25: I also would prefer this to be spelled out as `GarbageCollector`.

src/server/pfs/server/driver_v2.go
```diff
@@ -62,37 +63,57 @@ func newDriverV2(env *serviceenv.ServiceEnv, txnEnv *txnenv.TransactionEnv, etcd
...
 	go d2.compactionWorker()
 	return d2, nil
 }
 
-func newDB() (*gorm.DB, error) {
+func newDB() (db *sqlx.DB, retErr error) {
+	defer func() {
+		if db != nil {
+			db.MustExec(`DROP SCHEMA IF EXISTS storage CASCADE`)
+			fileset.PGStoreApplySchema(db)
```
brycemcanally, line 86: Minor nit, but can we us the word `Create` rather than `Apply`. That lines up more with the Postgres operation name.

src/server/pkg/dbutil/dbutil.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func NewDB(x DBParams) (*sqlx.DB, error) {
+	dsn := fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=disable", x.Host, x.Port, x.User, x.Pass, x.DBName)
+	return sqlx.Open("postgres", dsn)
+}
+
+// NewGORMDB creates a database client.
+// TODO: Remove GOARM and switch to sql x.
+func NewGORMDB(host, port, user, pass, dbName string) (*gorm.DB, error) {
```
brycemcanally, line 54: This is going to be removed in this PR right? If I recall correctly it was just the old garbage collector that needed this.

src/server/pkg/dbutil/dbutil.go
```diff
@@ -0,0 +1,65 @@
...
+
+// WithTestDB connects to postgres using the default settings, creates a database with a unique name
+// then calls cb with a sqlx.DB configured to use the newly created database.
+// After cb returns the database is dropped.
+func WithTestDB(t testing.TB, cb func(db *sqlx.DB)) {
+	dsn := fmt.Sprintf("host=%s port=%d user=%s sslmode=disable", DefaultPostgresHost, DefaultPostgresPort, TestPostgresUser)
+	db := sqlx.MustOpen("postgres", dsn)
+	dbName := fmt.Sprintf("test_%d", time.Now().UnixNano())
+	db.MustExec("CREATE DATABASE " + dbName)
+	t.Log("database", dbName, "successfully created")
```
brycemcanally, line 30: Was this just for debugging?

src/server/pkg/dbutil/dbutil.go
```diff
@@ -0,0 +1,65 @@
...
+	db2 := sqlx.MustOpen("postgres", dsn+" dbname="+dbName)
+	cb(db2)
```
brycemcanally, line 32: We probably want the testing object passed into the callback.

src/server/pkg/dbutil/dbutil.go
```diff
@@ -0,0 +1,65 @@
...
+	require.Nil(t, db2.Close())
+	if !devDontDropDatabase {
+		db.MustExec("DROP DATABASE " + dbName)
+	}
+	require.Nil(t, db.Close())
```
brycemcanally, line 37: We should probably use the `NoError` check here and in general.

src/server/pkg/dbutil/dbutil.go
```diff
@@ -0,0 +1,65 @@
...
+func WithTestDB(t testing.TB, cb func(db *sqlx.DB)) {
+	dsn := fmt.Sprintf("host=%s port=%d user=%s sslmode=disable", DefaultPostgresHost, DefaultPostgresPort, TestPostgresUser)
+	db := sqlx.MustOpen("postgres", dsn)
+	dbName := fmt.Sprintf("test_%d", time.Now().UnixNano())
+	db.MustExec("CREATE DATABASE " + dbName)
+	t.Log("database", dbName, "successfully created")
+	db2 := sqlx.MustOpen("postgres", dsn+" dbname="+dbName)
+	cb(db2)
+	require.Nil(t, db2.Close())
+	if !devDontDropDatabase {
```
brycemcanally, line 34: Probably should put this in a defer before the callback.

src/server/pkg/storage/chunk/metadata.go
```diff
@@ -0,0 +1,89 @@
...
+func (id ChunkID) HexString() string {
+	return hex.EncodeToString(id)
+}
+
+type ChunkMetadata struct {
+	Size     int
+	PointsTo []ChunkID
+}
+
+type MetadataStore interface {
```
brycemcanally, line 32: I kind of think we should be more consistent between the fileset layer and chunk layer in terms of the store naming convention. Right now we have a notion of a fileset store which will store metadata about a fileset including its root index, and a chunk metadata store which stores metadata about chunks. I think it might make sense to call the fileset store a metadata store as well. Another change related to this is that it is not idiomatic Go to have getters prefixed with get, and I think the `Metadata` suffix is a bit redundant considering the type has metadata in the name. So, function names like `SetChunk`, `Chunk`, `DeleteChunk` seem better to me. Also, it might even be clean to change `ChunkMetadata` -> `Metadata` since it is implicit through the package name what the metadata is associated with. Also, I think the fileset table should be called filesets rather than paths.

src/server/pkg/storage/fileset/index/util.go
```diff
@@ -30,3 +34,14 @@ func Generate(s string) []string {
 	sort.Strings(fileNames)
 	return fileNames
 }
+
+// IndexPointsTo returns a list of all the chunks this index references
+func IndexPointsTo(idx *Index) (ids []chunk.ChunkID) {
```
brycemcanally, line 39: I like this helper function, but it probably would be cleaner to just be `PointsTo`.

src/server/pkg/storage/fileset/pg_store.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	return idx, nil
+}
+
+func (s *pgStore) Walk(ctx context.Context, prefix string, cb func(string) error) error {
+	rows, err := s.db.QueryContext(ctx, `SELECT path from storage.paths WHERE path LIKE $1 || '%'`, prefix)
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
```
brycemcanally, line 61: We should probably capture any errors here.

src/server/pkg/storage/chunk/chunk.proto
brendoncarroll: The identifying information for the chunk is not going to change from it's hash.  So we are never going to realize the benefit of having a Chunk object that we can pack more information into over time. I call it a Ref because I envision it being exchanged for chunk data by Create and Get functions which will deal with compression/encryption, and operate on a chunk.Client.

src/server/pkg/storage/fileset/pg_store.go
brycemcanally: Hmm, it looks like you might not have to call `Close` based on the documentation, but it doesn't seem clear to me what happens when errors are encountered. It might be safer to have something like this:
```
defer func() {
  if err := rows.Close(); retErr == nil {
    retErr = err
  }
  if err := rows.Err(); retErr == nil {
    retErr = err
  }
}
```

src/server/pkg/storage/chunk/client.go
brendoncarroll: Yeah I guess we don't need to.

src/server/pfs/server/driver_v2.go
brendoncarroll: It might not just be create at some point, it could include other migrations. What about `SetupPostgresStore(db)` ?
This will likely change as the database migration stuff is fleshed out.

src/server/pkg/dbutil/dbutil.go
brendoncarroll: It's so you have access to the database name in the test.

brendoncarroll: It is for debugging, but I think it's good to have in the test logs all the time.

src/server/pkg/storage/chunk/metadata.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"context"
+	"crypto/sha512"
+	"encoding/hex"
+
+	"github.com/jmoiron/sqlx"
+)
+
+type ChunkID []byte
```
brycemcanally, line 11: Similar to what I mentioned below, I think we should lean on the package name for type namespacing when it makes sense. It is cleaner and idiomatic Go. So, I think this could just be `chunk.ID`.

src/server/pkg/dbutil/dbutil.go
brendoncarroll: I don't think that sort of threading through is necessary. I can't foresee a situation where it would be useful to modify the testing object or produce another one and then inject it into the callback.  The reason we pass it into the outer function is to make test failures during setup easier to debug. (they create failed assertions and short circuit the test, just like everything else)

src/server/pkg/storage/fileset/pg_store.go
brendoncarroll: I can't be the first person to think of this, but this is great
```go
defer errutil.SetRetErrIfNil(&retErr, rows.Close())
```

brendoncarroll: might need to wrap in function, but still it's very convenient.

brendoncarroll: yeah you have to 
```go
defer func() { errutil.SetRetErrIfNil(&retErr, rows.Close()) }()
```
to prevent the arguments from being evaluated

src/server/pkg/storage/chunk/metadata.go
brendoncarroll: In the chunk package we should keep the name `MetadataStore`. Rename the methods to `Set`, `Get`, and `Delete`.  I agree with `ChunkMetadata` to `Metadata`.

I'm less sure about the fileset package.  There is only one store for the filesets, the information isn't really spread between 2 stores, the way chunks are. (object storage, and metadata in postgres)

src/server/pkg/storage/tracker/tracker.go
```diff
@@ -0,0 +1,150 @@
...
+	// It errors if id has other objects referencing it.
+	// Marking something as a tombstone which is already a tombstone is not an error
+	MarkTombstone(ctx context.Context, id string) error
+
+	// FinishDelete deletes the object
+	// It is an error to call DeleteObject without calling MarkTombstone.
+	FinishDelete(ctx context.Context, id string) error
+
+	// IterateExpired calls cb with all the objects which have expired or are tombstones.
+	IterateExpired(ctx context.Context, cb func(id string) error) error
```
brycemcanally, line 55: This should probably be renamed. It is iterating over the unreferenced objects that have expired, not all of the expired objects. This should probably be something that points more to being eligible for deletion, so maybe `IterateDeletes`?

src/server/pkg/storage/tracker/gc.go
```diff
@@ -0,0 +1,93 @@
...
+
+func NewGarbageCollector(tracker Tracker, period time.Duration, deleter Deleter) *GarbageCollector {
+	return &GarbageCollector{
+		tracker: tracker,
+		period:  period,
+		deleter: deleter,
+	}
+}
+
+func (GarbageCollector *GarbageCollector) Run(ctx context.Context) error {
```
brycemcanally, line 39: We don't need to spell out `GarbageCollector` for the local variables, just the struct type name. The general pattern I like to follow, and I think that you usually follow, is a letter per word in the struct type name.

src/server/pkg/storage/tracker/pg_tracker.go
```diff
@@ -0,0 +1,222 @@
...
+	"database/sql"
+	"time"
+
+	"github.com/jmoiron/sqlx"
+	"github.com/lib/pq"
+)
+
+var _ Tracker = &PGTracker{}
+
+type PGTracker struct {
```
brycemcanally, line 14: `PGTracker` -> `PostgresTracker`

src/server/pkg/storage/chunk/metadata.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func NewPostgresStore(db *sqlx.DB) *PostgresStore {
+	return &PostgresStore{db: db}
+}
+
+func (s *PostgresStore) Set(ctx context.Context, chunkID ID, md Metadata) error {
+	_, err := s.db.ExecContext(ctx,
+		`INSERT INTO storage.chunks (hash_id, size) VALUES ($1, $2)
+		ON CONFLICT (hash_id) DO UPDATE SET size = $2 WHERE storage.chunks.hash_id = $1
```
brycemcanally, line 54: When would the size of a chunk change?

brendoncarroll: It wouldn't but since the method is called `Set`, It should have set/put semantics.

src/server/pkg/storage/chunk/metadata.go
brycemcanally: Okay, I guess I am fine with the Postgres stores being `fileset.Store` and `chunk.MetadataStore`.

src/server/pkg/storage/fileset/pg_store.go
brycemcanally: I am a bit concerned about this negatively impacting readability. Lets not do this for the time being.

src/server/pkg/storage/chunk/metadata.go
brycemcanally: We should probably error if there is a conflict. I think we only want to set the hash and size once, and potentially update other fields (such as last accessed for archiving purposes). We probably need to draw a distinction between the queries that set the immutable fields and the queries that set the mutable fields.

src/server/pkg/storage/chunk/client.go
```diff
@@ -0,0 +1,172 @@
...
+	"time"
+
+	"github.com/pachyderm/pachyderm/src/client/pkg/errors"
+	"github.com/pachyderm/pachyderm/src/server/pkg/obj"
+	"github.com/pachyderm/pachyderm/src/server/pkg/storage/tracker"
+)
+
+// Client allows manipulation of individual chunks, by maintaining consistency between
+// a tracker and an obj.Client.
+type Client struct {
```
brycemcanally, line 20: I still think the creation of temporary objects and renewal should be refactored into a general purpose abstraction in the tracker package.

src/server/pfs/server/driver_v2.go
```diff
@@ -62,37 +63,57 @@ func newDriverV2(env *serviceenv.ServiceEnv, txnEnv *txnenv.TransactionEnv, etcd
...
 	go d2.compactionWorker()
 	return d2, nil
 }
 
-func newDB() (*gorm.DB, error) {
+func newDB() (db *sqlx.DB, retErr error) {
+	defer func() {
+		if db != nil {
+			db.MustExec(`DROP SCHEMA IF EXISTS storage CASCADE`)
+			fileset.SetupPostgresStore(db)
```
brycemcanally, line 86: It might be cleaner to just have all three of these be `SetupPostgres` and scope them through the package naming.

brycemcanally: Also it might make more sense to execute this if retErr == nil, rather than checking the db handle. Although I don't really think that would change the behavior as is, I think it is a better pattern to follow for potential changes in the future.

src/server/pfs/server/driver_v2.go
```diff
@@ -534,6 +555,9 @@ func (d *driverV2) compactionWorker() {
 				Upper: shard.Range.Upper,
 			}
 			_, err = d.storage.Compact(ctx, shard.OutputPath, shard.Compaction.InputPrefixes, defaultTTL, index.WithRange(pathRange))
+			if err != nil {
```
brycemcanally, line 558: We should not panic here. Just return the error.

src/server/pkg/dbutil/dbutil.go
```diff
@@ -0,0 +1,53 @@
...
+	dbName := fmt.Sprintf("test_%d", time.Now().UnixNano())
+	db.MustExec("CREATE DATABASE " + dbName)
+	t.Log("database", dbName, "successfully created")
+	t.Cleanup(func() {
+		if !devDontDropDatabase {
+			db.MustExec("DROP DATABASE " + dbName)
+		}
+		require.NoError(t, db.Close())
+	})
+	db2 := sqlx.MustOpen("postgres", dsn+" dbname="+dbName)
```
brycemcanally, line 36: Why two handles?

src/server/pkg/errutil/errutil.go
```diff
@@ -61,3 +61,11 @@ func IsInvalidPathError(err error) bool {
 	return strings.Contains(err.Error(), "only printable ASCII characters allowed") ||
 		strings.Contains(err.Error(), "not allowed in path")
 }
+
+// SetRetErrIfNil is used to set the returned error (conventionally retErr) to
+// err if it is not already set. This is useful in defers.
+func SetRetErrIfNil(retErr *error, err error) {
```
brycemcanally, line 67: Lets remove this for now.

src/server/pkg/obj/util.go
```diff
@@ -2,15 +2,34 @@ package obj
...
+	})
+	objC, err := NewLocalClient(dir)
+	require.Nil(t, err)
+	return objC
+}
+
 // WithLocalClient constructs a local object storage client for testing during the lifetime of
 // the callback.
 func WithLocalClient(f func(objC Client) error) (retErr error) {
+	fmt.Println("WithLocalClient implements a pattern deprecated since go 1.14. consider switching to NewTestClient")
```
brycemcanally, line 32: This seems like it should be a TODO (or at least just add a TODO).

src/server/pkg/storage/chunk/client.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil, err
+	}
+	objW, err := c.objc.Writer(ctx, p)
+	if err != nil {
+		return nil, err
+	}
+	if _, err = objW.Write(chunkData); err != nil {
+		return nil, err
+	}
+	if err := objW.Close(); err != nil {
```
brycemcanally, line 76: I think we should have kept the deferred close (but also capture the error state). We are going to want to close on error to be sure resources are cleaned up. We might need to take a careful pass over our clients to make sure we won't be in a situation where an object gets persisted if the writer errored, but I think that is the pattern we decided on a while back.

src/server/pkg/storage/chunk/client.go
```diff
@@ -0,0 +1,128 @@
...
+	return chunkID, nil
+}
+
+func (c *Client) Get(ctx context.Context, chunkID ID, w io.Writer) error {
+	p := chunkPath(chunkID)
+	objR, err := c.objc.Reader(ctx, p, 0, 0)
+	if err != nil {
+		return err
+	}
+	defer objR.Close()
```
brycemcanally, line 88: Probably should capture any errors here.

src/server/pkg/storage/chunk/client.go
```diff
@@ -0,0 +1,128 @@
...
+
+func chunkPath(chunkID ID) string {
+	if len(chunkID) == 0 {
+		panic("chunkID cannot be empty")
+	}
+	return path.Join(prefix, chunkID.HexString())
+}
+
+func ChunkObjectID(chunkID ID) string {
+	return "chunk/" + chunkID.HexString()
```
brycemcanally, line 105: This prefix should probably be a const and in one place. I see it referenced in a few places and it looks like it is a local const in the deleter (although the one here ends with an s, which I think may be a bug).

src/server/pkg/storage/chunk/storage.go
```diff
@@ -37,75 +38,28 @@ func NewStorage(objClient obj.Client, opts ...StorageOption) *Storage {
 
 // NewReader creates a new Reader.
 func (s *Storage) NewReader(ctx context.Context, dataRefs []*DataRef) *Reader {
-	return newReader(ctx, s.objClient, dataRefs)
+	// using the empty chunkset for the reader
```
brycemcanally, line 41: This comment seems unnecessary.

src/server/pkg/storage/chunk/util.go
```diff
@@ -2,16 +2,22 @@ package chunk
...
 
-// WithLocalStorage creates a local storage instance for testing during the lifetime of
+// NewTestStorage creates a local storage instance for testing during the lifetime of
 // the callback.
-func WithLocalStorage(f func(obj.Client, *Storage) error, opts ...StorageOption) error {
-	return obj.WithLocalClient(func(objClient obj.Client) error {
-		return f(objClient, NewStorage(objClient, opts...))
-	})
+func NewTestStorage(t testing.TB, db *sqlx.DB, tr tracker.Tracker, opts ...StorageOption) (obj.Client, *Storage) {
+	// metadata store
```
brycemcanally, line 15: The two comments in here are unnecessary.

src/server/pkg/storage/fileset/pg_store_test.go
```diff
@@ -0,0 +1,14 @@
+package fileset
+
+import (
+	"testing"
+
+	"github.com/pachyderm/pachyderm/src/server/pkg/dbutil"
+)
+
+func TestPGStore(t *testing.T) {
```
brycemcanally, line 9: `PG` -> `Postgres`. This change should be applied across all names (including files).

src/server/pkg/storage/chunk/util.go
```diff
@@ -2,16 +2,22 @@ package chunk
...
+	db.MustExec(schema)
```
brycemcanally, line 16: This should probably be replaced with a function similar to `fileset.NewTestStore`.

src/server/pkg/storage/chunk/metadata.go
```diff
@@ -0,0 +1,89 @@
...
+		`, chunkID, md.Size)
+	return err
+}
+
+func (s *PostgresStore) Get(ctx context.Context, chunkID ID) (*Metadata, error) {
+	type chunkRow struct {
+		size int `db:"size"`
+	}
+	var x chunkRow
+	if err := s.db.GetContext(ctx, &x, `SELECT size FROM storage.chunks WHERE hash_id = $1`, chunkID); err != nil {
```
brycemcanally, line 64: We might want something like `ErrChunkNotExists` similar to `ErrPathNotExists`

src/server/pkg/storage/fileset/pg_store.go
```diff
@@ -0,0 +1,105 @@
...
+		idx = &index.Index{}
+	}
+	data, err := proto.Marshal(idx)
+	if err != nil {
+		return err
+	}
+	_, err = s.db.ExecContext(ctx,
+		`INSERT INTO storage.paths (path, index_pb)
+		VALUES ($1, $2)
+		ON CONFLICT (path) DO UPDATE SET index_pb = $2
```
brycemcanally, line 36: Similar to uploading the same chunk, it might make more sense for this to be an error. Particularly if the pb is supposed to be immutable.

src/server/pkg/storage/fileset/storage.go
```diff
@@ -316,74 +300,60 @@ func (s *Storage) compactSpec(ctx context.Context, fileSet string, compactedFile
...
+		if err := s.store.Delete(ctx, name); err != nil {
 			return err
 		}
-		return s.chunks.DeleteSemanticReference(ctx, name)
+		return s.tracker.MarkTombstone(ctx, oid)
 	})
 }
 
 // WalkFileSet calls f with the path of every primitive fileSet under prefix.
 func (s *Storage) WalkFileSet(ctx context.Context, prefix string, f func(string) error) error {
```
brycemcanally, line 313: Nit, `f` -> `cb`

src/server/pkg/storage/fileset/storage.go
```diff
@@ -316,74 +300,60 @@ func (s *Storage) compactSpec(ctx context.Context, fileSet string, compactedFile
 
 // Delete deletes a fileset.
 func (s *Storage) Delete(ctx context.Context, fileSet string) error {
-	fileSet = applyPrefix(fileSet)
-	return s.objC.Walk(ctx, fileSet, func(name string) error {
-		if err := s.objC.Delete(ctx, name); err != nil {
+	return s.store.Walk(ctx, fileSet, func(name string) error {
+		oid := filesetObjectID(name)
+		if err := s.store.Delete(ctx, name); err != nil {
```
brycemcanally, line 305: We can get this in a later PR, but we are going to want this to just be mark tombstone on the prefix (fileSet). The file set deleter and tracker objects we setup should handle the recursive deletion of sub file sets, then ultimately the chunks.

src/server/pkg/storage/renew/renewer.go
```diff
@@ -0,0 +1,83 @@
...
+	return r
+}
+
+// Context returns a context which will be cancelled when the renewer is closed
+func (r *Renewer) Context() context.Context {
+	return r.ctx
+}
+
+// IsClosed returns true if the renewer is closed
+func (r *Renewer) IsClosed() bool {
```
brycemcanally, line 45: Is this going to be used somewhere?

src/server/pkg/storage/renew/renewer.go
```diff
@@ -0,0 +1,83 @@
...
+	ttl       time.Duration
+
+	ctx    context.Context
+	cancel context.CancelFunc
+	done   chan struct{}
+	err    error
+}
+
+func NewRenewer(ttl time.Duration, renewFunc RenewFunc) *Renewer {
+	ctx, cancel := context.WithCancel(context.Background())
```
brycemcanally, line 22: We should have a passed in context here, not the background.

src/server/pkg/storage/renew/stringset.go
```diff
@@ -0,0 +1,64 @@
...
+	defer ss.mu.Unlock()
+	delete(ss.set, x)
+}
+
+// WithStringSet creates a StringSet using ttl and rf. It calls cb with the StringSets context and the new StringSet.
+// If ctx is cancelled, the StringSet will be Closed, and the cancellation will propagate down to the context passed to cb.
+func WithStringSet(ctx context.Context, ttl time.Duration, rf RenewStringFunc, cb func(ctx context.Context, ss *StringSet) error) error {
+	ss := NewStringSet(ttl, rf)
+	defer ss.Close()
+	go func() {
```
brycemcanally, line 53: I think if `ss.Context()` is derived from the passed in context, then you don't need this. We would just want a deferred close, and capture the error.

src/server/pkg/storage/track/gc.go
```diff
@@ -0,0 +1,93 @@
...
+		period:  period,
+		deleter: deleter,
+	}
+}
+
+func (gc *GarbageCollector) Run(ctx context.Context) error {
+	ticker := time.NewTicker(gc.period)
+	defer ticker.Stop()
+	for {
+		ctx, cf := context.WithTimeout(ctx, gc.period/2)
```
brycemcanally, line 43: It would be good to follow the same pattern as the renewer where you have a wrapper function so you can defer the cancel function and just return from runUntilEmpty. Also, we discussed this offline, but lets have this always return an error if it returns, which I think should help us address the messiness in the pfs master. We also might want a retry in this code path somewhere, I am guessing where we do the `IterateDeletable`.

src/server/pkg/storage/track/pg_tracker.go
```diff
@@ -0,0 +1,222 @@
...
+	for _, dwn := range pointsTo {
+		if dwn == id {
+			return ErrSelfReference
+		}
+	}
+	return t.withTx(ctx, func(tx *sqlx.Tx) error {
+		var oid int
+		if err := func() error {
+			if ttl > 0 {
+				return tx.GetContext(ctx, &oid,
```
brycemcanally, line 32: The lowest ttl should probably be a second, and we should error similarly to the way we do through the renewal pfs api.

src/server/pkg/storage/track/pg_tracker.go
```diff
@@ -0,0 +1,222 @@
...
+	if err == sql.ErrNoRows {
+		return nil
+	}
+	return nil
+}
+
+func (t *PostgresTracker) IterateDeletable(ctx context.Context, cb func(id string) error) error {
+	rows, err := t.db.QueryxContext(ctx,
+		`SELECT str_id FROM storage.tracker_objects
+		WHERE (expires_at <= CURRENT_TIMESTAMP OR tombstone)
```
brycemcanally, line 174: We should still restructure this such that we first get the set of objects that are not referenced, then delete if the object is expired or there is a tombstone.

src/server/pkg/storage/track/pg_tracker.go
```diff
@@ -0,0 +1,222 @@
...
+		AND int_id NOT IN (SELECT DISTINCT to_id FROM storage.tracker_refs)`)
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
```
brycemcanally, line 179: Capture the error here.

src/server/pkg/storage/track/tracker.go
```diff
@@ -0,0 +1,146 @@
...
+	// It errors with ErrDanglingRef if any of the elements in pointsTo do not exist
+	// It errors with ErrTombstone if the object exists and is marked as a tombstone. Callers should retry until
+	// they successfully create the object.
+	// CreateObject should always be called *before* adding an object to an auxillary store.
+	CreateObject(ctx context.Context, id string, pointsTo []string, ttl time.Duration) error
+
+	// SetTTLPrefix sets the expiration time to current_time + ttl for all objects with ids starting with prefix
+	SetTTLPrefix(ctx context.Context, prefix string, ttl time.Duration) (time.Time, error)
+
+	// TODO: thoughts on these?
```
brycemcanally, line 34: I think the interface we have now makes sense.

src/server/pkg/storage/track/tracker.go
```diff
@@ -0,0 +1,146 @@
...
+	GetUpstream(ctx context.Context, id string) ([]string, error)
+
+	// MarkTombstone causes the object to appear deleted.  It cannot be created, and objects referencing it cannot
+	// be created until it is deleted.
+	// It errors if id has other objects referencing it.
+	// Marking something as a tombstone which is already a tombstone is not an error
+	MarkTombstone(ctx context.Context, id string) error
+
+	// FinishDelete deletes the object
+	// It is an error to call DeleteObject without calling MarkTombstone.
```
brycemcanally, line 51: `DeleteObject` -> `FinishDelete`?

src/server/pkg/storage/track/tracker.go
```diff
@@ -0,0 +1,146 @@
...
+	FinishDelete(ctx context.Context, id string) error
+
+	// IterateDeletable calls cb with all the objects which have expired or are tombstones.
```
brycemcanally, line 54: This should maybe be worded as, `objects which are no longer referenced and have expired or are tombstoned`

src/server/pfs/server/driver_v2.go
brendoncarroll: All of this setup will change with the migrations stuff I'm working on.

src/server/pkg/dbutil/dbutil.go
brendoncarroll: Each db is scoped to a postgres database.  The database is part of the DSN.  The first db is for setting up the database, then db2 is created scoped to that database, then db does the teardown.

src/server/pkg/storage/chunk/client.go
brendoncarroll: I'll add it back. We should change the obj.Client to `Write(ctx, path, io.Reader)` so that non EOF's can signal an abort.  Or have close alternative called `Abort`

src/server/pkg/storage/chunk/util.go
brendoncarroll: We need a way to inject the tracker, since in order for filesets to depend on chunks they have to share a tracker. It also makes sense to check that everything works on a single DB.

src/server/pkg/storage/track/tracker.go
brendoncarroll: Yeah if I saw DeleteObject in my autocomplete I would assume it does what it says. By using FinishDelete it lets a potential caller know they have to do something else to start the deletion process (MarkTombstone)

src/server/pkg/storage/track/pg_tracker.go
brendoncarroll: I don't think we need to enforce that constraint at this level. Postgres supports intervals with resolution 1 microsecond.    I would argue only the system up a level can make reasonable assumptions about minimum latency.

src/server/pkg/storage/chunk/storage.go
brendoncarroll: This is worth noting because the empty string disables the renewer.  Where should we document that so someone doesn't come along and add a more descriptive string trying to match what we do for the writer?

src/server/pkg/storage/chunk/util.go
brendoncarroll: See the function `newTestStorage` in `chunk_test.go`, which only takes `testing.TB` and is used for the chunk package tests, but would not be useful for export.

src/server/pkg/storage/chunk/storage.go
brycemcanally: I'm just saying the comment does not tell you any more than you can infer from the usage. You could add a comment that says why we use the empty chunk set, which could provide some value.

src/server/pkg/storage/track/pg_tracker.go
brycemcanally: It is not really a big deal, I just cannot imagine needing a ttl less than a second for anything and having a check for anything less than that might help catch an error if something incorrectly sets the ttl.

src/server/pkg/storage/fileset/storage.go
```diff
@@ -316,74 +299,61 @@ func (s *Storage) compactSpec(ctx context.Context, fileSet string, compactedFile
...
+func (s *Storage) GC(ctx context.Context) error {
+	const period = 10 * time.Second
+	chunkDeleter := s.chunks.NewDeleter()
+	filesetDeleter := &deleter{
+		store: s.store,
 	}
-	return fileSet[len(semanticPrefix):]
+	mux := track.DeleterMux(func(id string) track.Deleter {
+		switch {
+		case strings.HasPrefix(id, "chunk/"):
```
brycemcanally, line 343: These prefixes should be consts in the relevant packages.

src/server/pkg/storage/track/tracker.go
brycemcanally: I was just referring to the comment needing to be updated. I agree with the name `FinishDelete`

src/server/pkg/storage/chunk/util.go
brycemcanally: I was just referring to the schema setup and store constructor. For file sets, we have one function (`NewTestStore`) that does both, which takes a db and returns a store.

src/server/pkg/storage/track/gc.go
brendoncarroll: The GC loop will log any errors and the ticker keeps going, so I don't think we need another retry loop right now.

src/server/pkg/storage/track/pg_tracker.go
brendoncarroll: If we are going to have an arbitrary limit I think it should only be enforced in one place (right now the PFS driver). At this level the worst thing a small interval would cause is and object that expires before the caller can do anything with it. Also setting the ttl to a nanosecond could be useful for testing.

brycemcanally: Yeah, so if we had an internal call that set it for less than a second, then it is probably a bug that would get caught pretty obviously by a check for less than a second.

brycemcanally: Discussed offline, we will leave it as a TODO for now.

brycemcanally (approved):
lgtm

PR #5340, opened 2020-10-12
Add verbose s3 logging 1.11

msteffen:
Add deploy option to enable verbose logging in S3 client

This is the 1.11 version of #5341


msteffen:
Moving to JD as Bryce is OOO

jdoliner (approved):
LGTM

PR #5347, opened 2020-10-14
Refactor temporary fileset renewal and add client wrappers

brycemcanally:
This PR refactors the temporary fileset renewal to make it available client side and adds the client side wrappers. This also shuffles some code around to resolve some dependency issues (mainly packages depending on the client package directly or indirectly).


src/client/pfs/pfs.proto
```diff
@@ -551,7 +551,7 @@ message CreateTmpFileSetResponse {
 
 message RenewTmpFileSetRequest {
   string fileset_id = 1;
-  int64 ttl_seconds = 2;
+  int64 ttl = 2;
```
brendoncarroll, line 554: We should probably add a comment here, explaining the units.

src/server/pfs/server/driver_v2.go
```diff
@@ -1083,6 +1083,9 @@ func (d *driverV2) createTmpFileSet(server pfs.API_CreateTmpFileSetServer) (stri
 }
 
 func (d *driverV2) renewTmpFileSet(ctx context.Context, id string, ttl time.Duration) error {
+	if ttl.Seconds() == 0 {
```
brendoncarroll, line 1086: When we talked offline I meant to do this on the client, and keep the units as seconds. If you are changing it to nanoseconds just do `ttl < time.Second`, which is more clear.

src/client/pfs/pfs.proto
brendoncarroll: Or rename to `ttl_ns`

src/server/pkg/serviceenv/option.go
```diff
@@ -0,0 +1,68 @@
+package serviceenv
```
brendoncarroll, line 1: 👍

src/server/pkg/storage/fileset/merge.go
```diff
@@ -213,7 +213,7 @@ func (fmr *FileMergeReader) Index() *index.Index {
 }
 
 // Header returns the tar header for the merged file.
-func (fmr *FileMergeReader) Header() (*tar.Header, error) {
+func (fmr *FileMergeReader) Header() (_ *tar.Header, retErr error) {
```
brendoncarroll, line 216: Named returns appear unused. Revert.

src/server/pkg/storage/fileset/merge.go
```diff
@@ -243,7 +243,7 @@ func (fmr *FileMergeReader) Header() (*tar.Header, error) {
 }
 
 // WriteTo writes the merged file to the passed in fileset writer.
-func (fmr *FileMergeReader) WriteTo(w *Writer) error {
+func (fmr *FileMergeReader) WriteTo(w *Writer) (retErr error) {
```
brendoncarroll, line 246: Same here as above

src/server/pkg/storage/fileset/option.go
```diff
@@ -54,8 +53,9 @@ func WithMaxOpenFileSets(max int) StorageOption {
 // UnorderedWriterOption configures an UnorderedWriter.
 type UnorderedWriterOption func(f *UnorderedWriter)
 
-// WithRenewer sets the UnorderedWriter's Renewer.
-func WithRenewer(ttl time.Duration, r *Renewer) UnorderedWriterOption {
+// WithRenewal configures the UnorderedWriter to renew subfileset paths
+// with the provided renewer.
+func WithRenewal(ttl time.Duration, r *Renewer) UnorderedWriterOption {
```
brendoncarroll, line 58: 😐

src/client/pfs/pfs.proto
brycemcanally: I'm going to convert this back to seconds and use the check you mention below in the client. The field being seconds makes more sense to me after thinking it over a bit.

src/server/pfs/server/driver_v2.go
brycemcanally: Refer to the comment above.

src/server/pkg/storage/fileset/option.go
brycemcanally: Yeah, I had a bit of a name conflict. I think this is a reasonable change.

brendoncarroll (approved):
LGTM

PR #5351, opened 2020-10-14
Add proto fields for job and pipeline filtering.

PFedak:
I figured I'd make a PR for this since the implementation given the proto change will be rather straightforward (I have a version written). My main reasons for going for the overengineered/YAGNI end of the spectrum are:

- ListJobF etc. pass request fields as arguments, and if we eventually expect more filtering logic putting everything else in a new filter struct avoids more function signature changes
- adding the additional level of indirection makes it easy to tell which fields are present, which shouldn't be true by default (I notice there's very little use of "oneof" in existing protos, which is another way of accomplishing the same thing), but I don't know how sensitive we are to avoiding nested structs

Happy to just do what's needed for filtering by state instead.


msteffen (approved):
LGTM

msteffen (approved):
LGTM lives!

PR #5356, opened 2020-10-16
Refactor RetryUntilCancel and add two utility Notify functions

msteffen:
This is part of a series of changes I'm making to PPS master, but this part was easy to break out, so I'm doing that to spread out the review and make the PPS master PR smaller

The 1.11.x version is #5357

### DIFF
--- a/src/server/pkg/backoff/retry.go
+++ b/src/server/pkg/backoff/retry.go
@@ -43,21 +43,26 @@ func NotifyCtx(ctx context.Context, name string) Notify {
 	}
 }
 
-var Continue = errors.New("looping through backoff")
+// ErrContinue is a sentinel error designed to be used with NotifyContinue.
+// NotifyContinue always returns nil (causing operation() to be retried) when
+// operation() returns ErrContinue. The two combined allow semantics similar to
+// 'continue' in a regular loop: operation() is re-run from the beginning, as a
+// loop's body would be.
+var ErrContinue = errors.New("looping through backoff")
 
 // NotifyContinue is a convenience function for use with RetryUntilCancel. If
 // 'inner' is set to a Notify function, it's called if 'err' is anything other
-// than Continue. If 'inner' is a string or any other non-nil value of another
-// type, any error other than Continue is logged, and the loop is re-run (in
+// than ErrContinue. If 'inner' is a string or any other non-nil value of another
+// type, any error other than ErrContinue is logged, and the loop is re-run (in
 // this case, there is no way to escape the backoff--RetryUntilCancel must be
 // used to avoid an infinite loop). If 'inner' is nil, any error other than
-// Continue is returned.
+// ErrContinue is returned.
 //
 // This is useful for e.g. monitoring functions that want to repeatedly execute
 // the same control loop until their context is cancelled.
 func NotifyContinue(inner interface{}) Notify {
 	return func(err error, d time.Duration) error {
-		if errors.Is(err, Continue) {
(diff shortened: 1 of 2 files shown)


src/server/pkg/backoff/retry.go
```diff
@@ -19,6 +22,54 @@ type Operation func() error
...
+// the same control loop until their context is cancelled.
+func NotifyLoop(inner interface{}) Notify {
+	return func(err error, d time.Duration) error {
+		if errors.Is(err, Loop) {
+			return nil
+		}
+		switch n := inner.(type) {
+		case Notify:
+			return n(err, d) // fallthrough doesn't work for type switches
+		case func(error, time.Duration) error:
```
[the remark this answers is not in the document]
Tryneus: Couldn't you just have the caller pass in a `func(error, time.Duration) error` (or `nil`) instead of an `interface{}`?

src/server/pkg/backoff/retry.go
```diff
@@ -19,6 +22,54 @@ type Operation func() error
...
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			log.Errorf("error in %s: %v: retrying in: %v", name, err, d)
+		}
+		return nil
+	}
+}
+
+var Loop = errors.New("looping through backoff")
```
Tryneus, line 46: I feel like `Continue` or `Retry` might have a clearer meaning for this value.

src/server/pkg/backoff/retry.go
```diff
@@ -49,19 +109,18 @@ func RetryNotify(operation Operation, b BackOff, notify Notify) error {
...
-// RetryUntilCancel is the same as RetryNotify, except that it will not retry if
-// the given context is canceled.
-func RetryUntilCancel(ctx context.Context, operation Operation, b BackOff, notify Notify) error {
-	return RetryNotify(operation, b, func(err error, d time.Duration) error {
 		select {
+		case <-time.After(next):
+			if ctx.Err() != nil {
+				// Corner case: if 'b' is ZeroBackOff, and 'ctx' is cancelled inside of
+				// notify(), then it's random which case is chosen, and the cancellation
+				// won't necessarily prevent opration() from running again. We can't
```
Tryneus, line 117:
```suggestion
				// won't necessarily prevent operation() from running again. We can't
```

src/server/pkg/backoff/retry.go
```diff
@@ -49,19 +109,18 @@ func RetryNotify(operation Operation, b BackOff, notify Notify) error {
...
+				// guarantee that 'operation' is called with a live context, but this
```
Tryneus, line 118: We _could_ guarantee that it's a live context if we do a non-blocking wait on `ctx.Done()` - I think that would be the right thing to do for this corner case.

Tryneus (changes requested):
A few nits/questions, but overall looks pretty good.

src/server/pkg/backoff/retry.go
Tryneus: I mean, yeah, it would always be a race condition in a multi-threaded model, but I feel like having a race condition when there is a single thread of execution is silly.

src/server/pkg/backoff/retry.go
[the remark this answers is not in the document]
msteffen: I do actually need to handle the `nil` case though, which I just added a commit for

Tryneus (approved):
LGTM

PR #5357, opened 2020-10-16
Refactor RetryUntilCancel and add two utility Notify functions

msteffen:
This is the 1.11.x version of #5356


msteffen:
merging, as the corresponding change for `master` (#5356) has already been reviewed and approved

PR #5363, opened 2020-10-19
Fix deadlock in monitorPipeline & refactor

### LINKED ISSUE #4371: standby pipeline not removed after run
**What happened?**:
Sometimes a standby pipeline is not removed. Unfortunately this is not consistent: sometimes it is, sometimes not. I check that `pachctl inspect pipeline xxx --raw | grep standby` and it shows `true`. Of course I check that there was no input to be processed. I can see that after the job is complete, the replicationController remains set to `1` instead of going down to `0`.

**Environment?**:
- Kubernetes version (use `kubectl version`): 1.11.10
- Pachyderm CLI and pachd server version (use `pachctl version`): 1.9.7

msteffen:
The main change is the creation of `monitor.go`, and the convention that functions in `monitor.go` not call outside of monitor.go.

I think it might be possible to refactor the PPS master so that those functions are in their own package and can't call outside the package without introducing a circular dependency, but that seemed more complicated, and may make sense to persue separately.

The 1.11.x version of this PR #5364 

Fixes #4371

### DIFF
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -2192,6 +2192,77 @@ func TestRecreatePipeline(t *testing.T) {
 	createPipeline()
 }
 
+// TestRecreateStandbyPipeline ensures that deleting and recreating a pipeline
+// with standby:true works. Because the PPS master now blocks after cancelling a
+// standby pipeline's 'monitorPipeline' goroutine, successfully recreating a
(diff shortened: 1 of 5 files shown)


nitinjainsj:
Potentially fixes #4371

jdoliner (approved):
LGTM

PR #5366, opened 2020-10-21
Clean glob before extracting capture groups

### LINKED ISSUE #5365: Glob patterns ending in '/' work differently inside and outside joins
Via slack:
"If I run `pachctl glob file A@master:/(*)` and then pachctl glob file `A@master:/(*)/`, though, they both return the same thing, which implies that `/(*)` and `/(*)/` are equivalent."

However, when using `/(*)/` as a capture pattern to join two inputs, we see the cross of the two repos rather than the inner join.

msteffen:
Fixes #5365

This is the 1.11 version of #5367


adelelopez (approved):
LGTM

PR #5367, opened 2020-10-21
Clean glob before extracting capture groups

### LINKED ISSUE #5365: Glob patterns ending in '/' work differently inside and outside joins

msteffen:
Fix #5365


msteffen:
@PFedak noticed that we weren't actually running any of our worker tests! This PR now fixes that as well (`test-worker-helper` now tests `s/s/worker/...`

adelelopez (approved):
LGTM

PR #5370, opened 2020-10-22
Enable worker tests in master

msteffen:
(Credit goes to @PFedak who noticed these tests weren't running in https://github.com/pachyderm/pachyderm/pull/5367 and what we needed to do to fix it)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -335,7 +335,7 @@ test-worker: launch-stats test-worker-helper
 
 test-worker-helper:
 	PROM_PORT=$$(kubectl --namespace=monitoring get svc/prometheus -o json | jq -r .spec.ports[0].nodePort) \
-	  go test -v -count=1 ./src/server/worker/ -timeout $(TIMEOUT)
+	  go test -v -count=1 ./src/server/worker/... -timeout $(TIMEOUT)
 
 clean: clean-launch clean-launch-kube
 
--- a/Makefile
+++ b/Makefile
@@ -334,8 +334,7 @@ test-tls:
 test-worker: launch-stats test-worker-helper
 
 test-worker-helper:
-	PROM_PORT=$$(kubectl --namespace=monitoring get svc/prometheus -o json | jq -r .spec.ports[0].nodePort) \
-	  go test -v -count=1 ./src/server/worker/... -timeout $(TIMEOUT)
+	./etc/testing/test_worker.sh $(TIMEOUT)
 
 clean: clean-launch clean-launch-kube
 
--- a/etc/testing/test_worker.sh
+++ b/etc/testing/test_worker.sh
@@ -2,10 +2,11 @@
 set -euxo pipefail
 
 export TIMEOUT="${1}"
-export PROM_PORT="$(
+PROM_PORT="$(
   kubectl --namespace=monitoring get svc/prometheus -o json \
     | jq -r .spec.ports[0].nodePort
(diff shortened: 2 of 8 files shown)


etc/testing/test_worker.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -euxo pipefail
+
+export TIMEOUT="${1}"
+PROM_PORT="$(
+  kubectl --namespace=monitoring get svc/prometheus -o json \
+    | jq -r .spec.ports[0].nodePort
+)"
```
Tryneus, line 8: This doesn't work on my machine and I'm not sure what I would need to do to get it to work.

Makefile
```diff
@@ -334,8 +334,7 @@ test-tls:
 test-worker: launch-stats test-worker-helper
 
 test-worker-helper:
-	PROM_PORT=$$(kubectl --namespace=monitoring get svc/prometheus -o json | jq -r .spec.ports[0].nodePort) \
-	  go test -v -count=1 ./src/server/worker/ -timeout $(TIMEOUT)
+	./etc/testing/test_worker.sh $(TIMEOUT)
```
Tryneus, line 337: thanks for moving this into a script

etc/testing/test_worker.sh
Tryneus: Also, I commented this out and the tests ran just fine?

etc/testing/test_worker.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -euxo pipefail
+
+export TIMEOUT="${1}"
```
Tryneus, line 4: Would be nice to error with a helpful message if the timeout is omitted (or just have a default value).

Tryneus (changes requested):
I opened #5374 with my suggestions

etc/testing/test_worker.sh
msteffen: Default added by #5374

Makefile
msteffen: :+1:

etc/testing/test_worker.sh
msteffen: This is for TestPrometheusStats: https://github.com/pachyderm/pachyderm/blob/master/src/server/worker/stats/stats_test.go#L93

Peter and I disabled that test before we decided to move this change into its own PR, because deploying Prometheus sounded annoying and TBH I don't even know if anyone uses our Prometheus integration anymore. To get it to work, you need to run `make launch-stats`, which is one of the dependencies of the `test-worker` make target that runs this test.

TBH I'm not sure what the best thing to do here is. If we want to support prometheus stats, then we need to re-add this code and fix the test. If we don't want to support prometheus stats, we should delete `src/server/worker/stats/` and everything that calls it. I'll leave it like this, raise a question in eng, and deal with it in a followup PR

msteffen: Update: per slack, hub uses Prometheus stats. I'll try to fix this test in the PR then. Maybe after these changes are made I'll ask to do another review pass?
- [ ] add a note to `test_worker.sh` clarifying that it should only be run via `make` so that Prometheus is also running.
- [ ] Fix & deflake `TestPrometheusStats`

msteffen:
Closing this PR as obsolete

Issue #5345, opened 2020-10-13
TestPipelineBuildLifecycleGo flake


PFedak:
TestPipelineBuildLifecycleGo failed in https://travis-ci.com/github/pachyderm/pachyderm/jobs/399167346

```
=== RUN   TestPipelineBuildLifecycleGo
../../../../etc/testing/pipeline-build/input/111.txt: 3 B / 3 B  100.00% ? p/s 0
../../../../etc/testing/pipeline-build/input/1.txt: 1 B / 1 B [ 100.00% ? p/s 0s
../../../../etc/testing/pipeline-build/input/11.txt: 2 B / 2 B  100.00% ? p/s 0s
go.mod: 131 B / 131 B [=======================================] 100.00% ? p/s 0s
go.sum: 169 B / 169 B [=======================================] 100.00% ? p/s 0s
main.go: 795 B / 795 B [======================================] 100.00% ? p/s 0s
file /1.txt not found in repo test-pipeline-build at commit d4ec0aec17fa49248db2a5d388e5bfc1
failed to find
  "0001"
in the text:
  """
  """
```

msteffen:
I've been seeing this flake quite a bit as well (especially in the 1.11.x branch—not sure if the underlying issue was fixed in master but not 1.11.x somehow)

msteffen:
I'm now noticing this regularly with TestPipelineBuildLifecyclePython as well

msteffen:
It looks like sometimes the test fails to load the source code into the `source` branch:
```
file /1.txt not found in repo test-pipeline-build at commit a010b8bfd21d4f5280628e0de454c3e0
failed to find
  "___1"
in the text:
  """

  """
--- FAIL: TestPipelineBuildLifecycleGo (46.42s)
    cmds_test.go:630: No error is expected but got exit status 1
    cmds_test.go:630: current stack:
        goroutine 14 [running]:
        runtime/debug.Stack(0x25a47a0, 0xc000480780, 0xc00038bd98)
                /usr/local/go/src/runtime/debug/stack.go:24 +0x9d
        github.com/pachyderm/pachyderm/src/client/pkg/require.fatal(0x262e8c0, 0xc0005db500, 0x0, 0x0, 0x0, 0x2252bca, 0x1f, 0xc000209a10, 0x1, 0x1)
[...]

PR #5376, opened 2020-10-22
Flush proper commit in testPipelineBuildLifecycle

PFedak:


PFedak:
Maaaybe fixes #5345

msteffen (approved):
Thanks for taking a look at this!

TBH I wouldn't expect the semantics of these two calls to be different, but it's worth trying (and this way seems more idiomatic)

msteffen:
To explain my force push: the tests all failed because some kubernetes binary couldn't be downloaded. I thought maybe the remote mirror was temporarily unavailable or something to I rebased this branch onto master and pushed basically just to force CI to re-run.

Unfortunately, this all didn't fix the download issue, which I now think was due to using an old version of Minikube. I think it messed up the commit ownership, which is hopefully not a big deal, but my bad :grimacing:

PR #5380, opened 2020-10-23
Fix tracing in master

msteffen:
master version of #5331


msteffen:
Merging this PR without review as it's identical to 5331 (rebased onto master), which was already reviewed

PR #5381, opened 2020-10-26
Refactor 2.0 load testing and performance / bug fixes

brycemcanally:
This PR refactors the 2.0 load testing to improve the understandability and scalability. It also adds some new load testing functionality and fixes several performance issues and a bug found during the load testing. 

Load testing:

To improve the understandability, the fuzzing model was changed and a load spec is now defined using a YAML file. An example spec is:
```
count: 5
operations:
  - count: 5
    fuzzOperations:
      - operation:
          putFile:
              files:
                  count: 5
                  fuzzFile:
                      - file:
                          source: "random"
                        prob: 1
        prob: 0.7
      - operation:
          deleteFile:
              count: 5
        prob: 0.3 
validator: {}
fileSources:
  - name: "random"
    random:
      fuzzSize:
        - size:
            min: 1000
            max: 10000
          prob: 0.3
        - size:
            min: 10000
            max: 100000
          prob: 0.3
        - size:
            min: 1000000
            max: 10000000
          prob: 0.3
        - size:
            min: 10000000
            max: 100000000
          prob: 0.1
```
Workloads are specified by specifying components that represent parts of a workload (operations, files, sizes) and composing them. The fuzzing model is such that you provide a set of components and a probability per component in place of a regular component (i.e. you can insert a `fuzzOperations` anywhere you can insert a `operation`). Fuzzing components are labeled with `fuzz*` and can be inserted anywhere the regular component can be inserted. Also, the load testing was refactored such that the validation / throttling / simulated failures are implemented by wrapper clients and added a load environment abstraction for global state.

Performance issues / bug fixes:

This PR also fixes several performance issues and a bug that was found during the testing:
- Removed an assertion that checks if a chunk is modified during a read and a copy of the full chunk for a read (these both had a significant impact on small file reads).
- Refactored the compaction such that the compaction property is applied at the top level filesets, rather than the shard filesets (this made it such that we essentially did a full compaction at every step).
- Removed the proto clone from the index accessor for fileset readers (this had a significant impact on small file reads).
- Fixes directory deletion which did not work correctly when performing accesses to only files contained within the directory. This was fixed by converting directory deletes to point deletes by resolving the files under the directory at deletion time and creating deletes for each file.
- Fixes a race condition that could occur in get file, which would result with partial results being returned. With this fix, I also refactored the common path for fileset wrapping and the full configuration into the source abstraction. This gives a stronger guarantee that the filesets wrappers are applied in the appropriate order and gives us one place for the full configuration (compute directory / file hashes and sizes).
- Short term fix for jobs hanging when a parent commit is squashed. This issue occurs because we resolve the commits outside of the job retry loop. The "fix" is to traverse the ancestor jobs in the job chain until we find one that doesn't correspond with a squashed commit. The longer term fix would be to make this traversal process operate on the commit chain since this problem still exists if all of the ancestor jobs in the job chain have squashed output commits.

This PR also adds fixed compaction delay configuration and refactors the in memory fileset into a buffer abstraction that is used by both the unordered writer and validator for tracking in-memory file state.


src/internal/storage/fileset/compaction.go
```diff
@@ -70,30 +38,22 @@ func indexOfCompacted(factor int64, inputs []*Primitive) int {
 	return l
 }
 
-// Compact compacts a set of filesets into an output fileset.
+// Compact writes the contents of ids to a new fileset with the specified ttl and returns the ID
+// Compact always returns the ID of a primitive fileset.
 func (s *Storage) Compact(ctx context.Context, ids []ID, ttl time.Duration, opts ...index.Option) (*ID, error) {
```
brendoncarroll, line 43: Where did the level based compaction go?

brendoncarroll: Okay I see it down below. So there is no way to perform level based compaction without the distributed compactor?

src/internal/storage/fileset/compaction.go
```diff
@@ -128,6 +88,26 @@ func NewDistributedCompactor(s *Storage, maxFanIn int, workerFunc CompactionBatc
 
 // Compact runs a compaction on the ids
 func (c *DistributedCompactor) Compact(ctx context.Context, ids []ID, ttl time.Duration) (*ID, error) {
```
brendoncarroll, line 90: We need a comment for this function, because it is not just a distributed version of Storage.Compact anymore.  It's the only thing that actually does the exponential levels.  Also note that it can return a Composite fileset, unlike Storage.Compact

src/internal/storage/fileset/option.go
```diff
@@ -55,6 +55,15 @@ func WithRenewal(ttl time.Duration, r *renew.StringSet) UnorderedWriterOption {
 	}
 }
 
+// WithParentID sets the parent fileset ID for the unordered writer.
+// This is used for converting directory deletions into a set of point
+// deletes for the files contained within the directory.
+func WithParentID(parentID *ID) UnorderedWriterOption {
```
brendoncarroll, line 61: Shouldn't this be an option on the Writer?

src/internal/storage/fileset/transforms.go
```diff
@@ -3,27 +3,46 @@ package fileset
...
+	}
+	return idxf
 }
 
-func (fil *indexFilter) Iterate(ctx context.Context, cb func(File) error, _ ...bool) error {
-	return fil.x.Iterate(ctx, func(fr File) error {
-		idx := fr.Index()
-		if fil.pred(idx) {
-			return cb(fr)
+func (idxf *indexFilter) Iterate(ctx context.Context, cb func(File) error, _ ...bool) error {
```
brendoncarroll, line 31: Does the last argument need to be propagated through?

src/internal/tarutil/util.go
```diff
@@ -277,3 +277,27 @@ func Export(storageRoot string, w io.Writer, opts ...ExportOption) (retErr error
 		})
 	})
 }
+
+// Reader converts a set of files to a tar stream.
+// TODO: Probably should just go to disk for this.
+func Reader(files []File) (io.Reader, error) {
```
brendoncarroll, line 283: nit: NewReader, or MakeReader for constructors.  Reserves Reader for an exported type later.

src/internal/stream/priority_queue.go
```diff
@@ -47,7 +47,7 @@ func (pq *PriorityQueue) Iterate(cb func([]Stream, ...string) error) error {
 func (pq *PriorityQueue) isHigherPriority(i, j int) bool {
 	si := pq.queue[i]
 	sj := pq.queue[j]
-	return si.Key() < sj.Key() || (si.Key() == sj.Key() && si.Priority() < sj.Priority())
+	return si.Key() < sj.Key() || (si.Key() == sj.Key() && si.Priority() > sj.Priority())
```
brendoncarroll, line 50: Why was this made?

src/server/pfs/server/driver.go
```diff
@@ -819,57 +819,49 @@ func (d *driver) finishCommit(txnCtx *txnenv.TransactionContext, commit *pfs.Com
...
-		empty := strings.Contains(commitInfo.Description, pfs.EmptyStr)
-		if err := d.updateProvenanceProgress(txnCtx, !empty, commitInfo); err != nil {
-			return err
-		}
-		if err := d.writeFinishedCommit(txnCtx.Stm, commit, commitInfo); err != nil {
-			return err
-		}
-		triggeredBranches, err := d.triggerCommit(txnCtx, commitInfo.Commit)
-		if err != nil {
+	}
```
brendoncarroll, line 829: So now we are compacting without the lock.  I don't know that it was necessarily the correct lock to be holding, but now it's gone, nonetheless.

We have to make sure no one is adding filesets to the commit when we are compacting. Or at least check that nothing has changed, and call SetTotalFileset in a transaction.

src/server/pfs/server/driver_compact.go
```diff
@@ -14,48 +14,63 @@ import (
...
+func (d *driver) compact(ctx context.Context, ids []fileset.ID) (*fileset.ID, error) {
+	compacted, err := d.storage.IsCompacted(ctx, ids)
+	if err != nil {
+		return nil, err
+	}
+	if compacted {
+		return d.storage.Compose(ctx, ids, defaultTTL)
+	}
+	var id *fileset.ID
+	if err := d.compactionQueue.RunTaskBlock(ctx, func(master *work.Master) error {
```
brendoncarroll, line 26: I do like this better here instead of in finish commit. But I think not having will create concurrency issues in finishCommit. See my other comment.

src/server/pfs/server/option.go
```diff
@@ -0,0 +1,22 @@
...
+import "github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset"
+
+type SourceOption func(*sourceConfig)
+
+type sourceConfig struct {
+	full   bool
+	filter func(fileset.FileSet) fileset.FileSet
+}
+
+func WithFull() SourceOption {
```
brendoncarroll, line 12: What do either of these do? Add comments to the constructors

src/server/pfs/server/source.go
```diff
@@ -27,11 +27,22 @@ type source struct {
 }
 
 // NewSource creates a Source which emits FileInfos with the information from commit, and the entries return from fileSet.
-func NewSource(commitInfo *pfs.CommitInfo, fileSet fileset.FileSet, full bool) Source {
+func NewSource(storage *fileset.Storage, commitInfo *pfs.CommitInfo, fs fileset.FileSet, opts ...SourceOption) Source {
```
brendoncarroll, line 30: This information could easily be communicated through a wrapped fileset.FileSet passed in the constructor.  None of the new state is used by any of sources methods.

I understand the goal of this is to consolidate the index resolver and dir inserter.  That can be done with a helper function that takes these additional arguments, without adding complexity to Source.

src/internal/storage/fileset/compaction.go
brycemcanally: Yeah.

src/internal/storage/fileset/compaction.go
brycemcanally: Yeah, it probably makes sense to pull out the level based part of this into a separate function / abstraction so that it is split such that the level based, distributed, and regular compaction are clearly divided into separate concepts.

src/internal/storage/fileset/option.go
brycemcanally: I think it makes more sense for the unordered writer. The writer has more restrictions on the ordering of paths, so it is going to be cleaner to have it operate on the fully resolved paths.

src/internal/storage/fileset/transforms.go
brycemcanally: Yeah, it does. I'll fix this for all of the transforms. Ideally we would be able to get rid of this method of specifying the deletive set some time in the near future.

src/internal/stream/priority_queue.go
brycemcanally: We had two implementations of this abstraction, one for the fileset merge and one for the datum merge (skipping). I merged them (pun intended) into one abstraction and had to choose a priority ordering. I went with the way the fileset merge determined priority (higher value is higher priority), so I changed the stream package implementation which was only used by the datum merge originally (I also had to change the datum merge logic to account for this).

src/server/pfs/server/driver.go
brycemcanally: Yeah, ideally we would just mark the commit as finished in a transaction (compaction would come later at read time) and that would block later (in the transaction serialization) attempts to modify filesets. I'm not sure what lock you are referring to though. I don't think this was ever locked in a way that would prevent that from occurring. We need to incrementally compact commits anyways to avoid the high fan in of reads on open commits, so I think this would be a reasonable change to include with that change since we would be moving the compaction to read time.

src/server/pfs/server/driver_compact.go
brycemcanally: This doesn't lock anything. It just sets up the task in the work package and blocks until it is complete. We need to solve the concurrency issue through Postgres as mentioned above.

src/server/pfs/server/option.go
brycemcanally: Sure, I'm surprised lint didn't complain about this.

src/server/pfs/server/source.go
brycemcanally: Discussed offline. We are sticking with the code as is to guarantee that the dir inserter is setup and to capture the full configuration in one place.

src/internal/storage/fileset/transforms.go
brendoncarroll: This is evidence of the optional argument introducing a bug in practice, so I'll suggest removing it again.

src/internal/stream/priority_queue.go
brendoncarroll: Okay so any comments which talk about merging left to right, where leftmost is the lowest index and rightmost is the last index still make sense?

src/internal/storage/fileset/compaction.go
brendoncarroll: Okay so something like:
```go
type Compactor interface {
  Compact(context.Context, ids []ID, ttl time.Duration) (*ID, error)
}

func DelayedCompact(ctx, c Compactor, ids []ID, ttl time.Duration) ([]ID, error) {
  // level based logic here, call c.Compact
}
```

src/server/pfs/server/driver.go
brendoncarroll: Okay I guess mark this as a TODO.  There are a couple of invariants we need to enforce.
1. You cannot write to a closed commit. (probably fixed with Postgres collecitons)
2. If you compact the diff fileset, you cannot write to the total until you check that the diff has not changed in a transaction.

src/server/pfs/server/option.go
brendoncarroll: Yeah I think linting broke, or this was disabled.  Comments on exports was one of the more important rules in my opinion.

src/server/pfs/server/testing/load/specs/load-4.yaml
```diff
@@ -0,0 +1,36 @@
...
+        prob: 1
+validator: {}
+fileSources:
+  - name: "random"
+    random:
+      fuzzSize:
+        - size:
+            min: 1000
+            max: 10000
+          prob: 0.3
```
brendoncarroll, line 22: A better way to do this is with weights. So you don't have to return an error if it doesn't sum to 1.

brendoncarroll (commented):
I don't see where the load tests are being run, other than in that global variable.
There should be single test that walks a directory and runs all the test yamls there as sub tests.

It looks like validation is an after thought here, and none of the specs actually define any validators.  The whole approach to validation as separate from input generation is flawed in my opinion.  Since we aren't really using it right now, it can be addressed later.

brendoncarroll (approved):

src/internal/storage/fileset/compaction.go
brycemcanally: This is pretty similar to what I went with, the main difference being that I used a compaction callback function rather than an interface. I don't really have a strong preference for interface vs callback, so if you prefer interface I can switch it over to that.

src/internal/storage/fileset/transforms.go
brycemcanally: Yeah, I don't think the deletive set should go through the same function as a normal iteration. I don't think we should do that in this PR though.

src/server/pfs/server/testing/load/specs/load-4.yaml
brycemcanally: I think probabilities are easier to understand for non-engineers (even engineers as well). I don't think it is a big deal that they need to sum to 1.

brycemcanally:
Yeah, that was the plan, but I am thinking about moving the load test runner into a gRPC endpoint such that we generate the workload in Pachd and can run the tests asynchronously (this should make testing workloads in hub, and in general, easier). Right now, we are just running the basic load test through the `TestLoad` test and the other tests are just sitting in a directory that I have been using for manual runs.

All of the specs are using the default validator configuration which stores the files that went in as hashes. I like this validation approach because I don't need to think about it for whatever workloads I add in the future, and it will work out of the box for whatever workloads other people specify. This isn't the end all be all for load testing because we cannot really do workload tailored validation with this model, but that isn't really the purpose. The goal is to be able to generate a large variety of workloads shapes and have basic validation that the content that went in looks as expected.

src/internal/stream/priority_queue.go
brycemcanally: Yeah, I'm going to include this in the file parts removal PR. I'm tweaking this again with the goal of just having a consistent ordering for the items going in and coming out of the merge. The inversion we do during the merge is confusing and error prone.

PR #5388, opened 2020-10-27
Use non-expiring tokens for PPS pipelines

msteffen:

PR #5389, opened 2020-10-27
Use non-expiring tokens for PPS pipelines

msteffen:
This is the master version of #5388


msteffen:
I accidentally merged this PR without the review, so I reverted it so it can be reviewed. I'm going to create a followup PR with the same changes, applied on top of the revert

PR #5399, opened 2020-10-29
Revert "Use non-expiring tokens for PPS pipelines"

msteffen:
Reverts pachyderm/pachyderm#5389


msteffen:
Accidentally merged #5389 without review. I'm reverting that PR so it can be reviewed

PR #5400, opened 2020-10-29
Use non-expiring tokens for PPS pipelines

msteffen:
This is the recreation of #5389 (which I reverted due to my accidentally merging it) and the master version of #5388


jdoliner (commented):
LGTM

PR #5402, opened 2020-10-30
Fix race condition when writing datum list job artifact in the worker master

Tryneus:
The `Close` of this file was happening after the subtask was written to the task channel, so a worker could pick up the subtask before object storage finished writing the data.  Fixed by moving the write into a separate call that happens synchronously.


jdoliner (approved):
LGTM

msteffen (approved):
LGTM

PR #5404, opened 2020-11-03
Refactor storage layer

brycemcanally:
This PR includes a pretty significant refactor of the storage layer.

The three main goals are:
- Simplify the code that manages tagged bytes.
- Change the storage layer abstractions to be stateless.
- Put the file set readers behind a common interface (FileSet and File).


src/server/pkg/storage/chunk/chain.go
```diff
@@ -0,0 +1,68 @@
...
+	close(prevChan)
+	return &TaskChain{
+		eg:       eg,
+		ctx:      errCtx,
+		prevChan: prevChan,
+	}
+}
+
+// CreateTask creates a new task in the task chain.
+func (c *TaskChain) CreateTask(cb func(context.Context, func(func() error) error) error) error {
```
brendoncarroll, line 32: This seems way more complicated than the problem at hand.  I think we should implement this with the more common "channel of channels" pattern.

brycemcanally: You want to just show me a simpler implementation. This seems pretty close to the simplest representation of the problem being solved (a chain of tasks that run in parallel, but finish serially). Keep in mind that this also handles context cancellation and errors within the tasks.

brendoncarroll: This is pattern is normally used to process a bunch of x's into y's in parallel and spit out the y's via callback.  I replaced the `interface{}`s with `func() error`.

It's a little longer, but the main advantage I see is the Do method is _much_ easier to understand than the CreateTask method.  I think that the complexity of that signature will prevent reuse. I don't think it will cross anyone's mind to express their operations as a triple nested function call.

```go
package workerpool

import (
	"golang.org/x/sync/errgroup"
)

type request struct {
	par, ser func() error
	respChan chan<- func() error
}

type WorkerPool struct {
	queue           chan request
	responses       chan (<-chan func() error)
	workers, cbExec *errgroup.Group
}

func newWorkerPool(n int) *WorkerPool {
	workers := &errgroup.Group{}
	cbExec := &errgroup.Group{}
	wp := &WorkerPool{
		queue:     make(chan request),
		responses: make(chan (<-chan func() error)),
		workers:   workers,
		cbExec:    cbExec,
	}
	// parallel requests
	for i := 0; i < n; i++ {
		workers.Go(func() error {
			for req := range wp.queue {
				if err := req.par(); err != nil {
					return err
				}
				req.respChan <- req.ser
			}
			return nil
		})
	}
	// serial callbacks
	cbExec.Go(func() error {
		for respChan := range wp.responses {
			ser := <-respChan
			if err := ser(); err != nil {
				return err
			}
		}
		return nil
	})
	return wp
}

func (wp *WorkerPool) Do(par, ser func() error) {
	respChan := make(chan func() error)
	wp.queue <- request{
		par:      par,
		ser:      ser,
		respChan: respChan,
	}
	wp.responses <- respChan
}

func (wp *WorkerPool) Close() (retErr error) {
	close(wp.queue)
	if err := wp.workers.Wait(); err != nil {
		retErr = err
	}
	close(wp.responses)
	if err := wp.cbExec.Wait(); err != nil && retErr == nil {
		retErr = err
	}
	return nil
}
```

brycemcanally: This would get messy on the client side because we would either need to declare all of the state that will get generated by the parallel portion such that it would be captured in a closure for both callbacks or setup a side channel for sending the data between the parallel and serial callbacks. The structure for the task chain is such that the closure is built into the pattern in that you have a wrapper function that blocks a part of a single callback for the serial part. Also, this does not handle context cancellation or errors, and has way more channel activity (also you have to be explicit about the number of workers and queue sizes). The model of running a set of callbacks in parallel, with a function wrapper that blocks a part of that callback based on the serial order of those callbacks is pretty simple and maps really well to our use case. This just doesn't seem like a good fit for the problem we are trying to solve.

src/server/pkg/storage/fileset/util.go
```diff
@@ -108,3 +122,84 @@ func DirUpperBound(p string) string {
 	}
 	return strings.TrimRight(p, "/") + "0"
 }
+
+// Iterator provides functionality for imperative iteration over a file set.
```
brendoncarroll, line 126: This needs to be better documented if it's exposed. How do I use this? When is iteration done? , if I don't close it, does it leak?  A code snippet with a for loop showing typical use is probably the shortest way to document it.

brendoncarroll (approved):

PR #5406, opened 2020-11-04
Go releaser

nitinjainsj:


goreleaser/docker-build-pipelines.yml
```diff
@@ -0,0 +1,50 @@
...
+builds:
+  -
+    id: pachctl
+    dir: src/server/cmd/pachctl
+    main: main.go
+    binary: pachctl
+    ldflags:
+      - -X {{ .Env.LD_FLAGS }} -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser
+    gcflags:
+      - all=-trimpath={{ .Env.PWD }}
```
jrockway, line 16: Apparently to do this right there also needs to be asmflags with a similar trimpath directive?  Maybe unnecessary since we set CGO_ENABLED=0.

goreleaser/docker-build-pipelines.yml
```diff
@@ -0,0 +1,50 @@
...
+changelog:
+  skip: true
+
+release:
+  disable: true
+
+dockers:
+  -
+    image_templates:
+      - pachyderm/go-build:{{.Env.VERSION}}
```
jrockway, line 39: Why {{.Env.VERSION}} throughout and not {{.Version}}?  I worry that there are two sources for which version is being built; goreleaser's own detection based on tags, and the environment.  It would be most reliable to consistently populate one from the other.  (i.e. in env sections, set VERSION={{.Version}}.)

goreleaser/pachctl.yml
```diff
@@ -0,0 +1,90 @@
...
+  name_template: "{{ .Env.VERSION }}"
+  prerelease: auto
+  disable: false
+
+brews:
+  -
+    name: pachctl@{{.Major}}.{{.Minor}}
+    ids:
+      - pachctl-archives
+    homepage: "github.com/pachyderm/pachyderm"
```
jrockway, line 77: This is not the same as the homepage listed in the debian package above.

goreleaser/pachctl.yml
```diff
@@ -0,0 +1,90 @@
...
+      owner: pachyderm
+      name: homebrew-tap
+    url_template: "https://github.com/pachyderm/pachyderm/releases/download/v{{ .Env.VERSION }}/{{ .ArtifactName }}"
+    commit_author:
+      name: pachyderm
+      email: nitin@pachyderm.com
+    test: |
+      system "#{bin}/pachctl version"
+    install: |
+      bin.install buildpath/"pachctl"
```
jrockway, line 89: Are the quotes necessary here?

jrockway (commented):
I think we should try to only have one config file, and vary the different builds with arguments passed to goreleaser.  If someone needs to update the pachctl build, they have to do it in 3 places, and it's highly likely that one will be forgotten and cause problems.

I don't have a good plan for the case where we put pachctl in a docker image.  We should probably drive the entire process inside docker, but use the same goreleaser configuration for consistency (but then it doesn't have the .git directory from which to guess the version number, so some work will be required there).  I also am not sure we should be passing GO_VERSION in as an ldflag -- if we want to print the go version as part of help (etc.) then we should get it from `runtime.Version()`.  Right now it looks easy to make a binary that says it was built with a version that it wasn't built with, which would be very confusing.

I still need to look into the docker image generation a little bit.  I've mostly reviewed the pachctl-related stuff so far.

jrockway (commented):
OK, I dove into the docker stuff and I understand how it works.  First, goreleaser builds binaries according to the `build` rules, and then it just copies those binaries into a docker container, with a configuration like:

```
FROM scratch
COPY mybin /
ENTRYPOINT ["/mybin"]
```

I am generally OK with this approach, but it's different than how we do it today, I think.  How we make this work depends on what runs the builds; it will look different if you expect someone to run this on their workstation versus it being done in CI somehow.  (I don't know how to do it cleanly in CI.  My mental model, which I don't think goreleaser supports is this: 1) Click a button to trigger a release 2) Something comes up with the correct version number.  3) Documentation/config is edited to refer to the new version (maybe irrelevant for us, but at the very least the Changelog typically gets the version at the top incremented) 4) This is committed, and the tag that was generated in step 2 is applied.  5) The build happens.  6) The commit from step 4 is pushed. 7) The containers are pushed and the Github release is created. 8) A message is posted to slack saying that we did the release :))

Makefile
```diff
@@ -40,73 +45,47 @@ install-clean:
...
-	docker tag pachyderm/worker pachyderm/worker:local
-
-docker-build-pachctl:
-	DOCKER_BUILDKIT=1 docker build \
-		--build-arg GO_VERSION=`cat etc/compile/GO_VERSION` \
-		--build-arg LD_FLAGS="$(LD_FLAGS)" \
-		--build-arg GC_FLAGS="$(GC_FLAGS)" \
-		$(DOCKER_BUILD_FLAGS) \
-		--progress plain -f Dockerfile.pachctl -t pachyderm/pachctl .
+	DOCKER_BUILDKIT=1 goreleaser release -p 1 $(GORELSNAP) --skip-publish --rm-dist -f goreleaser/docker.yml
```
msteffen, line 85: **Note for future**: This is probably outside the scope of this PR, but do you think it would make sense, in a subsequent PR, to set DOCKER_BUILDKIT inside of goreleaser if possible (so that `goreleaser release` doesn't necessarily need to be run from the Makefile)?

Makefile
```diff
@@ -7,16 +7,21 @@
...
 
-MINIKUBE_MEM=8192 # MB of memory allocated to minikube
-MINIKUBE_CPU=4 # Number of CPUs allocated to minikube
+MINIKUBE_MEM = 8192 # MB of memory allocated to minikube
+MINIKUBE_CPU = 4 # Number of CPUs allocated to minikube
+
+CHLOGFILE = /tmp/pachyderm/release/changelog.diff
+export GOVERSION = $(shell cat etc/compile/GO_VERSION)
+GORELSNAP = #--snapshot # uncomment --snapshot if you want to do a dry run.
+SKIP = #\# # To skip push to docker and github remove # in front of \#
```
msteffen, line 24: **Note for future**:This also would probably go into its own PR anyway, but as a matter of ergonomics, do you think it would be controlling both of these with a single var? e.g.
```
ifdef DRY_RUN
  GORELEASER_FLAGS += --snapshot
  DOCKER_SKIP = \#
endif
```

Makefile
```diff
@@ -40,73 +45,47 @@ install-clean:
...
 release-candidate:
-	@if [ $${VERSION_ADDITIONAL:0:1} != - ]; then \
-	  echo "VERSION_ADDITIONAL must start with a '-'"; \
-	  exit 1; \
-	fi
+	@make custom-release
+
+custom-release:
+	echo "" > $(CHLOGFILE)
+	@VERSION_ADDITIONAL=$(VERSION_ADDITIONAL) ./etc/build/make_release.sh
```
msteffen, line 66: I thought about this a little more, and is it both `VERSION` and `VERSION_ADDITIONAL` that isn't getting passed through? If it's just VERSION, maybe all that's needed is to put `export VERSION` at the top?

etc/build/make_changelog.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+if [ $# -ne 1 ]
+then
+    echo "Need to specify change log file used in goreleaser"
+    exit 1
+fi
+
+mkdir -p "$(dirname "$1")" && touch "$1"
+git diff HEAD^ HEAD -- CHANGELOG.md | sed -n /##/,/##/p | grep -v "##" | cut -d '+' -f 2 > "$1"
```
msteffen, line 12: **Note for future**: I just now noticed, after reading this, that all of the 1.V.x release notes get written into the 1.V branch (`CHANGELOG.md` in `master` stops at 1.11.0, because we write those updates into the 1.11.x branch). I just wanted to note here that whatever process we end up using for `make doc` to get `1.11.x` docs changes served from master, we might need to apply to `CHANGELOG.md` as well.

etc/testing/travis_build.sh
```diff
@@ -5,13 +5,16 @@ set -ex
 # We can't run the build step if there's no access to the secret env vars
 if [[ "$TRAVIS_SECURE_ENV_VARS" == "true" ]]; then
     docker login -u pachydermbuildbot -p "${DOCKER_PWD}"
-    make install docker-build
+    make install
     version=$(pachctl version --client-only)
+    git tag -f -am "Travis test v${version}" v"${version}"
```
msteffen, line 10: Just trying to understand this piece, what does `git tag` do here? Does `make docker-build` need it?

etc/testing/lint.sh
```diff
@@ -18,7 +18,7 @@ done
 
 go get -u golang.org/x/lint/golint
 find "./src" \
-  \( -path "*.pb.go" -o "${skip_paths[@]}" \) -prune -o -name '*.go' -print \
+  \( -path "*.pb.go" "${skip_paths[@]}" \) -prune -o -name '*.go' -print \
```
msteffen, line 21: I made a slightly longer fix for this in #5434: https://github.com/pachyderm/pachyderm/pull/5434/files#diff-c645c5f5616b86e3827ef72cb6fe3815d38df279da2bcda31ec731155926debd

Hopefully git is smart enough to just merge them, since I think those changes are line-wise a superset of this change. But if you do get a merge conflict, that's what's going on.

msteffen (approved):
Just a few high-level questions, superficial suggestions, and ideas for future extensions. Overall LGTM. It will be great to get away from goxc!

Makefile
```diff
@@ -7,16 +7,21 @@
 RUN= # used by go tests to decide which tests to run (i.e. passed to -run)
 # Label it w the go version we bundle in:
 export VERSION_ADDITIONAL = -$(shell git log --pretty=format:%H | head -n 1)
-LD_FLAGS = -X github.com/pachyderm/pachyderm/src/client/version.AdditionalVersion=$(VERSION_ADDITIONAL)
+export LD_FLAGS = github.com/pachyderm/pachyderm/src/client/version.AdditionalVersion=$(VERSION_ADDITIONAL)
```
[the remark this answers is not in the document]
msteffen: Sorry, I wrote this comment before our pragli call (for posterity: the goreleaser validator requires `LD_FLAGS` arguments to start with `-X`, so the `-X` had to be moved out of the make var and into the goreleaser config, in front of the make-provided `LD_FLAGS` arguments). Marking resolved.

etc/pipeline-build/go/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM golang:latest
 
 WORKDIR /app
-ADD build.sh .
-ADD run.sh .
+ADD etc/pipeline-build/go/build.sh .
+ADD etc/pipeline-build/go/run.sh .
```
lukemarsden, line 5: wouldn't a better change be to make whatever does `docker build` of this first `cd` into `etc/pipeline-build/go`? So that anyone can build this docker file from that directory.

etc/pipeline-build/python/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM python:3
 
 WORKDIR /app
-ADD build.sh .
-ADD run.sh .
+ADD etc/pipeline-build/python/build.sh .
+ADD etc/pipeline-build/python/run.sh .
```
lukemarsden, line 5: same comment as above: wouldn't a better change be to make whatever does `docker build` of this first `cd` into `etc/pipeline-build/go`? So that anyone can build this docker file from that directory.

etc/testing/circle_tests_inner.sh
```diff
@@ -9,7 +9,7 @@ VM_IP="localhost"
 export VM_IP
 PACH_PORT="30650"
 export PACH_PORT
-GOPATH="${HOME}/go"
+GOPATH=/root/go
```
lukemarsden, line 12: why is this change necessary?

etc/testing/vault-s3-client/test.sh
```diff
@@ -5,7 +5,11 @@ SCRIPT_DIR="$(dirname "${0}")"
 set -ex
 
 # Build pachyderm and push the relevant images
-make install && make docker-build && make docker-push || exit 1
+make install || exit 1
+VERSION=$(pachctl version --client-only)
+git tag -f -am "Travis test v${VERSION}" v"${VERSION}" || exit 1
+make docker-build || exit 1
+make docker-push || exit 1
```
lukemarsden, line 12: aren't all these `|| exit 1`s not doing anything, given that `-e` is set?

IOW, doing "or fail" on something that will fail if it fails anyway is the same as doing nothing?

etc/worker/Dockerfile
```diff
@@ -1,7 +1,7 @@
 ARG GO_VERSION
 FROM golang:${GO_VERSION} AS worker_build
 WORKDIR /app
-COPY init.go .
+COPY etc/worker/init.go .
```
lukemarsden, line 4: same comment as above: Let's make the thing that builds this image cd into the right working directory for the build context to relative be `./`, if possible?

lukemarsden (approved):
Nice job @nitinjainsj! Left some purely nice-to-have review comments. Consider addressing them optional, and merge at will!

.testfaster.yml
```diff
@@ -47,7 +47,10 @@ base:
   os_dockerfile: |-
     FROM quay.io/testfaster/kube-ubuntu
     ENV cache-bust 19
-    RUN apt-get update && apt-get install -y build-essential gettext-base socat shellcheck golang-go rsync jq pv silversearcher-ag
+    RUN apt-get update && apt-get install -y build-essential gettext-base socat shellcheck rsync jq pv silversearcher-ag
+    RUN cd /root && wget -nv -O - https://dl.google.com/go/go1.15.4.linux-amd64.tar.gz | tar -zxf -
```
lukemarsden, line 51: this will fix some errors i've seen in circleci regarding some newer branches requiring 1.15, nice work!

PR #5410, opened 2020-11-04
goreleaser support in 1.11.x

nitinjainsj:
Add goreleaser for doing point/custom/release-candidate builds. Also support build and push of docker images using goreleaser


nitinjainsj:
We don't need it for 1.11.x.

PR #5416, opened 2020-11-05
CircleCI and Testfaster

lukemarsden:
These are the code changes to start using CircleCI and Testfaster for core pach master.

Buttons will need to be clicked in CircleCI to enable this repo and set the right secrets (TESTFASTER_TOKEN and ENT_ACT_CODE at least come to mind), but this is the code part. Update: I've clicked the buttons and CircleCI should start running as soon as this PR is merged.

Ready for review.


lukemarsden:
Ok, circle is running on the latest commit now. Let's iterate on this branch until we get a green run in CircleCI.

msteffen (commented):
I'll do a fuller review soon, but I have two high-level questions, which I think would help me understand the goals of this PR better:

1. Why use `RUN_BAD_TESTS` instead of `TRAVIS`? Is it a goal to disable flaky tests in Travis?
2. How did you identify which tests to disable? E.g. `TestSyncFile` and `TestGlobFile` I haven't seen to be flaky.

I think if the goal of this PR is to get CircleCI+TestFaster running regularly so we can start iterating on them, then that sounds like a big step forward to me. My general level of anxiety around bugs in Pachyderm that we're ignoring due to tests we never wrote, disabled tests, flaky tests that we ignore, etc, is high, so I have reservations about merging this if it's potentially taking us further in that direction, but I don't think/can't tell if that's on the table or not.

Makefile
```diff
@@ -384,7 +391,7 @@ launch-logging: check-kubectl check-kubectl-connection
 	kubectl --namespace=monitoring port-forward `kubectl --namespace=monitoring get pods -l k8s-app=kibana-logging -o json | jq '.items[0].metadata.name' -r` 35601:5601 &
 
 launch-loki:
-	helm repo add --force-update loki https://grafana.github.io/loki/charts
+	helm repo add loki https://grafana.github.io/loki/charts
```
msteffen, line 394: Do newer versions of help have a replacement for this flag? The reason we added it is that retries in Travis break if it's not there. `travis_retry` runs the tests for a second time, which includes running this command, and then it fails twice with `repo "loki" already exists`. Maybe adding something like `helm repo delete loki || true` beforehand would work?

Makefile
```diff
@@ -5,10 +5,17 @@
...
-# Label it w the go version we bundle in:
-export VERSION_ADDITIONAL = -$(shell git log --pretty=format:%H | head -n 1)
-LD_FLAGS = -X github.com/pachyderm/pachyderm/src/client/version.AdditionalVersion=$(VERSION_ADDITIONAL)
-export GC_FLAGS = "all=-trimpath=${PWD}"
+
+# Don't set the version to the git hash in CI, as it breaks the go build cache.
+ifdef CIRCLE_BRANCH
+	export VERSION_ADDITIONAL = "ci_build"
+	LD_FLAGS = -X github.com/pachyderm/pachyderm/src/client/version.AdditionalVersion=$(VERSION_ADDITIONAL)
+	export GC_FLAGS = ""
```
msteffen, line 13: Not a problem, really, but just curious: why not set `trimpath` in circle? It seems like it would make the logs shorter?

etc/kube/push-to-minikube.sh
```diff
@@ -6,6 +6,11 @@ if [[ $# -ne 1 ]]; then
   echo "error: need the name of the docker image to push"
 fi
 
+if [ -f /TESTFASTER_PREWARM_COMPLETE ]; then
```
msteffen, line 9: Does testfaster use Minikube? Again, not a problem, but I guess I'd expect the `if` statement below to do the same job. Happy to leave this in if it doesn't.

etc/testing/testctl-rsync.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -euo pipefail
+IP=$(grep "##EXTERNAL_IP=" "$KUBECONFIG" |cut -d '=' -f 2-)
+PORT=$(grep "##SSH_FORWARDED_PORT=" "$KUBECONFIG" |cut -d '=' -f 2-)
+grep "##PRIVATE_KEY_ONELINER=" "$KUBECONFIG" |cut -d '=' -f 2- |base64 -d > id_rsa
+chmod 0600 id_rsa
+local_path="$1"
+remote_path="$2"
+exec rsync -a -e "ssh -i $(pwd)/id_rsa -p $PORT -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" "$local_path" root@"$IP":"$remote_path"
```
msteffen, line 9: Is this...sshing over rsync? I didn't know that was possible...

src/server/pachyderm_test.go
```diff
@@ -7142,7 +7158,9 @@ func TestSkippedDatums(t *testing.T) {
 }
 
 func TestOpencvDemo(t *testing.T) {
-	t.Skip("flaky")
```
msteffen, line 7145: I think we may actually need to leave this in, so the test doesn't start running in Travis

src/server/pachyderm_test.go
```diff
@@ -4910,7 +4921,9 @@ func TestDatumStatusRestart(t *testing.T) {
 }
 
 func TestUseMultipleWorkers(t *testing.T) {
-	t.Skip("flaky")
```
msteffen, line 4913: Same comment as below, I think we may need to leave this in for now

msteffen (approved):
Approved, with just a few questions. Also, apologies for taking until this morning to review it—it honestly just got late while I was working on #5434 and I forgot about this and fell asleep. If you want I can try to merge it during the day today?

Makefile
```diff
@@ -5,10 +5,17 @@
...
 
 RUN= # used by go tests to decide which tests to run (i.e. passed to -run)
-# Label it w the go version we bundle in:
-export VERSION_ADDITIONAL = -$(shell git log --pretty=format:%H | head -n 1)
-LD_FLAGS = -X github.com/pachyderm/pachyderm/src/client/version.AdditionalVersion=$(VERSION_ADDITIONAL)
-export GC_FLAGS = "all=-trimpath=${PWD}"
+
+# Don't set the version to the git hash in CI, as it breaks the go build cache.
+ifdef CIRCLE_BRANCH
+	export VERSION_ADDITIONAL = "ci_build"
```
msteffen, line 11: Instead of setting `CIRCLE_BRANCH=1` in testfaster, and having that change the `VERSION_ADDITIONAL` here, I think you could just set `VERSION_ADDITIONAL=ci_build` in testfaster directly and it would have the same effect (and I think wouldn't need any special handling)

etc/testing/testctl-rsync.sh
lukemarsden: it's rsync over ssh :)

etc/kube/push-to-minikube.sh
lukemarsden: it does, and i'm not sure why it wasn't working. i can try and figure that out and remove this as a followup.

Makefile
lukemarsden: because the path differs between the prewarm build and the real build, which i feared would break the go build cache

src/server/pachyderm_test.go
lukemarsden: oh, I thought I added this skip as part of this branch! That's odd.

src/server/pachyderm_test.go
lukemarsden: same as above. Will check in master and put it back if it's definitely there too.

Makefile
lukemarsden: They don't. I'll do as you suggest. Thanks!

chainlink: You can also pre-load this file, have a look at `cat ~/.helm/repository/repositories.yaml`

chainlink:
Congrats on getting this across the line! Looks super clean overall 🚀 

One potential iteration we might want to make:
Using CircleCI workflows to build pachctl/pachd/docker first, then pull them into each matrix run (vs compiling in each matrix run)

lukemarsden:
Thanks @chainlink!

The build happens before the CircleCI run starts ;) using the older checkout of the code defined in testfaster.yml. Then the build with the latest version of the code then only takes 12-40 seconds (x2). I'm pretty sure that's quicker than doing the build from scratch in CircleCI and stashing the artifacts somewhere.

Makefile
lukemarsden: done this

Makefile
lukemarsden: I went with Matthew's suggestion for expediency.

Makefile
lukemarsden: Reverted this as it broke things

lukemarsden:
Ready to merge as soon as Travis passes!

chainlink:
> The build happens before the CircleCI run starts ;) using the older checkout of the code defined in testfaster.yml. Then the build with the latest version of the code then only takes 12-40 seconds (x2), which is quick because it reuses the old build's go and docker caches. I'm pretty sure that's quicker than doing the build from scratch in CircleCI and stashing the artifacts somewhere.

Oh man, that's really nice. You typically lose the go build cache with docker (unless you use buildkit, which isn't well supported) so this is a really nice workaround!

lukemarsden:
It warms up the buildkit cache this way :)

PR #5420, opened 2020-11-06
Update README.md

pappasilenus:
Added the step where you actually create the darned secret.


echohack (approved):
Minor, but `sh` is not a grammar. We should use  ` ```shell` instead. Since this is consistent, let's go ahead and merge this. I'll follow up with a more comprehensive fix to the grammars.

Reference:
https://github.com/github/linguist/blob/master/vendor/README.md
https://github.com/atom/language-shellscript

PR #5422, opened 2020-11-06
Add transaction versions of pipeline helpers.

PFedak:
I anticipate a PR with a handful of changes like this, wanted to make sure there's no obviously cleaner approach before I get too much further


Tryneus (approved):
LGTM

PR #5425, opened 2020-11-10
[Docs] [Getting-started] fixed some minor issues on getting-started

echohack:

I think this document needs quite a bit of work, but it's a little too early (for me) to do a more thorough overhaul. For now, I've stuck with a few changes while keeping to the current format.

Improvements:

1. macOS does not need to specify the full path to an application to launch it with `open`. We can avoid a potentially confusing situation by modifying this command. This also avoids supporting N number of specific operating systems on a go-forward basis (macOS Big Sur is likely to have some large changes as well...). Overall: cleaner, more simple, and more likely to work and avoid bad copypasta.

2. Removed any references to PFS. Get this out of the way and don't bog down the user with too many details before getting to the next action. Stay focused on the pipeline specification instead.

3. JSON does not support comments, so remove the comments from JSON to avoid copy-pasta errors in other people's editors.

4. ` ```bash ` is not typically a supported grammar in markdown code fences. If it is, it is an alias for ` ```shell `, which references  https://github.com/atom/language-shellscript -- See also: https://github.com/github/linguist/blob/master/vendor/README.md

5. Some minor spelling errors and grammar cleanup.

### DIFF
--- a/doc/docs/master/getting_started/beginner_tutorial.md
+++ b/doc/docs/master/getting_started/beginner_tutorial.md
@@ -133,54 +133,48 @@ We can check to make sure the data we just added is in Pachyderm.
   **System response:**
 
   ```
-  COMMIT                           NAME         TYPE COMMITTED          SIZE 
+  COMMIT                           NAME         TYPE COMMITTED          SIZE
   d89758a7496a4c56920b0eaa7d7d3255 /liberty.png file About a minute ago 57.27KiB
   ```
 
-Also, you can view the file you have just added to Pachyderm. Because this is an
+Now you can view the file by retrieving it from Pachyderm. Because this is an
 image, you cannot just print it out in the terminal, but the following
-commands will let you view it easily:
+command will let you view it:
 
+* On macOS, run:
 
-* On macOS prior to Catalina, run:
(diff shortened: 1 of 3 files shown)


doc/docs/master/getting_started/beginner_tutorial.md
```diff
@@ -199,22 +193,24 @@ through the details.
...
+process them. In this case, `images` is the repo, and `/*` is the glob pattern.
+
+The glob pattern defines how the input data can be broken up if you want
+to distribute computation. `/*` means that each file can be
 processed individually, which makes sense for images. Glob patterns are
-one of the most powerful features in  Pachyderm.
+one of the most powerful features in Pachyderm.
 
-The following text is the Python code that we run in this pipeline:
+The following text is the Python code run in this pipeline:
```
nadegepepin, line 211: The following text is the Python code `edges.py` run in this pipeline:

nadegepepin (changes requested):
Looks awesome to me! Just added a small comment - Please keep the reference on the python file `edges.py`.

nadegepepin (approved):
LGTM

PR #5428, opened 2020-11-10
[Docs] Fix code fence grammar

echohack:

Code fence grammars are supported through Github's [linguist project](https://github.com/github/linguist). Across our documentation, we are using a number of different aliases for `shell` for these grammars. While this is not "incorrect" per se, we should only use the root name of the grammar, instead of an alias. Using the name of the grammar instead of a number of different aliases makes it easier to programmatically identify these code blocks. Ultimately consistency allows us to write better documentation across the entire project.

This PR is laser focused on fixing these grammars. It's common for folks to copy-paste or reference existing documentation when writing new documentation, so this also helps keep a single usage across the documentation on a go-forward basis. A linter to check for aliased grammars in the future might be a good idea.

I have inspected these changes with a visual diffing tool to ensure it did not step on anything else by accident. Finally, the only files effected are `*.md` markdown files, indicating that no grammars are used in actual production code, and that this change effects documentation only.

Reference:

The aliases that Linguist uses: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L5172
Linguist Full Grammar List: https://github.com/github/linguist/blob/master/vendor/README.md


nadegepepin (approved):
Nice! Go ahead.

PR #5430, opened 2020-11-11
Adds encryption and compression to the chunk refs.

brendoncarroll:
This makes the chunk tests about 4x slower. From 30seconds to 120seconds.  We probably need to add the in memory chunk cache in this P.R. before merging this in.

It is based off of sv2-chunk-tracker #5339 and is blocked on merging that.


src/server/pkg/pachhash/hash.go
```diff
@@ -0,0 +1,34 @@
...
+func New() hash.Hash {
+	h, err := blake2b.New256(nil)
+	if err != nil {
+		panic(err)
+	}
+	return h
+}
+
+// Sum computes a hash sum for a set of bytes.
+func Sum(data []byte) Output {
```
brycemcanally, line 27: It seems like it would be better to return a byte slice here, and just handle grabbing the 32 bytes under the hood. Is there a reason why we shouldn't do that?

src/server/pkg/pachhash/hash.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Output is the output of the hash function.
+// Sum returns an Output
+type Output = [OutputSize]byte
+
+// New creates a new hasher.
+func New() hash.Hash {
+	h, err := blake2b.New256(nil)
+	if err != nil {
+		panic(err)
```
brycemcanally, line 21: This is getting out of hand lol. Maybe we should just change the signature to return an error?

src/server/pkg/storage/chunk/storage.go
```diff
@@ -17,20 +17,24 @@ const (
 
 // Storage is the abstraction that manages chunk storage.
 type Storage struct {
+	memObj    obj.Client
```
brycemcanally, line 20: Why does chunk storage need two object storage clients? Is it ever going to operate on the one without the memory caching wrapper?

src/server/pkg/storage/chunk/transform.go
```diff
@@ -0,0 +1,201 @@
...
+		err := func() error {
+			gw, err := gzip.NewWriterLevel(lw, gzip.BestSpeed)
+			if err != nil {
+				return err
+			}
+			_, err = gw.Write(src)
+			if err != nil {
+				return err
+			}
+			return gw.Close()
```
brycemcanally, line 96: We probably should defer this. I don't think it really matters in this case, but I think we should always follow that pattern unless it would result with incorrect behavior (like closing the object storage writers).

src/server/pkg/storage/chunk/transform.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	return hex.EncodeToString(Hash(data)[:16])
+}
+
+func getFromCache(ctx context.Context, objs obj.Client, ref *Ref, w io.Writer) error {
+	rc, err := objs.Reader(ctx, ref.Key(), 0, 0)
+	if err != nil {
+		return err
+	}
+	defer rc.Close()
```
brycemcanally, line 194: error

src/server/pkg/storage/chunk/transform.go
```diff
@@ -0,0 +1,201 @@
...
+)
+
+type CreateOptions struct {
+	Secret      []byte
+	Compression CompressionAlgo
+}
+
+func Create(ctx context.Context, opts CreateOptions, data []byte, createFunc func(ctx context.Context, data []byte) (ID, error)) (*Ref, error) {
+	buf := make([]byte, len(data))
+	// compression
```
brycemcanally, line 23: I think most of the comments in this file are just filling up space without adding value. The code communicates what is happening clearly enough without them. I think comments are only warranted when it is not obvious what the code is doing. Plus, I think it is better to read the code rather than the comments to understand how something works because it is the source of truth.

src/server/pkg/storage/chunk/chunk.proto
```diff
@@ -15,10 +15,16 @@ message DataRef {
...
+  GZIP_BEST_SPEED = 1;  
+}
+
 message Ref {
   bytes id = 1;
   int64 size_bytes = 2;
   bool edge = 3;
 
-  // TODO: compression, encryption
+  bytes dek = 4;
```
brycemcanally, line 28: I think this should be spelled out. I knew this was a key associated with encryption, but I did not understand what the `de` part was. I'm guessing it is supposed to be an abbreviation for derived key, which I think we should just spell out here.

src/server/pkg/pachhash/hash.go
brendoncarroll: An error from this constructor is a configuration error. We want to know as soon as possible, and there is no point in checking for it everywhere because there is no way to handle it, other than by using the constructor correctly.

src/server/pkg/storage/chunk/storage.go
brendoncarroll: They don't form a single storage hierarchy.  The memory store holds uncompressed cipher text.  The disk cache and object storage form a storage hierarchy for ciphertexts.

src/server/pkg/storage/chunk/chunk.proto
brendoncarroll: It stands for Data Encryption Key.  The acronym is pretty widely used.
https://cloud.google.com/kms/docs/envelope-encryption

brendoncarroll: The key is derived in our case, but it could also be randomly generated.

src/server/pkg/pachhash/hash.go
brendoncarroll: https://pkg.go.dev/golang.org/x/crypto/blake2b#New256
If you give a key that is too large it errors.

src/internal/obj/mem_client.go
```diff
@@ -0,0 +1,90 @@
...
+	"bytes"
+	"context"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+	"sync"
+)
+
+type memClient struct {
```
brycemcanally, line 13: We discussed this offline, but I just want to make a note about going ahead and moving to the key / value store model for the purposes of the in-memory cache. We just need a simple in-memory, LRU cache that stores the plaintext chunks. This would just be used by the chunk storage layer.

brycemcanally: Also, we wouldn't need the void client.

src/server/pkg/pachhash/hash.go
brendoncarroll: Yeah there are a few reasons.
1. This can prevent allocations, if output is just copied to its destination.
2. It can be used in map and sync.Map. `[]byte` cannot.  We are making use of this for the in memory cache.
3. It is a correctly sized hash, no chance of getting a short buffer if you expect this type.
4. Returning `[]byte` is a lowest common interface thing for `hash.Hash` you'll notice most of the cryptographic hash implementations include a `Sum256` or `Sum512`.

src/server/pkg/storage/chunk/transform.go
brendoncarroll: The object client stuff is now taken care of in `storage/kv/obj_adapter.go`

src/server/pkg/storage/chunk/transform.go
brendoncarroll: I removed some of the comments that are just describing what a section of code is doing. But I want to keep the function descriptions.

src/internal/obj/mem_client.go
brendoncarroll: These changes are done.

src/internal/storage/kv/kv.go
```diff
@@ -0,0 +1,40 @@
...
+// ErrKeyNotFound is returned when the store does not contain a key
+var ErrKeyNotFound = errors.New("key not found")
+
+// ValueCallback is the type of functions used to access values
+type ValueCallback = func([]byte) error
+
+// GetPut supports the basic GetF and Put operations
+type GetPut interface {
+	Put(ctx context.Context, key, value []byte) error
+	GetF(ctx context.Context, key []byte, cb ValueCallback) error
```
brycemcanally, line 18: Could this just be `Get`? We have been moving away from the `F` suffix for callback based functions in other parts of the codebase. Also, really minor nit, but I think this interface should be `PutGet` rather than `GetPut`.

src/server/pfs/server/driver.go
```diff
@@ -118,7 +119,8 @@ func newDriver(env *serviceenv.ServiceEnv, txnEnv *txnenv.TransactionEnv, etcdPr
 	if err != nil {
 		return nil, err
 	}
-	chunkStorage := chunk.NewStorage(objClient, chunk.NewPostgresStore(db), tracker, chunkStorageOpts...)
+	memCache := kv.NewMemCache(10)
```
brycemcanally, line 122: We should probably go ahead and setup the configuration for this.

src/internal/storage/kv/memcache.go
```diff
@@ -0,0 +1,66 @@
...
+	mc.mu.Lock()
+	defer mc.mu.Unlock()
+	v, ok := mc.cache.Get(string(key))
+	if !ok {
+		return nil
+	}
+	return v.([]byte)
+}
+
+func (mc *memoryCache) onEvict(key, value interface{}) {
```
brycemcanally, line 65: I think we should just pass in nil for the `onEvict` function in `NewLRU` since we are not using it. Then get rid of this.

src/internal/storage/kv/kv.go
brendoncarroll: Yeah it can be `Get` instead

I don't know where you're getting `PutGet`
In the `io` package, there is `io.ReadWriter` so any analogy of Read -> Get and Put -> Write has that convention.
There is also just concatenating them in alphabetically order, which supports `GetPut`.

src/internal/storage/chunk/storage.go
```diff
@@ -18,50 +19,59 @@ const (
...
+		tracker:   tracker,
+		createOpts: CreateOptions{
+			Compression: CompressionAlgo_GZIP_BEST_SPEED,
+		},
 	}
 	for _, opt := range opts {
 		opt(s)
 	}
+	s.store = kv.NewFromObjectClient(s.objClient)
+	s.objClient = nil
```
brycemcanally, line 45: Why do we still have this field if we are just going to nil it out?

src/internal/storage/kv/kv.go
```diff
@@ -0,0 +1,40 @@
...
+	GetPut
+	Delete(ctx context.Context, key []byte) error
+	Exists(ctx context.Context, key []byte) (bool, error)
+	Walk(ctx context.Context, prefix []byte, cb func(key []byte) error) error
+}
+
+var crc64Tab = crc64.MakeTable(crc64.ISO)
+
+// AssertNotModified calls cb with x, but panics if x changed during the execution of cb
+func AssertNotModified(x []byte, cb ValueCallback) error {
```
brycemcanally, line 32: Is this worth the cost? We have to do a checksum for every chunk just to check if we have code that is modifying the in-memory chunk?

src/internal/storage/kv/kv.go
brycemcanally: Yeah, that is a fair point. I was mostly pulling that from the way I think about it (you write a chunk before you can get it) and because you laid out the functions with `Put` before `Get`. I agree we should follow the Go conventions, so if you just make the `GetF` -> `Get` change and list `Get` before `Put` then we should be good to go.

src/internal/storage/chunk/storage.go
brendoncarroll: So that the options can configure it.  I set it to nil so anything still referencing it would break.

src/internal/storage/chunk/transform.go
```diff
@@ -0,0 +1,215 @@
...
+func cryptoXOR(key, dst, src []byte) {
+	nonce := [chacha20.NonceSize]byte{}
+	ciph, err := chacha20.NewUnauthenticatedCipher(key, nonce[:])
+	if err != nil {
+		panic(err) // this only happens if you pass in an invalid key/nonce size, which we shouldn't do
+	}
+	ciph.XORKeyStream(dst, src)
+}
+
+func verifyData(id ID, x []byte) error {
```
brycemcanally, line 187: Not really sure if this is worth the cost as well.

brycemcanally: TODO is fine here as well.

src/internal/storage/chunk/storage.go
brycemcanally: Okay, not too sure how I feel about this, but I guess it is fine for now.

src/internal/storage/chunk/transform.go
brendoncarroll: This 100% has to stay.  None of the magical security/integrity properties of content-addressed storage apply without this.

src/server/pfs/server/driver.go
brendoncarroll: `StorageMemoryCacheSize` denominated in chunks, same as `StorageDiskCacheSize`

brycemcanally (approved):
lgtm

PR #5431, opened 2020-11-11
Create pipeline info transactionally

PFedak:
This is very rough (obviously doesn't even build), but hopefully makes clear what I'm planning


PFedak:
Oh whoops, assume that I'm threading through and returning specCommit (the new commit with the pipeline info file) through the function whenever it is created, even if there's an error later. This feels pretty awkward.

src/server/pps/server/api_server.go
```diff
@@ -2391,150 +2467,156 @@ func (a *apiServer) CreatePipeline(ctx context.Context, request *pps.CreatePipel
...
+			if !request.Reprocess {
+				pipelineInfo.Salt = oldPipelineInfo.Salt
+			}
+			specCommit, err := createOrValidateSpecCommit()
+			if err != nil {
+				return err
+			}
+
+			// move the spec branch for this pipeline to the new commit
+			txnCtx.Pfs().CreateBranchInTransaction(txnCtx, &pfs.CreateBranchRequest{
```
PFedak, line 2504: This consistently errors from not being able to find the commit we just made, but ends up working (in the STM retry, maybe not the first time). I assume this means I'm accidentally creating an entry in the transaction's STM for it which prevents it from going to PFS for real, but I'm not sure.

src/server/pps/server/api_server.go
```diff
@@ -2362,27 +2369,96 @@ func (a *apiServer) CreatePipeline(ctx context.Context, request *pps.CreatePipel
...
+			return errors.Errorf("the HEAD commit of this pipeline's spec branch " +
 				"is open. Either another CreatePipeline call is running or a previous " +
 				"call crashed. If you're sure no other CreatePipeline commands are " +
 				"running, you can run 'pachctl update pipeline --clean' which will " +
 				"delete this open commit")
 		}
 
-		// Remove provenance from existing output branch, so that creating a new
-		// spec commit doesn't create an output commit in the old output branch.
-		if err := a.hardStopPipeline(pachClient, pipelineInfo); err != nil {
```
PFedak, line 2382: I still need to work out the right way to close open commits. Straightforwardly converting the logic here into transaction versions still seems like it could miss jobs that aren't created yet, unless I'm missing some defense against that.

src/server/pps/server/api_server.go
```diff
@@ -1822,7 +1814,7 @@ func (a *apiServer) validatePipelineRequest(request *pps.CreatePipelineRequest)
 	return nil
 }
 
-func (a *apiServer) validatePipeline(pachClient *client.APIClient, pipelineInfo *pps.PipelineInfo) error {
+func (a *apiServer) validatePipelineInTransaction(txnCtx *txnenv.TransactionContext, pipelineInfo *pps.PipelineInfo) error {
```
PFedak, line 1817: a double pointer always just looks wrong to me, but it seemed a lot more natural than making sure an extra return value is included on all the exit paths, even with errors

src/server/pps/server/api_server.go
```diff
@@ -2362,27 +2369,96 @@ func (a *apiServer) CreatePipeline(ctx context.Context, request *pps.CreatePipel
...
+			specCommit = *prevSpecCommit
+		}
+		if specCommit != nil {
+			if !update {
+				// for new pipelines, we'll check for newness when the pipeline collection entry is created
+				return specCommit, nil
+			}
+			// make sure the pipeline branch hasn't changed outside of the transaction
+			// since the spec commit was created
+			if err := a.sudo(txnCtx.Client, func(superClient *client.APIClient) error {
```
PFedak, line 2388: I'm starting to suspect some of this logic is misguided and should rely on the STM transaction protections to just work™, but I don't quite see how

src/client/transaction.go
```diff
@@ -35,6 +35,30 @@ func (c APIClient) WithTransaction(txn *transaction.Transaction) *APIClient {
 	return c.WithCtx(ctx)
 }
 
+// WithoutTransaction returns a new APIClient which will run all future operations
+// outside of any active transaction
+// Removing from both incoming and ougoing metadata is necessary because Ctx() merges them
```
PFedak, line 40: This feels really messy, but I don't see a way around it as long as Ctx() indiscriminately combines metadata. I could also explicitly skip that key in AddMetadata but I wasn't confident about all GRPC paths that we care about.

PFedak (commented):
Overall, I'm a little uneasy at adding more STM retries, which slows everything down. Is there a good way to monitor that/rule of thumb for what would be unacceptable?

src/server/pps/server/api_server.go
PFedak: FWIW it's definitely resulting from the actual etcd fetch not finding the commit that was just made. I'm not entirely sure I understand the model, here - the etcd client being used in pfs through the grpc is different than the one used in the transaction, but is the underlying data store distributed, leading to an eventual-consistency situation?

src/server/pps/server/api_server.go
Tryneus: You could have it always pass in a commit, but a zero-value commit is treated the same as a `nil` is now.

src/server/pps/server/api_server.go
```diff
@@ -1979,13 +1952,22 @@ func (a *apiServer) hardStopPipeline(pachClient *client.APIClient, pipelineInfo
...
 			})
 		}
 	}
-	return nil
+
+	// We're not stopping the old pipeline right away, so inspect the output branch directly to ensure
+	// we don't miss any new commits that might be created
+	_, err = txnCtx.Pfs().InspectBranchInTransaction(txnCtx, &pfs.InspectBranchRequest{
+		Branch: client.NewBranch(pipelineInfo.Pipeline.Name, pipelineInfo.OutputBranch),
+	})
```
Tryneus, line 1966: I'm not sure what this is doing or preventing really?

src/server/pps/server/api_server.go
```diff
@@ -1979,13 +1952,22 @@ func (a *apiServer) hardStopPipeline(pachClient *client.APIClient, pipelineInfo
 		}
 		if ci.Finished == nil {
 			// Finish the commit and don't pass a tree
-			pachClient.PfsAPIClient.FinishCommit(pachClient.Ctx(), &pfs.FinishCommitRequest{
+			txnCtx.Pfs().FinishCommitInTransaction(txnCtx, &pfs.FinishCommitRequest{
```
Tryneus, line 1955: This isn't really a 'safe' operation, in my opinion, because the set of commits might change because the `ListCommitStream` is done outside of the transaction.  This may need to use the commits collection directly, although it would be kind of ugly to do that from PPS.  Also (just something to think about, not sure if it's actually a good idea), but would this function make sense to do deferred to the end of the transaction the same way we do `PropagateCommit` in PFS - we could append a similar task to be performed at the end, and let all the code for that stay in PFS.

src/server/pps/server/api_server.go
Tryneus: Yes, this `sudo` stuff is kind of sketchy, but for this branch in particular (when we already know the commit ID and we (probably) don't need to recreate it, I think this can be done in the transaction - that would require a `InspectCommitInTransaction`, but reads are still kind of ugly to do in transactions.

src/client/transaction.go
Tryneus: I don't like that we have to do this, but this is ~fine for now in my opinion.

Tryneus (commented):
Looks pretty good.  We're definitely running up against some limitations of transactions as they stand, but a lot of that is due to legacy behavior, I think.  Left a few comments with suggestions, but I don't see any big problems - let me know what you think.

Tryneus (approved):
LGTM with one change

src/server/pfs/server/transaction_defer.go
```diff
@@ -42,3 +43,61 @@ func (t *Propagater) PropagateCommit(branch *pfs.Branch, isNewCommit bool) error
...
+	return &PipelineFinisher{
+		d:      a.driver,
+		txnCtx: txnCtx,
+	}
+}
+
+// FinishPipelineCommits marks a branch as belonging to a pipeline which is being modified and thus
+// needs old open commits finished
+func (f *PipelineFinisher) FinishPipelineCommits(branch *pfs.Branch) error {
+	if branch == nil {
```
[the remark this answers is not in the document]
PFedak: Ah, of course

PR #5435, opened 2020-11-12
Fix TestPipelineBuildLifecycle & fix race in PutFile

msteffen:
This is the 1.11.x version of #5434


msteffen:
Merging without review as this is just a rebase of #5434, which was already reviewed, onto 1.11.x (minus a testing-specific commit)

PR #5446, opened 2020-11-17
[docs] update style guide to use correct shell indicator

echohack:

Directing a user to invoke a command via documentation is done by using a prompt indicator--`$` is a standard convention. Using a prompt indicator avoids confusing situations where commands are followed by output, other commands, or commands run on a remote shell.

Using the prompt indicator offers an advantage as it allows you to specify the output of commands in the same code block. The proximity of the command and output prevents unnecessarily breaking the output apart from the command, while retaining clarity of the command that was executed.

Using a prompt indicator introduces a usability problem, where users cannot easily copy and paste the command from the documentation. This is an accepted drawback. One workaround is that static site generators may provide a copy button for code blocks, which will copy the code without the prompt indicator or any output following the command.

If this pull request is accepted, I will followup with another PR to fix all existing code examples to use the correct prompt indicator.

### DIFF
--- a/doc/docs/master/contributing/docs-style-guide.md
+++ b/doc/docs/master/contributing/docs-style-guide.md
@@ -104,7 +104,7 @@ a preposition.
 | ---------- | --- |
 | The `put file` API includes an option for splitting **up** the file into separate datums automatically. | The `put file` API includes an option for splitting the file into separate datums automatically. |
 
-##Use meaningful links
+## Use meaningful links
 
 Link text should mean something to the users when they read it. Phrases
 like **Click here** and **Read more** do not provide useful information.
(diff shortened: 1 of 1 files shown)


JoeyZwicker:
Are we sure this is an acceptable drawback? We used to have all codeblocks done this way and purposely changed it (knowing it would reduce readability slightly) in favor of better copy/paste usability. If we want to reverse that decision, we can, but this was explicitly changed to it's existing form after users expressing interest in improve copy/paste functionality. 

This is not something I feel very strongly about either way, but it was a conscious decision, not just what we decided to do on a first pass. 

There may be a different implementation of the copy/paste functionality that would marry the two, but we might have to develop it ourselves. Specifically, having the copy button, instead of copying whatever is in the codeblock, to specifically cope a defined command that is built into the docs. This is more annoying & time consuming for anyone writing documentation, but it probably the best user experience.

echohack:
I'll go ahead and close this issue. I think learning more around the direction of the docs, tooling, and framework, will inform this topic in the near future.

PR #5449, opened 2020-11-18
Fixes issue with boltdb cursor usage when deleting directories

brycemcanally:
Deleting keys while traversing with a cursor results with undefined behavior. The recommended solution is to reposition (seek) the cursor after each deletion according to the docs: https://godoc.org/github.com/boltdb/bolt#Cursor. I added a test that surfaces the original issue this is addressing (a directory deletion not deleting all of the files) and creates a scenario where a deletion progresses the cursor and one that doesn't (my original attempt at solving this was based on the observation that deleting a key progressed the cursor, but that is not consistent behavior).


jdoliner (approved):
LGTM

PR #5451, opened 2020-11-18
Add tests for object clients and deployments

Tryneus:
This adds new tests in `src/server/pkg/obj/testing` (in a subdirectory to avoid a circular dependency with the `testutil` package):

* `client_test.go` directly tests each type of client with a set of basic read/write operations
* `deployment_test.go` runs an end-to-end test of spinning up a pachyderm deployment and running a basic pipeline

The client tests are fairly straightforward.  The deployment tests took a lot of tweaking and trial-and-error.  We generate the manifest by directly calling into the `assets` package, then use the JSON Encoder interface to rewrite `NodePort` services to use dynamically-allocated external ports, and to reduce CPU resource requests to "0" for all pods so that pods can be scheduled more-easily.  After that, we deploy the manifest to a unique namespace and look up the exposed pachd port to connect a pachClient for running the test.

A few tangential changes:
* no longer forbid a local persistent disk from accompanying a non-local object storage
* moved several kubernetes-related calls from `pachyderm_test.go` to the `testutil` package
* fixed some error comparisons from `os.IsNotExist` which doesn't support our stackful errors
* made our cli confirmation code more consistent using `cmdutil.InteractiveConfirm`, have it write to stderr rather than stdout
* exported some enum values from the `assets` package so that already-exported functions can be called
* update the Minio client to the latest `v6`, which lets us override the `PartSize` of uploads to avoid a huge allocation per upload

The new tests uncovered several issues that are not going to be fixed in this PR, the relevant tests are just skipped at the moment:
* Object client interruption does not work in most of our clients (Amazon, Minio, Microsoft)
* The Microsoft client cannot do a range read of only the first byte of an object
* The Minio client does not correctly write out zero-size objects on S3v2 on the Amazon or ECS backends
* The Amazon client gets `InvalidArgument` errors when running against the Google backend (which should be S3-compatible, the Minio client handles it fine)


src/server/pfs/server/obj_block_api_server.go
```diff
@@ -299,19 +299,27 @@ func (s *objBlockAPIServer) PutObjectSplit(server pfsclient.ObjectAPI_PutObjectS
 	}
 	var done bool
 	for !done {
+		logrus.Infof("PutObjectSplit 1")
```
msteffen, line 302: I think these are some debug logs that got left in?

Disappointed by the lack of swearing tbh

src/server/pkg/obj/microsoft_client.go
```diff
@@ -57,6 +57,12 @@ func blobRange(offset, size uint64) *storage.BlobRange {
 	} else if size == 0 {
 		return &storage.BlobRange{Start: offset}
 	}
+	// TODO: This doesn't handle a read with offset: 0, size: 1 correctly - it
+	// will end up with Start: 0, End: 0, which the client library interprets as
+	// the full range of the object. Tried a workaround with manually providing
+	// the 'Range' header, but could not make it work, it may be impossible at the
+	// Azure service level. Possible workaround - read an extra byte, and wrap the
+	// reader to throw away the extra byte.
```
msteffen, line 65: Yet another tragic story of a programmer lost to the darkness of not using right-open intervals.

Though our work will never be done, we must continue to spread the light of right-open, that we may save as many as possible from this fate.

src/server/pkg/obj/testing/deployment_test.go
```diff
@@ -0,0 +1,359 @@
...
+		if innerCb != nil {
+			err = innerCb(data)
+		}
+		rewriteNodePort(data)
+		return err
+	}
+}
+
+func rewriteNodePort(data map[string]interface{}) {
+	if data["kind"] == "Service" {
```
msteffen, line 47: I'm still not 100% sure I understand these tests, but would it make sense to declare a const string JSON patch and then apply it? (and just make sure the patch was applied without errors)

I'll try to look up the library that does this, but we use it in PPS for `pod_patch`

src/server/pkg/obj/testing/util.go
```diff
@@ -0,0 +1,82 @@
...
+		DisableSSL:     obj.DefaultDisableSSL,
+		NoVerifySSL:    obj.DefaultNoVerifySSL,
+		LogOptions:     obj.DefaultAwsLogOptions,
+	}
+}
+
+func LoadAmazonParameters(t *testing.T) (string, string, string, string) {
+	id := os.Getenv("AMAZON_DEPLOYMENT_ID")
+	secret := os.Getenv("AMAZON_DEPLOYMENT_SECRET")
+	bucket := os.Getenv("AMAZON_DEPLOYMENT_BUCKET")
```
msteffen, line 28: A few high-level questions:
- Do we create a new AWS bucket for each test run? Is that in this PR (and I just missed it) or is that separate?
- IIUC, this is all passed in via env vars so that we can run it in CI without exposing our cloud provider creds? Is that right? By extension, we'll have to set up these env vars in e.g. Travis?

src/server/pkg/testutil/kubeclient.go
```diff
@@ -146,3 +148,58 @@ func DeletePipelineRC(t testing.TB, pipeline string) {
...
+	pod, ok := e.Object.(*v1.Pod)
+	if !ok {
+		return false, errors.Errorf("unexpected object type in watch.Event")
+	}
+	return pod.Status.Phase == v1.PodRunning, nil
+}
+
+// WaitForPachdReady finds the pachd pods within the kubernetes namespace and
+// blocks until they are all ready.
+func WaitForPachdReady(t testing.TB, namespace string) {
```
msteffen, line 174: It'll be really nice to have a Go implementation of this logic. I think right now we have similar logic implemented in a few places, and maybe this will let us consolidate later? I'm thinking of:
- etc/reset.py
- etc/kube/start-minikube-vm.sh
- The hub backend, I think? (though I think the hub backend uses kubectl in some of the places that this uses the kubernetes Go client?)

src/server/pkg/obj/testing/deployment_test.go
```diff
@@ -0,0 +1,359 @@
...
+	require.NoError(t, err)
+
+	if cleanup {
+		defer func() {
+			err := kubeClient.CoreV1().Namespaces().Delete(namespaceName, nil)
+			require.NoError(t, err)
+		}()
+	}
+
+	cmd := tu.Cmd("kubectl", "apply", "--namespace", namespaceName, "-f", "-")
```
msteffen, line 195: At one point I spent a lot of time trying to figure out how to do this with the Go client and never figured it out.

src/server/pkg/obj/testing/client_test.go
```diff
@@ -0,0 +1,282 @@
...
+	t.Run("AmazonObjectStorage", func(t *testing.T) {
+		t.Parallel()
+		id, secret, bucket, region := LoadAmazonParameters(t)
+		amazonTests(t, id, secret, bucket, region, "")
+	})
+
+	// Test the Amazon client against ECS
+	t.Run("ECSObjectStorage", func(t *testing.T) {
+		t.Parallel()
+		id, secret, bucket, region, endpoint := LoadECSParameters(t)
```
msteffen, line 212: Can we deploy ECS inside of CI? Is ECS open-source (or like, generally deployable)?

Or is this just here in case we need to do like one-off testing later?

src/server/pkg/obj/testing/client_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+	t.Run("TestSubdirectory", func(t *testing.T) {
+		t.Parallel()
+		object := path.Join(tu.UniqueString("test-subdirectory-"), "object")
+		doWriteTest(t, client, object, []string{"foo", "bar"})
+	})
+
+	// TODO: implement walk test
+
+	t.Run("TestInterruption", func(t *testing.T) {
```
msteffen, line 139: Smart!

src/server/pkg/obj/testing/deployment_test.go
```diff
@@ -0,0 +1,359 @@
...
+	require.NoError(t, err)
+
+	err = assets.WriteSecret(encoder, secrets, opts)
+	require.NoError(t, err)
+
+	return manifest.String()
+}
+
+func withManifest(t *testing.T, backend assets.Backend, secrets map[string][]byte, callback func(namespace string, pachClient *client.APIClient)) {
+	namespaceName := tu.UniqueString("deployment-test-")
```
msteffen, line 165: Is the idea behind testing in separate namespaces that we can test all of the object storage clients simultaneously? I guess this is an empirical question, but is CI big enough to run like 10(?) Pachyderms simultaneously?

Actually, if so, maybe we should start parallelizing the PPS and auth tests...

msteffen (commented):
Just a quick initial pass, and a few comments/questions based on first impressions. Mostly snark.

src/server/pkg/obj/testing/deployment_test.go
Tryneus: Yeah, I looked briefly and it seemed like you just have to create each type of object manually, so I noped out.

src/server/pkg/obj/testing/client_test.go
Tryneus: I actually got a developer account with portal.ecstestdrive.com, and this uses the creds for that.

src/server/pkg/obj/testing/deployment_test.go
Tryneus: Since there's a significant amount of waiting, I got quite a bit of benefit from the parallelism locally when testing - a big part of the goal of this was to make these tests parallelizable.  They needed to be in a separate namespace anyway to avoid stepping on a developer's default deploy, so I figured I'd go all out.  On my machine, these tests can all run in under a minute.

src/server/pfs/server/obj_block_api_server.go
Tryneus: Yep yep, cleaned this up before I even saw this comment =P

src/server/pkg/obj/testing/deployment_test.go
Tryneus: There are a couple hurdles with that approach.  First, the manifest isn't actually json, it's a set of newline-delimited json objects, so I'd have to parse and apply the patch to each separately (annoying but fine).  Second, the rewriting here is conditional, but JSON patches don't support that (I think) - we could either test the type of the object before applying (which means we need to parse the JSON), or just ignore errors (which may break silently if we aren't careful).  Finally, I don't see a way to write a patch that applies to all elements in a subarray (such as replacing the cpu for all containers in a deployment template), which would make the patch much harder to apply.

I gave a shot at moving this to jsonpatch (after writing half this comment) and I don't think it is worth it at the moment. 
 This code is a little fragile and ugly, but it's fairly straightforward and does the job for now.  If we ever have new requirements for this, we can reevaluate =(

src/server/pkg/obj/testing/util.go
Tryneus: * creating a new bucket is hard and it makes cleanup harder, so these just use preconfigued buckets
  * this lets us pare down our creds to the minimal set of permissions to R/W access in one bucket, which is pretty nice
  * I'm not certain that there won't be flakiness from multiple deployments using the same bucket, but if it comes up we can add namespacing to our object clients.
* Yep - I'm trying to get this running in CircleCI at the moment - the creds are also in 1password

src/server/config/cmds.go
```diff
@@ -48,6 +48,7 @@ func Cmds() []*cobra.Command {
 	commands = append(commands, cmdutil.CreateAlias(getMetrics, "config get metrics"))
 
 	setMetrics := &cobra.Command{
+		Use:   "{{alias}} (true | false)",
```
msteffen, line 51: Nice catch! Were these commands just unusable before?

src/server/config/cmds.go
```diff
@@ -345,7 +350,7 @@ func Cmds() []*cobra.Command {
 		Short: "Manages the pachyderm config.",
 		Long:  "Gets/sets pachyderm config values.",
 	}
-	commands = append(commands, cmdutil.CreateDocsAlias(configDocs, "config", "^pachctl config "))
+	commands = append(commands, cmdutil.CreateDocsAlias(configDocs, "config", " config "))
```
msteffen, line 353: I don't really understand what `CreateDocsAlias` does, but should the second `" config "` have spaces around it?

src/server/pkg/obj/microsoft_client.go
msteffen: Marking resolved, though this darkness will not be resolved in our lifetimes, if ever.

msteffen (approved):
Just a few general questions, but overall LGTM

.circleci/config.yml
```diff
@@ -76,3 +76,4 @@ workflows:
               - PPS5
               - PPS6
               - EXAMPLES
+              - OBJECT
```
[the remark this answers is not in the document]
Tryneus: I'm working on getting these running in Travis at the moment.

src/server/config/cmds.go
Tryneus: No, they just didn't have the correct usage strings if you messed up or used `--help`.

src/server/config/cmds.go
Tryneus: This adds all of the matching commands in the 'Associated Commands' section of the help.  Specifically this didn't work on my machine because this assumes the name of the executable to be `pachctl`, whereas for me the help is:

```
$ pachctl help config
Gets/sets pachyderm config values.

Associated Commands:
  C:\Users\grey\go\bin\pachctl.exe config delete context      Deletes a context.
  C:\Users\grey\go\bin\pachctl.exe config get active-context  Gets the currently active context.
  C:\Users\grey\go\bin\pachctl.exe config get context         Gets a context.
  C:\Users\grey\go\bin\pachctl.exe config get metrics         Gets whether metrics are enabled.
  C:\Users\grey\go\bin\pachctl.exe config list context        Lists contexts.
  C:\Users\grey\go\bin\pachctl.exe config set active-context  Sets the currently active context.
  C:\Users\grey\go\bin\pachctl.exe config set context         Set a context.
  C:\Users\grey\go\bin\pachctl.exe config set metrics         Sets whether metrics are enabled.
  C:\Users\grey\go\bin\pachctl.exe config update context      Updates a context.
```

Not that this situation is good, but it's _more_ correct.

PR #5459, opened 2020-11-20
Add identity server alpha

actgardner:
Adds an embedded identity server with a new API and CLI for configuration, based on Dex. The goal is to replace the existing auth integrations with a more modular architecture, without adding a new pod to deployments.

- This PR adds the identity service behind a feature flag, which is enabled for integration tests but disabled for customers by default. 
- The postgres connection is instantiated on the first API or web request - this is necessary because of the structure of our test scripts. This way pachd is able to start before postgres is available. 
- The web server is also started lazily when the first web request comes in. If no IDP connectors are configured dex will return an error. We retry starting the server for each subsequent web request until it succeeds.
- The web server is restarted if the configuration is changed. This avoids the need to redeploy pachd to change the `Issuer` config.

The full steps to configure Dex and Pachyderm to do Github auth on minikube are here: https://gist.github.com/actgardner/d11545796112003c637882571df4b357


etc/pachd/Dockerfile
```diff
@@ -1,4 +1,5 @@
 FROM scratch
 COPY --from=pachyderm_build /app/pachd /pachd
+COPY --from=pachyderm_build /app/web /web
```
actgardner, line 3: Dex has static web assets that need to be included in the Docker image.

go.mod
```diff
@@ -119,3 +121,5 @@ replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+inc
 replace google.golang.org/grpc => google.golang.org/grpc v1.26.0
 
 replace k8s.io/client-go => k8s.io/client-go v0.0.0-20190718183610-8e956561bbf5
+
+replace github.com/dexidp/dex => github.com/pachyderm/dex v0.0.0-20201118163231-aa20083dfa7f
```
actgardner, line 125: For now we have a fork that has one change - it supports starting with no IDP configured, and adding one later. We can probably upstream this or work around it another way in the future to avoid maintaining a fork.

src/client/identity/identity.proto
```diff
@@ -0,0 +1,67 @@
+syntax = "proto3";
+
+package identity;
+option go_package = "github.com/pachyderm/pachyderm/src/client/identity";
```
actgardner, line 4: The identity API is basically a more pachyderm-idiomatic version of the dex grpc API. One addition is the ability to configure IDP connectors - we should be able to upstream this but it may happen post 2.0.

src/client/identity/identity.proto
```diff
@@ -0,0 +1,67 @@
...
+
+/* ConnectorConfig represents an IDP to delegate authentication to */
+message ConnectorConfig {
+  string id = 1;
+  string name = 2;
+  string type = 3;
+  int64 configVersion = 4;
+  string jsonConfig = 5;
```
actgardner, line 12: Conveniently Dex just treats IDP config as a JSON blob, like our existing API. This API should be able to configure any provider they support upstream with no changes.

src/server/auth/server/config.go
```diff
@@ -42,6 +42,7 @@ type canonicalGitHubIDP struct{}
 
 type canonicalOIDCIDP struct {
 	Issuer              string
+	IssuerOverride      string
```
actgardner, line 45: `IssuerOverride` tells pachd to try and reach Dex at a different hostname than is specified in `Issuer`. This is necessary if DNS resolves differently within the cluster and outside of it. For example, when we're running in minikube Dex is reachable at `localhost:30658` from the host machine, but at `pachd:658` from within the pachd pod. `Issuer` should be configured so it's accurate from the host machine's perspective, so the OAuth redirect links resolve properly in the user's browser.

src/server/auth/server/http_client.go
```diff
@@ -0,0 +1,20 @@
+package server
+
+import (
+	"net/http"
+)
+
+type RewriteRoundTripper struct {
+	RewriteHost string
```
actgardner, line 8: TODO: This should support rewriting the protocol (`http` versus `https`) and it should check the URL to be rewritten.

src/server/auth/server/http_client.go
```diff
@@ -0,0 +1,20 @@
...
+}
+
+func RewriteClient(rewriteHost string) *http.Client {
```
actgardner, line 11: Returns an `http.Client` we can inject into the OIDC and OAuth libraries which silently rewrites requests where the host is `Issuer` to `IssuerOverride`.

src/server/identity/server/api_server.go
```diff
@@ -0,0 +1,141 @@
...
+		pachLogger: log.NewLogger("identity.API"),
+		server:     server,
+	}, nil
+}
+
+func (a *apiServer) CreateConnector(ctx context.Context, req *identity.CreateConnectorRequest) (resp *identity.CreateConnectorResponse, retErr error) {
+	a.LogReq(req)
+	defer func(start time.Time) { a.LogResp(req, resp, retErr, time.Since(start)) }(time.Now())
+
+	if err := a.server.createConnector(req.Config.Id, req.Config.Name, req.Config.Type, int(req.Config.ConfigVersion), []byte(req.Config.JsonConfig)); err != nil {
```
actgardner, line 56: These are thin wrappers around the existing Dex API that we an add auth to.

src/server/identity/server/auth.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/dexidp/dex/storage/sql"
+	logrus "github.com/sirupsen/logrus"
+)
+
+const (
+	dexHttpPort = ":658"
+)
+
+type dexServer struct {
+	api.DexServer
```
actgardner, line 21: The existing Dex grpc API supports configuring clients but not connectors. If we upstream connector support we may be able to merge this file with `api_server.go`.

src/server/identity/server/auth.go
```diff
@@ -0,0 +1,112 @@
...
+			User:     pgUser,
+			Password: pgPwd,
+			Host:     pgHost,
+			Port:     uint16(pgPort),
+		},
+		SSL: sql.SSL{
+			Mode: pgSSL,
+		},
+	}
+	logger := logrus.NewEntry(logrus.New()).WithField("source", "dex")
```
actgardner, line 39: TODO: better logging

src/server/identity/server/auth.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	return &dexServer{
+		dexStorage: storage,
+		DexServer:  server.NewAPI(storage, nil),
+	}, nil
+}
+
+func (s *dexServer) createConnector(id, connType, name string, resourceVersion int, jsonConfig []byte) error {
+	conn := storage.Connector{
```
actgardner, line 75: TODO: validate inputs

src/server/pkg/serviceenv/config.go
```diff
@@ -69,7 +69,16 @@ type PachdSpecificConfiguration struct {
 	RequireCriticalServersOnly bool   `env:"REQUIRE_CRITICAL_SERVERS_ONLY",default=false"`
 	MetricsEndpoint            string `env:"METRICS_ENDPOINT",default="`
 	// TODO: Merge this with the worker specific pod name (PPS_POD_NAME) into a global configuration pod name.
-	PachdPodName string `env:"PACHD_POD_NAME,required"`
+	PachdPodName        string `env:"PACHD_POD_NAME,required"`
+	PostgresServiceHost string `env:"POSTGRES_SERVICE_HOST,default=172.17.0.1"`
+	PostgresServicePort int    `env:"POSTGRES_SERVICE_PORT,default=32228"`
+	PostgresServiceSSL  string `env:"POSTGRES_SERVICE_SSL,default=disable"`
+
+	EnterpriseServerEnabled bool   `env:"ENTERPRISE_SERVER_ENABLED,default=true"`
```
actgardner, line 77: TODO: wire these options into `pachctl deploy`

Makefile
```diff
@@ -400,15 +401,6 @@ launch-loki:
 clean-launch-loki:
 	helm uninstall loki
 
-launch-dex:
```
actgardner, line 403: We no longer need an external Dex server because we have the embedded server.

src/server/auth/server/testing/oidc_test.go
```diff
@@ -101,38 +142,70 @@ func TestOIDCAuthCodeFlow(t *testing.T) {
...
+	// pachd is configured to reach dex with kube dns, but the tests might be
+	// outside the cluster.
+	c := &http.Client{}
+	c.CheckRedirect = func(_ *http.Request, via []*http.Request) error {
+		return http.ErrUseLastResponse
 	}
 
-	c := &http.Client{}
-	resp, err := c.PostForm(fmt.Sprintf("http://%v/token", dexHost(testClient)), vals)
+	oauthConfig := oauth2.Config{
```
actgardner, line 166: I rewrote this to do a normal auth code flow for the test app instead of the implicit flow. Dex doesn't support enabling/disabling implicit flow at runtime, and we only needed it for this test.

src/server/auth/server/oidc.go
```diff
@@ -178,16 +189,7 @@ func (o *InternalOIDCProvider) GetOIDCLoginURL(ctx context.Context) (string, str
 	if o == nil {
 		return "", "", errors.WithStack(errNotConfigured)
 	}
-	// TODO(msteffen, adelelopez): We *think* this 'if' block can't run anymore:
```
actgardner, line 181: I also don't think this can run anymore.

src/client/identity/identity.proto
```diff
@@ -0,0 +1,123 @@
+syntax = "proto3";
+
+package identity;
+option go_package = "github.com/pachyderm/pachyderm/src/client/identity";
+
+message IdentityConfig {
+	string issuer = 1;
```
actgardner, line 7: Normally in a dex server the issuer is configured with a config file at startup, but when you're initially setting up it can be tricky to get the right value. I pulled out this config and made the web server hot reloadable so users don't need to redeploy to iterate on the correct value.

Makefile
```diff
@@ -215,7 +215,7 @@ launch: install check-kubectl
 
 launch-dev: check-kubectl check-kubectl-connection install
 	$(eval STARTTIME := $(shell date +%s))
-	pachctl deploy local --no-guaranteed -d --dry-run $(LAUNCH_DEV_ARGS) | kubectl $(KUBECTLFLAGS) apply -f -
+	pachctl deploy local --no-guaranteed -d --dry-run --id-server $(LAUNCH_DEV_ARGS) | kubectl $(KUBECTLFLAGS) apply -f -
```
actgardner, line 218: Enable the id server feature flag when we're running tests

Makefile
```diff
@@ -326,6 +326,7 @@ test-local:
 	CGOENABLED=0 go test -count=1 -cover -short $$(go list ./src/server/... | grep -v '/src/server/pfs/fuse') -timeout $(TIMEOUT)
 
 test-auth:
+	./etc/testing/start_postgres.sh
```
actgardner, line 329: Auth tests now require postgres as a backing store for the embedded Dex server.

src/server/cmd/pachd/main.go
```diff
@@ -424,6 +426,32 @@ func doFullMode(config interface{}) (retErr error) {
 				}
 			}
 		}
+		if env.IdentityServerEnabled {
```
actgardner, line 429: Add a feature flag to enable the identity server (off by default).

Makefile
msteffen: Are the auth tests separate from the identity tests? Both require postgres, or just auth?

goreleaser/docker.yml
```diff
@@ -47,6 +47,7 @@ dockers:
       - go.mod
       - go.sum
       - src
+      - web
```
msteffen, line 50: I know this is superficial, but would it make sense to call this like `internal_dex.config` or something?

src/server/auth/server/config.go
msteffen: This comment is super helpful (especially because it kind of explains both `Issuer` and `IssuerOverride`)—how would you feel about using it as a field comment (i.e. in the code)?

msteffen (commented):
Still going through this, but I thought I'd post the few random comments I had

src/client/auth/auth.proto
```diff
@@ -101,6 +101,7 @@ message IDProvider {
      string redirect_uri = 4 [(gogoproto.customname) = "RedirectURI"];
      repeated string additional_scopes = 5;
      bool ignore_email_verified = 6;
+     string issuer_override = 7;
```
msteffen, line 104: This is kind of speculative, but this generally seems like an appealing attack vector to me (since if you can convince Pachyderm to verify a token with your compromised MITM server, you can authenticate as anyone). I know the auth config should only be settable by Pachyderm cluster admins anyway, but I wonder if it would make sense to restrict this field to dev clusters (the same kind that use fake GitHub auth). If customers need it in prod, maybe there's a good way to limit the values that are set (e.g. it must resolve to a private IP address).

src/server/identity/server/dex_api.go
```diff
@@ -0,0 +1,334 @@
...
+	"github.com/pkg/errors"
+	logrus "github.com/sirupsen/logrus"
+
+	"github.com/pachyderm/pachyderm/src/client/identity"
+)
+
+// dexAPI wraps an api.DexServer and extends it with CRUD operations
+// that are currently missing from the dex api. It also handles lazily
+// instantiating the api and storage on the first request.
+type dexAPI struct {
```
msteffen, line 22: Is this the file that implements hot-reloading of the dex config? How does that happen?

msteffen (commented):
A few more questions (making more progress!)

Makefile
actgardner: Auth tests rely on the identity server, which relies on postgres.

src/client/identity/identity.proto
actgardner: I renamed this to `IdentityServerConfig` - this is configuration for the Dex web server which handles OIDC requests.

src/client/auth/auth.proto
actgardner: Yeah I have two conflicting impulses here - I wanted to make things as configurable as possible, because once code is out in the field configuration is much easier than spinning a new patch to handle an unexpected edge case. But on reflection this does seem like an easy way to escalate admin access to the cluster into impersonating any user (which is already kind of possible but at least it's logged).

I've replaced the `issuer_override` with a boolean `localhost_issuer`, which only supports querying the embedded identity server over the loopback interface. This seems harder to exploit but users can still use the embedded server in minikube.

actgardner:
@msteffen After talking with brendan and bryce I've moved this to a new PR based off the 2.0 alpha branch: https://github.com/pachyderm/pachyderm/pull/5561

echohack:
New PR and commit has merged. Deleting branch.

Issue #5458, opened 2020-11-20
DNS failure causes pachd not to connect


pappasilenus:
**What happened?**:

A user attempted to connect via pachctl 1.11.7 and got a connect error because DNS had a strange "server misbehaving" failure looking up localhost.
Changing the DNS server from the router's default to 8.8.8.8 solved the issue.

This failure does not occur on pachctl 1.10.5.

**What you expected to happen?**:
Expected the connection to occur.

**How to reproduce it (as minimally and precisely as possible)?**:

Getting the brand of the router or dhcp server version and the configured DNS settings. Here's a log of what occurred:

```
olly> pachctl version -v

COMPONENT           VERSION

pachctl             1.11.7

[0000] DEBUG Implicit port forwarder listening on port 55528

[0000]  INFO parsed scheme: "dns" source=etcd/grpc

[0001]  INFO grpc: error from DNS TXT record lookup: lookup _grpc_config.localhost on 192.168.1.1:53: server misbehaving source=etcd/grpc

[0001]  INFO ccResolverWrapper: sending update to cc: {[{127.0.0.1:55528  <nil> 0 <nil>} {[::1]:55528  <nil> 0 <nil>}] 0xc000abe680 <nil>} source=etcd/grpc

could not connect to pachd at "grpc://localhost:55528": context deadline exceeded

github.com/pachyderm/pachyderm/src/client.NewOnUserMachine

                src/client/client.go:463

github.com/pachyderm/pachyderm/src/server/cmd/pachctl/cmd.PachctlCmd.func2

                src/server/cmd/pachctl/cmd/cmd.go:411

github.com/pachyderm/pachyderm/src/server/pkg/cmdutil.RunFixedArgs.func1

                src/server/pkg/cmdutil/cobra.go:31

github.com/spf13/cobra.(*Command).execute

                /home/pach/go/pkg/mod/github.com/spf13/cobra@v0.0.6-0.20191202130430-b04b5bfc50cb/command.go:833

github.com/spf13/cobra.(*Command).ExecuteC

                /home/pach/go/pkg/mod/github.com/spf13/cobra@v0.0.6-0.20191202130430-b04b5bfc50cb/command.go:917

github.com/spf13/cobra.(*Command).Execute

                /home/pach/go/pkg/mod/github.com/spf13/cobra@v0.0.6-0.20191202130430-b04b5bfc50cb/command.go:867

main.main.func1

                src/server/cmd/pachctl/main.go:20

main.main

[...]

PR #5467, opened 2020-11-24
Disable TXT service config lookup in grpc dialer

actgardner:
Resolves https://github.com/pachyderm/pachyderm/issues/5458

It seems like some DNS servers incorrectly return SERVFAIL instead of NXDOMAIN when there's no TXT record for `_grpc_config.<server>`. GRPC 1.26 treats this as a retryable error and backs off until the connection times out.

Example dig output from a bad DNS server:

```
> dig _grpc_config.hub-c0-92zz2shvqr.clusters-staging.pachyderm.io
; <<>> DiG 9.10.6 <<>> _grpc_config.hub-c0-92zz2shvqr.clusters-staging.pachyderm.io
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 14473
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1
;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;_grpc_config.hub-c0-92zz2shvqr.clusters-staging.pachyderm.io. IN A
;; AUTHORITY SECTION:
clusters-staging.pachyderm.io. 900 IN   SOA     ns-526.awsdns-01.net. awsdns-hostmaster.amazon.com. 1 7200 900 1209600 86400
;; Query time: 80 msec
;; SERVER: 192.168.1.1#53(192.168.1.1)
;; WHEN: Tue Nov 24 15:02:11 EST 2020
;; MSG SIZE  rcvd: 173
```

Related GRPC issue: https://github.com/grpc/grpc-go/issues/3207

AFAIK nobody uses the service config feature so it's safe to just disable this.

### DIFF
--- a/src/client/client.go
+++ b/src/client/client.go
@@ -629,6 +629,7 @@ func DefaultDialOptions() []grpc.DialOption {
 	return []grpc.DialOption{
 		// Don't return from Dial() until the connection has been established.
 		grpc.WithBlock(),
+		grpc.WithDisableServiceConfig(),
 		grpc.WithKeepaliveParams(keepalive.ClientParameters{
 			Time:                20 * time.Second,
 			Timeout:             20 * time.Second,


brendoncarroll (approved):
LGTM

Tryneus:
Could you remove https://github.com/pachyderm/pachyderm/blob/1ed494b800228f2251560e0ee2bd84e7fb4c6842/src/client/client.go#L673 in this case?  TL;DR: I had a similar problem in windows, but only in `pachctl` - there are other places that use the default dial options, though (but I didn't have any proof that it was an issue elsewhere).

actgardner:
I just asked someone to try and repro this on master, I'm assuming the issue was they only tried 1.11.x builds where this change hadn't landed. Maybe we should also backport it to the 1.11 branch.

In general turning off service config discovery won't be harmful since we never use it. I think moving this into the default dial options is fine.

PR #5471, opened 2020-11-25
Make debug API compatible with auth

brycemcanally:
This PR makes the debug API compatible with auth and limits the debug API to admin clients.


msteffen (approved):
LGTM

PR #5473, opened 2020-11-30
Make debug API compatible with auth (1.11.x)

brycemcanally:
1.11.x PR for #5471

### DIFF
--- a/src/server/auth/server/testing/auth_test.go
+++ b/src/server/auth/server/testing/auth_test.go
@@ -1,7 +1,9 @@
 package server
 
 import (
+	"archive/tar"
 	"bytes"
+	"compress/gzip"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


msteffen (approved):
LGTM

FWIW, I typically don't seek approval for 1.11.x PRs if they're identical (i.e. a rebase) of changes that were already reviewed and approved in the `master` branch. Maybe that's a mistake, though my understanding is that this is Pachyderm's recommended practice.

Is this branch different from the `master` version? I didn't notice the differences if so.

brycemcanally:
I think it would be a good idea to seek a review / approval for all PRs that will be going into a release branch. Even if the change is identical to a change on master, it would be good to have a sanity check to make sure it is indeed identical (and that it should be identical in the release branch). @jdoliner and / or @nitinjainsj should probably decide on this, but my opinion is that a review / approval should be required for anything that is going into a release.

PR #5475, opened 2020-11-30
Additive / Deletive Set

brycemcanally:
This PR pulls file operation information out of the indexing layer, and stores it in the fileset layer (in the form of an additive and deletive set that is stored in a fileset protocol buffer). Also, this removes tar headers from the content stream, and therefore simplifies the handling of files and directories at merge time. Directory metadata is generated at read time and will take advantage of caching.


src/server/pkg/storage/fileset/api.go
```diff
@@ -24,7 +21,7 @@ var _ File = &FileReader{}
 // FileSet represents a set of files.
 type FileSet interface {
 	// Iterate iterates over the files in the file set.
-	Iterate(ctx context.Context, cb func(File) error) error
+	Iterate(ctx context.Context, cb func(File) error, deletive ...bool) error
```
brendoncarroll, line 24: I don't like how there is a parameter, which I have 50/50 shot of getting wrong, that is also implicit.

src/server/pkg/storage/fileset/fileset.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package fileset;
+option go_package = "github.com/pachyderm/pachyderm/src/server/pkg/storage/fileset";
+
+import "server/pkg/storage/fileset/index/index.proto";
+
+message Metadata {
```
brendoncarroll, line 8: We have talked offline about this. It's not really Metadata. This object is in essence the fileset, not additional information.  `Layer`, `ODF`, `Op` `OpSet` `ChangeSet` `Patch` `Delta` are all better names.

src/server/pkg/storage/fileset/index/index.proto
```diff
@@ -20,22 +19,13 @@ message Range {
...
-}
-
-// TODO: Overwrite op is technically unecessary, but it might be better to be explicit.
-enum Op {
-  APPEND = 0;
-  DELETE = 1;
-  OVERWRITE = 2;
+message File {
+  repeated Part parts = 1;
+  repeated chunk.DataRef data_refs = 2;
```
brendoncarroll, line 24: Why do we also need data_refs for a File?  I would prefer using a Part with an empty tag or something like that.

src/server/pkg/storage/fileset/index/util.go
```diff
@@ -53,17 +53,17 @@ func PointsTo(idx *Index) (ids []chunk.ID) {
 	return ids
 }
 
-func resolveDataOps(idx *Index) {
-	if idx.FileOp.DataRefs == nil {
+func resolveParts(idx *Index) {
```
brendoncarroll, line 56: What are the performance implications of not doing this in place? I think it would be clearer if this returned a new value.

src/server/pkg/storage/fileset/merge.go
```diff
@@ -28,153 +27,169 @@ func newMergeReader(chunks *chunk.Storage, fileSets []FileSet) *MergeReader {
...
+					break
+				}
+			}
 		}
-		if len(files) == 1 {
-			return cb(files[0])
+		if len(fss) == 0 {
+			return nil
+		}
+		if IsDir(fss[0].key()) {
```
brendoncarroll, line 69: I didn't think there were directories at this point?

src/server/pkg/storage/fileset/merge.go
```diff
@@ -28,153 +27,169 @@ func newMergeReader(chunks *chunk.Storage, fileSets []FileSet) *MergeReader {
...
 	}
-	mergeIdx.FileOp.DataOps = append([]*index.DataOp{hdo}, mergeDataOps(dos)...)
+	// Merge the parts based on the lexicograhical ordering of the tags.
+	mergeIdx.File.Parts = mergeParts(ps)
 	return mergeIdx
 }
 
-func mergeDataOps(dataOps [][]*index.DataOp) []*index.DataOp {
-	if len(dataOps) == 0 {
+func mergeParts(pss []*partStream) []*index.Part {
```
brendoncarroll, line 111: It looks like we are performing a merge on streams which are entirely in memory.  Essentially performing heap sort on a slice.  Could we use the "sort" package instead? Is there a path to having these stream from out of memory that we are preparing for?

src/server/pkg/storage/fileset/reader.go
```diff
@@ -1,43 +1,56 @@
...
-func (r *Reader) Iterate(ctx context.Context, cb func(File) error) error {
-	ir, err := r.indexFactory(ctx)
-	if err != nil {
+func (r *Reader) setup(ctx context.Context) error {
+	if r.additive == nil {
+		md, err := r.store.Get(ctx, r.path)
+		if err != nil {
+			return err
+		}
+		r.additive = index.NewReader(r.chunks, md.Additive, r.indexOpts...)
```
brendoncarroll, line 37: Can we return these? `setup(ctx) (additive, deletive, error)` and remove the associated fields from the reader?

src/server/pkg/storage/fileset/reader.go
```diff
@@ -1,43 +1,56 @@
...
+	return nil
+}
+
+// Iterate iterates over the files in the file set.
+func (r *Reader) Iterate(ctx context.Context, cb func(File) error, deletive ...bool) error {
+	if err := r.setup(ctx); err != nil {
 		return err
 	}
-	return ir.Iterate(ctx, func(idx *index.Index) error {
+	if len(deletive) > 0 && deletive[0] {
```
brendoncarroll, line 48: Thinking about this more, we should probably have a second method `IterateDeletes`.

brendoncarroll: RE: PPS `Iterate` vs `ForEach` this might be a good chance to make that change.

src/server/pkg/storage/fileset/writer.go
```diff
@@ -141,76 +103,57 @@ func (w *Writer) nextIdx(idx *index.Index) error {
...
+		if w.idx.Path > p {
+			return errors.Errorf("cannot write path (%s) after (%s)", p, w.idx.Path)
 		}
 	}
 	return nil
 }
 
 // Delete creates a delete operation for a file.
-func (w *Writer) Delete(p string) error {
+// TODO: Check path order.
```
brendoncarroll, line 117: All operations should operate on paths in order.  So we would have a single `lastPath` field for the whole Writer?

src/server/pkg/storage/fileset/writer.go
```diff
@@ -141,76 +103,57 @@ func (w *Writer) nextIdx(idx *index.Index) error {
...
 		SizeBytes: idx.SizeBytes,
 	}
 	if err := w.nextIdx(copyIdx); err != nil {
 		return err
 	}
-	for _, dataRef := range idx.FileOp.DataRefs {
-		if err := w.cw.Copy(dataRef); err != nil {
-			return err
+	// Copy the file data refs if they are resolved.
+	if idx.File.DataRefs != nil {
```
brendoncarroll, line 143: See my comment at the beginning. I think we should use the empty tag for this and just store it as parts.

src/server/pkg/storage/fileset/writer.go
```diff
@@ -264,17 +207,28 @@ func (w *Writer) Close() error {
...
 	var pointsTo []string
-	for _, cid := range index.PointsTo(idx) {
+	for _, cid := range index.PointsTo(additiveIdx) {
+		pointsTo = append(pointsTo, chunk.ObjectID(cid))
+	}
+	for _, cid := range index.PointsTo(deletiveIdx) {
 		pointsTo = append(pointsTo, chunk.ObjectID(cid))
 	}
 	if err := w.tracker.CreateObject(w.ctx, filesetObjectID(w.path), pointsTo, w.ttl); err != nil {
 		return err
```
brendoncarroll, line 227: What do we want to do if the fileset already exists?  `tracker.CreateObject` will return an err, but w.store.Set should be fine.

src/server/pkg/storage/fileset/reader.go
brendoncarroll: If we are introducing `NewIterator` I definitely think we should change these methods to `ForEach`.  Iterate would seems like it should return an object similar to `NewIterator`

src/server/pkg/storage/fileset/api.go
brycemcanally: You could just as easily use the wrong iterate function if we exposed it as `IterateDeletes`. This approach allows for less boilerplate in our current FileSet implementations. I do want to separate the notion of deletions from the File API long term, but this is a pretty low friction way to get this functionality up and running.

src/server/pkg/storage/fileset/index/index.proto
brycemcanally: I think this will get a bit cleaner if we end up moving to the parts only model (index per part rather than per file). With the current model, I think it is clearer to have the data refs for the full file associated with the file data structure rather than a special part. The code would get pretty wonky if that was how we managed the full file data refs.

src/server/pkg/storage/fileset/index/util.go
brycemcanally: I don't think the performance hit would be much of a concern. I agree with this sentiment in general, but I think I will just leave this as a TODO because this is going to get re-written / thrown away if we move to the index entry per part model.

src/server/pkg/storage/fileset/merge.go
brycemcanally: Yeah, but directory deletions are still captured in the deletive set as directory paths. Deleting a directory means deleting all of the files underneath it.

src/server/pkg/storage/fileset/merge.go
brycemcanally: The priority queue implementation handles the heap sort and captures the behavior of collecting entries with the same keys and ordering them by priority (which is important for ordering additive / deletive entries). Although the streaming functionality is not being taken advantage of, I think the features I mentioned above and the simple interface (although I think there is opportunity for improvement here) make it a good choice here.

src/server/pkg/storage/fileset/reader.go
brycemcanally: Yeah, that is a good idea.

src/server/pkg/storage/fileset/reader.go
brycemcanally: Refer to the first thread for `IterateDeletes`, and I think we should separate the global name change from this set of changes.

src/server/pkg/storage/fileset/writer.go
brycemcanally: That is not currently true. Additive operations are ordered with respect to other additive operations, but not with respect to deletive operations (and vice versa). The unordered writer has its own notion of an additive and deletive set and it is serialized additive first followed by deletive. I suppose we could merge them, but that seems unnecessary seeing as they are written to separate index writers. I think we might end drawing a clearer distinction between additive / deletive operations rather than having them operate on the same writer.

src/server/pkg/storage/fileset/writer.go
brycemcanally: Same as above. I disagree, but this is likely to change so it does not really matter.

src/server/pkg/storage/fileset/writer.go
brycemcanally: Wouldn't it be an error to try and create a file set that already exists?

src/server/pkg/storage/fileset/api.go
brendoncarroll: Can you define a separate delete method and comment it out with a TODO

src/server/pkg/storage/fileset/index/util.go
brendoncarroll: OK. TODO is fine.

src/server/pkg/storage/fileset/reader.go
brendoncarroll: I'm fine with adding this as a TODO.

src/server/pkg/storage/fileset/writer.go
brendoncarroll: I think we should add checks for this.  The Writer isn't threadsafe or anything so it should just be a straightforward conditional check. A lot depends on the ordering here being correct.

src/server/pkg/storage/fileset/writer.go
brendoncarroll: The issue is if you successfully create the object, but you don't actually finish adding it to the fileset table.  If the paths are write once, the correct thing to do is ignore `tracker.ErrObject` exists and recreate the object.  If the paths are not write once then we could be in an inconsistent state after an error.

src/server/pkg/storage/fileset/fileset.proto
brendoncarroll: you could call it `INode`

src/server/pkg/storage/fileset/writer.go
brycemcanally: Yeah, I think we need to have these in a single transaction. I added a TODO for this.

PR #5478, opened 2020-12-01
Refactor auth test methods into testutils

actgardner:
The `src/server/auth/server/testing` package has methods to manage activating auth, minting auth tokens for arbitrary users, and resetting the cluster to a pristine state. These are potentially useful when integration testing any module that interacts with the auth system.

This PR refactors those methods into the testutils package. I've made a few comments inline where the change is more than just moving the method between modules.

### DIFF
--- a/src/server/pkg/testutil/auth.go
+++ b/src/server/pkg/testutil/auth.go
@@ -1,0 +1,364 @@
+package testutil
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/pachyderm/pachyderm/src/client"
+	"github.com/pachyderm/pachyderm/src/client/auth"
+	"github.com/pachyderm/pachyderm/src/client/pkg/errors"
+	"github.com/pachyderm/pachyderm/src/client/pkg/require"
+	"github.com/pachyderm/pachyderm/src/server/pkg/backoff"
+)
+
+const (
+	// AdminUser is the sole cluster admin after GetAuthenticatedPachClient is called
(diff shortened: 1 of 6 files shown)


src/server/pkg/testutil/auth.go
```diff
@@ -0,0 +1,364 @@
...
+	return auth.GitHubPrefix + user
+}
+
+// helper method to generate a super admin role
+func superClusterRole() *auth.ClusterRoles {
+	return &auth.ClusterRoles{Roles: []auth.ClusterRole{auth.ClusterRole_SUPER}}
+}
+
+// UpdateAuthToken manually updates the auth token for a given user
+func UpdateAuthToken(user string, token string) {
```
actgardner, line 44: A few activation tests involve minting new admin tokens - this provides a way to update the cache so `DeleteAll` still succeeds in the test teardown.

msteffen (approved):
LGTM.

One idea I had after reading this is that maybe the auth tests could start storing the admin token in the local pach config instead of `/tmp/pach-auth-test_admin-token`, so that `pachctl` commands work automatically after test failure?

We have a retry mechanism in Travis that doesn't currently work for the auth tests because, before retrying tests, it runs `pachctl delete-all` which fails if auth is still on (and likewise `etc/testing/kube_debug.sh`, which runs after test failure, can't run any tests either). 

This is a preexisting problem from our old auth tests, but I guess I'm worried that if we start having more auth-on tests then it'll get even more annoying than it already is, and using the pach config might be an ergonomic workaround.

actgardner:
Yeah caching the admin token using the normal token caching in pachctl makes a lot of sense to me as a follow-on. Once the identity and extract/restore stuff is landed I can take a look at that.

PR #5481, opened 2020-12-01
Improve Travis failure output:

msteffen:
- Fix 'kubectl logs' bug (when you use a selector, e.g. `-l app=pachd`, with `kubectl logs`, it changes the default behavior to only output 10 lines :facepalm:. Setting the `--tail` flag is necessary to get more)
- Get logs from pachd using --previous, in case pachd crashed
- Try to get job and commit info after a failure (e.g. check if a job is stuck in merging)
- beautify failure output (run each command surrounded by `===` so you can quickly scan through them)


etc/testing/kube_debug.sh
```diff
@@ -1,19 +1,29 @@
...
-top -b -n 1 | head -n 20
-df -h
+cmds=(
+  'kubectl get all --all-namespaces'
+  'kubectl version'
+  'kubectl get all --all-namespaces'
+  'kubectl describe pod -l suite=pachyderm,app=pachd'
+  'kubectl describe pod -l suite=pachyderm,app=etcd'
+  # Set --tail b/c by default 'kubectl logs' only outputs 10 lines if -l is set
+  'kubectl logs --tail=100 -l suite=pachyderm,app=pachd'
```
nitinjainsj, line 17: I'd add a tail of 1000. It's better to get more logs.

nitinjainsj (approved):
On minor recommendation.

LGTM

lukemarsden:
Merging as this fixes CircleCI for master (except OBJECT), and doesn't seem like it makes Travis worse.

PR #5487, opened 2020-12-02
Fix windows build following goreleaser migration

Tryneus:
This actually went smoother than expected, but still a couple changes to fix the windows build:

* Change `LD_FLAGS` back to specifying `-X` within the env var, `LD_FLAGS` should be concatenable
* Update the `vscode_compile_task.sh` build script to use `goreleaser`


Makefile
```diff
@@ -14,7 +14,7 @@ else
 	export GC_FLAGS = "all=-trimpath=${PWD}"
 endif
 
-export LD_FLAGS = github.com/pachyderm/pachyderm/src/client/version.AdditionalVersion=$(VERSION_ADDITIONAL)
+export LD_FLAGS="-X github.com/pachyderm/pachyderm/src/client/version.AdditionalVersion=$(VERSION_ADDITIONAL)"
```
nitinjainsj, line 17: The problem with -X here is that it will fail the pachctl release (goreleaser/pachctl.yaml). Hence the need to remove it from the Makefile.

Tryneus: In that case, we should export an environment variable for the key/value pair without the `-X` (which is not `LD_FLAGS`), because this is an improper format for `LD_FLAGS`.  I'll update this.

etc/compile/vscode_compile_task.sh
```diff
@@ -1,25 +1,17 @@
...
\ No newline at end of file
+REPO_DIR=$(git rev-parse --show-toplevel)
+COMPILE_DIR=$(dirname "${BASH_SOURCE[0]}")
+VERSION_ADDITIONAL=-$(git log --pretty=format:%H | head -n 1)
+
+export DOCKER_BUILDKIT=1
+export GOVERSION=$(cat "${COMPILE_DIR}/GO_VERSION")
+export LD_FLAGS="-X github.com/pachyderm/pachyderm/src/client/version.AdditionalVersion=${VERSION_ADDITIONAL}"
+
+goreleaser release -p 1 --snapshot --skip-publish --rm-dist -f "${REPO_DIR}/goreleaser/docker.yml"
```
nitinjainsj, line 17: I am guessing there is a good reason why we cannot call make docker-build here? If you could, this code would look like...
cd $REPO_DIR
make docker-build

nitinjainsj (approved):

etc/compile/vscode_compile_task.sh
Tryneus: So I wrote this to require as few extra tools as possible, and `make` doesn't exist on windows without jumping through a few hoops - unfortunately I think what you said is just the right thing to do, so I rewrote this to use `make` and we can get rid of all this duplication.

nitinjainsj (approved):
Code review that delete code are the best.

LGTM

PR #5480, opened 2020-12-01
Add a mutex to make stm threadsafe which should stop pachd crashing.

lukemarsden:
Matthew and I figured out a [test failure](https://app.circleci.com/pipelines/github/pachyderm/pachyderm/812/workflows/3a508c8a-09cc-48dd-a1db-ed7b575f0bc6/jobs/11453) in TestListDatum was caused by concurrent map writes:

```
2020-12-01T17:17:17Z INFO PPS master: processing event for "alice-pipeline1e1569f6d904"
fatal error: concurrent map writes

goroutine 2646 [running]:
runtime.throw(0x2ed1e64, 0x15)
        /usr/local/go/src/runtime/panic.go:1116 +0x72 fp=0xc001c51730 sp=0xc001c51700 pc=0x4379b2
runtime.mapassign(0x2a3f180, 0xc0015773e0, 0xc001c518e8, 0x47)
        /usr/local/go/src/runtime/map.go:585 +0x5c5 fp=0xc001c517b0 sp=0xc001c51730 pc=0x410125
github.com/pachyderm/pachyderm/src/server/pkg/collection.(*stm).Put(0xc000bff110, 0xc0010da400, 0x3d, 0xc001bd8cd0, 0x47, 0x0, 0xc0008e8c30, 0x0, 0x0)
        /app/src/server/pkg/collection/transaction.go:209 +0x1f1 fp=0xc001c51b28 sp=0xc001c517b0 pc=0x2285291
github.com/pachyderm/pachyderm/src/server/pkg/collection.(*readWriteCollection).PutTTL(0xc0005c2a00, 0xc001bb87c0, 0x19, 0x3293f60, 0xc0008e8c30, 0x0, 0x47046b0, 0xc0005a89a0)
        /app/src/server/pkg/collection/collection.go:329 +0xb95 fp=0xc001c51ca8 sp=0xc001c51b28 pc=0x227e615
github.com/pachyderm/pachyderm/src/server/pkg/collection.(*readWriteCollection).Put(0xc0005c2a00, 0xc001bb87c0, 0x19, 0x3293f60, 0xc0008e8c30, 0xc00071e270, 0x23)
        /app/src/server/pkg/collection/collection.go:238 +0x5c fp=0xc001c51cf8 sp=0xc001c51ca8 pc=0x227d89c
[...]

PR #5488, opened 2020-12-02
Test flakes and reducing pool build verbosity

lukemarsden:
1. Mark some tests that are flaky as flaky
2. Bump the commit reference in testfaster to one which doesn't have `--debug` set for goreleaser, to make pool builds less noisy
3. A change I missed from review comments of https://github.com/pachyderm/pachyderm/pull/5480 - removing Lock/Unlock from the other stm commit()


Tryneus:
LGTM

PR #5490, opened 2020-12-03
Fix flaw in documentation

daniel-falk:
It seems like there are some errors in the documentation about datum and file relationships. The numbers in the text does not seem to correspond to the numbers in the images. As is now can be seen here:

https://docs.pachyderm.com/latest/concepts/pipeline-concepts/datum/relationship-between-datums/

PR #5492, opened 2020-12-03
Typos fix in datum doc - Back propagation of fix in PR #5490

nadegepepin:


echohack (approved):
👍 I think changes like this should fall under an OBVIOUS_FIX policy (that doesn't exist right now), so approved!

PR #5494, opened 2020-12-04
Auth extract/restore

actgardner:
Add support for extract/restore to the 1.12 admin API. Once this is merged we can backport it to 1.11 as well.
- [x] Add RPCs to ~extract the enterprise license~ (whoops, we already had this) and auth tokens for robot users.
- [x] Add operations to restore the enterprise license, activate auth, and restore ACLs and robot tokens.
- [x] Wire up the new endpoints in the extract and restore methods
- [x] Integration tests

Design decisions:
- On clusters where auth is enabled, Extract will generate a new root auth token and display it to the user. This prevents the case where a restored cluster has no admins because the IDP configuration is no longer valid.
- Only robot tokens are extracted. User tokens are regenerated by logging in, and pipeline tokens are regenerated when the pipelines are recreated.
- Group membership isn't extracted because it's synced from the IDP on login.
- ~Added a `pach auth login --token` flag which works like OTP but accepts a robot token, so users can auth to clusters wiht a backup token where the IDP integration has failed.~ (we already had this)
- Enabling auth and restoring ACLs is done last, after pipelines have already been created. Otherwise restoring a pipeline fails because the output repo already exists, which isn't allowed when auth is enabled.
- Added an argument to the auth Activate RPC where callers can provide a hashed token to restore instead of generating a random one.


Makefile
```diff
@@ -319,7 +319,7 @@ test-auth:
 	go test -v -count=1 ./src/server/auth/server/testing -timeout $(TIMEOUT) $(RUN)
 
 test-admin:
-	go test -v -count=1 ./src/server/admin/server -timeout $(TIMEOUT)
+	go test -v -count=1 ./src/server/admin/server -timeout $(TIMEOUT) $(RUN)
```
actgardner, line 322: Added the option to filter admin tests.

src/client/admin.go
```diff
@@ -21,8 +21,12 @@ func (c APIClient) InspectCluster() (*admin.ClusterInfo, error) {
 }
 
 // Extract all cluster state, call f with each operation.
-func (c APIClient) Extract(objects bool, f func(op *admin.Op) error) error {
-	extractClient, err := c.AdminAPIClient.Extract(c.Ctx(), &admin.ExtractRequest{NoObjects: !objects})
+func (c APIClient) Extract(objects, enterprise, auth bool, f func(op *admin.Op) error) error {
+	extractClient, err := c.AdminAPIClient.Extract(c.Ctx(), &admin.ExtractRequest{
```
actgardner, line 25: Added an option to not extract the enterprise token or auth token, as a safety valve. With these disabled extract/restore is the same as previous releases.

src/client/auth/auth.go
```diff
@@ -33,6 +33,9 @@ const (
 	// (with this prefix) is a logical PPS pipeline (even though the pipeline may
 	// not exist).
 	PipelinePrefix = "pipeline:"
+
+	// RootUser is the robot user created when auth is initialized.
+	RootUser = "robot:root"
```
actgardner, line 38: In 2.0 I want to make this the hard-coded default robot account that always provisions a token when auth is activated.

src/client/pkg/config/config.go
```diff
@@ -221,3 +221,19 @@ func (c *Config) Write() error {
 	value = proto.Clone(c).(*Config)
 	return nil
 }
+
+func WritePachTokenToConfig(token string) error {
```
actgardner, line 225: Refactored this out of the auth command-line in the config client package, so the auth test utils can configure an auth token for pachctl. This is useful for debugging when tests fail, and prevents clusters from getting stuck in an unusable state.

src/server/admin/cmds/cmds.go
```diff
@@ -32,20 +34,46 @@ $ {{alias}} -u s3://bucket/backup`,
...
+
+			// If the cluster has auth enabled, mint a robot token as a backup
+			// in case auth is misconfigured on restore.
+			authActive, err := c.IsAuthActive()
+			if err != nil {
+				return errors.Wrapf(grpcutil.ScrubGRPC(err), "error")
+			}
+
+			if authActive {
+				resp, err := c.GetAuthToken(c.Ctx(), &auth.GetAuthTokenRequest{
```
actgardner, line 46: Because of the way the API is structured, the only place to generate a new token and show the user the plaintext is in the CLI itself. Only the hashed robot token is stored in the extract.

src/server/admin/server/admin_test.go
```diff
@@ -192,22 +193,52 @@ func RepoInfoToName(repoInfo interface{}) interface{} {
...
 
-// testExtractRestore effectively implements both TestExtractRestoreObjects
-// TestExtractRestoreNoObjects, as their logic is mostly the same
-func testExtractRestore(t *testing.T, testObjects bool) {
-	if os.Getenv("RUN_BAD_TESTS") == "" {
-		t.Skip("Skipping because RUN_BAD_TESTS was empty")
-	}
+// testExtractRestore effectively implements TestExtractRestoreObjects,
+// TestExtractRestoreNoObjects, TestExtractRestoreAuth, as their logic is mostly the same
+func testExtractRestore(t *testing.T, testObjects, testAuth bool) {
```
actgardner, line 198: Add optional auth support to the existing extract/restore tests. This way we don't have to duplicate the logic to build repos, pipelines, etc.

src/server/admin/server/admin_test.go
```diff
@@ -192,22 +193,52 @@ func RepoInfoToName(repoInfo interface{}) interface{} {
...
 
+	// If we're testing auth, give all the pipeline users access to the first repo
+	// Also make some admin and FS admin users
+	if testAuth {
+		_, err := c.SetACL(c.Ctx(), &auth.SetACLRequest{
+			Repo: dataRepo,
+			Entries: []*auth.ACLEntry{
+				&auth.ACLEntry{Username: "github:pipeline-user-0", Scope: auth.Scope_READER},
+				&auth.ACLEntry{Username: "github:pipeline-user-1", Scope: auth.Scope_WRITER},
+				&auth.ACLEntry{Username: "github:pipeline-user-2", Scope: auth.Scope_OWNER},
```
actgardner, line 224: All these users don't _need_ access to the initial repo, but we can assert that this ACL is restored correctly later.

src/server/admin/server/api_server.go
```diff
@@ -257,6 +263,87 @@ func (a *apiServer) Extract(request *admin.ExtractRequest, extractServer admin.A
...
+		} else {
+			logrus.Warnf("Enterprise license state: %v, not extracting", state.State)
+		}
+	}
+
+	if !request.NoAuth {
+		config, err := pachClient.GetConfiguration(pachClient.Ctx(), &auth.GetConfigurationRequest{})
+		if err == nil {
+			// Activate authentication as `robot:extractrestore`. This token will be used by the Restore process for subsequent operations.
+			if err := writeOp(&admin.Op{Op1_12: &admin.Op1_12{ActivateAuth: &auth.ActivateRequest{Subject: "robot:extractrestore"}}}); err != nil {
```
actgardner, line 285: This has the slightly gross property that every time we extract/restore we end up creating another token in etcd and throwing it away. Because these have no TTL they could accumulate over time. Realistically doesn't seem like a huge concern, we could always filter tokens for this user out of extracts in the future, or delete the token at the end of the restore.

src/server/admin/server/api_server.go
```diff
@@ -257,6 +263,87 @@ func (a *apiServer) Extract(request *admin.ExtractRequest, extractServer admin.A
...
+
+			// Extract all robot tokens. This includes the backup root token created by pachctl, which can be used to recover the cluster if the IDP configuration is wrong.
+			robotTokens, err := pachClient.ExtractAuthTokens(pachClient.Ctx(), &auth.ExtractAuthTokensRequest{})
+			if err != nil {
+				return err
+			}
+
+			for _, t := range robotTokens.Tokens {
+				if err := writeOp(&admin.Op{Op1_12: &admin.Op1_12{RestoreAuthToken: &auth.RestoreAuthTokenRequest{Token: t}}}); err != nil {
+					if !auth.IsErrExpiredToken(err) {
```
actgardner, line 297: We have to do some gross mapping of TTLs to absolute timestamps, and it's possible for an extract to contain an expired robot token. In that case we just log and move on rather than blocking the restore.

src/server/auth/cmds/cmds.go
```diff
@@ -194,6 +179,14 @@ func LoginCmd() *cobra.Command {
 				resp, authErr = c.Authenticate(
 					c.Ctx(),
 					&auth.AuthenticateRequest{OneTimePassword: code})
+			} else if useToken {
```
actgardner, line 182: This is identical to how the OTP code works, but it just stuffs a token into the pachd config directly.

src/server/auth/server/api_server.go
```diff
@@ -2802,3 +2802,120 @@ func (a *apiServer) SetConfiguration(ctx context.Context, req *auth.SetConfigura
...
+
+		token := &auth.HashedAuthToken{
+			HashedToken: hash,
+			TokenInfo: &auth.TokenInfo{
+				Subject: val.Subject,
+				Source:  val.Source,
+			},
+		}
+		if ttl != -1 {
+			expiration, err := types.TimestampProto(time.Now().Add(time.Second * time.Duration(ttl)))
```
actgardner, line 2857: This feels weird, maybe we should put the absolute expiration timestamp in the TokenInfo in 2.0?

src/server/pkg/testutil/auth.go
```diff
@@ -155,6 +156,7 @@ func activateAuth(tb testing.TB) {
 	}
 	tokenMap[AdminUser] = resp.PachToken
 	ioutil.WriteFile(adminTokenFile, []byte(resp.PachToken), 0644)
+	config.WritePachTokenToConfig(resp.PachToken)
```
actgardner, line 159: As discussed earlier, when we enable auth on a test cluster this will also add the auth token to the pach context. This should reduce the risk of a cluster becoming unusable after a failed auth test.

src/server/admin/server/admin_test.go
```diff
@@ -192,33 +193,84 @@ func RepoInfoToName(repoInfo interface{}) interface{} {
...
 	if os.Getenv("RUN_BAD_TESTS") == "" {
 		t.Skip("Skipping because RUN_BAD_TESTS was empty")
 	}
 	if testing.Short() {
 		t.Skip("Skipping integration tests in short mode")
 	}
 
+	numPipelines := 3
 	c := tu.GetPachClient(t)
+	pipelineClients := make([]*client.APIClient, numPipelines)
```
actgardner, line 208: Initialize all the clients up front, because after much suffering I realized `GetAuthenticatedPachClient` resets the set of admins every time :|

src/server/admin/server/admin_test.go
```diff
@@ -192,33 +193,84 @@ func RepoInfoToName(repoInfo interface{}) interface{} {
...
+			Repo: dataRepo,
+			Entries: []*auth.ACLEntry{
+				&auth.ACLEntry{Username: "github:pipeline-user-0", Scope: auth.Scope_READER},
+				&auth.ACLEntry{Username: "github:pipeline-user-1", Scope: auth.Scope_WRITER},
+				&auth.ACLEntry{Username: "github:pipeline-user-2", Scope: auth.Scope_OWNER},
+			},
+		})
+		require.NoError(t, err)
+
+		_, err = c.ModifyClusterRoleBinding(c.Ctx(), &auth.ModifyClusterRoleBindingRequest{
```
actgardner, line 239: Test that cluster role bindings are restored.

src/server/admin/cmds/cmds.go
msteffen: I don't totally understand the big picture here. It looks like the server-side `Extract()` implementation generates an `Activate` message if auth is active, which I'd expect to create the first admin in the target cluster during `Restore()`, so I don't totally understand what the user is supposed to do with this token after we print it? Is `resp` used for other stuff in this function and I just missed it?

src/server/admin/server/admin_test.go
```diff
@@ -230,6 +282,19 @@ func testExtractRestore(t *testing.T, testObjects bool) {
...
 			false,
 		))
+
+		if testAuth {
+			_, err := c.SetACL(c.Ctx(), &auth.SetACLRequest{
+				Repo: pipeline,
+				Entries: []*auth.ACLEntry{
+					&auth.ACLEntry{Username: fmt.Sprintf("github:pipeline-user-%d", i+1), Scope: auth.Scope_READER},
+					&auth.ACLEntry{Username: fmt.Sprintf("github:pipeline-user-%d", i), Scope: auth.Scope_OWNER},
+					&auth.ACLEntry{Username: fmt.Sprintf("pipeline:%v", pipeline), Scope: auth.Scope_WRITER},
```
msteffen, line 292: Does it break the test to remove this entry? I think `CreatePipeline` is supposed to add the pipeline user—if it's not that may be a bug.

src/server/admin/server/api_server.go
```diff
@@ -257,6 +263,88 @@ func (a *apiServer) Extract(request *admin.ExtractRequest, extractServer admin.A
...
+		if state.State == enterprise.State_ACTIVE {
+			if err := writeOp(&admin.Op{Op1_12: &admin.Op1_12{ActivateEnterprise: &enterprise.ActivateRequest{ActivationCode: state.ActivationCode}}}); err != nil {
+				return err
+			}
+		} else {
+			logrus.Warnf("Enterprise license state: %v, not extracting", state.State)
+		}
+	}
+
+	// Once all the repos and pipelines have been restored, then try to enable auth. Our auth checks block creating a pipeline if the output repo already exists,
```
msteffen, line 281: I remember Yusuf talking about this with build pipelines, and while I don't think there's nothing unique to auth that requires  this check (so we could get rid of it without breaking anything new, I think), it actually strikes me as a bug that we don't check this in all cases.

Creating a pipeline when an existing output repo exists seems broken even without auth. What if you create two pipelines with the same name and they're writing on top of each other? This is maybe a JD question.

src/server/admin/server/api_server.go
msteffen: I'm still trying to understand the mechanics here a bit, but IIUC this creates the token during `Restore()`, and my expectation is that a user would only call restore once (otherwise I think they're creating repos and pipelines on top of existing copies of the same repos and pipelines, which I think errors). Basically, I agree it's not a huge deal if we do leak tokens, and I actually think even that's not a risk :+1:

src/server/auth/server/api_server.go
msteffen: I originall wrote this for the comment above, but moved it here since this seems like the more relevant place.

In you view, is the grossness due to skew (i.e. the TTL you get may not be the TTL anymore after getting the current time and adding the two, etc)? There's a bit of a limitation that comes from etcd's lease API dealing in durations, but we could always work around that by checking the token's expiration and disregarding it if it's in the past, even if the etcd lease hasn't expired yet.

src/server/pkg/testutil/auth.go
msteffen: :+1:

msteffen (commented):
First pass, let me know what you think

src/server/admin/server/api_server.go
actgardner: This token is a robot token so it's extracted and restored every time the user upgrades. So if they do an extract/restore for 2.0, then 2.1, etc. each of those would call Activate again and create a new robot token which would be included in subsequent extracts.

Like I said, not the end of the world, and we could not extract these in the future as well.

src/server/auth/cmds/cmds.go
actgardner: Yup, I just didn't know we had a use-auth-token. I'll remove this for now.

src/server/auth/server/api_server.go
actgardner: Yup, the introduction of skew feels weird. I don't think we need to address it here,  but like you said in the future we could do a belt-and-suspenders by setting a TTL and also checking a timestamp encoded into the token info.

AFAIK this PR should address any issues with trying to restore expired tokens.

src/server/admin/server/admin_test.go
actgardner: When the pipeline is created the pipeline user is included in the ACL, but we need to include it here because `SetACL` doesn't just append entries to the ACL, it replaces the whole thing.

src/server/admin/cmds/cmds.go
actgardner: `Activate` creates the first admin in the restored cluster, but how do we plumb that information (the plaintext for the first auth token) back to the user doing the restore? The `Restore` RPC returns `Empty`, which seems like an oversight. We could change that to a `RestoreResponse`, and try and put the auth token in there - not sure if this counts as a breaking API change.

My other concern with putting the only known-good auth token in the RPC response to `Restore` is that, if something later in the restore blows up, that response won't be sent to the user, and the cluster may be totally useless. If we generate a token for the user as part of the Extract process, as long as Activate and restoring the auth tokens succeeds, the user has all their data, pipelines and an auth identity they could use to manually clean up the mess.

A third option I considered was having `Activate` accept a hashed auth token to use for the initial admin. In this scenario the hashed root token would be baked into the extract, and the user restoring the extract would supply the plaintext for the token in the auth context from pachctl. This has the nice property that it doesn't generate any new tokens during the restore process.

src/server/admin/server/api_server.go
actgardner: > What if you create two pipelines with the same name and they're writing on top of each other?

Things like the PipelineInfo collection seem to be keyed on the pipeline name, so I don't believe it's possible to have two pipelines with the same name.

actgardner:
@msteffen I think I've replied to everything - I'm going to remove the `pachctl auth login -t` command for now. LMK if you want to hop on a call tomorrow to talk through the lifecycle of the token we generate during the extract process.

src/server/admin/server/admin_test.go
msteffen: Oh, right. I'm now realizing that this seems like kind of a shortcoming in the auth design, actually, in that it's a pretty easy way to shoot yourself in the foot (maybe we should enforce a R+M+W workflow by including a version token in SetACL/GetACL, and if you call SetACL without a version token you get rejected? Or require you to set the version token to -1, so at least you know you're YOLOing. I think we do something similar with the auth config, to prevent simultaneous writes).

For a test it doesn't really matter, but one alternative could be to use `SetScope`, (i.e. `SetScope("github:...", READER); SetScope("github:...", OWNER)`).

src/server/admin/server/api_server.go
msteffen: oooh, got it. Thanks!

src/server/admin/cmds/cmds.go
msteffen: Ah, now I see the issue, thanks! One question: above, you mentioned that you wanted to make `RootUser` the hard-coded default robot account in 2.0 (meaning they're always a super admin?). Will pach 1.12 will work that way? I ask because my understanding of the mechanism here is: any token for `RootUser` is an escape hatch, and this gives users an escape hatch by getting a new `RootUser` token and just printing it, and I want to make sure that's right. If that's how this works, does `Activate()` guarantee that `RootUser` is a cluster admin in 1.12? This approach broadly makes sense to me, and I'm on board with it, though I outlined a few alternatives that occurred to me below as well, in case they're appealing.

- One high-level option that occurs to me is that we could make `Restore()` a streaming API (or add a new `RestoreAuthCompatible()` API that is streaming, as an alternative to the existing API, per semver. If someone calls the old API and it encounters an `Activate` op, or we use the no-op auth marker described below and the old API encounters that, it could simply error). This would let `Restore()` generate a token if it encounters the no-op auth marker at the beginning and send it back right away, but then keep going with the restore. A benefit of this approach is that it avoids user-generated tokens (as described below).

- I do also like the idea of setting up auth using a token the user already has during `Restore()`. Is the idea that the user would send the hash, and we'd call `Activate(hash) -> tokens.Put(hash, 'robot:extractrestore')` and then auth is activated and the user's token authenticates them as `robot:extractrestore`?
  - This is a small modification, but if that understanding is right, it seems like one risk with using the user's current token during `Restore()` is that if we activate auth and place `(user_token_hash, robot:extractrestore)` in the tokens collection, and then later call `RestoreToken(user_token_hash, github:alice)` and `alice` isn't an admin (e.g. different user restoring than extracting) then the user might be downgraded from admin to non-admin partway through the restore, which might mess up subsequent restore ops. One way we could work around that is by generating a token client-side before calling `Restore()` and printing _that_ out at the start of `Restore()`. Restore would use the client-generated temporary token as the initial admin token (`robot:extractrestore`), and use that for everything. Another benefit of that approach is that when `Restore()` puts `(client_generated_token_hash, robot:extractrestore)` in the tokens collection, we could put e.g. a 30-day TTL on it, which prevents the token-leaking issue. Alternatively, `Restore()` could put `(user_generated_token_hash, RootUser)` in the tokens collection with no lease?
  - A closely-related corner case is that if the user doesn't have an auth token in their config at all, but you encounter an `Activate()` request partway through a big restore, then you have to send the user an error after already restoring a bunch of stuff. But one solution that occurs to me is to have the first message in the Restore dump be a no-op marker that just indicates "auth will be activated later, watch out," and if `Restore()` sees that but the user didn't send `Restore()` a token, we can error right away (also, encountering an `Activate()` request if there was no auth marker at the beginning would be an error, but that would be a bug in `Extract()` rather than user error).
  - One final idea: on an API level, as far as activating auth with a user-provided token hash, we could lean on `Activate()`, or we could modify `RestoreToken()` to activate auth if it's not already activated. The internal implementation could be roughly the same—Activate and RestoreToken could use a shared helper—but it would restrict the use of client-generated tokens to a less-used API.

src/server/admin/server/api_server.go
msteffen: That makes sense. FWIW this check is in `authorizePipelineOpInTransaction`. All we have to do is disable it is change the check for `Create` so that it require either a) the output repo doesn't exist or b) the user has WRITER access to it.

src/client/admin/admin.proto
```diff
@@ -17,6 +17,15 @@ import "client/admin/v1_11/pfs/pfs.proto";
 import "client/admin/v1_11/pps/pps.proto";
 import "client/pfs/pfs.proto";
 import "client/pps/pps.proto";
+import "client/auth/auth.proto";
+import "client/enterprise/enterprise.proto";
+
+// CheckAuthToken is the first message in an extract on a cluster
+// where auth is enabled. It verifies that the user provided the correct
+// auth token for the restore before we start restoring. 
+message CheckAuthToken {
```
actgardner, line 26: Because we restore auth relatively late in the restore process, the first Op on a cluster with auth is this placeholder that verifies the auth token in the user's context matches the hash used for Activate later on.

src/client/auth/auth.go
```diff
@@ -213,3 +231,25 @@ func IsErrTooShortTTL(err error) bool {
...
+	if !ok {
+		return "", ErrNoMetadata
+	}
+	if len(md[ContextTokenKey]) > 1 {
+		return "", errors.Errorf("multiple authentication token keys found in context")
+	} else if len(md[ContextTokenKey]) == 0 {
+		return "", ErrNotSignedIn
+	}
+	return md[ContextTokenKey][0], nil
+}
```
actgardner, line 255: Refactor these out of the auth server into the auth client package. They're useful in tests as well as the admin logic.

src/client/auth/auth.proto
```diff
@@ -39,14 +39,23 @@ message ActivateRequest {
   // - If set to a robot user (i.e. it has a 'robot:' prefix), then Pachyderm
   //   will generate a new token for the robot user; this token will be the only
   //   way to administer this cluster until more admins are added.
+  // 
+  //   Deprecated: in 2.0 Activate will always provision a fixed robot user.
```
actgardner, line 43: Add some deprecation notices for fields I want to remove from Activate in 2.0.

src/server/admin/cmds/cmds.go
actgardner: OK, I've taken another swing at this:

- Activate now accepts an optional token hash to use as the root token. I realized I want this for integration tests anyways, so we don't have to deal with randomly generated tokens. 

- If auth is activated on the cluster, the client generates a token and includes it in the Extract request.

- Extract includes a `CheckAuthToken` op in the extract file which verifies the auth token during Restore.

- Once Extract has succeeded the client spits out the token on stdout.

- The user must set up their context for Restore by activating the token ahead of time.

- The user calls Restore and the auth token is included in the GRPC context.

- Server-side the Restore method immediately sees the `CheckAuthToken` op and verifies that the token in the context matches the hashed token in the extract.

- If the tokens match, the server will call Activate with the hashed token. This atomically enables auth, sets the root token and makes RootUser an admin. This way the cluster is never in an unrecoverable state.

> Ah, now I see the issue, thanks! One question: above, you mentioned that you wanted to make RootUser the hard-coded default robot account in 2.0 (meaning they're always a super admin?). Will pach 1.12 will work that way? 

I think this is too much of a breaking change for a minor release. In 2.0 I want to take away the ability to remove RootUser from the set of admins, and always use that account for Activate.

> If that's how this works, does Activate() guarantee that RootUser is a cluster admin in 1.12?

I realized the previous implementation didn't guarantee this. The new one does. There's nothing that happens after Activate that removes admins, Restore only adds new ones.

> One high-level option that occurs to me is that we could make Restore() a streaming API (or add a new RestoreAuthCompatible() API that is streaming, as an alternative to the existing API, per semver.

In my experience bidirectional streaming grpc APIs are kind of gross to work with. I'm hoping we can come up with a solution where we don't need to stream data back to the client mid-Restore.

> One way we could work around that is by generating a token client-side before calling Restore() and printing that out at the start of Restore(). Restore would use the client-generated temporary token as the initial admin token (robot:extractrestore), and use that for everything.

The upside of generating the token in Restore is there's no risk a user will lose their token (and consequently their ability to restore a backup). The downside is that Restore doesn't really know whether or not a given extract includes auth until after it's been extracted. We would either print the token immediately when Restore is called (even if auth is not enabled), print it when we get some streaming message (potential timing issues?) or print it after the restore is successful (half-restored clusters can be in a bad state).

> Another benefit of that approach is that when Restore() puts (client_generated_token_hash, robot:extractrestore) in the tokens collection, we could put e.g. a 30-day TTL on it, which prevents the token-leaking issue. Alternatively, Restore() could put (user_generated_token_hash, RootUser) in the tokens collection with no lease?

I would prefer not to put a TTL on the only admin token we know a user has, as it would basically make the cluster a time bomb. I guess they could always restore again, but if they did anything in the meantime they could end up in a bad state. In the future if we're concerned about the token leaking issue we can avoid restoring any other RootUser tokens _except_ the one printed by Extract.

msteffen (approved):
A few small questions, but nothing blocking. Overall LGTM.

Thanks again for making this happen!

src/client/auth/auth.proto
```diff
@@ -47,6 +47,11 @@ message ActivateRequest {
   // is that string (unless this "looks like" a GitHub access code, in which
   // case Pachyderm does retrieve the corresponding GitHub username)
   string github_token = 1 [(gogoproto.customname) = "GitHubToken"];
+
+  // If set, activate auth for the root user (`robot:root`), and add this
```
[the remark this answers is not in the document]
actgardner: Yup, on reviewing this I agree that hashing client-side didn't confer any benefits. I've updated this to accept a plaintext token to use.

src/server/auth/server/testing/admin_test.go
```diff
@@ -2174,38 +2230,3 @@ func TestFSAdminOneTimePasswordOtherUser(t *testing.T) {
 	require.YesError(t, err)
 	require.Matches(t, "must be an admin", err.Error())
 }
-
-// TestInitialRobotUserGetOTP tests that Pachyderm can be activated with a robot
-// user, which will have no session expiration, and that when this user gets an
-// OTP, the token associated with their OTP will expire
-func TestInitialRobotUserGetOTP(t *testing.T) {
```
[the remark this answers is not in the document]
actgardner: I had to rework it a bit to activate auth as a robot, since `pach:root` can't get new tokens. It's back now.

actgardner:
@msteffen I hate to do this to you, but when I was making the change to have Activate accept a plaintext token I realized I might as well go all the way and have that token be generated in Restore instead of Extract. 99% of the code is the same, but instead of baking the token into the extract data we generate a token in the client during Restore, and that token is used for Activate. To make sure the client doesn't lose the token if there's an error we set the token in the active pachctl context _and_ print it on the command-line.

I know you've spent a bunch of time on this, the only new stuff you need to worry about is this block: https://github.com/pachyderm/pachyderm/pull/5494/files#diff-911fb402bb5f7c2cbd739ee7d2d3477d84e24ea799c4414eaf9bde455f3b79b8R73-R98

src/client/auth/auth.go
```diff
@@ -213,3 +236,25 @@ func IsErrTooShortTTL(err error) bool {
...
 		strings.Contains(errMsg, ")")
 }
+
+// HashToken converts a token to a cryptographic hash.
+// We don't want to store tokens verbatim in the database, as then whoever
+// that has access to the database has access to all tokens.
+func HashToken(token string) string {
+	sum := sha256.Sum256([]byte(token))
+	return fmt.Sprintf("%x", sum)
+}
```
msteffen, line 246: Since `Activate()` takes a root token instead of a hash now, do you think it would make sense to keep this in the server? I guess I worry a little that it'll be harder to change if it's in the client package (though if it's not in our API, maybe not).

src/server/admin/cmds/cmds.go
```diff
@@ -63,6 +69,33 @@ $ {{alias}} -u s3://bucket/backup`,
 			if err != nil {
 				return err
 			}
+
+			// Generate a root auth token and cache it in the current context. If the restore operation
+			// activates auth this will become the new root token.
+			if !noToken {
+				authToken := uuid.NewWithoutDashes()
```
msteffen, line 76: One arguable benefit of this approach that occurred to me is that if you restore the same dump twice into two different clusters, they'll have different root tokens (which will strengthen the invariant that all clusters have a unique root token)

msteffen (approved):
LGTM! (with one totally optional suggestion)

src/server/admin/cmds/cmds.go
actgardner: Yeah I had this thought as well. Sharing tokens between clusters and the risk of losing a token between extracting and restoring seem to outweigh the danger of a partial restore.

PR #5495, opened 2020-12-04
Add some load tests using the k6 framework

jrockway:
I have been playing with k6 on other projects and wanted to try out the gRPC support.  It works!  This change adds some tests that call pfs.PutFile and pfs.GetFile on some small files.  (k6 doesn't support streaming very well, it seems.)

To "productionize" this, we should run a few seconds worth of tests with each CI run (`k6 --vus=10, --duration=5s`), and send the performance results somewhere, so we can look at per-PR performance changes.  It would also be nice to collect go CPU profiles during the tests, so that you can see where the server is spending its time.

We should also consider doing some overnight tests, to ensure that memory usage stays under control even with heavy workloads.

Of course, it would also be good to test something other than PFS.  That was the first thing that popped into my mind when I was thinking about this this morning... but obviously we do more than getting and putting files :)

cc: @lukemarsden for advice on continuously running these, @echohack because you were investigating performance problems recently


echohack:
I love k6! Something like k6 gets really powerful when you combine it with a performance budget and fail the build when you're over budget. (We'll need to block on merge for failed CI, but that's another topic)

What I've looked into so far is very much dipping my toe into the water from an integration, user-driven perspective using `pachctl` and `BATS`. These tests really center around `pachctl delete pipeline` and the new `--split-txn` flag. Here's a few results from those tests:

```
| Test data | gRPC transaction size | overwrite? | Setup Time |
-------------------------
| 4000 commits and 10 chained branches | 4202715 bytes | --overwrite | 50m |
| 1000 commits and 12 chained branches | 1814377 bytes | --overwrite | 24m10s |
| 1000 commits and 11 chained branches | 1801413 bytes | --overwrite | 22m31s |
| 2000 commits and 04 chained branches | 2023879 bytes | --overwrite | 22m9s |
| 1750 commits and 03 chained branches | 1596061 bytes | --overwrite | 12m42s |
```

PR #5497, opened 2020-12-04
Don't use bash double-bracket conditional in the Makefile because it doesn't work in CI

Tryneus:
CI logged this error:
```
Dec 04 18:12:22 /bin/sh: 1: [[: not found
```

This appears to be because the shell used by `make` doesn't support double-bracket conditional syntax, so change it to use a different command.  The error itself is more noise than anything else, as the conditional is only there to prevent accidentally deploying pachyderm to the wrong kubernetes cluster which has no reason to even be configured on a CI machine, and it doesn't appear to be blocking anything.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,7 +150,7 @@ check-kubectl:
 		echo "error: kubectl not found"; \
 		exit 1; \
 	}
-	@if [[ $(shell kubectl config current-context) == *gke_pachub* ]]; then \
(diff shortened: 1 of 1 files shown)


lukemarsden (approved):

lukemarsden:
Lgtm, thanks!

lukemarsden:
Please merge when tests pass

PR #5502, opened 2020-12-07
Join Examples

JoeyZwicker:


examples/joins/README.md
```diff
@@ -1,193 +1,231 @@
...
+Have a quick look at your repositories: 
+```shell
+$ pachctl list file stores@master
+$ pachctl list file purchases@master
+$ pachctl list file returns@master
+```
+You should see the following files:
+- Stores:
+
+![stores_repository](./img/pachctl_list_file_stores_master.png)
```
JoeyZwicker, line 106: do we prefer this (and below) as screenshots of the terminal versus just text block that shows the same info? I think they both work -- the only downside to images is they slowly bloat the repo more over time, are a little harder to update, and may not work as well on mobile or slower loading envs.

examples/joins/README.md
```diff
@@ -1,193 +1,231 @@
...
+![returns_repository](./img/pachctl_list_file_returns_master.png)
+
+## 4. Example 1 - An Inner-Join pipeline creation 
+***Goal***
+List all purchases by zip code. 
+>![pach_logo](./img/pach_logo.svg) Remember that we will only see the zip code of the stores where purchases actually happened. That is what inner joins do.
+
+1. **Pipeline input repositories**: `stores` and `purchases` - Inner join by STOREID.
+2. **Pipeline**: Executes a python code reading the `zipcode` in the matching STOREIDx.txt file and appending the matched purchase file's content to a text file named after the zip code.
+3. **Pipeline output repository**: `inner_join`- list of text files named after the stores' zip codes in which purchases have been made and containing the list of said purchases.
```
JoeyZwicker, line 123: I love this structure and summary view of what's going on. very informative.

examples/joins/README.md
```diff
@@ -1,193 +1,231 @@
...
+3. **Pipeline output repository**: `inner_join`- list of text files named after the stores' zip codes in which purchases have been made and containing the list of said purchases. 
+
+In the diagram below, we have mapped out the data of our example. We will expect the 2 following files in the output repository of our `inner_join` pipeline: 
+```shell
+    └── 02108.txt
+    └── 90210.txt
+```
+Each should contain 3 purchases.
+
+![inner_join](./img/inner_join.png)
```
JoeyZwicker, line 132: typo in the image. red box should say 90210 but it has 09210.

examples/joins/README.md
```diff
@@ -1,193 +1,231 @@
...
+
+***Step 3*** - Now let's create your pipeline. 
+
+Because unprocessed data are awaiting in your entry repositories, the pipeline creation will automatically trigger a job.
+In the `examples/joins` directory, run:
+```shell
+$ make inner-join
```
JoeyZwicker, line 139: IMO, having a make target when its one line is a little weird. It's fine to leave in the makefile, when it's only one line, I think it's better for the tutorial to show the user the pachctl cmd (they'll learn more) rather than the make target as the primary/preferred cmd to run. The make is better when we're scripting a bunch of setup.

examples/joins/README.md
```diff
@@ -1,193 +1,231 @@
...
+```
+You should see our 2 expected text files. 
+
+![output_repository](./img/pachctl_list_file_inner_join_master.png)
+
+Now for a visual confirmation of their content:
+```shell
+$ pachctl get file inner_join@master:/02108.txt
+```
+>![pach_logo](./img/pach_logo.svg) Want to take this example to the next level? Practice using joins AND [groups](Add link to group). You can create a 2 steps pipeline that will group Returns and Purchases by storeID then join the output repo with Stores to agregate by location.
```
JoeyZwicker, line 185: love it!

examples/joins/README.md
```diff
@@ -1,193 +1,231 @@
...
+
+1. **Pipeline input repositories**: `stores` and `returns` - Outer join by STOREID: 3 scenarios
+2. **Pipeline**: Executes a python code reading the `zipcode` (if any...) in the matching STOREIDx.txt file and appending the matched return file's content (if any...) to a text file named after the zip code. 
+3. **Pipeline output repository**: `outer_join`- list of text files named after the stores' zip codes. What list will be produced depends on where the outer join's focus was put on. We will get into the detail of each case to help you figure out the best scenario for you.
+
+In the diagram below, we have mapped out the data of our example and the expected match in each of the 3 cases. 
+>![pach_logo](./img/pach_logo.svg) We have created an edge case here, with a return made in a store (Store 0) that is not in our list and therefore has an unknow zipcode. This should help highlight how outer join's matches are broader that inner joins. In real life, see this as one of your products being returned to a store that does not belong to your geographic unit... Again, this is an over simplified/over stretched example.
```
JoeyZwicker, line 202:
```suggestion
>![pach_logo](./img/pach_logo.svg) We have created an edge case here, with a return made in a store (Store 0) that is not in our list and therefore has an unknown zipcode. This should help highlight how outer join's matches are broader that inner joins. 
```

JoeyZwicker: minor typo and not needed to over qualify that it's a stretched example.

examples/joins/README.md
```diff
@@ -1,193 +1,231 @@
...
+- the zip of **all stores** (listed in our repo or not) **where at least one return was made**
+- all **zip of stores listed** in our repo, **with or without returns**
+- the zip of **all stores** listed or unlisted in our repo, **with or without returns**
+
+The end result will vary depending on where we will focus the outer join. Although the behaviour might slightly differ from SQL because, again, we are working at the file path level, think about a left outer, right outer, and full outer.
+
+1. **Pipeline input repositories**: `stores` and `returns` - Outer join by STOREID: 3 scenarios
+2. **Pipeline**: Executes a python code reading the `zipcode` (if any...) in the matching STOREIDx.txt file and appending the matched return file's content (if any...) to a text file named after the zip code. 
+3. **Pipeline output repository**: `outer_join`- list of text files named after the stores' zip codes. What list will be produced depends on where the outer join's focus was put on. We will get into the detail of each case to help you figure out the best scenario for you.
+
```
JoeyZwicker, line 200: I feel like there's something missing before you jump into the diagram. Specifically, I think we should be showing the pipeline spec input, and telling the user that "outer is a boolean field that can be set independently on each input repo. Setting outer: true means that files in that repo, even if they don't match, will still be processed as a datum."

Also, I'm not sure we need to cover all 3 out join scenarios at once. One idea might be to have the default be outer: true on both repos (like it is) and have them run just that and show the diagram. Then, have Next Steps be:

"experiment with only having outer set on one repo or the other. This can mirror real world situations such as...<as you describe above>. We've also drawn the diagrams for you to help visualize what is happening in each of the scenarios where only one of the inputs has `outer:true` set."

What do you think?

examples/joins/README.md
nadegepepin: You are 100% right. I restructured the makefile to make sure that anyone could run both examples in any order without running into a repo that has already been created and ended up with this one line. Makes no sense. Removing it. Thx for the feedback.

examples/joins/README.md
```diff
@@ -1,193 +1,231 @@
...
+***Pre-requisite***
+- A workspace on [Pachyderm Hub](https://docs.pachyderm.com/latest/pachhub/pachhub_getting_started/) (recommended) or Pachyderm running [locally](https://docs.pachyderm.com/latest/getting_started/local_installation/).
+- [pachctl command-line ](https://docs.pachyderm.com/latest/getting_started/local_installation/#install-pachctl) installed, and your context created (i.e., you are logged in)
+
+***Getting started***
+- Clone this repo.
+- Make sure Pachyderm is running. You should be able to connect to your Pachyderm cluster via the `pachctl` CLI. 
+Run a quick:
+```shell
+$ pachctl version
```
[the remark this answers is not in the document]
nadegepepin: Will add it in the readme for now as it will likely help users. Ultimately, will create an issue to keep this idea of a warning to the user in our pipeline.

examples/joins/README.md
nadegepepin: ? Not sure to understand that one. The red box does say 90120.txt

examples/joins/README.md
nadegepepin: I like the idea of the other outer-join options presented in "Want to take this example to the next level?" with the expected results listed in each case. Will add the pipeline spec for more clarity. I initially thought that it belongs to the documentation itself but there is no harm in adding it here. Some people would rather learn by example before taking a look at the doc.

examples/joins/README.md
```diff
@@ -1,19 +1,19 @@
 # Inner and Outer Join Pipelines 
 >![pach_logo](./img/pach_logo.svg) The outer join functionality is available in version **1.12 and higher**.
 - Our first example will walk you through a typical inner join case. In a similar way to SQL, "inner-join" pipelines **only** run your code on the files, in your joined repositories, that match a specific naming pattern  (i.e., ***match your glob pattern/capture groups***). In Pachyderm's terms, inner joins will only create a datum (see Key concepts below) if there is a match in all join repos.
-- Our second example will showcase 3 variations of "outer-join" pipelines and outline how they differ from the first. In short, outer joins specify that the input repo with `"outer_join": true` set will still see a datum even if it does not have a match.
+- Our second example will showcase 3 variations of "outer-join" pipelines and outline how they differ from the first. In short, outer joins specify that the input repo with `"outer_join": true` set in the pipeline specifications will still see a datum even if it does not have a match.
```
JoeyZwicker, line 4:
```suggestion
- Our second example will showcase 3 variations of "outer-join" pipelines and outline how they differ from the first. In short, outer joins specify that for an input repo with `"outer_join": true` set in the pipeline specifications, datums will still be created even if it does not have a match.
```

JoeyZwicker: This is admittedly really hard to describe in normal english --  struggled with it a ton too.

JoeyZwicker (commented):
Did a quick readthrough of the changes since last review and LGTM. I can't approve the PR because it was technically created by me. :P

Overall, seems good.

mindthevirt (commented):
LGTM

PR #5506, opened 2020-12-07
Build kafka-go modules to popule go.mod/sum

msteffen:


lukemarsden (approved):

msteffen:
Merging without waiting for CI, as this relates to a bug fix and doesn't change any code (and also CI is just really unreliable atm, so waiting for CI would delay this substantially)

PR #5509, opened 2020-12-07
Cluster Migrations

brendoncarroll:
Adds migrations package for coordinating changes to cluster state.  (Most importantly, right now, changes to postgres)

### DIFF
--- a/src/server/pkg/migrations/migrations.go
+++ b/src/server/pkg/migrations/migrations.go
@@ -90,14 +90,14 @@ func ApplyMigrations(ctx context.Context, db *sqlx.DB, baseEnv Env, state State)
 		if _, err := tx.ExecContext(ctx, `INSERT INTO migrations (id, start_time) VALUES ($1, CURRENT_TIMESTAMP)`, state.n); err != nil {
 			return err
 		}
-		logrus.Info("applying migration", state.n)
+		logrus.Infof("applying migration %d", state.n)
 		if err := state.change(ctx, env); err != nil {
 			return err
 		}
 		if _, err := tx.ExecContext(ctx, `UPDATE migrations SET end_time = CURRENT_TIMESTAMP WHERE id = $1`, state.n); err != nil {
 			return err
 		}
-		logrus.Info("successfully applied migration", state.n)
+		logrus.Infof("successfully applied migration %d", state.n)
 		return nil
 	}(); err != nil {
 		if err := tx.Rollback(); err != nil {
--- a/src/server/pkg/migrations/migrations.go
+++ b/src/server/pkg/migrations/migrations.go
@@ -12,20 +12,27 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// Env contains all the objects that can be manipulated during a migration.
+// The Tx field will be overwritten with the transaction that the migration should be performed in.
 type Env struct {
 	// TODO: etcd
 	ObjectClient obj.Client
 	Tx           *sqlx.Tx
 }
 
+// MakeEnv returns a new Env
+// The only advantage to this contructor is you can be sure all the fields are set.
+// You can also create an Env using a struct literal.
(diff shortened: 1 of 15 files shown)


src/server/pkg/storage/chunk/metadata.go
```diff
@@ -103,17 +105,21 @@ func (s *postgresStore) Delete(ctx context.Context, chunkID ID) error {
...
-func SetupPostgresStore(db *sqlx.DB) {
-	db.MustExec(schema)
-}
-
-const schema = `
-	CREATE SCHEMA IF NOT EXISTS storage;
-
-	CREATE TABLE storage.chunks (
+// SetupPostgresStoreV0 sets up tables in db
+func SetupPostgresStoreV0(ctx context.Context, tableName string, tx *sqlx.Tx) error {
```
brendoncarroll, line 109: This is how I was thinking of parameterize the table locations.  But it doesn't seem like we need this right now and I think it harms the readability of the SQL so this is more just to get feedback on this method. We should probably pull it out before merging.

src/server/pkg/storage/chunk/util.go
```diff
@@ -19,7 +21,11 @@ func NewTestStorage(t testing.TB, db *sqlx.DB, tr track.Tracker, opts ...Storage
 
 // NewTestStore creates a store for testing.
 func NewTestStore(t testing.TB, db *sqlx.DB) MetadataStore {
-	db.MustExec(schema)
+	ctx := context.Background()
+	tx := db.MustBegin()
+	tx.MustExec(`CREATE SCHEMA IF NOT EXISTS STORAGE`)
```
brendoncarroll, line 26: I want to avoid the conditional creates in the production setup SQL, so this has to be added to the test methods

src/server/pkg/migrations/migrations.go
```diff
@@ -0,0 +1,172 @@
...
+			);
+			INSERT INTO migrations (id, name, start_time, end_time) VALUES (0, 'init', CURRENT_TIMESTAMP, CURRENT_TIMESTAMP) ON CONFLICT DO NOTHING;
+			`)
+			return err
+		},
+	}
+}
+
+func ApplyMigrations(ctx context.Context, db *sqlx.DB, baseEnv Env, state State) error {
+	if state.prev != nil {
```
brycemcanally, line 71: We discussed this offline, add a wrapper to convert the linked list to a slice that we iterate through in the code below. This would improve the readability.

src/server/pkg/storage/chunk/metadata.go
brycemcanally: Discussed offline, not parameterizing table names for now.

src/server/pkg/storage/chunk/util.go
```diff
@@ -19,7 +21,11 @@ func NewTestStorage(t testing.TB, db *sqlx.DB, tr track.Tracker, opts ...Storage
...
+	SetupPostgresStoreV0(ctx, "storage.chunks", tx)
```
brycemcanally, line 27: Check error.

src/server/pkg/storage/fileset/postgres_store.go
```diff
@@ -100,12 +99,17 @@ const schema = `
...
+	return err
 }
 
 // NewTestStore returns a Store scoped to the lifetime of the test.
 func NewTestStore(t testing.TB, db *sqlx.DB) Store {
-	SetupPostgresStore(db)
+	ctx := context.Background()
+	tx := db.MustBegin()
+	tx.MustExec(`CREATE SCHEMA IF NOT EXISTS storage`)
+	SetupPostgresStoreV0(ctx, tx)
```
brycemcanally, line 112: Check error.

brycemcanally (approved):
lgtm

PR #5510, opened 2020-12-08
Update transaction documentation for pipelines.

PFedak:


nadegepepin (changes requested):
Such a micro edit.
you --may--? need to change multiple pipelines.

nadegepepin (commented):
Could you please provide an example? It is not super clear if you are not familiar with transactions.

PR #5511, opened 2020-12-08
Revert golden manifests on master back to 1.12

actgardner:
Tests are failing on master because the AdditionalVersion isn't set to -rc1 in typical master builds. This was a mistake committed when releasing rc1.


nitinjainsj (approved):
LGTM

PR #5516, opened 2020-12-10
Document how to enable groups support in OIDC [ci skip]

actgardner:
Update the docs to mention the `additional_scopes` field, where users can request the `groups` scope to sync group membership.


nadegepepin (approved):
LGTM

PR #5522, opened 2020-12-11
Deflake dex tests

actgardner:
Install dex on the TestFaster k8s clusters when we're running an auth test bucket.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -384,7 +384,7 @@ clean-launch-loki:
 	helm uninstall loki
 
 launch-dex:
-	helm repo add --force-update stable https://charts.helm.sh/stable
+	helm repo add stable https://charts.helm.sh/stable
 	helm repo update
 	helm upgrade --install dex stable/dex -f etc/testing/auth/dex.yaml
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -384,7 +384,7 @@ clean-launch-loki:
 	helm uninstall loki
 
 launch-dex:
-	helm repo add --force-update stable https://charts.helm.sh/stable
```
actgardner, line 387: I'm not sure what the difference is between the version of helm in testfaster and the version in Travis - I think we're OK to drop this for now. Once the identity server alpha lands these will use the identity server instead of dex anyways.

lukemarsden (approved):
nice work :tada:! i didn't realise we were just failing to launch dex! sorry :)

as a slight optimization, you could move `make launch-dex` into https://github.com/pachyderm/pachyderm/blob/master/.testfaster.yml#L115 so that it gets started ahead of the tests running, which would speed things up a bit, but that has the possible downside of running dex in places it's not needed - so your call. feel free to merge as-is, or make that change and merge without further review once CI is green. thanks!

lukemarsden:
Btw as long as you are reasonably confident changes aren't making Travis work, I don't mind if you merge deflaking changes with failing Travis builds. Insisting on Travis being green for every single change will slow us down.

PR #5523, opened 2020-12-11
Allow pipeline creation with extant output repo.

PFedak:
With auth enabled, this resulted in an error, preventing the creation of
build pipelines.


actgardner (approved):

actgardner:
Is this covered by any integration tests? If not it might be good to add one to avoid regressions

PR #5525, opened 2020-12-12
Various CI stability fixes

lukemarsden:
- pass through `retain_slots` param (translates to `--retain-slots` on `testctl get`, which allows certain CircleCI runs e.g. those triggered by https://github.com/pachyderm/ci-analytics/blob/main/loop.py to say "leave failed VMs around for longer" to make it easier to deflake them - they won't be cleaned up as quickly)
- add docker image that was being hammered by prewarm docker build which was causing docker hub rate limits (also fixed by fixing docker login in prewarm script on the testfaster side)
- bump to a newer version of pachyderm hash to include Makefile fixes in prewarm script
- build the pipeline build makefile target which fixes some tests
- bump mem to 16GB to fix OOMs in some tests
- switch curl to wget to make it less flaky downloading the minio client (!)
- mark another test as flaky
- a mascot easter egg

please merge if approved as i'm on vacation


etc/testing/images/ubuntu_with_s3_clients/Dockerfile
```diff
@@ -8,11 +8,12 @@ RUN apt-get update -y && apt-get install -y \
...
   groff \
   apt-transport-https \
-  ca-certificates gnupg
+  ca-certificates gnupg \
+  wget
 RUN python3 -m easy_install pip
 RUN pip3 install --upgrade pip
 RUN pip3 install awscli --upgrade
-RUN curl -L https://dl.min.io/client/mc/release/linux-amd64/mc >/usr/local/bin/mc \
+RUN wget -O /usr/local/bin/mc https://dl.min.io/client/mc/release/linux-amd64/mc \
```
nitinjainsj, line 16: nit: Unless we need to know the download progress, I'd use -nv to silence all the download logs.

nitinjainsj (approved):

etc/testing/images/ubuntu_with_s3_clients/Dockerfile
lukemarsden: thanks! fixed!

PR #5527, opened 2020-12-14
Use updated version of crewjam/saml

actgardner:
crewjam/saml has had two CVEs in 2020 (https://github.com/crewjam/saml/security/advisories), bump the version to fix both of them.


src/server/auth/server/config.go
```diff
@@ -517,7 +517,6 @@ func (a *apiServer) setCacheConfig(config *auth.AuthConfig) error {
 	for _, idp := range newConfig.IDPs {
 		if idp.SAML != nil {
 			a.samlSP = &saml.ServiceProvider{
-				Logger:      logrus.StandardLogger(),
```
actgardner, line 520: They've removed the `logger` from ServiceProvider because they no longer log errors during validation.

nitinjainsj (approved):
LGTM.

Issue #5448, opened 2020-11-18
Hashtree cache is shared across workers in tests


msteffen:
I believe this is the underlying cause of https://github.com/pachyderm/pachyderm/issues/5391 and #5503 

Basically, when we deploy pachyerm with local object storage (`pachctl deploy local`), we have a hostpath mount in all pods connecting `/var/pachyderm/pachd` on the host to `/pach` inside every container (`pachd`, `user`, `storage`). We also set the env var `STORAGE_ROOT` to `/pach` in all containers. Finally we use `STORAGE_ROOT` for a lot of things, including
1. as the base directory of the local object storage client (probably the originally intended use—all pods should use a shared hostdir for local object storage, as regular object storage is also shared across workers), and
2. as the base directory of the worker cache, which is the problem.
3. (and, I believe but haven't confirmed, as a prefix for all object storage paths when using non-local object storage, which, if true, is a little inefficient but I don't think hurts us much)

Because of 2), when we run our test suite and worker A deletes a chunk from their worker cache (i.e. `/pach/chunk/a1b2c3/chunk-d4e5f6`), the chunk disappears in worker B's cache (because `/pach` mounts a common hostpath in both containers). This causes some mysterious failures, with the most common one I've seen that a pipeline gets stuck in merging due to `os.Open` returning an error and then getting retried indefinitely.

I think this is likely affecting all tests that run pipelines with more than one worker (either because of parallelism > 1, or because of multiple connected pipelines, as the different pipelines' workers are also sharing a hashtree cache). I see two TODOs here:
1. If `os.Open` returns an error in `localcache.Get()`, instead of `Get()` passing the error up and causing the worker to (I believe?) get stuck in an infinite retry loop, where it calls `fetchChunk -...-> localcache.Get() -> os.Open() -> "not found"` indefinitely, it should remove the entry from `localcache`'s in-memory map, causing `fetchChunk` to re-download the data from object storage
2. we shouldn't use one env var (`STORAGE_ROOT`) and one volume (`pach-disk`) for both object storage and the worker cache. The worker cache should probably be inside an EmptyDir mount, which means two mounts (`pach-disk -> hostPath`, and `pach-cache -> emptyDir` or something) and two env vars (`STORAGE_ROOT`, and `CACHE_ROOT` or something) for all pods

msteffen:
(I think it's *possible* that this is hurting us in production clusters too, but I'd have to read the code more carefully/do some experiments to be sure.

[...]

PR #5530, opened 2020-12-14
Don't use the same host path for hashtree caches across workers in a local deployment

Tryneus:
This adds a new environment variable for controlling where the worker stores its hashtree caches, using an `EmptyDir` volume that won't be shared with other nodes in a local cluster.  This is responsible for some test flakes, as discussed in #5448, and could potentially affect any test with pipelines with concurrency > 1.


msteffen (approved):
LGTM. Thanks for fixing this! I think this will fix `TestExtractRestore*` too (which doesn't have parallelism, but does have multiple pipelines that run together, and therefore still multiple workers on the same node, and is also super flaky) and possibly other tests. I'm very excited about this fix, and I'm really hoping that it boosts CI reliability a ton.

I also think it might make sense to add `delete(c.keys, key)` in localcache here: https://github.com/pachyderm/pachyderm/blob/master/src/server/pkg/localcache/cache.go#L77 (i.e. if getting the key didn't work, then make sure future calls to `Has(key)` and such stop returning true) but you're more familiar with that code than I am.

PR #5538, opened 2020-12-16
Fix incorrect trigger behavior on pipeline update and race condition on triggering with rapid commits

Tryneus:
Debugged with @PFedak and @jdoliner.  The `TestTrigger` test was broken in a couple ways in PPS and PFS.

First, when updating a pipeline with a trigger, it would not set a new `Head` which would cause the trigger code to reset the trigger condition (as if it were a new branch).

Secondly, when finishing commits, the trigger code would only proceed if the new commit was the head of a branch.  With pipelines, this is actually not guaranteed because there may be several output commits created before the first output commit finishes (and at this point it is no longer the head of its branch).  To avoid this, we always consider a commit to be the head of its branch when running the trigger code within `FinishCommit`.

### DIFF
--- a/src/server/pfs/server/trigger.go
+++ b/src/server/pfs/server/trigger.go
@@ -34,7 +34,9 @@ func (d *driver) triggerCommit(
 	}
 	// find which branches this commit is the head of
 	headBranches := make(map[string]bool)
-	headBranches[newHead.Branch.Name] = true
+	if newHead.Branch != nil {
+		headBranches[newHead.Branch.Name] = true
+	}
 	for _, b := range repoInfo.Branches {
 		bi := &pfs.BranchInfo{}
 		if err := branches.Get(b.Name, bi); err != nil {
--- a/src/server/pfs/server/trigger.go
+++ b/src/server/pfs/server/trigger.go
@@ -35,6 +35,12 @@ func (d *driver) triggerCommit(
 	// find which branches this commit is the head of
 	headBranches := make(map[string]bool)
 	if newHead.Branch != nil {
+		// If the commit was made as part of a branch, then it _was_ the branch head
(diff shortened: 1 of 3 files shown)


src/server/pps/server/api_server.go
```diff
@@ -2607,8 +2607,20 @@ func (a *apiServer) CreatePipeline(ctx context.Context, request *pps.CreatePipel
...
+			_, err := pfsClient.InspectBranch(ctx, &pfs.InspectBranchRequest{
+				Branch: client.NewBranch(input.Pfs.Repo, input.Pfs.Branch),
+			})
+
+			if err != nil && !isNotFoundErr(err) {
+				visitErr = err
+				return
+			} else if err == nil {
+				prevHead = client.NewCommit(input.Pfs.Repo, input.Pfs.Branch)
+			}
```
PFedak, line 2620: kind of silly to critique my own code, but I guess I didn't really need the extra `err` variable

src/server/pfs/server/trigger.go
```diff
@@ -34,6 +34,7 @@ func (d *driver) triggerCommit(
 	}
 	// find which branches this commit is the head of
 	headBranches := make(map[string]bool)
+	headBranches[newHead.Branch.Name] = true
```
PFedak, line 37: I feel like this deserves a comment

PFedak (approved):
LGTM

src/server/pps/server/api_server.go
Tryneus: I think you do, or you would need to clear it if `isNotFoundErr(err)`

Tryneus:
@PFedak - added a comment and reorganized the `CreatePipeline` changes a bit

PR #5539, opened 2020-12-17
Only check admin witnesses once

actgardner:
The way this code was structured, we checked whether `len(adminWitnesses) == 0` for each admin we added or removed. Since we're iterating over a map, the order is not deterministic. If you're trying to atomically swap the only admin, there's a 50-50 chance you'll do a remove before an add, which would fail even though at the end of the transaction you'll still have one admin. This is the source of the flakiness in `TestAdmins` in the auth cmd tests.

Pull the witness check out of the loop over role bindings, so we add and remove all the role bindings first, then we check if there's at least one admin remaining.

### DIFF
--- a/src/server/auth/server/api_server.go
+++ b/src/server/auth/server/api_server.go
@@ -942,22 +942,22 @@ func (a *apiServer) applyClusterRoleBindings(ctx context.Context, roleBindings m
 			} else if err := fsAdmins.Delete(principal); err != nil && !col.IsErrNotFound(err) {
 				return err
 			}
+		}
 
-			// select an arbitrary existential witness to guarantee that that there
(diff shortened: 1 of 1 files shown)


src/server/auth/server/api_server.go
```diff
@@ -942,22 +942,22 @@ func (a *apiServer) applyClusterRoleBindings(ctx context.Context, roleBindings m
 			} else if err := fsAdmins.Delete(principal); err != nil && !col.IsErrNotFound(err) {
 				return err
 			}
+		}
```
nitinjainsj, line 945: LOL. This is the bug fix? I was trying to see what changed in the next section.

nitinjainsj (approved):
If the tests are passing. LGTM.

PR #5541, opened 2020-12-18
Use channel for pipelinePoller to reduce etcd contention

msteffen:
This or should unblock the PFS 2.0 alpha by reducing the amount of etcd contention caused by pollPipelines.

Each commit builds, so they can be reviewed independently


src/server/pps/server/poller.go
```diff
@@ -119,27 +147,133 @@ func (a *apiServer) pollPipelines(pachClient *client.APIClient) {
...
 		if ctx.Err() == nil {
 			panic("pollPipelines is exiting prematurely which should not happen; restarting pod...")
 		}
 	}
 }
+
+func (m *ppsMaster) pollPipelinePods(pollClient *client.APIClient) {
+	ctx := pollClient.Ctx()
+	if err := backoff.RetryUntilCancel(ctx, func() error {
+		// watchChan will be nil if the Watch call below errors, this means
```
PFedak, line 168: seems like we don't need this comment anymore

src/server/pps/server/poller.go
```diff
@@ -119,27 +147,133 @@ func (a *apiServer) pollPipelines(pachClient *client.APIClient) {
...
+		kubePipelineWatch, err := m.a.env.GetKubeClient().CoreV1().Pods(m.a.namespace).Watch(
+			metav1.ListOptions{
+				LabelSelector: metav1.FormatLabelSelector(metav1.SetAsLabelSelector(
+					map[string]string{
+						"component": "worker",
+					})),
+				Watch: true,
+			})
+		if err != nil {
+			return errors.Wrap(err, "failed to watch kubernetes pods")
```
PFedak, line 182: not sure what type of errors we expect here - is there any concern about this just constantly looping if there's some k8s connection issue?

src/server/pps/server/poller.go
```diff
@@ -119,27 +147,133 @@ func (a *apiServer) pollPipelines(pachClient *client.APIClient) {
...
+				if condition.Type == v1.PodScheduled &&
+					condition.Status != v1.ConditionTrue && failures[condition.Reason] {
+					if err := m.a.setPipelineCrashing(ctx, pipelineName, condition.Message); err != nil {
+						return errors.Wrap(err, "error moving pipeline to CRASHING")
+					}
+				}
+			}
+		}
+		return backoff.ErrContinue // keep polling until cancelled (RetryUntilCancel)
+	}, &backoff.ZeroBackOff{}, backoff.NotifyContinue("pollPipelinePods"),
```
PFedak, line 220: similar to above, is there a reason to make this instant rather than the exponential infinite backoff it was before? not sure the failure modes here

src/server/pps/server/poller.go
```diff
@@ -119,27 +147,133 @@ func (a *apiServer) pollPipelines(pachClient *client.APIClient) {
...
+// writeEv and deleteEv events to the PPS master when it sees them.
+//
+// pollPipelinesEtcd is unlike the other poll and monitor goroutines in that it
+// sees the result of other poll/monitor goroutines' writes. For example, when
+// pollPipelinePods (above) observes that a pipeline is crashing and updates its
+// state in etcd, the flow for starting monitorPipelineCrashing is:
+//
+//  k8s watch ─> pollPipelinePods  ╭──> pollPipelinesEtcd  ╭──> m.run()
+//                      │          │            │          │      │
+//                      ↓          │            ↓          │      ↓
```
PFedak, line 236: Between this and the invariant in that createBranch comment, I wonder if you have the most unique unicode codepoints across commits.

PFedak (approved):
LGTM, the reorganization makes sense. Only comments are about the backoff/error handling in the monitor functions.

src/server/pps/server/master.go
```diff
@@ -181,43 +118,84 @@ func (a *apiServer) setPipelineCrashing(ctx context.Context, pipelineName string
...
+}
+
+func (m *ppsMaster) deletePipelineResources(pipelineName string) (retErr error) {
 	log.Infof("PPS master: deleting resources for pipeline %q", pipelineName)
-	span, ctx := tracing.AddSpanToAnyExisting(ctx, //lint:ignore SA4006 ctx is unused, but better to have the right ctx in scope so people don't use the wrong one
+	span, ctx := tracing.AddSpanToAnyExisting(m.masterClient.Ctx(),
 		"/pps.Master/DeletePipelineResources", "pipeline", pipelineName)
 	defer func() {
-		tracing.TagAnySpan(span, "err", retErr)
+		tracing.TagAnySpan(ctx, "err", retErr)
```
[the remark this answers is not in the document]
msteffen: Yup, exactly

src/server/pps/server/poller.go
msteffen: :+1: Removed

src/server/pps/server/poller.go
msteffen: This is a good question; reflecting on it a little, I *think* this is the right choice. Since this gets run when the PPS master starts, the options I see are:
1. Restart all of pachd (i.e. crash & reboot)
2. Restart the PPS master (or yield the PPS master lock and start the PPS master somewhere else; i.e. return an error)
3. Retry the k8s `Watch()` RPC (what this does)

My rationale for preferring option 3 is this: it seems like the high-level goal is to start a functional k8s watch as quickly as possible. I don't see any reason why a k8s watch is more likely to succeed if we retry it on a different machine. Also, because this code runs on master startup (which happens on pachd startup, and possibly other times as well), I don't think there's any broader context for this code that would need to be part of a retry loop. And shrinking the retry loop to just this function seems a lot faster than the other two options (since we don't have to deal with scheduling or master election), so I think this is the best option.

src/server/pps/server/poller.go
msteffen: I'm not totally sure what the right answer is either; I thought about this some more too and changed it back to exponential backoff because this code is making an RPC, so if the `Watch` RPC is failing because k8s API server is overloaded, then at least pachd won't be contributing to the overload.

src/server/pps/server/poller.go
msteffen: :bow: Thank you for your kind words. As soon as I heard Go supported unicode source files, I knew what my calling was in life.

PR #5545, opened 2020-12-18
GoReleaser fixes

msteffen:
- Make GoReleaser work on Ubuntu desktops (which keep $HOME and therefore $HOME/go in a mounted ecryptfs volume by default)
- Fix build pipeline docker images

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,8 +94,9 @@ release-pachctl:
 docker-build:
 	DOCKER_BUILDKIT=1 goreleaser release -p 1 --snapshot $(GORELDEBUG) --skip-publish --rm-dist -f goreleaser/docker.yml
 
-docker-build-pipeline-build:
-	DOCKER_BUILDKIT=1 goreleaser release -p 1 --snapshot $(GORELDEBUG) --skip-publish --rm-dist -f goreleaser/docker-build-pipelines.yml
+docker-build-pipeline-build: install
+	VERSION=$$(pachctl version --client-only) DOCKER_BUILDKIT=1 \
+	  goreleaser release -p 1 --snapshot $(GORELDEBUG) --skip-publish --rm-dist -f goreleaser/docker-build-pipelines.yml
 
 docker-build-proto:
 	docker build $(DOCKER_BUILD_FLAGS) -t pachyderm_proto etc/proto
@@ -139,9 +140,11 @@ docker-push: docker-tag
 	$(SKIP) docker push pachyderm/worker:$(VERSION)
 	$(SKIP) docker push pachyderm/pachctl:$(VERSION)
 
-docker-push-pipeline-build:
-	$(SKIP) docker push pachyderm/go-build:$(VERSION)
-	$(SKIP) docker push pachyderm/python-build:$(VERSION)
+docker-push-pipeline-build: install
+	$(SKIP) ls etc/pipeline-build | xargs -I {} docker push pachyderm/{}-build:$$(pachctl version --client-only)
+
+docker-push-pipeline-build-to-minikube: install
+	$(SKIP) ls etc/pipeline-build | xargs -I {} etc/kube/push-to-minikube.sh pachyderm/{}-build:$$(pachctl version --client-only)
 
 check-kubectl:
 	@# check that kubectl is installed
--- a/etc/pipeline-build/Makefile
+++ b/etc/pipeline-build/Makefile
@@ -1,9 +1,0 @@
-.PHONY: docker-build push-to-minikube
-
-export DOCKER_BUILDKIT = 1
-
-docker-build:
-	ls | grep -v Makefile | xargs -I {} bash -c 'cd ../..; docker build -t pachyderm/{}-build:$(shell $(GOPATH)/bin/pachctl version --client-only) -f etc/pipeline-build/{}/Dockerfile .'
-
-push-to-minikube: docker-build
-	ls | grep -v Makefile | xargs -I {} ../kube/push-to-minikube.sh pachyderm/{}-build:$(shell $(GOPATH)/bin/pachctl version --client-only)
--- a/etc/pipeline-build/Makefile
+++ b/etc/pipeline-build/Makefile
@@ -1,7 +1,9 @@
 .PHONY: docker-build push-to-minikube
 
+export DOCKER_BUILDKIT = 1
+
 docker-build:
-	ls | grep -v Makefile | xargs -I {} docker build -t pachyderm/{}-build:$(shell $(GOPATH)/bin/pachctl version --client-only) {}
+	ls | grep -v Makefile | xargs -I {} bash -c 'cd ../..; docker build -t pachyderm/{}-build:$(shell $(GOPATH)/bin/pachctl version --client-only) -f etc/pipeline-build/{}/Dockerfile .'
 
 push-to-minikube: docker-build
 	ls | grep -v Makefile | xargs -I {} ../kube/push-to-minikube.sh pachyderm/{}-build:$(shell $(GOPATH)/bin/pachctl version --client-only)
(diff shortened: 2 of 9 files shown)


nitinjainsj (commented):
Remind me: Why not just get rid of this Makefile and add a target in the top-level Makefile to explicitly build go and python targets? If we ever add new lang support we can a new target or add another line to the same target?

msteffen:
> Remind me: Why not just get rid of this Makefile and add a target in the top-level Makefile to explicitly build go and python targets? If we ever add new lang support we can a new target or add another line to the same target?

My mistake, I forgot that we decided to take that approach. We actually already have make targets (`docker-build-pipeline-build` and `docker-push-pipeline-build`)

I can remove `etc/pipeline-build/Makefile`, but one question I have is about the `VERSION` environment variable. How is that set? It seems like some of the release targets depend on it as well. I can set it easily enough with `export VERSION=$(pachctl version --client-only)` (if that's right), but I was wondering what the normal flow is?

nitinjainsj:
I was also playing with the same code. Getting rid of the Makefile seems clean. Your version looks fine. We have this pattern of doing `make install`; `VERSION=pachctl version --client-only`; `export VERSION`; `make docker...`

I want to clean this up so we won't require the first 3 steps. I'll try out a few things. If can make it work in a clean way, I can clean up all the build steps. For now, the pattern to export VERSION seems the right thing to do.

etc/pipeline-build/Makefile
```diff
@@ -1,7 +0,0 @@
-.PHONY: docker-build push-to-minikube
-
-docker-build:
-	ls | grep -v Makefile | xargs -I {} docker build -t pachyderm/{}-build:$(shell $(GOPATH)/bin/pachctl version --client-only) {}
-
-push-to-minikube: docker-build
-	ls | grep -v Makefile | xargs -I {} ../kube/push-to-minikube.sh pachyderm/{}-build:$(shell $(GOPATH)/bin/pachctl version --client-only)
```
nitinjainsj, line 7: I think you will need to call in the top makefile. Maybe create a target push-build-pipeline-to-minikube

`./etc/kube/push-to-minikube.sh pachyderm/go-build:$(VERSION)`
`./etc/kube/push-to-minikube.sh pachyderm/python-build:$(VERSION)`

msteffen: Done

msteffen:
> I was also playing with the same code. Getting rid of the Makefile seems clean. Your version looks fine. We have this pattern of doing `make install`; `VERSION=pachctl version --client-only`; `export VERSION`; `make docker...`
> 
> I want to clean this up so we won't require the first 3 steps. I'll try out a few things. If can make it work in a clean way, I can clean up all the build steps. For now, the pattern to export VERSION seems the right thing to do.

This makes sense. As I've been working on this code, I've started understanding the goreleaser code a bit better. I have some ideas and I'm curious what you think:
1. I think instead of having goreleaser run on the host and build the `pachyderm_build` image followed by the `pachd`, `pachctl` and `worker` images, it might make sense to build the `pachyderm_build` image on host and run `goreleaser` inside of that. I think then we'd be able to get rid of the `notusedbuild` targets—`goreleaser` would build `pachd`, `worker`, etc with `go build` as build targets and then package them into docker containers from inside the `pachyderm_build` container and release them (or not). I think this might actually be faster than what we do now

2. It looks like the way releases work now is, you run `make point-release` which calls `etc/build/make_release.sh` which calls `make release`. I wonder if it would make sense to get rid of `make release` (since it's not supposed to be called directly anyway) and move it into `make_release.sh`. Then `make point-release` calls `make_release.sh` and that does all the work, which is a fairly common pattern in our repo. I think this would limit the use of `VERSION` to `make_release.sh` (the other make targets that use it, `docker tag` and `docker push`, could use `$$(pachctl version --client-only)` instead of `$(VERISON), which would make it easier to call them directly anyway IMO)

msteffen:
Final update: because `goreleaser/docker-build-pipelines.yml` doesn't include `.dockerignore` in `extra_files`, it's not copied to the goreleaser temp dir, and `etc` is therefore not ignored when goreleaser runs `docker build`. So the extra `etc/pipeline-build/{go,python}/Dockerfile.dockerignore` files aren't necessary and I removed them.

Makefile
```diff
@@ -95,8 +95,9 @@ release-pachctl:
 docker-build:
 	DOCKER_BUILDKIT=1 goreleaser release -p 1 --snapshot $(GORELDEBUG) --skip-publish --rm-dist -f goreleaser/docker.yml
 
-docker-build-pipeline-build:
-	DOCKER_BUILDKIT=1 goreleaser release -p 1 --snapshot $(GORELDEBUG) --skip-publish --rm-dist -f goreleaser/docker-build-pipelines.yml
+docker-build-pipeline-build: install
```
nitinjainsj, line 98: This looks cleaner than what we have today 👏 

Even though I don't like that we are calling `install` when version should be available. It just adds to build times. Since you are changing this, you may want to fix the places where we call these targets and remove the `make install` part before these targets.

nitinjainsj:
> > I was also playing with the same code. Getting rid of the Makefile seems clean. Your version looks fine. We have this pattern of doing `make install`; `VERSION=pachctl version --client-only`; `export VERSION`; `make docker...`
> > I want to clean this up so we won't require the first 3 steps. I'll try out a few things. If can make it work in a clean way, I can clean up all the build steps. For now, the pattern to export VERSION seems the right thing to do.
> 
> This makes sense. As I've been working on this code, I've started understanding the goreleaser code a bit better. I have some ideas and I'm curious what you think:
> 
> 1. I think instead of having goreleaser run on the host and build the `pachyderm_build` image followed by the `pachd`, `pachctl` and `worker` images, it might make sense to build the `pachyderm_build` image on host and then run `goreleaser` inside of `pachyderm_build` (https://goreleaser.com/install/#running-with-docker). I think then we'd be able to get rid of the `notusedbuild` targets—`goreleaser` would build `pachd`, `worker`, etc with `go build`, as build targets, and then package them into docker containers and release them (or not) from inside the `pachyderm_build` container. I think this might actually be faster too, as the builds would all happen inside the same container.
> 2. It looks like the way releases work now is, you run `make point-release` which calls `etc/build/make_release.sh` which calls `make release`. I wonder if it would make sense to get rid of `make release` (since it's not supposed to be called directly anyway) and move it into `make_release.sh`. Then `make point-release` calls `make_release.sh` and that does all the work, which is a fairly common pattern in our repo. I think this would limit the use of `VERSION` to `make_release.sh` (the other make targets that use it, `docker tag` and `docker push`, could use `$$(pachctl version --client-only)` instead of `$(VERISON), which would make it easier to call them directly anyway IMO)

I am trying to unpack the pachyderm_build container -- I am unsure how this works right now. But high level what you described for 1) makes sense. 

For 2), we can do what you described if that makes the Makefile targets cleaner and easier to understand.

Makefile
nitinjainsj: We should do this after you come back from the holidays. But something to think about. The docker-* targets are all used for internal testing and CI. We can put in any version number in the images ... like 9.99.9-<commitID>. This way we get rid of the whole make install step everywhere.

nitinjainsj (approved):
Fix the caller (travis)

msteffen:
I'm deciding to merge this PR even though Travis isn't passing, for two reasons:
- This is a tooling PR, not a product PR. It will be harder for me to develop locally until it's merged, and shouldn't affect customers, which makes it relatively higher value and relatively lower risk
- Now the the 2.0 branch is merged (and this was rebased on master after the merge), Travis tests are unreliable, so we're relying on CircleCI, which is passing.

I do think the failures I was seeing in Travis will need to be addressed, and I filed #5590, #5601, #5602, and #5603 to document them

PR #5548, opened 2020-12-21
Fileset Tests

brendoncarroll:
Adds tests to fileset package including an a gc test with a deleter for all the objects used by the storage layer, and brings back some of the old fileset tests.


src/server/pkg/storage/fileset/writer.go
```diff
@@ -68,13 +69,15 @@ func newWriter(ctx context.Context, store Store, tracker track.Tracker, chunks *
...
-	w.deletive = index.NewWriter(ctx, chunks, "deletive-index-writer-"+uuidStr)
-	w.cw = chunks.NewWriter(ctx, "chunk-writer-"+uuidStr, w.callback, chunkWriterOpts...)
+	w.additive = index.NewWriter(ctx, chunks, "additive-index-writer")
+	w.deletive = index.NewWriter(ctx, chunks, "deletive-index-writer")
+	w.cw = chunks.NewWriter(ctx, "chunk-writer", w.callback, chunkWriterOpts...)
 	return w
 }
 
 // Append creates an append operation for a file and provides a scoped file writer.
+// TODO: providing the tag here, since it is required, is a better API than getting a *FileWriter
```
brycemcanally, line 79: We need to be able to support multiple tags for a single file, so just associating a single tag with a file writer would not work. I do agree that this API could probably be improved, and I am leaning towards inverting the control by taking a reader instead of implementing the io.Writer interface which would allow us to provide a function that takes a tag and the reader for the tagged data.

src/server/pkg/storage/fileset/fileset_test.go
```diff
@@ -1,274 +1,192 @@
...
+				tag:  tag,
+				data: data,
+			})
+		}
+		files = append(files, &testFile{
+			path: "/" + fileName,
+		})
+	}
+
+	// Write out ten filesets where each subsequent fileset has the content of one random file changed.
```
brycemcanally, line 124: We should have a test that does what this comment describes. That is how this test worked previously. I am fine with leaving this test as a basic write / read check, then having a test called `TestWriteThenReadFuzz` which does what is described here. We also need a merge test.

src/server/pkg/storage/fileset/storage.go
```diff
@@ -126,6 +126,12 @@ func (s *Storage) Open(ctx context.Context, fileSets []string, opts ...index.Opt
 	return newMergeReader(s.chunks, fss), nil
 }
 
+// TODO: public?
+func (s *Storage) exists(ctx context.Context, fileset string) bool {
```
brycemcanally, line 130: This type of check seems sketchy. Interpreting any error as doesn't exist seems error prone.

src/server/pkg/storage/fileset/writer.go
brendoncarroll: I will change the comment to mention taking a tag + io.Reader

src/server/pkg/storage/fileset/fileset_test.go
brendoncarroll: I'm going to hold of on the merge / compaction test until the new fileset API stuff is done.  There is actually a Merge method which takes a slice of fileset IDs and returns a single one.  So it will be pretty straightforward to write test vectors for that I think.

src/server/pkg/storage/fileset/storage.go
brendoncarroll: Yeah I was thinking about that. We do this in the obj.Client.  I'll change it to return `(bool, error)`

src/server/pkg/testutil/context.go
```diff
@@ -0,0 +1,20 @@
+package testutil
+
+import (
+	"context"
+	"testing"
+)
+
+// NewTestContext returns a context scoped to the current test
```
brycemcanally, line 8: Is this really necessary? Go will just kill the binary if the timeout is exceeded. Also, if we keep this, the cleanup will need to be within the scope of the deadline check because it panics as is with a timeout of 0.

src/server/pkg/storage/fileset/storage.go
brycemcanally: We should probably do the same for our object clients at some point.

src/server/pkg/testutil/context.go
brendoncarroll: removing this

src/internal/storage/fileset/fileset_test.go
```diff
@@ -1,274 +1,224 @@
...
+	eg := errgroup.Group{}
+	eg.Go(func() error {
+		return f.Content(w)
+	})
+	eg.Go(func() (retErr error) {
+		defer func() {
+			if retErr != nil {
+				r.CloseWithError(retErr)
+			} else {
+				r.Close()
```
brycemcanally, line 72: Is this close necessary?

brendoncarroll: Probably not in the test, I think your are supposed to close it so that any writers will get errors instead of hanging.

brycemcanally (approved):
lgtm

PR #5550, opened 2020-12-22
Build steps cleanup after goreleaser

nitinjainsj:
Remove stale build file.


msteffen (approved):
It looks like commit 2436330 might have reverted a lot of the changes in this PR (to `Makefile`, as well as deleting some of the inner Makefiles). I just wanted to check that that was deliberate. 

Other than that, LGTM

nitinjainsj:
> It looks like commit [2436330](https://github.com/pachyderm/pachyderm/commit/2436330ab51417668bf599a76e4b1b46e9c4ee73) might have reverted a lot of the changes in this PR (to `Makefile`, as well as deleting some of the inner Makefiles). I just wanted to check that that was deliberate.
> 
> Other than that, LGTM

Yes. It was intentional. I went down the path of finding where we were calling push-to-minikube. There is another target in the s3 gateway dir. The changes to Makefile are easy, but I am not sure if the cleanup was necessary -- hence the revert. The other set of changes were similar/same to your PR, so I took those out.

PR #5552, opened 2020-12-22
Change ubuntu to 18.04 and golang to 1.15.4

nitinjainsj:
This seems like a big review, but it is mostly cleanup.

For test faster... we got are not on golang 1.15.4. There is one test that requires ubuntu:16.04 because that was the DefaultUserImage in 1.11 & 1.7

Also fixed a few dangling tags for images in our examples dir. -- will get the customer team to review that part.


lukemarsden:
stale

PR #5556, opened 2020-12-23
An intro-level example of spout 2.0

nadegepepin:
First version - needs code clean up/ WIP

Overall, a couple of thoughts:
I used **python-pachyderm** to put files in the output repo of my spout:
 `client = python_pachyderm.Client(host="pachd", port=650)
...
  with client.commit('spout', 'master') as c:
 ...
      client.put_file_bytes(c, filename, msg)`

1. I could not connect to pachd without mentioning the host (pachd) and port number (650). Ideally, I should not need to do that.
2. Why do we need to mention the name of the output repo in the commit? Shouldn't it default to the pipeline's name in the case of a spout?
3. Also, using `client.commit('ANYNAME', 'master')`, I was able to commit in another repo that I just created (ANYNAME), unrelated to the spout. I don't think that there is any mechanism that prevents us from doing that, but I somehow had understood that it would not be possible (I might have misunderstood though. )
2. In an ideal world, I should have used the SpoutManager -> **python-pachyderm** needs to be updated. (it is still using the tarstream/named pipe)
3. I was unable to use build-pipelines.
3. Last, about data lineage. I wanted to add a `pachctl inspect commit` on any commit in the last output repo `processor` to showcase where the data came from. Is there anything specific that shows that it came from an outside source? Is it worth adding in the example? Below an example of what I got.
Screen Shot 2020-12-23 at 5 16 44 PM


adelelopez:
Everything here looks really great!
1. I think that's a bug where it's not doing the appropriate port-forwarding
2. You mean in the put-file?
3. Yeah, we should prevent that.
4. The entire spout-manager is deprecated now
5. That's another bug.
6. That's a good idea, I'll see if there's a way we can note that info.

adelelopez (approved):

examples/spouts/spout101/README.md
```diff
@@ -0,0 +1,181 @@
...
+between an external stream of data and Pachyderm's repo.
+In Martin Fowler's enterprise integration patterns' terms,
+a Pachyderm spout is a 
+*[Polling Consumer](https://www.enterpriseintegrationpatterns.com/patterns/messaging/PollingConsumer.html)* 
+(actively polls for a message, processes it, then polls for another).
+
+Generally, you use spouts for situations 
+when the interval between new data generation
+is large or sporadic, but the latency requirement
+to start the processing is short.
```
JoeyZwicker, line 19:
```suggestion
Generally, spout pipelines are ideal for situations 
when the frequency of new incoming data 
is sporadic and the latency requirement
to start the processing is short. 
```

examples/spouts/spout101/README.md
```diff
@@ -0,0 +1,181 @@
...
+a Pachyderm spout is a 
+*[Polling Consumer](https://www.enterpriseintegrationpatterns.com/patterns/messaging/PollingConsumer.html)* 
+(actively polls for a message, processes it, then polls for another).
+
+Generally, you use spouts for situations 
+when the interval between new data generation
+is large or sporadic, but the latency requirement
+to start the processing is short. 
+Therefore, a regular pipeline with a cron input
+that polls for new data might not be an optimal solution.
```
JoeyZwicker, line 21:
```suggestion
In these workloads, a regular pipeline with a cron input
that polls for new data at a consistent time interval might not be an optimal solution.
```

examples/spouts/spout101/README.md
JoeyZwicker: I don't think of spouts as particularly suited or not for large time intervals between data inputs. They are totally fine for this, but it's actually just as common (or sometimes moreso) use spouts for streaming data that might be coming in at a very high frequency. 

Do you think it's worth making a not here (or somewhere, I haven't finished reading the whole docs yet) that users should be careful not to use spouts to make "too many" commits too frequently. While small, there is an overhead per commit and per job (sophisticated versioning and lineage tracking aren't free unfortunately) and therefore we generally recommend making commits no more frequently that a few per minute. Pachyderm in general can handle micro batches much better that full streaming.

examples/spouts/spout101/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+A spout pipeline has three major differences from regular pipelines:
+- it runs continuously
```
JoeyZwicker, line 24:
```suggestion
- it runs continuously, waiting for new events from any subscribed queues
```

examples/spouts/spout101/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+You configure a spout in the 
+[pipeline specification](https://docs.pachyderm.com/latest/reference/pipeline_spec/)
+file by adding a `spout` section after its pipeline section.
+At a minimum, a spout pipeline will need a pipeline name,
+an empty spout section, and a transform section
+where you will specify the transformation you want
+to execute on the messages consumed.
+
+>![pach_logo](./img/pach_logo.svg) Note that the use
```
JoeyZwicker, line 37: do we have some methodology to track notices like this such that when build + spouts work we remember to remove this?

examples/spouts/spout101/README.md
```diff
@@ -0,0 +1,181 @@
...
+```
+Ideally, have your pachctl and pachd versions match. At a minimum, you should always use the same major & minor versions of your pachctl and pachd. 
+
+## Example - Spout 101 
+***Goal***
+In this example,
+we will keep generating two random strings, 
+one of 1KB and one of 2KB,
+at intervals varying between 10s and 30s. 
+A spout pipeline will actively poll those data,
```
JoeyZwicker, line 81: per other comment, I don't quite see this is polling. What do you think? Perhaps, "Our spout pipeline will actively receive those events and commit them as text files to the output repo" -- or something like that.

examples/spouts/spout101/README.md
```diff
@@ -0,0 +1,181 @@
...
+then commit them to its output repo -
+in the form of a text file - using the **pachctl put file** command.
```
JoeyZwicker, line 83: One thing to be a little careful of in terminology. `put file` is the API command. pachctl is technically the name of the CLI for the Pachdyerm API. So it doesn't 100% makes sense to say "use pachctl put file with the python client"... it probably more accurate to say "use the `put file` API call via the python client." I think JD or someone on the eng side could probably tell you what sounds best -- I'm not the best judge because I'm not a real developer.

JoeyZwicker: This comment applies to line 91 too

JoeyZwicker (commented):
Looks great. A few very minor comments and then LGTM. I did not fully run the example myself, just reviewed the readme.

IMO, the polling-style code we have written to make it all work nicely with the "fake" messaging queue makes this less of a copy/paste example and more of a conceptual example. I think that's totally fine and you do a goo job of clarifying that in the readme.

examples/spouts/spout101/README.md
```diff
@@ -0,0 +1,181 @@
...
+>![pach_logo](./img/pach_logo.svg) This new implementation of the spout functionality is available in version **1.12 and higher**.
+
+
+## Intro
+A spout is a type of pipeline that ingests 
+streaming data (message queue, database transactions logs,
+event notifications... ). 
+See spout pipelines as a bridge
+between an external stream of data and Pachyderm's repo.
+In Martin Fowler's enterprise integration patterns' terms,
```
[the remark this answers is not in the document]
mindthevirt: I think technically it is polling but should probably be described as "subscribes to a stream of data", which then makes it "pull"-like.

mindthevirt:
After Joey's suggestions are incorporated, I think it will be good to go. I like the idea of this being an example that explains the workings of a Spout and the other examples explaining how a Spout works with SQS, Kafka etc.

examples/spouts/spout101/README.md
nadegepepin: I have created a feature request for this: **Build pipelines support for spouts** #5560, the merge of which will alert me.

examples/spouts/spout101/README.md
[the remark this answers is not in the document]
nadegepepin: It is a matter of perspective.
Data lineage's perspective: A spout is a source of data pushed into Pachyderm's repo and the starting point of the data lineage. It is, in this view, an entry point.
Integration perspective:  A spout is a by-the-book use case of a **Polling Consumer** as defined in the Enterprise Integration Pattern. It is a consumer of external data polled from a third-party messaging system. 
As suggested by Jan, I am replacing the term "polling" with the notion of subscription.

mindthevirt:
> * Yeah, we should prevent that.

@adelelopez did you open an issue for this?

PR #5561, opened 2021-01-04
Embedded identity server

actgardner:


src/client/auth/auth.proto
```diff
@@ -106,6 +106,11 @@ message IDProvider {
      string redirect_uri = 4 [(gogoproto.customname) = "RedirectURI"];
      repeated string additional_scopes = 5;
      bool ignore_email_verified = 6;
+
+     // localhost_issuer ignores the contents of the issuer claim and makes all
+     // OIDC requests to the embedded OIDC provider. This is necessary to support
+     // some network configurations like Minikube.
```
msteffen, line 112: :+1:

msteffen (commented):
I'm still working on understanding some of the changes, but I added a few questions/comments.

web/static/img/atlassian-crowd-icon.svg
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" standalone="no"?>
```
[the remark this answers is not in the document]
actgardner: +1 for making the name clearer.

Mainline Dex requires that the directory structure and the template files exist (see https://github.com/dexidp/dex/blob/fc7f1ef6cc34b0ce181c964d076df57b4bf0b171/server/templates.go#L17-L35). At least one case where Dex would need to serve a UI is if a user had multiple auth providers configured. Most of the templates and directories we could probably just `touch` in the Dockerfile instead of including working templates.

We could also try and maintain a fork of Dex, they're not really amenable to the "embedded" use case so PRs that would make our lives easier probably won't make it upstream.

src/server/identity/server/dex_web.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (w *dexWeb) getServer() *dex_server.Server {
+	var server *dex_server.Server
+	// Get a read lock to check if the server is nil (this is rare)
+	w.RLock()
+	if w.server == nil {
+		// If the server is nil, unlock and acquire the write lock
+		w.RUnlock()
+		w.Lock()
```
msteffen, line 92: One more place where I think you could do `defer w.Unlock()`

src/server/pkg/deploy/assets/postgres.go
```diff
@@ -120,6 +130,10 @@ func PostgresDeployment(opts *AssetOpts, hostPath string) *apps.Deployment {
 									Name:      "postgres-storage",
 									MountPath: "/var/lib/postgresql/data",
 								},
+								{
+									Name:      postgresInitVolumeName,
+									MountPath: "/docker-entrypoint-initdb.d",
```
msteffen, line 135: Just curiosity, but why the reference to docker in the volume name? Does docker interact with postgres somehow?

src/server/pkg/serviceenv/config.go
```diff
@@ -68,7 +69,14 @@ type PachdSpecificConfiguration struct {
 	RequireCriticalServersOnly bool   `env:"REQUIRE_CRITICAL_SERVERS_ONLY",default=false"`
 	MetricsEndpoint            string `env:"METRICS_ENDPOINT",default="`
 	// TODO: Merge this with the worker specific pod name (PPS_POD_NAME) into a global configuration pod name.
-	PachdPodName string `env:"PACHD_POD_NAME,required"`
+	PachdPodName        string `env:"PACHD_POD_NAME,required"`
+	PostgresServiceHost string `env:"POSTGRES_SERVICE_HOST,default=172.17.0.1"`
```
msteffen, line 73: I'm surprised that there's a sensible default for this, but I'm not sure I understand this part totally. This seems like the IP for a kubernetes service called "postgres"—can you control service IPs? Or is this like different entirely?

src/server/identity/server/dex_web.go
```diff
@@ -0,0 +1,118 @@
...
+	dex_server "github.com/dexidp/dex/server"
+	logrus "github.com/sirupsen/logrus"
+)
+
+// webDir is the path to find the static assets for the web server.
+// This is always /dex-assets in the docker image, but it can be overriden for testing
+var webDir = "/dex-assets"
+
+// dexWeb wraps a Dex web server and hot reloads it when the
+// issuer is reconfigured.
```
msteffen, line 19: Maybe just add a note to the effect of `'issuer' must be the address at which this identity server is reachable from all pachd nodes (unused if 'localhost_issuer' is set in pachd's OIDC IDP config). This value may not be knowable at cluster creation time, or may change if the cluster is reconfigured` (either here or over `issuer`, just so there's something about when/why issuer might be reconfigured)?

msteffen (approved):
LGTM. Thanks for being patient with this one!

I left a few suggestions, mostly about comments and stuff.

src/server/pkg/deploy/assets/postgres.go
actgardner: This is part of the postgres Docker image - this is the well-known path where you can mount setup scripts to run when the database is initialized.

src/server/pkg/serviceenv/config.go
actgardner: Removed this default, it was a leftover from before we had a better story for deploying postgres.

PR #5564, opened 2021-01-05
[2.0.0-alpha] Remove ability to activate auth as users other than root

actgardner:
Make activating auth simpler in 2.0 by only supporting one happy path - generating (or supplying) a token for `pach:root`, which has irrevocable cluster admin status. This paves the way to some other simplifications, like eliminating the "make sure we haven't removed all admins" check.


actgardner:
Nevermind, I'll base this on master since we're cutting 1.12 today. See #5565

PR #5565, opened 2021-01-05
Remove ability to activate auth as users other than root

actgardner:
Remove the `subject` and `github_user` options from auth.Activate. This means all 2.0 clusters with auth activated are guaranteed to have a specific root user (`pach:root`).

 In a future PR I want to make it impossible to revoke the root user's admin status, which will be much simpler than checking whether there's always at least one admin. This will also ensure it's always possible to recover a cluster, even if the auth config is bad, as long as you securely store the root token.


src/server/auth/cmds/cmds_test.go
```diff
@@ -45,8 +45,8 @@ func activateAuth(t *testing.T) {
 	// TODO(msteffen): Make sure client & server have the same version
 	// Logout (to clear any expired tokens) and activate Pachyderm auth
 	require.NoError(t, tu.Cmd("pachctl", "auth", "logout").Run())
-	cmd := tu.Cmd("pachctl", "auth", "activate")
-	cmd.Stdin = strings.NewReader("admin\n")
+	cmd := tu.Cmd("pachctl", "auth", "activate", "--supply-root-token")
+	cmd.Stdin = strings.NewReader("iamroot\n")
```
msteffen, line 49: would it make sense to do `tu.RootToken + "\n"` (or maybe plain `tu.RootToken` would work?) in case we change it for some reason?

src/server/auth/cmds/cmds_test.go
```diff
@@ -80,12 +79,12 @@ func TestAuthBasic(t *testing.T) {
 	activateAuth(t)
 	defer deactivateAuth(t)
 	require.NoError(t, tu.BashCmd(`
-		echo "{{.alice}}" | pachctl auth login
-		pachctl create repo {{.repo}}
-		pachctl list repo \
-			| match {{.repo}}
-		pachctl inspect repo {{.repo}}
-		`,
+			echo "{{.alice}}" | pachctl auth login
```
msteffen, line 82: I don't really object but why the extra spacing? Should we be doing this in general?

src/server/auth/server/api_server.go
```diff
@@ -524,51 +524,6 @@ func (a *apiServer) Activate(ctx context.Context, req *auth.ActivateRequest) (re
 		return nil, errors.Errorf("already activated")
 	}
 
-	// The Pachyderm token that Activate() returns will have the TTL
```
msteffen, line 527: Awesome

src/server/pkg/testutil/auth.go
```diff
@@ -100,20 +92,16 @@ func getPachClientInternal(tb testing.TB, subject string) *client.APIClient {
 	if subject == "" {
 		return resultClient // anonymous client
 	}
+	if subject == AdminUser {
+		resultClient.SetAuthToken(RootToken)
+		return resultClient
+	}
+
 	if token, ok := tokenMap[subject]; ok {
```
msteffen, line 100: Does this code make sense to keep around if `UpdateAuthToken` is going away? IIUC, there isn't a way to put tokens into `tokenMap` anymore, which means this code won't run anymore either

msteffen (approved):
Overall LGTM. There are a few nits in the code, but one high-level question I have is: are there any codepaths left where we test GitHub auth, and if not would it ironically make sense to add some tests for it?

src/client/auth/auth.proto
```diff
@@ -28,29 +28,10 @@ import "google/protobuf/timestamp.proto";
...
-  // When Pachyderm is deployed locally, setting this value to a given string
-  // will automatically authenticate the caller as a GitHub user whose username
-  // is that string (unless this "looks like" a GitHub access code, in which
-  // case Pachyderm does retrieve the corresponding GitHub username)
-  string github_token = 1 [(gogoproto.customname) = "GitHubToken"];
-
-  // If set, activate auth for the root user (`pach:root`), and add this
-  // token as the initial login token. The root user will have super
-  // admin permissions on the cluster.
+  // If set, this token is used as the root user login token.
```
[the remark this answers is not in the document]
msteffen: After reviewing the rest of the PR, I now know that it is, so you can ignore this. Or maybe it would make sense to add like "if unset pachd will generate and return a token" or something to the comment?

src/server/pkg/testutil/auth.go
actgardner: I'm going to refactor this further in a follow-on, once we make the "root is always an admin" change most of this file can go away because we just use the root client to make a token for whoever.

actgardner:
> are there any codepaths left where we test GitHub auth, and if not would it ironically make sense to add some tests for it?

Right now we still have code to test the Github login flow, but once we merge the identity server my plan is to remove Github auth altogether, and we'll only need to test OIDC. In the future we can just have the root user call `GetToken` for any user token we want to test with, removing the need for the Github login backdoor that we use for testing today.

PR #5566, opened 2021-01-05
Fix pachctl put file paths when run from windows

Tryneus:
When running on Windows, `pachctl put file` would result in mangled paths due to preserving windows path separators.  This converts the path to use Unix separators for the PFS path.

Before:
```
$ pachctl put file input@master:/foo/bar/baz.txt -f input/baz.txt
$ pachctl list file input@master:/**
NAME                            TYPE SIZE 
/\foo\bar\baz.txt file 2B
```

After:
```
$ pachctl put file input@master:/foo/bar/baz.txt -f input/baz.txt
$ pachctl list file input@master:/**
NAME                            TYPE SIZE 
/foo/bar/baz.txt  file 2B
```


msteffen (approved):
LGTM

PR #5568, opened 2021-01-05
Use a unique pipeline name in filter test.

PFedak:
Seems like just another issue with rapidly deleting and creating a
replication controller for the same pipeline. This was the second of two
consecutive tests that shared a pipeline name. Maybe it's worth going
through and removing all other occurrences.


msteffen (approved):
LGTM

PR #5569, opened 2021-01-06
Migration update 2.0

nitinjainsj:
Seems like we were some missing step in the instructions. Got this to working. Should be ready to merge. I'll send out another PR with changes to instructions.


msteffen (approved):
LGTM

PR #5572, opened 2021-01-06
Require users to be logged in to call secrets API

actgardner:
Add auth checks to the secrets API.


chainlink:
Nice find! this looks like a serious security issue (especially for hub)

actgardner:
It's not too severe because the API doesn't provide a way to get the plaintext of a secret. You can only list pach secrets, test for the existence of a non-pach k8s secret, and delete any k8s secret the pach SA can delete.

msteffen (approved):

PR #5580, opened 2021-01-06
Require authentication to get the cluster's enterprise token

actgardner:
This was tricky because we need an unauthenticated endpoint to check whether the license is active, so we can decide whether to check auth. We also use send the activation code to Segment, but I've confirmed we only use the name and expiration fields.

For now I split this in two, one is only used by admins for extract/restore and returns the enterprise license data. The other strips the license signature out of the response, so the license can't be replayed to activate enterprise features on a new cluster.

### DIFF
--- a/src/client/enterprise/enterprise.proto
+++ b/src/client/enterprise/enterprise.proto
@@ -51,6 +51,17 @@ enum State {
 message GetStateResponse {
   State state = 1;
   TokenInfo info = 2;
+
+  // activation_code will always be an empty string,
+  // call GetEnterpriseCode to get the activation code
+  string activation_code = 3;
+}
+
+message GetActivationCodeRequest {}
+
+message GetActivationCodeResponse {
+  State state = 1;
+  TokenInfo info = 2;
   string activation_code = 3;
 }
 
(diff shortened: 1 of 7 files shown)


src/client/enterprise/enterprise.proto
```diff
@@ -51,6 +51,17 @@ enum State {
 message GetStateResponse {
   State state = 1;
   TokenInfo info = 2;
+
+  // activation_code will always be an empty string,
+  // call GetEnterpriseCode to get the activation code
+  string activation_code = 3;
```
msteffen, line 57: You could also remove the field by deleting it and adding `reserved: 3;` to the top of the message. I'm not sure if we should avoid doing that in the 1.12 branch for semver (so that client code that reads the field continues to compile? Though if it's no longer being set then that client code would still break, so leaving it actually seems like introducing a silent failure, so maybe we should remove it? I'm honestly not sure...). But either way I think it could make sense to remove and reserve the field in `master`, and I think doing so could allow you to catch any internal calls that read it.

msteffen (approved):
LGTM, with one minor suggestion about the proto

src/client/enterprise/enterprise.proto
actgardner: This is a good point - I didn't think the enterprise code was used anywhere, but we're using it for our Segment metrics 🤦‍♀️ 

I'll talk to the marketing team about how we want to handle this, maybe we can hash the token to get a unique identifier that random people from the internet can't grab for themselves.

actgardner: OK, I've updated this after talking to Dan - now `GetState` strips the signature but leaves in all the fields Marketing uses. In 2.0 I would like to move GetState to an internal service and/or revisit how we identify clusters but I think it's worth having this fix on master and 1.12.x as a stopgap.

PR #5581, opened 2021-01-06
Minor changes to PFS gRPC API (2.0) and readd put file from URL

brycemcanally:


src/client/pfs/pfs.proto
```diff
@@ -307,16 +307,15 @@ enum Delimiter {
   CSV = 4;
 }
 
-message AppendFileRequest {
+message AppendFile {
   bool overwrite = 1;
   string tag = 2;
```
brendoncarroll, line 312: In the storage layer I think you can do a tag per file. Should this be in the sources?

src/client/pfs/pfs.proto
```diff
@@ -351,16 +350,32 @@ message AppendFileRequest {
...
+}
+
+message DeleteFile {
   string file = 1; 
   string tag = 2;
 }
 
-message FileOperationRequest {
+message ModifyFileRequest {
   Commit commit = 1;
```
brendoncarroll, line 375: I think this should go inside as `Commit set_commit = 1`

brendoncarroll (approved):

src/client/pfs/pfs.proto
brycemcanally: An append file modification will correspond to one file, a tar stream, or a url. If you want to add different tags for a file (or set of files), then those should be separate append file modifications.

src/client/pfs/pfs.proto
brycemcanally: The commit and first modification message are both set in the initial modify file request message. Also, I don't think putting that in the same oneof as the modifications makes sense because they are not conceptually related.

PR #5584, opened 2021-01-07
2.0.0 alpha merge

brendoncarroll:
This is the 2.0.0 alpha branch with master merged in from today (2021-01-07).


brendoncarroll:
It would be great to get a some eyes on a couple of areas.  There was some work going on in master that we weren't totally familiar with and we want to make sure we didn't regress any of it.

There were a couple of changes to extract restore which was mostly removed in 2.0.
@msteffen can you take a look at the removed files under `./src/admin/`.

@PFedak there were some changes to `CreatePipelineInTransaction`. Can you review those changes?

We noticed an auth check that was added in master to fsck in the PFS driver.  I think that made it in correctly, but @actgardner can you double check that?

src/client/admin.go
```diff
@@ -19,158 +14,3 @@ func (c APIClient) InspectCluster() (*admin.ClusterInfo, error) {
 	}
 	return clusterInfo, nil
 }
-
-// Extract all cluster state, call f with each operation.
```
actgardner, line 23: Is the plan to remove Extract/Restore and re-implement it from scratch in 2.0? Do we have a list somewhere of all the things we'll need to get feature parity in 2.0?

actgardner:
The fsck auth check LGTM.

src/client/admin.go
brycemcanally: There is not a plan in place for extract / restore yet. We may end up using the 1.x extract / restore as a reference implementation for what ends up going into 2.0, but for now it is easier to just remove it. We have a TODO list in the 2.0 alpha project, which I am going to add more details to after we get this merged in.

PR #5586, opened 2021-01-07
Fileset API

brendoncarroll:
RPCs for associating filesets with commits.


src/client/pfs/pfs.proto
```diff
@@ -161,14 +161,13 @@ message FileInfo {
...
+message CompactionTask {
+  repeated string inputs = 1;
+  PathRange range = 2;
 }
 
-message Shard {
-  Compaction compaction = 1;
-  PathRange range = 2;
-  string output_path = 3;
+message CompactionResult {
```
brycemcanally, line 169: Nit, we should probably call this `CompactionTaskResult`

src/server/pfs/server/api_server.go
```diff
@@ -307,6 +307,7 @@ func (a *apiServer) DeleteBranch(ctx context.Context, request *pfs.DeleteBranchR
 }
 
 func (a *apiServer) ModifyFile(server pfs.API_ModifyFileServer) (retErr error) {
```
brycemcanally, line 309: This is going to be refactored to use the fileset API / functionality under the hood right?

src/server/pkg/storage/fileset/storage.go
```diff
@@ -328,30 +274,60 @@ func (s *Storage) GC(ctx context.Context) error {
...
+	for _, id := range comp.Layers {
+		pointsTo = append(pointsTo, filesetObjectID(id))
+	}
+	if err := s.tracker.CreateObject(ctx, filesetObjectID(id), pointsTo, ttl); err != nil {
+		return nil, err
+	}
+	if err := s.store.Set(ctx, id, md); err != nil {
+		return nil, err
+	}
+	return &id, nil
```
brycemcanally, line 314: Why do we need to return a pointer here? If we can just return the ID, then that seems preferable.

src/server/pfs/server/api_server.go
```diff
@@ -648,3 +524,186 @@ func (a *apiServer) RenewFileset(ctx context.Context, req *pfs.RenewFilesetReque
...
+
+func modifyFile(ctx context.Context, server modifyFileSource, uw *fileset.UnorderedWriter) (int64, error) {
+	var total int64
+	for {
+		req, err := server.Recv()
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				return total, nil
+			}
+			return -1, err
```
brycemcanally, line 536: I think this should still return the bytes read, also I prefer `bytesRead` over `total`.

src/server/pfs/server/commit_store.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+}
+
+func (cs *postgresCommitStore) AddFileset(ctx context.Context, commit *pfs.Commit, id fileset.ID) error {
+	// clone to remove the ttl.
+	id2, err := cs.s.Clone(ctx, id, track.NoTTL)
+	if err != nil {
+		return err
+	}
+	var num int
```
brycemcanally, line 41: I'm assuming this is going to be used for something at some point?

src/server/pfs/server/commit_store.go
```diff
@@ -0,0 +1,145 @@
...
+		RETURNING num
+	`, commit.Repo.Name, commit.ID, *id2); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (cs *postgresCommitStore) GetFileset(ctx context.Context, commit *pfs.Commit) (*fileset.ID, error) {
+	id, err := cs.getTotal(ctx, commit)
+	if err == nil {
```
brycemcanally, line 54: We should change this to check for a specific type of error (probably no rows) then return any other errors. So:
```
if err != nil {
  if errors.Is(err, ...) {
    return cs.s.Clone(ctx, *id, defaultTTL)
  }
  return err
}
```
I could imagine a scenario where we could return an incorrect fileset when a transient error occurs.

src/server/pfs/server/commit_store.go
```diff
@@ -0,0 +1,145 @@
...
+		return cs.s.Clone(ctx, *id, defaultTTL)
+	}
+	ids, err := cs.getDiff(ctx, commit)
```
brycemcanally, line 57: The diff + parent total would be the fileset for the commit. This seems to just return the diff when the total does not exist, so I think we need to grab the parent total and add it to the composition.

src/server/pfs/server/driver.go
```diff
@@ -298,11 +300,8 @@ func (d *driver) listRepo(pachClient *client.APIClient, includeAuth bool) (*pfs.
 }
 
 func (d *driver) deleteRepo(txnCtx *txnenv.TransactionContext, repo *pfs.Repo, force bool) error {
-	if err := d.storage.Store().Walk(txnCtx.ClientContext, repo.Name, func(p string) error {
-		return d.storage.Delete(txnCtx.ClientContext, p)
-	}); err != nil {
-		return err
-	}
+	// TODO: impoement deleteRepo with commitStore
```
brycemcanally, line 303: This misspell got a chuckle out of me. Idempotent?

src/server/pkg/work/work.go
```diff
@@ -267,7 +268,7 @@ func NewWorker(etcdClient *etcd.Client, etcdPrefix string, taskNamespace string)
 }
 
 // ProcessFunc is a callback that is used for processing a subtask in a task.
-type ProcessFunc func(context.Context, *Task) error
+type ProcessFunc func(context.Context, *Task) (*types.Any, error)
```
brycemcanally, line 271: I think we would be better off having one proto for a particular task, and having result / status information as a part of the proto (another field, a oneof with a task / result, or whatever). Each additional proto type introduced in the work package will require a separate collection that needs to be managed.

src/server/pkg/work/work.go
```diff
@@ -363,9 +365,11 @@ func (w *Worker) subtaskFunc(subtaskKey string, processFunc ProcessFunc) subtask
 							}
 							subtaskInfo.Task = subtask
 							subtaskInfo.State = State_SUCCESS
+							subtaskInfo.Result = result
 							if retErr != nil {
 								subtaskInfo.State = State_FAILURE
 								subtaskInfo.Reason = retErr.Error()
+								subtaskInfo.Result = nil
```
brycemcanally, line 372: Lets not nil out the result here because there may be useful information that we want to collect when a failure occurs. Also, the process below should probably look like:
```
var err error
result, err = processFunc(claimCtx, subtask)
return err
```

src/server/pkg/storage/track/tracker.go
```diff
@@ -23,6 +23,9 @@ var (
 	ErrSelfReference = errors.Errorf("object cannot reference itself")
 )
 
+// NoTTL will cause the object to live forever
+const NoTTL = time.Duration(0)
```
brycemcanally, line 27: :+1:

src/server/pkg/storage/fileset/storage.go
```diff
@@ -107,194 +99,148 @@ func (s *Storage) newReader(fileSet string, opts ...index.Option) *Reader {
...
+	for _, id := range ids {
+		md, err := s.store.Get(ctx, id)
+		if err != nil {
 			return nil, err
 		}
+		switch x := md.Value.(type) {
+		case *Metadata_Primitive:
+			fss = append(fss, s.newReader(id, opts...))
+		case *Metadata_Composite:
+			fs, err := s.Open(ctx, x.Composite.Layers)
```
brycemcanally, line 113: I think we should flatten here. Nesting merge readers might just work, but I am not too sure about it and I think we might be better off having the same simplification that we have for composite filesets, which is no nesting.

src/server/pkg/storage/fileset/storage.go
```diff
@@ -107,194 +99,148 @@ func (s *Storage) newReader(fileSet string, opts ...index.Option) *Reader {
...
+	for _, id := range ids {
+		md, err := s.store.Get(ctx, id)
+		if err != nil {
+			return nil, err
+		}
+		switch x := md.Value.(type) {
+		case *Metadata_Primitive:
+			flattened = append(flattened, id)
+		case *Metadata_Composite:
+			ids2, err := s.Flatten(ctx, x.Composite.Layers)
```
brycemcanally, line 170: Are you thinking we should support nesting composite filesets?

src/server/pkg/storage/fileset/storage.go
```diff
@@ -107,194 +99,148 @@ func (s *Storage) newReader(fileSet string, opts ...index.Option) *Reader {
...
-	return s.store.Walk(ctx, fileSet, func(name string) error {
-		oid := filesetObjectID(name)
-		if err := s.store.Delete(ctx, name); err != nil {
-			return err
-		}
-		return s.tracker.MarkTombstone(ctx, oid)
-	})
+// Drop allows a fileset to be deleted if it is not otherwise referenced.
+func (s *Storage) Drop(ctx context.Context, id ID) error {
+	_, err := s.SetTTL(ctx, id, -1)
```
brycemcanally, line 217: Maybe the negative TTL should also be a constant?

src/server/pfs/server/driver_file.go
```diff
@@ -52,64 +52,50 @@ func (d *driver) oneOffModifyFile(ctx context.Context, repo, branch string, cb f
...
-		if err != nil {
-			return "", err
-		}
-		renewer.Add(outputPath)
+	if !compact {
+		renewer.Add(*id)
+		return id, nil
 	}
-	return id, nil
+	compactedID, err := d.storage.Compact(ctx, []fileset.ID{*id}, defaultTTL)
```
brycemcanally, line 90: I think we will be able to remove this if we move to a compaction at read time model, so we would just always return the composite fileset. That doesn't really change anything in this PR, I just wanted to make a note.

src/server/pfs/server/driver_file.go
```diff
@@ -118,15 +104,47 @@ func (d *driver) withWriter(pachClient *client.APIClient, commit *pfs.Commit, cb
...
-	return fsw.Close()
+	return d.commitStore.AddFileset(ctx, commitInfo.Commit, *id)
+}
+
+func (d *driver) getDefaultTag() string {
+	// TODO: change this to a constant like "input" or "default"
+	return fmt.Sprintf("%012d", time.Now().UnixNano())
+}
+
+func (d *driver) openCommit(pachClient *client.APIClient, commit *pfs.Commit, opts ...index.Option) (*pfs.Commit, fileset.FileSet, error) {
```
brycemcanally, line 128: :+1:

src/server/pfs/server/commit_store.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return nil
+}
+
+func (cs *postgresCommitStore) getDiff(ctx context.Context, commit *pfs.Commit) ([]fileset.ID, error) {
+	var ids []fileset.ID
+	if err := cs.db.SelectContext(ctx, &ids,
+		`SELECT fileset_id FROM pfs.commit_diffs
+		WHERE commit_id = $1 AND repo_name = $2
+		ORDER BY num
```
brycemcanally, line 107: :+1:

src/server/pfs/server/driver.go
```diff
@@ -804,59 +804,35 @@ func (d *driver) finishCommit(txnCtx *txnenv.TransactionContext, commit *pfs.Com
...
-			var exists bool
-			if err := d.storage.Store().Walk(m.Ctx(), p, func(_ string) error {
-				exists = true
-				return nil
-			}); err != nil {
-				return false, err
-			}
-			return exists, nil
+	var parentFSID *fileset.ID
+	if commitInfo.ParentCommit != nil {
```
brycemcanally, line 808: Why don't we just do this in the task callback, then just append to the ids when there is a parent?

src/server/pfs/server/driver_compact.go
```diff
@@ -12,188 +11,74 @@ import (
...
-		return nil, err
-	}
-	if err := d.storage.Shard(ctx, fs, func(pathRange *index.PathRange) error {
-		shardOutputPath := path.Join(scratch, strconv.Itoa(len(subtasks)))
-		shard, err := serializeShard(&pfs.Shard{
-			Compaction: compaction,
+func (d *driver) compact(master *work.Master, ids []fileset.ID) (*fileset.ID, error) {
+	// serialize access to RunSubtasks, the compactor may call workerFunc concurrently
+	mu := sync.Mutex{}
+	workerFunc := func(ctx context.Context, task fileset.CompactionTask) (*fileset.ID, error) {
```
brycemcanally, line 22: We probably should change this function to take a set of tasks. Right now we are not processing shards in parallel and we kind of cannot because the work package expects you to not have parallel `RunSubtasks` calls. The expected usage is to provide a batch, or to setup a channel that you eventually close for a batch.

brycemcanally: I'm guessing you set things up like this to make keeping the order of shards simpler?

src/server/pkg/storage/fileset/compaction.go
```diff
@@ -0,0 +1,168 @@
...
+			Inputs:    ids,
+			PathRange: pathRange,
+		})
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	results := make([]ID, len(tasks))
+	eg := errgroup.Group{}
+	for i, task := range tasks {
```
brycemcanally, line 154: If we keep this pattern, you need to add a locally scoped i.

src/server/pkg/storage/fileset/compaction.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+		layers, err := s.getPrimitiveBatch(ctx, ids)
+		if err != nil {
+			return false, err
+		}
+		return isCompacted(s.levelFactor, layers), nil
+	case *Metadata_Primitive:
+		return true, nil
+	default:
+		// TODO: should it be?
```
brycemcanally, line 32: I guess it would be compacted, but we should probably discuss empty filesets more.

src/server/pkg/storage/fileset/compaction.go
```diff
@@ -0,0 +1,168 @@
...
+func isCompacted(factor int64, layers []*Primitive) bool {
+	return indexOfCompacted(factor, layers) == len(layers)
+}
+
+// indexOfCompacted returns the last value of i for which the "compacted relationship"
+// is maintained for all layers[:i+1]
+// the "compacted relationship" is defined as leftSize >= (rightSize * factor)
+// If there is an element at i+1. It will be the first element which does not satisfy
+// the compacted relationship with i.
+func indexOfCompacted(factor int64, inputs []*Primitive) int {
```
brycemcanally, line 58: A concern I brought up about this when you initially proposed it was the duplicate work that can occur when you hit the situation where a small change results with compacting multiple of the larger levels (with this algorithm we would recursively compact them vs compacting them in one pass like the current algorithm). I think the simplicity we get from this outweighs that concern for me, and the extra work is the sum of the exponential decrease of the necessary work in the worst case (I'm not super strong at math, but I think the additional overhead is not going to be that big of a deal). Also, I think we could add the constant compaction delay (don't compact until > x number of primitive filesets) by just wrapping this with a check that counts the primitives before calling into this functionality.

src/server/pkg/storage/fileset/compaction.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	inputs, err := s.getPrimitiveBatch(ctx, ids)
+	if err != nil {
+		return nil, err
+	}
+	i := indexOfCompacted(s.levelFactor, inputs)
+	if i == len(inputs) {
+		return s.Compose(ctx, ids, ttl)
+	}
+	// merge everything from i-1 onward into a single layer
```
brycemcanally, line 84: I think `from i onward` makes more sense in this comment.

src/server/pkg/storage/fileset/compaction.go
```diff
@@ -0,0 +1,168 @@
...
+	merged, err := s.Merge(ctx, ids[i:], ttl)
+	if err != nil {
+		return nil, err
+	}
+	// replace everything from i-1 onward with the merged version
```
brycemcanally, line 89: same as above.

src/server/pfs/server/api_server.go
brendoncarroll: Yeah that's the plan.  I'm running into some problems with how commits are referenced, I think I screwed up branches somehow.  I don't want to refactor until that's passing.  I also just fixed something with commits in transactions which may prevent us from refactoring this too much.

src/server/pfs/server/commit_store.go
brendoncarroll: I think I was logging it. I will remove it.

src/server/pfs/server/commit_store.go
```diff
@@ -0,0 +1,145 @@
...
+	return &id, nil
+}
+
+// SetupPostgresCommitStoreV0 runs SQL to setup the commit store.
+func SetupPostgresCommitStoreV0(ctx context.Context, tx *sqlx.Tx) error {
+	_, err := tx.ExecContext(ctx, `
+		CREATE TABLE pfs.commit_diffs (
+			repo_name VARCHAR(250) NOT NULL,
+			commit_id VARCHAR(64) NOT NULL,
+			num SERIAL NOT NULL,
```
brendoncarroll, line 132: This might need to be BIGSERIAL so we aren't limited to 2^32 calls to AddFileset.

src/server/pfs/server/driver.go
brendoncarroll: implement

src/server/pfs/server/driver.go
brendoncarroll: I didn't want to do another call to postgres in the callback.  It seems like the callback that gets the master is contended?

src/server/pfs/server/driver_compact.go
brendoncarroll: I mostly wanted to express the compaction concurrently in the fileset package.  And then impose the one call constraint here, hence the Mutex.  Is there really that much of a performance benefit to batching these tasks?  I was hoping we could change the work package to allow single calls concurrently.

My prior model for a work coordination package is that tasks are given an identifier and there is essentially a pub-sub channel setup for that identifier, the callers wait for the callee, or timeout.
It's trivial to layer caching on top of that, and that is a great way to get checkpointing, without complicated the algorithm.  There is no state for a job, only work skipping based on cached computations.

src/server/pfs/server/driver_file.go
brendoncarroll: Yeah I agree. I think we need to keep this for the time being to get the reasonable behavior we talked about for temporary filesets and commits.

src/server/pkg/storage/fileset/compaction.go
brendoncarroll: There is no way to create an empty fileset through the API, so for the moment I think an error is more useful than considering it trivially compacted.

src/server/pkg/storage/fileset/compaction.go
brendoncarroll: I'm going to make a ticket for Compaction Optimizations and add it to the 2.0.0 alpha project.  And keep it out of this P.R.

src/server/pkg/storage/fileset/storage.go
brendoncarroll: I'm thinking of making the ID a UUID. Either way I want it to be opaque to the caller. I would rather code outside the fileset package return nil, than have to know IDs are strings at the moment.

src/server/pfs/server/commit_store.go
brendoncarroll: So I fixed this, and now we can read open commits.  But it is potentially expense. We can also read commits with open parents, and grandparents, etc. which is potentially even more expensive.

src/internal/dbutil/tx.go
```diff
@@ -0,0 +1,37 @@
...
+	"database/sql"
+
+	"github.com/jmoiron/sqlx"
+	"github.com/sirupsen/logrus"
+)
+
+// WithTx calls cb with a transaction,
+// The transaction is committed IFF cb returns nil.
+// If cb returns an error the transaction is rolled back.
+func WithTx(ctx context.Context, db *sqlx.DB, cb func(tx *sqlx.Tx) error) (retErr error) {
```
brycemcanally, line 14: :+1:

src/server/pkg/storage/fileset/storage.go
brendoncarroll: Composite filesets do allow nesting.  During compaction we flatten before the merge.  In practice we do flatten everything that gets stored long term.

src/server/pkg/storage/fileset/storage.go
brendoncarroll: Yeah, why not? I think we've talked about always compacting the flatten representation, but why do the extra work flattening if you don't have to.

src/internal/storage/fileset/storage.go
```diff
@@ -107,194 +99,148 @@ func (s *Storage) newReader(fileSet string, opts ...index.Option) *Reader {
...
+	for _, id := range ids {
+		md, err := s.store.Get(ctx, id)
+		if err != nil {
 			return nil, err
 		}
+		switch x := md.Value.(type) {
+		case *Metadata_Primitive:
+			fss = append(fss, s.newReader(id, opts...))
+		case *Metadata_Composite:
+			fs, err := s.Open(ctx, x.Composite.Layers)
```
brycemcanally, line 113: Missing options here.

src/internal/storage/fileset/storage.go
```diff
@@ -328,30 +274,60 @@ func (s *Storage) GC(ctx context.Context) error {
...
+	md := &Metadata{
+		Value: &Metadata_Primitive{
+			Primitive: prim,
+		},
+	}
+	var pointsTo []string
+	for _, chunkID := range prim.PointsTo() {
+		pointsTo = append(pointsTo, chunk.ObjectID(chunkID))
+	}
+	if err := s.tracker.CreateObject(ctx, filesetObjectID(id), pointsTo, ttl); err != nil {
```
brycemcanally, line 288: It would be good if we could get the object creation and store set in the same transaction.

brendoncarroll: I think we can get by without transactions, I have a ticket for a helper function that creates a temporary object, add the auxilary data, then upgrade the object to permanent.   That workflow prevents leaks, and doesn't need transactions.

src/server/pfs/server/driver.go
brendoncarroll: This is now implemented with a note to switch to transactions.

src/server/pfs/server/driver_compact.go
brendoncarroll: Added `CompactionBatchWorker` and switched back to the old model. Left a TODO next to `CompactionWorker` in the fileset package.

src/server/pkg/storage/fileset/compaction.go
brendoncarroll: Removed in favor of `CompactionBatchWorker`

src/internal/storage/fileset/storage.go
brendoncarroll: Okay nevermind. We do need transactions, or at least it is easier with them.  There is a TODO in the driver, and after the tracker supports `sql.Tx` all these spots will be replaced.

src/server/pfs/server/driver_compact.go
```diff
@@ -11,189 +10,82 @@ import (
...
-		if err != nil {
 			return nil, err
 		}
-		if err := fileset.CopyFiles(ctx, fsw, fs, true); err != nil {
-			return nil, err
-		}
-	}
-	if err := fsw.Close(); err != nil {
-		return nil, err
+		fmt.Println(tasks, "->", results)
```
brycemcanally, line 57: Remove.

src/internal/storage/fileset/compaction.go
```diff
@@ -0,0 +1,168 @@
...
+		})
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	results, err := c.workerFunc(ctx, tasks)
+	if err != nil {
+		return nil, err
+	}
+	if len(results) != len(tasks) {
```
brycemcanally, line 164: :+1:

src/internal/storage/track/renewer.go
```diff
@@ -59,8 +59,15 @@ func (r *Renewer) Add(ctx context.Context, id string) error {
 }
 
 // Close stops the background renewal
-func (r *Renewer) Close() error {
-	return r.r.Close()
+func (r *Renewer) Close() (retErr error) {
```
brycemcanally, line 62: The use of `retErr` in this function seems pretty weird. You could accomplish the same thing and be more consistent with our codebase by structuring it like this:
```
defer func() {
  if err := r.r.Close(); retErr == nil {
    retErr = err
  }
}()
ctx := context.Background()
_, err := r.tracker.SetTTLPrefix(ctx, r.id+"/", ExpireNow)
return err
```

src/internal/storage/track/tracker.go
```diff
@@ -114,7 +120,7 @@ func TestTracker(t *testing.T, newTracker func(testing.TB) Tracker) {
 			"DeleteSingleObject",
 			func(t *testing.T, tracker Tracker) {
 				id := "test"
-				require.Nil(t, tracker.CreateObject(ctx, id, []string{}, 0))
+				require.Nil(t, tracker.CreateObject(ctx, id, []string{}, -1))
```
brycemcanally, line 123: Replace with ExpireNow.

src/internal/storage/track/tracker.go
```diff
@@ -124,14 +130,14 @@ func TestTracker(t *testing.T, newTracker func(testing.TB) Tracker) {
 			"ExpireSingleObject",
 			func(t *testing.T, tracker Tracker) {
 				require.Nil(t, tracker.CreateObject(ctx, "keep", []string{}, time.Hour))
-				require.Nil(t, tracker.CreateObject(ctx, "expire", []string{}, time.Microsecond))
-				time.Sleep(time.Millisecond)
+				require.Nil(t, tracker.CreateObject(ctx, "expire", []string{}, -1))
```
brycemcanally, line 133: Same as above.

src/server/pfs/server/driver_file.go
```diff
@@ -52,64 +53,50 @@ func (d *driver) oneOffModifyFile(ctx context.Context, repo, branch string, cb f
...
+	if !compact {
+		renewer.Add(id.HexString())
+		return id, nil
 	}
-	return id, nil
+	compactedID, err := d.storage.Compact(ctx, []fileset.ID{*id}, defaultTTL)
+	if err != nil {
+		return nil, err
+	}
+	renewer.Add(string(*compactedID))
```
brycemcanally, line 95: I believe this should be a call to `HexString`

src/server/pfs/server/driver_file.go
```diff
@@ -118,18 +105,53 @@ func (d *driver) withWriter(pachClient *client.APIClient, commit *pfs.Commit, ta
...
 		return pfsserver.ErrCommitFinished{commitInfo.Commit}
 	}
-	commit = commitInfo.Commit
-	n := d.getSubFileset()
-	subFileSetStr := fileset.SubFileSetStr(n)
-	if tag == "" {
-		tag = subFileSetStr
+	fsw := d.storage.NewWriter(ctx)
+	if err := cb(d.getDefaultTag(), fsw); err != nil {
+		if _, err := fsw.Close(); err != nil {
```
brycemcanally, line 110: I don't think we want to persist the changes if an error occurs. I think this is the case where we need the `Abort` function to just handle the clean up. For now, I think we would be better off making a note of this and not performing a `Close` when an error occurs.

src/internal/storage/track/renewer.go
brendoncarroll: The defer just obscures the control flow.  This just returns the first error, but ensures that both of the calls are made.

brycemcanally: It is more consistent with the rest of the codebase and it is a more common pattern for ensuring the execution of code than what you have. Also, it does ensure execution of both.

brycemcanally (approved):
lgtm

PR #5588, opened 2021-01-08
Ensure pipelines don't lose permissions on output repo.

PFedak:
Build pipelines read and write from different branches of the same repo.
With auth enabled, setting the read and write permission ACLs could
race, possibly leaving the pipeline unable to produce output.


msteffen (approved):
Approved because fixing CI is important and this seems like a reasonable change in and of itself, but I don't totally understand yet how the pipeline gets removed from its output repo. If the pipeline's output repo is in `add` (b/c build pipelines take their output repo as an input), then why isn't it removed from `remove` on line 2071?

Also, I didn't notice this in #5523, but it's unintuitive to me to define `TestPipelineBuildLifecycle` inside `testutil`—the package is a library, but the test (even if it's technically called by other tests) is for Pachyderm rather than for the library that it's in.

An alternative that occurs to me is to move `activateAuth(t)` into `testutil`, then call it from `pps/cmds/cmds_test.go`, so that the `pps` test is running `TestPipelineBuildLifecycle` with auth both on and off. @actgardner moved some of the auth activation functions into `testutil` in https://github.com/pachyderm/pachyderm/pull/5478, so moving `activateAuth(t)` (essentially the CLI version of the same thing) into testutil seems like a natural extension of that change. Let me know what you think, and either way we can do it in a followup PR, but I thought I'd raise the question here.

actgardner:
Based on the CI output this doesn't appear to have fixed the problem, I'm going to take a look this morning - please hold off on merging.

> Also, I didn't notice this in #5523, but it's unintuitive to me to define TestPipelineBuildLifecycle inside testutil—the package is a library, but the test (even if it's technically called by other tests) is for Pachyderm rather than for the library that it's in.

Agree, this should be in one of the test suites and use the utility methods from testutil to setup/teardown the authenticated cluster.

actgardner (approved):
+1 on this to fix CI but I would love to revisit `fixPipelineInputRepoACLsInTransaction` - the parallelism in setting the ACLs is pretty unintuitive and I can see how reading the code one would assume it sets reader on all the inputs _then_ sets writer on the output.

It's also not clear to me that we would ever want to clobber a pipeline's existing permissions on a repo - if the pipeline has OWNER on an input, why would we silently downgrade it to READER? Plus this runs as a superuser and it seems like an avenue for permissions issues to creep in.

PFedak:
> The test is still failing in Circle because the image referenced for that pipeline doesn't exist:
> 
> ```
> default       pod/pipeline-test-pipeline-build19b6051b3a76-build-v1-ppzps   1/2     ImagePullBackOff   0          9m56s
> default       pod/pipeline-test-pipeline-build19b6051b3a76-v1-2hnc6         1/2     ImagePullBackOff   0          9m54s
> ```


Yeah, I realized it's actually failing for separate reasons in circle and travis. The circle issue was actually fine before we went to 2.0.0, now the bake script that builds those images is stale.

PFedak:
I'll definitely make the code-moving changes you both recommend, shouldn't be hard. 

You're right, @msteffen, that currently for build pipelines the delete from `remove` isn't necessary, but if `update pipeline` can change the input arbitrarily (which seems like a rare use case, admittedly) then we could still lose it. As @actgardner said, I was very confused by both the concurrency in this code and the potential for reducing a pipeline's permissions.

actgardner (approved):

PR #5595, opened 2021-01-08
Bump grpc to 1.29

actgardner:
I had to do this to get support for chaining interceptors, I figured I should get it into master so nobody else spends an hour messing with modules. 1.29 is as high as I could go without needing to bump the etcd client, which introduced yet more dependency problems.


go.mod
```diff
@@ -109,14 +109,16 @@ require (
 	sigs.k8s.io/yaml v1.1.0
 )
 
-// Holy shit, the docker library versions are a clusterfuck, see https://github.com/moby/moby/issues/39302
+// Docker library versioning is not straightforward, see https://github.com/moby/moby/issues/39302
```
jrockway, line 112: that's been bugging me for a while :)

go.mod
```diff
@@ -109,14 +109,16 @@ require (
...
 // For the moment, the windows build requires a fix that has not been tagged with an official release
 replace github.com/docker/docker => github.com/docker/docker v1.4.2-0.20191213113251-3452f136aa68
 
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible
 
-replace google.golang.org/grpc => google.golang.org/grpc v1.26.0
+replace google.golang.org/grpc => google.golang.org/grpc v1.29.1
```
jrockway, line 118: Is a replace directive necessary here?  We must depend directly on grpc, so should be able to bump the version that way.

jrockway (approved):
I added the interceptor chaining stuff and never had any problems; not sure what it is I did to not have problems :)

go.mod
actgardner: It seems like the simplest way to pin the dependency given that it's both a direct and indirect dependency, but admittedly I do not know everything about go modules.

jrockway: I can't say for sure without testing, but I believe the version resolution algorithm ensures that only the latest is chosen, so if the main module says 1.29.1, then everything in the program is forced to use that version.  (It changes for v2, though.)

https://golang.org/ref/mod#minimal-version-selection

I would try it and see.  If it doesn't work, submit your change as-is :)

actgardner:
> I added the interceptor chaining stuff and never had any problems; not sure what it is I did to not have problems :)

https://godoc.org/google.golang.org/grpc#ChainUnaryInterceptor is a method introduced in 1.28 that makes it simpler to chain multiple interceptors server-side. It was always possible to write your own chaining but I don't see that in the codebase, I'm not sure what I'm missing.

jrockway:
Makes sense!  I was chaining client interceptors, you're chaining server interceptors.

PR #5596, opened 2021-01-09
Prevent root user admin status from being removed

actgardner:
Make it impossible to change cluster role bindings for the root user, which makes two things much simpler:
1) We no longer need complex logic when modifying role bindings to ensure at least one user is a cluster admin (the root user is always an admin).
2) We no longer need complex logic to find an admin when enabling auth for tests. All tests run on a cluster where `pach:root` is guaranteed to be an admin, and the root token is fixed in tests, so it's trivial to recover from any failed test and reset the cluster to a good state with `DeleteAll`.

This PR mostly simplifies the test utils we provide for testing with auth on:
- `ActivateAuth` can be used to enable auth explicitly if a root client isn't needed. This configures the root token in the pachctl config as well.
- `GetAuthenticatedClient` can be used to get a client authenticated as an arbitrary subject.
- `GetUnauthenticatedClient` can be used to get a client with no auth token, even if a token is configured in the pachctl config.
- `DeleteAll` resets a cluster to the initial state with auth disabled.

We also switch `GetAuthenticatedClient` to use the root client and call `GetAuthToken` for the specified subject, instead of relying on the env var to call `Login` with an arbitrary user. Once we remove Github auth this will no longer be available, we'll also need to switch the cmd tests that rely on `Login` at that time.


msteffen:
Just starting this review now, but both of the improvements you point out seem like big wins. `pach:root` is a huge improvement!

src/server/pkg/testutil/auth.go
```diff
@@ -17,315 +14,60 @@ import (
...
 		tb.Fatalf("could not activate auth service: %v", err.Error())
 	}
-	tokenMap[AdminUser] = resp.PachToken
-	config.WritePachTokenToConfig(resp.PachToken)
+	config.WritePachTokenToConfig(RootToken)
 
 	// Wait for the Pachyderm Auth system to activate
 	require.NoError(tb, backoff.Retry(func() error {
-		if isAuthActive(tb, true) {
+		if isActive, err := client.IsAuthActive(); err != nil {
```
msteffen, line 38: I actually didn't know we had this :joy:. I went and looked it up, and it looks like Yusuf added it for the S3 gateway.

I'm pretty sure I wrote at least some code that checks whether auth is active by calling `WhoAmI`. It's out of scope for the PR, but FWIW there are probably calls to `WhoAmI` that discard the result and should maybe be calling this instead.

msteffen (approved):
LGTM!

Btw, I have to double check with Joey but I'm pretty sure you can expense [these](https://www.amazon.com/Replacement-Individual-Applicable-MacBook-Keyboard/dp/B08PYZGMV2/ref=sr_1_3) if you need one

PR #5597, opened 2021-01-11
Add a basic auth interceptor

actgardner:
This PR adds a GRPC server-side interceptor to enforce basic authorization rules, moving a lot of our authorization logic out of the server itself. There are a few benefits to this:

- The auth interceptor "fails closed" - if a new RPC is added, it must be added to the interceptor or it cannot be called. This prevents the simplest security vulnerability where an endpoint is accessible to unauthenticated remote users.
- It's easier to see at a glance what permissions are required for a given RPC. This isn't fool-proof because an RPC may fan out to other RPCs, and the server may enforce additional checks, but it simplifies the common cases.
- Reduces boilerplate auth code. Every service no longer needs to include boilerplate to get a pach client and check whether a user is authenticated.

In the future I'd like to extend the interceptor to support repo-level permissions as well, which would eliminate most auth checks in PFS and PPS.


src/server/pkg/auth/handler.go
```diff
@@ -0,0 +1,141 @@
...
+type authHandler interface {
+	unary(ctx context.Context, info *grpc.UnaryServerInfo, req interface{}) error
+	stream(ctx context.Context, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error
+	requireAuthEnabled()
+}
+
+func requireAuthEnabled(f authHandlerFn) authHandlerFn {
+	return func(i *Interceptor) authHandler {
+		h := f(i)
+		h.requireAuthEnabled()
```
msteffen, line 23: It seems like, if you wanted, you could remove `requireAuthEnabled` from the `authHandler` interface (and remove it as a field from the other handlers, so that each handler doesn't have to check if auth is enabled), and change the types a bit (this also uses the suggestion in the last comment of consolidating the interface):
```
// Calls 'WhoAmI' and then calls an innerAuthHandler
type authHandler func(context.Context, *grpc.UnaryServerInfo) error

// innerAuthHandler must be wrapped by 'requireAuthEnabled' or 'permitAuthDisabled'
type innerAuthHandler func(context.Context, *grpc.UnaryServerInfo, *auth.WhoAmIResponse) error
type authHandler func(*Interceptor) innerAuthHandler

func newAuthEnabledHandler(required bool, inner innerAuthHandler) authHandler {
	return func(ctx context.Context, info *grpc.UnaryServerInfo) {
		pachClient := h.interceptor.env.GetPachClient(ctx)
		resp, err := pachClient.WhoAmI(pachClient.Ctx(), &auth.WhoAmIRequest{})
		if err != nil {
			if auth.IsErrNotActivated(err) && !h.required {
				return nil
			}
			return err
		}
		return h.inner(ctx, info, resp)
	}
} 

func requireAuthEnabled(f authHandlerFn) authHandlerFn {
	return func(i *Interceptor) authHandler {
		return newAuthEnabledHandler(/* auth required: */ true, f(i))
	}
}

func permitAuthDisabled(f authHandlerFn) authHandlerFn {
	return func(i *Interceptor) authHandler {
		return newAuthEnabledHandler(/* auth required: */ false, f(i))
	}
}
...
...
// The other handlers are mostly reduced to 'isPermitted' (which could probably just be defined inline, if you wanted)
func authenticated(i *Interceptor) innerAuthHandler {
	a := &authenticatedHandler{
		interceptor: i,
	}
	return a.isPermitted
}
...
...
// this needs a wrapper now, but I think that's the only change
"/auth.API/Deactivate":               permitAuthDisabled(adminOnly),
...
...
// This is what the interceptors looks like afterwards:
func (i *Interceptor) InterceptUnary(...) (interface{}, error) {
	...
	if err := a(i)(ctx, info); err != nil {
	...
}

func (i *Interceptor) InterceptStream(...) (interface{}, error) {
	...
	if err := a(i)(ctx, info); err != nil {
	...
}
```

I think that would get you the current behavior, and the other handlers could know whether auth is enabled based on whether `who` is nil (or, if always wrapped in `requireAuthEnabled()`, assume `who` is never nil and assume auth is on).

msteffen (commented):
One high-level comment on the interceptor itself. I wanted to get it out but I'm still going through the rest

src/server/pkg/auth/handler.go
msteffen: as an aside, what do you think of the names `authEnabledAnd` and `authDisabledOr`, and then `allowed`, `authenticated` and `admin` for the inner auth handlers (the idea is `enabled` means we don't need a variadic argumement—there's always an inner handler—and the table looks like:
```
....
"/auth.API/Activate":            authDisabledOr(allowed),
....
"/auth.API/GetOneTimePassword":  authEnabledAnd(authenticated),
...
"/auth.API/Deactivate":          authDisabledOr(admin),
"/auth.API/SetConfiguration":    authEnabledAnd(admin),
...
"/debug.Debug/Profile":          authDisabledOr(admin),
```

It means extra typing in the table, because all of the non-auth APIs become `authDisabledOr`, like:
```
"/pfs.API/CreateRepo":      authDisabledOr(authenticated),
"/pfs.API/InspectRepo":     authDisabledOr(authenticated),
"/pfs.API/ListRepo":        authDisabledOr(authenticated),
"/pfs.API/DeleteRepo":      authDisabledOr(authenticated),
"/pfs.API/StartCommit":     authDisabledOr(authenticated),
...
```

so it's maybe a little more repetitive, but I thought it might read nicely. Anyway, just an idea)

actgardner:
Shortened this up a little bit, I think I took most of your suggestions. Instead of having the outer wrapper types for both cases I opted for `admin` and `authenticated` failing closed when auth is disabled, and the wrappers `authDisabledOr` and `authPartialOr` catching those errors and recovering. I think it's a good trade off of clarity and verbosity.

src/server/pkg/auth/handler.go
```diff
@@ -0,0 +1,73 @@
...
+			return nil
+		}
+		return err
+	}
+}
+
+// authPartialAnd wraps an authHandler and permits the RPC if authHandler succeeds or
+// if auth is partially enabled on the cluster. Handlers only report partial activation
+// if they would otherwise succeed, so this
+func authPartialAnd(h authHandler) authHandler {
```
msteffen, line 25: One more naming idea: I think you could call this `ppsUserOr`.

Basically, the only way for `me.FullyActivated` in `authenticated()` or `admin()` to be false is if the PPS user is calling it during partial activation (otherwise `WhoAmI` returns `ErrNoToken`). AFAICT, if an interceptor doesn't use this, the PPS user is the thing being kept out.

`ppsUserOr(authenticated)` reads a little weirdly becuase the PPS user is authenticated, but maybe that's an implicit implementation detail of `authenticated`

msteffen (commented):
One more thought. Overall LGTM, though I want to do one more pass this afternoon, if that's OK. Thanks again for being a little patient with me here! Just trying to wrap my head around the two versions.

msteffen (approved):
LGTM!

PR #5600, opened 2021-01-12
Reference counting for hashtrees in cache

actgardner:
Add reference counting to the hashtree DBs - instead of calling Destroy on trees that aren't in the cache, and depending on eviction to destroy trees in the cache, now we call Destroy on every hashtree in the pfs driver.

- A tree in the cache that isn't being used in the driver, or a tree that is instantiated outside of the cache, has 1 reference. Calling `Destroy` deletes it immediately.
- When a tree is fetched from the cache we increment the references to 2. When the caller is done with the tree they call `Destroy` and decrement the references back to 1. The tree in the cache remains live.
- When a tree is evicted from the cache we call `Destroy`. If the tree is still in use elsewhere, it will remain live until the caller also calls `Destroy`.

This makes both cases safe - whether the tree is closed by caller first, or whether it's evicted from the cache while in use.

### DIFF
--- a/src/server/pfs/server/driver.go
+++ b/src/server/pfs/server/driver.go
@@ -1365,6 +1365,7 @@ func (d *driver) makeCommit(
 			if err != nil {
 				return nil, err
 			}
+			destroyHashtree(parentTree)
 			defer destroyHashtree(tree)
 			for i, record := range records {
 				if err := d.applyWrite(recordFiles[i], record, tree); err != nil {
@@ -1602,6 +1603,7 @@ func (d *driver) finishCommit(txnCtx *txnenv.TransactionContext, commit *pfs.Com
 		}
 
 		defer func() {
+			destroyHashtree(parentTree)
 			if finishedTree != nil {
 				destroyHashtree(finishedTree)
 			}
@@ -3722,12 +3724,7 @@ func (d *driver) getTreeForFile(pachClient *client.APIClient, file *pfs.File) (h
 			return err
 		}
 		if commitInfo.Finished != nil {
-			t, err := d.getTreeForCommit(txnCtx, file.Commit)
-			if err != nil {
-				return err
-			}
-
-			result, err = t.Copy()
+			result, err = d.getTreeForCommit(txnCtx, file.Commit)
 			return err
 		}
 
--- a/src/server/pfs/server/driver.go
+++ b/src/server/pfs/server/driver.go
@@ -3727,7 +3727,6 @@ func (d *driver) getTreeForFile(pachClient *client.APIClient, file *pfs.File) (h
 			return err
 		}
 		if commitInfo.Finished != nil {
-			fmt.Printf("Finished commit\n")
 			result, err = d.getTreeForCommit(txnCtx, file.Commit)
 			return err
 		}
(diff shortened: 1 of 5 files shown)


src/server/pkg/hashtree/cache.go
```diff
@@ -115,6 +125,7 @@ func (c *Cache) GetOrAdd(key interface{}, generator func() (HashTree, error)) (H
 
 		return nil, err
 	}
+	newValue.GetRef()
```
Tryneus, line 128: I think since the tree comes with a default reference, this isn't needed?

Tryneus (changes requested):
I have some FUD about default references, and a couple nits.

src/server/pkg/hashtree/db.go
```diff
@@ -147,7 +149,10 @@ func newDBHashTree(file string) (HashTree, error) {
 	}); err != nil {
 		return nil, errors.EnsureStack(err)
 	}
-	return &dbHashTree{db}, nil
+	return &dbHashTree{
+		DB:   db,
+		refs: 0,
```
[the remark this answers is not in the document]
actgardner: Changed, now we initialize this value to 1 and compare to 0 references.

src/server/pkg/hashtree/db.go
```diff
@@ -646,8 +651,21 @@ func (h *dbHashTree) Copy() (HashTree, error) {
 	return result, nil
 }
 
-// Destroy cleans up the on disk structures for the hashtree.
+// GetRef attempts to increment the reference count for this
+// db but fails if the tree is already being deleted.
+func (h *dbHashTree) GetRef() bool {
+	refs := atomic.AddInt64(&h.refs, 1)
+	return refs != -1
```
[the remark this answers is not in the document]
actgardner: Yup this was an oversight, I've replaced it with an atomic compare-and-swap in a loop so once the count falls to 0 it never goes back up.

src/server/pkg/hashtree/cache.go
actgardner: The default reference covers the fact that the tree is in the cache, this one covers the fact that it's returned from `GetOrAdd`. Otherwise on the initial cache miss the caller would decrement the refs to 0 and destroy the newly cached tree.

src/server/pkg/hashtree/db.go
```diff
@@ -651,18 +651,26 @@ func (h *dbHashTree) Copy() (HashTree, error) {
...
-	refs := atomic.AddInt64(&h.refs, 1)
-	return refs != -1
+	for {
+		refs := atomic.LoadInt64(&h.refs)
+		if refs > 0 {
+			if atomic.CompareAndSwapInt64(&h.refs, refs, refs+1) {
+				return true
+			}
+		}
+		return false
```
Tryneus, line 665: this should be in an else - otherwise there's no fallthrough if the `CompareAndSwap` fails

Tryneus (approved):
LGTM once CI passes

src/server/pfs/server/driver.go
```diff
@@ -3722,20 +3727,22 @@ func (d *driver) getTreeForFile(pachClient *client.APIClient, file *pfs.File) (h
 			return err
 		}
 		if commitInfo.Finished != nil {
-			t, err := d.getTreeForCommit(txnCtx, file.Commit)
-			if err != nil {
-				return err
-			}
-
-			result, err = t.Copy()
+			fmt.Printf("Finished commit\n")
```
msteffen, line 3730: Sorry for the drive-by comment, but this looks like it might be a stray debug log? (PR seems interesting!)

actgardner: Thanks!

PR #5609, opened 2021-01-13
Performance improvements for reading output repo metadata

### LINKED ISSUE #3315: pachd can overwhelm AWS metadata service when running with IAM role
Reported by Russel Howe in our users channel. After compiling pachd with `aws.Config.CredentialsChainVerboseErrors`, he started seeing logs of the form:
```
errored downloading data after 2.92478ms: NoCredentialProviders: no valid providers in chain
caused by: EnvAccessKeyNotFound: AWS_ACCESS_KEY_ID or AWS_ACCESS_KEY not found in environment
SharedCredsLoad: failed to load shared credentials file
caused by: open /.aws/credentials: no such file or directory
EC2RoleRequestError: no EC2 instance role found
caused by: EC2MetadataError: failed to make EC2Metadata request
caused by: <?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title>429 - Too Many Requests</title>
 </head>
 <body>
  <h1>429 - Too Many Requests</h1>
 </body>
</html>
```

Perhaps there is some client-side credentials caching that we could enable to avoid this issue?

brycemcanally:
This PR contains two performance improvements for reading output repo metadata:
- Setup the object client once during driver setup, rather than once per request (fixes #3315)
- Adds a caching layer above the object storage client (this comes in the form of a cache object storage client). This caching layer is only going to apply to the reading of output repo metadata in the interest of keeping this change simple and because we already have caching that applies to input repo metadata as well as Pachyderm objects. In 2.0, this caching strategy will apply to object storage across the board.

### DIFF
--- a/src/server/pfs/server/driver.go
+++ b/src/server/pfs/server/driver.go
@@ -89,7 +89,8 @@ type CommitStream interface {
 type collectionFactory func(string) col.Collection
 
 type driver struct {
-	env *serviceenv.ServiceEnv
+	env       *serviceenv.ServiceEnv
+	objClient obj.Client
 	// etcdClient and prefix write repo and other metadata to etcd
 	etcdClient *etcd.Client
 	txnEnv     *txnenv.TransactionEnv
@@ -130,10 +131,22 @@ func newDriver(
 		return nil, errors.Errorf("cannot initialize driver with nil treeCache")
 	}
 	// Initialize driver
+	objClient, err := obj.NewClientFromSecret(storageRoot)
+	if err != nil {
+		return nil, err
+	}
+	if env.StorageDiskCacheSize > 0 {
+		diskCache, err := obj.NewLocalClient(filepath.Join(os.TempDir(), "pfs-cache", uuid.NewWithoutDashes()))
+		if err != nil {
+			return nil, err
+		}
+		objClient = obj.NewCacheClient(objClient, diskCache, env.StorageDiskCacheSize)
+	}
 	etcdClient := env.GetEtcdClient()
 	d := &driver{
 		env:            env,
 		txnEnv:         txnEnv,
+		objClient:      objClient,
 		etcdClient:     etcdClient,
 		prefix:         etcdPrefix,
 		repos:          pfsdb.Repos(etcdClient, etcdPrefix),
--- a/src/server/pkg/obj/cache_client.go
+++ b/src/server/pkg/obj/cache_client.go
@@ -2,6 +2,7 @@ package obj
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"sync"
 	"time"
@@ -15,9 +16,8 @@ var _ Client = &cacheClient{}
 type cacheClient struct {
 	slow, fast Client
 
-	mu           sync.Mutex
-	cache        *simplelru.LRU
-	populateOnce sync.Once
+	mu    sync.Mutex
+	cache *simplelru.LRU
 }
 
 // NewCacheClient returns slow wrapped in an LRU write-through cache using fast for storing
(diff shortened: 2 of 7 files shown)


src/server/pfs/server/driver.go
```diff
@@ -130,10 +131,22 @@ func newDriver(
 		return nil, errors.Errorf("cannot initialize driver with nil treeCache")
 	}
 	// Initialize driver
+	objClient, err := obj.NewClientFromSecret(storageRoot)
+	if err != nil {
+		return nil, err
+	}
+	if env.StorageDiskCacheSize > 0 {
+		diskCache, err := obj.NewLocalClient(filepath.Join(os.TempDir(), "pfs-cache", uuid.NewWithoutDashes()))
```
brycemcanally, line 139: @Tryneus Does the tmp directory seem like a reasonable place for this as compared to `StorageRoot` or `CacheRoot`?

src/server/pkg/obj/cache_client.go
```diff
@@ -40,28 +40,52 @@ func NewCacheClient(slow, fast Client, size int) Client {
 }
 
 func (c *cacheClient) Reader(ctx context.Context, p string, offset, size uint64) (io.ReadCloser, error) {
-	c.doPopulateOnce(ctx)
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	if _, exists := c.cache.Get(p); exists {
-		return c.fast.Reader(ctx, p, offset, size)
+	cacheP := cachePath(p, offset, size)
```
brycemcanally, line 45: @Tryneus A change I needed to make to our cache client for 2.0 was to make the support for offset / size actually cache just the offset / size rather than the full object. This is unnecessary for 2.0 since we only operate on chunks with a pre-defined max size, but for 1.x the object can be the full hashtree which may be too big to cache in the local filesystem (particularly since we can have multiple hashtrees in the cache). The change I made is intended to store the path, offset, size as the key in the local cache so we only download and cache the object parts that are relevant to the request (this should work well with the path range indexing that is used for output repos).

src/server/pkg/obj/cache_client.go
```diff
@@ -84,19 +108,6 @@ func (c *cacheClient) IsRetryable(err error) bool {
 	return c.fast.IsRetryable(err) || c.slow.IsRetryable(err)
 }
 
-func (c *cacheClient) deleteFromCache(ctx context.Context, p string) error {
```
brycemcanally, line 87: @Tryneus Stale reads are technically possible (even with this, a different client could perform the deletion), but if we ever get into a situation where that happens then something else went wrong in Pachyderm. I am leaning on the deletion of the commit metadata to prevent this from happening and on the eviction to eventually remove it from the cache.

src/server/pfs/server/driver.go
Tryneus: This seems fine to me

src/server/pkg/testpachd/real_env.go
```diff
@@ -78,6 +78,9 @@ func WithRealEnv(cb func(*RealEnv) error, customConfig ...*serviceenv.PachdFullC
 
 		realEnv.LocalStorageDirectory = path.Join(realEnv.Directory, "localStorage")
 		config.StorageRoot = realEnv.LocalStorageDirectory
+		if err := os.Setenv("STORAGE_BACKEND", "LOCAL"); err != nil {
+			panic(err)
```
Tryneus, line 82: It's better to return the error so the test process can do cleanup (i.e. in defer blocks), but communicating via env vars is pretty bad because it means we can't do tests in parallel if the environment variable ever has to change.

Tryneus (changes requested):
Some concerns and nits, but otherwise LGTM

src/server/pfs/server/driver.go
```diff
@@ -130,10 +131,22 @@ func newDriver(
...
+		if err != nil {
+			return nil, err
+		}
+		objClient = obj.NewCacheClient(objClient, diskCache, env.StorageDiskCacheSize)
```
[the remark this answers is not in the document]
brycemcanally: Yeah, I think this is a fair concern, but I don't think trying to unify the caches in 1.x is going to be worth the effort. Right now we essentially have three PFS caches: object, input repo metadata, and output repo metadata. The object one takes into account size when caching and the input / output repo metadata ones don't. I think this is fine for now since in general the metadata will be much smaller than the data, and in 2.0 we don't need to account for the size of individual chunks because we can bound it based on the max chunk size configuration (which is why `env.StorageDiskCacheSize` is for the number of objects). Also, the output repo metadata objects that are cached in general will be smaller than the input repo ones since it is only caching a subset of metadata (path range) not a full hashtree (boltDB), which is why that default is much higher than the default for the input repo metadata cache.

Tryneus (approved):
LGTM

PR #5610, opened 2021-01-13
Removes Travis CI

brendoncarroll:


brendoncarroll:
@msteffen @lukemarsden There are a couple more references to travis env variables (particularly in the Makefile). e.g. `TRAVIS_BUILD_NUMBER`
I'm sort of hesitant to remove the declarations inline with the Make recipes unless you think it's a good idea?

The travis scripts seemed to be out of the circle call tree, so those are gone.  Circle is passing, so I think that's a safe move.

msteffen:
Overall LGTM. Most of my concerns here are around losing as little test coverage in the transition as possible. In particular, I'd like to compare travis.sh with the circle scripts (I can do it locally) so that anything missing from Circle is at least documented.

Separately, I think we should re-enable all the tests in Circle that weren't identified by the last ci-analytics run.

etc/testing/travis_install.sh
```diff
@@ -1,88 +0,0 @@
-#!/bin/bash
-
-set -ex
-
-# install latest version of docker
```
msteffen, line 5: @lukemarsden how does docker get installed in the testfaster VMs, actually? Is is part of the base VM image they use? Is that based on a public docker image?

etc/testing/travis_install.sh
```diff
@@ -1,88 +0,0 @@
...
-  jq \
-  silversearcher-ag \
-  python3 \
-  python3-pip \
-  python3-setuptools \
-  pkg-config \
-  fuse
-
-# Install fuse
-sudo modprobe fuse
```
msteffen, line 33: @lukemarsden Likewise how does fuse get installed in testfaster?

etc/testing/travis_build_check_and_stash.sh
```diff
@@ -1,87 +0,0 @@
...
-tar cf /tmp/save_git_tarball/pachyderm.tar pachyderm
-cd /tmp/save_git_tarball
-
-cat <<EOT >Dockerfile
-FROM ubuntu:xenial
-COPY pachyderm.tar /
-EOT
-
-docker build -t pachyderm/ci_code_bundle:"${TRAVIS_BUILD_NUMBER}" .
-docker push pachyderm/ci_code_bundle:"${TRAVIS_BUILD_NUMBER}"
```
msteffen, line 87: I...did not realize that we were pushing our entire git repo to docker hub in every Travis run to get around the "outdated commit" issue.

So sad to be getting rid of Travis.

msteffen (approved):
A few questions for @lukemarsden but overall LGTM.

This PR really provides a breathtaking retrospective view of the amount of clowning we tolerated from Travis. I was really looking for stuff that we do in Travis but don't do in Testfaster yet, and everything I found was so dumb.

etc/testing/travis_install.sh
lukemarsden: it's baked in, the contract with testfaster is that it gives you a VM with docker installed and also the version of kubernetes specified in `kubernetes_version` in `.testfaster.yml`

etc/testing/travis_install.sh
lukemarsden: oh, I don't know if it does. what would be failing if it weren't?

etc/testing/travis_build_check_and_stash.sh
lukemarsden: hahahaha
yes, i put that in there

PR #5611, opened 2021-01-13
Fix postgres deployment manifest on windows using incorrect path separators

Tryneus:
The `assets` package was using `filepath` for the disk path, but the manifest needs to always use unix-like separators rather than whichever OS is running `pachctl deploy`.  Changed to use `path` instead, which is always unix-like.


brycemcanally (approved):
lgtm

PR #5613, opened 2021-01-14
[1.12] Spouts use PutFileClient 

actgardner:


src/server/worker/pipeline/common.go
```diff
@@ -84,6 +84,12 @@ func ReceiveSpout(
 				return err
 			}
 			return withSpoutCommit(cancelCtx, pachClient, pipelineInfo, logger, f, func(commit *pfs.Commit, tr *tar.Reader) error {
+				putFileClient, err := pachClient.NewPutFileClient()
+				if err != nil {
+					return err
+				}
+				defer putFileClient.Close()
```
brycemcanally, line 91: We should capture the error state here. I like to use this pattern:
```
defer func() {
  if err := putFileClient.Close(); retErr == nil {
    retErr = err
  }
}()
```

brycemcanally:
We need to get rid of the `withSpoutFile` function since the retry doesn't really make sense when using the `PutFileClient` (we wouldn't be able to know which file needs to be retried). So, I think you would just want to use the tar reader as the source for the put file calls on the `PutFileClient` and get rid of the `withSpoutFile` function wrapping the put file calls and just delete it since it is dead code at that point.

src/server/worker/pipeline/common.go
```diff
@@ -83,7 +83,19 @@ func ReceiveSpout(
...
 				return err
 			}
-			return withSpoutCommit(cancelCtx, pachClient, pipelineInfo, logger, f, func(commit *pfs.Commit, tr *tar.Reader) error {
+			return withSpoutCommit(cancelCtx, pachClient, pipelineInfo, logger, f, func(commit *pfs.Commit, tr *tar.Reader) (retErr error) {
+				putFileClient, err := pachClient.NewPutFileClient()
+				if err != nil {
+					return err
+				}
+				defer func() {
+					if err := putFileClient.Close(); err != nil {
```
brycemcanally, line 92: This is going to overwrite the original error state if an error occurs. We should only set `retErr` if it is nil, which is why I like the pattern where the only check for determining whether to set `retErr` is `retErr == nil` since it is simple and always correct.

brycemcanally (approved):
lgtm

PR #5618, opened 2021-01-15
Remove Github authentication

actgardner:
In 2.0 our plan is to only support OIDC authentication in pachd, and handle Github/SAML/LDAP/etc. with the embedded Dex identity server. Github authentication is currently tightly coupled to our auth system, so this PR just aims to remove Github auth and make all tests work with OIDC or, in most cases, using robot users with no auth provider configured at all.

One major user-facing consequence of this change is that we no longer qualify usernames for users, we just throw an error. A user has to explicitly specify `<idp>:<username>` for every API call, and the username is case sensitive. We only ever supported fixing up usernames for Github anyways, so this isn't much of a regression, and in general automatically changing user inputs for role bindings feels like a potential footgun. 

Once we add a mapping of "user -> tokens" to support revoking users acccess we can leverage that to list users on a cluster, and potentially to check whether the user in a role binding exists.


src/client/auth/auth.proto
```diff
@@ -18,8 +18,8 @@ import "google/protobuf/timestamp.proto";
  * ListAdmins) will have their own conventions
  *
  * The current user formats are:
- * 1) GitHub usernames:
- *      "github:MyGitHubUsername"
+ * 1) Users synced from an identity provider:
+ *      "<idp>:username"
```
actgardner, line 22: I think long-term we won't need to support multiple IDPs in pachd at all, so we may even be able to simplify this to just "user:<username>".

src/server/auth/server/testing/oidc_test.go
```diff
@@ -4,143 +4,29 @@ import (
...
 
-	"github.com/pachyderm/pachyderm/src/client"
 	"github.com/pachyderm/pachyderm/src/client/auth"
-	"github.com/pachyderm/pachyderm/src/client/identity"
 	"github.com/pachyderm/pachyderm/src/client/pkg/require"
-	"github.com/pachyderm/pachyderm/src/server/pkg/backoff"
 	tu "github.com/pachyderm/pachyderm/src/server/pkg/testutil"
 )
 
-// Dex's mock connector always returns the same identity for all authentication
```
actgardner, line 20: Refactored this code into testutils so we can enable OIDC auth for the CLI tests as well.

src/server/auth/cmds/cmds_test.go
```diff
@@ -142,52 +107,50 @@ func TestAdmins(t *testing.T) {
...
 		pachctl auth list-admins \
-			| match -v "^github:admin$" \
-			| match "^github:admin2$"
+			| match -v "^robot:admin$" \
+			| match "^robot:admin2$"
 		`).Run())
 
 	// Now 'admin2' is the only admin. Login as admin2, and swap 'admin' back in
-	// (so that deactivateAuth() runs), and call 'list-admin' (to make sure it
+	// (so that detu.ActivateAuth() runs), and call 'list-admin' (to make sure it
```
msteffen, line 130: `detu` I think is a typo/find-and-replace mangling?

src/server/auth/cmds/cmds_test.go
```diff
@@ -351,23 +315,6 @@ func TestGetOneTimePasswordTTL(t *testing.T) {
...
-	if testing.Short() {
-		t.Skip("Skipping integration tests in short mode")
-	}
-	activateAuth(t)
-	defer deactivateAuth(t)
-
-	require.NoError(t, tu.BashCmd(`
-		pachctl auth get-config -o yaml \
-		  | match 'live_config_version: 1' \
-		  | match 'id_providers:' \
```
msteffen, line 364: Do you think it would make sense to keep this test but just delete the github part? I suppose this isn't like our most exhaustive test of yaml rendering, but it makes me a little anxious to remove it completely because when I've changed our yaml renderer in the past (it's wierdly hard to render and parse yaml in Go, particularly to and from protobufs because they don't generate the struct tags you need), it tended to break obscure commands.

src/server/auth/server/api_server.go
```diff
@@ -29,19 +27,11 @@ import (
...
 	"golang.org/x/net/context"
-	"golang.org/x/oauth2"
-	"golang.org/x/sync/errgroup"
 )
 
 const (
-	// DisableAuthenticationEnvVar specifies an environment variable that, if set, causes
-	// Pachyderm authentication to ignore github and authmatically generate a
-	// pachyderm token for any username in the AuthenticateRequest.GitHubToken field
-	DisableAuthenticationEnvVar = "PACHYDERM_AUTHENTICATION_DISABLED_FOR_TESTING"
```
msteffen, line 43: :pray:

src/server/auth/server/api_server.go
```diff
@@ -1787,19 +1677,11 @@ func (a *apiServer) SetACLInTransaction(
 		if user == ppsUser {
 			continue
 		}
-		eg.Go(func() error {
```
msteffen, line 1790: This is way better now

src/server/auth/server/testing/auth_test.go
```diff
@@ -2876,135 +2871,25 @@ func TestDeletePipelineMissingRepos(t *testing.T) {
 	})
 }
 
-func TestDisableGitHubAuth(t *testing.T) {
```
msteffen, line 2879: :boom:

msteffen (approved):
LGTM! All the comments are questions or little nits about comments and stuff

src/server/pkg/testutil/identity.go
```diff
@@ -0,0 +1,166 @@
...
+	require.NoError(t, err)
+
+	_, err = adminClient.CreateOIDCClient(adminClient.Ctx(), &identity.CreateOIDCClientRequest{
+		Client: &identity.OIDCClient{
+			Id:           "testapp",
+			Name:         "testapp",
+			RedirectUris: []string{"http://test.example.com:657/authorization-code/callback"},
+			Secret:       "test",
+		},
+	})
```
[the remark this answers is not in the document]
actgardner: It's used to test the TrustedPeers functionality where you get an ID token from one client (testapp), but it includes another audience (pachd). We only use it in one test but it's very cheap to set up.

src/server/auth/cmds/cmds_test.go
actgardner: I moved some YAML testing into `TestConfig` so we check the JSON and YAML formats there.

PR #5621, opened 2021-01-18
Remove the partial auth activation state

actgardner:
The `auth.Activate` RPC used to follow this process: 
- enable partial auth, which locks out all users except the PPS token
- iterate over all existing PPS pipelines, set permissions and generate auth tokens for the pipelines by calling `pps.ActivateAuth`
- enable full auth, make the root user an admin and return the root token to the user

The partial auth state was designed to avoid a user rapidly calling `Authenticate` and `CreatePipeline` on a cluster during `pps.ActivateAuth`. If this happened the newly created pipeline would not have an auth token and would fail to start. 

In 2.0 there are no authentication methods configured by default  - a user can't call `Authenticate` until the root user configures auth. This means the cluster is locked from the time `auth.Activate` starts. 

This PR removes the partial activation state and removes the call to `pps.ActivateAuth` from `auth.Activate`. Instead, the user who activates auth on the cluster will get the root token and use that token to call `ActivateAuth`. This has a few benefits:
- simplifies auth checks, since auth state is now binary instead of ternary
- removes latency - previously `auth.Activate` had 2 one second sleep calls to wait for the admin cache to converge
- removes the dependency on PPS from `auth.Activate` - in the standalone Enterprise Server PPS won't be running so this would be unnecessary

### DIFF
--- a/src/client/auth/auth.go
+++ b/src/client/auth/auth.go
@@ -62,10 +62,9 @@ var (
 	// it also needs to be updated in the UI code
 	ErrNotActivated = status.Error(codes.Unimplemented, "the auth service is not activated")
 
-	// ErrPartiallyActivated is returned by the auth API to indicated that it's
-	// in an intermediate state (in this state, users can retry Activate() or
-	// revert with Deactivate(), but not much else)
-	ErrPartiallyActivated = status.Error(codes.Unavailable, "the auth service is only partially activated")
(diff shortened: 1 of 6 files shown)


msteffen (approved):
LGTM. Just to note it, I think we'll need to add the requirement to call `pps.ActivateAuth()` to our docs, too (probably `doc/docs/master/enterprise/auth/enable-auth.md`) so the language client users know to do it.

PR #5623, opened 2021-01-19
Refactor license parsing and caching out of enterprise server

actgardner:
This PR lays some groundwork for the license server by extracting functionality that will be used in both the license server and enterprise server:
- parsing and validating activation codes now lives in `server/pkg/license`
- watching and cacheing the value of a single frequently read, infrequently written key is in `server/pkg/keycache`

`keycache` will also be useful in the auth server refactor to cache the OIDC config, and to cache role bindings.


jdoliner (approved):
LGTM

PR #5631, opened 2021-01-21
Move migrations out of PFS

actgardner:
We want to move to Postgres as a datastore for all services in pachd, but currently the migration code runs in the PFS master. This moves the definition of all the migrations to `src/server/pkg/clusterstate` and moves the database migration logic to the pachd main so it runs before any services are started.

Right now every pachd will attempt to run the migrations and block on acquiring a lock on the migrations table. This seems fine for now but long-term we may want to move this logic to the deploy process instead.

### DIFF
--- a/src/server/cmd/pachd/main.go
+++ b/src/server/cmd/pachd/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/pachyderm/pachyderm/src/server/pkg/clusterstate"
 	"github.com/pachyderm/pachyderm/src/server/pkg/cmdutil"
 	col "github.com/pachyderm/pachyderm/src/server/pkg/collection"
+	"github.com/pachyderm/pachyderm/src/server/pkg/dbutil"
 	"github.com/pachyderm/pachyderm/src/server/pkg/deploy/assets"
 	logutil "github.com/pachyderm/pachyderm/src/server/pkg/log"
 	"github.com/pachyderm/pachyderm/src/server/pkg/metrics"
@@ -292,7 +293,7 @@ func doFullMode(config interface{}) (retErr error) {
 
 	identityStorageProvider := identity_server.NewLazyPostgresStorage(
 		env.PostgresServiceHost,
-		env.IdentityServerDatabase,
+		dbutil.DefaultDBName,
 		env.IdentityServerUser,
 		env.IdentityServerPassword,
 		env.PostgresServiceSSL,
--- a/src/server/cmd/pachd/main.go
+++ b/src/server/cmd/pachd/main.go
@@ -30,11 +30,13 @@ import (
 	"github.com/pachyderm/pachyderm/src/server/health"
 	identity_server "github.com/pachyderm/pachyderm/src/server/identity/server"
 	pfs_server "github.com/pachyderm/pachyderm/src/server/pfs/server"
+	"github.com/pachyderm/pachyderm/src/server/pkg/clusterstate"
 	"github.com/pachyderm/pachyderm/src/server/pkg/cmdutil"
 	col "github.com/pachyderm/pachyderm/src/server/pkg/collection"
 	"github.com/pachyderm/pachyderm/src/server/pkg/deploy/assets"
 	logutil "github.com/pachyderm/pachyderm/src/server/pkg/log"
 	"github.com/pachyderm/pachyderm/src/server/pkg/metrics"
+	"github.com/pachyderm/pachyderm/src/server/pkg/migrations"
 	"github.com/pachyderm/pachyderm/src/server/pkg/netutil"
 	"github.com/pachyderm/pachyderm/src/server/pkg/serviceenv"
 	txnenv "github.com/pachyderm/pachyderm/src/server/pkg/transactionenv"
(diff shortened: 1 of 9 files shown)


src/server/pkg/deploy/assets/postgres.go
```diff
@@ -347,26 +329,3 @@ func PostgresService(local bool, opts *AssetOpts) *v1.Service {
...
-		},
-		ObjectMeta: objectMeta(postgresInitConfigMapName, labels(postgresName), nil, opts.Namespace),
-		Data: map[string]string{
-			"init-db.sh": `
-#!/bin/bash
-set -e
-
-psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-EOSQL
-    CREATE DATABASE dex;
-    GRANT ALL PRIVILEGES ON DATABASE dex TO postgres;
```
actgardner, line 367: This was a hacky solution to set up the dex schema - dex does it's own migrations.

src/server/pkg/migrations/migrations.go
```diff
@@ -68,28 +68,44 @@ func InitialState() State {
...
-			return err
+			return nil
 		},
 	}
 }
 
 // ApplyMigrations does the necessary work to actualize state.
 // It will manipulate the objects available in baseEnv, and use the migrations table in db.
 func ApplyMigrations(ctx context.Context, db *sqlx.DB, baseEnv Env, state State) error {
+	tx, err := db.BeginTxx(ctx, &sql.TxOptions{})
```
actgardner, line 79: Previously we ran each migration step in it's own transaction. This runs this risk of leaving a customer database in a weird in-between state where they can't run a previous release or the current release. It also meant we needed to re-acquire the lock on the migrations table for each step in the migration, which would make for confusing logs in pachd if multiple instances were competing to make progress.

brendoncarroll: More complicated migrations involving things other than the database might benefit from having checkpoints.  We can't do that if everything rolls back.  If we are shuffling things around in object storage or doing ETL to populate a table we would probably do it in steps to avoid losing hours of work.

Confusing logs, but otherwise correct behavior doesn't seem like an issue considering this is a transition step to something better like a single container executing the migration while everyone blocks.

This seems orthogonal to rollback, which can also be done state by state.  And I totally agree that we need a way to rollback.

src/server/pkg/testpachd/real_env.go
```diff
@@ -55,6 +58,13 @@ func WithRealEnv(db *sqlx.DB, cb func(*RealEnv) error, customConfig ...*servicee
 		config.PeerPort = uint16(realEnv.MockPachd.Addr.(*net.TCPAddr).Port)
 		servEnv := serviceenv.InitServiceEnv(config)
 
+		if err := migrations.ApplyMigrations(context.Background(), db, migrations.Env{}, clusterstate.DesiredClusterState); err != nil {
```
brendoncarroll, line 61: 👍

src/server/pkg/migrations/migrations.go
```diff
@@ -68,28 +68,44 @@ func InitialState() State {
...
+		return err
+	}
+	logrus.Infof("acquired lock on migrations table")
+
 	for _, state := range collectStates(make([]State, 0, state.n+1), state) {
-		if err := applyMigration(ctx, db, baseEnv, state); err != nil {
-			return err
+		if err := applyMigration(ctx, tx, env, state); err != nil {
+			logrus.Error(err)
+			return tx.Rollback()
```
brendoncarroll, line 105: I like doing the rollback here, it does seem cleaner.

src/server/cmd/pachd/main.go
```diff
@@ -256,6 +259,15 @@ func doFullMode(config interface{}) (retErr error) {
 	if env.EtcdPrefix == "" {
 		env.EtcdPrefix = col.DefaultPrefix
 	}
+
+	// TODO: currently all pachds attempt to apply migrations, we should coordinate this
```
brendoncarroll, line 263: 👍

src/server/pkg/migrations/migrations.go
```diff
@@ -68,28 +68,44 @@ func InitialState() State {
...
 func ApplyMigrations(ctx context.Context, db *sqlx.DB, baseEnv Env, state State) error {
+	tx, err := db.BeginTxx(ctx, &sql.TxOptions{})
+	if err != nil {
+		return err
+	}
+
+	env := baseEnv
+	env.Tx = tx
+
+	if _, err := tx.ExecContext(ctx, `CREATE TABLE IF NOT EXISTS migrations (
```
brendoncarroll, line 87: I'm sort of indifferent as to whether it's easier to special case the creation of the table here or in the initial state change.  We need to special case isFinished of state 0 to check for the existence of the table, and special case the execution of state 0 to create the table, without the usual locking.

So this shouldn't be executed during every call to ApplyMigration, it should get skipped by isFinished in the common case.

I want to avoid special-casing state 0 to callers of this API.  It should represent whether the migrations table has been created. If it has been actualized, then the table exists, if not then it hasn't.  We shouldn't have behavior where InitialState is a thing you have to call to use the API, but the state you get back is actually totally meaningless because the creation of the table doesn't correspond to state 0's actualization.

src/server/pkg/migrations/migrations.go
brendoncarroll: Although looking at it again, we were returning the proximate error before, and now we are returning the error from rollback.

I could be wrong here, but my understanding is that this would return nil for a successful rollback?  I think we want to return the error that caused the rollback, and log whether there was a problem rolling back.

brendoncarroll:
I think we were running the migration in the PFS master (`master.go`).  That can be removed, and the blocking call in the constructor`driver.go` can go as well.  The change you made to WithRealEnv should take care of it.

actgardner:
I've reverted everything but the bare minimum back to the way it was, should be good for another pass once CI is green.

brendoncarroll (approved):
LGTM

PR #5633, opened 2021-01-21
Remove SAML and simplify auth config

actgardner:
This PR removes support for any auth providers except a single OIDC configuration (which will always be an identity server, either the embedded one or a remote one). Since we no longer support multiple IDPs, we can refer to all users as `user:<email>` and groups as `group:<group>`. If a user needs to support multiple IDPs they can configure them in Dex.

This also simplifies the way we configure the OIDC server - instead of copying the configuration values multiple times, we validate them in `SetConfiguration` and use `keycache` to cache the `auth.AuthConfig` struct. The OIDC methods are now just attached to the `apiServer` struct, and they load the cached config as needed.

One outstanding TODO item is to cache the `oidc.Provider` to avoid doing multiple round-trips to the identity server when a user logs in. User log ins are relatively rare and not latency-sensitive so I'm punting on this for now.


src/server/auth/server/oidc.go
```diff
@@ -118,6 +63,31 @@ func CryptoString(n int) string {
 	return base64.RawURLEncoding.EncodeToString(b)
 }
 
+// validateOIDC validates an OIDC configuration before it's stored in etcd.
+func validateOIDCConfig(config *auth.AuthConfig) error {
```
actgardner, line 67: This was the only validation we were doing in `config.go` for OIDC, so I pulled it out here.

src/server/auth/server/testing/config_test.go
```diff
@@ -1,813 +0,0 @@
...
-	for _, idp := range a.IDProviders {
-		formatIDP(idp)
-	}
-	require.True(t, proto.Equal(e, a),
-		"expected: %v\n           actual: %v", e, a)
-}
-
-// TestSetGetConfigBasic sets an auth config and then retrieves it, to make
-// sure it's stored propertly
-func TestSetGetConfigBasic(t *testing.T) {
```
msteffen, line 46: IIUC, we'll still have a `SetConfiguration` endpoint—would it make sense to at least keep around this test for it? (or is there like another test elsewhere?)

src/server/auth/server/oidc.go
```diff
@@ -118,6 +63,31 @@ func CryptoString(n int) string {
...
+	if _, err := url.Parse(config.Issuer); err != nil {
+		return errors.Wrapf(err, "OIDC issuer must be a valid URL")
+	}
+
+	// this does a request to <issuer>/.well-known/openid-configuration to see if it works
+	ctx := context.Background()
+	_, err := oidcProvider(ctx, config)
```
msteffen, line 74: I know this is in the OIDC library, but do you know offhand if this potentially blocks (or leaks the goroutine) if the issuer is unavailable?

src/server/auth/server/oidc.go
```diff
@@ -136,75 +106,56 @@ func half(state string) string {
...
-			nil,
-			nil,
-		),
-	}
-	var err error
-	ctx := context.Background()
-	if localhostIssuer {
-		client, err := LocalhostRewriteClient(issuer)
+// oidcProvider creates a short-lived oidc.Provider tied to a request context.
+// TOOD: we should cache and reuse the provider as long as the config hasn't changed.
```
msteffen, line 110: typo (should be DOOT)

src/server/auth/server/oidc.go
```diff
@@ -384,42 +340,53 @@ func (o *InternalOIDCProvider) validateIDToken(ctx context.Context, rawIDToken s
...
-	return o.a.setGroupsForUserInternal(ctx, claims.Email, groups)
+	return a.setGroupsForUserInternal(ctx, claims.Email, groups)
 }
 
 // handleOIDCExchangeInternal is a convenience function for converting an
 // authorization code into an access token. The caller (handleOIDCExchange) is
 // responsible for storing any responses from this in etcd and sending an HTTP
 // response to the user's browser.
-func (a *apiServer) handleOIDCExchangeInternal(ctx context.Context, sp *InternalOIDCProvider, authCode, state string) (nonce, email string, retErr error) {
+func (a *apiServer) handleOIDCExchangeInternal(ctx context.Context, authCode, state string) (nonce, email string, retErr error) {
```
msteffen, line 362: Maybe out of scope for this PR, but I just noticed that `state` isn't used and can be removed (or the etcd write could be moved into this function, which is kind of how I expected it to work :shrug:)

msteffen (approved):
LGTM. Just few small suggestions

src/client/auth/auth.proto
```diff
@@ -47,106 +47,20 @@ message ActivateResponse {
...
-     // OIDC requests to the embedded OIDC provider. This is necessary to support
-     // some network configurations like Minikube. 
-     bool localhost_issuer = 7;
-  }
-  OIDCOptions oidc = 5 [(gogoproto.customname) = "OIDC"];
-}
 
-// Configure Pachyderm's auth system (particularly authentication backends
+// Configure Pachyderm's auth system with an OIDC provider 
 message AuthConfig {
```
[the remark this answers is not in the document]
actgardner: Renamed to `OIDCConfig` since that's what it really is.

src/server/auth/server/api_server.go
```diff
@@ -196,6 +175,15 @@ func NewAuthServer(
 	public bool,
 	requireNoncriticalServers bool,
 ) (APIServer, error) {
+
+	authConfig := col.NewCollection(
```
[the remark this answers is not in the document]
actgardner: We need it for the keycache and the field in the struct, this was a little less verbose.

src/server/auth/server/oidc.go
actgardner: Updated this to accept the incoming ctx from `SetConfiguration` so it can be cancelled. AFAICT there's no risk of a goroutine leaking from this, it looks like `NewProvider` fetches the well-known URI during the constructor using the provider context, and the JWKS library only makes a request if you call Verify for an un-cached kid (https://github.com/coreos/go-oidc/blob/v3.0.0/oidc/jwks.go#L118)

src/server/auth/server/oidc.go
actgardner: That method is pretty complicated, I'll make a note to try and simplify it in a future PR so it can get the appropriate level of attention.

src/server/auth/server/testing/config_test.go
actgardner: Reinstated `config_test.go` and removed all but the few tests that were still relevant.

PR #5641, opened 2021-01-22
set gopath using go env

brendoncarroll:


nitinjainsj (approved):
Another change "-f" in git push. We made that change in master already.

LGTM

PR #5649, opened 2021-01-27
fix) examples.kubeflow.pipeline.pipeline_id bug

crescent98:
The original code had matters because the variable p gets two roles. (one for pipeline lists, one for a single pipeline)
Based on this issue, fixed one of the p (for the list) to p_list.


JoeyZwicker:
Thanks for much @crescent98!

msteffen (approved):
Thanks for the fix, @crescent98!

crescent98:
> Thanks for much @crescent98!

No prob, if there's no conflicts, can I get it merged?

PR #5652, opened 2021-01-27
Reduce the number of test buckets

actgardner:
We've removed a lot of auth tests, and identity tests run quickly, so combine them into one bucket. Also reduce the PPS buckets, to try and make all buckets complete in about the same amount of time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,8 +19,7 @@ jobs:
         type: string
 
     environment:
-      PPS_BUCKETS: "6"
-      AUTH_BUCKETS: "2"
+      PPS_BUCKETS: "3"
       GOPROXY: https://proxy.golang.org
       BUCKET: << parameters.bucket >>
       RUN_BAD_TESTS: << pipeline.parameters.run_flaky_tests >>
@@ -97,18 +96,13 @@ workflows:
             parameters:
               bucket:
               - MISC
-              # If you want to update the number of PPS or auth buckets, you'll neet to
-              # update the value of PPS_BUCKETS or AUTH_BUCKETS above
+              # If you want to update the number of PPS buckets, you'll neet to
+              # update the value of PPS_BUCKETS above
               - ADMIN
-              - AUTH1
-              - AUTH2
-              - IDENTITY
+              - AUTH
               - PFS
               - PPS1
               - PPS2
               - PPS3
-              - PPS4
-              - PPS5
-              - PPS6
               - EXAMPLES
               - OBJECT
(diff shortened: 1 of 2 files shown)


lukegrid:
Nice!

build-PPS3 - 20m 27s
build-PPS2 - 23m 58s
build-PPS1 - 18m 36s
build-PFS - 14m 29s
build-OBJECT - 13m 29s
build-MISC - 20m 50s
build-EXAMPLES - 12m 48s
build-AUTH - 21m 0s
build-ADMIN - 11m 11s

PPS2 is now the longest test bucket, I wonder if we should experiment with bumping that 3 (down from 6) back up to 4?

lukemarsden:
Nice!

build-PPS3 - 20m 27s
build-PPS2 - 23m 58s
build-PPS1 - 18m 36s
build-PFS - 14m 29s
build-OBJECT - 13m 29s
build-MISC - 20m 50s
build-EXAMPLES - 12m 48s
build-AUTH - 21m 0s
build-ADMIN - 11m 11s

PPS2 is now the longest test bucket, I wonder if we should experiment with bumping that 3 (down from 6) back up to 4?

I should be able to reduce queuing in Testfaster...

lukemarsden (changes requested):
Nice!

```
build-PPS3 - 20m 27s
build-PPS2 - 23m 58s
build-PPS1 - 18m 36s
build-PFS - 14m 29s
build-OBJECT - 13m 29s
build-MISC - 20m 50s
build-EXAMPLES - 12m 48s
build-AUTH - 21m 0s
build-ADMIN - 11m 11s
```

PPS2 is now the longest test bucket, I wonder if we should experiment with bumping that 3 (down from 6) back up to 4?

I should separately be able to reduce queuing (increase parallelism) in Testfaster, once I've got that working it will make sense to _increase_ the number of test buckets, but of course care should be taken to only split up the longest ones :)

actgardner:
I've bumped the PPS buckets to 4. 

> I should separately be able to reduce queuing (increase parallelism) in Testfaster, once I've got that working it will make sense to increase the number of test buckets

Isn't the number of concurrent jobs limited by the amount of RAM on the test box? Because we need a minikube + pachd + etcd + postgres per bucket? I know if the startup cost for a new VM was 0 we would just want to run every test in a separate parallel VM to get the best utilization, but it seems like there's a non-zero startup latency _and_ some kind of scheduling preference that doesn't just run each bucket as a FIFO?

lukemarsden (approved):

PR #5659, opened 2021-02-01
Make `make lint` faster

actgardner:
I call `make lint` pretty frequently while iterating on code, and it takes ~15s to run each time. This PR gets it down to 6 seconds, which is still not great but it's almost exclusively spent in `staticcheck`. The big improvement is not calling `go get` for golint and staticcheck if they exist on the path and they're less than a day old, but I also added some xargs parallelism.


jdoliner (approved):
LGTM

PR #5349, opened 2020-10-14
Double stats job file

brendoncarroll:
This includes a test, the additive merge workaround from @Tryneus and an improved error message should this come up again.

PR #5669, opened 2021-02-02
Add files to stats commits for ListDatum.

### LINKED ISSUE #5386: Cannot find job file in ListDatum
I believe @Tryneus and @brendoncarroll have the most context, but IIUC if a job with an additive workload runs, it may merge its output into its parent commit's hashtree. However, this is not implemented correctly for stats.

There is a proposed fix in https://github.com/pachyderm/pachyderm/pull/5349, however we're concerned that disabling the additive workload optimization will be too costly for users, so we're looking for alternative solutions

PFedak:
Fix #5386 (or at least, new instances, I've left out the code I was working on to try and disambiguate the concatenated data) by adding the job ID to files in stats commits. The old files will still be written to avoid breaking any other dependencies.

### DIFF
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -6797,6 +6797,151 @@ func TestPipelineWithStats(t *testing.T) {
 	require.Equal(t, pps.DatumState_SUCCESS, datum.State)
 }
 
+func TestPipelineWithStatsError(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration tests in short mode")
+	}
+
+	c := tu.GetPachClient(t)
+	require.NoError(t, c.DeleteAll())
+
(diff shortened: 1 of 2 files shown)


PFedak:
This isn't enough right now, the fundamental issue is that stats for skipped datums get merged with the copy of themselves already in the parent. Either we need the stats hashtree to overwrite the parent (with special merge logic) or different handling of stats data for skipped datums. Leaning towards the second for now since it will involve less mucking around with hashtree internals.

PFedak:
Ok, now I think this is good to go. I'm writing a test which exercises at least the basic failed-datum-which-later-succeeds case, happy to hear any other suggestions. Should probably include the test from #5349.

PFedak:
Nevermind, still doomed, doesn't seem like there's a way to make this approach work without directly checking the parent tree. Not sure that's preferable to adding new merge logic.

PR #5671, opened 2021-02-03
[ci skip]Doc reference refresh script closes #4804

### LINKED ISSUE #4804: Create separate `make doc` targets
We need to have separate `make doc` targets for building docs for patches and major/minor releases.

For patches, we can have something like make doc-patch that will only generate the `pachctl` reference and output the results to the `doc/docs/master/reference/pachctl/` and `doc/docs/minor-version/reference/pachctl/` directories. All other docs should not be touched. This needs to be run in the master branch because it is the only branch that we are publishing on the docs website.

For minor/major version, we need to run the whole script that is being run at the moment as part of the release process (updates pachctl and some of the docs).

**Bonus:** While the script generates the reference, it does not add the files to the table of contents which is stored in the mkdocs-<version>.yaml([example](https://github.com/pachyderm/pachyderm/blob/master/doc/mkdocs-master.yml#L222)). Right now, we have to add them manually which is not optimal and would be great to automate with a script that parses that directory and outputs the list of files from that directory under ` Pachctl Reference:`. That being said, this can be addressed in a separate issue.

nadegepepin:


echohack (approved):
Looks great! Happy to have these little tweaks to improve our docs automation further.

PR #5673, opened 2021-02-03
Join & Group Inputs in 2.0

brycemcanally:
This PR implements join & group inputs in 2.0. There is some more design work that is needed for datum IDs. The current approach is not going to work (at least if we want it to be consistent across input types) because the datum ID could be very long for group inputs (or large crosses). For now, we are special casing the datum ID for group inputs to be the group ID, and we are going to circle back to this problem later, but before 2.0 GA.


jdoliner (approved):
LGTM

PR #5675, opened 2021-02-03
[ci skip] Add supported release and features doc

echohack:
Signed-off-by: echohack <git@echohack.app>

This document covers Pachyderm's expected levels of support for releases and features, and defines the nomenclature for each.

Historically, Pachyderm has used "alpha" and "beta" to refer to both features and releases, and this can cause confusion as the level of support provided, and the expectations for each of these is unknown.

There's a few goals for this document:

1. Establish a standard nomenclature around releases and features.
2. Ensure that there are no collisions in the nomenclature around releases and features.
3. Begin to mark and categorize features where appropriate.
4. Set expectations for support, including when a release or a feature is not supported.
5. Set a level of caution for features that are not production-ready.

Ultimately this document can give users clearer expectations around releases and features going forward.


nadegepepin (commented):
LGTM

PR #5679, opened 2021-02-04
Support prefix="/..." in S3 gateway

lukemarsden:
Fixes https://github.com/pachyderm/pachyderm/issues/5655

Do we need to merge this into master as well? S3G is disabled there. Need to check if the codepath this affects exists in 2.0 as well. If not, we need to make sure that we port this fix to whatever replaces it too.


msteffen (approved):
I think there's a library in PFS for normalizing paths too, but can add that later

lukemarsden:
@brycemcanally please make sure this fix goes into 2.0 as well. Thanks!

PR #5680, opened 2021-02-04
Remove OTPs

actgardner:
Dash 2.0 and hub are both going to move to using OIDC for users to auth to pachd, so we can (hopefully) eliminate support for OTPs in 2.0.

Normally I would wait longer to do this in case we decide we want OTPs after all, but I'm trying to prune down as much functionality in the auth server as possible before I tackle role bindings because it's a big structural change. If we decide we need OTPs in the future we can revert this and deal with the merge conflicts then.


msteffen (approved):
LGTM.

One use case that may be particularly affected by this is the dash: AFAIK you can't sign in to dash 1.0 via OIDC. If we want dash 1.0 to work with auth 2.0, merging this may undermine that.

actgardner:
> One use case that may be particularly affected by this is the dash: AFAIK it's not possible to sign in to dash 1.0 via OIDC. If we want dash 1.0 to work with auth 2.0 (which may be necessary for hub auth, i.e. going from the Hub dashboard to a cluster dashboard? Though you know way more about that than I do), merging this may undermine that.

My understanding is that Dash 1.0 won't work with 2.0 - we're making other API changes that would break it anyways. The plan for Hub and Dash 2.0 is to move to OIDC and eliminate OTPs. This is a pretty easy change to revert if we need to walk it back.

PR #5690, opened 2021-02-09
Fix pollPipelines crash

msteffen:
This is the 1.12.x version of #5692

### DIFF
--- a/src/server/pkg/backoff/retry.go
+++ b/src/server/pkg/backoff/retry.go
@@ -102,11 +102,15 @@ func RetryUntilCancel(ctx context.Context, operation Operation, b BackOff, notif
 
 	b.Reset()
 	for {
-		if err = operation(); err == nil {
+		err = operation()
+		if err == nil {
 			return nil
 		}
 		if ctx.Err() != nil {
 			return ctx.Err() // return if cancel() was called inside operation()
+		}
+		if errors.Is(err, ErrContinue) {
+			b.Reset()
 		}
 
 		if next = b.NextBackOff(); next == Stop {
--- a/src/server/pkg/backoff/retry.go
+++ b/src/server/pkg/backoff/retry.go
@@ -43,11 +43,19 @@ func NotifyCtx(ctx context.Context, name string) Notify {
 	}
 }
 
-// ErrContinue is a sentinel error designed to be used with NotifyContinue.
-// NotifyContinue always returns nil (causing operation() to be retried) when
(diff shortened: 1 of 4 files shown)


src/server/pkg/backoff/retry_test.go
```diff
@@ -71,6 +71,39 @@ func TestRetryUntilCancelZeroBackoff(t *testing.T) {
...
+	var b mustResetBackOff = 0
+	var results []int
+	RetryUntilCancel(context.Background(), func() error {
+		results = append(results, len(results)+1)
+		if len(results) < 3 {
+			return ErrContinue
+		}
+		return nil
+	}, &b, NotifyContinue(t.Name()))
+	require.Equal(t, []int{1, 2, 3}, results)
```
PFedak, line 104: I feel like I've seen this pattern a few places in our tests, is the motivation to get more informative output compared to just incrementing a counter?

src/server/pkg/backoff/retry.go
```diff
@@ -102,12 +102,16 @@ func RetryUntilCancel(ctx context.Context, operation Operation, b BackOff, notif
...
+		err = operation()
+		if err == nil {
 			return nil
 		}
 		if ctx.Err() != nil {
 			return ctx.Err() // return if cancel() was called inside operation()
 		}
+		if errors.Is(err, ErrContinue) {
+			b.Reset()
+		}
```
PFedak, line 114: This is definitely closer to how I expected `ErrContinue` to behave. I think it's worth updating its comment to make it clear that it no longer requires `NotifyContinue`. I see a couple of places in our codebase where `b.Reset()` is explicitly called in the loop to get this behavior, but I'm assuming no one would use the `notify` function to decide whether to turn this effect on or off.

PFedak (approved):
LGTM, but I think the comments for `ErrContinue` should be updated to reflect that it works at the backoff level now, rather than the weird notify handling. More important for the main PR, though.

src/server/pkg/backoff/retry.go
msteffen: Done

src/server/pkg/backoff/retry_test.go
msteffen: TBH, I used the pattern without thinking too much about it. Using a counter does seem equivalent, though I feel like there's probably a better pattern than either, for checking that the iterations behaved as expected rather than just checking that they happened. If I think of something better I'll send a followup PR?

PR #5691, opened 2021-02-09
Support Pachyderm services with no port

msteffen:
While services are normally meant for e.g. dashboards, the facts that they
- must have one datum, and
- don't skip datums

make them useful for propagating branch changes to the outside world. Currently our seldon demo requires a complicated `start transaction; create branch ...; update pipeline --reprocess; finish transaction` sequence to roll back to a prior version of a model and deploy it (without `--reprocess`, the deploy pipeline skips the prior model's datum and exits without deploying anything). With this change, the seldon deployment pipeline can be packaged into a service, and then simply running `pachctl create branch model@master model@master~1` does the right thing: consistently redeploy the prior model to seldon.

~Services actually mostly work for this already (if the user code running in a service exits, the job goes into SUCCESS)~ This is a bug in 1.12.x. However, independent of whether the user code exits, a benefit of services is that they can keep the s3 gateway open for the duration of the service. In these cases, it's not necessary to expose a public port, as the data being served is internal: the job's input data via the S3 gateway is going to another k8s pod (Seldon in this example)

### DIFF
--- a/src/server/worker/pipeline/service/service.go
+++ b/src/server/worker/pipeline/service/service.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"context"
+	"fmt"
 
 	"golang.org/x/sync/errgroup"
 
@@ -104,7 +105,11 @@ func Run(driver driver.Driver, logger logs.TaggedLogger) error {
 			}
 
 			if err := eg.Wait(); err != nil {
-				logger.Logf("error running user code: %+v", err)
+				reason := fmt.Sprintf("error running user code: %v", err)
+				if err := driver.UpdateJobState(job.ID, pps.JobState_JOB_FAILURE, reason); err != nil {
+					logger.Logf("error updating job progress: %+v", err)
+				}
+				logger.Logf(reason)
 			}
 
 			// Only want to update this stuff if we were canceled due to a new commit
--- a/src/server/worker/pipeline/service/service.go
+++ b/src/server/worker/pipeline/service/service.go
@@ -105,11 +105,7 @@ func Run(driver driver.Driver, logger logs.TaggedLogger) error {
 			}
 
 			if err := eg.Wait(); err != nil {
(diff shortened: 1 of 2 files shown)


src/server/worker/pipeline/service/service.go
```diff
@@ -104,7 +105,11 @@ func Run(driver driver.Driver, logger logs.TaggedLogger) error {
 			}
 
 			if err := eg.Wait(); err != nil {
-				logger.Logf("error running user code: %+v", err)
+				reason := fmt.Sprintf("error running user code: %v", err)
+				if err := driver.UpdateJobState(job.ID, pps.JobState_JOB_FAILURE, reason); err != nil {
+					logger.Logf("error updating job progress: %+v", err)
+				}
+				logger.Logf(reason)
```
Tryneus, line 112: It appears to be a variable-shadowing bug that services do not reattempt on the user code exiting, and we should fix that rather than setting the job to failed here.

Tryneus (changes requested):

src/server/worker/pipeline/service/service.go
msteffen: Done. Removed this part of the change and renamed the PR. The only thing it does now is: if no externalPort is set in the service spec, then no k8s service is created

src/server/worker/pipeline/service/service.go
```diff
@@ -2,6 +2,7 @@ package service
 
 import (
 	"context"
+	"fmt"
```
Tryneus, line 5: I imagine this is breaking the build, but otherwise LGTM

Tryneus (approved):

src/server/worker/pipeline/service/service.go
msteffen: :joy: Whoops, thanks!!

PR #5693, opened 2021-02-09
Role bindings

actgardner:
Historically there have been 4 "scopes" (roles) in Pachyderm: reader, writer, owner and admin. These roles mapped into a number of permissions: owners can modify ACLs and delete a repo, while writers can add files and edit pipeline specs, and readers can list, view and inspect files. Roles were enforced in a somewhat ad-hoc way - for instance, the spec repo was special-cased to be world-readable, or the PPS user was special-cased to be a cluster admin. We had the notion of an FS admin, which was Owner but on the entire cluster. 

This PR has two goals:
- Role bindings are standardized across all nouns. "Cluster roles" and "repo roles" are represented and evaluated the same way. Cluster role bindings are cached because they're commonly used, and there's a hierarchical relationship (cluster bindings apply to all repos). When we add "projects" as a new noun, it will be simple to extend this system to a three-tiered hierarchy and add new permissions that apply to projects.
- More granular, customizable roles are possible. The FS Admin role was a one-off solution to give all users access to edit ACLs but not to modify cluster configuration. Custom roles open the door to more user stories like "an operator can view logs for all pipelines and start and stop jobs, but not view the files or modify the job specifications". Or allowing users to view pachd logs but not to change OIDC configuration, etc.

Conceptually, the way role bindings work is:
- There's a set of operations users can perform on different nouns (listing files, deleting a repo, stopping a pipeline). Each of these operations has a corresponding `Permission`.
- `Permissions` are grouped together into `Roles`, which are identified by a string.
- Every noun in the system, called a `Resource`, (right now this means repos and the cluster itself) has a set of `RoleBindings` which are a map of `<principal, list of roles>`.
- To evaluate whether a principal can perform an operation on a `Resource`, we check the set of `Permissions` required for that operation against all of the `Roles` that principal has bound on the `Resource`, as well as `Roles` that the principal's groups have bound, and also the `Roles` the principal and groups have bound to any parent `Resources` (the cluster is a parent to all repos, eventually Projects will be parents to repos and clusters will be parents to Projects). If the set of `Permissions` is satisfied by some combination of the `RoleBindings`, then the operation is allowed.


src/auth/auth.go
```diff
@@ -41,18 +41,28 @@ const (
...
-			return Scope(value), nil
-		}
-	}
-	return Scope_NONE, errors.Errorf("unrecognized scope: %s", s)
-}
+	// PpsUser is a special, unrevokable cluster administrator account used by PPS
+	// to create pipeline tokens, close commits, and do other necessary PPS work.
+	// It's not possible to authenticate as ppsUser (pps reads the auth token for
+	// this user directly from etcd).
+	PpsUser = `pach:pps`
```
actgardner, line 49: Renamed the PPS user to `pach:pps` - all subjects with a `pach` prefix have cluster role bindings that can't be modified by users.

src/auth/auth.go
```diff
@@ -139,44 +149,20 @@ func IsErrExpiredToken(err error) bool {
...
 
-	Repo     string // Repo that the user is attempting to access
-	Required Scope  // Caller needs 'Required'-level access to 'Repo'
-
-	// Group 2:
-	// AdminOp indicates an operation that the caller couldn't perform because
-	// they're not an admin
-	AdminOp string
+	Resource Resource     // Resource that the user is attempting to access
+	Required []Permission // Caller needs 'Required'-level access to 'Resource'
```
actgardner, line 157: We no longer have two types of permissions errors - instead we return the resource and the set of missing permissions required.

src/auth/auth.proto
```diff
@@ -72,50 +72,6 @@ message SetConfigurationRequest {
...
-message GetAdminsResponse{
-  repeated string admins = 1;
-}
-
-// Deprecated. Add and remove users from the set of cluster super admins.
-message ModifyAdminsRequest {
-  repeated string add = 1;
-  repeated string remove = 2;
-}
-message ModifyAdminsResponse {}
```
actgardner, line 117: There's no longer a notion of "this group of users are admins" - we could re-add it as "the set of users with a specific role bound on the cluster" but it's not used for anything right now.

src/auth/auth.proto
```diff
@@ -157,26 +113,56 @@ message WhoAmIRequest {}
...
+  // principal -> roles. All principal names include the structured prefix indicating their type.
+  map<string, Roles> entries = 1;
+}
+
+// Permission represents the ability to perform a given operation on a Resource 
+enum Permission {
+  UNKNOWN = 0;
+
+  // CLUSTER_ADMIN encompasses all "admin" APIs, which are generally cluster configuration
+  CLUSTER_ADMIN        = 100;
```
actgardner, line 137: To keep things simple there's currently one "admin" permission in the admin role, but in a follow-on PR it makes sense to split this into a permission per admin RPC. I didn't tackle it here because it also involves rewriting the auth interceptor.

src/auth/auth.proto
```diff
@@ -157,26 +113,56 @@ message WhoAmIRequest {}
...
+  REPO_MODIFY_BINDINGS = 202;
+  REPO_DELETE          = 203;
+  REPO_INSPECT_COMMIT  = 204;
+  REPO_LIST_COMMIT     = 205;
+  REPO_DELETE_COMMIT   = 206;
+  REPO_CREATE_BRANCH   = 207;
+  REPO_LIST_BRANCH     = 208;
+  REPO_DELETE_BRANCH   = 209;
+  REPO_INSPECT_FILE    = 210;
+  REPO_LIST_FILE       = 211;
```
actgardner, line 150: This is a first pass at splitting out individual permissions from the PFS roles. I suspect we'll need to do a couple revisions to get to something useful.

src/auth/auth.proto
```diff
@@ -190,92 +176,51 @@ message Groups {
...
-  // username is the principal posessing this level of access to this ACL's
-  // repo (despite the name, this principal may be for a human user or a
-  // pachyderm robot)
-  string username = 1;
+message ModifyRoleBindingResponse {}
 
-  // scope is the level of access that the owner of 'principal' has to this
-  // ACL's repo (actually a role in typical security terminology)
-  Scope scope = 2;
+message DeleteRoleBindingRequest {
```
actgardner, line 209: Added an explicit RPC for `DeleteRoleBinding` - the previous behaviour had a subtle flaw where if you removed all users from the ACL, it would delete the ACL itself, and any user in the cluster could add themself as an Owner (because of the "new repo" logic).

src/server/auth/server/testing/admin_test.go
```diff
@@ -111,8 +95,11 @@ func TestActivate(t *testing.T) {
 	// Check that the token 'c' received from pachd authenticates them as "pach:root"
 	who, err := rootClient.WhoAmI(rootClient.Ctx(), &auth.WhoAmIRequest{})
 	require.NoError(t, err)
-	require.Equal(t, auth.ClusterRole_SUPER, who.ClusterRoles.Roles[0])
 	require.Equal(t, auth.RootUser, who.Username)
+
+	bindings, err := rootClient.GetClusterRoleBindings()
+	require.NoError(t, err)
+	require.Equal(t, buildClusterBindings(), bindings)
```
actgardner, line 102: Most of the changes in these tests are just 1:1 swaps of "check an ACL/check a cluster role binding" -> "check a role binding".

src/server/auth/server/testing/admin_test.go
```diff
@@ -187,28 +177,18 @@ func TestSuperAdminRWO(t *testing.T) {
 	require.Equal(t, 1, CommitCnt(t, aliceClient, repo)) // check that no commits were created
 
 	// bob can't update the ACL
-	_, err = bobClient.SetScope(bobClient.Ctx(), &auth.SetScopeRequest{
-		Repo:     repo,
-		Username: robot("carol"),
-		Scope:    auth.Scope_READER,
-	})
+	err = bobClient.ModifyRepoRoleBinding(repo, robot("carol"), []string{auth.RepoReaderRole})
```
actgardner, line 180: `SetScope` has been replaced by `ModifyRepoRoleBinding` which is a thin helper in the client.

src/server/auth/server/testing/admin_test.go
```diff
@@ -187,28 +177,18 @@ func TestSuperAdminRWO(t *testing.T) {
...
 	require.YesError(t, err)
 	require.Matches(t, "not authorized", err.Error())
-	require.Equal(t, entries(alice, "owner"), getACL(t, aliceClient, repo)) // check that ACL wasn't updated
+	require.Equal(t, buildBindings(alice, auth.RepoOwnerRole), getRepoRoleBinding(t, aliceClient, repo)) // check that ACL wasn't updated
 
 	// 'admin' makes bob a super admin
-	_, err = rootClient.ModifyClusterRoleBinding(rootClient.Ctx(),
-		&auth.ModifyClusterRoleBindingRequest{Principal: bob, Roles: superClusterRole()})
-	require.NoError(t, err)
+	require.NoError(t, rootClient.ModifyClusterRoleBinding(bob, []string{auth.ClusterAdminRole}))
```
actgardner, line 186: `ModifyClusterRoleBinding` is also a helper method in the client that follows the same structure as `ModifyRepoRoleBinding`.

src/server/auth/server/testing/admin_test.go
```diff
@@ -464,47 +389,6 @@ func TestFSAdminFixBrokenRepo(t *testing.T) {
 	require.Equal(t, 1, CommitCnt(t, rootClient, repo)) // check that a new commit was created
 }
 
-func TestModifyAdmins(t *testing.T) {
```
actgardner, line 467: `ModifyAdmins` is removed so this is no longer necessary.

src/server/auth/server/testing/admin_test.go
```diff
@@ -187,28 +177,18 @@ func TestSuperAdminRWO(t *testing.T) {
...
+	require.Equal(t, buildBindings(alice, auth.RepoOwnerRole), getRepoRoleBinding(t, aliceClient, repo)) // check that ACL wasn't updated
 
 	// 'admin' makes bob a super admin
-	_, err = rootClient.ModifyClusterRoleBinding(rootClient.Ctx(),
-		&auth.ModifyClusterRoleBindingRequest{Principal: bob, Roles: superClusterRole()})
-	require.NoError(t, err)
+	require.NoError(t, rootClient.ModifyClusterRoleBinding(bob, []string{auth.ClusterAdminRole}))
 
 	// wait until bob shows up in admin list
-	require.NoError(t, backoff.Retry(func() error {
```
actgardner, line 205: We block until the cached set of cluster bindings has been updated, so it's no longer necessary to spin in every test like this.

src/server/auth/server/testing/admin_test.go
```diff
@@ -932,129 +783,29 @@ func TestGetSetScopeAndAclWithExpiredToken(t *testing.T) {
...
-		},
-	})
+	bindings, err = rootClient.GetRepoRoleBindings(repo)
 	require.NoError(t, err)
-	require.ElementsEqual(t,
-		entries(alice, "owner", robot("carol"), "writer"), getACL(t, rootClient, repo))
-}
-
-// TestAdminWhoAmI tests that when an admin calls WhoAmI(), the ClusterRoles reflects their admin roles
-func TestAdminWhoAmI(t *testing.T) {
```
actgardner, line 1017: There's no longer an "admin" field in `WhoAmI` so this test is unnecessary.

src/server/auth/server/testing/admin_test.go
```diff
@@ -1137,17 +891,12 @@ func TestGetAuthToken(t *testing.T) {
 	who, err = robotClient2.WhoAmI(robotClient2.Ctx(), &auth.WhoAmIRequest{})
 	require.NoError(t, err)
 	require.Equal(t, robotUser, who.Username)
-	require.Equal(t, 0, len(who.ClusterRoles.Roles))
-	if version.IsAtLeast(1, 10) {
-		require.True(t, who.TTL >= 0 && who.TTL < secsInYear)
-	} else {
-		require.Equal(t, -1, who.TTL)
-	}
+	require.True(t, who.TTL >= 0 && who.TTL < secsInYear)
```
actgardner, line 894: Token TTLs are now just "whatever you asked for" without the version logic.

src/server/auth/server/testing/admin_test.go
```diff
@@ -1586,64 +1304,20 @@ func TestGetAuthTokenErrorFSAdminUser(t *testing.T) {
...
 		Subject: robot(tu.UniqueString("t-1000")),
 	})
 	require.Nil(t, resp)
 	require.YesError(t, err)
-	require.Matches(t, "must be an admin", err.Error())
-}
-
-// TestGetAuthTokenDefaultTTL tests the default TTL of a token returned from
-// GetAuthToken
-func TestGetAuthTokenDefaultTTL(t *testing.T) {
```
actgardner, line 1613: Removed these commented-out tests.

src/server/auth/server/testing/admin_test.go
```diff
@@ -1569,7 +1287,7 @@ func TestGetAuthTokenErrorNonAdminUser(t *testing.T) {
 	})
 	require.Nil(t, resp)
 	require.YesError(t, err)
-	require.Matches(t, "must be an admin", err.Error())
+	require.Matches(t, "needs permissions \\[CLUSTER_ADMIN\\] on CLUSTER", err.Error())
```
actgardner, line 1290: The text of the permission-denied error changed, so these had to be updated.

src/server/auth/server/testing/auth_test.go
```diff
@@ -27,54 +27,11 @@ import (
 	globlib "github.com/pachyderm/ohmyglob"
 )
 
-// aclEntry mirrors auth.ACLEntry, but without XXX_fields, which make
```
actgardner, line 30: Repos now use the same structure as the cluster role bindings, so these helpers aren't necessary.

src/server/auth/server/testing/auth_test.go
```diff
@@ -1595,49 +1334,6 @@ func TestCreatePipelineRepoAlreadyExistsPermissions(t *testing.T) {
 	))
 }
 
-// TestAuthorizedNoneRole tests that Authorized(user, repo, NONE) yields 'true',
-// even for repos with no ACL
-func TestAuthorizedNoneRole(t *testing.T) {
```
actgardner, line 1600: There's no longer a notion of None scope, so this isn't necessary. 

Also "None scope" makes me think of

src/server/pfs/server/driver.go
```diff
@@ -229,41 +227,34 @@ func (d *driver) inspectRepo(txnCtx *txnenv.TransactionContext, repo *pfs.Repo,
...
-	if who.ClusterRoles != nil {
-		for _, s := range who.ClusterRoles.Roles {
-			if s == auth.ClusterRole_SUPER || s == auth.ClusterRole_FS {
-				return auth.Scope_OWNER, nil
-			}
-		}
-	}
 
-	resp, err := pachClient.AuthAPIClient.GetScope(ctx, &auth.GetScopeRequest{Repos: []string{repo.Name}})
+	resp, err := pachClient.AuthAPIClient.Authorize(ctx, &auth.AuthorizeRequest{
```
actgardner, line 245: `getPermissions` just tries to `Authorize` with every permission,  and returns the list of permissions that were covered.

src/server/pps/server/api_server.go
```diff
@@ -3276,18 +3245,17 @@ func (a *apiServer) ActivateAuth(ctx context.Context, req *pps.ActivateAuthReque
 	defer func(start time.Time) { a.Log(req, resp, retErr, time.Since(start)) }(time.Now())
 	pachClient := a.env.GetPachClient(ctx)
 
+	// Set the permissions on the spec repo so anyone can read it
+	if err := pachClient.ModifyRepoRoleBinding(ppsconsts.SpecRepo, auth.AllClusterUsersSubject, []string{auth.RepoReaderRole}); err != nil {
```
actgardner, line 3249: Removed the special casing for the spec repo, it now has normal role bindings like every other repo.

src/server/pps/server/api_server.go
```diff
@@ -471,21 +461,23 @@ func (a *apiServer) authorizePipelineOpInTransaction(txnCtx *txnenv.TransactionC
...
-			} else if !isNotFoundErr(err) {
+				required = auth.Permission_REPO_WRITE
+			} else if isNotFoundErr(err) {
+				return nil
+			} else {
 				return err
 			}
 		case pipelineOpListDatum, pipelineOpGetLogs:
-			required = auth.Scope_READER
+			required = auth.Permission_REPO_READ
```
actgardner, line 478: `authorizePipelineOp` is something I would love to eliminate but it's a little outside the scope of this PR.

src/server/pps/server/api_server.go
```diff
@@ -3276,18 +3245,17 @@ func (a *apiServer) ActivateAuth(ctx context.Context, req *pps.ActivateAuthReque
...
 
+	// Set the permissions on the spec repo so anyone can read it
+	if err := pachClient.ModifyRepoRoleBinding(ppsconsts.SpecRepo, auth.AllClusterUsersSubject, []string{auth.RepoReaderRole}); err != nil {
+		return nil, errors.Wrapf(grpcutil.ScrubGRPC(err), "cannot configure role binding on spec repo")
+	}
+
 	// Unauthenticated users can't create new pipelines or repos, and users can't
 	// log in while auth is in an intermediate state, so 'pipelines' is exhaustive
 	var pipelines []*pps.PipelineInfo
-	if err := a.sudo(pachClient, func(superUserClient *client.APIClient) error {
```
actgardner, line 3282: Removed the sudoing from `pps.ActivateAuth`, since this needs to be callled by a cluster admin anyways.

src/server/auth/server/api_server.go
```diff
@@ -31,16 +27,12 @@ import (
 )
 
 const (
-	allClusterUsersSubject = "allClusterUsers"
-
-	tokensPrefix    = "/tokens"
-	aclsPrefix      = "/acls"
-	adminsPrefix    = "/admins"
-	fsAdminsPrefix  = "/fs-admins"
```
actgardner, line 39: ACLs, admins and fsAdmins are all just role bindings now.

src/server/auth/server/api_server.go
```diff
@@ -89,22 +74,14 @@ type apiServer struct {
 	txnEnv     *txnenv.TransactionEnv
 	pachLogger log.Logger
 
-	adminCache map[string]auth.ClusterRoles // cache of current cluster admins
-	adminMu    sync.Mutex                   // guard 'adminCache'
-
-	configCache *keycache.Cache
+	configCache             *keycache.Cache
+	clusterRoleBindingCache *keycache.Cache
```
actgardner, line 78: Use a keycache to cache the set of role bindings for the cluster, for performance.

src/server/auth/server/api_server.go
```diff
@@ -298,103 +269,9 @@ func (a *apiServer) retrieveOrGeneratePPSToken() {
...
-	}, b, func(err error, d time.Duration) error {
-		logrus.Errorf("error watching admin collection: %v; retrying in %v", err, d)
-		return nil
-	})
-}
-
-func (a *apiServer) getEnterpriseTokenState() (enterpriseclient.State, error) {
-	pachClient := a.env.GetPachClient(context.Background())
-	resp, err := pachClient.Enterprise.GetState(context.Background(),
+func (a *apiServer) getEnterpriseTokenState(ctx context.Context) (enterpriseclient.State, error) {
```
actgardner, line 272: Pass a context to `getEnterpriseTokenState` so we can cancel it.

src/server/auth/server/api_server.go
```diff
@@ -436,11 +313,16 @@ func (a *apiServer) Activate(ctx context.Context, req *auth.ActivateRequest) (re
...
+	// initialize the root user as a cluster admin
 	if _, err = col.NewSTM(ctx, a.env.GetEtcdClient(), func(stm col.STM) error {
-		admins := a.admins.ReadWrite(stm)
+		roleBindings := a.roleBindings.ReadWrite(stm)
 		tokens := a.tokens.ReadWrite(stm)
-		if err := admins.Put(auth.RootUser, epsilon); err != nil {
+		if err := roleBindings.Put(clusterRoleBindingKey, &auth.RoleBinding{
+			Entries: map[string]*auth.Roles{
+				auth.RootUser: &auth.Roles{Roles: map[string]bool{auth.ClusterAdminRole: true}},
+				auth.PpsUser:  &auth.Roles{Roles: map[string]bool{auth.ClusterAdminRole: true}},
```
actgardner, line 323: Explicitly include the PPS user in the cluster role bindings. These bindings can't be removed, but they make it clearer how the "PPS superuser" code works IMO.

src/server/auth/server/api_server.go
```diff
@@ -475,55 +354,10 @@ func (a *apiServer) Activate(ctx context.Context, req *auth.ActivateRequest) (re
...
-		// state, in case activation fails and they need to revert to "none"
-
-		// TODO: we should consider if we want to allow deactivation to proceed
-		// even if the state is "none", in case they're in a bad state
-		return nil, auth.ErrNotActivated
-	}
-
-	// Check if the cluster is in a partially-activated state. If so, allow it to
-	// be completely deactivated so that it returns to normal
-	var ppsUserIsAdmin bool
```
actgardner, line 489: I missed this when I removed partial activation. There's no auth logic in `Deactivate` anymore, it's all handled in the interceptor.

src/server/auth/server/api_server.go
```diff
@@ -533,188 +367,61 @@ func (a *apiServer) Deactivate(ctx context.Context, req *auth.DeactivateRequest)
...
+	// so that Deactivate() is less likely to race with subsequent calls that
+	// expect auth to be deactivated.
 	if err := backoff.Retry(func() error {
-		if a.isActive() {
+		if err := a.isActive(); err == nil {
 			return errors.Errorf("auth still activated")
 		}
 		return nil
-	}, backoff.RetryEvery(time.Second)); err != nil {
+	}, backoff.RetryEvery(100*time.Millisecond)); err != nil {
```
actgardner, line 378: Shorten these timeouts from 1 second to 100 milliseconds, to hopefully reduce test latency.

src/server/auth/server/api_server.go
```diff
@@ -533,188 +367,61 @@ func (a *apiServer) Deactivate(ctx context.Context, req *auth.DeactivateRequest)
...
+	callerInfo, err := a.getAuthenticatedUser(ctx)
 	if err != nil {
 		return err
 	}
-	if state != enterpriseclient.State_ACTIVE && !isAdmin {
-		return errors.New("Pachyderm Enterprise is not active in this " +
-			"cluster (until Pachyderm Enterprise is re-activated or Pachyderm " +
-			"auth is deactivated, only cluster admins can perform any operations)")
+
+	if callerInfo.Subject == auth.RootUser || callerInfo.Subject == auth.PpsUser {
```
actgardner, line 401: `expiredClusterAdminCheck` now permits only the PPS user or root user to make calls to a cluster where the enterprise license has run out.

src/auth/auth.go
msteffen: Finally! Thank you!!

src/auth/auth.proto
msteffen: `SetScope` had this bug? Or a prior implementation of `DeleteRoleBinding`? If it's `SetScope` we should fix it in 1.12.x too (and probably have a test? That sounds bad)

actgardner: I thought that `SetScope` had a bug where an "orphaned" repo could be "adopted" by any authenticated user, but I realized we handle this case by calling `InspectRepoInTransaction`. So the existing logic is fine, we should remove DeleteRoleBinding and just use that.

src/auth/auth.go
```diff
@@ -41,18 +41,28 @@ const (
...
-	return Scope_NONE, errors.Errorf("unrecognized scope: %s", s)
-}
+	// PpsUser is a special, unrevokable cluster administrator account used by PPS
+	// to create pipeline tokens, close commits, and do other necessary PPS work.
+	// It's not possible to authenticate as ppsUser (pps reads the auth token for
+	// this user directly from etcd).
+	PpsUser = `pach:pps`
+
+	// ClusterAdminRole is the role for cluster admins, who have full access to all APIs
+	ClusterAdminRole = "clusterAdmin"
```
msteffen, line 52: No preference on my part, but just curious: why define roles in Go as strings, but permissions in the proto as enums? Is the thinking that roles may need more flexibility or multiple parts later on?

src/auth/auth.proto
```diff
@@ -190,92 +177,54 @@ message Groups {
...
+  // resource is the resource to modify the role bindings on 
+  Resource resource = 1;
 
-message SetScopeResponse {}
+  // principal is the principal to modify the roles binding for
+  string principal = 2;
 
-message GetACLRequest {
-  string repo = 1;
+  // roles is the set of roles for principal
```
msteffen, line 207: Just checking my understanding: to revoke someone's access to a resource, you'd leave `roles` empty?

src/auth/auth.proto
msteffen: Phew! Makes sense

src/auth/auth_err_test.go
```diff
@@ -1,88 +0,0 @@
...
-	"github.com/pachyderm/pachyderm/v2/src/internal/require"
-)
-
-// grpcify returns an error e such that e.Error() is similar to what grpc
-// errors emit (though this doesn't convert 'error' to an actual GRPC error)
-func grpcify(err error) error {
-	return errors.Errorf("rpc error: code = Unknown desc = %s", err.Error())
-}
-
-func TestIsErrNotActivated(t *testing.T) {
```
msteffen, line 16: Just curious, why delete these tests? They are kind of pedantic—is it that, or do they no longer work?

src/auth/auth.proto
```diff
@@ -156,27 +112,58 @@ message AuthenticateResponse {
...
-  READER = 1;
-  WRITER = 2;
-  OWNER = 3;
+// Roles represents the set of roles a principal has
+message Roles {
+  map<string, bool> roles = 1;
+}
+
+// RoleBinding represents the set of roles principals have on a given Resource
+message RoleBinding {
```
msteffen, line 128: This is like, a very very minor nit, but I think the naming scheme here is slightly different at least from the one Kubernetes uses (I haven't checked GCP, actually. I can check that and update this comment). My understanding is that in Kubernetes, a RoleBinding is a single pair: `(principle, role)`, so if I have three roles on the same resource, that's technically three RoleBindings.

I think it's fine to keep this as-is, but I wanted to register the possible discrepency in case you think it might matter.

src/pfs/pfs.proto
```diff
@@ -49,7 +49,7 @@ message RepoInfo {
 message RepoAuthInfo {
   // The callers access level to the relevant repo (e.g. may be OWNER even if
   // the user isn't an OWNER of the repo, if they're an admin for the cluster)
-  auth.Scope access_level = 1;
+  repeated auth.Permission permissions = 1;
```
msteffen, line 52: This may be another API that it may make sense to makeover in 2.0 (not necessarily in this PR, just an opportunity that occured to me). I think of the main consumer of this field as being the dash, which wants to gray out repos that the user can't access. A list of permissions would definitely work, but I wonder if there's a more UI-friendly thing it would make sense to return?

src/server/auth/server/api_server.go
```diff
@@ -1477,70 +867,17 @@ func (a *apiServer) GetAuthToken(ctx context.Context, req *auth.GetAuthTokenRequ
 
 // GetAuthToken implements the protobuf auth.GetAuthToken RPC
 func (a *apiServer) GetAuthTokenInTransaction(txnCtx *txnenv.TransactionContext, req *auth.GetAuthTokenRequest) (resp *auth.GetAuthTokenResponse, retErr error) {
-	if !a.isActive() {
+	if err := a.isActive(); err != nil {
```
msteffen, line 870: Just checking my understanding here: it looks like this function is changing to drop support for the "no subject" case (also based on the fact that `TestGetAuthTokenNoSubject` is getting removed). Does this need to validate `req.Subject`? (or does that happen elsewhere? I might not have gotten to it yet).

FWIW I'm actually realizing that a consequence of dropping support for github auth is that a bunch of APIs that used to pretty flexible about subjects might need to do more validation now (i.e. subjects/principles should always have a prefix now. Apologies if that is generally checked already, I just need to revisit the code to remind myself).

src/server/auth/cmds/cmds_test.go
```diff
@@ -192,25 +193,6 @@ func TestConfig(t *testing.T) {
 		`).Run())
 }
 
-// TestGetAuthTokenNoSubject tests that 'pachctl get-auth-token' infers the
-// subject from the currently logged-in user if none is specified on the command
-// line
-func TestGetAuthTokenNoSubject(t *testing.T) {
```
msteffen, line 198: Just checking: it looks like support for this is getting dropped? I added a note to `GetAuthTokenInTransaction` too.

msteffen (commented):
First pass—thanks for being patient!

src/server/auth/server/testing/auth_test.go
msteffen: :joy:

msteffen: Literally laughing

src/auth/auth.go
actgardner: Yeah, my thinking here is that user-defined roles are an inevitable next step. Having roles be strings we validate in the server paves the way for that. Permisisons are an enum because they're tightly coupled to the actual code.

src/auth/auth.proto
actgardner: Correct, I'll add a comment to that effect.

src/auth/auth.proto
actgardner: I originally had `RoleBinding` pluralized, and then in some instances that changed a single binding I had it singular, and it made it harder to keep track of what APIs and structs had each spelling. I figured for simplicity it could just always be singular, even though this is really a collection of role bindings.

src/pfs/pfs.proto
actgardner: I'll talk to the Dash 2.0 team about this - I do think they'll want to know the permissions so they can display appropriate UI elements?

src/server/auth/server/api_server.go
actgardner: This is definitely a good point. Yes, this drops support for getting a token for yourself. Long-term I was going to split this into "get a token for a pipeline" and "get a token for a robot", since they have pretty different permissions models IMO.

src/server/auth/cmds/cmds_test.go
actgardner: Yup, dropping support for this  - it was easier to do it here and simplify the permissions checks but I think this API will also change more before 2.0.

src/auth/auth_err_test.go
actgardner: Re-added these, I ripped them out to get things compiling and forgot to circle back to update them.

src/server/pps/server/api_server.go
actgardner: Maybe I'm missing something - pipeline workers have the pipeline token, no? The alternative would be that it was possible for unauthenticated users to read all spec commits in 1.12.x which doesn't seem right?

src/server/auth/server/api_server.go
```diff
@@ -1477,70 +867,17 @@ func (a *apiServer) GetAuthToken(ctx context.Context, req *auth.GetAuthTokenRequ
...
-	// bad security practice (it might be). However, preventing admins from
-	// extending their own session here wouldn't improve security: admins can
-	// currently manufacture a buddy account, promote it to admin, and then extend
-	// their session indefinitely with the buddy account. Pachyderm cannot prevent
-	// this with the information is currently stores, so preventing admins from
-	// extending their session indefinitely would require larger changes to the
-	// auth model.
-	if !isAdmin {
-		// Caller is getting OTP for themselves--use TTL of their current token
-		ttl, err := a.getCallerTTL(txnCtx.ClientContext)
```
msteffen, line 1536: Just checking: is this getting removed? Can users still get an auth token for themselves, or is this locked to admins now?

TBH If users can get auth tokens for themselves, I do think that this is valuable from a security perspective (e.g. even though we're unifying auth under OIDC, I think ID tokens can have an `expires_in` constraint that, because we swap the ID token for an access token, it's on us to enforce. Users could work around it, though, if they could get a new token for themselves that had a longer expiration than their current token).

src/server/auth/server/api_server.go
msteffen: **Update** I looked through the existing code and realized that the set of functions that fail if enterprise is deactivated predates this PR and is kind of random:
```
GetACLInTransaction
SetACLInTransaction
SetScopeInTransaction
GetScopeInTransaction
SetACLInTransaction
```

I think ideally it would just be Authenticate(), but changing this seems like it's outside the scope of this PR

src/server/auth/server/api_server.go
actgardner: The way I'm thinking about auth tokens in 2.0 is: 

- `user` tokens are only available by logging in through the IDP. Nobody, not even admins, can impersonate a real person, and a user can't extend their own session by generating tokens for their own identity. 

- `robot` tokens can be created by admins for things like CI. This is a privileged action because a robot can live beyond the lifetime of a user, even indefinitely.

- `pipeline` tokens can be created by a user if they have appropriate permissions on the pipeline itself. 

In a future PR I'd like to split out `CreateRobotToken` and `CreatePipelineToken` into two separate RPCs so we don't have complicated string parsing - instead the user will provide the robot/pipeline name and we'll build the principal name and validate their access.

actgardner:
> Yeah I don't love it either. Maybe this isn't what you meant, but the problem I could never figure out is, if pipelines are resources in and of themselves, how should pachd decide whether it's OK to create a pipeline, since the pipeline inputs aren't stored anywhere except in the CreatePipelineRequest? Mapping a PipelineInfo to a bunch of auth requests on its inputs was kind of a workaround, but it does put a fair amount of auth logic in PPS. I mean I kind of wondered if we should just allow people to create a pipeline that consumes repos they can't read, and then let their pipeline fail, but that seems like bad UX?

So my thinking on this is - right now the `Reader` role on a repo implicitly includes a permission that means "you can add a pipeline as a reader on this repo". If we pulled that out as a separate permission and RPC, then you could get the auth check for free on the input repos when you called `AddPipelineAsReaderToRoleBindings(<pipeline>, <repo>)`. This lets us kill the superuser for setting up input repo ACLs.

> Just checking: this just affects Authenticate()? FWIW, the main reason we only disable login (on license expiration) in 1.x is so that pipelines keep running. In other words, if a production user of Pachyderm forgets to renew their license, we didn't want to break production for them, but we did want to put them in a state where the only way to keep using Pachyderm is to give everyone admin creds, thus effectively undermining auth.

The behaviour here is intended to be that when the enterprise token expires, only the root user and pipelines can perform any RPCs against the cluster - pipelines keep running and the root user can disable auth or renew the license, but that's it. 

> Also, would you need to be the root user to re-activate your license? TBH, I think the odds are decent that someone locks themselves out of their cluster by losing their root token and letting their license expire. If they can get a two-week trial license from us (which I think just requires an email) and reactivate enterprise long enough to deactivate auth (which I think would clear the root token?) then that seems like an OK enough escape hatch.

This is interesting - we could include a carve-out to allow any user to activate with a new enterprise license, even if the license is expired. I think in general with concerns about "what if the user loses the root token" my response is two-fold: (a) a lot of stuff can go wrong and this is the escape hatch, so we need to emphasize the importance of the root token in the docs and CLI (b) users can always do postgres surgery to make a new root token if they really lose it. The root token isn't used to encrypt any data or anything, there's always a manual process where someone can recover their cluster.

> Also I think ideally the error message should tell them they can keep using Pachyderm by using the root token.

Agree, I'll change this.

src/server/auth/server/api_server.go
```diff
@@ -820,58 +509,56 @@ func (a *apiServer) AuthorizeInTransaction(
...
 ) (resp *auth.AuthorizeResponse, retErr error) {
-	if !a.isActive() {
-		return nil, auth.ErrNotActivated
+	if err := a.isActive(); err != nil {
+		return nil, err
 	}
 
-	callerInfo, err := a.getAuthenticatedUser(txnCtx.ClientContext)
-	if err != nil {
+	if err := a.expiredEnterpriseCheck(txnCtx.ClientContext); err != nil {
```
msteffen, line 516: Per the comment above about keeping pipelines running, I think this `expiredEnterpriseCheck` would be important to remove (I think the expired enterprise check should be removed from the other non-Authenticate() APIs as well, but I think the reason we never noticed they were there in tests is that pipelines don't call those APIs).

I'm surprised that `TestPipelinesRunAfterExpiration` passed—we may need to debug that test too

src/server/auth/server/api_server.go
msteffen: This makes sense to me now, thanks!

I'm starting to feel like we need a list of product changes in auth 2.0, because there are enough that customer team will need a bit of time and tutoring before they can sell and support it (especially e.g. Joey and JK, who I think are pretty used to auth 1.0 and will need to be deprogrammed a bit).

I don't know what the best place for this to live would be, but to illustrate (and hopefully help), I started just a skeleton here: https://docs.google.com/document/d/1NaRJC_4kPbwSa629_-u8ytb1eZ1qu7uwkt5uImrJ2vQ/edit

msteffen:
> So my thinking on this is - right now the `Reader` role on a repo implicitly includes a permission that means "you can add a pipeline as a reader on this repo". If we pulled that out as a separate permission and RPC, then you could get the auth check for free on the input repos when you called `AddPipelineAsReaderToRoleBindings(<pipeline>, <repo>)`. This lets us kill the superuser for setting up input repo ACLs.

Oh, interesting. I mean this makes sense to me. Getting rid of the superuser token would definitely be cool. We might still need some version of authorizePipelineOp, just because we'd presumably want to check all input repos before changing any of them, but maybe we could do some kind of batch-ModifyRoleBindings request that could do all of the auth checks before making any changes? Just an idea. Basically this sounds great

> The behaviour here is intended to be that when the enterprise token expires, only the root user and pipelines can perform any RPCs against the cluster - pipelines keep running and the root user can disable auth or renew the license, but that's it. We might need to engage with the customer team to understand what "enterprise license is expired" should look like.

Makes sense. That proposed behavior does fit my product intuition—I guess I was just worried about the pipelines part, and just didn't totally understand how it worked previously

> This is interesting - we could include a carve-out to allow any user to activate with a new enterprise license, even if the license is expired. I think in general with concerns about "what if the user loses the root token" my response is two-fold: (a) a lot of stuff can go wrong and this is the escape hatch, so we need to emphasize the importance of the root token in the docs and CLI (b) users can always do postgres surgery to make a new root token if they really lose it. The root token isn't used to encrypt any data or anything, there's always a manual process where someone can recover their cluster.

Yeah, this makes total sense to me. I have done the equivalent etcd surgery to unlock a cluster for at least one user, so your point that that's always an option is well-taken. I guess one thought I've had in the past: if we come back around to encrypting all internal communication with mTLS, does that remove the ability to do postgres surgery? Maybe emergency surgey is just the backup plan to stick with, though, and we should just make sure to support emergency surgery as a use-case when we get around to internal mTLS? Like preserve the ability to mint new TLS certs if the cluster administrator is OK with it or something? Maybe certmanager has some kind of feature for that.

src/auth/auth.proto
msteffen: This makes sense. Just checking my understanding of one piece of all this: this PR gets rid of the `FS` cluster role—is that replaced with binding a user to the `repoOwner` role, but at the cluster level?

src/server/auth/server/testing/admin_test.go
```diff
@@ -222,27 +202,18 @@ func TestSuperAdminRWO(t *testing.T) {
...
-		Repo:     repo,
-		Username: robot("carol"),
-		Scope:    auth.Scope_READER,
-	})
+	err = bobClient.ModifyRepoRoleBinding(repo, robot("carol"), []string{auth.RepoReaderRole})
 	require.NoError(t, err)
 	// check that ACL was updated
-	require.ElementsEqual(t,
-		entries(alice, "owner", robot("carol"), "reader"), getACL(t, aliceClient, repo))
+	require.Equal(t, buildBindings(alice, auth.RepoOwnerRole, robot("carol"), "repoReader"), getRepoRoleBinding(t, aliceClient, repo))
```
msteffen, line 208: nit: why `"repoReader"` vs `auth.RepoReaderRole`?

src/server/auth/server/api_server.go
```diff
@@ -921,549 +608,252 @@ func (a *apiServer) WhoAmI(ctx context.Context, req *auth.WhoAmIRequest) (resp *
...
-	callerInfo, err := a.getAuthenticatedUser(txnCtx.ClientContext)
-	if err != nil {
+	if err := a.checkCanonicalSubject(req.Principal); err != nil {
 		return nil, err
 	}
-	callerIsAdmin, err := a.hasClusterRole(txnCtx.ClientContext, callerInfo.Subject, auth.ClusterRole_FS)
-	if err != nil {
-		return nil, err
+
+	if strings.HasPrefix(req.Principal, auth.PachPrefix) && req.Resource.Type == auth.ResourceType_CLUSTER {
```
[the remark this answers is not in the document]
actgardner: It prevents the root user from creating repos, which breaks a bunch of tests. I don't forsee customers creating repos as the root user a lot, but I also don't know of a good reason why we would forbid it.

actgardner:
For some reason I can't reply to some of the comments (maybe they're on individual commits versus the PR itself?)

> This makes sense. Just checking my understanding of one piece of all this: this PR gets rid of the FS cluster role—is that replaced with binding a user to the repoOwner role, but at the cluster level?

Correct, one of the goals of moving to all role-bindings is to make the roles composable like this, so we don't need to special-case "user who can read every repo in the cluster/project". Assigning repoReader at the project or cluster level applies to all repos in that project/cluster.

actgardner:
@msteffen Following up on our discussion this morning, I made a couple small changes:
- The enterprise license check now only happens in `Authenticate`, so basically any existing user/pipeline can continue to work but new users can't log in. I'll follow up with the customer team about how we want this to behave in 2.0 GA.
- I removed the `DeleteRoleBinding` RPC and made it just an in-process method call from PFS. I'd like to adopt this pattern more generally in transactionenv, so we have clearly "internal only" methods.

src/server/auth/server/api_server.go
```diff
@@ -921,549 +608,252 @@ func (a *apiServer) WhoAmI(ctx context.Context, req *auth.WhoAmIRequest) (resp *
...
-		return auth.Scope_NONE, errors.Wrapf(err, "could not retrieve caller's group memberships")
-	}
-	for _, g := range groups {
-		groupScope := acl.Entries[g]
-		if scope < groupScope {
-			scope = groupScope
-		}
+	roles := &auth.Roles{Roles: make(map[string]bool)}
+	for _, r := range req.Roles {
+		roles.Roles[r] = true
```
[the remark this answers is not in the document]
actgardner: We check that all the roles are valid earlier in the method :)

msteffen (approved):
LGTM

PR #5696, opened 2021-02-09
Fix GOPATH variable in Makefile when running on Windows

Tryneus:
`go env GOPATH` returns a windows-formatted path on windows, but the rest of the makefile expects unix-style paths.  Use `cygpath` to translate the path when running the makefile in windows.


brendoncarroll (approved):
LGTM

PR #5697, opened 2021-02-09
Run all tests in CI again.

PFedak:
Looks like this got lost during the bucketing change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,5 +105,6 @@ workflows:
               - PPS1
               - PPS2
               - PPS3
+              - PPS4
               - EXAMPLES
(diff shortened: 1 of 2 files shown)


PFedak:
Not sure if it makes sense to bump down the testfaster pool size to 11 (the number of buckets after this PR).

actgardner:
Oh, good catch 🤦 @lukemarsden do we need to change the number of prewarms to correspond to the number of test buckets?

actgardner (approved):

lukemarsden:
You might as well have the number of prewarms be a multiple (in this case,
1x) of the expected test parallelism

PR #5698, opened 2021-02-09
Add pipeline metrics

nitinjainsj:
This is a work in progress. I don't know what data we get from PIpelineInfo that we can add to our metrics. The implementation covers most of the things we discussed in the meeting. There are a few things I am not sure if they are correct.

- Should we look at ResourceRequests or ResourceLimits to send back?
- Is s3_out same as s3 gateway enabled?
- Do we only get elements that have been defined in the pipeline spec or does PipelineInfo also have defaults returned? If so, how do we distinguish it?
- Right now I am capturing most of the info as a bool. Are there any we should capture as numbers?
- Are there any more stats/config we can capture?


jdoliner (approved):
This all looks good to me.

PR #5699, opened 2021-02-10
Doc- Prometheus configuration draft

nadegepepin:


doc/docs/1.12.x/deploy-manage/deploy/prometheus/prometheus.md
```diff
@@ -0,0 +1,88 @@
+!!! note 
+    To monitor a Pachyderm cluster
+    with Prometheus, 
+    a ***Pachyderm Enterprise License*** is required. 
+
+# Monitor your Pachyderm cluster with Prometheus
```
mindthevirt, line 6: We should indicate that this is one way to deploy Prometheus but we really have no preference and don't want to be responsible for the deployment of Prometheus.

mindthevirt: Also, above the first point, we should call out that pachd and the pipeline pods expose Prometheus metrics

doc/docs/1.12.x/deploy-manage/deploy/prometheus/prometheus.md
```diff
@@ -0,0 +1,88 @@
...
+            In pachd.json, find:
+            ```shell
+            - name: prometheus-metrics
+                nodePort: 30656
+                port: 656
+                protocol: TCP
+                targetPort: 656
+            ```
+    
+
```
mindthevirt, line 68: We should say something like:

If you followed the instruction above, you can connect to Prometheus by using kubectl port-forward....
If you have an existing Prometheus deployment, please navigate to your Prometheus GUI.

doc/docs/1.12.x/deploy-manage/deploy/prometheus/prometheus.md
```diff
@@ -0,0 +1,88 @@
...
+    ```
+
+1. You can now browse your targets (http://localhost:9090/targets).
+    Run a pipeline of your choice. The `pachyderm-scraper` should be visible:
+
+    ![pachyderm scraper target](./img/prometheus_target_pachyderm_scaper.png)
+
+    In the ClassicUI tab, you should be able to see the new pachyderm metrics:
+
+    ![pachyderm metrics](./img/prometheus_pachyderm_metrics.png)
```
mindthevirt, line 82: we should replace this image with a list of metrics that are being exposed for pachd vs the pipeline.

mindthevirt: For a list of worker stats, see: https://github.com/pachyderm/pachyderm/blob/9b7c365665477ce73c514c9e1d902a4fa1555322/src/server/worker/stats/stats.go

For a list of pachd stats: 
https://github.com/pachyderm/pachyderm/blob/9b7c365665477ce73c514c9e1d902a4fa1555322/src/internal/storage/metrics/metrics.go

^ the above needs to be confirmed with eng

mindthevirt (commented):
Made some comments.

doc/docs/master/deploy-manage/deploy/prometheus/job_metrics.md
```diff
@@ -0,0 +1,24 @@
+Back to the [Prometheus set up page](../index).
+# List of job metrics exposed to Prometheus.
```
mindthevirt, line 2: Let’s change it to pipeline metrics. Instead of job metrics, I think it makes it clearer, even though you’re correct that these are metrics of jobs within a pipeline

PR #5709, opened 2021-02-12
Fix error when email_verified is not set in OIDC

actgardner:
`errors.Wrapf` has the annoying property that, if you wrap `nil`, the result is `nil`. In this case if the OIDC provider didn't set the `email_verified` claim, `validateIDToken` would return `(nil, nil, nil)` and callers would get an NPE because `err` was `nil` but no claims were returned.


brendoncarroll (approved):
LGTM

PR #5710, opened 2021-02-12
Open commit tests and bug fix

brycemcanally:
This PR adds at least one open commit test for each PFS file accessor. It also fixes a bug when accessing an open commit without a parent.


brendoncarroll (approved):
LGTM

PR #5711, opened 2021-02-12
Removes Object API, and tar package fork.

brendoncarroll:
Also creates pfs server protos, and renames DeleteCommit to SquashCommit


brycemcanally (approved):
lgtm

PR #5719, opened 2021-02-16
doc - Dash ingress - update

nadegepepin:


mindthevirt (changes requested):
The new format is great, thank you for cleaning this up. The graphic is easily understandable but a bit too small, is there a way to make it clickable, and then it opens up bigger? Also, the green font is difficult to read in the diagram. Everything else looks great

mindthevirt (approved):
LGTM!

PR #5720, opened 2021-02-17
Avoid metrics report deadlock.

PFedak:
Metrics reporting functions failed to close a wait channel when metrics
were disabled, causing commands like pachctl version to hang for ten
seconds. Also move the client creation to after the metrics check,
though it doesn't actually do anything.


brendoncarroll (approved):
LGTM

Issue #5704, opened 2021-02-11
TestFlake: TestCreatePipelineTransaction (1.12.x)


msteffen:
CircleCI run: https://app.circleci.com/pipelines/github/pachyderm/pachyderm/2397/workflows/03b1cfe0-ed4c-401f-a0b1-539c2c2525c4/jobs/32906

Test Logs: [test_logs.txt](https://github.com/pachyderm/pachyderm/files/5968087/test_logs.txt)
Failure Logs: [failure_logs.txt](https://github.com/pachyderm/pachyderm/files/5968089/failure_logs.txt)

Excerpt (looks like the protos are out of date in CircleCI, including that as well in case it's relevant):
```
Feb 09 21:12:17 + echo -e '\e[1;31m~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\e[0m'
Feb 09 21:12:17 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Feb 09 21:12:17 + echo -e '\e[1;31m~\e[0m \e[1;31mWARNING:\e[0m pachyderm_proto is out of date'
Feb 09 21:12:17 ~ WARNING: pachyderm_proto is out of date
Feb 09 21:12:17 + echo -e '\e[1;31m~\e[0m please run'
Feb 09 21:12:17 ~ please run
Feb 09 21:12:17 + echo -e '\e[1;31m~\e[0m   make DOCKER_BUILD_FLAGS=--no-cache proto'
Feb 09 21:12:17 ~   make DOCKER_BUILD_FLAGS=--no-cache proto
Feb 09 21:12:17 + echo -e '\e[1;31m~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\e[0m'
Feb 09 21:12:17 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Feb 09 21:12:17 + exit 1
Feb 09 21:12:17 tar: This does not look like a tar archive
Feb 09 21:12:17 tar: Exiting with failure status due to previous errors
Feb 09 21:12:17 make[1]: Leaving directory '/root/project/pachyderm'
Feb 09 21:12:17 Protos need to be recompiled; run 'DOCKER_BUILD_FLAGS=--no-cache make proto'.
Feb 09 21:12:17 go test -count=1 ./src/server/transaction/server/testing -timeout 3600s
Feb 09 21:12:17 make[1]: *** [Makefile:236: proto] Error 2
[...]

PR #5722, opened 2021-02-17
Flush / subscribe commit go client wrappers in 2.0.

brycemcanally:
This PR updates the flush / subscribe commit go client wrappers to follow the pattern we are establishing in 2.0 (callback based with an *All variant that collects everything). This change was made now to get ahead of some flakiness that has showed up in our testing due to the fact that flush commit does not return commits sorted by provenance (https://github.com/pachyderm/pachyderm/issues/5704). I have seen this show up multiple times because it is common to flush a commit then check against the first returned with the expectation that it is the output commit (in 2.0 it can be the stats commit). The `FlushCommitAll` wrapper in this PR sorts by provenance and all of the tests that rely on this behavior have been updated to use this (up to this point there have been a bunch of implementations of commit collection and all of them had to be updated over time to sort the commits by provenance). If a test is using this pattern, then it should use `FlushCommitAll` rather than the callback based `FlushCommit`.

Fixes https://github.com/pachyderm/pachyderm/issues/5704


jdoliner (approved):
LGTM

PR #5725, opened 2021-02-18
[WIP] Stop copying metadata on disk from output repos

actgardner:
**holding off on merging this because we don't actually have evidence of a speedup yet**

Remove the unix-specific `downloadTree` implementation which wrote the block to disk. This was redundant because `objClient` now caches metadata blocks on disk, so we were going block storage -> disk -> disk -> merger.


brycemcanally (approved):
lgtm

actgardner:
I tested a few improvements:
- removing the copy on disk from `downloadTree`
- avoiding deserializing every proto in a block in `nodes`

Neither of them seemed to put a big dent in the total latency for `GetFile`. So the theory that the bulk of the time is streaming back data blocks seems feasible.

PR #5728, opened 2021-02-19
Simplify activating auth on localhost

actgardner:
Make `pachctl auth activate`:
- activate auth
- set the issuer in dex to `localhost:30658`
- create a `pachd` client in dex
- configure pachd with the OIDC config for dex

This covers the common case where the user is running pachd in minikube. The only remaining step is for the user to configure the IDP in dex. If the user wants manual control over each step they can call `pachctl auth activate --only activate` to stop after auth is enabled.

The dex web server wouldn't start until a connector was configured. So I also wrote a mock connector that redirects to a static page which just says "no connectors are configured", which is only configured when no other connectors are configured.

### DIFF
--- a/dex-assets/static/not-configured.html
+++ b/dex-assets/static/not-configured.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+    <title>No identity providers configured</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link href="/static/main.css" rel="stylesheet">
+    <link href="/theme/styles.css" rel="stylesheet">
+    <link rel="icon" href="/theme/favicon.png">
+  </head>
+
+  <body class="theme-body">
+    <div class="theme-navbar">
+    </div>
+
+    <div class="dex-container">
+      <h2>No identity providers are configured, add one with pachctl idp add-connector</h2>
+    </div>
+  </body>
+</html>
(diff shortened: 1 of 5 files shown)


dex-assets/static/not-configured.html
```diff
@@ -0,0 +1,21 @@
...
+    <link href="/theme/styles.css" rel="stylesheet">
+    <link rel="icon" href="/theme/favicon.png">
+  </head>
+
+  <body class="theme-body">
+    <div class="theme-navbar">
+    </div>
+
+    <div class="dex-container">
+      <h2>No identity providers are configured, add one with pachctl idp add-connector</h2>
```
actgardner, line 18: This isn't pretty but all these static assets need a going-over for 2.0.

cpepin (approved):
LGTM

src/server/identity/server/dex_web.go
```diff
@@ -58,11 +62,33 @@ func (w *dexWeb) stopWebServer() {
...
 
-// startWebServer must called while holding the write mutex
-func (w *dexWeb) startWebServer() {
+// startWebServer returns a new dex web server, and a boolean for whether it should be cached.
+func (w *dexWeb) startWebServer() (*dex_server.Server, bool) {
+	cache := true
 	storage, err := w.storageProvider.GetStorage(w.logger)
 	if err != nil {
-		return
+		return nil, false
```
msteffen, line 70: nit: would it make sense to log this error or something? Is it every meaningful?

msteffen (approved):
LGTM, with one minor question

src/server/identity/server/dex_web.go
actgardner: The storage provider logs the error right now. It would probably make sense to consolidate all the logging in one place - I think I can actually eliminate the storage provider altogether now that postgres is part of every deployment, so we don't need to start lazily.

PR #5732, opened 2021-02-23
Fall back to object storage later for worker hashtrees.

PFedak:
The initial call to GetChunk for worker clients will succeed even if the
requested chunk isn't present. Since we don't rely on streaming
behavior, just read the whole chunk before deciding whether to fall back
to object storage.


brycemcanally:
I think this should probably be restructured so that we read directly into the cache instead of into an in-memory buffer. The cache `Put` call should delete the entry if an error occurs during the read.

PR #5737, opened 2021-02-23
Make CreateRoleBinding into an internal API

actgardner:
Simplify `ModifyRoleBindings` by adding the invariant that every resource also has a role binding:
- When creating a resource, the service that creates the resource also creates the role binding in the same transaction, using an internal API. 
- When deleting a resource, the service deletes the role binding explicitly in the same transaction using an internal API.
- When a user calls `ModifyRoleBindings` we can always assume the role binding exists if the resource exists, which removes the need to for a cross-service call to check whether the resource exists, whether it's being newly created, etc.

I think this makes it simpler to reason about the role bindings system - there's very little conditional logic in the Modify RPC now. The downside is that it does require adding an explicit `ActivateAuth` to PFS to add role bindings for all the existing repos.

This PR also moves us further in the direction of creating internal auth APIs that aren't exposed to users as RPCs and don't have their own auth checks. We have this in PFS but it's new ground for the auth subsystem so I'm open to feedback on structure, documentation, etc. if we're worried the internal API will create footguns. I'd like to start using this pattern more broadly.


src/server/auth/server/api_server.go
```diff
@@ -657,16 +657,53 @@ func (a *apiServer) DeleteRoleBindingInTransaction(txnCtx *txnenv.TransactionCon
 	return nil
 }
 
-func bindingFromRequest(req *auth.ModifyRoleBindingRequest) *auth.Roles {
-	if len(req.Roles) == 0 {
-		return nil
+// rolesFromRoleSlice converts a slice of strings into *auth.Roles,
```
actgardner, line 660: Small refactor: validate the roles when we're converting from a slice of strings to an `auth.Roles` since we need to do it in Create and Modify.

src/internal/auth/interceptor.go
```diff
@@ -207,7 +208,7 @@ var authHandlers = map[string]authHandler{
 	"/pps.API/GarbageCollect":  authDisabledOr(authenticated),
 	"/pps.API/UpdateJobState":  authDisabledOr(authenticated),
 	"/pps.API/ListPipeline":    authDisabledOr(authenticated),
-	"/pps.API/ActivateAuth":    authDisabledOr(authenticated),
+	"/pps.API/ActivateAuth":    admin,
```
actgardner, line 211: It doesn't make sense to call `ActivateAuth` unless auth is enabled and the user is an admin, make this consistent between PFS and PPS.

src/server/auth/server/api_server.go
```diff
@@ -618,34 +618,8 @@ func (a *apiServer) DeleteRoleBindingInTransaction(txnCtx *txnenv.TransactionCon
...
-	})
-	if err != nil {
-		return err
-	}
-	if !authorized.Authorized {
-		return &auth.ErrNotAuthorized{
-			Subject:  authorized.Principal,
-			Resource: *resource,
-			Required: permissions,
-		}
```
actgardner, line 648: To be consistent with CreateRoleBinding I removed the auth check from DeleteRoleBinding - they're both part of the internal API that should only be called by other services with their own auth checks. Likewise I removed the TODO about avoiding "orphan" repos because this should always run in a transaction along with the deletion of the resource itself.

msteffen (approved):
LGTM, with one question.

src/server/auth/server/api_server.go
```diff
@@ -683,6 +698,11 @@ func (a *apiServer) ModifyRoleBindingInTransaction(
 		return nil, err
 	}
 
+	roles, err := rolesFromRoleSlice(req.Roles)
```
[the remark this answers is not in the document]
actgardner: No, but since it validates the list of roles now I wanted to put it closer to the top to avoid doing work for invalid requests.

PR #5739, opened 2021-02-23
Split generating pipeline tokens out into a separate internal API

actgardner:
Add an unauthenticated, internal API which only generates tokens for PPS pipelines.

This is part of two larger auth initiatives:
- Eliminate the PPS superuser token.
- Replace the `GetToken` RPC with `GetRobotToken` to reduce the types of users that can be impersonated.


src/server/auth/server/api_server.go
```diff
@@ -872,6 +872,22 @@ func (a *apiServer) GetAuthTokenInTransaction(txnCtx *txnenv.TransactionContext,
 	}, nil
 }
 
+// GetPipelineAuthTokenInTransaction is an internal API used to create a pipeline token for a given pipeline.
+// Not an RPC.
```
msteffen, line 876: Maybe add an extra note about how it's insecure? Something like `does not authenticate the caller, and should only be used in code paths that have already authenticated the caller (currently just CreatePipeline)` or something?

msteffen (approved):
LGTM.

PR #5740, opened 2021-02-24
Introduce granular permissions for adding pipelines as repo readers

actgardner:
In 1.x we relied on the PPS superuser to add and remove pipeline principals from the input repo ACLs using general-purpose calls to set the ACLs. One of the goals in 2.0 is to eliminate the PPS superuser and use the user's permissions instead.

This PR adds new, very fine-grained permissions that used to be explicit in the PPS API: 
- Adding a pipeline as a reader on a repo. In 1.x the Reader role implied being able to add a pipeline to the ACL. 
- Removing a pipeline as a reader. In 1.x having writer on the pipeline's output repo implied being able to remove that pipeline from an input repo's ACL.
- Adding a pipeline as a writer on it's own output repo. This seems so narrowly defined that it should be safe for anyone to do - in 1.x being able to write to the output repo implied being able to add the pipeline as a writer.

The methods to add/remove pipelines on repos live in the auth package, so we don't have to expose a general-purpose method to modify role bindings without permissions checks. They're only exposed as an internal API to be used by PPS, although they would be safe to expose to end users - they don't rely on PPS to do the auth checks for them.


src/auth/auth.go
```diff
@@ -148,6 +148,27 @@ func IsErrExpiredToken(err error) bool {
 	return strings.Contains(err.Error(), status.Convert(ErrExpiredToken).Message())
 }
 
+const errNoRoleBindingMsg = "no role binding exists for"
```
actgardner, line 151: Add a new type of error for when the specified role binding doesn't exist. This is necessary because we delete the output repo before we remove the pipeline from the input repo bindings.

src/auth/auth.proto
```diff
@@ -137,18 +137,21 @@ enum Permission {
...
+  REPO_DELETE                 = 203;
+  REPO_INSPECT_COMMIT         = 204;
+  REPO_LIST_COMMIT            = 205;
+  REPO_DELETE_COMMIT          = 206;
+  REPO_CREATE_BRANCH          = 207;
+  REPO_LIST_BRANCH            = 208;
+  REPO_DELETE_BRANCH          = 209;
+  REPO_INSPECT_FILE           = 210;
+  REPO_LIST_FILE              = 211;
+  REPO_ADD_PIPELINE_READER    = 212;
```
actgardner, line 152: Add new permissions for adding a pipeline as a reader on a repo, removing a pipeline as a reader, and adding a pipeline as a writer. These are implicit in the 1.x roles model.

src/internal/transactionenv/env.go
```diff
@@ -141,6 +141,12 @@ type AuthTransactionServer interface {
 	ModifyRoleBindingInTransaction(*TransactionContext, *auth.ModifyRoleBindingRequest) (*auth.ModifyRoleBindingResponse, error)
 	GetRoleBindingInTransaction(*TransactionContext, *auth.GetRoleBindingRequest) (*auth.GetRoleBindingResponse, error)
 
+	// Methods to add and remove pipelines from input and output repos. These do their own auth checks
+	// for specific permissions required to use a repo as a pipeline input/output.
+	AddPipelineReaderToRepoInTransaction(*TransactionContext, string, string) error
+	AddPipelineWriterToRepoInTransaction(*TransactionContext, string) error
+	RemovePipelineReaderFromRepoInTransaction(*TransactionContext, string, string) error
```
actgardner, line 148: These are very specific methods in the internal API that have their own auth. The alternative was to put the auth logic in PPS and expose an API that supports modifying all role bindings without auth, which feels like a bigger possible footgun.

src/server/auth/server/api_server.go
```diff
@@ -543,7 +543,10 @@ func (a *apiServer) AuthorizeInTransaction(
 
 	// Get the role bindings for the resource to check
 	var roleBinding auth.RoleBinding
-	if err := a.roleBindings.ReadWrite(txnCtx.Stm).Get(resourceKey(req.Resource), &roleBinding); err != nil && !col.IsErrNotFound(err) {
```
actgardner, line 546: If the role binding doesn't exist in Authorize, throw ErrNoRoleBinding rather than trying to evaluate the cluster binding alone. Every resource should always have a corresponding role binding.

src/server/auth/server/api_server.go
```diff
@@ -698,11 +741,6 @@ func (a *apiServer) ModifyRoleBindingInTransaction(
 		return nil, err
 	}
 
-	roles, err := rolesFromRoleSlice(req.Roles)
```
actgardner, line 701: I refactored the non-auth guts of `ModifyRoleBindingInTransaction` into `setUserRoleBindingInTransaction` so it can be reused in each of the pipeline methods.

src/server/pps/server/api_server.go
```diff
@@ -1680,51 +1680,20 @@ func (a *apiServer) fixPipelineInputRepoACLsInTransaction(txnCtx *txnenv.Transac
 	for repo := range remove {
 		repo := repo
 		eg.Go(func() error {
-			return a.sudoTransaction(txnCtx, func(superTxnCtx *txnenv.TransactionContext) error {
```
actgardner, line 1683: The payoff - remove sudoing for `fixPipelineInputRepoACLs`

src/server/auth/server/testing/auth_test.go
```diff
@@ -2498,6 +2505,10 @@ func TestDebug(t *testing.T) {
 	tu.DeleteAll(t)
 	defer tu.DeleteAll(t)
 
+	// Activate auth at the beginning of the test, so pipelines are created with auth tokens.
+	// Otherwise when we activate auth the pps master will asynchronously update RCs with tokens,
+	// which can cause pods to stop existing while we collect the debug logs.
+	adminClient := tu.GetAuthenticatedPachClient(t, auth.RootUser)
```
actgardner, line 2511: /cc @brycemcanally - you mentioned the Debug dump test was flaky with auth enabled - this explains why the containers were disappearing during the test and fixes it.

brycemcanally: I'm not following why this addresses the issue. Does the `tu.GetAuthenticatedPachClient` call below when setting up the alice client not activate auth?

actgardner: Yeah I realized my comment is not quite accurate. Every call to `GetAuthenticatedPachClient` calls `pps.ActivateAuth`, which rotates the auth tokens for any existing pipelines, even if auth is already enabled. So setting up the alice client enables auth, but getting the root token attempts to activate auth again, which causes the specs for the existing pipelines to changge and bounces the pods.

src/server/auth/server/api_server.go
```diff
@@ -698,11 +741,6 @@ func (a *apiServer) ModifyRoleBindingInTransaction(
...
-	if err != nil {
-		return nil, err
-	}
-
 	if strings.HasPrefix(req.Principal, auth.PachPrefix) && req.Resource.Type == auth.ResourceType_CLUSTER {
```
msteffen, line 744: Proposal: something I've been thinking about with this change is that with the old 3-role system, if you were a READER on a repo, the only way to add a pipeline to that repo is if you also had access to all of the pipeline's other inputs & output (and therefore could update the pipeline—i.e. you couldn't directly add `pipeline:foo` to one ACL unless you were an OWNER on it).

Reflecting on it, I think I was basically uncomfortable with the idea of adding a `pipeline` user to an ACL via something other than `CreatePipeline`, so what if we made `ModifyRoleBinding` return an error if `req.Principal` is not a user principal, and then _only_ `CreatePipeline` can add `pipeline` users to ACLs, via `CreatePipeline -> AddPipelineReaderToRepoInTransaction -> setUserRoleBindingInTransaction`?

If we do implement that, a corrolary idea is that maybe we should let `pach:root` circumvent this restriction and add anyone to an ACL (in case of Pachyderm bugs basically).

WDYT?

msteffen (approved):
~LGTM~ (edit: I had one question about `fixPipelineInputRepoACLsInTransaction`)

One proposal about `ModifyRoleBindings` as well.

src/server/pps/server/api_server.go
```diff
@@ -1680,51 +1680,33 @@ func (a *apiServer) fixPipelineInputRepoACLsInTransaction(txnCtx *txnenv.Transac
...
-				_, err := superTxnCtx.Auth().ModifyRoleBindingInTransaction(
-					superTxnCtx,
-					&auth.ModifyRoleBindingRequest{
-						Resource:  &auth.Resource{Type: auth.ResourceType_REPO, Name: pipelineName},
-						Principal: auth.PipelinePrefix + pipelineName,
-						Roles:     []string{auth.RepoWriterRole},
-					})
+			// If we get an `ErrNoRoleBinding` that means the output repo no longer exists - this can happen if we're removing dangling role bindings
+			// from input repos but the output repo is already gone.
```
msteffen, line 1705: I may be misunderstanding this comment, but does this code block run in the case where a pipeline is being deleted? I think this should only run in the new pipeline case, where I think there should should already be an output repo (because it was created earlier in the request (and maybe ealier in the txn? I'll have to check `CreatePipeline`)

msteffen (commented):
One question about `fixPipelineInputRepoACLsInTransaction`, actually :grimacing:

src/server/pps/server/api_server.go
msteffen: I also just realized: this may not be cleaning up pipelines' output repos in the case where `pipelineInfo == nil`, which I think is just a bug. Hopefully not too bad, since it's only an invalid state if the pipeline has been deleted but the repo is still around, which a) I'd guess is rare, and b) I don't think can really be exploited. If you created a new pipeline with the same name, it'd just start writing the same output repo anyway.

actgardner: > I may be misunderstanding this comment, but does this code block run in the case where a pipeline is being deleted?

I missed the `prevPipelineInfo == nil` and assumed this was a belt-and-suspenders block to make sure the writer can _always_ write to the output repo. But I think you're right that the repo should always exist when this is called.

> I also just realized: this may not be cleaning up pipelines' output repos in the case where pipelineInfo == nil, which I think is just a bug.

I think the case of "create a new pipeline with an existing output repo" works with auth enabled, but yeah deleting a pipeline and leaving the repo leaves a weird dangling reference to the pipeline in the repo's role bindings. As you said, not a big deal because of constraints on pipeline output repo names.

src/server/auth/server/api_server.go
actgardner: > I was basically uncomfortable with the idea of adding a `pipeline` user to an ACL via something other than `CreatePipeline`

I don't think this is a security concern, right? It's mostly that we automatically manage pipeline permissions such that users _shouldn't_ need to change them. But I don't really forsee a huge downside to users manually adding/removing pipelines from repos where they have permissions to modify role bindings? The worst outcome I can see is a user breaks their own pipelines, and they can trivially fix it by updating the pipeline spec to fix the role bindings. 

One possible upside is being able to use the pipeline token for more than just reading inputs and writing output repos. It's possible users want to perform arbitrary RPCs against Pachyderm inside a pipeline, and I don't think we discourage that, but right now we make them generate a robot token manually and mount it in the pipeline. Is there a good reason they shouldn't be able to use the pipeline token for those operations, since it's already mounted and identified with the pipeline? In my mind robot tokens are mostly for external systems to talk to Pachyderm, and compute tasks inside Pachyderm use pipeline tokens.

I'm going to merge this so I can get all my changes in before PTO and they can get into the next alpha, but I'm open to discussing this more. Right now my instinct is that I want to avoid adding a special case (ModifyRoleBindings isn't allowed to work with pipeline principals) and then a special case to the special case (root user can modify role bindings for pipeline principals) but I'm not sure I fully understand the consequences of users modifying the role bindings manually.

msteffen: You make good points. Particularly the point about adding a pipeline to make `GetFile` work from inside the pipeline—I definitely see how adding the pipeline is cleaner than generating a robot token and having the pipeline use two tokens.

I guess the case I was thinking about is one where there's a pipeline I don't own and am not associated with. Could I change the behavior of that pipeline by creating a malicious repo and adding the pipeline to it? Maybe this is too contrived an example to worry about, but I guess what I'm wondering is, maybe it would make sense to say "I can add a pipeline to a repo if I can read the repo _and_ I can modify the pipeline?

Your point about not adding too many special cases also makes a ton of sense, though, and maybe the best option is to just wait until we've figured out pipeline permissions a bit more and see if a natural solution presents itself.

actgardner: > I guess the case I was thinking about is one where there's a pipeline I don't own and am not associated with. Could I change the behavior of that pipeline by creating a malicious repo and adding the pipeline to it? Maybe this is too contrived an example to worry about, but I guess what I'm wondering is, maybe it would make sense to say "I can add a pipeline to a repo if I can read the repo and I can modify the pipeline?

FWIW this is how GCP handles permissions - you can add any arbitrary service account to your resource role bindings, even across projects and organizations, because you're permitting them to make API calls but you can't force them to make those calls. So like, I could have a bucket full of malicious code and make it world-readable, but I can't force other people's GCE instances to download that code using only role bindings. 

I think the key distinction is that granting a pipeline user access to read from a repo is distinct from adding that repo to the pipeline spec.

PR #5745, opened 2021-02-25
Remove sudo from PPS monitor, pipeline controller and worker

actgardner:
Run `monitorPipeline` and `finishPipelineOutputCommits` using the pipeline's auth token instead of the superuser token. Also remove what seems like an extraneous `GetAuthToken` from the worker.

### DIFF
--- a/src/server/pps/server/monitor.go
+++ b/src/server/pps/server/monitor.go
@@ -61,7 +61,7 @@ const crashingBackoff = time.Second * 15
 // corresponding goroutine running monitorPipeline() that puts the pipeline in
 // and out of standby in response to new output commits appearing in that
 // pipeline's output repo.
-func (m *ppsMaster) startMonitor(pipelineInfo *pps.PipelineInfo) {
+func (m *ppsMaster) startMonitor(pipelineInfo *pps.PipelineInfo, ptr *pps.EtcdPipelineInfo) {
 	pipeline := pipelineInfo.Pipeline.Name
 	m.monitorCancelsMu.Lock()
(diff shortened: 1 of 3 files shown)


src/server/pps/server/worker_rc.go
```diff
@@ -825,26 +823,6 @@ func (a *apiServer) createWorkerSvcAndRc(ctx context.Context, ptr *pps.EtcdPipel
 		}
 	}
 
-	// Generate pipeline's auth token & add pipeline to the ACLs of input/output
```
msteffen, line 828: Good catch!

src/server/pps/server/monitor.go
```diff
@@ -70,14 +70,8 @@ func (m *ppsMaster) startMonitor(pipelineInfo *pps.PipelineInfo) {
...
-				// TODO(msteffen): run the pipeline monitor as the pipeline user, rather
-				// than as the PPS superuser
-				if err := m.a.sudo(pachClient, func(superUserClient *client.APIClient) error {
-					m.monitorPipeline(superUserClient, pipelineInfo)
-					return nil
-				}); err != nil {
-					log.Errorf("error monitoring %q: %v", pipeline, err)
-				}
+				pachClient.SetAuthToken(ptr.AuthToken)
+				m.monitorPipeline(pachClient, pipelineInfo)
```
msteffen, line 74: I like that this is shorter than the code that had the TODO on it. It's like, `// TODO: delete this todo`

msteffen (approved):
LGTM. Nice!

PR #5746, opened 2021-02-25
Remove PFS user as cluster admin

actgardner:
At this point the PFS token should only be used for accessing the spec repo - remove the cluster-wide admin permission and give it writer on the spec repo instead.


msteffen (approved):
LGTM (I had a few comments on https://github.com/pachyderm/pachyderm/pull/5740, but if I'm reading this right, this is like a <10 line change on top of that)

actgardner:
Yup, this is a pretty small change but I had it WIP until the other stuff lands. Thanks for reviewing it :)

PR #5753, opened 2021-02-26
Add go mod file to word count example

brycemcanally:
The latest version of Go has modules on by default, so this example started failing recently since there wasn't a module file.


jdoliner (approved):
lgtm

PR #5754, opened 2021-02-26
Fix issue when downloading chunk hashtree (1.12.x)

brycemcanally:
This PR fixes an issue that occurs when we need to fallback on object storage when downloading a chunk hashtree. The issue was that the code was only accounting for errors when setting up the streaming client, not for the actual read (where the `hashtree chunk not found` error will occur). The fix was to fallback when an error occurs in either case, and to also avoid the in-memory buffering since writes to the local cache are idempotent (and it deletes / cleans up the entry if an error occurs).

### DIFF
--- a/src/server/worker/pipeline/transform/worker.go
+++ b/src/server/worker/pipeline/transform/worker.go
@@ -627,13 +627,18 @@ func fetchChunkFromWorker(driver driver.Driver, logger logs.TaggedLogger, addres
 	return grpcutil.NewStreamingBytesReader(getChunkClient, cancel), nil
 }
 
-func fetchChunk(driver driver.Driver, logger logs.TaggedLogger, cache *hashtree.MergeCache, chunkID string, info *HashtreeInfo, shard int64, stats bool) error {
+func fetchChunk(driver driver.Driver, logger logs.TaggedLogger, cache *hashtree.MergeCache, chunkID string, info *HashtreeInfo, shard int64, stats bool) (retErr error) {
 	if info.Address != "" {
-		err := func() error {
+		err := func() (retErr error) {
 			reader, err := fetchChunkFromWorker(driver, logger, info.Address, info.SubtaskID, shard, stats)
 			if err != nil {
 				return err
 			}
+			defer func() {
+				if err := reader.Close(); retErr == nil {
+					retErr = err
+				}
+			}()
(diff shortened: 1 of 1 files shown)


src/server/worker/pipeline/transform/worker.go
```diff
@@ -627,20 +627,26 @@ func fetchChunkFromWorker(driver driver.Driver, logger logs.TaggedLogger, addres
...
 		logger.Logf("error when fetching cached chunk (%s) from worker (%s) - fetching from object store instead: %v", info.Object, info.Address, err)
 	}
 
 	reader, err := driver.PachClient().DirectObjReader(info.Object)
 	if err != nil {
-		return nil, errors.EnsureStack(err)
+		return err
 	}
-	return reader, nil
+	return cache.Put(chunkID, reader)
```
Tryneus, line 649: I think you still need to close the reader in both cases here (fetch from worker and direct read).

src/server/worker/pipeline/transform/worker.go
```diff
@@ -681,24 +687,7 @@ func handleMergeTask(driver driver.Driver, logger logs.TaggedLogger, data *Merge
...
-					}()
-
-					// TODO: this only works if it is read into a buffer first?
-					buf := &bytes.Buffer{}
-					if _, err := io.Copy(buf, reader); err != nil {
-						return errors.EnsureStack(err)
-					}
-
-					return errors.EnsureStack(cache.Put(chunkID, bytes.NewBuffer(buf.Bytes())))
+					return errors.EnsureStack(fetchChunk(driver, logger, cache, chunkID, hashtreeInfo, data.Shard, data.Stats))
```
Tryneus, line 690: I'd prefer if the `errors.EnsureStack` is closer to where the error is instantiated - if we get the stack here we won't know if it's from fetching from a worker or from a direct read without additional log statements helping out.

Tryneus (changes requested):
A couple concerns but overall it looks fine to me.

src/server/worker/pipeline/transform/worker.go
brycemcanally: We will know it is from a direct read since we will fallback on that for any error that occurs when fetching from the worker.

Tryneus (approved):
LGTM

PR #5763, opened 2021-03-01
Use fixed sized array for fileset.IDs

brendoncarroll:


src/internal/storage/fileset/compaction.go
```diff
@@ -16,7 +16,11 @@ func (s *Storage) IsCompacted(ctx context.Context, id ID) (bool, error) {
 	}
 	switch x := md.Value.(type) {
 	case *Metadata_Composite:
-		ids, err := s.Flatten(ctx, stringsToIDs(x.Composite.Layers))
+		ids, err := stringsToIDs(x.Composite.Layers)
```
brycemcanally, line 19: Shouldn't this be `PointsTo`?

src/internal/storage/fileset/fileset.go
```diff
@@ -1,42 +1,73 @@
...
+	id := ID{}
+	if _, err := rand.Read(id[:]); err != nil {
+		panic(err)
+	}
+	return id
 }
 
 // ParseID parses a string into an ID or returns an error
 func ParseID(x string) (*ID, error) {
 	if len(x) < 32 {
```
brycemcanally, line 29: Should we also check if the id is too long?

src/internal/storage/fileset/fileset.go
```diff
@@ -1,42 +1,73 @@
...
 
+// Scan implements sql.Scanner
+func (id *ID) Scan(src interface{}) error {
+	x, ok := src.([]byte)
+	if !ok {
+		return errors.Errorf("scanning fileset.ID: can't turn %T into fileset.ID", src)
+	}
+	x = bytes.Replace(x, []byte{'-'}, []byte{}, -1)
+	*id = ID{}
+	if len(x) < 32 {
```
brycemcanally, line 59: Maybe this should be a helper function that we can call to validate a hex encoded ID?

brendoncarroll: what other place would we use that?

brycemcanally: Where I commented above. It seems like the validation should check for an exact length and be implemented in one place through a helper function.

brycemcanally (approved):
lgtm

PR #5764, opened 2021-03-01
Let fuse inodes be generated automatically

brendoncarroll:


jdoliner (approved):
This lgtm. Did you find any docs that said that fuse would auto gen inodes if you pass 0 for the inode? I'm cuirious how that actually works.

brendoncarroll:
It says you can use 0 in the documentation for `StableAttr`
https://pkg.go.dev/github.com/hanwen/go-fuse/fs#StableAttr

It might use the pointer value of the node to prevent allocating new inodes, but I'm not sure exactly how it works.

PR #5772, opened 2021-03-04
look up spec commits concurrently

brendoncarroll:


src/server/pps/server/api_server.go
```diff
@@ -2492,37 +2492,79 @@ func (a *apiServer) listPipeline(pachClient *client.APIClient, request *pps.List
...
-				pipelineInfo, err = ppsutil.GetPipelineInfo(pachClient, name, ptr)
-				if err != nil {
+				if err := func() error {
+					mu.Lock()
+					defer mu.Unlock()
+					// the filtering shares that buffer thing, and it's CPU bound so why not do it with the lock
+					if !filterPipeline(pinfo) {
+						return nil
+					}
+					return f(pinfo)
```
Tryneus, line 2551: The only problem I see is that we may call `f` after it has already errored.  Perhaps we should have a bool to bail out after acquiring `mu` if an error has occurred (or the `eg` has been canceled)?

Tryneus (changes requested):
LGTM with one concern

src/server/pps/server/api_server.go
brendoncarroll: Is there a guarantee that f won't be called after it errors?  f won't be called after the function returns.  But I think the caller already doesn't know how many times f will be called.

To get the behavior you're describing, I guess we add another channel and then check the context for cancellation?

Tryneus:
LGTM

PR #5775, opened 2021-03-05
Doc - Getting Started with PachHub clean up

nadegepepin:


mindthevirt (changes requested):
Pachhub has to be renamed to Hub.

Also, the screenshot pachhub_cluster_connect.png should only show step 1-3

doc/docs/1.12.x/pachhub/pachhub_getting_started.md
```diff
@@ -1,181 +1,101 @@
...
       pachctl config get active-context
       ```
-
-      **System Response:**
-
-      ```shell
-      test-svet-cc0mi51i52
+      **System response**
+      ```
+      jolly-caribou
```
jrockway, line 72: It might be valuable to note that jolly-caribou is whatever you typed into the "create workspace" dialog.  (Maybe that's clear from the screenshots though.)

doc/docs/master/pachhub/pachhub_getting_started.md
```diff
@@ -1,181 +1,101 @@
...
+!!! Note
+    Let us know how to improve Pachyderm Hub. Join our [Slack channel](http://slack.pachyderm.io).
+## Get started in 4 simple steps
+![Pachyderm Hub Steps](../images/pachhub_steps.png)
+## Before you start
+Log in with your GitHub or Google account to start using [hub.pachyderm.com](https://hub.pachyderm.com). 
+![Pachyderm Hub Login](../images/pachhub_login.png)
+## 1- Create a Cluster 
+Click the **Create a 4-hr Workspace** button and fill out the form.
+![PachHub cluster](../images/pachhub_create_workspace.png)
```
jrockway, line 22: Hub workspace

doc/docs/master/pachhub/pachhub_getting_started.md
```diff
@@ -1,181 +1,101 @@
...
+
+## 2- Install pachctl
+Pachyderm Hub enables you to access your cluster through Pachyderm 
+CLI `pachctl` and the web interface called the Dashboard.
+Although you can perform most simple actions directly in the Dashboard,
 `pachctl` provides full functionality. Most likely, you will use
 `pachctl` for any operation beyond the most basic workflow.
-Pachyderm recommends that you use `pachctl` for all data operations and
-the dashboard to view your data and graphical representation of your
+We recommend that you use `pachctl` for all data operations and
```
jrockway, line 39: I think there is some documentation style guide floating around that says to never use "we" in documentation... but I hate that convention and prefer what you wrote here ;)

jrockway (commented):
Looks good to me.  Please resolve Jan's concerns before submitting. 

There have been several iterations on what things are called, so it would be good to do a pass over the docs and make sure that we are internally consistent :/.  I think with this PR, things are good, but just to be extra clear:

Workspaces were once called pachs or clusters.  I see some images that call them "clusters" still.  I don't think we ever call them that in the UI.

Teams were once called organizations or orgs.

Hub was once called PachHub or PacHub.  (I don't think anyone ever decided how many "h"s where in there 😓)

jrockway (approved):
The image updates look good.  PachHub still appears in the filenames of the images, but I don't think that's a big concern.  It can be cleaned up in the future if we really care.

mindthevirt:
@nadegepepin I think we're 99% there. Can we just make sure to remove PachHub from the URL and just call it Hub or Pachyderm Hub?

PR #5780, opened 2021-03-08
Deprecate Vault Plugin in 1.12.5

actgardner:
Add a notice to the docs that the Vault plugin is deprecated as of the next release (1.12.5).


msteffen (approved):
LGTM

echohack:
I'm late to the party! This is great!

PR #5781, opened 2021-03-08
Remove vault plugin from 2.0

actgardner:
Remove the Vault plugin in 2.0, which makes it easier to make some auth API changes. In the future we could bring it back and adapt it to the new API if there's demand for that.


msteffen (approved):
LGTM

PR #5784, opened 2021-03-08
[WIP] Split up the admin role into permissions

actgardner:


src/internal/auth/interceptor.go
```diff
@@ -37,45 +38,42 @@ var authHandlers = map[string]authHandler{
...
-	"/auth.API/GetAuthToken":      admin,
-	"/auth.API/SetConfiguration":  admin,
-	"/auth.API/ExtendAuthToken":   admin,
-	"/auth.API/SetGroupsForUser":  admin,
-	"/auth.API/ModifyMembers":     admin,
-	"/auth.API/GetUsers":          admin,
-	"/auth.API/ExtractAuthTokens": admin,
-	"/auth.API/RestoreAuthToken":  admin,
-	"/auth.API/Deactivate":        admin,
+	"/auth.API/GetConfiguration":  authenticated,
```
msteffen, line 48: There's a permission defined for `AUTH_GET_CONFIG`, but it looks like this doesn't use it—is it for futureproofing or would it maybe make sense to get rid of it?

msteffen (approved):
One small question, but otherwise LGTM

src/internal/auth/interceptor.go
msteffen: Nevermind, just saw https://github.com/pachyderm/pachyderm/pull/5785, which addresses this. Marking resolved

PR #5792, opened 2021-03-09
Track users authenticated via Dex

actgardner:
Add an interceptor for the Dex web server that handles the `/approval` page.  The lifespan of a Dex request looks like:
- the client app redirects the user to Dex with a state, nonce and redirect
- Dex stores the client's state and nonce and generates a new request ID
- Dex redirects the user through the IDP login flow and they authenticate
- Dex adds the data about the authenticated user's identity to the request record and redirects them to the approval page
- the approval page ties together the authenticated identity with the state, nonce and redirect and redirects the user back to the client app

By hooking into the approval route we can catch auth requests after they've logged in to the IDP but before an ID token has been issued for the client app. This lets us record the user's identity and, in the future, enforce the seat limit.


etc/testing/port-forward.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
actgardner, line 1: The web server tests don't need to run in k8s, but now they rely on postgres which is deployed on k8s. This script port-forwards postgres from k8s so the tests can use it when they're running outside, which makes iteration must faster.

msteffen (approved):
One nit, but otherwise LGTM

PR #5800, opened 2021-03-10
Add GetRobotToken auth RPC and CLI

actgardner:
In 2.0 `GetAuthToken` will be replaced with `GetRobotToken`, which can only be used for robot users. The default behaviour is to create a token that lasts forever (robot tokens are for external services), but tokens can have a TTL.

The goal is to give users more granular control over robot tokens used to integrate with external services. As a stretch goal in 2.0 we'll add the "Robots as Nouns" story where users can set role bindings for individual robots.

This PR is the first of several:
- Add `GetRobotToken`
- Switch all tests to use `GetRobotToken`
- Remove `GetAuthToken`


src/server/auth/server/api_server.go
```diff
@@ -870,6 +870,31 @@ func (a *apiServer) GetRoleBinding(ctx context.Context, req *auth.GetRoleBinding
 	return response, nil
 }
 
+// GetRobotToken implements the protobuf auth.GetRobotToken RPC
+func (a *apiServer) GetRobotToken(ctx context.Context, req *auth.GetRobotTokenRequest) (resp *auth.GetRobotTokenResponse, retErr error) {
+	a.LogReq(req)
+	defer func(start time.Time) { a.LogResp(req, resp, retErr, time.Since(start)) }(time.Now())
+
+	subject := auth.RobotPrefix + req.Robot
```
msteffen, line 878: Would it make sense to canonicalize the subject? (i.e. in case a user accidentally sends `robot:foo` and then the token is for `robot:robot:foo` or something?)

msteffen (approved):
One question, but otherwise LGTM

src/server/auth/server/api_server.go
```diff
@@ -875,7 +875,10 @@ func (a *apiServer) GetRobotToken(ctx context.Context, req *auth.GetRobotTokenRe
 	a.LogReq(req)
 	defer func(start time.Time) { a.LogResp(req, resp, retErr, time.Since(start)) }(time.Now())
 
-	subject := auth.RobotPrefix + req.Robot
+	subject := req.Robot
+	if !strings.HasPrefix(subject, auth.RobotPrefix) {
+		subject = auth.RobotPrefix + subject
```
msteffen, line 880: would it make sense to error/trim if a user sends e.g. `robbot:foo`?

actgardner: I'm hesitant to trim the user input arbitrarily, so I made colons illegal in robot names. You can send `robot:rosie` or `rosie` but not `robbot:rosie`.

PR #5811, opened 2021-03-11
Make it possible to deploy the Enterprise Server stand-alone

actgardner:
The Enterprise Server is a stripped-down pachd that doesn't have the PFS or PPS services - it's designed to register a bunch of other pachds with the license service and provide a centralized identity service for them.

This PR:
- adds a new mode for the pachd binary, `enterprise`, which only runs the services for the enterprise server
- adds a new flag, `pachctl deploy --enterprise-server`, which deploys the enterprise service instead of a normal pachd

Most of the changes are to make it possible for pachd and the enterprise-server to coexist on the same minikube cluster - this is necessary for testing and for the customer team. The enterprise server uses it's own postgres and etcd  (with their own host paths) and different NodePorts to avoid conflicts.


src/server/cmd/pachd/main.go
```diff
@@ -90,6 +92,286 @@ func doReadinessCheck(config interface{}) error {
 	return env.GetPachClient(context.Background()).Health()
 }
 
+func doEnterpriseMode(config interface{}) (retErr error) {
```
actgardner, line 95: This is all copy-paste of `doFullMode` - at some point it seems like it would be good to DRY this up, but I think it's bigger surgery than I want to undertake in the 2.0 timeframe.

src/internal/transactionenv/env.go
```diff
@@ -397,10 +397,12 @@ func (env *TransactionEnv) WithWriteContext(ctx context.Context, cb func(*Transa
 			Client:        pachClient,
 			ClientContext: pachClient.Ctx(),
 			Stm:           stm,
-			pfsPropagater: env.pfsServer.NewPropagater(stm),
 			txnEnv:        env,
 		}
-		txnCtx.commitFinisher = env.pfsServer.NewPipelineFinisher(txnCtx)
+		if env.pfsServer != nil {
```
actgardner, line 402: The auth service creates it's own transaction contexts to run `InTransaction` methods, so we need to support a transaction context with no PFS server.

src/internal/deploy/cmds/cmds.go
```diff
@@ -537,7 +546,19 @@ func standardDeployCmds() []*cobra.Command {
 				// Serve the Pachyderm object/block API locally, as this is needed by
 				// our tests (and authentication is disabled anyway)
 				opts.ExposeObjectAPI = true
+
+				// Set the postgres and etcd nodeports explicitly for developers
+				opts.PostgresOpts.Port = 32228
+				opts.EtcdOpts.Port = 32379
```
actgardner, line 552: We were setting these nodeports in a kind of obscure way before, but I assumed the goal was to have fixed nodeports for developers so we can easily connect to postgres and etcd?

src/internal/deploy/assets/assets.go
```diff
@@ -657,26 +662,34 @@ func PachdDeployment(opts *AssetOpts, objectStoreBackend Backend, hostPath strin
...
+		command = append(command, "--mode=enterprise")
+	}
+
 	return &apps.Deployment{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Deployment",
 			APIVersion: "apps/v1",
 		},
-		ObjectMeta: objectMeta(pachdName, labels(pachdName), nil, opts.Namespace),
+		ObjectMeta: objectMeta(name, labels(name), nil, opts.Namespace),
```
actgardner, line 678: Give the enterprise server a different deployment name - it should be in a different namespace at least, but this makes it look a little less like another random pachd pod.

src/internal/deploy/assets/etcd.go
```diff
@@ -355,11 +358,7 @@ func EtcdVolumeClaim(size int, opts *AssetOpts) *v1.PersistentVolumeClaim {
 
 // EtcdNodePortService returns a NodePort etcd service. This will let non-etcd
 // pods talk to etcd
-func EtcdNodePortService(local bool, opts *AssetOpts) *v1.Service {
-	var clientNodePort int32
```
actgardner, line 359: We were setting these nodeports in kind of an obscure way - I changed this to only set the ports in dev mode, see below.

src/internal/deploy/assets/assets.go
```diff
@@ -730,6 +743,58 @@ func PachdDeployment(opts *AssetOpts, objectStoreBackend Backend, hostPath strin
 	}
 }
 
+// EnterpriseService returns a service for the Enterprise Server
+func EnterpriseService(opts *AssetOpts) *v1.Service {
```
brendoncarroll, line 747: I think we have started pulling these functions out of `assets.go`; see `postgres.go` and `etcd.go`.  Can we make a separate file in the same package for this?

src/internal/deploy/assets/assets.go
```diff
@@ -730,6 +743,58 @@ func PachdDeployment(opts *AssetOpts, objectStoreBackend Backend, hostPath strin
...
+					Name:     "api-grpc-port",
+					NodePort: 31650,
+				},
+				{
+					Port:     651, // also set in cmd/pachd/main.go
+					Name:     "trace-port",
+					NodePort: 31651,
+				},
+				{
+					Port:     auth.OidcPort,
```
brendoncarroll, line 777: So this runs it's own Dex.  I assume the configuration for that will be the same as for pachd?

src/internal/deploy/cmds/cmds.go
```diff
@@ -537,7 +546,19 @@ func standardDeployCmds() []*cobra.Command {
 				// Serve the Pachyderm object/block API locally, as this is needed by
 				// our tests (and authentication is disabled anyway)
 				opts.ExposeObjectAPI = true
```
brendoncarroll, line 548: I thought the ObjectAPI was gone?

src/internal/deploy/cmds/cmds.go
```diff
@@ -537,7 +546,19 @@ func standardDeployCmds() []*cobra.Command {
...
+
+				// Set the postgres and etcd nodeports explicitly for developers
+				opts.PostgresOpts.Port = 32228
+				opts.EtcdOpts.Port = 32379
 			}
+
+			// Put the enterprise server backing data in a different path,
```
brendoncarroll, line 555: What sort of data goes here? Is this just forward looking at the moment? I thought the state was in postgres?

brendoncarroll:
One thing I think would make this architecture easier to understand for customers is if the enterprise server had it's own docker image and binary.  Even if that just calls into a parameterized pachd function behind the scenes.

src/internal/deploy/assets/assets.go
actgardner: I'm not sure what you mean.

src/internal/deploy/assets/postgres.go
```diff
@@ -352,7 +351,7 @@ func PostgresService(local bool, opts *AssetOpts) *v1.Service {
 				{
 					Port:     5432,
 					Name:     "client-port",
-					NodePort: clientNodePort,
+					NodePort: opts.PostgresOpts.Port,
```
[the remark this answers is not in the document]
actgardner: As a follow-on PR I could see making postgres deployments more configurable - any thoughts on parameters we should expose?

src/internal/deploy/cmds/cmds.go
actgardner: If so we should make an issue to remove all the references to it.

src/internal/deploy/assets/postgres.go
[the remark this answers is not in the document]
brendoncarroll: Probably just CPU, Memory, and Storage.

src/internal/deploy/cmds/cmds.go
brendoncarroll: Yeah I think that's my bad, I didn't get all the configuration, when I removed it.  Just confirming it's not used for anything here.

brendoncarroll: I threw this onto the 2.0 board. https://github.com/pachyderm/pachyderm/projects/11#card-57128360

brendoncarroll (approved):

chainlink:
cc @robert-uhl we might want to add this as an option in the 2.x helm chart

PR #5817, opened 2021-03-12
Add enterprise server context to pachctl config

actgardner:
Add the concept of a separate context in the pachctl config which points to the organization's enterprise server:
- In a single-server deployment (like `pachctl deploy local`), the enterprise context just points to the same active context. All the commands run against one pachd.
- In a multi-server deployment, the enterprise context points at the enterprise server, and the active context points to a pachd. Identity and license commands run against the enterprise server, while everything else happens in the pachd.


src/client/client.go
```diff
@@ -455,11 +455,6 @@ func NewForTest() (*APIClient, error) {
 
 // NewOnUserMachine constructs a new APIClient using $HOME/.pachyderm/config
 // if it exists. This is intended to be used in the pachctl binary.
-//
-// TODO(msteffen) this logic is fairly linux/unix specific, and makes the
-// pachyderm client library incompatible with Windows. We may want to move this
-// (and similar) logic into src/server and have it call a NewFromOptions()
-// constructor.
 func NewOnUserMachine(prefix string, options ...Option) (*APIClient, error) {
```
actgardner, line 458: Refactor `NewOnUserMachine` so we can reuse the lgoic for `NewEnterpriseClientOnUserMachine`.

src/server/enterprise/cmds/cmds.go
```diff
@@ -99,29 +98,40 @@ func DeactivateCmd() *cobra.Command {
 
 // RegisterCmd returns a cobra.Command that registers this cluster with a remote Enterprise Server.
 func RegisterCmd() *cobra.Command {
-	var server, id string
+	var id, pachdAddr, enterpriseAddr string
```
brendoncarroll, line 101: Where are these set?

src/server/config/cmds_test.go
```diff
@@ -85,6 +85,22 @@ func TestActiveContext(t *testing.T) {
...
+		t.Skip("Skipping integration tests in short mode")
+	}
+
+	require.YesError(t, run(t, `
+		pachctl config set active-enterprise-context foo 2>&1 | match "context does not exist: foo"
+	`))
+
+	require.NoError(t, run(t, `
+		echo '{}' | pachctl config set context foo --overwrite
+		pachctl config set active-enterprise-context foo
```
brendoncarroll, line 99: So anyone who runs this test will have their foo context clobbered?

src/internal/config/config.go
```diff
@@ -69,6 +70,39 @@ func (c *Config) ActiveContext(errorOnNoActive bool) (string, *Context, error) {
...
+		return "", nil, errors.Errorf("cannot get active context from non-v2 config")
+	}
+	if envContext, ok := os.LookupEnv(enterpriseContextEnvVar); ok {
+		context := c.V2.Contexts[envContext]
+		if context == nil {
+			return "", nil, errors.Errorf("pachctl config error: `%s` refers to a context (%q) that does not exist", contextEnvVar, envContext)
+		}
+		return envContext, context, nil
+	}
+	context := c.V2.Contexts[c.V2.ActiveEnterpriseContext]
```
brendoncarroll, line 85: I don't think this needs to change now. But naming things context can become annoying if anyone ever needs to important the context package.

src/client/client.go
```diff
@@ -469,7 +464,29 @@ func NewOnUserMachine(prefix string, options ...Option) (*APIClient, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "could not get active context")
 	}
+	return newOnUserMachine(cfg, context, prefix, options...)
+}
 
+// NewEnterpriseClientOnUserMachine constructs a new APIClient using $HOME/.pachyderm/config
```
brendoncarroll, line 470: This might be out of scope, but since we were talking about UX. I personally don't like it when applications write to my $HOME. I think there are standards for `.config` and `.local` but no one ever follows them.

brendoncarroll (approved):
One possible correctness issue with the argument parsing.  Some non-blocking suggestions, and then a few questions for my own understanding. Okay to merge though.

src/server/enterprise/cmds/cmds.go
actgardner: Nope, the flags are parsed automatically by cmdutils. We just define the variables and provide a pointer when defining the flags below.

src/server/config/cmds_test.go
actgardner: This is a convention in the config CLI tests right now - we could probably change them all to use `UniqueString` in the future.

src/client/client.go
actgardner: I agree, another thing to reconsider for the product in the future. In this case we're reusing the existing pachyderm config and just adding a new field for the enterprise context, this comment is copy-paste from `NewOnUserMachine`.

PR #5818, opened 2021-03-12
Keeps Object Storage and Postgres in Sync

brendoncarroll:
These changes let us create and delete chunks without ever ending up with a chunk without a corresponding object in object storage.  It is possible to wind up with multiple objects per chunk, and objects without a Postgres entry, but the existence of these anomalies will not impact correctness, and there are background services to eliminate them eventually.

### DIFF
--- a/src/internal/storage/chunk/client.go
+++ b/src/internal/storage/chunk/client.go
@@ -40,6 +40,7 @@ func NewClient(store kv.Store, db *sqlx.DB, tr track.Tracker, name string) Clien
 	}
 	c := &trackedClient{
 		store:   store,
+		db:      db,
 		tracker: tr,
 		renewer: renewer,
 		ttl:     defaultChunkTTL,
@@ -50,6 +51,9 @@ func NewClient(store kv.Store, db *sqlx.DB, tr track.Tracker, name string) Clien
 // Create creates a new chunk from metadata and chunkData.
 // It returns the ID for the chunk
 func (c *trackedClient) Create(ctx context.Context, md Metadata, chunkData []byte) (_ ID, retErr error) {
+	if c.renewer == nil {
+		panic("client must be named to create chunks")
+	}
 	chunkID := Hash(chunkData)
 	var pointsTo []string
 	for _, cid := range md.PointsTo {
@@ -64,7 +68,7 @@ func (c *trackedClient) Create(ctx context.Context, md Metadata, chunkData []byt
 		}
 		var ents []Entry
 		if err := tx.Select(&ents, `
-		SELECT hash_id, gen
+		SELECT chunk_id, gen
 		FROM storage.chunk_objects
 		WHERE uploaded = TRUE AND tombstone = FALSE AND chunk_id = $1`, chunkID); err != nil {
 			return err
@@ -91,14 +95,16 @@ func (c *trackedClient) Create(ctx context.Context, md Metadata, chunkData []byt
 	if !needUpload {
 		return chunkID, nil
 	}
-	// TODO: need to check for existence of specific objects.
 	key := chunkKey(chunkID, gen)
-	if exists, err := c.store.Exists(ctx, key); err != nil {
-		return nil, err
-	} else if exists {
-		return chunkID, nil
-	}
 	if err := c.store.Put(ctx, key, chunkData); err != nil {
+		return nil, err
+	}
+	_, err := c.db.Exec(`
+	UPDATE storage.chunk_objects
+	SET uploaded = TRUE
+	WHERE chunk_id = $1 AND gen = $2
+	`, chunkID, gen)
+	if err != nil {
 		return nil, err
 	}
 	return chunkID, nil
@@ -110,7 +116,7 @@ func (c *trackedClient) Get(ctx context.Context, chunkID ID, cb kv.ValueCallback
 	err := c.db.Get(&gen, `
 	SELECT gen
 	FROM storage.chunk_objects
-	WHERE chunk_id = %1
+	WHERE uploaded = TRUE AND tombstone = FALSE AND chunk_id = $1
 	LIMIT 1
 	`, chunkID)
 	if err != nil {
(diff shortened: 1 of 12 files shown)


src/internal/storage/chunk/client.go
```diff
@@ -43,73 +50,105 @@ func NewClient(store kv.Store, mdstore MetadataStore, tr track.Tracker, name str
...
 	chunkID := Hash(chunkData)
 	var pointsTo []string
 	for _, cid := range md.PointsTo {
-		pointsTo = append(pointsTo, ObjectID(cid))
+		pointsTo = append(pointsTo, cid.TrackerID())
 	}
-	chunkOID := ObjectID(chunkID)
-	if err := dbutil.WithTx(ctx, c.mdstore.DB(), func(tx *sqlx.Tx) error {
-		if err := c.mdstore.SetTx(tx, chunkID, md); err != nil && err != ErrMetadataExists {
+	chunkOID := chunkID.TrackerID()
```
brycemcanally, line 62: We should probably use `TID` as the suffix for tracker ID to avoid confusion between object storage objects and tracker objects. This is pretty minor, but I think it might be important for readability.

src/internal/storage/chunk/gc.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (gc *GarbageCollector) deleteOne(ctx context.Context, ent Entry) error {
+	if err := gc.deleteObject(ctx, ent.ChunkID, ent.Gen); err != nil {
+		return err
+	}
+	return gc.deleteEntry(ctx, ent.ChunkID, ent.Gen)
+}
+
+func (gc *GarbageCollector) deleteObject(ctx context.Context, chunkID ID, gen uint64) error {
+	return gc.s.store.Delete(ctx, chunkKey(chunkID, gen))
```
brycemcanally, line 84: I think we could run into issues if we return an error here when the object does not exist. For example, if the machine crashes immediately after this call and before deleting the entry, then I think the next garbage collector would be stuck in a loop once it reached this point again for the object.

src/internal/storage/chunk/gc.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (gc *GarbageCollector) deleteEntry(ctx context.Context, chunkID ID, gen uint64) error {
+	_, err := gc.s.db.ExecContext(ctx, `
+	DELETE FROM storage.chunk_objects
+	WHERE chunk_id = $1 AND gen = $2 AND tombstone = TRUE
+	LIMIT 1
```
brycemcanally, line 91: Why is this necessary?

src/internal/storage/chunk/gc.go
brendoncarroll: I think delete is idempotent.  If it's not at the object level; I propose changing it at the kv level.

src/internal/storage/chunk/gc.go
brendoncarroll: It's not.  Would you like to join the add limits to point deletes cargo cult?

src/internal/storage/chunk/client.go
brendoncarroll: Yeah this makes sense.

src/internal/storage/chunk/gc.go
brendoncarroll: Okay so delete is not idempotent on GCS
https://cloud.google.com/storage/docs/xml-api/delete-object

But it is in AWS.
https://docs.aws.amazon.com/sdkfornet1/latest/apidocs/html/M_Amazon_S3_AmazonS3Client_DeleteObject.htm

So I changed my mind, this needs to be uniform before it leaves the obj package.

src/internal/storage/chunk/client.go
```diff
@@ -43,73 +50,105 @@ func NewClient(store kv.Store, mdstore MetadataStore, tr track.Tracker, name str
...
+func (c *trackedClient) Get(ctx context.Context, chunkID ID, cb kv.ValueCallback) (retErr error) {
+	var gen uint64
+	err := c.db.Get(&gen, `
+	SELECT gen
+	FROM storage.chunk_objects
+	WHERE uploaded = TRUE AND tombstone = FALSE AND chunk_id = $1
+	LIMIT 1
+	`, chunkID)
+	if err != nil {
+		err = errors.Errorf("no objects for chunk %v", chunkID)
```
brycemcanally, line 123: It seems like this should be handling sql no rows and just returning the error otherwise.

src/internal/storage/chunk/gc.go
```diff
@@ -0,0 +1,149 @@
...
+	`, chunkID, gen)
+	return err
+}
+
+// ReduceObjectsPerChunk eliminates redundant objects
+func ReduceObjectsPerChunk(ctx context.Context, db *sqlx.DB) error {
```
brycemcanally, line 97: I think we should punt on these two functions for now. They seem pretty low leverage since we expect this issue to be pretty rare. It looks like you are doing that anyways since they are not in use outside of a test.

src/internal/storage/chunk/gc.go
brycemcanally: Okay, so are you thinking we should have a standard error for this case or just requiring that the clients don't return an error if the object does not exist?

src/internal/storage/chunk/gc.go
brycemcanally: lol, no thanks.

src/internal/storage/chunk/gc.go
brendoncarroll: I like no error. I think that is more useful generally. I can't think of a use case where we care if there was something there.

brendoncarroll: Okay, for the sake of this P.R. lets assume this is the behavior. And track it here:
https://github.com/pachyderm/pachyderm/pull/5866#issuecomment-804299195

brycemcanally (approved):
lgtm

PR #5822, opened 2021-03-12
Rework `pachctl deploy ide` for 2.0

actgardner:
Update `pachctl deploy ide` to use the identity server over OIDC for 2.0. 

Using OIDC presents a lot of complications, because we need to know:
- the URL for the identity server (from the user's laptop)
- the URL for the identity server (from the IDE pod) 
- the URL for the IDE (for the redirect)

I've tried to make this simple for users getting started by having everything go through `pachctl port-forward`, which makes the URLs predictable from the user's perspective. In a production deployment users will need to customize the public URLs so everyone can access them (or I guess have every port-forward).

We also need to register a new client ID with the appropriate redirect and secret. This is idempotent, so a user can re-run the deploy and we'll reuse the existing client ID unless the redirect has changed. We also register the client ID as a peer of pachd so the IDE can get a token for the pachd audience.

The end result is that users can still run the following to get a working IDE:
```
pachctl deploy local
pachctl enterprise activate
pachctl auth activate
pachctl idp add-connector <config for some auth connector>
pachctl deploy ide
pachctl port-forward
```

### DIFF
--- a/src/client/auth.go
+++ b/src/client/auth.go
@@ -11,12 +11,16 @@ func (c APIClient) IsAuthActive() (bool, error) {
 		Resource: &auth.Resource{Type: auth.ResourceType_CLUSTER},
 	})
 	switch {
+	case err == nil:
+		return true, nil
+	case auth.IsErrNotAuthorized(err):
+		return true, nil
 	case auth.IsErrNotSignedIn(err):
 		return true, nil
 	case auth.IsErrNotActivated(err):
 		return false, nil
 	default:
-		return true, grpcutil.ScrubGRPC(err)
+		return false, grpcutil.ScrubGRPC(err)
 	}
 }
 
--- a/src/client/auth.go
+++ b/src/client/auth.go
@@ -16,7 +16,7 @@ func (c APIClient) IsAuthActive() (bool, error) {
 	case auth.IsErrNotActivated(err):
 		return false, nil
 	default:
-		return false, grpcutil.ScrubGRPC(err)
+		return true, grpcutil.ScrubGRPC(err)
 	}
 }
 
--- a/src/client/portforwarder.go
+++ b/src/client/portforwarder.go
@@ -70,10 +70,14 @@ func NewPortForwarder(context *config.Context, namespace string) (*PortForwarder
 
 // Run starts the port forwarder. Returns after initialization is begun with
 // the locally bound port and any initialization errors.
-func (f *PortForwarder) Run(appName string, localPort, remotePort uint16) (uint16, error) {
+func (f *PortForwarder) Run(appName string, localPort, remotePort uint16, selectors ...string) (uint16, error) {
 	podNameSelector := map[string]string{
 		"suite": "pachyderm",
 		"app":   appName,
+	}
+
+	for i := 1; i < len(selectors); i += 2 {
+		podNameSelector[selectors[i-1]] = selectors[i]
 	}
 
 	podList, err := f.core.Pods(f.namespace).List(metav1.ListOptions{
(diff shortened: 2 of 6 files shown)


src/client/portforwarder.go
```diff
@@ -70,12 +70,16 @@ func NewPortForwarder(context *config.Context, namespace string) (*PortForwarder
 
 // Run starts the port forwarder. Returns after initialization is begun with
 // the locally bound port and any initialization errors.
-func (f *PortForwarder) Run(appName string, localPort, remotePort uint16) (uint16, error) {
+func (f *PortForwarder) Run(appName string, localPort, remotePort uint16, selectors ...string) (uint16, error) {
```
actgardner, line 73: The JupyterHub chart uses a slightly different label scheme, so we need to be able to filter by `component: proxy` as well.

src/client/auth.go
```diff
@@ -11,6 +11,10 @@ func (c APIClient) IsAuthActive() (bool, error) {
 		Resource: &auth.Resource{Type: auth.ResourceType_CLUSTER},
 	})
 	switch {
+	case err == nil:
+		return true, nil
+	case auth.IsErrNotAuthorized(err):
+		return true, nil
```
actgardner, line 17: This method wasn't working in the case where nothing bad happened, which blocked deploying the IDE.

msteffen (commented):
Sorry, I'm just coming back to this PR and realized I had a bunch of comments I hadn't submitted

src/internal/deploy/cmds/cmds.go
```diff
@@ -1114,8 +1109,10 @@ func Cmds() []*cobra.Command {
 						"name": hubImageName,
 						"tag":  hubImageTag,
 					},
-					"extraConfig": map[string]interface{}{
-						"templates": "c.JupyterHub.template_paths = ['/app/templates']",
+					"extraEnv": map[string]interface{}{
+						"OAUTH2_AUTHORIZE_URL": idAddr + "/auth",
+						"OAUTH2_TOKEN_URL":     internalIDAddr + "/token",
```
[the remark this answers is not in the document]
actgardner: This is all just oauthenticator config, the only customization we do now is putting the access token in the environment.

src/internal/deploy/cmds/cmds.go
```diff
@@ -64,14 +65,8 @@ access, you can check progress with:
...
-Once all of the pods are in the 'Ready' status, you can access the IDE in the
-following manners:
-
-* If you're on docker for mac, it should be accessible on 'localhost'.
-* If you're on minikube, run 'minikube service proxy-public --url' -- one or
-  both of the URLs printed should reach the IDE.
-* If you're on a cloud deployment, use the external IP of
-  'kubectl get service proxy-public'.
+Once all of the pods are in the 'Ready' status, you can access the IDE
+by running 'pachctl port-forward' and visiting 'localhost:30659'
```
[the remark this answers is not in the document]
actgardner: Looking at the docs for the helm chart, they do have you deploy a load-balanced service with no TLS and no authenticator to the cloud by default 🤕 

My thinking about this is that we probably shouldn't make that foot-gun case easy? It would be hard anyways because you'd need to know the redirect URI, which will be the random load-balancer IP. I think for a cloud deployment the minimum standard is going to be:

- be able to assign a hostname or fixed IP for the service
- some kind of TLS

It sucks that deploying a web-accessible service in k8s is not straightforward, but the alternative is that we're telling customers to put their data on the internet with no security, which doesn't seem great.

actgardner:
I've done a release of IDE 2.0.0-alpha1, everything is working testing locally:
Screen Shot 2021-03-30 at 11 02 09 AM

src/identity/identity.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+var (
+	// ErrInvalidID is returned if the client or connector ID does not exist
+	ErrInvalidID = status.Error(codes.Internal, "ID does not exist")
+
+	// ErrAlreadyExists is returned if the clientor connector ID already exists
```
msteffen, line 14: minor nit: `client or connector`

msteffen (approved):
LGTM, with one very minor nit

PR #5834, opened 2021-03-16
Add integration tests with a remote enterprise server

actgardner:
Add an integration test that tests against an external enterprise server, in the same cluster in a different namespace. This lays the groundwork to also test other enterprise server features.


Makefile
```diff
@@ -217,6 +217,14 @@ launch-dev: check-kubectl check-kubectl-connection install
 	until timeout 1s ./etc/kube/check_ready.sh app=pachd; do sleep 1; done
 	@echo "pachd launch took $$(($$(date +%s) - $(STARTTIME))) seconds"
 
+launch-enterprise: check-kubectl check-kubectl-connection install
```
brendoncarroll, line 220: I think this change is also in the other P.R.  Double check this merges cleanly before it goes in.

src/client/client.go
```diff
@@ -453,6 +453,34 @@ func NewForTest() (*APIClient, error) {
 	return client, nil
 }
 
+// NewEnterpriseClientForTest constructs a new APIClient for tests.
+func NewEnterpriseClientForTest() (*APIClient, error) {
```
brendoncarroll, line 457: If this is only supposed to be used in tests, I like the pattern of depending on the interface `testing.TB` and then not returning an error and just using the require package to short-circuit on errors.

We use the naming scheme `NewTest____(t testing.TB) _____` elsewhere.

brendoncarroll: Additionally you can register cleanup on the `testing.TB`

src/internal/deploy/cmds/cmds.go
```diff
@@ -557,8 +557,10 @@ func standardDeployCmds() []*cobra.Command {
 				opts.ExposeObjectAPI = true
 
 				// Set the postgres and etcd nodeports explicitly for developers
-				opts.PostgresOpts.Port = 32228
-				opts.EtcdOpts.Port = 32379
+				if !enterpriseServer {
+					opts.PostgresOpts.Port = 32228
```
brendoncarroll, line 561: Aren't these set elsewhere? I think they should be in constants.

brendoncarroll (changes requested):
A couple things to address.

For the test constructor this is what I had in mind.
https://github.com/pachyderm/pachyderm/blob/master/src/internal/storage/fileset/util.go#L18
https://github.com/pachyderm/pachyderm/blob/master/src/internal/storage/chunk/util.go#L18
https://github.com/pachyderm/pachyderm/blob/master/src/internal/dbutil/testdb.go#L25

src/client/client.go
actgardner: Agree, this is an exact copy of `NewForTest` which has the same problems. I'm going to add a TODO, I think we could have much simpler implementations for both of them and move them to testutil.

src/internal/deploy/cmds/cmds.go
actgardner: I switched the postgres port to `dbutil.DefaultPort`. In general our story around what connects to what in tests, etc. seems pretty chaotic - we seemingly have a bunch of scripts to port-forward postgres, run it in a docker container, and we run it with a nodeport so it's accessible at the minikube IP. More than reducing the use of constants in the codebase I would love to have a more consistent development flow, which might just eliminate these nodeports altogether.

brendoncarroll: I agree that we could parameterize some of the dbutil stuff so that you can better configure the test environment, Maybe through environment variables.  It would be nice if it wasn't hard coded to use the local forwarded port.

Aren't there 2 Postgres instances though?  They can't both be on the default port if we want to forward them both?

src/client/client.go
brendoncarroll: I don't think a package with test constructors is a good pattern generally. It produces dependency cycles.  If everything has a constructor in testutil then you can't import testutil for your unit test.  The testutil package currently has methods for generating random values, which has utility almost wherever you would write a test.  Something like serviceenv does make sense for black box testing, but my point is that constructors like this can't only live in a single package.

I didn't realize this was part of the public API, that is more of a problem.  I am generally fine with marking things as TODO, but this TODO would have to be addressed before 2.0, if we are serious about API compatibility. What is stopping this from just being moved into where the tests are?

`getUserMachineAddrAndOpt` <- Is that a better method to make public? Would users get value from that? 
Right now it seems like we are exporting 2 methods with no real value to users, which are just waiting to be depended on in a breaking way.

src/internal/deploy/cmds/cmds.go
actgardner: This is inside a `!enterpriseServer` block - it's only set for pachd. We have no reliance on the nodeport for the enterprise server tests, so we don't care what port gets assigned.

src/client/client.go
actgardner: The reason I copied the existing pattern and marked this TODO is to avoid having this discussion hold up this PR - I think we have technical debt around testing and the structure of the codebase, but my goal here is to try and land as much of the 2.0 feature work as possible so we can get feedback. I agree we should get this out of the public API before 2.0.

brendoncarroll: Sure, as long as we are on the same page about the path forward, this doesn't need to be a blocker right now.

brendoncarroll (approved):

PR #5837, opened 2021-03-16
Add CLI command to log in with an ID token

actgardner:
Add a way for users to convert an ID token from a trusted peer app into a pach auth token using the CLI. We'll use this in the IDE to configure pachctl.

### DIFF
--- a/src/internal/testutil/identity.go
+++ b/src/internal/testutil/identity.go
@@ -1,10 +1,13 @@
 package testutil
 
 import (
+	"context"
 	"net/http"
 	"net/url"
 	"strings"
 	"testing"
+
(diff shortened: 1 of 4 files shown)


src/internal/testutil/identity.go
```diff
@@ -127,6 +130,60 @@ func DoOAuthExchange(t testing.TB, loginURL string) {
 	require.NoError(t, err)
 }
 
+func GetOIDCTokenForTrustedApp(t testing.TB) string {
```
actgardner, line 133: This code is all just moved from the OIDC test into testutils, so we can reuse it in the CLI tests.

msteffen (approved):
LGTM

PR #5840, opened 2021-03-16
[Doc] Pipeline spec - side car - added details about cache size and behavior 

nadegepepin:


RaananHadar:
But what is the default cache size?

nadegepepin:
> But what is the default cache size?

If not explicitly specified, cache_size defaults to 64M.
Additionally, files > 25% of the cache size are not cached.
Will update the documentation today.

PR #5842, opened 2021-03-17
[1.12] Use a mutex to prevent writes to the logger after close

actgardner:
Users were seeing an error where writes to msgCh could happen after it was closed, causing a panic:
```
panic: send on closed channel
goroutine 273764 [running]:
 github.com/pachyderm/pachyderm/src/server/worker/logs.(*taggedLogger).Logf(0xc0008d8780, 0x2b0a484, 0x2, 0xc002ea1de0, 0x1, 0x1)
   /app/src/server/worker/logs/logger.go:218 +0x34d
 github.com/pachyderm/pachyderm/src/server/worker/logs.(*taggedLogger).Write(0xc0008d8780, 0xc003a36000, 0x131e, 0x8000, 0x131e, 0x0, 0x0)
   /app/src/server/worker/logs/logger.go:263 +0x216
 io.copyBuffer(0x7f5c466afdc0, 0xc0008d8780, 0x2e29660, 0xc008284028, 0xc003a36000, 0x8000, 0x8000, 0xc0007c8300, 0x0, 0x1)
   /usr/local/go/src/io/io.go:411 +0x206
 io.Copy(...)
   /usr/local/go/src/io/io.go:368
 github.com/pachyderm/pachyderm/src/server/pkg/exec.(*Cmd).writerDescriptor.func1(0x0, 0x0)
   /app/src/server/pkg/exec/exec.go:266 +0x65
 github.com/pachyderm/pachyderm/src/server/pkg/exec.(*Cmd).Start.func1(0xc0002262c0, 0xc003b74080)
  /app/src/server/pkg/exec/exec.go:382 +0x27
 github.com/pachyderm/pachyderm/src/server/pkg/exec.(*Cmd).Start
  /app/src/server/pkg/exec/exec.go:381 +0x570
```

This change adds an RW mutex - many producers can write to the channel concurrently, but Close will block until all the current writers have finished, then it will close the channel and set putObjClient to nil.

### DIFF
--- a/src/server/worker/logs/logger.go
+++ b/src/server/worker/logs/logger.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/gogo/protobuf/jsonpb"
@@ -57,6 +58,8 @@ type TaggedLogger interface {
 }
 
 type taggedLogger struct {
+	// Mutex is used to ensure no loggers are writing to msgCh when it's closed
+	sync.RWMutex
 	template  pps.LogMessage
 	stderrLog *log.Logger
 	marshaler *jsonpb.Marshaler
@@ -214,6 +217,8 @@ func (logger *taggedLogger) Logf(formatString string, args ...interface{}) {
 		return
 	}
 	fmt.Println(msg)
+	logger.RLock()
+	defer logger.RUnlock()
 	if logger.putObjClient != nil {
 		logger.msgCh <- msg + "\n"
 	}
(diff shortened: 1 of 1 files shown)


src/server/worker/logs/logger.go
```diff
@@ -268,8 +273,10 @@ func (logger *taggedLogger) Write(p []byte) (_ int, retErr error) {
 // statements to object storage.  Returns a pointer to the generated pfs.Object
 // as well as the total size of all written messages.
 func (logger *taggedLogger) Close() (*pfs.Object, int64, error) {
-	close(logger.msgCh)
+	logger.Lock()
+	defer logger.Unlock()
 	if logger.putObjClient != nil {
+		close(logger.msgCh)
```
msteffen, line 279: Just curious about how this code works, why does this go inside the `if` statement? IIUC, `putObjClient` is just for when we want to log to object storage (i.e. for writing logs to the stats commit, I think? Maybe that's wrong though)

msteffen (approved):
One question, but otherwise LGTM. This was fast!

src/server/worker/logs/logger.go
actgardner: The way the existing code worked, we used whether `putObjClient` is nil to signal whether Close has been called (the channel has been closed) - if a new log event arrived after Close, we would hit the `logger.putObjClient != nil` on line 222 and skip sending it. 

The race was because we close the channel, but then we block until the error group is finished, then we block until the client is done recieving messages, _then_ we set the client to nil. So there was potentially a period of time where the channel was closed and putObjClient was not nil, which led to a send on a closed channel.

I moved `close(logger.msgCh)` inside the block because otherwise calling Close would close the channel twice, which also panics.

PR #5843, opened 2021-03-17
Fixed an issue where mkdocs bug causes misalignment of code fenced tables

echohack:
Signed-off-by: echohack <git@echohack.app>


nadegepepin:
LGTM

PR #5845, opened 2021-03-17
Supporting rewriting the issuer for split-plane networking

actgardner:
It's necessary to support more flexible rewriting of the issuer hostname to support scenarios like "pachd and enterprise server in the same cluster, both port-forwarded". This might not be a common use case but customer team will need it to experiment.


msteffen (approved):
LGTM.

To be honest, I don't totally understand why `localhost_issuer` doesn't work if you're port-forwarding to pachd as well as the enterprise server (`issuer` is the path from the user's browser to the enterprise server, which is also the OIDC IdP, if I understand correctly?). I would be super interested in more color, but I don't want to block the PR on my curiosity.

actgardner:
Untitled drawing

Hopefully this helps? You're right, `issuer` is the path from the user's browser to the enterprise server. The override is the path from pachd to the enterprise server, which isn't localhost in deployment example 3. 

My hope would be that nobody actually does this in real life, and they do pattern 1, but I know customer team likes to have self-contained demos that don't depend on DNS, etc.

actgardner:
Closing this for now, it wasn't necessary for testing.

PR #5851, opened 2021-03-18
File API and Go client wrapper changes in 2.0

brycemcanally:
This PR has several changes that relate to the file API and Go client wrapper in 2.0.

The `PutFileClient` interface is replaced by the `ModifyFileClient` interface, which serves a similar purpose but has slightly different semantics. `ModifyFileClient` is not thread safe, so the general pattern for parallelism is to use multiple, parallel `ModifyFileClient`s (this is how parallel `PutFile*` calls are implemented). In general, the `ModifyFileClient` functions should be used when a user wants to create a large set of small modifications (put / delete). For large files, parallel `PutFile*` calls will give better performance.

The put file functionality is now an overwrite operation by default and the append functionality is exposed through an append flag. The PFS tests were updated such that the `PutFile` calls that needed the append functionality were updated to set the append flag, but for the PPS tests all of the `PutFile` calls were updated to set the append flag since the testing loop is much slower. It may make sense to unset this over time for the tests that don't need this.

The copy file overwrite functionality has been implemented in 2.0 and is now the default similarly to put file (the append functionality is available through an append flag).

The configuration of some of the client wrapper calls are now based on functional options. The calls that have been moved over are `PutFile*`, `DeleteFile` and `CopyFile`. The functional options that have similar prefixes (i.e. `WithAppend`) are suffixed by the corresponding call (i.e. `WithAppendPutFile`). We could maybe restructure the packages so we don't need this, but this seems like the way to go if these are going to be global with respect to all of the client wrappers.

Also, I cherry picked the bug fix from here: https://github.com/pachyderm/pachyderm/pull/5658


examples/opencv/goclient-example/opencv-example.go
```diff
@@ -35,15 +35,15 @@ func main() {
...
 		panic(err)
 	}
 
-	if err := c.PutFileURL("images", "master", "AT-AT.png", "http://imgur.com/8MN9Kg0.png", false, false); err != nil {
+	if err := c.PutFileURL("images", "master", "AT-AT.png", "http://imgur.com/8MN9Kg0.png", false); err != nil {
 		panic(err)
 	}
 
-	if err := c.PutFileURL("images", "master", "kitten.png", "http://imgur.com/g2QnNqa.png", false, false); err != nil {
+	if err := c.PutFileURL("images", "master", "kitten.png", "http://imgur.com/g2QnNqa.png", false); err != nil {
```
brendoncarroll, line 46: Someday CI will fail because a kitten cannot be retrieved from imgur.  But not today.

src/client/pfs_file.go
```diff
@@ -1,76 +1,70 @@
...
-	if err != nil {
-		return err
-	}
-	if err := mfc.AppendFileTar(overwrite, r, tag...); err != nil {
-		return err
-	}
-	return mfc.Close()
+// PutFileWriter sets up a writer for putting a file in PFS.
+func (c APIClient) PutFileWriter(repo, commit, path string, opts ...PutFileOption) (io.WriteCloser, error) {
+	return nil, errV1NotImplemented
```
brendoncarroll, line 27: What's the deal here?  If this is not in v2, just remove the method.

src/client/pfs_file.go
```diff
@@ -1,76 +1,70 @@
...
 }
 
 // ModifyFileClient is used for performing a stream of file modifications.
 // The modifications are not persisted until the ModifyFileClient is closed.
 // ModifyFileClient is not thread safe. Multiple ModifyFileClients
 // should be used for concurrent modifications.
-type ModifyFileClient struct {
-	client pfs.API_ModifyFileClient
-	modifyFileCore
+type ModifyFileClient interface {
```
brendoncarroll, line 57: If this interface is user facing, we need comments on these methods.

src/client/pfs_file.go
```diff
@@ -84,7 +78,7 @@ func (c APIClient) WithModifyFileClient(repo, commit string, cb func(*ModifyFile
 }
```
brendoncarroll, line 78: We need to close regardless of the error right? Is it a leak otherwise?

src/client/pfs_file.go
```diff
@@ -120,29 +111,24 @@ type modifyFileCore struct {
...
-			}
-			af.Tag = tag[0]
+		for _, opt := range opts {
+			opt(pf)
 		}
-		if err := mfc.sendAppendFile(af); err != nil {
+		if err := mfc.sendPutFile(pf); err != nil {
 			return err
 		}
 		if _, err := grpcutil.ChunkReader(r, func(data []byte) error {
```
brendoncarroll, line 129: Wouldn't it be cool if this also did content address chunks? So that the gRPC messages line up perfectly with the chunks we are going to create.  Maybe that's a good onboarding issue for someone.

src/client/pfs_file.go
```diff
@@ -174,35 +160,36 @@ func (mfc *modifyFileCore) maybeError(f func() error) (retErr error) {
...
-			AppendFile: req,
+		Modification: &pfs.ModifyFileRequest_PutFile{
+			PutFile: req,
 		},
 	})
 }
 
-// AppendFileTar appends a set of files from a tar stream.
-func (mfc *modifyFileCore) AppendFileTar(overwrite bool, r io.Reader, tag ...string) error {
+var errV1NotImplemented = errors.Errorf("V1 method not implemented")
```
brendoncarroll, line 171: See above, I think we should just break the API.  This was useful when v1 and v2 lived in the same codebase.

src/client/pfs_file.go
```diff
@@ -258,6 +236,14 @@ func (mfc *modifyFileCore) sendDeleteFile(req *pfs.DeleteFile) error {
 	})
 }
 
+// Close closes the ModifyFileClient.
+func (mfc *modifyFileClient) Close() error {
```
brendoncarroll, line 240: Can this be called more than once; what are the semantics here?

src/internal/storage/fileset/unordered_writer.go
```diff
@@ -170,8 +170,7 @@ func newUnorderedWriter(ctx context.Context, storage *Storage, memThreshold int6
 	return uw, nil
 }
 
-// Append appends a file to the file set.
-func (uw *UnorderedWriter) Append(p string, overwrite bool, r io.Reader, customTag ...string) error {
+func (uw *UnorderedWriter) Put(p string, appendFile bool, r io.Reader, customTag ...string) error {
```
brendoncarroll, line 173: Can we just separate `Put` and `Append`?

brendoncarroll: Especially since this is an internal API lets just be explicit, there is no continuity we need to keep for users with regard to what Put does.

src/server/pps/cmds/cmds.go
```diff
@@ -1294,54 +1294,49 @@ func buildHelper(pc *pachdclient.APIClient, request *ppsclient.CreatePipelineReq
...
-		f, err := progress.Open(srcFilePath)
-		if err != nil {
-			return errors.Wrapf(err, "failed to open file %q for source code in build step-enabled pipeline", srcFilePath)
-		}
-		defer func() {
-			if err := f.Close(); err != nil && retErr == nil {
-				retErr = err
+			destFilePath, err := filepath.Rel(buildPath, srcFilePath)
+			if err != nil {
+				return errors.Wrapf(err, "failed to discover relative path for %s", srcFilePath)
```
brendoncarroll, line 1313: both of these should be q or s

src/client/pfs_file.go
brycemcanally: I wanted to keep it around because it seems like a useful method that we would want to implement eventually, but it probably makes more sense to remove it for now. I'll go ahead and remove it and add a TODO to the 2.0 GitHub project.

src/client/pfs_file.go
brycemcanally: We don't want to close because that would indicate success. I'll add a cancel context so we clean up resources immediately.

src/client/pfs_file.go
brycemcanally: The same as the underlying gRPC stream. Should we support calling this more than once?

src/internal/storage/fileset/unordered_writer.go
brycemcanally: The way I was thinking about this is that it is the entry point into the storage layer, so the interface lines up with what the client needs (PFS). I guess we could change the interface to Add / Delete like the fileset writer and PFS would have a helper function that converts a Put to the appropriate Add / Delete, but I kind of think what we have is fine. I think we should revisit this later because I don't think it is particularly important now.

src/client/pfs_file.go
brycemcanally: Actually, this unfortunately needs to be handled through the pach client context. So, to do this properly you would have to do `pachClient := pachClient.WithCtx(cancelCtx)` to get the appropriate cleanup upon failure. I think these semantics make sense as compared to having an abort function, and they line up with the semantics of the underlying gRPC. It does suck that you cannot just pass in a context for the call though. We have discussed taking the context out of the pach client, but that might be a large undertaking.

brycemcanally: We could use this pattern under the hood for the scoped calls, but the unscoped ones will need to be handled by the user.

brendoncarroll: Most of the obj.Clients use the context for cancellation too.  Maybe we should take a context and mark it as TODO.  As long as it doesn't leak resources in practice, I think it's fine.

src/client/pfs_file.go
brendoncarroll: I guess the underlying semantics are fine.  It's not going to panic.  I want to be able to defer Close it without having the multiple close problems that we had in the storage layer.

brendoncarroll (approved):

src/client/pfs_file.go
brycemcanally: I think it is reasonable for multiple closes to be a client error, and for failing the call the context should be canceled without a close call. That lines up with the gRPC stream behavior.

brendoncarroll (approved):
LGTM

PR #5866, opened 2021-03-22
Change `obj.Client` interface to take readers and writers.  Adds `pacherr` package for common error types.

brendoncarroll:
Changes the `obj.Client` interface to have `Put` and `Get` which replace `Writer` and `Reader` respectively.
These new methods take a reader or writer as the last argument and manage filling or reading from it.  This is a better fit for our new storage layer because we only write modest sized objects (<20MB) to object storage.  There is little benefit to using the more complicated APIs that enable recovering during a large upload.  The clients also use the standard error types from `pacherr` to signal when it is appropriate to retry or an object cannot be found.


brendoncarroll:
The uniform client needs to ensure that NotExist errors from Delete are nil'd out.  This behavior is expected by the `kv` package.

src/internal/obj/api.go
```diff
@@ -0,0 +1,31 @@
...
+	// Put writes the data from r to an object at name
+	// It should error if the object already exists or we don't have sufficient
+	// permissions to write it.
+	Put(ctx context.Context, name string, r io.Reader) error
+
+	// Get writes the data for an object to w
+	// If `size == 0`, the reader should read from the offset till the end of the object.
+	// It should error if the object doesn't exist or we don't have sufficient
+	// permission to read it.
+	Get(ctx context.Context, name string, offset uint64, size uint64, w io.Writer) error
```
brycemcanally, line 19: I think we should remove the offset / size parameters. We may want this as a functional option later, but I kind of think we shouldn't have this because I cannot think of a good use case for it. Also, it requires more from object storage providers and / or client implementations.

src/internal/pacherr/retry.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"errors"
+	"net"
+	"time"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type Retry struct {
```
brendoncarroll, line 12: I'm thinking of renaming this `Transient`

Makefile
```diff
@@ -289,7 +289,7 @@ test-client:
 test-object-clients:
 	# The parallelism is lowered here because these tests run several pachd
 	# deployments in kubernetes which may contest resources.
-	go test -count=1 ./src/internal/obj/testing -timeout $(TIMEOUT) -parallel=2
+	go test -count=1 ./src/internal/obj/obj_intgtest -timeout $(TIMEOUT) -parallel=2
```
Tryneus, line 292: Out of curiosity - why was this renamed?   I have no clue what `intgtest` means - some googling got me to 'integration test'?  Can we not repeat `obj` in the path, at least?  Also, since the test code is now split between `obj` and `obj_intgtest`, this should probably run both or else we might miss some stuff in CI.

src/internal/pacherr/retry.go
Tryneus: If you do, can you include `Err` or `Error` in the name like ~every other error we have?

src/internal/obj/testsuite.go
```diff
@@ -0,0 +1,161 @@
...
+	if err := func() (retErr error) {
+		buf := bytes.NewBuffer(nil)
+		return c.Get(ctx, testObj, 0, 0, buf)
+	}(); err != nil {
+		return errors.Wrapf(err, "unable to read from object storage")
+	}
+	if err := c.Delete(ctx, testObj); err != nil {
+		return errors.Wrapf(err, "unable to delete from object storage")
+	}
+	// Try reading a non-existant object to make sure our IsNotExist function
```
Tryneus, line 95:
```suggestion
	// Try reading a non-existent object to make sure our IsNotExist function
```

src/internal/obj/obj_intgtest/tests.go
```diff
@@ -0,0 +1,83 @@
...
+	LocalClient     ClientType = "Local"
+)
+
+// testInterruption
+// Interruption is currently not implemented on the Amazon, Microsoft, and Minio clients
+//  Amazon client - use *WithContext methods
+//  Microsoft client - move to github.com/Azure/azure-storage-blob-go which supports contexts
+//  Minio client - upgrade to v7 which supports contexts in all APIs
+//  Local client - interruptible file operations are not a thing in the stdlib
+func testInterruption(t *testing.T, client obj.Client) {
```
Tryneus, line 44: Not that it works in the first place, but moving this here - this test will never run for local clients which seems incorrect to me - this behavior _should_ be supported by all clients in the long run.

Tryneus: Also, this appears to have removed the `t.Skip()` logic for all the broken clients - are these tests just failing?  This should probably be run in a sub-test regardless, rather than calling this directly from the client test suite.

Tryneus (commented):
A few questions about the testing stuff, everything else looks pretty good to me.

I'm mildly concerned that the new interface is copying the data more times which will hurt performance.

src/internal/pacherr/retry.go
brendoncarroll: Yeah does `TransientError` sound okay?

Makefile
brendoncarroll: I don't really care about the name, other than `testing` being a bad name because of the conflict with the standard library.  I thought it might be useful to distinguish these kinds of tests as integration tests (expensive to run, require a lot of config, use external services).  Maybe you would call them end-to-end tests.  Let me know if you have a strong preference for a particular name; I will change it.

I will also run the obj package tests.  The idea is that it is easier to iterate on the tests when they are in the obj package and running against the local client.  So that is where the test suite should live.  And then we call into the test suite + additional tests in the other package.

src/internal/obj/obj_intgtest/tests.go
brendoncarroll: I didn't call it for the broken clients. That is why it's not in the test suite.

brendoncarroll:
> I'm mildly concerned that the new interface is copying the data more times which will hurt performance.

@brycemcanally what do you think about the performance impact?

src/internal/obj/obj_intgtest/tests.go
Tryneus: Hmmm - this is called by the microsoft client which doesn't support contexts - I wonder if the test is broken.

Tryneus: Also, the old behavior was to have a `t.Skip()` which would provide some visibility into what tests were missing or broken and now the only indication is the comment above.  I'm not sure it's worth changing back at this point, but I think there was some value in it.

src/internal/pacherr/retry.go
Tryneus: sgtm

Makefile
Tryneus: I'd be fine with `integration_tests`.

I use `testing` packages to get around circular dependencies, usually, so they often don't need to be included (and the name conflict with the `testing` stdlib shouldn't matter).  In the case where we have behavior that other tests might want, it might be worth putting in a `testutil` package.

brendoncarroll: There is an idiom in go: `_test` packages that handle the circular dependency case.

https://tip.golang.org/cmd/go/#hdr-Test_packages
>  Test files that declare a package with the suffix "_test" will be compiled as a separate package, and then linked and run with the main test binary. 

That is the only time you are allowed to have two packages in the same directory.

So in the pfs server, I would be in favor of renaming the testing package to pfs_test and just have those tests in the server package.  But here I think we should keep them in a separate package because they are more costly to run, and are a different kind of test. We aren't just getting around the circular dependency.

I will change to `integration_tests`

src/internal/obj/amazon_client.go
```diff
@@ -272,23 +274,24 @@ func newAmazonClient(region, bucket string, creds *AmazonCreds, cloudfrontDistri
...
+		Body:            r,
+		Bucket:          aws.String(c.bucket),
+		Key:             aws.String(name),
+		ContentEncoding: aws.String("application/octet-stream"),
+	})
+	return err
 }
 
-func (c *amazonClient) Walk(_ context.Context, name string, fn func(name string) error) error {
+func (c *amazonClient) Walk(_ context.Context, name string, fn func(name string) error) (retErr error) {
```
brycemcanally, line 291: We probably should add TODOs for the implementations that do not respect the context.

src/internal/obj/amazon_client.go
```diff
@@ -374,17 +372,19 @@ func (c *amazonClient) Reader(ctx context.Context, name string, offset uint64, s
 		}
 		getObjectOutput, err := c.s3.GetObject(objIn)
 		if err != nil {
-			return nil, err
+			return err
 		}
 		reader = getObjectOutput.Body
 	}
-	return newBackoffReadCloser(ctx, c, reader), nil
+	defer reader.Close()
```
brycemcanally, line 379: Check error.

brycemcanally: Also, you could probably simplify the return below.

src/internal/obj/amazon_client.go
```diff
@@ -404,99 +401,45 @@ func (c *amazonClient) Exists(ctx context.Context, name string) bool {
...
-	}
-	return false
-}
-
-func (c *amazonClient) IsIgnorable(err error) bool {
-	return false
-}
-
-func (c *amazonClient) IsNotExist(err error) bool {
+func (c *amazonClient) transformError(err error, objectPath string) error {
```
brycemcanally, line 404: We should probably take a pass over these error checks to see if they still make sense and if we should add any new ones. I would like to get rid of the string comparison ones if we can. We can do this later, but it might be good to have a TODO for it.

src/internal/obj/limited_client.go
```diff
@@ -36,44 +36,20 @@ func NewLimitedClient(client Client, maxReaders, maxWriters int) Client {
 	}
 }
 
-func (loc *limitedClient) Writer(ctx context.Context, name string) (io.WriteCloser, error) {
-	if err := loc.writersSem.Acquire(ctx, 1); err != nil {
-		return nil, err
+func (loc *limitedClient) Put(ctx context.Context, name string, r io.Reader) error {
+	const cost = 1
```
brycemcanally, line 40: Shouldn't this be global?

src/internal/obj/local_client.go
```diff
@@ -37,50 +38,56 @@ func (c *localClient) normPath(path string) string {
...
 	}
-
 	file, err := os.Create(fullPath)
 	if err != nil {
-		return nil, errors.EnsureStack(err)
+		return err
 	}
-
-	return file, nil
+	defer file.Close()
```
brycemcanally, line 53: Check error. Why are there two closes?

src/internal/obj/local_client.go
```diff
@@ -37,50 +38,56 @@ func (c *localClient) normPath(path string) string {
...
-func (c *localClient) Reader(_ context.Context, path string, offset uint64, size uint64) (io.ReadCloser, error) {
+func (c *localClient) Get(_ context.Context, path string, offset uint64, limit uint64, w io.Writer) (retErr error) {
+	defer func() { retErr = c.transformError(retErr, path) }()
 	file, err := os.Open(c.normPath(path))
 	if err != nil {
-		return nil, errors.EnsureStack(err)
+		return errors.EnsureStack(err)
 	}
-
+	defer file.Close()
```
brycemcanally, line 66: Check error.

src/internal/obj/microsoft_client.go
```diff
@@ -39,16 +39,30 @@ func newMicrosoftClient(container string, accountName string, accountKey string)
 	return &microsoftClient{container: (&blobSvc).GetContainerReference(container)}, nil
 }
 
-func (c *microsoftClient) Writer(ctx context.Context, name string) (io.WriteCloser, error) {
-	return newMicrosoftWriter(ctx, c, name), nil
+func (c *microsoftClient) Put(ctx context.Context, name string, r io.Reader) error {
+	w := newMicrosoftWriter(ctx, c, name)
+	if _, err := io.Copy(w, r); err != nil {
+		w.Close()
```
brycemcanally, line 45: A TODO to remove this and respect the context would probably be good here. Since this operation is idempotent, closing a partial write and returning the error should be fine for now.

src/internal/obj/microsoft_client.go
```diff
@@ -39,16 +39,30 @@ func newMicrosoftClient(container string, accountName string, accountKey string)
...
-		return c.container.GetBlobReference(name).Get(nil)
+		r, err = c.container.GetBlobReference(name).Get(nil)
+	} else {
+		r, err = c.container.GetBlobReference(name).GetRange(&storage.GetBlobRangeOptions{Range: blobRange})
+	}
+	if err != nil {
+		return err
 	}
-	return c.container.GetBlobReference(name).GetRange(&storage.GetBlobRangeOptions{Range: blobRange})
+	defer r.Close()
```
brycemcanally, line 63: Check error.

src/internal/obj/minio_client.go
```diff
@@ -40,6 +41,89 @@ func newMinioClientV2(endpoint, bucket, id, secret string, secure bool) (*minioC
 	}, nil
 }
 
+func (c *minioClient) Put(ctx context.Context, name string, r io.Reader) (retErr error) {
+	defer func() { retErr = c.transformError(retErr, name) }()
+	wc := newMinioWriter(ctx, c, name)
+	if _, err := io.Copy(wc, r); err != nil {
```
brycemcanally, line 47: We should probably close here similarly to the microsoft client, then add a TODO to handle the context appropriately.

brycemcanally: Actually, I think we discussed removing the minio client in 2.0 and just having the amazon client for s3 compatible object storage providers. We might want to figure out if we still want to do that.

src/internal/obj/monkey_client.go
```diff
@@ -119,20 +79,8 @@ func (c *monkeyClient) Walk(ctx context.Context, dir string, walkFn func(name st
 
 // Exists wraps the existance check.
 func (c *monkeyClient) Exists(ctx context.Context, path string) bool {
+	if enabled && localRand.Float64() < failProb {
```
brycemcanally, line 82: The monkey client isn't supposed to give the wrong results, just to occasionally fail the retryable operations (putting / getting).

src/internal/obj/uniform_client.go
```diff
@@ -0,0 +1,58 @@
...
+
+var _ Client = &uniformClient{}
+
+// uniformClient is for ensuring uniform behavior across all the object clients
+type uniformClient struct {
+	c Client
+}
+
+func newUniformClient(c Client) Client {
+	if c == nil {
```
brycemcanally, line 19: I kind of think this shouldn't be a thing. We could error / panic if it is nil, but I think that would add a lot of bloat to our code if we followed that pattern. I think we should instead just use the pattern you suggested in the datum ID PR where we only create a wrapper client when the client being wrapped is not nil. That would allow us to get rid of these extra checks that just add bloat.

Makefile
brycemcanally: I think we should avoid the underscore in the package name like @Tryneus suggested.

src/internal/pacherr/common.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (e ErrNotExist) GRPCStatus() *status.Status {
+	return status.New(codes.NotFound, e.Error())
+}
+
+func IsNotExist(err error) bool {
+	target := ErrNotExist{}
+	return errors.As(err, &target) ||
+		os.IsNotExist(err)
```
brycemcanally, line 36: Why is `os.IsNotExist` included here?

src/internal/obj/util.go
```diff
@@ -28,24 +29,33 @@ func NewTestClient(t testing.TB) (Client, string) {
...
+		if err := wcb(pw); err != nil {
+			return pw.CloseWithError(err)
 		}
-	}()
-	_, err = io.Copy(wc, rc)
-	return err
+		return pw.Close()
+	})
+	eg.Go(func() error {
+		return rcb(pr)
```
brycemcanally, line 58: I think we need a close with error for the read side as well. That should be enough to make it such that we don't need to thread a context through this helper function.

src/internal/obj/util.go
```diff
@@ -28,24 +29,33 @@ func NewTestClient(t testing.TB) (Client, string) {
...
 // Copy copys an object from src at srcPath to dst at dstPath
 func Copy(ctx context.Context, src, dst Client, srcPath, dstPath string) (retErr error) {
-	rc, err := src.Reader(ctx, srcPath, 0, 0)
-	if err != nil {
-		return err
-	}
-	defer func() {
-		if err := rc.Close(); retErr == nil {
-			retErr = err
+	pr, pw := io.Pipe()
```
brycemcanally, line 32: Why not use `WithPipe` here?

brycemcanally:
I think the performance characteristics are fine. We already copy when downloading the chunks, but now this is occurring through the obj package rather than in the obj adapter. Either way, most of our abstractions are moving to be callback based so I think this will compose pretty well. As far as the pipe is concerned, I think that is going to work out fine as well. The cost of a goroutine per object is fine since the majority of the time will be spent setting up the object in object storage, and the cost of scheduling per write should be fine since we will perform a write per chunk (or the full file write if it is small, which will then force the time to be dominated by setting up the object in object storage) when we are pulling from the storage layer. Ideally we wouldn't need a pipe, but I think it is worth it for the simplicity we get in other places from the callback based interface. Also, I think the way you are handling it through a common helper function makes a lot of sense so we only have to get it right in one place.

src/internal/obj/testsuite.go
```diff
@@ -0,0 +1,161 @@
...
+		buf := &bytes.Buffer{}
+		require.NoError(t, client.Get(ctx, name, 0, 0, buf))
+		actualHash := pachhash.Sum(buf.Bytes())
+		require.Equal(t, expectedHash, actualHash)
+	})
+}
+
+// TestStorage is a defensive method for checking to make sure that storage is
+// properly configured.
+func TestStorage(ctx context.Context, c Client) error {
```
brycemcanally, line 78: We should probably run this during Pachd start up.

src/internal/obj/local_client.go
```diff
@@ -37,50 +38,56 @@ func (c *localClient) normPath(path string) string {
...
+	return file.Close()
 }
 
-func (c *localClient) Reader(_ context.Context, path string, offset uint64, size uint64) (io.ReadCloser, error) {
+func (c *localClient) Get(_ context.Context, path string, offset uint64, limit uint64, w io.Writer) (retErr error) {
+	defer func() { retErr = c.transformError(retErr, path) }()
 	file, err := os.Open(c.normPath(path))
 	if err != nil {
-		return nil, errors.EnsureStack(err)
+		return errors.EnsureStack(err)
```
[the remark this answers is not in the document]
brendoncarroll: Remove and add to the UniformClient

src/internal/obj/minio_client.go
brendoncarroll: https://github.com/pachyderm/pachyderm/issues/5919

src/internal/obj/monkey_client.go
brendoncarroll: Failure is equivalent to false here, so we should test for it.

src/internal/pacherr/common.go
brendoncarroll: To make it more generally useful.  I don't think this would ever return a false positive.

src/internal/obj/api.go
brendoncarroll: I think this will be difficult to add back.  Were we ever using this? Are there any use cases coming up?

src/internal/obj/testsuite.go
brendoncarroll: That has the potential to litter object storage with undeleted objects.  If object storage doesn't work, I'm sure we will find out about it soon enough.  What is the advantage of failing startup? Wouldn't you want to keep retrying and troubleshoot other things without having to restart?

brendoncarroll: Adding some sort of integrity checks to the storage layer, tracked and cleaned up by the usual mechanisms, might be worth doing.

src/internal/obj/api.go
brendoncarroll: I've thought about this more. I think it might become useful for PutFile from object storage.  It is easy to take out if we ever want to, but definitely more work to put back in.  I think we should keep it in for now.

brycemcanally: I disagree. I doubt it would be that useful for PutFile from object storage, but even if we still want to support it for that, the offset and size should not be required parameters for the interface. The default should be a full read and the size 0 being a full read shouldn't be a thing.

src/internal/obj/monkey_client.go
brycemcanally: Now that we are returning an error, we should be good. We need to avoid giving incorrect results and reporting no errors, because that is not the correct semantics for the monkey client.

src/internal/obj/testsuite.go
brycemcanally: We want to catch object storage issues as soon as possible and in a place where it is obvious what the root problem is. I think a preflight check for object storage makes sense, but I don't have a strong opinion about where it occurs. Having it in the storage layer and using the tracking mechanism to handle cleanup makes sense to me.

src/internal/obj/factory.go
```diff
@@ -184,36 +175,9 @@ var EnvVarToSecretKey = []struct {
...
-	// IsNotExist returns true if err is a non existence error
-	IsNotExist(err error) bool
-	// IsIgnorable returns true if the error can be ignored
-	IsIgnorable(err error) bool
-}
-
 // NewGoogleClient creates a google client with the given bucket name.
 func NewGoogleClient(bucket string, opts []option.ClientOption) (c Client, err error) {
-	defer func() { c = newCheckedClient(c) }()
+	defer func() { c = newUniformClient(c) }()
```
brycemcanally, line 180: This needs to be:

```
c, err := newGoogleClient(bucket, opts)
if err != nil {
  return nil, err
}
return newUniformClient(c), nil
```

Similar for all of the constructors below.

src/internal/pacherr/common.go
brycemcanally: I would interpret `pacherr.IsNotExist` as telling me whether the type is `pacherr.ErrNotExist`, not whether it is either `pacherr.ErrNotExist` or `os.ErrNotExist`.

brendoncarroll: The idea is that over time we migrate to this package for checking if an error is part of a certain class of errors.  If you want to know the type, you can use errors.Is, or cast from the interface.

brycemcanally: Okay, that might make sense. I think errors.Is is favored over these types of functions, but I guess it is fine for now.

brycemcanally (approved):
lgtm

PR #5871, opened 2021-03-23
Implement NoSkip in CreatePipelineRequest in 1.x

msteffen:
This also pulled in some recent commits from the 1.12.x branch (which is ahead of 1.13.x right now). Let me know if you want to try and merge those separately.

Also, just FYI, despite all the random stuff from other commits, the new worker code really did make the operative part of this change two lines long, which is kind of awesome.


Tryneus (approved):
LGTM

PR #5873, opened 2021-03-23
Log HeartbeatRequest without secret instead of logging the HeartbeatResponse

acohen4:
To avoid logging the enterprise token stored in the HeartbeatResponse, the HeartbeatRequest is logged instead with its `Secret` field value stripped.

### DIFF
--- a/src/server/license/server/api_server.go
+++ b/src/server/license/server/api_server.go
@@ -237,8 +237,9 @@ func stripSecretFromRequest(req *lc.HeartbeatRequest) *lc.HeartbeatRequest {
 }
 
 func (a *apiServer) Heartbeat(ctx context.Context, req *lc.HeartbeatRequest) (resp *lc.HeartbeatResponse, retErr error) {
-	a.LogReq(nil)
-	defer func(start time.Time) { a.pachLogger.Log(stripSecretFromRequest(req), nil, retErr, time.Since(start)) }(time.Now())
+	redactedRequest := stripSecretFromRequest(req)
+	a.LogReq(redactedRequest)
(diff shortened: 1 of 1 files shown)


src/server/license/server/api_server.go
```diff
@@ -230,9 +230,15 @@ func (a *apiServer) AddCluster(ctx context.Context, req *lc.AddClusterRequest) (
...
 }
 
+func stripSecretFromRequest(req *lc.HeartbeatRequest) *lc.HeartbeatRequest {
+	r := *req
+	r.Secret = ""
+	return &r
+}
+
 func (a *apiServer) Heartbeat(ctx context.Context, req *lc.HeartbeatRequest) (resp *lc.HeartbeatResponse, retErr error) {
 	a.LogReq(nil)
```
actgardner, line 240: Can you also update `LogReq` to log the request with the secret redacted?

acohen4: absolutely

src/server/license/server/api_server.go
```diff
@@ -230,9 +230,16 @@ func (a *apiServer) AddCluster(ctx context.Context, req *lc.AddClusterRequest) (
 	}, nil
 }
 
+func stripSecretFromRequest(req *lc.HeartbeatRequest) *lc.HeartbeatRequest {
+	r := *req
```
acohen4, line 234: based on the code in the Heartbeat method below, I'm making the assumption that a nil check on the request isn't necessary. Does that seem right @actgardner ?

actgardner: We don't conventionally do it, although it's not a great habit because a malformed request could cause the server to restart. We should probably just add a GRPC middleware to require non-nil requests everywhere.

actgardner (approved):

PR #5879, opened 2021-03-24
Move copy file functionality into modify file and change PPS workers to use the fileset API

### LINKED ISSUE #5898: ModifyFile consumes first message on stream without reading content
**What happened?**:

When I was working on the Python client, I was sending ModifyFile streams to add files to a commit. The first file in the stream was being skipped, because it turns out `ModifyFile` reads off the first message in the stream to get the commit without looking at the content: https://github.com/pachyderm/pachyderm/blob/master/src/server/pfs/server/api_server.go#L329

It's not clear whether it's valid to send an AppendFile in the first message of the stream. One way to make this foolproof would be to make the "commit" message a separate type in the `oneof` and require it to be the first message of the stream.

Either it should be an error to send an AppendFile/DeleteFile as the first message in the stream, or we could pass through the first message instead of dropping it.

**Environment?**:
- Others:

brycemcanally:
This PR moves the copy file functionality into modify file and changes the PPS workers to use the fileset API.

Copy file is no longer a separate gRPC endpoint. It is now a modification supported by modify file. I was initially considering keeping this separate and adding a new endpoint for copying filesets, but as I started to sketch out the changes it seemed like moving the copy file functionality into modify file made more sense and would be much cleaner. The main downside to this approach is that the performance characteristics might be a bit unclear since we have to serialize the buffered puts / deletes and resolve the copy (as compared to just buffering it with whatever puts / deletes are currently buffered). This should be addressable through documentation. This change also opens the door to copy file optimizations such as buffering a set of copies if they come from the same commit.

The PPS workers now operate on the fileset API (the missing piece was the copy file functionality). Workers now create filesets for the datumsets they process and return the fileset IDs back to the master. The master will then add the filesets to the output / meta commit. This is a better model for handling the output since we avoid potential issues with workers writing to commits, but failing to communicate the success back to the master.

Miscellaneous:
- `ModifyFileClient` -> `ModifyFile` and the removal of the `Close` function for the interface. We want to use this interface for both `ModifyFile` and `CreateFileset`, which have different `Close` signatures.
- Removed `datum.Client`. It is no longer needed since copy file is a part of modify file.
- Datums are now evenly distributed among the workers if the `ChunkSpec` (soon to be `DatumsetSpec`) isn't set.
- Removed Object API from auth interceptor.
- Slight change in copy file parameters (src, dst -> dst, src) to be more Go like.

TODO:
- ~~Merge https://github.com/pachyderm/pachyderm/pull/5816 and https://github.com/pachyderm/pachyderm/pull/5851 since this is based on those changes.~~

Fixes #5898 and #5901


src/server/pfs/s3/object.go
```diff
@@ -104,7 +104,7 @@ func (c *controller) CopyObject(r *http.Request, srcBucketName, srcFile string,
 		return "", s2.NotImplementedError(r)
 	}
 
-	if err = pc.CopyFile(srcBucket.Repo, srcBucket.Commit, srcFile, destBucket.Repo, destBucket.Commit, destFile); err != nil {
+	if err = pc.CopyFile(destBucket.Repo, destBucket.Commit, destFile, srcBucket.Repo, srcBucket.Commit, srcFile); err != nil {
```
brendoncarroll, line 107: How did this not break something?

src/server/pfs/server/paths.go
```diff
@@ -59,10 +57,11 @@ func cleanPath(x string) string {
...
 }
 
-func checkFilePath(path string) error {
-	path = filepath.Clean(path)
-	if strings.HasPrefix(path, "../") {
-		return errors.Errorf("path (%s) invalid: traverses above root", path)
-	}
-	return nil
-}
+// TODO: Readd?
```
brendoncarroll, line 60: What happened here?

src/server/worker/pipeline/transform/transform.proto
```diff
@@ -11,10 +11,11 @@ import "server/worker/datum/datum.proto";
 message DatumSet {
   // Inputs
   string job_id = 1 [(gogoproto.customname) = "JobID"];
-  string file_set = 2;
+  string fileset_id = 2;
```
brendoncarroll, line 14: 👍

src/server/worker/pipeline/transform/registry.go
```diff
@@ -441,32 +440,38 @@ func (reg *registry) processJobRunning(pj *pendingJob) error {
 			return err
 		}
 	}
-
-	// TODO: This is a hack to ensure that deletions are generated before any output is uploaded.
-	// This may be resolved by either explicitly generating deletes first (somewhat similar to this hack) or
```
brendoncarroll, line 446: So we are now doing the first method, which is "similar to this hack" ?

src/server/pfs/server/val_server.go
```diff
@@ -73,39 +73,40 @@ func (a *validatedAPIServer) SquashCommitInTransaction(txnCtx *txnenv.Transactio
...
-	}
-	if err := authserver.CheckRepoIsAuthorized(a.env.GetPachClient(ctx), dst.Commit.Repo.Name, auth.Permission_REPO_WRITE); err != nil {
-		return nil, err
-	}
-	if err := checkFilePath(dst.Path); err != nil {
-		return nil, err
-	}
-	return a.APIServer.CopyFile(ctx, request)
-}
+// TODO: Put this somewhere else, CopyFile gone.
```
brendoncarroll, line 76: Can we get rid of this? Why are we saving it? that's what git is for.

brendoncarroll (commented):
The actual protocol change itself looks good.  A couple small things and questions, otherwise.

src/server/pfs/s3/object.go
brycemcanally: Not sure I understand the question. The parameters for copy file are now (dst, src) rather than (src, dst) which means the arguments needed to be shuffled around.

src/server/pfs/server/paths.go
brycemcanally: The linter complained about it being unused. I think it probably makes sense to just throw it away for now similar to the copy file validation in the validation server (like you mentioned below).

src/server/pfs/server/val_server.go
brycemcanally: Yeah.

src/server/worker/pipeline/transform/registry.go
brycemcanally: Yeah, I kind of considered it a hack since it was the only way to ensure the correct ordering of deletes when workers were operating directly on the commits. It seems like a reasonable approach now because we have a path forward for caching the computation of the datums to process and in a lot of cases we need to know the number of datums to determine how to divide them into datumsets (when a user does not set a datumset spec). Also, it allows us to immediately add the fileset IDs to the commits since the deletions will already be applied.

brendoncarroll (approved):

PR #5882, opened 2021-03-25
1.13.x migration

nitinjainsj:
Most of this code is to support extract and restore. I need a sanity check on golden manifest. I was not expecting to add "replicationcontrollers/scale" during the release process. The only change I was expecting is the change in version numbers.

cc @jdoliner @robert-uhl


jdoliner (approved):
LGTM, the reason that `rcs/scale` was added to our permissions was to support the autoscaling feature.

robert-uhl (approved):
LGTM. I’ve opened a PR adding this to the upcoming Helm chart. Thanks for the heads-up!

PR #5884, opened 2021-03-25
Caching WhoAmI result in the Request Context 

acohen4:
Here's what I'm thinking for caching WhoAmI requests. 

Limitation - 'cache inserts' can only happen in Middleware (Interceptor)

### DIFF
--- a/src/internal/auth/handler.go
+++ b/src/internal/auth/handler.go
@@ -5,63 +5,58 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/client"
 )
 
-// an authHandler can optionally return a key-value that will be cached in the request's context
-type authHandler func(*client.APIClient, string) (*keyValue, error)
+// an authHandler can optionally return a username string that will be cached in the request's context
+type authHandler func(*client.APIClient, string) (string, error)
 
-type ctxKey string
+type ContextKey string
 
-type keyValue struct {
-	k ctxKey
-	v interface{}
-}
-
-const whoAmIKey ctxKey = "ctxKey_WhoAmI"
+const WhoAmIResultKey = ContextKey("WhoAmI")
 
 // authDisabledOr wraps an authHandler and permits the RPC if authHandler succeeds or
 // if auth is disabled on the cluster
 func authDisabledOr(h authHandler) authHandler {
-	return func(pachClient *client.APIClient, fullMethod string) (*keyValue, error) {
-		memoized, err := h(pachClient, fullMethod)
+	return func(pachClient *client.APIClient, fullMethod string) (string, error) {
+		username, err := h(pachClient, fullMethod)
 
 		// TODO: check this
 		if auth.IsErrNotActivated(err) {
-			return nil, nil
+			return "", nil
 		}
-		return memoized, err
+		return username, err
 	}
 }
 
 // unauthenticated permits any RPC even if the user has no authentication token
-func unauthenticated(pachClient *client.APIClient, fullMethod string) (*keyValue, error) {
-	return nil, nil
+func unauthenticated(pachClient *client.APIClient, fullMethod string) (string, error) {
+	return "", nil
 }
 
 // authenticated permits an RPC if auth is fully enabled and the user is authenticated
-func authenticated(pachClient *client.APIClient, fullMethod string) (*keyValue, error) {
+func authenticated(pachClient *client.APIClient, fullMethod string) (string, error) {
 	// consider the request authenticated if WhoAmI has a value
-	if v := pachClient.Ctx().Value(whoAmIKey); v != nil {
-		return nil, nil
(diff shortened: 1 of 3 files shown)


acohen4:
@actgardner is this along the line of what you were thinking?

src/internal/auth/handler.go
```diff
@@ -5,47 +5,63 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/client"
 )
 
-type authHandler func(*client.APIClient, string) error
+// an authHandler can optionally return a key-value that will be cached in the request's context
+type authHandler func(*client.APIClient, string) (*keyValue, error)
```
actgardner, line 9: Returning an arbitrary key-value pair might be overkill - I think we could just return the username as a string for now to make this a little simpler.

src/internal/auth/handler.go
```diff
@@ -5,47 +5,63 @@ import (
...
+
+type ctxKey string
+
+type keyValue struct {
+	k ctxKey
+	v interface{}
+}
+
+const whoAmIKey ctxKey = "ctxKey_WhoAmI"
```
actgardner, line 18: Go's contexts kind of suck because the keys aren't namespaced or anything - using string keys it's possible to have collisions with any package that touches a context. One way to get around this is to exploit the fact that Go's equality comparison includes the fully-qualified type - you can make a type alias like `type contextKey string` and it won't be equal to a string with the same value, or even a type called contextKey in another package.

Example: https://play.golang.org/p/OPOG_btut_O

src/internal/auth/interceptor.go
```diff
@@ -250,10 +250,18 @@ func (i *Interceptor) InterceptUnary(ctx context.Context, req interface{}, info
 		return nil, fmt.Errorf("no auth function for %q, this is a bug", info.FullMethod)
 	}
 
-	if err := a(pachClient, info.FullMethod); err != nil {
+	authResp, authErr := a(pachClient, info.FullMethod)
+
+	if authErr != nil {
 		logrus.Errorf("denied unary call %q\n", info.FullMethod)
```
actgardner, line 256: If we get back the username we should include it in these logs (or "unauthenticated" if the user wasn't authed at all).

src/internal/auth/interceptor.go
```diff
@@ -266,7 +274,7 @@ func (i *Interceptor) InterceptStream(srv interface{}, stream grpc.ServerStream,
 		return fmt.Errorf("no auth function for %q, this is a bug", info.FullMethod)
 	}
 
-	if err := a(pachClient, info.FullMethod); err != nil {
+	if _, err := a(pachClient, info.FullMethod); err != nil {
```
actgardner, line 277: We should also be able to set the context for the incoming stream, so this works for streaming and unary calls.

src/internal/auth/interceptor.go
```diff
@@ -250,25 +281,48 @@ func (i *Interceptor) InterceptUnary(ctx context.Context, req interface{}, info
...
 
-	if err := a(pachClient, info.FullMethod); err != nil {
-		logrus.Errorf("denied streaming call %q\n", info.FullMethod)
+	username, err := a(pachClient, info.FullMethod)
+
+	if err != nil {
+		logUser := "unauthenticated"
+		if username != "" {
+			logUser = username
+		}
```
actgardner, line 318: it might be possible to dry this up a little with a function, not really a big deal

```
func nameOrUnauthenticated(name string) string {
    if name == "" {
        return unauthenticated
    }
    return name
}
```

src/internal/auth/handler.go
```diff
@@ -1,51 +1,67 @@
...
 	"github.com/pachyderm/pachyderm/v2/src/client"
 )
 
-type authHandler func(*client.APIClient, string) error
+// an authHandler can optionally return a username string that will be cached in the request's context
+type authHandler func(*client.APIClient, string) (string, error)
+
+type ContextKey string
+
+const WhoAmIResultKey = ContextKey("WhoAmI")
```
actgardner, line 14: I think it makes sense to make these package-private, and expose a "getWhoAmI" method that accepts a context and returns the username if it's set? That way we know it can only be set in this package.

src/internal/auth/interceptor.go
```diff
@@ -235,6 +236,36 @@ func NewInterceptor(env *serviceenv.ServiceEnv) *Interceptor {
 	}
 }
 
+// we use ServerStreamWrapper to set the stream's Context with added values
+type ServerStreamWrapper struct {
```
actgardner, line 240: 👍

acohen4:
@actgardner could you give this another pass? I responded to your latest suggestions and did a little more DRYing up

actgardner (approved):
LGTM 👍

PR #5897, opened 2021-03-29
Add postgres implementation of the collections package

Tryneus:
This PR adds an implementation of the `collections` and `watch` package that works with postgres, as well as expanding the tests in the `collections` package to ensure feature and behavior parity between the two implementations.

There are still a few missing tests that I'm working on, but this should be mostly reviewable.  I'll do a pass today to add comments and a `README.md`.


src/server/pfs/server/api_server.go
```diff
@@ -255,7 +255,7 @@ func (a *apiServer) SubscribeCommit(request *pfs.SubscribeCommitRequest, stream
 func (a *apiServer) ClearCommit(ctx context.Context, request *pfs.ClearCommitRequest) (_ *types.Empty, retErr error) {
 	func() { a.Log(request, nil, nil, 0) }()
 	defer func(start time.Time) { a.Log(request, nil, retErr, time.Since(start)) }(time.Now())
-	return nil, a.driver.clearCommit(a.env.GetPachClient(ctx), request.Commit)
+	return &types.Empty{}, a.driver.clearCommit(a.env.GetPachClient(ctx), request.Commit)
```
Tryneus, line 258: Unrelated: this was causing this RPC to crash.

src/auth/auth.proto
```diff
@@ -86,6 +86,8 @@ message TokenInfo {
     GET_TOKEN = 2;  // returned by GetToken()--revokeable.
   }
   TokenSource source = 2;
+
+  string hash = 3;
```
brendoncarroll, line 90: Is this relevant? Is this just because there was work done on master?

src/internal/collection/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Collections are a library for storing protobufs in a persistent database with a simple interface for reading, writing, and watching for changes.  There are two existing implementations, one for etcd and one for postgres.
+
+A collection can be constructed via `NewEtcdCollection` or `NewPostgresCollection`, depending on the backend to be used.  This function must be given:
+ * a template protobuf `Message`, which will be clones for use in serialization and deserialization
+ * a set of fields to create secondary indexes for (used by `GetByIndex` and `WatchByIndex`)
+
+## API
+
+The collection API is split into two means of accessing a collection - `ReadOnly` and `ReadWrite`.  A `ReadOnly` collection does not make any consistency guarantees, while a `ReadWrite` collection is guaranteed to do all reads and writes within a single transaction, although this comes with trade-offs.
```
brendoncarroll, line 11: So not even ReadCommitted? I could be reading data from transactions in progress?

src/internal/collection/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### ReadOnly
+
+A `ReadOnlyCollection` is constructed with a context that can be used for cancellation.
+
+Operations:
```
brendoncarroll, line 17: This is excellent documentation, but all of it is less discoverable than it would be on the interface methods.  That shows up in autocomplete and godoc.  I don't have strong opinions on how to keep those in sync, but it's worth mentioning.

src/internal/collection/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+The postgres implementation is written to mirror the functionality of the etcd implementation rather than to provide an ORM-style table of the protobuf fields inside a database.
+
+### sqlx.Tx
+
+The `sqlx.Tx` can be constructed via `NewSQLTx`.  This will occupy a sql session for its lifetime.  Constructing tehe `sqlx.Tx` will result in a `BEGIN;` at the start, and a `COMMIT;` at the end of the callback, unless an error occurs, in which case a `ROLLBACK;` will be issued.  This will reattempt up to three times in the case of a commit error.
+
+### Tables
+
+A table will be created for each postgres collection.  Construction of the collection will check that the table exists, as well as all relevant triggers and indexes.  If any are missing, they will be created.
```
brendoncarroll, line 66: I don't think this is a great idea.  I can imagine difficult to fix bugs coming from lazily creating the tables.  I think we should split that functionality out of the constructor, and have the constructor just fail if the table doesn't exist.  It would be expected that the setup code has already been called before an STM can be created.

src/internal/collection/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Triggers
+
+Two triggers exist on each postgres collection table, `update_modified_trigger` and `notify_watch_trigger`.
+
+`update_modified_trigger` will modify the row before any write to update the `updatedat` column to the current time.
+
+`notify_watch_trigger` will issue `NOTIFY` commands to one or several channels in case any clients are running watches on the collection.  For table-wide watches, the channel is straightforwardly `pwc_<table>`.  For point-watches, the trigger will also notify one channel for each index on the collection.  The channel name must depend on the value of the field being indexed so that clients do not have to discard O(n) events per interesting event.  Unfortunately, channel names must be valid postgres identifiers which puts several limitations on the channel name, so the value must be hashed.  This still leaves the possibility open for a collision in the channel name, so the client must still check that the event belongs to the watch in question (see [Listeners](#listeners) for details).
```
brendoncarroll, line 83: I think there is a bug with pq and names longer than 31 characters.  Not sure if you've run into that.

src/internal/collection/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Each `Watch` operation will listen on a postgres channel according to the watch type:
+ * `Watch`, `WatchF` - listens to the table-wide `pwc_<table>` channel
+ * `WatchOne`, `WatchOneF`, `WatchByIndex` - listens to a hashed channel `pwc_<table>_<hash>` based on the index key
+
+ Once the `LISTEN` has returned, events must be buffered until the user has read out the current state of objects in the collections.  The `Watch` operation will then issue a `Get` or `List` operation to load the current state and keep track of the last modified timestamp.  The existing items will be provided to the `Watch` as `EventPut` events, and once these are done, the buffered events will be forwarded to the `Watch`, filtering out any that arrived before loading the existing items.
```
brendoncarroll, line 101: What is the failure mode for this? Does postgres drop from the front or back of the queue. That sort of constrains what we can do.

src/internal/collection/collection_test.go
```diff
@@ -1,699 +1,1352 @@
...
+		require.True(t, ok, "row '%s' was present but was not expected", k)
+	}
 }
 
-func TestDeletePrefix(t *testing.T) {
-	etcdClient := getEtcdClient()
-	uuidPrefix := uuid.NewWithoutDashes()
+func collectionTests(
+	parent *testing.T,
+	newCollection func(context.Context, *testing.T) (ReadCallback, WriteCallback),
```
brendoncarroll, line 171: 👍

src/internal/collection/collection_test.go
```diff
@@ -1,699 +1,1352 @@
...
+				for _, cb := range cbs {
+					if err := writer(ctx, cb); err != nil {
+						return err
+					}
+				}
+				return nil
+			})
+
+			t.Cleanup(func() {
+				require.NoError(t, eg.Wait())
```
brendoncarroll, line 390: Are you allowed to fail the test during the Cleanup call like this?

src/internal/collection/collection_test.go
```diff
@@ -1,699 +1,1352 @@
...
+					return errutil.ErrBreak
+				}
+				return nil
+			}
 		}
 
-		jobInfos.Put(j1.Job.ID, j1)
-		if err := jobInfos.Get(j1.Job.ID, job); err != nil {
-			return err
+		collectEventsChannel := func(watcher watch.Watcher, count int, out *[]*watch.Event) {
```
brendoncarroll, line 405: I think this should return a channel and then the callback is out of the picture and you can test by reading a certain number from the channel or timing out.

src/internal/collection/collection_test.go
```diff
@@ -1,699 +1,1352 @@
...
+				count, err = emptyRead.Count()
+				require.NoError(t, err)
+				require.Equal(t, int64(0), count)
+			})
+		})
+
+		// 'Watch' functions with callbacks don't have a good hook to trigger
+		// events. This is a bit racy, but we can sleep a bit to let the listen get
+		// started, then trigger some writes.
+		asyncWrite := func(t *testing.T, writer WriteCallback, cbs ...func(rw col.ReadWriteCollection) error) {
```
brendoncarroll, line 376: This pattern seems like it will become a source of flaky tests.  I think the way to do this is set up a listener that copies to a channel.  Then write in a goroutine, and receive in the main.  You can also select from a context so that it times out in the failing case.

src/internal/collection/postgres_collection.go
```diff
@@ -0,0 +1,711 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/version"
+)
+
+const (
+	modifiedTriggerName  = "update_modified_trigger"
+	modifiedFunctionName = "update_modified_time"
+	watchTriggerName     = "notify_watch_trigger"
+	watchBaseName        = "pwc" // "Pachyderm Watch Channel"
+	indexBaseName        = "idx"
+	pgIdentBase64Values  = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_$"
```
brendoncarroll, line 31: That dollar sign seems like a problem.  We can't use `-`?  Postgres uses $ for passing arguments.

src/internal/collection/postgres_listener.go
```diff
@@ -0,0 +1,395 @@
...
+		value:      value,
+	}
+}
+
+func (pw *postgresWatcher) Watch() <-chan *watch.Event {
+	return pw.c
+}
+
+func (pw *postgresWatcher) Close() {
+	// Close the 'done' channel to interrupt any waiting writes - because of this,
```
brendoncarroll, line 67: This comment needs to be visible to users either on the interface or above the method.  Alternatively make it safe using sync.Once.

src/internal/collection/types.go
```diff
@@ -81,37 +82,53 @@ type ReadWriteCollection interface {
...
-// ReadWriteIntCollection is a ReadonlyCollection interface specifically for ints.
-type ReadWriteIntCollection interface {
-	Create(key string, val int) error
-	Get(key string) (int, error)
-	Increment(key string) error
-	IncrementBy(key string, n int) error
-	Decrement(key string) error
-	DecrementBy(key string, n int) error
-	Delete(key string) error
+type PostgresReadWriteCollection interface {
```
brendoncarroll, line 88: Won't these only ever be implemented by one type? They probably don't need to be interfaces.

src/internal/collection/README.md
Tryneus: I haven't seen any issues with this, maybe because we use their listener object directly rather than doing this through the `sql` package.

src/internal/collection/README.md
Tryneus: That sounds fine to me - I suppose I can just move all the initialization into a migration, but it becomes a bit more cumbersome to declare a new collection.

src/internal/collection/README.md
Tryneus: If we want identical semantics to etcd, we should use `Serializable` - right now it's just using the default of `ReadCommitted`, but I think we should probably move it to `ReadRepeatable` - I don't think we need to guarantees of `Serializable` in postgres.

src/auth/auth.proto
Tryneus: I made a slight change to the collection interface that we no longer return the key in various callbacks - individual collections are expected to be able do derive the key from the stored protobuf (if they need it).  This might be a bad idea and maybe I should revert this.

src/internal/collection/README.md
Tryneus: I don't think postgres drops items from the queue.  Locally, if a watcher is not keeping up with its events, the channel will fill up and we will error out if the write ever blocks.  This lets us keep the listening goroutine rather tight, so it should have no problem keeping up with the listen events from postgres.

src/internal/collection/types.go
Tryneus: They don't need to be, but this file serves as a good entry point to see the differences between the interfaces.  There should probably be better doc strings for each call (like you mentioned in the `README.md`), so I can add those in.

src/internal/dbutil/option.go
```diff
@@ -1,27 +1,27 @@
 package dbutil
 
 // Option configures a DB.
-type Option func(*dBConfig)
+type Option func(*dbConfig)
```
brendoncarroll, line 4: lol

src/internal/dbutil/testdb.go
```diff
@@ -1,79 +0,0 @@
-package dbutil
```
brendoncarroll, line 1: I think this is gone in master? I'm not really sure what's going on with this diff.

src/internal/deploy/assets/postgres.go
```diff
@@ -51,6 +53,70 @@ type PostgresOpts struct {
 	Port int32
 }
 
+// WritePostgresAssets generates all of the postgres-related parts of the
+// kubernetes manifest according to the given options and writes it into the
+// given encoder.
+func WritePostgresAssets(encoder serde.Encoder, opts *AssetOpts, objectStoreBackend Backend,
```
brendoncarroll, line 59: 👍 This might be worth just merging to master in it's own P.R.

src/internal/keycache/cache.go
```diff
@@ -10,21 +10,20 @@ import (
...
 	col "github.com/pachyderm/pachyderm/v2/src/internal/collection"
-	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/watch"
 )
 
 // Cache watches a key in etcd and caches the value in an atomic value
 // This is useful for frequently read but infrequently updated values
 type Cache struct {
-	c            col.Collection
+	c            col.EtcdCollection
```
brendoncarroll, line 19: Is this because we don't have revisions in Postgres? What functionality is missing?

src/internal/testpachd/mock_env.go
```diff
@@ -19,59 +21,44 @@ type MockEnv struct {
...
-func WithMockEnv(cb func(*MockEnv) error) error {
-	return testetcd.WithEnv(func(etcdEnv *testetcd.Env) (err error) {
-		// Use an error group with a cancelable context to supervise every component
-		// and cancel everything if one fails
-		ctx, cancel := context.WithCancel(etcdEnv.Context)
-		defer cancel()
-		eg, ctx := errgroup.WithContext(ctx)
+// NewMockEnv constructs a MockEnv for testing, which will be destroyed at the
+// end of the test.
+func NewMockEnv(t *testing.T) *MockEnv {
```
brendoncarroll, line 26: 👍

brendoncarroll (commented):
I finished a first pass of this.  I want to dig into the implementation more, but I've been looking at this a while.  It is a pretty large P.R. and I think there are some sections that could be broken out.  Mainly because there are changes that widely affect the codebase, and they might be relevant/of concern to the other team members, who could more reasonably be expected to review them separately.

- The deploy/assets change is a pretty clean refactor; I think that is ready to merge. Maybe the postgres deployment test stuff can go with it.
- The NewEnv and NewRealEnv look good and are pretty self contained.  The t.Cleanup change has been a long time coming.

I have some concerns about the `testutil` stuff.  The usefulness of that package actually decreases the more constructors you put in it.  Once testutil references a package, it is useless to that package because it would create a circular dependency to import testutil.
I vote against moving test constructors in there, and would prefer to revert the db, chunk, and fileset test constructor changes.

src/internal/collection/postgres_collection.go
brendoncarroll: What about base32?  128bits/16bytes of entropy takes up ~25 bytes. That's 64bit collision resistance.  That should be enough right?  With the twc_ it's still less than the weird 31 character issue I was running into.

src/auth/auth.proto
brendoncarroll: If every collection has a primary key then I would say revert.  But if they don't have a primary key conceptually, and we just choose one in postgres then this is a better fit for the model.

src/internal/collection/postgres_collection.go
Tryneus: Oh dammit - this was from and old implementation anyway and I need to remove it - I started using hex-encoded md5 sums instead.

Tryneus:
> I have some concerns about the `testutil` stuff. The usefulness of that package actually decreases the more constructors you put in it. Once testutil references a package, it is useless to that package because it would create a circular dependency to import testutil.
I vote against moving test constructors in there, and would prefer to revert the db, chunk, and fileset test constructor changes.

Regarding this - moving stuff to the `testutil` package was actually a follow-up on a comment you made - that you can work around circular dependencies by putting tests into a `_test` sub-package which is explicitly allowed by the golang spec.  I think this is the way forward, rather than cluttering up libraries with exported test code, we put the test code into `testutil` and use `_test` packages.

src/internal/keycache/cache.go
Tryneus: This has been moved over to `PostgresCollection` in the `postgres-intergration` branch - this branch isn't supposed to move anything over to postgres collections yet.

src/internal/collection/collection_test.go
Tryneus: I think I have an alternative here that will be a bit more readable and concise, will upload it tonight.  Going to merge back all the changes to `src/internal/collection` that came up in the integration branch.

src/internal/collection/collection_test.go
Tryneus: Reworked these tests to get rid of `asyncWrite`, they are now in `watch_test.go` and I deduped the tests between the channel-based and callback-based API.

brendoncarroll:
> we put the test code into `testutil` and use `_test` packages.

This means we can only do black-box testing.  I don't think you can share private methods or inspect private variables when the tests are in a separate package.

PR #5892, opened 2021-03-26
[doc] [skip-ci] document no-skip setting

echohack:
Signed-off-by: echohack <git@echohack.app>

TODO:

- Add additional language around this setting per @nadegepepin 
- Add to version 1.13.x folder
- Push to version 1.13.x branch

PR #5903, opened 2021-03-29
Rename pipeline option NoSkip to ReprocessSpec

Tryneus:
`NoSkip` was confusing and underspecified - `ReprocessSpec` describes how datums are chosen to be reprocessed.  There are two possible values currently, `"until_success"` (the default) and `"every_commit"` (equivalent to `NoSkip=true`).  I tried making these enums in the protobuf, but we don't have good structures for dealing with enums in the pipeline spec, so it was simpler to leave them as strings.

As for the values themselves, we can adjust these, but if we're going to we should do it now.  Another suggestion we had for `"until_success"` was `"only_on_request"`, which I decided against because we do actually attempt to reprocess errored datums even if it's not explicitly requested.  Curious if anyone else has a preference.


echohack (approved):
Thank you so much for making these changes. Although at first glance this might seem a bit `SuperVerbosityMode`, this is much improved over the previous `no_skip` negative boolean in a negative use case (skipping behavior), and much easier for our users to understand.

msteffen (commented):
I think `src/server/pkg/ppsutil/util.go` and `src/server/pachyderm_test.go` may need to be updated as well? (just based on their presence in https://github.com/pachyderm/pachyderm/pull/5871/files)

Tryneus:
@msteffen - yes, that will need to be addressed in the `1.13.x` branch (but I'm getting this merged first so we can sort out the naming argument).

msteffen:
> @msteffen - yes, that will need to be addressed in the `1.13.x` branch (but I'm getting this merged first so we can sort out the naming argument).

To clarify, I think this might currently be a bug in master, but re: `src/server/pkg/ppsutil/util.go`, I think if [PipelineReqFromInfo](https://github.com/pachyderm/pachyderm/blob/master/src/internal/ppsutil/util.go#L303-L328) doesn't set `ReprocessSpec` then `pachctl extract pipeline` drops the field (I could be misunderstanding the code though—apologies if so!)

nadegepepin:
Documented in PR https://github.com/pachyderm/pachyderm/pull/5892 and https://github.com/pachyderm/pachyderm/pull/5906

Tryneus:
> > @msteffen - yes, that will need to be addressed in the `1.13.x` branch (but I'm getting this merged first so we can sort out the naming argument).
> 
> To clarify, I think this might currently be a bug in master, but re: `src/server/pkg/ppsutil/util.go`, I think if [PipelineReqFromInfo](https://github.com/pachyderm/pachyderm/blob/master/src/internal/ppsutil/util.go#L303-L328) doesn't set `ReprocessSpec` then `pachctl extract pipeline` drops the field (I could be misunderstanding the code though—apologies if so!)

Good catch - fixed!

echohack:
Tossing the 1.13.x label on here so this doesn't get forgotten to merge back into 1.13.x branch once it gets merged. :)

jdoliner:
Will the datum get reprocessed if you update the pipeline? If so I think `every_job` might be better than `every_commit`. Technically updating the pipeline does create a commit but users mostly don't think of it that way.

Tryneus:
> Will the datum get reprocessed if you update the pipeline? If so I think `every_job` might be better than `every_commit`. Technically updating the pipeline does create a commit but users mostly don't think of it that way.

sgtm - updated

PR #5912, opened 2021-03-30
Define the Prometheus metrics port

robert-uhl:
Deployment port names are limited by Kubernetes to 15 characters, so use
`prom-metrics` for both the deployment and service port names.
Cf. https://github.com/pachyderm/helmchart/pull/42


brendoncarroll (approved):
LGTM

PR #5918, opened 2021-03-31
[Doc] [skip-ci] Fixed error on major/minor/point release terminology in Migration and Update doc

nadegepepin:


dgeorg42:
Can we change the following:
`Migrations - A migration that you must perform to move between major or minor releases, such as between version 1.11.7 and 1.12.5.`

To:
`Migrations - A migration must be performed when you are moving between major releases, such as moving from 1.13.x to 2.0
`

PR #5921, opened 2021-03-31
Refactor serviceenv to an interface

actgardner:
Refactor serviceenv to an interface so we can stub out the client constructor methods.


msteffen (approved):
LGTM, this is awesome!

PR #5922, opened 2021-03-31
Few changes that should improve put file performance for small files

brycemcanally:
This PR will hold a few changes that should improve put file performance. I'll update this comment with more information once we are ready to commit to a set of changes.

Current changes:
- Added buffering of single message in code that receives PutFile messages.
- Removed rpc from PFS gRPC server to Block gRPC server when doing a put file. PFS gRPC server now makes calls directly on the Block gRPC server struct.
- Fixed bug where pachctl put file with the recursive flag was not respecting the parallelism limiter in terms of the goroutines being spawned. (This will probably be the most impactful for users seeing extremely slow put file recursive calls).
- Upgraded gRPC-Go to 1.36.1.

### DIFF
--- a/src/server/pfs/cmds/cmds.go
+++ b/src/server/pfs/cmds/cmds.go
@@ -1462,12 +1462,14 @@ func putFileHelper(c *client.APIClient, pfc client.PutFileClient,
 				return nil
 			}
 			childDest := filepath.Join(path, strings.TrimPrefix(filePath, source))
+			limiter.Acquire()
 			eg.Go(func() error {
+				defer limiter.Release()
 				// don't do a second recursive 'put file', just put the one file at
 				// filePath into childDest, and then this walk loop will go on to the
 				// next one
 				return putFileHelper(c, pfc, repo, commit, childDest, filePath, false,
-					overwrite, limiter, split, targetFileDatums, targetFileBytes,
+					overwrite, limit.New(0), split, targetFileDatums, targetFileBytes,
 					headerRecords, filesPut)
 			})
 			return nil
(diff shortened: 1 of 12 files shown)


src/server/pfs/cmds/cmds.go
```diff
@@ -1462,12 +1462,14 @@ func putFileHelper(c *client.APIClient, pfc client.PutFileClient,
...
 			childDest := filepath.Join(path, strings.TrimPrefix(filePath, source))
+			limiter.Acquire()
 			eg.Go(func() error {
+				defer limiter.Release()
 				// don't do a second recursive 'put file', just put the one file at
 				// filePath into childDest, and then this walk loop will go on to the
 				// next one
 				return putFileHelper(c, pfc, repo, commit, childDest, filePath, false,
-					overwrite, limiter, split, targetFileDatums, targetFileBytes,
+					overwrite, limit.New(0), split, targetFileDatums, targetFileBytes,
```
jdoliner, line 1472: So for this to work it looks like it depends on the fact that this callsite will never use the limiter because `recursive` is false. That feels a little bit fragile to me. It's definitely worth a note that it shouldn't be changed but maybe it also makes sense to pass a `nil` limiter? I think a panic is probably preferable to a deadlock, especially in a command like `put file` which can block for while so the user might not figure out they're deadlocked and just assume it's taking a really long time to upload.

jdoliner (approved):
One small change requested, after that LGtm

src/server/pfs/cmds/cmds.go
brycemcanally: Creating a limiter with 0 returns a no op limiter which has no limit. The recursive calls will use the passed in limiter but it will be a no op. The top level limiter is the one that matters. It would probably be clearer if there was a separate constructor for this rather than passing in 0 to new though. I think this is fine for now since it keeps the change small and is pretty understandable if you know that 0 means no op.

PR #5923, opened 2021-03-31
Fix enterprise check loophole with update pipeline

### LINKED ISSUE #5916: 1.13.0 rc1: pachctl update pipeline Doesn't Enforce Parallelism Limits
**What happened?**:

I had an existing pipeline running with no parallelism spec set. I edited the JSON file for the pipeline and added a parallelism_spec with constant = 200. I then ran `pachctl update pipeline` and was able to successfully update the pipeline to that level of parallelism, despite not having an Enterprise token.

I expected to receive an error message informing me that parallelism greater than 8 requires an Enterprise token.

Do what I did above. :)  Create a pipeline with parallelism < 8, then update the spec to have paralelism > 8 and run `pachctl update pipeline`

Nope. :)

**Environment?**:
- Kubernetes version (use `kubectl version`): 1.18.6
- Pachyderm CLI and pachd server version (use `pachctl version`): 1.13.0-rc.1
- Cloud provider (e.g. aws, azure, gke) or local deployment (e.g. minikube vs dockerized k8s): AWS
- Others:

jdoliner:
Fixes #5916 

This fixes an issue where `update pipeline` could set any parallelism it wanted. The new rule is that `update pipeline` can't increase the parallelism over the existing parallelism but it can update a pipeline and keep the same parallelism.

### DIFF
--- a/src/server/pps/server/api_server.go
+++ b/src/server/pps/server/api_server.go
@@ -1865,7 +1865,13 @@ func (a *apiServer) validatePipelineRequest(request *pps.CreatePipelineRequest) 
 
 func (a *apiServer) validateEnterpriseChecks(ctx context.Context, pipelineInfo *pps.PipelineInfo) error {
 	pachClient := a.env.GetPachClient(ctx)
-	if _, err := pachClient.InspectPipeline(pipelineInfo.Pipeline.Name); err == nil {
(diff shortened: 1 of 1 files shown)


nitinjainsj (approved):
LGTM

PR #5925, opened 2021-04-01
Move identity config to postgres

actgardner:
Because we don't support storing durable data in etcd in postgres in 2.0, migrate the identity server config to postgres. Originally I was just going to use a postgres collection + watch, but there was a bug where Dex requires a server restart every time an IDP connector is changed. If you're running multiple HA identity servers we don't have a good way to notify all the servers that they need to restart whenever a single server sees an IDP connector change.

Instead I did a bit of re-architecting:
- We don't need to lazily create the postgres connector, postgres is required for pachd to start.
- Store the config in a single row in postgres using sqlx directly.
- The web server calls into the API server to check whether the config or connectors have changed. If it changes, the web server is restarted.

This adds some latency to every request to the identity server, but this is a generally low-traffic server that users only interact with during the login process, so I don't expect it to bog down postgres.


src/server/identity/server/dex_web.go
```diff
@@ -23,41 +24,35 @@ var webDir = "/dex-assets"
 type dexWeb struct {
 	sync.RWMutex
 
-	// `issuer` must be a well-known URL where all pachds can reach this server.
-	// Dex usually loads it from a config file, but it can be clumsy to find the
-	// exact right value. Instead of requiring the user to change the config
-	// we support updating it via an RPC.
```
msteffen, line 29: Any chance you'd be OK preserving this comment in `src/server/identity/db.go`? (to document what's stored in the config table, basically)

src/server/identity/db.go
```diff
@@ -16,3 +16,14 @@ CREATE TABLE IF NOT EXISTS identity.users (
 `)
 	return err
 }
+
+// CreateConfigTable sets up the postgres table which stores IDP configuration
+func CreateConfigTable(ctx context.Context, tx *sqlx.Tx) error {
+	_, err := tx.ExecContext(ctx, `
+CREATE TABLE IF NOT EXISTS identity.config (
```
msteffen, line 23: I think this lets you remove `IdentityServerConfig` from `src/identity/identity.proto` too?

msteffen (approved):
Overall LGTM, with two small suggestions/requests

src/server/identity/db.go
actgardner: We still use `IdentityServerConfig` in the `(Set|Get)IdentityServerConfig` RPCs.

PR #5935, opened 2021-04-05
Add status-only flag to list datum.

PFedak:
List datum is a particularly bad performance case for get file, fetching
two small files separately for every datum. The most common usage for
list datum is to see failed datums, which only requires checking the
existence of some files per datum and so is much faster.


src/client/pps.go
```diff
@@ -457,23 +457,29 @@ func (c APIClient) listDatum(job *pps.Job, input *pps.Input, pageSize, page int6
 
 // ListDatumF returns info about datums in a Job, calling f with each datum info.
 func (c APIClient) ListDatumF(jobID string, pageSize int64, page int64, f func(di *pps.DatumInfo) error) error {
-	return c.listDatumF(NewJob(jobID), nil, pageSize, page, f)
+	return c.listDatumF(NewJob(jobID), nil, pageSize, page, false, f)
+}
+
+// ListDatumStatusF returns only status info about datums in a Job, calling f with each datum info.
+func (c APIClient) ListDatumStatusF(jobID string, pageSize int64, page int64, f func(di *pps.DatumInfo) error) error {
```
msteffen, line 464: How hard do you think it would be to implement a functional-options version of ListDatumF instead? (i.e. `ListDatumF(...normal arguments..., WithStatusOnly())`)

I'm kind of hoping that that becomes a normal way of adding these kinds of performance options to functions (like `--split-transaction` for `DeleteRepo()`, or a hypothetical future `--dont-fetch-pipeline-spec` version of `ListPipeline()`). Otherwise I worry we'll drown in slight renames of our API calls.

msteffen (approved):
On suggestion about the client API, but otherwise LGTM

src/client/pps.go
PFedak: I think your concerns about incremental changes like this make sense, but in this particular case the flag and method call won't be carried over to 2.0, since it doesn't have the same performance issues. Maybe I should be worried about our deprecation process for this, come to think of it.

In any case, I'm happy to make that change.

PR #5936, opened 2021-04-05
Remove GetAuthToken and ExtendAuthToken RPC

actgardner:
`GetAuthToken` and `ExtendAuthToken` don't have any security vulnerabilities, but being able to impersonate users seems like an anti-pattern. The product has robot tokens so long-lived processes outside the cluster can authenticate with unique identities, and we've added the `GetRobotAuthToken` RPC to get a robot token (and `GetPipelineAuthTokenInTransaction` for PPS to get pipeline tokens). Having more specific RPCs will hopefully make the authorization logic very simple.


src/auth/auth.proto
```diff
@@ -79,13 +79,6 @@ message TokenInfo {
   // Subject (i.e. Pachyderm account) that a given token authorizes.
   // See the note at the top of the doc for an explanation of subject structure.
   string subject = 1;
-
-  enum TokenSource {
-    INVALID = 0;
-    AUTHENTICATE = 1; // returned by Authenticate()--non-revokeable
-    GET_TOKEN = 2;  // returned by GetToken()--revokeable.
-  }
-  TokenSource source = 2;
```
actgardner, line 88: There's only one way to get each token type, and all tokens are revocable, so I eliminated this concept.

src/internal/auth/interceptor.go
```diff
@@ -31,8 +31,6 @@ var authHandlers = map[string]authHandler{
 	// Activate only has an effect when auth is not enabled
 	// Authenticate, Authorize and WhoAmI check auth status themselves
 	// GetOIDCLogin is necessary to authenticate
-	// GetAuthToken and Deactivate are necessary in the partial activation state
```
actgardner, line 34: We no longer have the partial activation state, this was stale.

msteffen (approved):
LGTM

PR #5938, opened 2021-04-05
Disable auth service watches in sidecar

actgardner:
We're severely limited in how many postgres watches we can have, unlike in etcd, so sidecars cannot use watches. The auth service uses watches to monitor the OIDC config and the cluster role bindings.

This PR disables the watches in the auth service. This makes it impossible to get/set the OIDC config from the sidecar. It also makes `Authorize` do a postgres query for the cluster role bindings on each request. 

We discussed proxying all the sidecar auth requests to a full pachd, but that seemed like a bigger lift that might not scale well as the number of pipelines grows. This approach has the advantage that we could slap a small LRU cache on the Authorize results in the sidecar and every request should be for the pipeline user and either one of the input repos or output repo.

### DIFF
--- a/src/internal/testutil/local/pachd.go
+++ b/src/internal/testutil/local/pachd.go
@@ -215,7 +215,7 @@ func RunLocal() (retErr error) {
 		var authAPIServer authserver.APIServer
 		if err := logGRPCServerSetup("Auth API", func() error {
 			authAPIServer, err = authserver.NewAuthServer(
-				env, txnEnv, path.Join(env.Config().EtcdPrefix, env.Config().AuthEtcdPrefix), true, requireNoncriticalServers)
+				env, txnEnv, path.Join(env.Config().EtcdPrefix, env.Config().AuthEtcdPrefix), true, requireNoncriticalServers, true)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


msteffen:
> This approach has the advantage that we could slap a small LRU cache on the Authorize results in the sidecar and every request should be for the pipeline user and either one of the input repos or output repo.

Just trying to get my bearings a little, is the LRU cache in this PR or will that come later?

Overall this plan makes sense to me—how should we handle cache invalidation in the sidecars? Like evict every 10 seconds or something?

msteffen (approved):
Overall LGTM—if I understand this PR correctly, this doesn't add any kind of caching, it just puts the clusterRoleBinding cache behind a `watchesEnabled` flag. That all makes sense to me!

actgardner:
> Just trying to get my bearings a little, is the LRU cache in this PR or will that come later?

Sorry yeah, the LRU cache is a possible improvement we could add in the future. This is just to make sure we can run ~100 workers in 2.0 GA. I do think we could just throw a 30s TTL on the cache entries and everything would be good, but I'll hold off for now.

PR #5940, opened 2021-04-05
Remove enterprise server from sidecar

actgardner:
Watches are expensive in postgres, so we can't use them in the sidecar. We also don't want to heartbeat. It turns out we don't need the enterprise service in the sidecar at all, since we only call auth.Authorize and it doesn't check the enterprise license state.


msteffen (approved):
LGTM

PR #5945, opened 2021-04-06
Fix flaky enterprise test TestGetActivationCode

actgardner:
Fix an issue where `TestGetActivationCode` would sometimes fail because the cached value hadn't caught up to etcd writes:
- the test would call `Activate` to enable auth on the cluster with a license and no expiration
- the test would then call `Activate` _again_ with an expiration, to check whether the cluster state would change
- the second call to `Activate` would immediately succeed because there was a license, but the expiration may not have taken effect yet

This change blocks `Activate` until the exact value written to etcd appears in the cache, instead of just blocking until any not-nil value appears.

### DIFF
--- a/src/internal/auth/interceptor.go
+++ b/src/internal/auth/interceptor.go
@@ -253,7 +253,7 @@ func (i *Interceptor) InterceptUnary(ctx context.Context, req interface{}, info 
 	username, err := a(pachClient, info.FullMethod)
 
 	if err != nil {
-		logrus.Errorf("denied unary call %q to user %v\n", info.FullMethod, nameOrUnauthenticated(username))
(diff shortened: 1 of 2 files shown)


src/internal/auth/interceptor.go
```diff
@@ -253,7 +253,7 @@ func (i *Interceptor) InterceptUnary(ctx context.Context, req interface{}, info
 	username, err := a(pachClient, info.FullMethod)
 
 	if err != nil {
-		logrus.Errorf("denied unary call %q to user %v\n", info.FullMethod, nameOrUnauthenticated(username))
+		logrus.WithError(err).Errorf("denied unary call %q to user %v\n", info.FullMethod, nameOrUnauthenticated(username))
```
acohen4, line 256: thank you :)

acohen4:
LGTM :)

PR #5948, opened 2021-04-07
Change default OIDC behaviour to allow unverified email

actgardner:
In 1.12 we changed the default behaviour of the OIDC connector to require the `email_verified` claim by default. This is a good idea if the provider supports it, but it turns out customers keep hitting this stumbling block because their IDPs don't provide this claim. 

This changes the field from `ignore_email_verified` to `require_email_verified`. The default behaviour is now to accept ID tokens without the `email_verified` claim, but we still allow users to enforce this check.


msteffen (approved):
LGTM

PR #5949, opened 2021-04-07
Move Auth Token Storage from etcd to Postgres

acohen4:


src/server/auth/db.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/jmoiron/sqlx"
+	"golang.org/x/net/context"
+)
+
+// TODO(acohen4): Decide on hash as primary key or unique column
+// CreateTokensTable sets up the postgres table which tracks active clusters
+func CreateAuthTokensTable(ctx context.Context, tx *sqlx.Tx) error {
+	_, err := tx.ExecContext(ctx, `
+CREATE TABLE IF NOT EXISTS auth.auth_tokens (
+	token_hash VARCHAR(4096) PRIMARY KEY,
```
acohen4, line 13: @actgardner any thoughts on having token_hash be a `primary key` vs a `unique field`?

src/server/auth/server/api_server.go
```diff
@@ -1416,22 +1368,147 @@ func (a *apiServer) RestoreAuthToken(ctx context.Context, req *auth.RestoreAuthT
...
+		}
+	}(); err != nil {
 		return nil, errors.Wrapf(err, "error restoring auth token")
 	}
+
 	return &auth.RestoreAuthTokenResponse{}, nil
 }
+
+// returns -1 if token is not found, and the TTL otherwise
+func (a *apiServer) lookupAuthTokenTTLSeconds(ctx context.Context, tokenHash string) (int64, error) {
```
acohen4, line 1392: @actgardner should I move these helper methods dealing with PG to a different file?

src/server/auth/server/api_server.go
```diff
@@ -1416,22 +1368,147 @@ func (a *apiServer) RestoreAuthToken(ctx context.Context, req *auth.RestoreAuthT
...
+				return nil, err
+			}
+			token.Expiration = expiration
+		}
+		robotTokens = append(robotTokens, token)
+	}
+	return robotTokens, nil
+}
+
+func (a *apiServer) generateAndInsertAuthToken(ctx context.Context, subject string, ttlSeconds int64) (string, error) {
```
acohen4, line 1451: I should add a comment here to clarify that we interpret an expiration value of NULL as "lives forever". Now we will sometimes have expiration values set in the passed, since we only remove those values in the purgeExpiredTokens() goroutine

src/server/auth/server/api_server.go
```diff
@@ -613,16 +596,20 @@ func (a *apiServer) WhoAmI(ctx context.Context, req *auth.WhoAmIRequest) (resp *
...
 		token, err := auth.GetAuthToken(ctx)
 		if err != nil {
 			return nil, err
 		}
-		ttl, err = a.tokens.ReadOnly(ctx).TTL(auth.HashToken(token)) // lookup token TTL
+		ttl, err = a.lookupAuthTokenTTLSeconds(ctx, auth.HashToken(token))
 		if err != nil {
 			return nil, errors.Wrapf(err, "error looking up TTL for token")
 		}
+		if ttl < -1 {
```
acohen4, line 610: the semantics here are kind of strange. Let me know if you have a better idea to make this more clear.

a TTL of -1 is returned to the user to communicate "lives forever". This is to be consistent with the current behavior, and seems to make good sense. It's strange because if the TTL value is less than 0, it could also just be expired, so right now we're interpreting anything less than -1 as expired.

One way of writing this more clearly is to just return the ErrBadToken directly from `lookupAuthTokenTTLSeconds() `method

src/server/auth/db.go
actgardner: Primary key makes sense to me - the postgres docs say it's just uniqueness + not null, both of which apply to the token hash.

src/server/auth/db.go
```diff
@@ -0,0 +1,20 @@
...
+	subject VARCHAR(64) NOT NULL,
```
actgardner, line 14: We're probably going to want an index on the subject and expiration as well? I think subject could just be a normal b-tree index, we might benefit from a different type for expiration to do efficient deletions when we clean them up?

src/server/auth/server/api_server.go
```diff
@@ -214,6 +204,15 @@ func NewAuthServer(
 		// Watch for changes to the cluster role binding
 		go s.clusterRoleBindingCache.Watch()
 	}
+
+	// purge expired tokens every 24 hours
+	go func(a *apiServer, intervalSeconds int) {
```
actgardner, line 209: This is mostly just personal preference, but it feels more idiomatic to have this be a method attached to the apiServer struct, like `func (a *apiServer) purgeExpiredTokensRoutine()`. `intervalSeconds` could also be pulled out as a constant `time.Duration` at the top of this file.

src/server/auth/server/api_server.go
actgardner: I think having them in this file is fine.

src/server/auth/server/api_server.go
```diff
@@ -1416,22 +1368,147 @@ func (a *apiServer) RestoreAuthToken(ctx context.Context, req *auth.RestoreAuthT
...
+	}(); err != nil {
 		return nil, errors.Wrapf(err, "error restoring auth token")
 	}
+
 	return &auth.RestoreAuthTokenResponse{}, nil
 }
+
+// returns -1 if token is not found, and the TTL otherwise
+func (a *apiServer) lookupAuthTokenTTLSeconds(ctx context.Context, tokenHash string) (int64, error) {
+	var ttl float64
```
actgardner, line 1393: What do you think about changing the response struct to return the absolute expiration times instead of a TTL? The reason we expose a TTL in the first place is because that's how etcd did it, but if postgres knows the exact timestamp we could just send that to the user and it seems more useful.

src/server/auth/server/api_server.go
```diff
@@ -1416,22 +1368,147 @@ func (a *apiServer) RestoreAuthToken(ctx context.Context, req *auth.RestoreAuthT
...
+	return token, nil
+}
+
+// generates a token, and stores it's hash and supporting data in postgres
+func (a *apiServer) upsertAuthToken(ctx context.Context, tokenHash string, subject string, ttlSeconds int64) error {
+	// Register the pachd in the database
+	if _, err := a.env.GetDBClient().ExecContext(ctx,
+		`INSERT INTO auth.auth_tokens (token_hash, subject, expiration) 
+		VALUES ($1, $2, NOW() + $3 * interval '1 sec') 
+		ON CONFLICT (token_hash)
```
actgardner, line 1474: What's the use case for updating the subject and expiration if the hash collides? I do think we used to do this, but if a user is going to get a token that exactly matches another user's token we should probably throw an error?

src/server/auth/server/api_server.go
```diff
@@ -1416,22 +1368,147 @@ func (a *apiServer) RestoreAuthToken(ctx context.Context, req *auth.RestoreAuthT
...
+func (a *apiServer) deleteAuthToken(ctx context.Context, tokenHash string) error {
+	if _, err := a.env.GetDBClient().ExecContext(ctx, `DELETE FROM auth.auth_tokens WHERE token_hash=$1`, tokenHash); err != nil {
+		return errors.Wrapf(err, "error deleting token")
+	}
+	return nil
+}
+
+func (a *apiServer) purgeExpiredTokens() error {
+	logrus.Info("Purging expired auth tokens")
+	if _, err := a.env.GetDBClient().Exec(`DELETE FROM auth.auth_tokens WHERE NOW() > expiration`); err != nil {
```
actgardner, line 1510: I'm assuming this is safe for the case where `expiration=NULL`?

src/server/auth/server/api_server.go
acohen4: good idea

src/server/auth/db.go
actgardner: Oh no, I meant a different type of index, but it seems like a b-tree might actually be fine. Don't let me prematurely optimize this too hard ;)

src/server/auth/server/api_server.go
acohen4: I think it was needed in the GetAuthToken RPC that's now removed. I will simplify it here to a plain insert

src/server/auth/db.go
acohen4: On second thought, do we have an anticipated use-case in mind to query on subject? My only concern is adding unused indexes that may slow down writes. 

Although this is such a simple table it may not matter too much. What do you think @actgardner ?

src/auth/auth.proto
```diff
@@ -105,9 +105,9 @@ message AuthenticateResponse {
 message WhoAmIRequest {}
 
 message WhoAmIResponse {
-  reserved 2,4;
+  reserved 2,3,4;
   string username = 1;
-  int64 ttl = 3 [(gogoproto.customname) = "TTL"];
+  google.protobuf.Timestamp expiration = 5;
```
acohen4, line 110: @actgardner  do we also check in the generated proto.go file?

actgardner: Yes, add the generated proto go files :)

src/auth/auth.proto
```diff
@@ -140,6 +140,8 @@ enum Permission {
   CLUSTER_AUTH_GET_GROUP_USERS           = 111;
   CLUSTER_AUTH_EXTRACT_TOKENS            = 112;
   CLUSTER_AUTH_RESTORE_TOKEN             = 113;
+  CLUSTER_AUTH_DELETE_EXPIRED_TOKENS     = 140;
```
actgardner, line 143: 👍

src/client/transaction.go
```diff
@@ -588,3 +588,7 @@ func (c *debugBuilderClient) Binary(ctx context.Context, req *debug.BinaryReques
 func (c *debugBuilderClient) Dump(ctx context.Context, req *debug.DumpRequest, opts ...grpc.CallOption) (debug.Debug_DumpClient, error) {
 	return nil, unsupportedError("Dump")
 }
+
+func (c *authBuilderClient) DeleteExpiredAuthTokens(ctx context.Context, req *auth.DeleteExpiredAuthTokensRequest, opts ...grpc.CallOption) (*auth.DeleteExpiredAuthTokensResponse, error) {
+	return nil, unsupportedError("RestoreAuthToken")
```
actgardner, line 593: RestoreAuthToken -> DeleteExpiredAuthTokens

src/internal/clusterstate/current.go
```diff
@@ -51,4 +52,11 @@ var DesiredClusterState migrations.State = migrations.InitialState().
 	}).
 	Apply("create identity config table v0", func(ctx context.Context, env migrations.Env) error {
 		return identity.CreateConfigTable(ctx, env.Tx)
+	}).
+	Apply("create auth schema", func(ctx context.Context, env migrations.Env) error {
+		_, err := env.Tx.ExecContext(ctx, `CREATE SCHEMA auth`)
+		return err
+	}).
+	Apply("auth tokens", func(ctx context.Context, env migrations.Env) error {
```
actgardner, line 60: I think the convention we're following for the migration names is like `create auth tokens table v0`

src/server/auth/db.go
actgardner: I have a follow-on task to add an RPC for "revoke all tokens for a given user", I think indexing on subject is useful in that case.

src/server/auth/server/roles.go
```diff
@@ -24,6 +24,7 @@ func permissionsForRole(role string) ([]auth.Permission, error) {
 			auth.Permission_CLUSTER_AUTH_GET_GROUP_USERS,
 			auth.Permission_CLUSTER_AUTH_EXTRACT_TOKENS,
 			auth.Permission_CLUSTER_AUTH_RESTORE_TOKEN,
+			auth.Permission_CLUSTER_AUTH_DELETE_EXPIRED_TOKENS,
```
actgardner, line 27: 👍

src/server/auth/server/api_server.go
```diff
@@ -1416,22 +1364,164 @@ func (a *apiServer) RestoreAuthToken(ctx context.Context, req *auth.RestoreAuthT
...
 
-	// Check whether the token hash already exists - we don't want to replace an existing token
-	if _, err := col.NewSTM(ctx, a.env.GetEtcdClient(), func(stm col.STM) error {
-		tokens := a.tokens.ReadWrite(stm)
-		var existing auth.TokenInfo
-		err := tokens.Get(req.Token.HashedToken, &existing)
-		if err == nil {
+	if err := func() error {
+		// Check whether the token hash already exists - we don't want to replace an existing token
+		if _, err := a.lookupAuthTokenInfo(ctx, req.Token.HashedToken); err == nil {
```
actgardner, line 1369: You can probably eliminate this because of the uniqueness constraint now, right? There's no way to restore an existing token.

actgardner:
This is looking good! I made a few suggestions, mostly just typos.

src/server/auth/db.go
acohen4: sounds good :)

src/server/auth/server/api_server.go
```diff
@@ -1416,22 +1366,163 @@ func (a *apiServer) RestoreAuthToken(ctx context.Context, req *auth.RestoreAuthT
...
+			return nil, tsErr
+		}
+		return expiryTS, nil
+	}
+	return nil, nil
+}
+
+func (a *apiServer) lookupAuthTokenInfo(ctx context.Context, tokenHash string) (*auth.TokenInfo, error) {
+	var tokenInfo auth.TokenInfo
+	if err := a.env.GetDBClient().GetContext(ctx, &tokenInfo, `SELECT subject FROM auth.auth_tokens WHERE token_hash = $1`, tokenHash); err != nil {
```
actgardner, line 1423: I think this is missing the case where a token is expired but it hasn't been deleted yet? It might be good to add that to the test, call `WhoAmI` or something with the recently-expired-but-undeleted token and make sure it isn't accepted.

actgardner:
@acohen4 looks good, one more comment about making sure users can't use an expired token that hasn't been deleted yet.

src/server/auth/server/api_server.go
```diff
@@ -1416,22 +1354,157 @@ func (a *apiServer) RestoreAuthToken(ctx context.Context, req *auth.RestoreAuthT
...
+		return nil, col.ErrNotFound{Type: "auth_tokens", Key: tokenHash}
+	}
+
+	if err = rows.Scan(&subject, &expiration); err != nil {
+		return nil, errors.Wrapf(err, "error querying token")
+	}
+
+	tokenInfo := auth.TokenInfo{Subject: subject}
+	if expiration != nil {
+		expiryTS, tsErr := types.TimestampProto(*expiration)
```
actgardner, line 1404: There's a gogoproto extension that will generate this struct with a normal Go time, which should let you load the data directly into the struct using sqlx:

https://github.com/pachyderm/pachyderm/blob/master/src/license/license.proto#L61-L62

https://github.com/pachyderm/pachyderm/blob/master/src/server/license/server/api_server.go#L308-L309

actgardner (approved):
LGTM!

PR #5951, opened 2021-04-07
Support protocol in license/enterprise server config

actgardner:
When license server and enterprise server communicate, previously we only accepted an address and not a protocol. This meant users couldn't connect to a server using TLS. Update the codebase to accept a protocol and use the system CAs for a TLS connection if specified.


acohen4 (approved):
LGTM

PR #5952, opened 2021-04-07
Treat an empty enterprise context as a single cluster deployment

actgardner:
Initially the plan for the enterprise context was to try and keep it up-to-date with the active context when we suspected a user was working with individual clusters (ex. not in an enterprise setup). I've gotten some feedback that this is confusing. 

This PR changes `config.ActiveEnterpriseContext` to return the active (non-enterprise) context when the enterprise context is blank. It also makes it so `pachctl deploy` and `pachctl config set-active-context` _don't_ configure the enterprise context. The end result is that if a user never deploys an enterprise server then the active enterprise context will never be set, and all RPCs will go to the active context, which seems less confusing.

### DIFF
--- a/src/internal/config/config.go
+++ b/src/internal/config/config.go
@@ -70,7 +70,8 @@ func (c *Config) ActiveContext(errorOnNoActive bool) (string, *Context, error) {
 	return c.V2.ActiveContext, context, nil
 }
 
-// ActiveEnterpriseContext gets the active enterprise server context in the config
+// ActiveEnterpriseContext gets the active enterprise server context in the config.
+// If no enterprise context is configured, this returns the active context.
 func (c *Config) ActiveEnterpriseContext(errorOnNoActive bool) (string, *Context, error) {
 	if c.V2 == nil {
 		return "", nil, errors.Errorf("cannot get active context from non-v2 config")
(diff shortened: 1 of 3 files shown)


msteffen (approved):
LGTM, though I actually don't understand where the old semantics and the new semantics differ in practice. My understanding was:
**Old**
- `pachctl deploy` sets `active-context` and `active-enterprise-context` to the same thing
- `pachctl config get active-enterprise-context` returns the `active-enterprise-context` which is equal to the `active-context`
- `pachctl config set active-context` sets both `active-context` and `active-enterprise-context` to the same, new thing every time it's run (because they're the same to begin with).
- `pachctl config get active-enterprise-context` returns the `active-enterprise-context` which is still equal to the `active-context` (because that invariant is being maintained by `set`)

**New**
- `pachctl deploy` sets `active-context` and leaves `active-enterprise-context` blank
- `pachctl config get active-enterprise-context` returns the `active-context` which is equal to itself (because the `active-enterprise-context` is blank to begin with)
- `pachctl config set active-context` sets `active-context` and still leaves `active-enterprise-context` blank
- `pachctl config get active-enterprise-context` still returns the `active-context` which is still equal to itself (because `set` is maintaining the invariant that `active-enterprise-context` is blank)

It is just a matter of people looking at their config and being like "what is this new context?!"?

Final question, I should know this but just re-checking my understanding: the enterprise context sets the endpoint/auth token/etc for the license server, right? (and the `enterprise` server not related--users don't talk to that directly anymore?)

actgardner:
> It is just a matter of people looking at their config and being like "what is this new context?!"?

You're absolutely right that the semantics are mostly the same. People did seem to be confused by the new context. Also, for users who had an existing active context but not an enterprise context it broke the invariant and they had to manually fix it. 

Ironically, this change is going to re-break things for those users (who now have an enterprise context set), but that's a very small pool of internal users ATM.

PR #5954, opened 2021-04-08
Move S3 gateway from clientFactory to serviceenv

msteffen:
This gets us away from clientFactory, which was leaking TCP sessions, and moves us to serviceenv which is the intended library for initializing clients (and is designed to solve this problem)


PFedak (approved):
LGTM

msteffen:
Closing; this is obsolete

PR #5955, opened 2021-04-08
PG Bouncer Deployment and Service

brendoncarroll:
Adds a PG Bouncer k8s service and deployment.
Modifies `etc/reset.py` to push the image to minikube.
Sets the environment variable for the postgres host on the pachd deployment.


src/internal/deploy/assets/postgres.go
```diff
@@ -380,3 +385,75 @@ EOSQL
...
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: objectMeta(pgBouncerName, labels(pgBouncerName), nil, opts.Namespace),
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name:  pachdName,
+							Image: pgBouncerImage,
+							Env: []v1.EnvVar{
+								{Name: "DB_USER", Value: postgresUser},
+								{Name: "DB_PASSWORD", Value: postgresPassword},
```
actgardner, line 435: Do we need to specify a password? Aren't we using trust auth for everything anyways?

actgardner (approved):

src/internal/deploy/assets/postgres.go
```diff
@@ -374,9 +385,81 @@ set -e
...
+			Replicas: replicas(1),
+			Selector: &metav1.LabelSelector{
+				MatchLabels: labels(pgBouncerName),
+			},
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: objectMeta(pgBouncerName, labels(pgBouncerName), nil, opts.Namespace),
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name:  pachdName,
```
brycemcanally, line 437: pachdName -> pgBouncerName

src/internal/deploy/assets/postgres.go
```diff
@@ -374,9 +385,81 @@ set -e
...
+								{Name: "AUTH_TYPE", Value: "trust"},
+							},
+							Ports: []v1.ContainerPort{
+								{
+									ContainerPort: 5432,
+									Name:          "client-port",
+								},
+							},
+							ImagePullPolicy: "IfNotPresent",
+							Resources: v1.ResourceRequirements{
```
brycemcanally, line 452: We are just going to use the same resource requests as postgres?

etc/testing/start_postgres.sh
```diff
@@ -1,12 +1,12 @@
...
+if ! docker ps | grep -q postgres_test
 then
     echo "starting postgres..."
-    docker run -d \
-    -e POSTGRES_DB=pgc \
+    docker run -d --rm \
+    --name postgres_test \
     -e POSTGRES_HOST_AUTH_METHOD=trust \
-    -p 32228:5432 \
+    -p 5432:5432 \
```
brycemcanally, line 9: This is kind of annoying if you have an unrelated postgres instance using the default port. It might be better to just have our own postgres_test port.

src/internal/deploy/assets/assets.go
```diff
@@ -1224,6 +1236,14 @@ func WriteAssets(encoder serde.Encoder, opts *AssetOpts, objectStoreBackend Back
 		return err
 	}
 
+	// PG Bouncer
```
brycemcanally, line 1239: I would have no idea what the two lines below are doing without this comment.

src/internal/deploy/assets/postgres.go
```diff
@@ -374,9 +385,81 @@ set -e
 
 psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-EOSQL
     CREATE DATABASE dex;
-    GRANT ALL PRIVILEGES ON DATABASE dex TO postgres;
+    GRANT ALL PRIVILEGES ON DATABASE dex TO pachyderm;
```
brycemcanally, line 388: This should probably be using the PostgresUser variable.

brycemcanally:
We need to propagate this configuration to the worker environment as well. You need to make some changes to `src/server/pps/server/worker_rc.go`. I just ran a test locally to confirm that the postgres environment variables aren't getting set up for the workers.

src/internal/deploy/assets/postgres.go
brendoncarroll: I think you need the password to tell it to use password auth as opposed to one of the other auth methods.

etc/testing/start_postgres.sh
brendoncarroll: The check just looks for a container called postgres.  I think sharing the default port in this case will make problems with multiple postgres's more obvious.

PR #5956, opened 2021-04-08
Move S3 Gateway clientFactory -> ServiceEnv

msteffen:
1.13.x version of #5955

This gets us away from clientFactory, which was leaking ports, and moves us to serviceenv which is the intended library for initializing clients (and is designed to solve this problem)

### DIFF
--- a/etc/testing/travis.sh
+++ b/etc/testing/travis.sh
@@ -11,6 +11,9 @@ set -ex
     done
 ) &
 
+# Log in to docker, so we don't get rate-limited if we're pulling images
+docker login -u pachydermbuildbot -p "${DOCKER_PWD}"
+
 # Repeatedly restart minikube until it comes up. This corrects for an issue in
 # Travis, where minikube will get stuck on startup and never recover
 while true; do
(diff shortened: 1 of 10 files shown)


etc/testing/travis.sh
```diff
@@ -53,8 +56,9 @@ else
     popd
 fi
 
-make launch-loki
-
+if [[ "${BUCKET}" = "PPS3" ]]; then
+    make launch-loki
+fi
```
PFedak, line 61: I guess this isn't that important since we don't anticipate many test changes in 1.x, but the bucket a test is in isn't guaranteed to be stable, right?

PFedak (approved):
LGTM

etc/testing/travis.sh
msteffen: No, it's not. TBH my motivation was trying to make CI run faster, and I figured that despite the risk that TestLoki moves to another bucket due to adding/removing tests, this change was relatively low-risk because a) we'll know right away because CI starts failing, and b) it's easy to fix: you just change the bucket.

Issue #5964, opened 2021-04-09
DeletePipeline fails with auth enabled


actgardner:
**What happened?**:

Running `pachctl delete pipeline` on a cluster with auth enabled deletes the output repo but the pipeline remains. Trying to delete the pipeline a second time raises an error because the repo no longer exists.

Logs:
```
2021-04-09T16:18:43Z INFO pps.API.DeletePipeline {"request":{"pipeline":{"name":"inner_join"}}} []
2021-04-09T16:18:43Z INFO auth.API.WhoAmI {"duration":0.0040083,"request":{},"response":{"username":"user:nadege.pepin@pachyderm.io","ttl":2590592}} []
2021-04-09T16:18:43Z INFO pfs.API.GetFile {"request":{"file":{"commit":{"repo":{"name":"__spec__"},"id":"cf19cadfb0fc40aebfe6a5ad6f06a727"},"path":"spec"}}} []
2021-04-09T16:18:43Z INFO auth.API.WhoAmI {"duration":0.0045624,"request":{},"response":{"username":"user:nadege.pepin@pachyderm.io","ttl":2590592}} []
2021-04-09T16:18:43Z INFO auth.API.Authorize {"request":{"resource":{"type":2,"name":"__spec__"},"permissions":[200]}} []
2021-04-09T16:18:43Z INFO auth.API.Authorize {"duration":0.0698248,"request":{"resource":{"type":2,"name":"__spec__"},"permissions":[200]},"response":{"authorized":true,"satisfied":[200],"principal":"user:nadege.pepin@pachyderm.io"}} []
2021-04-09T16:18:43Z INFO auth.API.WhoAmI {"duration":0.0057644,"request":{},"response":{"username":"user:nadege.pepin@pachyderm.io","ttl":2590592}} []
2021-04-09T16:18:43Z INFO auth.API.Authorize {"request":{"resource":{"type":2,"name":"__spec__"},"permissions":[204]}} []
2021-04-09T16:18:43Z INFO auth.API.Authorize {"duration":0.0059386,"request":{"resource":{"type":2,"name":"__spec__"},"permissions":[204]},"response":{"authorized":true,"satisfied":[204],"principal":"user:nadege.pepin@pachyderm.io"}} []
2021-04-09T16:18:43Z INFO auth.API.WhoAmI {"duration":0.0035168,"request":{},"response":{"username":"user:nadege.pepin@pachyderm.io","ttl":2590592}} []
[...]

PR #5965, opened 2021-04-09
Fix DeletePipeline with auth enabled

actgardner:
Resolves https://github.com/pachyderm/pachyderm/issues/5964

When auth is enabled, calling InspectRepo or DeleteJob on a deleted repo will throw `auth.ErrNoRoleBinding` errors rather than a not-found error. This fixes the behaviour so DeletePipeline works.

### DIFF
--- a/src/server/pps/server/api_server.go
+++ b/src/server/pps/server/api_server.go
@@ -2713,7 +2713,7 @@ func (a *apiServer) deletePipeline(pachClient *client.APIClient, request *pps.De
 
 	// check if the output repo exists--if not, the pipeline is non-functional and
 	// the rest of the delete operation continues without any auth checks
-	if _, err := pachClient.InspectRepo(request.Pipeline.Name); err != nil && !isNotFoundErr(err) {
+	if _, err := pachClient.InspectRepo(request.Pipeline.Name); err != nil && !isNotFoundErr(err) && !auth.IsErrNoRoleBinding(err) {
 		return nil, err
 	} else if !isNotFoundErr(err) {
 		// Check if the caller is authorized to delete this pipeline. This must be
@@ -2770,7 +2770,7 @@ func (a *apiServer) deletePipeline(pachClient *client.APIClient, request *pps.De
 	if err := a.jobs.ReadOnly(ctx).GetByIndex(ppsdb.JobsPipelineIndex, request.Pipeline, jobPtr, col.DefaultOptions, func(jobID string) error {
 		eg.Go(func() error {
 			_, err := a.DeleteJob(ctx, &pps.DeleteJobRequest{Job: client.NewJob(jobID)})
-			if isNotFoundErr(err) {
+			if isNotFoundErr(err) || auth.IsErrNoRoleBinding(err) {
 				return nil
 			}
 			return err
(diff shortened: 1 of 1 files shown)


src/server/pps/server/api_server.go
```diff
@@ -2770,7 +2770,7 @@ func (a *apiServer) deletePipeline(pachClient *client.APIClient, request *pps.De
 	if err := a.jobs.ReadOnly(ctx).GetByIndex(ppsdb.JobsPipelineIndex, request.Pipeline, jobPtr, col.DefaultOptions, func(jobID string) error {
 		eg.Go(func() error {
 			_, err := a.DeleteJob(ctx, &pps.DeleteJobRequest{Job: client.NewJob(jobID)})
-			if isNotFoundErr(err) {
+			if isNotFoundErr(err) || auth.IsErrNoRoleBinding(err) {
```
actgardner, line 2773: This will never succeed with auth enabled since the repo was already deleted. AFAICT that's not a regression in the new auth system, we've been silently ignoring it forever. It seems like long-term we should move the job stopping logic to pollPipelines as the comment suggests?

src/server/pps/server/api_server.go
```diff
@@ -2713,7 +2713,7 @@ func (a *apiServer) deletePipeline(pachClient *client.APIClient, request *pps.De
 
 	// check if the output repo exists--if not, the pipeline is non-functional and
 	// the rest of the delete operation continues without any auth checks
-	if _, err := pachClient.InspectRepo(request.Pipeline.Name); err != nil && !isNotFoundErr(err) {
+	if _, err := pachClient.InspectRepo(request.Pipeline.Name); err != nil && !isNotFoundErr(err) && !auth.IsErrNoRoleBinding(err) {
 		return nil, err
 	} else if !isNotFoundErr(err) {
```
msteffen, line 2718: Should this check be weakened too? I.e. `else if !isNotFoundErr(err) && !auth.IsErrNoRoleBinding(err)` or even `else if err == nil`?

If I understand this correctly, it calls `DeleteRepo` in the `IsErrNoRoleBinding(err) == true` case, which I think it's not supposed to do?

src/server/pps/server/api_server.go
msteffen: ...whoops :grimacing:

re:moving the logic, TBH I'm not totally sure I understand, but I do remember writing the comment. FWIW, basically, the issue is that this operates on the jobs that exist right now, but if there are any output commits that don't have a job yet (the delay being a product of the fact the it's the worker master, running elsewhere, that propagates output commits -> jobs) then a new job could show up even after all existing jobs are deleted.

The idea was basically that `pollPipelines` could list all jobs and delete any job where the job's pipeline doesn't exist anymore, though I think this logic is compatible with that change—i.e. we could make that change and leave this as well if we wanted.

msteffen (commented):
Just one question about the condition around `InspectRepo`

src/server/pps/server/api_server.go
actgardner: Yup, good catch.

actgardner:
> The idea was basically that pollPipelines could list all jobs and delete any job where the job's pipeline doesn't exist anymore, though I think this logic is compatible with that change—i.e. we could make that change and leave this as well if we wanted.

My dream would be to move this to pollPipelines exclusively. I feel like we've been bitten by belt-and-suspenders solutions where neither of the solutions works perfectly, but they usually cover for each other, which results in really hard to debug edge cases, and just generally a proliferation of extra code that doesn't run very often. I don't think I've got the context to move this in the near term though.

msteffen (approved):
LGTM

PR #5966, opened 2021-04-09
Delete identity data during auth deactivate

### LINKED ISSUE #5929: 2.0.0-alpha.10 - Auth - `pachctl auth deactivate` then reactivate leaves cluster in inconsistent state
**What happened?**:
$ **pachctl auth  deactivate**
Are you sure you want to delete ALL auth information (ACLs, tokens, and admins) in this cluster, and expose ALL data? yN
y
$ **pachctl auth activate**
WARNING: DO NOT LOSE THE AUTH TOKEN BELOW. STORE IT SECURELY FOR THE LIFE OF THE CLUSTER.THIS TOKEN WILL ALWAYS HAVE ADMIN ACCESS TO FIX THE CLUSTER CONFIGURATION.
Pachyderm root token:
b508ba28d01345af8137049bc473506e
**failed to configure OIDC client ID: OIDC client with id "pachd" already exists**

It looks like the `pachctl auth deactivate` did not clean the auth properly.

actgardner:
Resolves #5929

### DIFF
--- a/src/server/auth/cmds/cmds.go
+++ b/src/server/auth/cmds/cmds.go
@@ -224,10 +224,10 @@ Prompt the user to input a root token on stdin, rather than generating a random 
 func DeactivateCmd() *cobra.Command {
 	var enterprise bool
 	deactivate := &cobra.Command{
-		Short: "Delete all ACLs, tokens, and admins, and deactivate Pachyderm auth",
-		Long: "Deactivate Pachyderm's auth system, which will delete ALL auth " +
-			"tokens, ACLs and admins, and expose all data in the cluster to any " +
-			"user with cluster access. Use with caution.",
+		Short: "Delete all ACLs, tokens, admins, IDP integrations and OIDC clients, and deactivate Pachyderm auth",
+		Long: "Deactivate Pachyderm's auth and identity systems, which will delete ALL auth " +
+			"tokens, ACLs and admins, IDP integrations and OIDC clients, and expose all data " +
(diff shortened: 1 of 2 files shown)


src/server/auth/cmds/cmds.go
```diff
@@ -243,6 +243,11 @@ func DeactivateCmd() *cobra.Command {
 				return errors.Wrapf(err, "could not connect")
 			}
 			defer c.Close()
+
+			// Delete any data from the identity server
+			if _, err := c.IdentityAPIClient.DeleteAll(c.Ctx(), &identity.DeleteAllRequest{}); err != nil {
```
msteffen, line 248: Maybe this is too much of a corner case to worry about, but is there any risk that by doing `identity.DeleteAll()` before `auth.Deactivate()`, we could get the user's cluster stuck? I.e. they call `c.Deactivate()`, but because that call is restricted they need to re-auth, but they can no longer re-auth because the OIDC info is gone?

msteffen (approved):
I had one question, but otherwise LGTM. Approving the PR because I think the issue raised is a corner case and don't think it should block merging if it doesn't need to be addressed.

src/server/auth/cmds/cmds.go
actgardner: I think this corner case is covered by (a) the root token (b) the ability to do postgres surgery to create a new root token if the original is lost.

PR #5967, opened 2021-04-09
Support adding groups to role bindings

actgardner:
In her testing @nadegepepin found that we had broken support for groups in role bindings, because we have no tests for groups 😿  This fixes support for groups and adds a basic test that we can add a group to role bindings, and that group membership is tested.


msteffen (approved):
LGTM!

PR #5968, opened 2021-04-09
Make groups claim default in pachd OIDC config

actgardner:
In 2.x customers will almost always be connecting pachd to dex, which supports the `groups` scope. In most cases users won't know about `pachctl auth set-config` at all, and will only interact with the identity API. It makes sense to include `groups` in the default so users automatically have group syncing if they configure it in dex.

In the unlikely event that a user isn't using dex and wants pachd to talk to a different OIDC provider directly, they can provide a custom list of scopes which may not include `groups`.


msteffen (approved):
LGTM

PR #5969, opened 2021-04-12
Add functions for writing postgres and etcd assets, move StorageClass to use kube API instead of JSON

Tryneus:


src/internal/deploy/assets/etcd.go
```diff
@@ -65,6 +68,62 @@ type EtcdOpts struct {
...
+		if err = encoder.Encode(EtcdVolumeClaim(volumeSize, opts)); err != nil {
+			return err
+		}
+		if err = encoder.Encode(EtcdDeployment(opts, "")); err != nil {
+			return err
+		}
+	} else {
+		return errors.Errorf("unless deploying locally, either --dynamic-etcd-nodes or --static-etcd-volume needs to be provided")
+	}
+	if err := encoder.Encode(EtcdNodePortService(opts)); err != nil {
```
brycemcanally, line 120: nit, you could just return the result of the last encode call

src/internal/deploy/assets/postgres.go
```diff
@@ -51,6 +54,66 @@ type PostgresOpts struct {
...
+		if err = encoder.Encode(PostgresVolumeClaim(volumeSize, opts)); err != nil {
+			return err
+		}
+		if err = encoder.Encode(PostgresDeployment(opts, "")); err != nil {
+			return err
+		}
+	} else {
+		return errors.Errorf("unless deploying locally, either --dynamic-postgres-nodes or --static-postgres-volume needs to be provided")
+	}
+	if err := encoder.Encode(PostgresService(opts)); err != nil {
```
brycemcanally, line 111: similar to above, just return the result of the encode call.

src/internal/deploy/assets/postgres.go
```diff
@@ -320,6 +383,19 @@ func PostgresStatefulSet(opts *AssetOpts, backend Backend, diskSpace int) interf
 	}
 }
 
+// TestPostgresVolume creates a persistent volume for use with StatefulSets in a
```
brycemcanally, line 386: Is there a way to create a statefulset with the volumes pre-defined? My understanding was that we would need a local storage class to do the local testing with a statefulset. Something I noticed when looking at this last was that we really shouldn't be using deployments for etcd or postgres because we run the risk of deadlocks due to the `ReadWriteOnce` access mode for the volume. It would be good if we could move all of this over to statefulsets and it looks like the local storage class may give us a way to move the local deployment to statefulsets as well.

Tryneus: I agree that we should move these all over to statefulsets, and I'd be happy to do that once I have some spare cycles, but it's a bit out of scope for me now.  I'd like it if we only did statefulsets for pachd as well, just to reduce the surface area of the assets code.

brycemcanally: Yeah, definitely out of scope for this PR. I just wasn't sure if this function would get any use since I'm not aware of a way to set up a statefulset with pre-defined volumes. Also, we should probably keep pachd as a deployment since it is not stateful.

brycemcanally (approved):
lgtm

PR #5970, opened 2021-04-12
Remove admin test bucket

actgardner:
There's no tests in `src/server/admin/...` since we removed extract/restore - remove the test bucket so we don't do a whole VM allocation and build for no reason.


msteffen (approved):
LGTM. Nice catch!!

PR #5971, opened 2021-04-12
pachd: set GOMAXPROCS according to CPU limit

jrockway:
People seem to run pachd with a CPU limit set, but don't remember to set GOMAXPROCS to that CPU limit.  The result is occasional high latency; GOMAXPROCS set greater than the CPU limit will cause the cpu quota to be exceeded before the cfs period is complete under periods of high load, and that results in some RPCs with a long sleep in the middle of them.  The problem that we mitigate is the default behavior of GOMAXPROCS; it is always set to the number of threads available on the underlying VM/machine, and pachd will use this much parallelism and exceed the CPU quota before the CFS throttling period ends.

https://k8s.af/ contains a number of horror stories involving CPU limits in general; this patch reduces the severity of that class of incidents.  You can still starve pachd of performance by not giving it enough CPU, of course, but at least response latency will be consistent in this case.

Note that automaxprocs does nothing in the case of CPU requests; setting requests without limits will use all available cores.  Users may want to explicitly set GOMAXPROCS in the case where they don't want to overcommit the entire machine.


actgardner (approved):
You might want to backport this to 1.13.x as well.

jrockway:
Yup, I'll backport.

PR #5974, opened 2021-04-12
Add get-groups CLI command

actgardner:
We've supported syncing group membership from SAML and OIDC since 1.x but there was no way for a user to see their group memberships with the CLI. This provides a simple way to debug group syncing.

### DIFF
--- a/src/server/auth/cmds/cmds.go
+++ b/src/server/auth/cmds/cmds.go
@@ -442,8 +442,8 @@ func GetGroupsCmd() *cobra.Command {
 	var enterprise bool
 	getGroups := &cobra.Command{
 		Use:   "{{alias}} [username]",
-		Short: "Get an auth token for the user with the specified name.",
-		Long:  "Get an auth token for the user with the specified name.",
+		Short: "Get the list of groups a user belongs to",
+		Long:  "Get the list of groups a user belongs to. If no user is specified, the current user's groups are listed.",
 		Run: cmdutil.RunBoundedArgs(0, 1, func(args []string) error {
 			c, err := newClient(enterprise)
 			if err != nil {
--- a/src/server/auth/cmds/cmds.go
+++ b/src/server/auth/cmds/cmds.go
@@ -438,6 +438,35 @@ func GetRobotTokenCmd() *cobra.Command {
 	return cmdutil.CreateAlias(getAuthToken, "auth get-robot-token")
 }
 
+func GetGroupsCmd() *cobra.Command {
+	var enterprise bool
+	getGroups := &cobra.Command{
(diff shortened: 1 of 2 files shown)


src/server/auth/cmds/cmds_test.go
```diff
@@ -275,15 +317,6 @@ func TestMain(m *testing.M) {
 			panic(err.Error())
 		}
 	}
-	time.Sleep(time.Second)
```
actgardner, line 278: Deactivating auth is synchronous now, this was adding latency to every test with no benefit.

src/server/auth/cmds/cmds.go
```diff
@@ -438,6 +438,35 @@ func GetRobotTokenCmd() *cobra.Command {
 	return cmdutil.CreateAlias(getAuthToken, "auth get-robot-token")
 }
 
+func GetGroupsCmd() *cobra.Command {
+	var enterprise bool
+	getGroups := &cobra.Command{
+		Use:   "{{alias}} [username]",
+		Short: "Get an auth token for the user with the specified name.",
```
msteffen, line 445: I think the Short/Long descriptions need to be updated

src/server/auth/cmds/cmds.go
```diff
@@ -438,6 +438,35 @@ func GetRobotTokenCmd() *cobra.Command {
...
+			}
+			resp, err := c.GetGroups(c.Ctx(), req)
+			if err != nil {
+				return grpcutil.ScrubGRPC(err)
+			}
+			fmt.Println(strings.Join(resp.Groups, "\n"))
+			return nil
+		}),
+	}
+	getGroups.PersistentFlags().BoolVar(&enterprise, "enterprise", false, "Get group membership info from the enterprise server")
```
msteffen, line 466: Just checking my mental model: this is for users who want to change the IDP configs, and want to make sure that the enterprise server will authorize them to do so (because a group that they're a member of has a ClusterRole, and the group membership must therefore be stored on the enterprise sierver for the user to assume that ClusterRole and modify the IDP config)?

msteffen (approved):
LGTM, with one nit and one question

src/server/auth/cmds/cmds_test.go
msteffen: Good catch!

src/server/auth/cmds/cmds.go
actgardner: Yeah, basically - the enterprise server has it's own auth server that users have to log in to, and that syncs group membership the same as the pachd. That group membership can be used in role bindings to support updating the enterprise license or configuring IDPs.

PR #5975, opened 2021-04-12
Remove unused ID field from pipelineinfo

actgardner:
AFAICT this field isn't used anywhere, and it was confusing to the front-end team.


jdoliner (approved):
LGTM

PR #5977, opened 2021-04-12
Fix Travis by making 'make launch-loki' work

msteffen:


jrockway (approved):
Good call on "docker login".

I think we're still installing the wrong version and that implies that we might need more test coverage in that area.  Or, 1.5.x and 2.0.0 are more compatible than we thought!

PR #5980, opened 2021-04-13
Fix activating enterprise with localhost

actgardner:
These tests weren't running, and consequently I didn't know that enterprise activation was broken because the license server got removed from the internal server.


msteffen (approved):
LGTM

PR #5981, opened 2021-04-13
Support port-forwarding to the enterprise server implicitly

actgardner:
Because the enterprise server pods have a different name, `pachctl` wasn't able to connect to them by implicitly port-forwarding when no `grpc_address` was set. This should fix that.


src/internal/config/config.proto
```diff
@@ -95,6 +95,10 @@ message Context {
     // we ensure this is the same as what the cluster reports back, to prevent
     // us from connecting to the wrong cluster.
     string cluster_deployment_id = 11 [(gogoproto.customname) = "ClusterDeploymentID"];
+
+    // A boolean that records whether the context points at an enterprise server,
```
msteffen, line 99: minor nit: IIUC, in the "unified" case (pachd and enterprise are on the same server), this should be `false`, since the app for that pod will be `pachd`. I wonder if it would make sense to say e.g. `if true, this context points to a standalone Pachyderm enterprise server, rather than a regular pachd pod` or something along those lines, to maybe make the default clearer?

msteffen (approved):
One minor nit, otherwise LGTM

PR #5982, opened 2021-04-13
Rework test env methods

Tryneus:
This changes the `testetcd.Env`, `testpachd.MockEnv` and `testpachd.RealEnv` to use `t.Cleanup` instead of being callback-based.  In addition:

 * `testpachd.NewRealEnv` will now populate the defaults of its serviceenv config
 * Added `Context()` and `Close()` methods to `ServiceEnv`
 * Added a method for creating a postgres deployment on-demand for tests (currently unused because it doesn't work on testfaster, still debugging)
 * Moved some things around to fix circular dependencies (mostly just having `assets` not depend on things in `server`).
 * Added a `testutil.MkdirTemp()` for a test-scoped temporary dir, which was being reimplemented in several places
 * Moved `pfs/server.NewObjClient` to `obj.NewClient`.


src/internal/cmdutil/env.go
```diff
@@ -71,10 +71,10 @@ func populateInternal(reflectValue reflect.Value, decoderMap map[string]string,
 	if reflectValue.Type().Kind() != reflect.Struct {
 		return errors.Errorf("%s: %v", expectedStructErr, reflectValue.Type())
 	}
-	numField := reflectValue.NumField()
-	for i := 0; i < numField; i++ {
+	for i := 0; i < reflectValue.NumField(); i++ {
 		structField := reflectValue.Type().Field(i)
-		if structField.Type.Kind() == reflect.Struct {
+		ptrToStruct := structField.Type.Kind() == reflect.Ptr && structField.Type.Elem().Kind() == reflect.Struct
+		if structField.Type.Kind() == reflect.Struct || ptrToStruct {
```
Tryneus, line 77: I'm a little confused by the code before - from what I can tell, it would never run on the `serviceenv.Configuration` object because all of its subfields are pointers to structs, but I'm pretty sure this would have broken `pachd` if it didn't work...

src/internal/log/log.go
```diff
@@ -156,14 +156,14 @@ func (l *logger) ReportMetric(method string, duration time.Duration, err error)
 			if err := prometheus.Register(runTime); err != nil {
 				// metrics may be redundantly registered; ignore these errors
 				if !errors.As(err, &prometheus.AlreadyRegisteredError{}) {
-					l.LogAtLevel(entry, logrus.WarnLevel, fmt.Sprintf("error registering prometheus metric %v: %v", runTime, runTimeName), err)
+					l.LogAtLevel(entry, logrus.WarnLevel, fmt.Sprintf("error registering prometheus metric %v: %v", runTimeName, err))
```
Tryneus, line 159: The changes in this file were just fixes for some warnings I saw in a build.

src/internal/serviceenv/config.go
```diff
@@ -10,22 +10,26 @@ type Configuration struct {
...
+	CacheRoot           string `env:"PACH_CACHE_ROOT,default=/pach-cache"`
+	GCPercent           int    `env:"GC_PERCENT,default=50"`
+	LokiHost            string `env:"LOKI_SERVICE_HOST"`
+	LokiPort            string `env:"LOKI_SERVICE_PORT"`
+	SamlPort            uint16 `env:"SAML_PORT,default=654"`
+	OidcPort            uint16 `env:"OIDC_PORT,default=657"`
+	PostgresServiceHost string `env:"POSTGRES_SERVICE_HOST"`
+	PostgresServicePort int    `env:"POSTGRES_SERVICE_PORT"`
+	PostgresServiceSSL  string `env:"POSTGRES_SERVICE_SSL,default=disable"`
+	PostgresDBName      string `env:"POSTGRES_DATABASE_NAME"`
```
Tryneus, line 32: Moved from `PachdSpecificConfiguration` because these should still be present for workers.

src/internal/serviceenv/service_env.go
```diff
@@ -199,19 +207,10 @@ func (env *NonblockingServiceEnv) initKubeClient() error {
...
-		}
-		port, err := strconv.Atoi(portStr)
-		if err != nil {
-			return err
-		}
-		db, err := dbutil.NewDB(dbutil.WithHostPort(host, port))
+		db, err := dbutil.NewDB(
+			dbutil.WithHostPort(env.config.PostgresServiceHost, env.config.PostgresServicePort),
+			dbutil.WithDBName(env.config.PostgresDBName),
+		)
```
Tryneus, line 213: Using the config values rather than pulling from the env because why would we pull from the env.

src/internal/storage/chunk/option.go
```diff
@@ -65,3 +69,19 @@ func WithNoUpload() WriterOption {
 		w.noUpload = true
 	}
 }
+
+// StorageOptions returns the chunk storage options for the config.
+func StorageOptions(conf *serviceenv.Configuration) ([]StorageOption, error) {
```
Tryneus, line 74: Moved from `serviceenv` to avoid a circular reference.

src/internal/storage/fileset/option.go
```diff
@@ -78,3 +79,18 @@ func WithTTL(ttl time.Duration) WriterOption {
 		w.ttl = ttl
 	}
 }
+
+// StorageOptions returns the fileset storage options for the config.
+func StorageOptions(conf *serviceenv.Configuration) []StorageOption {
```
Tryneus, line 84: Moved from `serviceenv` to avoid a circular reference.

src/internal/tracing/extended/extended_trace.go
```diff
@@ -99,8 +99,8 @@ func PersistAny(ctx context.Context, c *etcd.Client, pipeline string) {
 		tracesCol := TracesCol(c).ReadWrite(stm)
 		return tracesCol.PutTTL(pipeline, traceProto, int64(duration.Seconds()))
 	}); err != nil {
-		log.Errorf("could not persist extended trace for pipeline %q to etcd: %v. ",
-			vals[0], pipeline, err, defaultDuration)
+		log.Errorf("could not persist extended trace for pipeline %q to etcd: %v",
+			pipeline, err)
```
Tryneus, line 103: Another compile warning.

src/internal/deploy/assets/enterprise.go
```diff
@@ -41,9 +39,9 @@ func EnterpriseService(opts *AssetOpts) *v1.Service {
 					NodePort: 31651,
 				},
 				{
-					Port:     auth.OidcPort,
+					Port:     OidcPort,
```
brendoncarroll, line 42: I don't think we should get into the habit of putting configuration as constants in libraries.  If the auth package is parameterized in terms of an OidcPort, then let it accept it where applicable.  When it's time to wire everything up, we should keep all the arbitrary values in one place.

Tryneus: I agree, I think this should be in `serviceenv.Configuration`, but I was trying to avoid additional scope creep in this PR.

src/internal/serviceenv/config.go
```diff
@@ -45,7 +49,7 @@ type PachdFullConfiguration struct {
 type PachdSpecificConfiguration struct {
 	StorageConfiguration
 	NumShards                  uint64 `env:"NUM_SHARDS,default=32"`
-	StorageBackend             string `env:"STORAGE_BACKEND,default="`
+	StorageBackend             string `env:"STORAGE_BACKEND,default=LOCAL"`
```
brendoncarroll, line 52: I don't think it's helpful to have this default.  It seems like when this is not set, it is better to error than silently switch to local.

src/internal/serviceenv/config.go
```diff
@@ -69,10 +73,7 @@ type PachdSpecificConfiguration struct {
 	RequireCriticalServersOnly bool   `env:"REQUIRE_CRITICAL_SERVERS_ONLY",default=false"`
 	MetricsEndpoint            string `env:"METRICS_ENDPOINT",default="`
 	// TODO: Merge this with the worker specific pod name (PPS_POD_NAME) into a global configuration pod name.
```
brendoncarroll, line 75: Is this comment still applicable?

src/internal/serviceenv/option.go
```diff
@@ -1,52 +1,54 @@
...
-	var opts []chunk.StorageOption
-	if conf.StorageUploadConcurrencyLimit > 0 {
-		opts = append(opts, chunk.WithMaxConcurrentObjects(0, conf.StorageUploadConcurrencyLimit))
+// ChunkMemoryCache returns the in memory cache for chunks, pre-configured to the desired size
+func (conf *Configuration) ChunkMemoryCache() kv.GetPut {
+	size := conf.StorageMemoryCacheSize
+	if size < 1 {
+		size = 1
 	}
-	if conf.StorageDiskCacheSize > 0 {
```
brendoncarroll, line 20: What happened to the disk cache?

src/internal/serviceenv/config.go
Tryneus: We can do that, this was just the existing behavior but I made it explicit.  I'll change this to be required.

src/internal/serviceenv/option.go
Tryneus: This was moved into `chunk/option.go` to fix a circular dependency.

src/internal/serviceenv/service_env.go
brendoncarroll: Lol 👍 . Where is the magic that populates the struct with all the `env` tags?

src/internal/serviceenv/option.go
brendoncarroll: And the upload limiter?

src/internal/storage/chunk/option.go
brendoncarroll: Oh cool, these options survived.

src/internal/serviceenv/service_env.go
Tryneus: I had to ask msteffen to point me to it - it's in `cmdutil/env.go`

src/internal/deploy/assets/enterprise.go
Tryneus: Apparently this _is_ in the `serviceenv.Configuration`, and this is just for the default value.  Other ports have a hard-coded default value inlined in `assets.go`, so this is actually a step up from that.  I'll rename it to `DefaultOidcPort` and add a field in the asset options, but ugh.

src/internal/testutil/directory.go
```diff
@@ -0,0 +1,27 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/require"
+)
+
+// MkdirTemp creates a temporary directory with a random name under the OS's
+// temp directory and deletes it at the end of the test.
+func MkdirTemp(t testing.TB) string {
```
brendoncarroll, line 14: This is provided by the testing package.
https://golang.org/pkg/testing/#T.TempDir

Tryneus: That seems useful, I'll move over to that.

src/server/auth/server/api_server.go
```diff
@@ -53,9 +53,6 @@ const (
 	// GitHookPort is 655
 	// Prometheus uses 656
 
-	// OidcPort is the port where OIDC ID Providers can send auth assertions
-	OidcPort = 657
```
brendoncarroll, line 57: I think we are right to move this out of here, but I think it should just go into the deploy package.  The arbitrary config values should not be spread across the codebase.

Values like ports don't matter in isolation, they could be anything.  They matter, and become non-arbitrary when constrained by other port choices.  That happens in the deployment configuration, so that's where we should pick port numbers.

brendoncarroll:
This looks good.  2 things:

As far as config values go, someone needs to be able to rewire our service architecture from one place (the deploy package). Config parameters that connect things should never have a default.  It is almost never useful, and it makes it harder to see how configuration flows through the code.  Knobs, and other tuneables should have a default, so they don't have to be touched. 
So regarding the OidcPort, don't set it in the auth packages or provide any default there, in prod the value comes from the deploy package, in testing it can be a hardcoded value next to the test constructor, but the value used in testing should not be available for others to hardcode.

I'm still against the `testutil` package for reasons I outlined in the previous P.R. You made one argument in favor of it:
> rather than cluttering up libraries with exported test code
Which I understand as a discoverability/cognitive overhead/aesthetic argument.  I don't think that outweighs the loss of clear-box testing.
We can disagree and commit here. I don't want to hold this up.

brendoncarroll (approved):

PR #5983, opened 2021-04-14
Fixes datum skipping bug exposed by cancel many jobs test

brycemcanally:
When performing the second step of the skipping process where we determine whether we can actually skip the datums we recorded as skipped (the parent job output the results of the datums successfully), we need to account for the case where no output was uploaded (as compared to uploading failed, recovered, etc.). The cancel many jobs test exposed this bug because we would incorrectly skip the datum and finish the job before it was canceled.


Tryneus (approved):
LGTM

PR #5985, opened 2021-04-14
RevokeAuthTokensForUser RPC

acohen4:
This RPC is simply executing: `DELETE FROM auth.auth_tokens WHERE subject = $1`


src/server/auth/server/testing/admin_test.go
```diff
@@ -928,6 +928,88 @@ func TestTokenRevoke(t *testing.T) {
...
 }
 
+func TestRevokeTokensForUser(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration tests in short mode")
+	}
+	tu.DeleteAll(t)
+	defer tu.DeleteAll(t)
+
+	contains := func(tokens []*auth.TokenInfo, hashedToken string) bool {
```
actgardner, line 938: I know we use this in several tests, could you extract it out as it's own function? Could be a follow-up PR.

actgardner (approved):
LGTM, one comment but we can address that in the future as well.

src/server/auth/server/testing/admin_test.go
acohen4: good point - I'll merge and submit a follow-up PR

PR #5989, opened 2021-04-15
Usernames are fully qualified when syncing group membership from IDPs

actgardner:
Group syncing wasn't working because we weren't using the fully-qualfiied names for the users (with the `user:` prefix).


acohen4 (approved):
Looks good!

PR #5991, opened 2021-04-17
Respect GOBIN in Makefiles and scripts

msteffen:
Main change:
- Add two files, `etc/govars.mk` and `etc/govars.sh` that set default values for `GOPATH=$HOME/go` (if unset) and `GOBIN=$GOPATH/bin` (if unset)
- Rewrite our other scripts and makefiles to source (scripts) or include (makefiles) one of the above two files, and then use the above vars by replacing all instances of `$GOPATH/bin` with `$GOBIN`
- The result is that scripts' default behavior stays the same, but if `$GOBIN` is set, it overrides the default `$GOPATH/bin`, so that scripts/makefiles are using the same path that `go build` uses when building Go binaries (instead of assuming `$GOPATH/bin`).

Secondary change:
- `etc/govars.mk` and `etc/govars.sh` also set the default value `PACHCTL=$GOBIN/pachctl` (if unset). 
- Rewrite other scripts and makefiles to replace `$GOBIN/pachctl` with `$PACHCTL`
- This way, when using a custom pachctl, you can force scripts/makefiles to use it as well by setting that env var and overriding the default value 
  - This was already done by `examples/Makefile`, `examples/Makefile.testing` and `examples/deferred_processing/automated_deferred_processing/Makefile` but the pattern is now universal to everything that includes `govars.mk` and `govars.sh`

Other miscellaneous changes:
- remove `etc/user-job` (had a Dockerfile for a user-job container that AFAICT is no longer used)
- remove `etc/initdev` (had a bunch of vagrant configs that AFAICT are unused and no longer work)
- remove `src/cmd` from `.gitignore` (it no longer exists)
- remove `.goxc*` files, and remove all such paths from `.gitignore` (we no longer use goxc)
- Remove `src/server/tmp_vendorize_client_go` which AFAICT is no longer used
- Add `.envrc` and `_worktree_data` to `.gitignore`, which can be used to store/set branch-specific data when using git worktrees. 
  - This works well with the `$GOBIN` change: you can work on multiple pachyderm branches simultaneously using git worktrees, and each worktree can set its own value for `$GOBIN` in `.envrc`.
  - When you `cd` into that worktree, `.envrc` in that worktree can set `$GOBIN` to `_worktree_data/bin` and any binaries you build there will be private to the worktree. Because our scripts and makefiles now respect `$GOBIN`, all of our build/test workflows would continue to work, even while multiple branches are checked out and modified simultaneously.


Makefile
```diff
@@ -31,31 +33,17 @@ GORELSNAP = #--snapshot # uncomment --snapshot if you want to do a dry run.
...
-	TIMEOUT = 3600s
-else
-ifndef TIMEOUT
-	# You should be able to specify your own timeout, but by default we'll use the same bound as travis
-	TIMEOUT = 3600s
-endif
-endif
+# Upper bound for travis test timeout
+# You should be able to specify your own timeout, but by default, use the same bound as travis
+TIMEOUT ?= 3600s
```
msteffen, line 38: In Makefiles
```
X ?= foo
```
is the same as
```
ifndef X
  X = foo
endif
```

Makefile
```diff
@@ -31,31 +33,17 @@ GORELSNAP = #--snapshot # uncomment --snapshot if you want to do a dry run.
...
-ifdef TRAVIS_BUILD_NUMBER
-	# Upper bound for travis test timeout
-	TIMEOUT = 3600s
-else
-ifndef TIMEOUT
-	# You should be able to specify your own timeout, but by default we'll use the same bound as travis
-	TIMEOUT = 3600s
-endif
-endif
+# Upper bound for travis test timeout
```
brycemcanally, line 36: We don't use travis in master anymore right? Should the comment be changed to CircleCI?

.gitignore
```diff
@@ -2,6 +2,9 @@
 *~
 *#*
 
+# vim temporary files
+.*.sw*
```
brycemcanally, line 6: This pattern should probably be: `*.sw[a-p]`. It isn't matching some swap files I have locally.

brycemcanally (approved):
Some minor changes, but lgtm after that.

.gitignore
msteffen: Interesting. Both work on my machine, and I'm happy to change it.

I'm curious which swap files it wasn't picking up locally, just because I'm wondering what the bug in `.*.sw*` is, but it's a totally academic interest.

Makefile
msteffen: Done.

.gitignore
brycemcanally: It doesn't match `.swp` or `.swo`. I think the more completely correct gitignore for swap files can be found here: https://github.com/github/gitignore/blob/master/Global/Vim.gitignore. The pattern I provided seems to be a common pattern used for swap files in gitignore.

msteffen: Thanks!

msteffen:
Merging this PR because, despite the GitHub message that build-MISC failed, it actually passed in a subsequent run (that didn't update GitHub): https://app.circleci.com/pipelines/github/pachyderm/pachyderm/3870/workflows/b9b44dcd-10d7-46d7-9c26-32804056c877

I had to cancel the previous run because when CircleCI tried to provision a TestFaster VM, it got stuck in a state where TestFaster kept saying that the VM was assigned but not ready. I cancelled that run and re-ran the failed CircleCI job, and it passed. The link above shows the all-green CircleCI run for commit 9ac96a9

PR #5992, opened 2021-04-19
Upload go test output to BigQuery for test dashboard

actgardner:
This adds a step to CircleCI that picks up the JSON-formatted test output, puts in GCS and then loads it to BigQuery so we can do BI stuff to it. It doesn't handle failures and latency outside the Go tests (ex. VM startup overhead, tests timing out) but we could add scraping of the CircleCI API in the future to get that as well.

Dashboard: https://datastudio.google.com/u/1/reporting/aa374234-80d9-4b24-ac90-8aeab34760cc/page/ig4DC


etc/testing/circle_tests_inner.sh
```diff
@@ -15,6 +15,8 @@ GOPATH=/root/go
 export GOPATH
 PATH="${GOPATH}/bin:${PATH}"
 export PATH
+TESTFLAGS="-json | tee -a /tmp/results.json | jq -j 'select( .Output != null ) | .Output'"
```
brendoncarroll, line 18: So this undoes the JSON, and presents the normal test output?

brendoncarroll (approved):

etc/testing/circle_tests_inner.sh
actgardner: Yeah, it's a little backwards because `-json` actually parses the go-formatted test output and then turns it into JSON, which we turn back into the go-formatted output. Long-term we'll probably switch this and do the JSON (and junit) parsing later.

PR #5996, opened 2021-04-20
BigQuery test data warehouse (take 2)

actgardner:
Buffering in the pipe from `go test` to `tee` was causing tests to time out if they didn't produce enough output. Adding an `stdbuf` on stdin seemed to do the trick. Also, instead of emitting JSON and parsing out the output lines, this dumps all the Go-formatted output to a file for later conversion. This opens the door to produce JUnit reports for Circle summaries as well.


lukemarsden (approved):
This is great!

PR #5998, opened 2021-04-20
Integrating Postgres Collections with Auth

acohen4:
This PR aims to integrate the addition of Postgres Collections to the existing Auth Server

The changes include:

- Migrating all existing collections to postgres collections **except** OIDC logins which are ephemeral and can be kept in etcd
- When available, using the existing SQL TX in calls to the Postgres Auth Tokens table


src/server/auth/server/api_server.go
```diff
@@ -258,7 +264,13 @@ func (a *apiServer) getClusterRoleBinding(ctx context.Context) (*auth.RoleBindin
 }
 
 func (a *apiServer) isActive(ctx context.Context) error {
-	_, err := a.getClusterRoleBinding(ctx)
+	return a.txnEnv.WithReadContext(ctx, func(txnCtx *txnenv.TransactionContext) error {
+		return a.isActiveInTransaction(txnCtx)
```
actgardner, line 268: Is it free to open a transaction if we don't actually use it? The happy path for `getClusterRoleBinding` is that it hits the in-memory cache, so I just want to make sure this isn't adding overhead in that case.

src/server/auth/server/api_server.go
```diff
@@ -672,6 +683,13 @@ func (a *apiServer) CreateRoleBindingInTransaction(txnCtx *txnenv.TransactionCon
 	// Call Create, this will raise an error if the role binding already exists.
 	key := resourceKey(resource)
 	roleBindings := a.roleBindings.ReadWrite(txnCtx.SqlTx)
+	// TODO(acohen4): what should the collision semantics be here?
```
actgardner, line 686: Is it necessary to add a read here? If the role binding key has a uniqueness constraint, whatever transaction commits second should fail and the first transaction will succeed at creating the repo.

src/server/auth/server/testing/auth_test.go
```diff
@@ -2481,7 +2481,6 @@ func TestRestoreAuthToken(t *testing.T) {
 	require.True(t, auth.IsErrExpiredToken(err))
 
 	// restore a token with an expiration date in the future
-	req.Token.HashedToken = fmt.Sprintf("%x", sha256.Sum256([]byte("expiring-token")))
```
actgardner, line 2484: Why is this being removed?

src/server/auth/cmds/cmds.go
```diff
@@ -380,7 +380,7 @@ func WhoamiCmd() *cobra.Command {
 			}
 			fmt.Printf("You are \"%s\"\n", resp.Username)
 			if resp.Expiration != nil {
-				fmt.Printf("session expires: %v\n", *resp.Expiration)
+				fmt.Printf("session expires: %v\n", resp.Expiration.Format(time.RFC822))
```
actgardner, line 383: It's not worth pulling this out, but for unrelated small changes in the future it's better to make a separate PR. It makes understanding the history easier, and reverting things cleaner.

actgardner:
Looks like CI is failing for an unused import. I left a few comment. It would be helpful to have a high level summary in the PR description as well so we know what this was about in a month. For example, I think this is (a) moving oidc nonces back to etcd from postgres (b) propagating transactions to more methods? But it would be good to be clear what the goals are for the PR.

acohen4:
@actgardner thanks for the feedback. I definitely need to clean parts of this up - a few small file changes snuck into my commits and can be kicked out. I'll also be more descriptive

src/server/auth/server/api_server.go
acohen4: not free :( 

thanks for calling this out

src/server/auth/server/api_server.go
acohen4: in tests where we create multiple authenticated clients, each client creation triggers the pfsClient.ActivateAuth RPC. it seems that now whenever we the Creation fails, even though our code gracefully handles the error, the DB Tx is invalidated because part of it failed.

actgardner: Hmmm. If we just wrote SQL we could make it idempotent by using an `on conflict` clause, does the sql collections library support that?

src/server/auth/server/testing/auth_test.go
acohen4: I'm going to remove this change from the PR - i previously had other changes in this file too. 

I originally removed it because technically this line isn't doing anything since this value is being set a few lines earlier

src/server/auth/server/api_server.go
```diff
@@ -46,6 +50,8 @@ const (
 	// GitHookPort is 655
 	// Prometheus uses 656
 
+	// OidcPort is the port where OIDC ID Providers can send auth assertions
```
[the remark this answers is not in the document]
acohen4: Thanks for clarifying

I think this change can also be reverted since it isn't in master anyway

src/server/auth/server/api_server.go
acohen4: @actgardner good point

@Tryneus what do you think is the best solution here? Should we handle this which custom SQL, or add a collections method for Create and do nothing on conflict?

I suppose another alternative would be to see if we could get Postgres to keep the transaction valid even when one INSERT fails - although I'm not sure if that's possible

Tryneus: Ah, yeah, we should change `collection.Create()` to not invalidate the transaction, if possible.  Fix in PR #6002.

acohen4:
hey @actgardner could you give this another look? I cut down most extraneous changes and responded to the previous batch of comments. 

This branch is still being fixed, but the changes I have at least resolve the current small known issues in the AuthServer - specifically 

(1) integrating collections transactions with the auth token storage I worked on recently 
(2) reverting the OIDC collection to be backed by etcd instead of Postgres

cc @Tryneus

actgardner:
@acohen4 this looks good now!

PR #5999, opened 2021-04-20
Add GetPermissions APIs

actgardner:
The way `Authorize` is implemented, the caller sends the set of permissions they want to test. This allows us to exit early as soon as all the permissions are covered without necessarily looking at every role binding.

For `InspectRepo` and `ListRepo` we want the opposite behaviour - we want to test every permission and find all of the roles a principal has been granted, even if they're redundant. This PR adds new RPCs to test the current user's permissions, and also for an admin to check any user's permissions.

I updated the `pachctl auth check repo <repo>` command - it previous took a permission to test. Now it prints all a user's permissions, which may be more useful for debugging.


acohen4 (approved):
LGTM!

PR #6000, opened 2021-04-21
Split GetGroups into a reflexive RPC and GetGroupsForPrincipal

actgardner:
`GetGroups` in 1.x accepted an empty string, which meant "list my groups", or a username which meant "list groups for this user". Our auth system doesn't have a concept of a user as a resource, so this is difficult to express. Instead I just split the RPC in two: "get my groups" is something any user can do, "get someone else's groups" is a cluster-level permission.


acohen4 (approved):
Looks great!

PR #6001, opened 2021-04-21
File part removal

brycemcanally:
This PR removes file parts from the storage layer. After this change, a file will no longer be divided into parts and instead a file & tag pair are considered a distinct entry from a user standpoint all the way down to the storage format (multiple index entries with the same path can exist, but multiple index entries with the same path & tag pair will not). With this change, the merge process was changed such that it operates on a tuple comparison based on the path & tag pairs.


src/client/pfs.go
```diff
@@ -40,11 +40,15 @@ func NewCommitProvenance(repoName string, branchName string, commitID string) *p
 }
 
 // NewFile creates a pfs.File.
-func NewFile(repoName string, branchName string, commitID string, path string) *pfs.File {
-	return &pfs.File{
+func NewFile(repoName string, branchName string, commitID string, path string, tag ...string) *pfs.File {
```
brendoncarroll, line 43: This should just be a string.  The empty string has the same effect as this not being set.
Maybe include an example comment showing that it can be `""`.

brendoncarroll: Even if we use this pattern elsewhere because it's convenient, I don't think we should expose it in the public API.  A file does not optionally take a list of tags.

src/internal/storage/fileset/buffer.go
```diff
@@ -8,118 +8,102 @@ import (
...
 }
 
-func (b *Buffer) Add(p, tag string) io.Writer {
-	p = Clean(p, false)
-	if _, ok := b.additive[p]; !ok {
-		b.additive[p] = &file{
-			path:  p,
-			parts: make(map[string]*part),
-		}
+func (b *Buffer) Add(path string, tag ...string) io.Writer {
```
brendoncarroll, line 28: Same recommendation as above, but this is internal so I don't really care as much here.

src/internal/storage/fileset/buffer.go
```diff
@@ -8,118 +8,102 @@ import (
...
+	}
+	taggedFiles := b.deletive[path]
+	taggedFiles[fileTag] = &file{
+		path: path,
+		tag:  fileTag,
 	}
-	b.deletive[p].parts[tag[0]] = &part{tag: tag[0]}
 }
 
 func (b *Buffer) WalkAdditive(cb func(string, string, io.Reader) error) error {
```
brendoncarroll, line 78: Nit: fill in the parameter names or add a comment so we can determine which is tag and which is path with autocomplete.

src/internal/storage/fileset/index/util.go
```diff
@@ -59,57 +59,17 @@ func SizeBytes(idx *Index) int64 {
...
-			}
+		for _, dataRef := range idx.File.DataRefs {
+			size += dataRef.SizeBytes
 		}
 	}
 	return size
 }
 
-// TODO: Change this such that it returns a new index with the updated fields, rather than updating in place.
-func resolveParts(idx *Index) {
```
brendoncarroll, line 78: 👍

src/internal/storage/fileset/reader.go
```diff
@@ -74,7 +74,6 @@ func (fr *FileReader) Index() *index.Index {
 
 // Content writes the content of the file.
 func (fr *FileReader) Content(w io.Writer) error {
-	dataRefs := getDataRefs(fr.idx.File.Parts)
-	r := fr.chunks.NewReader(fr.ctx, dataRefs)
+	r := fr.chunks.NewReader(fr.ctx, fr.idx.File.DataRefs)
```
brendoncarroll, line 77: This is never modified in the chunk reader right, we just contract the slice without copying anything?

src/server/worker/datum/iterator.go
```diff
@@ -444,18 +444,15 @@ func (ds *datumStream) Next() error {
...
 }
 
-func (ds *datumStream) Key() string {
-	return common.DatumID(ds.meta.Inputs)
-}
-
-func (ds *datumStream) Priority() int {
-	return ds.priority
+func (ds *datumStream) Compare(s stream.Stream) int {
+	return strings.Compare(ds.id, s.(*datumStream).id)
```
brendoncarroll, line 455: This looks like it could panic if it is given something that's not a datumStream.

It's usually better to have a comparison function (like the sort package).  Because it will actually be in a position to know that it will never be comparing non datumStreams.  There is no way of knowing that from the context of this method.

src/client/pfs.go
brycemcanally: Yeah, I kind of have the same intuition about this not being the same as optional configuration. I guess it really is a decision about whether we should support defaults here, or require explicit specification of each of the fields. We could have the default file and tag be empty string since those are supported defaults by several operations. I do like keeping the tag optional though because most users should not need to think about tags when operating on files.

src/internal/storage/fileset/reader.go
brycemcanally: We just iterate over the datarefs, and read the fields. No modifications to the data.

src/server/worker/datum/iterator.go
brycemcanally: Yeah, that can happen in the callback as well. I think the streaming functionality may make moving to a model like that add more complexity than value since the client would need to manage a mapping from indexes to a set of streams. We could add a type check to the constructor to ensure that all of the stream are of the same type if you are concerned about this? I am up for changes to the priority queue if you have a cleaner model in mind though.

src/internal/storage/fileset/index/reader.go
```diff
@@ -16,6 +16,7 @@ type Reader struct {
 	chunks *chunk.Storage
 	filter *pathFilter
 	topIdx *Index
+	tag    string
```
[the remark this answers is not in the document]
brycemcanally: Actually, we should probably just have a default tag other than the empty string and treat the empty string as no filter. That will work better with the empty string default in the protos at the gRPC layer.

src/client/pfs.go
brendoncarroll: If someone sees a parameter they don't recognize, they will probably look at the function documentation and see that it is okay to pass `""`.  A functional option would be fine here too.  But the variadic tags argument is presenting the wrong model to users.

brycemcanally: Okay, I can see that. My general approach to adding new optional parameters to functions that are heavily used is to start with a variadic parameter for the new parameter then functional options if more are needed. We should probably get some more opinions about this considering how heavily used this will be. @jdoliner @Tryneus what do you guys think the signature of this function should look like? Should these just all be required or would having files and / or tags be functional options make sense? It might be a bit weird to have the file path as a functional option, but using the tag should be very rare for most users.

src/server/worker/datum/iterator.go
brendoncarroll: My recommendation is to pass a compare function to the priority queue.  so a `func(a, b Stream) int`.  This is basically the same thing as the current compare method.  However, this will be defined as a lambda in practice, and it will be defined right next to the `streams := []Stream{s1, s2, s3}` .  So the lambda in context will know that all the streams can be cast safely to whatever concrete type.  A public method `Compare(other Stream)`  on the other hand, just assumes it will never compare anything else, which is not a great assumption considering that it’s public.

This is pretty much the same reason why the newer sort methods like Slice take a comparison function, but some older data structure packages define interfaces that do a similar cast to their own concrete type.

src/client/pfs.go
Tryneus: We're currently doing some rework on the way these structures are built in https://github.com/pachyderm/pachyderm/pull/6152 so this may need some of the same treatment.  In general, I'm negative on using a variadic argument for non-variadic data (but I also don't like golang zero-values).  I think after #6152, since people will usually be building files off of commits, the common use case will be `commit.NewFile("path", "tag") which is less of a magical invocation of a tsunami of strings.  I would lean slightly towards not using a variadic argument, though.

brycemcanally: Okay, I am just going to remove the tag field from this constructor for now and readd it as a required parameter once those changes are in.

brendoncarroll (approved):
LGTM

src/client/pfs.go
jdoliner: Coming in late to this thread but I like the decision here.

PR #6002, opened 2021-04-21
Prevent transaction rollback on collection.Create ErrExists error

Tryneus:
Reported in https://github.com/pachyderm/pachyderm/pull/5998#discussion_r617015950.

### DIFF
--- a/src/internal/collection/postgres_collection.go
+++ b/src/internal/collection/postgres_collection.go
@@ -90,7 +90,7 @@ func (c *postgresCollection) ReadWrite(tx *sqlx.Tx) PostgresReadWriteCollection 
 // NewSQLTx starts a transaction on the given DB, passes it to the callback, and
 // finishes the transaction afterwards. If the callback was successful, the
 // transaction is committed. If any errors occur, the transaction is rolled
-// back.  This will reattempt the transaction up to three times.
+// back.  This will reattempt the transaction forever.
 func NewSQLTx(ctx context.Context, db *sqlx.DB, apply func(*sqlx.Tx) error) error {
 	var lastErr error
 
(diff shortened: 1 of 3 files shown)


src/internal/collection/postgres_collection_test.go
```diff
@@ -35,7 +35,7 @@ func TestPostgresCollections(suite *testing.T) {
 			require.NoError(t, listener.Close())
 		})
 
-		testCol := col.NewPostgresCollection(db, listener, &col.TestItem{}, []*col.Index{TestSecondaryIndex}, nil)
```
Tryneus, line 38: Unrelated - these tests had been broken on the integration branch.

src/internal/collection/postgres_collection.go
```diff
@@ -90,7 +90,7 @@ func (c *postgresCollection) ReadWrite(tx *sqlx.Tx) PostgresReadWriteCollection
 // NewSQLTx starts a transaction on the given DB, passes it to the callback, and
 // finishes the transaction afterwards. If the callback was successful, the
 // transaction is committed. If any errors occur, the transaction is rolled
-// back.  This will reattempt the transaction up to three times.
+// back.  This will reattempt the transaction forever.
```
acohen4, line 93: @Tryneus why do we want to re-attempt the transaction forever?

Tryneus: That's actually the existing behavior for `EtcdCollection`, this is just providing the same interface.

acohen4 (approved):
Looks good!

PFedak (approved):
LGTM
learning so much about databases lately, I guess an alternative to this would be a SAVEPOINT + ROLLBACK TO, but this seems much better.

PR #6004, opened 2021-04-22
Simplify constructor arguments

actgardner:
Reduce the amount of code in main, to make it easier to instantiate API servers without a lot of copy-paste. Moving the clusterID generation/fetching to an error group might also reduce time to pod readiness.


msteffen (approved):
LGTM

PR #6005, opened 2021-04-22
Increases the maximum request body size in the s3 gateway.

brendoncarroll:
This provides a workaround for #5942 when multipart uploading is disabled.

### DIFF
--- a/src/server/pfs/s3/master_test.go
+++ b/src/server/pfs/s3/master_test.go
@@ -1,6 +1,7 @@
 package s3
 
 import (
+	"crypto/rand"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


msteffen (approved):
LGTM.

This is probably an obvious question, but are there any cases where setting `requestTimeout` to 20 minutes could cause problems?

The main one I can think of is if there was a user-code bug where the client was connecting and then not sending anything (and the connection would hang around for a long time) but that seems reasonably unlikely to me.

brendoncarroll:
I think the user-code would have to be hanging around for a while, and instead of us closing and giving them an error, the user sees their program is taking a while to run, and has to kill it.  It's definitely not a better way to expose that problem to them.  But there could be other things that make their code spin forever that we don't have control over.

PR #6012, opened 2021-04-23
Recording User Contexts and exposing Synch  contexts command

acohen4:
Changes:
- adding license.clusters columns 
- exposing License.ListUserClusters RPC
- adding sync-contexts command


src/server/enterprise/cmds/cmds.go
```diff
@@ -149,8 +156,10 @@ func RegisterCmd() *cobra.Command {
 		}),
 	}
 	register.PersistentFlags().StringVar(&id, "id", "", "the id for this cluster")
-	register.PersistentFlags().StringVar(&pachdAddr, "pachd-address", "", "the address for the enterprise server to reach this pachd")
+	register.PersistentFlags().StringVar(&pachdAddr, "pachd-address", "", "the address for a user to reach this pachd")
+	register.PersistentFlags().StringVar(&pachdEntAddr, "pachd-enterprise-address", "", "the address for the enterprise server to reach this pachd")
 	register.PersistentFlags().StringVar(&enterpriseAddr, "enterprise-server-address", "", "the address for the pachd to reach the enterprise server")
+	register.PersistentFlags().StringVar(&clusterId, "cluster-deployment-id", "", "the deployment id of the cluster being registered")
```
actgardner, line 162: The admin running this command has access to the cluster being registered (from `client.NewOnUserMachine`). If they don't specify a cluster ID we should be able to get it from https://github.com/pachyderm/pachyderm/blob/master/src/admin/admin.proto#L15

src/server/enterprise/cmds/cmds.go
```diff
@@ -124,11 +125,17 @@ func RegisterCmd() *cobra.Command {
 			}
 			defer ec.Close()
 
+			if pachdEntAddr == "" {
+				pachdEntAddr = pachdAddr
```
actgardner, line 129: Would it make sense to use the address from `c` (the admin's connection to the pachd) rather than defaulting to the enterprise address? We could actually use that as the default for the enterprise connection too? The cases I'm imagining are either the pachd is on the internet - both the enterprise server and the admin can use the internet-facing address - or the users and the enterprise server have to use different addresses because users are port-forwarding or using a VPN or something similar.

src/server/enterprise/cmds/cmds.go
```diff
@@ -192,6 +201,50 @@ func GetStateCmd() *cobra.Command {
...
+	syncContexts := &cobra.Command{
+		Short: "Pull all available Pachyderm Cluster contexts into your pachctl config",
+		Long:  "Pull all available Pachyderm Cluster contexts into your pachctl config",
+		Run: cmdutil.Run(func(args []string) error {
+			cfg, err := config.Read(false, false)
+			if err != nil {
+				return err
+			}
+
+			c, err := client.NewOnUserMachine("user")
```
actgardner, line 214: This should get an enterprise server client with `client.NewEnterpriseClientOnUserMachine("user")`.

src/server/enterprise/cmds/cmds.go
```diff
@@ -192,6 +201,50 @@ func GetStateCmd() *cobra.Command {
...
+			// update the pach_address of all existing contexts, and add the rest as well.
+			for _, cluster := range resp.Clusters {
+				if context, ok := cfg.V2.Contexts[cluster.Id]; ok {
+					context.ClusterDeploymentID = cluster.ClusterDeploymentId
+					context.PachdAddress = cluster.Address
+				} else {
+					cfg.V2.Contexts[cluster.Id] = &config.Context{
+						ClusterDeploymentID: cluster.ClusterDeploymentId,
+						PachdAddress:        cluster.Address,
+						Source:              3, // what's the source? should we populate it here?
```
actgardner, line 233: I don't really know if we use the source for anything, but we could add a new value for "synced from the enterprise server" just in case.

src/server/enterprise/cmds/cmds.go
```diff
@@ -192,6 +201,50 @@ func GetStateCmd() *cobra.Command {
...
+			}
+			defer c.Close()
+			resp, err := c.License.ListUserClusters(c.Ctx(), &license.ListUserClustersRequest{})
+			if err != nil {
+				return err
+			}
+
+			// update the pach_address of all existing contexts, and add the rest as well.
+			for _, cluster := range resp.Clusters {
+				if context, ok := cfg.V2.Contexts[cluster.Id]; ok {
```
actgardner, line 226: I don't know if we need to check for an existing context, we could just blindly overwrite the existing value for the key?

actgardner:
This is looking good - to test against a deployment with the enterprise server separate from pachd, you can add a test here: `https://github.com/pachyderm/pachyderm/blob/master/src/server/enterprise/testing/cmd_test.go`.

To run those tests locally you can run `make launch-enterprise` to deploy the enterprise server in a separate namespace. You might need to give minikube more memory or change the allocations for the default pachd to fit everything into the cluster.

src/server/enterprise/cmds/cmds.go
acohen4: that works! 

my only concern was erasing any useful info that was already set for a context. Is there anything a user could manually set that we would want to preserve?

src/server/enterprise/cmds/cmds.go
acohen4: that makes a lot of sense :)

src/server/enterprise/cmds/cmds.go
acohen4: right now a Source value of 3 represents Imported. It makes sense to stick with that, right?

https://github.com/pachyderm/pachyderm/blob/master/src/internal/config/config.proto#L108

src/server/enterprise/cmds/cmds.go
```diff
@@ -124,11 +125,29 @@ func RegisterCmd() *cobra.Command {
 			}
 			defer ec.Close()
 
+			if pachdUsrAddr == "" {
+				pachdUsrAddr = c.GetAddress()
```
actgardner, line 129: `GetAddress` actually doesn't include the protocol (`grpc` versus `grpcs`). I'm looking at fixing our client package to make these cases easier to handle, can you add a TODO here and I can fix this once I'm done with the client package?

src/server/enterprise/cmds/cmds.go
actgardner: Oh yeah, I completely forgot about the user's auth token, we probably want to keep that 😓 

Thinking about it a little - if the cluster deployment id has changed, the auth token is probably no longer valid and we should set it to an empty string (so the user is prompted to log in). If the deployment id hasn't changed we should keep the auth token so the user doesn't have to re-log-in every time they sync.

src/server/enterprise/testing/cmd_test.go
```diff
@@ -234,3 +235,83 @@ func TestLoginPachd(t *testing.T) {
...
+	// get cluster ID from connection
+	clusterInfo, inspectErr := c.AdminAPIClient.InspectCluster(c.Ctx(), &types.Empty{})
+	require.NoError(t, inspectErr)
+	clusterId := clusterInfo.DeploymentID
+
+	// register a new cluster
+	require.NoError(t, tu.BashCmd(`
+		echo {{.license}} | pachctl enterprise activate
+		echo {{.token}} | pachctl auth activate --enterprise --issuer http://pach-enterprise.enterprise:658 --supply-root-token
+		pachctl enterprise register --id {{.id}} --enterprise-server-address grpc://pach-enterprise.enterprise:650 --pachd-address grpc://pachd.default:650
```
acohen4, line 302: @actgardner the one issue here is that I'm not testing leaving pachd-address argument blank. If I were to do so, the enterprise server will try to look for pachd at an address inaccessible to it. 

Do you think there's a good way to test leaving pachd-address blank?

actgardner: It doesn't feel great to say "don't test it" but we might need to skip it for now - I don't have any great ideas for how to handle this with our current black box approach to testing.

src/server/enterprise/cmds/cmds.go
```diff
@@ -192,6 +213,51 @@ func GetStateCmd() *cobra.Command {
...
+			}
+
+			err = cfg.Write()
+			if err != nil {
+				return err
+			}
+			return nil
+		}),
+	}
+	return cmdutil.CreateAlias(syncContexts, "enterprise sync-contexts")
```
acohen4, line 258: @actgardner what do you think of this command name? Is there anyone else we should consult with?

actgardner: I think that's fine, it fits the convention of how we've been naming enterprise/auth/license/identity CLI commands.

actgardner:
This looks good, one more thing to consider is the `Activate` command (which is used to activate the enterprise server itself), we should probably expose these flags there as well?(https://github.com/pachyderm/pachyderm/blob/master/src/server/enterprise/cmds/cmds.go#L28).

src/license/license.proto
```diff
@@ -65,6 +70,8 @@ message ClusterStatus {
 message UpdateClusterRequest {
   string id = 1;
   string address = 2;
+  string user_address = 3;
```
acohen4, line 73: FYI @actgardner , I just added the fields to update cluster as well

src/server/enterprise/testing/cmd_test.go
```diff
@@ -234,3 +235,113 @@ func TestLoginPachd(t *testing.T) {
...
+		"id", id,
+		"token", tu.RootToken,
+		"license", tu.GetTestEnterpriseCode(t),
+		"clusterId", clusterId,
+		"userAddress", "grpc://pachd.default:700",
+	).Run())
+
+	// re-register cluster with a new cluster ID
+	// the cluster id should be updated and the session token should be set to empty
+	// TODO(acohen4): set session_token so that it can be unset
```
acohen4, line 294: @actgardner didn't find a good way to do this. I think to test this we would need both an Active Cluster that we can log into, and be able to redeploy that cluster so that it has a new deployment ID. 

What do you think?

src/server/license/server/api_server.go
```diff
@@ -344,13 +346,48 @@ func (a *apiServer) UpdateCluster(ctx context.Context, req *lc.UpdateClusterRequ
 	a.LogReq(req)
 	defer func(start time.Time) { a.pachLogger.Log(req, resp, retErr, time.Since(start)) }(time.Now())
 
-	if err := a.validateClusterConfig(ctx, req.Address); err != nil {
-		return nil, err
+	// should we only validate the pachd Address accessible to the enterprise server?
```
acohen4, line 349: surfacing this comment/question:

should we only validate the pachd Address accessible to the enterprise server?

src/server/license/server/api_server.go
```diff
@@ -344,13 +346,48 @@ func (a *apiServer) UpdateCluster(ctx context.Context, req *lc.UpdateClusterRequ
...
+	if req.Address != "" {
+		if err := a.validateClusterConfig(ctx, req.Address); err != nil {
+			return nil, err
+		}
+	}
+
+	// is there semantic value to updating a cluster's address or user_address to empty string?
```
acohen4, line 356: surfacing this comment/question:

is there semantic value to updating a cluster's address or user_address to empty string?

acohen4:
now that I look at `enterprise activate`, I'm a little confused by the hard-coding of localhost for the cluster. 

Is someone who's activating a non-local cluster running a different flow?

src/server/enterprise/cmds/cmds.go
actgardner: This fix is actually in master, if you have the time to rebase? `GetAddress` now returns `grpcutil.PachdAddress` which has a method to get the fully-qualified address.

src/license/license.proto
actgardner: Good catch!

src/server/enterprise/testing/cmd_test.go
actgardner: Yeah I don't know that we have a good easy way to test that edge case, I'm OK to punt on it for now.

src/server/license/server/api_server.go
actgardner: I'm not sure I follow, is the question whether we should check whether the enterprise server can connect to the URI that the end users will get from syncing? I don't think so - the goal is to support a case where the users and the enterprise server have different network topologies, such that they reach the pachd via different addresses. In cases were the `enterprise -> pachd` and `client -> pachd` routes are different, I think it's likely we can't use the client URI to reach the pachd from the enterprise server.

src/server/license/server/api_server.go
actgardner: I don't think setting them to an empty string makes sense - in a lot of cases they're probably the same value though? I'm not sure if we can leverage that for updates at all.

actgardner:
> now that I look at enterprise activate, I'm a little confused by the hard-coding of localhost for the cluster.

> Is someone who's activating a non-local cluster running a different flow?

`enterprise activate` is used for two use cases:
- The user has a single pachd, and they're activating an enterprise license. In this case the pachd is also running the license service for itself.
- The user is activating a new enterprise server. Likewise, in this case the enterprise server is registered with itself.

actgardner:
I'm sorry to keep piling things on, but I was working on something else and I remembered that we added the `enterprise_server` field to the context. We should also add it to the table of registered servers and sync it to config: https://github.com/pachyderm/pachyderm/blob/master/src/internal/config/config.proto#L99-L101

src/server/license/server/api_server.go
acohen4: ok cool 👍  that's what I was thinking too, but wasn't sure if maybe there's something to consider there

src/server/license/server/api_server.go
acohen4: that makes total sense. I think I'm trying to get at whether there's some validation worth performing on the user_address. At least with the network accessible pachd address, we are confident in its validity. 

I guess the only thing we can really do is validate the protocol. Does that seem worth doing? I guess that's something that's implied by how we configure the server.

I think I answered my own question here lol - nothing to validate for that value. Feel free to chime in if I'm off.

src/license/license.proto
```diff
@@ -65,6 +72,9 @@ message ClusterStatus {
...
+  string cluster_deployment_id = 4;
+  bool enterprise_server = 5;
```
acohen4, line 77: Since `enterprise_server` is a boolean field in an update call, we always interpret a value even if the caller didn't explicitly set it. 

@actgardner let me know if you have any concern here. if it's not a useful enough field in the context of the Update RPC, we could also cut it out.

src/license/license.proto
```diff
@@ -43,7 +43,14 @@ message AddClusterRequest {
   // to authenticate to the license server. Otherwise a secret will be
   // generated and returned in the response. 
   string secret = 3;
+  // The pachd address for users to connect to. 
+  string user_address = 4;
+  // The deployment ID value generated by each Cluster
+  string cluster_deployment_id = 5;
+  // the enterprise server flags whether this cluster is registered with the enterprise server
```
actgardner, line 50: The comment is kind of misleading, this field indicates whether the address points to an enterprise server.

src/license/license.proto
actgardner: Yeah let's not support updating it - in the worst case a user can delete the cluster and recreate it.

actgardner (approved):

actgardner:
One nit about the enterprise field comment but otherwise this LGTM 👍

PR #6014, opened 2021-04-24
Add -0 to xargs in etc/testing/lint.sh, and fix lint errors

msteffen:
I suspect this is a difference between BSD xargs and GNU xargs, but previously `find ... --print0 | xargs` in `etc/testing/lint.sh` wasn't being parsed correctly in CI because `xargs` couldn't handle the null bytes that `find` was printing. This PR add `--null` to xargs, causing it to parse its arguments as a null-delimited list.

This caused `make lint` to start parsing all of `find`'s output in CI again, which in turn exposed some lint errors (mostly missing comments for exported functions, with two `Id -> ID` variable naming issues and one "function returns unexported type" issue) that had accumulated, which this PR also fixes.


actgardner:
Yeah this fails for me on OS X:
```
make lint
etc/testing/lint.sh
xargs: illegal option -- -
usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements] [-S replsize]]
             [-J replstr] [-L number] [-n number [-x]] [-P maxprocs]
             [-s size] [utility [argument ...]]
```

We can fix it by installing GNU xargs:

```
brew install findutils
export PATH="/usr/local/opt/findutils/libexec/gnubin:$PATH"
```

We could potentially add this to onboarding docs?

msteffen:
> Yeah this fails for me on OS X:

One option could be to set the args with
```
declare -a xargs_opts; if [[ $(uname) != "Darwin" ]]; then xargs_opts+=( "--null" ); fi
```
but after stealing Rosalind's mac, I think `-0` actually works for both, so I just replaced `--null` with `-0`—any chance that works on your machine?

actgardner:
`-0` seems to work on darwin, nice find :)

msteffen:
Closing this PR as obsolete

PR #6017, opened 2021-04-25
[DOC] fix protobufs location

jhawk28:
the client docs are pointing to an old/incorrect location


nadegepepin (approved):
LGTM! Thanks for the help!

nadegepepin:
Added PR #6040 to broadcast the change to other versions of the doc.

PR #6018, opened 2021-04-26
[Example] First pass at a simple example of S3 protocol enabled pipeline

nadegepepin:


nadegepepin:
This branch was shared to validate an example whose the pipeline's specification would land in the doc. The doc is now updated https://docs.pachyderm.com/latest/deploy-manage/manage/s3gateway/deploy-s3gateway-sidecar/#s3-enable-your-pipeline - The test example is no more needed.

PR #6019, opened 2021-04-26
Don't read STORAGE_UPLOAD_CONCURRENCY_LIMIT directly from the env

actgardner:
`STORAGE_UPLOAD_CONCURRENCY_LIMIT` and `STORAGE_BACKEND` are two environment variables we read directly from the environment instead of from the ServiceEnv. This only fixes the first one because https://github.com/pachyderm/pachyderm/blob/master/src/internal/obj/factory.go is pretty hairy by comparison.


msteffen (approved):
LGTM--nice!

PR #6020, opened 2021-04-26
Forward PFS & PPS sidecar requests that use watches to full pachd instances

brycemcanally:
This PR includes some changes that enable us to forward PFS & PPS sidecar requests to a full pachd instance (through the pachd service) so that we can limit the number of postgres connections dedicated to watches to the number of full pachd instances. The changes are:
- Adding a service flag to `GetPachClient` to indicate that the pach client should be connected to pachd through the pachd service.
- Adding sidecar api servers for PFS & PPS that forward requests that use watches. For PFS, the RPCs are `InspectCommit`, `FlushCommit`, and `SubscribeCommit`. For PPS, the RPCs are `InspectJob`.


Tryneus:
Closed in favor of #6403

brycemcanally:
Addressed by: https://github.com/pachyderm/pachyderm/pull/6403

PR #6022, opened 2021-04-26
Pipefail for go test

actgardner:
We're piping stdout to tee in the integration tests, but we don't have `pipefail` set so Make may still return with a 0 exit code. This ensures the tests report a failure if `go test` returns a non-zero value.


msteffen (approved):
Nice catch!

PR #6023, opened 2021-04-26
Tests produce Junit output

actgardner:
This is a freebie since we already pipe stdout into a file, we can convert it to JUnit XML and get test reports from Circle. This doesn't work for non-Go tests or some failures where the test binary stack traces, but in some cases it makes parsing the failure output simpler.

Example of useful and less-useful output depending on the nature of the test failure:
Screen Shot 2021-04-26 at 4 43 51 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,9 @@ jobs:
           command: |
             etc/testing/upload_stats.sh
 
+      - store_test_results:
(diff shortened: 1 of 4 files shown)


lukemarsden:
:partyparrot:

lukemarsden (approved):

msteffen (approved):
Very cool!

PR #6024, opened 2021-04-27
Remove the replace of the k8s library

chainlink:
Slowly working through go mod upgrades to the k8s package

In this edition, we bump autorest and remove the replace to the custom k8s version and back to mainline


chainlink:
@lukemarsden This builds locally, is there a testfaster cache we need to blow out here?

lukemarsden:
try bumping this hash to the commit on your branch https://github.com/pachyderm/pachyderm/blob/remove-k8s-replace/.testfaster.yml#L63

PR #6027, opened 2021-04-27
[Parallel Tests] -  Add Logger to serviceenv

actgardner:
Make it possible to redirect (most) logs to a file by injecting a logger from serviceenv.


msteffen (approved):
LGTM

PR #6029, opened 2021-04-27
[Parallel Tests] - Add Ready channel for test serviceenv

actgardner:
What it says on the tin. Also pass credentials through `GetPachClient` in the testing service env.

### DIFF
--- a/src/internal/serviceenv/testing.go
+++ b/src/internal/serviceenv/testing.go
@@ -22,6 +22,11 @@ type TestServiceEnv struct {
 	LokiClient    *loki.Client
 	DBClient      *sqlx.DB
 	Ctx           context.Context
+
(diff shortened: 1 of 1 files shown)


msteffen (approved):
Overall LGTM.

This seems slightly redundant with `NonblockingServiceEnv` but trying to factor this behavior out seems might be over-abstracting, unless you see like an awesome way to do it?

actgardner:
> This seems slightly redundant with NonblockingServiceEnv but trying to factor this behavior out seems might be over-abstracting, unless you see like an awesome way to do it?

Honestly the more I work with the testing serviceenv the more I feel like we could tweak the non-blocking service env and eliminate it altogether. My thinking right now is that I'll try and land the testing code and then see if I can eliminate the testing serviceenv.

PR #6030, opened 2021-04-27
[Parallel Tests] - Add postgres database to sidecar environment

actgardner:
Right now sidecars don't have the postgres database set as an environment variable, so they all assume the default database name.


msteffen (approved):
LGTM

PR #6031, opened 2021-04-27
[Parallel Tests] - Add optional filter label to worker pods

actgardner:
Right now pachd assumes every worker pod in a namespace belongs to it. I started looking at supporting multiple pachds using multiple namespaces, but it gets hairy to handle exposing the etcd and postgres services to the sidecars in every namespace. Debugging is also a pain.

This is an alternate approach where pachds can attach a label to their worker pods, and only care about pods with their label. This allows us to run pipelines from multiple pachds in a single namespace without them clobbering each other.

### DIFF
--- a/src/internal/serviceenv/config.go
+++ b/src/internal/serviceenv/config.go
@@ -19,6 +19,7 @@ type GlobalConfiguration struct {
 	S3GatewayPort       uint16 `env:"S3GATEWAY_PORT,default=600"`
 	PPSEtcdPrefix       string `env:"PPS_ETCD_PREFIX,default=pachyderm_pps"`
 	Namespace           string `env:"PACH_NAMESPACE,default=default"`
+	PipelineLabel       string `env:"PACH_WORKER_LABEL"`
 	StorageRoot         string `env:"PACH_ROOT,default=/pach"`
 	CacheRoot           string `env:"PACH_CACHE_ROOT,default=/pach-cache"`
 	GCPercent           int    `env:"GC_PERCENT,default=50"`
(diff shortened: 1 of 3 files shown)


chainlink:
Does this use one etcd and one PG for all the pachs?

actgardner:
This PR just adds the filter label to the worker pods, but yes in this model we would use one common etcd and postgres with different prefixes and databases for each test.

src/internal/serviceenv/config.go
```diff
@@ -19,6 +19,7 @@ type GlobalConfiguration struct {
 	S3GatewayPort       uint16 `env:"S3GATEWAY_PORT,default=600"`
 	PPSEtcdPrefix       string `env:"PPS_ETCD_PREFIX,default=pachyderm_pps"`
 	Namespace           string `env:"PACH_NAMESPACE,default=default"`
+	PipelineLabel       string `env:"PACH_WORKER_LABEL"`
```
msteffen, line 22: Apologies if I just missed it, but I couldn't figure out right away: how does this get set?

actgardner: This is only set by the testing framework, to specify which test pachd each worker belongs to. It's not really used in production at all.

actgardner:
@msteffen @chainlink Do y'all have strong objections to this approach? The PR kind of stalled out, this is the last thing blocking adding the parallel testing framework.

chainlink:
I'm good with this to unblock testing but would like to circle back on the multi namespace complexity as we're likely going to want to use namespaces for multi tenancy.

PR #6036, opened 2021-04-28
Validate parameters of StopJob and DeleteJob

Tryneus:
Better to not segfault if these parameters are omitted.


msteffen (approved):
LGTM, nice catch!

PR #6038, opened 2021-04-29
Test PPS master stuckness

msteffen:
From investigating some of the recent hangs, the evidence I've been able to collect suggests that the kubernetes API server isn't responding to pachd requests, and it's causing the pipeline controller to block on `GetRC`. This PR adds some debug logging to the pipeline controller to monitor these requests more closely and, more importantly, adds both client-side and server-side timeouts to these requests. I don't know if this new behavior is due to the (slightly, in my estimation) increased load from the autoscaling goro or if, possibly, GKE has just been unhealthy recently and all GKE-based pachyderm clusters are having this issue, or something else completely.

My plan is to build a one-off test image and deploy it into the clusters that are known to be having issues right now and see if it helps.

One note on implementation:
The (old) version of the kubernetes API client that we use only lets you set client-side timeouts during client creation AFAICT. To avoid the overhead of redundant TCP handshakes and such, I wanted to create a single kube-client with timeouts in serviceenv. I also wasn't sure it would be safe to impose timeouts on all Kubernetes requests (e.g. would this timeout cause the watches in `pollPipelinePods` or `NewPool` to time out after 30s?) so I altered serviceenv to initialize a second kube client with the timeout, and I changed pachd to use it in the places that seem to be locking up.


msteffen:
Closing PR, as it's obsolete.

PR #6039, opened 2021-04-29
Rename Job -> PipelineJob

Tryneus:
In preparation for global IDs, which will implement a new protobuf for Jobs.


PFedak (approved):
LGTM, I'm assuming you used some form of search (`\b(jobinfo)|(ji)`, maybe?) or automation so there's not too much value in me doing that, too.

Tryneus:
Yeah, it was mostly just some fancy greps. :+1:

PR #6042, opened 2021-04-29
[1.13.x] - Recreate eventCh when master restarts

actgardner:
We used to close `m.eventCh` in a defer in `run`, but we call `run` in an infinite loop. The first time the PPS master loses the etcd lock it would shut everything down, close the channel, and then re-acquire the lock and try to start them, only to have them send their events on a closed channel.


msteffen (approved):
Good catch!

PR #6048, opened 2021-05-03
Add more granular roles

actgardner:
Break up some of the permissions in `clusterAdmin` so users can delegate specific tasks without giving full admin privileges.

An example is Hub, where we want to give `allClusterUsers` the `debugger`, `repoOwner` and `robotUser` roles but not full cluster admin. Giving users the ability to register new OIDC apps without giving them access to read all repos is another use case.

/cc @nadegepepin - It's not a priority for 2.0 but we could document these additional roles?
/cc @seslattery - When the next alpha drops we should be able to use these in Hub.

### DIFF
--- a/src/auth/auth.go
+++ b/src/auth/auth.go
@@ -65,6 +65,26 @@ const (
 	// RepoReaderRole is a role which grants ability to both read from a repo
 	RepoReaderRole = "repoReader"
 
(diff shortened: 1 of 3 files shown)


acohen4 (approved):
Looks great!

PR #6050, opened 2021-05-03
Run tests in both Circle and TestFaster

actgardner:
Run tests in a large Circle VM and a TestFaster VM at the same time. This doesn't include any optimizations for Circle builds, but it passes consistently in ~20 minutes.


nitinjainsj (approved):
LGTM

PR #6051, opened 2021-05-03
Fix TestRegisterDefaultArgs to expect pachd address to point at the Client's configured host

acohen4:


actgardner (approved):
LGTM, thanks for turning this around quickly!

PR #6055, opened 2021-05-04
RevokeTokensForUser is a cluster permission

actgardner:
`/auth.API/RevokeAuthTokensForUser` was in the auth middleware as `authenticated`, but it has some abuse potential. This makes it a cluster permissions reserved for admins.


acohen4 (approved):
LGTM!

PR #6057, opened 2021-05-04
Prevent users from revoking root and pps tokens

actgardner:
This is a small change to make it harder for users to break their cluster.


acohen4 (approved):
good catch!

PR #6060, opened 2021-05-04
Rotate Auth Token gRPC + CLI

acohen4:
Changes:

- gRPC implementation to handle token rotation
- CLI command (`pachctl auth rotate-root-token`) that exposes gRPC
- gRPC + CLI tests
- Include gRPC in Admin Role permissions


src/server/auth/server/api_server.go
```diff
@@ -362,6 +362,48 @@ func (a *apiServer) Activate(ctx context.Context, req *auth.ActivateRequest) (re
...
+
+	// If the token hash was in the request, use it and return an empty response.
+	// Otherwise generate a new random token.
+	pachToken := req.PachToken
+	if pachToken == "" {
+		pachToken = uuid.NewWithoutDashes()
+	}
+
+	var err error
+	if req.TTL > 0 {
```
acohen4, line 395: I'm conflicted as to whether to expose TTL in the command. On one hand it makes sense to give the admin the power to control the TTL of a new token for a user. On the other by carelessly omitting a TTL value, they grant the token indefinite lifetime.

We could instead have the default value be a Standard length of time, and to declare indefinite lifetime, we could pass another flag, or pass a negative value.

What are your thoughts here @actgardner ?

acohen4:
hey @actgardner , the PR you recently made to prevent token revocation for root and pps users brought up a timely edge case to mind. 

Should token rotation work normally for the root user? I ask because a rotation implies a revocation of previous tokens followed by the creation of new ones

actgardner:
@acohen4 I think I didn't provide a great explanation of this ticket, that's my bad. The problem we're trying to solve is that for most users, if you revoke their tokens because they leaked, they can just log in again. Or if it's a robot user, you make a new token. That doesn't need to be an atomic operation because there's no risk of totally losing access to the cluster.

Atomically rotating tokens _only_ makes sense for the root token, because we have an invariant that there must always be one root token (to ensure users don't get locked out of the cluster). We need some way for the root token to be revoked if it was committed to a git repo, for instance, but we also need to ensure the user gets a new token. This isn't true for any other type of token, because there's no risk of users permanently losing access to the cluster.

acohen4:
@actgardner thanks for clarifying, I think I misunderstood one of our earlier conversations. I'll make the revision

src/server/auth/server/api_server.go
```diff
@@ -362,6 +362,34 @@ func (a *apiServer) Activate(ctx context.Context, req *auth.ActivateRequest) (re
 	return &auth.ActivateResponse{PachToken: pachToken}, nil
 }
 
+// RotateToken implements the protobuf auth.RotateToken RPC
+func (a *apiServer) RotateAuthToken(ctx context.Context, req *auth.RotateAuthTokenRequest) (resp *auth.RotateAuthTokenResponse, retErr error) {
```
actgardner, line 366: We probably want the name to be like `RotateRootToken` to make it clear it only works for the root user.

src/server/auth/server/api_server.go
```diff
@@ -362,6 +362,34 @@ func (a *apiServer) Activate(ctx context.Context, req *auth.ActivateRequest) (re
...
+	// We don't want to actually log the request/response since they contain credentials.
+	defer func(start time.Time) { a.LogResp(nil, nil, retErr, time.Since(start)) }(time.Now())
+
+	// TODO(acohen4): Transactionify so that token deletion and insertion are atomic
+	// First revoke this subject's existing auth_tokens
+	if err := a.deleteAuthTokensForSubject(ctx, auth.RootUser); err != nil {
+		return nil, err
+	}
+
+	// If the token hash was in the request, use it and return an empty response.
```
actgardner, line 378: The request will have the plaintext token, not the hash afaik.

src/server/auth/cmds/cmds.go
```diff
@@ -681,6 +681,36 @@ func GetEnterpriseRoleBindingCmd() *cobra.Command {
...
+			defer c.Close()
+
+			req := &auth.RotateAuthTokenRequest{
+				PachToken: pachToken,
+			}
+			resp, err := c.RotateAuthToken(c.Ctx(), req)
+			if err != nil {
+				return err
+			}
+			fmt.Printf("Pachyderm auth token:\n%s\n", resp.PachToken)
```
actgardner, line 705: We should make this explicit that it's the root token, with the same caveats as `Activate` that the old root token will be revoked and this is the only way to authenticate if auth is broken.

src/server/auth/cmds/cmds.go
```diff
@@ -681,6 +681,39 @@ func GetEnterpriseRoleBindingCmd() *cobra.Command {
 	return cmdutil.CreateAlias(get, "auth get enterprise")
 }
 
+// RotateRootToken returns a cobra command that rotates the auth token for the Root User
+func RotateRootToken() *cobra.Command {
+	var rootToken string
+	rotateToken := &cobra.Command{
+		Use:   "{{alias}}",
+		Short: "Rotate the auth token for a user",
```
actgardner, line 689: nit: The usage should reflect that this only works for the root user and that it revokes any existing root token.

src/server/auth/server/api_server.go
```diff
@@ -362,6 +362,34 @@ func (a *apiServer) Activate(ctx context.Context, req *auth.ActivateRequest) (re
 	return &auth.ActivateResponse{PachToken: pachToken}, nil
 }
 
+// RotateRootToken implements the protobuf auth.RotateRootToken RPC
+func (a *apiServer) RotateRootToken(ctx context.Context, req *auth.RotateRootTokenRequest) (resp *auth.RotateRootTokenResponse, retErr error) {
+	pachClient := a.env.GetPachClient(ctx)
+	ctx = pachClient.Ctx() // copy auth information
```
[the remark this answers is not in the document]
acohen4: Yeah this is unnecessary. I mistakenly though the pachClient was loading auth info into the context.

This same logic is being run here: https://github.com/pachyderm/pachyderm/blob/master/src/server/auth/server/api_server.go#L304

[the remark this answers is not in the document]
acohen4: got it! it's clear to me now. I was missing the bit of knowledge around Metadata getting dropped in outgoing context creation, without further intervention

src/server/auth/server/api_server.go
```diff
@@ -362,6 +362,32 @@ func (a *apiServer) Activate(ctx context.Context, req *auth.ActivateRequest) (re
 	return &auth.ActivateResponse{PachToken: pachToken}, nil
 }
 
+// RotateRootToken implements the protobuf auth.RotateRootToken RPC
+func (a *apiServer) RotateRootToken(ctx context.Context, req *auth.RotateRootTokenRequest) (resp *auth.RotateRootTokenResponse, retErr error) {
+	// We don't want to actually log the request/response since they contain credentials.
+	defer func(start time.Time) { a.LogResp(nil, nil, retErr, time.Since(start)) }(time.Now())
+
+	// TODO(acohen4): Transactionify so that token deletion and insertion are atomic
```
actgardner, line 370: Is it a big lift to make this atomic? I would feel better if this was wrapped in a transaction before it hits master, but I might be missing context.

acohen4: nah not a big lift. My take has been that it will be easy to add once postgres-integration is merged. 
I'll add the transaction wrapping and leave a note to migrate it to some of the library code written in postgres-integration

actgardner (approved):

Issue #6058, opened 2021-05-04
pachd correctly propagates new volumes to sidecars, but not mounts


pappasilenus:
**What happened?**:

We created an emptyDir volume and mounted it on /tmp to get Pachyderm to run in a restricted-privileges environment on OpenShift.  Pachd ran correctly, and propagated its custom-configured volumes configuration to the sidecars correctly, but didn't propagate the mounts.  We need to manually edit the RC's to make the mounts appear.

**What you expected to happen?**:

Pachd should correctly propagate custom configurations to sidecars.

**How to reproduce it (as minimally and precisely as possible)?**:

Add a custom volume and mount to pachd. 
Create a pipeline
Note that the mount doesn't exist in the pipeline's RC

**Anything else we need to know?**:

Essential to running on OpenShift

**Environment?**: OpenShift
- Kubernetes version (use `kubectl version`): 
- Pachyderm CLI and pachd server version (use `pachctl version`):
- Cloud provider (e.g. aws, azure, gke) or local deployment (e.g. minikube vs dockerized k8s):
- OS (e.g. from /etc/os-release):
- Others:

actgardner:
As mentioned in https://github.com/pachyderm/pachyderm/issues/6056, we could use a system-wide default pod patch to support mounting arbitrary volumes into every RC. 

For the specific case of `/tmp` I think it makes sense to just create the `/tmp` dir with the appropriate permissions in our docker images, so it exists all the time.

mindthevirt:
<div>Agent Dale Georg linked Freshdesk ticket <a href="https://pachyderm-support.freshdesk.com/a/tickets/262">262</a> for this issue.</div>

PR #6065, opened 2021-05-05
Ensure /tmp, /pach-cache exist in the pachd image

PFedak:
We still create directories under `os.TempDir()` in `withTmpFile`, which is used for legacy spouts as well as debug dumps. I imagine there's little harm in moving these under the same directory, but I wonder if @actgardner's suggestion [here](https://github.com/pachyderm/pachyderm/issues/6058#issuecomment-832107649) starts to make more sense.

### DIFF
--- a/src/server/pfs/server/driver.go
+++ b/src/server/pfs/server/driver.go
@@ -141,7 +141,7 @@ func newDriver(
 		return nil, err
 	}
 	if env.StorageDiskCacheSize > 0 {
-		diskCache, err := obj.NewLocalClient(filepath.Join(os.TempDir(), "pfs-cache", uuid.NewWithoutDashes()))
+		diskCache, err := obj.NewLocalClient(filepath.Join(env.CacheRoot, "pfs-cache", uuid.NewWithoutDashes()))
 		if err != nil {
 			return nil, err
 		}
--- a/src/server/pkg/serviceenv/option.go
+++ b/src/server/pkg/serviceenv/option.go
@@ -1,15 +1,12 @@
 package serviceenv
 
 import (
-	"os"
 	"path/filepath"
 
 	"github.com/pachyderm/pachyderm/src/server/pkg/obj"
 	"github.com/pachyderm/pachyderm/src/server/pkg/storage/chunk"
 	"github.com/pachyderm/pachyderm/src/server/pkg/storage/fileset"
 )
-
-var localDiskCachePath = filepath.Join(os.TempDir(), "pfs-cache")
 
 // ChunkStorageOptions returns the chunk storage options for the service environment.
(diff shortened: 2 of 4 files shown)


src/server/pkg/serviceenv/option.go
```diff
@@ -1,24 +1,21 @@
...
-
 // ChunkStorageOptions returns the chunk storage options for the service environment.
 func (env *ServiceEnv) ChunkStorageOptions() ([]chunk.StorageOption, error) {
 	var opts []chunk.StorageOption
 	if env.StorageUploadConcurrencyLimit > 0 {
 		opts = append(opts, chunk.WithMaxConcurrentObjects(0, env.StorageUploadConcurrencyLimit))
 	}
 	if env.StorageDiskCacheSize > 0 {
-		diskCache, err := obj.NewLocalClient(localDiskCachePath)
+		diskCache, err := obj.NewLocalClient(filepath.Join(env.CacheRoot, "pfs-cache"))
```
brycemcanally, line 18: This is the V2 code path, which isn't exercised in the 1.13.x branch, so this isn't really necessary but it is fine to leave it. We have a PR to remove this code, we just haven't merged it in.

brycemcanally:
Either approach seems fine to me, but I would probably lean towards guaranteeing that `/tmp` exists since it would not require us to take into account configuration for any temporary files we want to create.

actgardner:
If you want to check whether pachd comes up without root access, you can just add a security context to the pachd deployment and run as a non-root userId: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod

Whatever directory we pick we'll need to create in the root of the file system - the scratch image doesn't have `/pach` or `/tmp`.

src/server/pfs/server/driver.go
```diff
@@ -141,7 +141,7 @@ func newDriver(
 		return nil, err
 	}
 	if env.StorageDiskCacheSize > 0 {
-		diskCache, err := obj.NewLocalClient(filepath.Join(os.TempDir(), "pfs-cache", uuid.NewWithoutDashes()))
```
actgardner, line 144: If we're creating `/tmp` do we need to also move these to CacheRoot? I guess we're adding both to the image so it should work anyways, I'm just not sure if I'm missing anything.

actgardner (approved):
One comment about moving the caches into CACHE_ROOT (why weren't they there already) but I think this will solve the issue, +1.

src/server/pfs/server/driver.go
PFedak: No, it's definitely not necessary, it just seemed more consistent (and what I thought would fix the issue initially). I can put it back - I don't imagine too many users are concerned with the actual path we use.

PR #6066, opened 2021-05-05
Require PFS commits to be on a branch

Tryneus:
With Global IDs, multiple commits may be accessible in the same repo with the same ID (due to changing multiple branches in a repo within a single transaction, or other more convoluted ways).  While this will likely not be a common issue once we move stats to a system repo, it must still be possible to disambiguate between multiple commits in the same repo with the same global ID.  The first (and most tedious) step towards this is requiring all commits to be made on a branch.  Following this PR, a commit can no longer _not_ be on a branch in a repo.  In addition, in preparation for the aforementioned disambiguation, any place in the API where we specify a commit ID, we may also specify a branch name.

In this PR:
 * PFS proto changes:
   * Replaced `Commit.Repo` with `Commit.Branch`, as the repo can now be accessed in the `Branch`
   * Removed `CommitInfo.Branch`, it should now be accessed in `CommitInfo.Commit.Branch`
   * `CommitProvenance.Branch` removed, as the branch can now be accessed in `CommitProvenance.Commit.Branch`
   * `StartCommitRequest.Branch` is no longer optional and has been changed from a `string` to a `Branch` so that `StartCommitRequest.Commit` may be omitted entirely since we can get the repo info from the `Branch`
 * All PFS client wrapper functions that involve commits also include a branch name parameter
 * When specifying a commit by branch rather than ID, we will now construct a proto like:
```diff
Commit{
+ Branch: {
    Repo: {
      Name: "data"
    },
+   Name: "master",
+ },
- ID: "master",
}
```
 * As a side-effect, some code assumed anyone could start a commit on an output repo as long as it was not on a branch.  For the time being, I have removed the check that disallows starting a commit on a branch in an output repo so that this could still work.


src/server/pfs/server/driver.go
```diff
@@ -765,28 +739,23 @@ func (d *driver) makeCommit(
 // If a complete commit provenance is passed in it just uses that.
 // It accepts an STM so that it can be used in a transaction and avoids an
 // inconsistent call to d.inspectCommit()
-func (d *driver) resolveCommitProvenance(stm col.STM, userCommitProvenance *pfs.CommitProvenance) (*pfs.CommitProvenance, error) {
+func (d *driver) resolveCommitProvenance(stm col.STM, userCommitProvenance *pfs.CommitProvenance) (*pfs.CommitProvenance, *pfs.CommitInfo, error) {
```
jdoliner, line 742: This was already the case but it strikes me as a little bit weird that this mutates one of its arguments and then returns it rather than just mutating it and having the user use it? Is there a good reason to do that?

jdoliner (approved):
A few comments but overall nothing I think absolutely needs to be fixed.

src/server/pfs/server/driver.go
Tryneus: I also do not like this, but I was afraid of how many bugs I might introduce if I changed this.  I don't doubt that this has caused bugs by being the way it is either, though.

src/server/pfs/server/driver.go
```diff
@@ -1593,7 +1564,7 @@ func (d *driver) squashCommit(txnCtx *txnenv.TransactionContext, userCommit *pfs
 		// BFS upwards through graph for all non-deleted children
 		var next *pfs.Commit                            // next vertex to search
 		queue := []*pfs.Commit{lowestCommitInfo.Commit} // queue of vertices to explore
-		liveChildren := make(map[string]struct{})       // live children discovered so far
+		liveChildren := make(map[string]string)         // live children discovered so far
```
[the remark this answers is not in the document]
Tryneus: I believe multiple commits could have the same branch in `liveChildren` - do you see an issue with that?  I think what we actually want is for the key to be a `commitKey` and the value to be a `*pfs.Commit`, but at the moment that doesn't change anything until global IDs are in.

brycemcanally:
I think `src/internal/workload` is a dead package. Would you mind deleting it with this PR?

src/pfs/pfs.proto
```diff
@@ -58,15 +58,13 @@ message RepoAuthInfo {
 }
 
 message BranchInfo {
+  reserved 1;
```
brycemcanally, line 61: I don't think we need to reserve this with 2.0 since there is no expectation of compatibility. I kind of think that we would benefit from renumbering all of our protos for 2.0, then being stricter about the reserved numbers moving forward.

src/server/pfs/server/driver.go
```diff
@@ -424,28 +432,23 @@ func (d *driver) deleteRepo(txnCtx *txnenv.TransactionContext, repo *pfs.Repo, f
...
-//   commit, but 'branch' is still moved to point at the new commit
-// - If neither 'parent.ID' nor 'branch' are set, the new commit will have no
-//   parent
-// - If only 'parent.ID' is set, and it contains a branch, then the new commit's
-//   parent will be the HEAD of that branch, but the branch will not be moved
+// - 'parent' may be omitted, in which case the parent commit is inferred
+//   from 'branch'.
+// - If 'parent' is set, it determines the parent commit, but 'branch' is
+//   still moved to point at the new commit
 // TODO: Remove the v1 storage data structures from this function, they are not
```
brycemcanally, line 445: This can be deleted.

brycemcanally (approved):
Just a few minor changes for things not really related to the core change, other than that lgtm. I do have a few followup questions / thoughts:
- I agree with the change to use protos for the parameters of the Go client wrapper.
- For global IDs, it is not clear to me how we would go from a commit ID to the corresponding commits in the other repos unless the branch names were also a part of the global ID. Would you be able to access a commit in a repo without a branch parameter and potentially get back more than one? It seems like you would have to figure out the right branch otherwise.
- I am planning on adding a tag parameter for the file operations client side (it is just a field in the protos for the file operations as of now in the branch I am working with). It might make sense for that to be in the file proto somewhere. I guess this should be added to the string encoding in pachctl, but it seems like it is getting a bit unwieldy.

Tryneus:
> I agree with the change to use protos for the parameters of the Go client wrapper.
This is a separate change that @PFedak is working on - obviously we want to get this in first to avoid merge conflicts.

> For global IDs, it is not clear to me how we would go from a commit ID to the corresponding commits in the other repos unless the branch names were also a part of the global ID. Would you be able to access a commit in a repo without a branch parameter and potentially get back more than one? It seems like you would have to figure out the right branch otherwise.
You can access a commit without a branch parameter, but it is an error if there is an ambiguity.  For programmatic access, you should generally specify the branch if you know it.  The Global ID Job structure tracks each affected branch.

> I am planning on adding a tag parameter for the file operations client side (it is just a field in the protos for the file operations as of now in the branch I am working with). It might make sense for that to be in the file proto somewhere. I guess this should be added to the string encoding in pachctl, but it seems like it is getting a bit unwieldy.
I don't have any opinions on the command-line format without knowing the use cases, but putting it in the file proto seems right.  Once we change the client interface to accept the protos instead of strings, it shouldn't have many collateral code changes.

src/pfs/pfs.proto
Tryneus: I'll leave that for a separate PR because we'd need to sweep all the proto files anyway.

PR #6067, opened 2021-05-05
Bump PPS buckets to 8

lukemarsden:
Draft PR to see how it performs in Testfaster now that we've added two new servers.

Also cache a newer commit of the pachyderm source code, to avoid having to re-download new dependencies and such.


msteffen:
Closing PR, as it's obsolete.

PR #6068, opened 2021-05-05
Reenable PPS & PFS tests

brycemcanally:
This PR updates / fixes several tests that were not running in master. It also adds more information about the tests that are not running (through TODOs and skip messages for the relevant tests). The tests addressed in this PR are split into three groups: updated  / fixed, not updated / fixed but should be before 2.0, and not updated / fixed but might / should be after 2.0. Tests that were removed are in a section that include the reason for each test's removal. Also, several minor changes are included that are addressed in the miscellaneous section.

**Updated / fixed:**

PPS:
- TestPipelineWithLargeFiles
- TestChunkSpec/size
- TestPipelineWithDatumTimeout
- TestUserWorkingDir
- TestDatumTries
- TestCopyOutToIn
- TestMalformedPipeline
- TestListDatum
- TestUpdatePipeline
- TestPipelineVersions

PFS:
- PutFileDirectoryTraversal
- BigListFile
- MonkeyObjectStorage
- PutFileAtomic

Auth:
- TestS3GatewayAuthRequests

**Not updated / fixed, but should be before 2.0:**

PPS:
- TestRunPipeline
- Stats commit tests.
- TestCronPipeline/RunCronOverwrite
- TestListDatumDuringJob
- TestTrigger

PFS:
- Tests that check file path / directory path conflicts.
- Tests that use the offset read functionality
- PutFileValidCharacters
- PFS sync tests.
- PutFileCommit
- LargeDeleteRepo
- TestPanicOnNilArgs

**Not updated / fixed to run, but should / might be after 2.0:**

PPS:
- Git input tests.
- Put file split tests.
- File history tests.
- TestExtractPipeline
- TestEditPipeline

**Removed:**

PPS:
- TestEgressFailure: This was testing whether an egress error fails a job. This does not really make sense with V2 since egress is a part of the core job retry loop. We will retry the egress until some other job failure constraint is met, such as a timeout.
- TestLazyPipelineCPPipes: This was testing functionality that attempted to detect whether a user copied a special file (namely pipes for lazy files) and to error out when that is the case. I think this should be handled by documenting how certain features that cannot be black boxed are implemented (such as lazy files based on named pipes), rather than trying to catch all of the ways a user may be misusing them.
- TestGetFileWithEmptyCommits: Empty commits only serve the purpose of indicating that a job failed, they do not get interpreted as throwing away the parent commit content. When we have general purpose metadata associated with commits, then the empty commit idea should probably be discarded.
- TestMissingPipelineSpec: This test is checking if basic PPS functionality still works if the pipeline spec object is deleted. I think this type of error is critical enough that we should not try and come up with ways to sort of work when it happens.

PFS:
- TestCleanPath: This test checks that the relative elements of a path are removed and that the path would go through successfully. We are now erroring when a path with a relative element is used because we want to be explicit about this not being supported.
- PutFileLeak: This was testing for a goroutine leak in the V1 implementation of put file. This is no longer relevant in V2.

**Miscellaneous:**
- Implemented functionality for listing datums based on a passed in set of inputs.
- Removed a lot of dead code and V1 configuration that is no longer relevant in V2.
- Added validation for incoming PFS paths (prevents relative paths).
- Implements datumset splitting based on size.
- Added erroring when a file is not found.
- Added functionality for setting the ownership of files downloaded by PPS based on the configured user.


src/server/worker/datum/datum.go
```diff
@@ -64,7 +60,39 @@ func CreateSets(dit Iterator, storageRoot string, setSpec *SetSpec, upload func(
...
+				size += int64(input.FileInfo.SizeBytes)
+			}
+			if size >= setSpec.SizeBytes {
+				size = 0
+				return true
+			}
+			return false
+		}
+	default:
+		panic("must set a set spec field")
```
jdoliner, line 94: Do we really want to panic here? It seems like that will just put the worker in a crashloop that won't give any feedback to the user about why the pipelines isn't proceeding.

brycemcanally: This shouldn't ever happen because we default to evenly distributing the datums in the registry. I'm going to change this to a default of 1 when it is not set just in case. That default will apply at the datum package level, and the default of distributing the datums evenly will still apply at the transform package level.

jdoliner (approved):
All LGTM

PR #6070, opened 2021-05-05
Add initial support for system repos.

PFedak:
System repos represent a collection of related repos under a shared
name, differentiated by various types labelling their purpose. The main
user friendly client functions will only produce the default "user" repo
type to keep basic workflows simple. Internally, repo objects are
expected to have the type field filled in, and this is enforced by a key
check when accessing the repo collection.


src/pfs/pfs.go
```diff
@@ -8,11 +8,17 @@ import (
...
 )
 
-var (
+const (
 	// ChunkSize is the size of file chunks when resumable upload is used
 	ChunkSize = int64(512 * 1024 * 1024) // 512 MB
 	// EmptyStr is included in the description of output commits from failed jobs.
 	EmptyStr = "(empty)"
+
+	// default system repo types
```
Tryneus, line 17: Given the PFS-wide nature of these consts, I think the names should indicate that this is specifically a repo type, so maybe `UserRepoType`, etc?

src/pfs/pfs.proto
```diff
@@ -181,6 +182,8 @@ message InspectRepoRequest {
 
 message ListRepoRequest {
   reserved 1;
+  string type = 2;
+  bool all = 3;
```
Tryneus, line 186: I think we can exclude `all` if we interpret an empty `type` field as 'no filter'.

src/server/pfs/server/driver.go
```diff
@@ -217,6 +222,17 @@ func (d *driver) createRepo(txnCtx *txnenv.TransactionContext, repo *pfs.Repo, d
 		existingRepoInfo.Description = description
 		return repos.Put(repo.Name, &existingRepoInfo)
 	} else {
+		// if this is a system repo, make sure the corresponding user repo already exists
+		if repo.Type != pfs.UserType {
+			baseRepo := client.NewRepo(repo.Name)
+			err = repos.Get(pfsdb.RepoKey(baseRepo), &existingRepoInfo)
+			if err != nil && col.IsErrNotFound(err) {
```
Tryneus, line 229: I think these if blocks are backwards.

Tryneus (commented):
LGTM with a couple nits - I guess the only obstacle is having `GetByIndex` support in read-write collections?

src/pfs/pfs.go
PFedak: Ah sorry, meant to do that before I sent it

src/pfs/pfs.proto
PFedak: I think I was debating if it made the signature of ListReposByType confusing, but I suppose most people will just want ListRepos anyway.

src/server/pfs/server/driver.go
PFedak: You're absolutely right, whoops

src/client/pfs.go
```diff
@@ -88,12 +93,17 @@ func (c APIClient) InspectRepo(repoName string) (*pfs.RepoInfo, error) {
...
 }
 
-// ListRepo returns info about all Repos.
-// provenance specifies a set of provenance repos, only repos which have ALL of
-// the specified repos as provenance will be returned unless provenance is nil
-// in which case it is ignored.
+// ListRepo returns info about standard Repos
 func (c APIClient) ListRepo() ([]*pfs.RepoInfo, error) {
-	request := &pfs.ListRepoRequest{}
+	return c.ListRepoByType("", false)
```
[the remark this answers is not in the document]
PFedak: I had flipped where the default got set and forgot this, thanks

src/internal/collection/types.go
```diff
@@ -89,6 +89,9 @@ type PostgresReadWriteCollection interface {
 	ReadWriteCollection
 
 	DeleteByIndex(index *Index, indexVal string) error
+	// GetByIndex can have a large impact on database contention if used to retrieve
+	// a large number of rows. Consider using a read-only collection if possible
+	GetByIndex(index *Index, indexVal string, val proto.Message, opts *Options, f func(string) error) error
```
PFedak, line 94: I'm also open to renaming this `ExpensiveGetByIndex` or something just to make it harder to blindly switch away from read only

Tryneus: I was leaning more towards renaming all uses to `ListByIndex`, but that's kind of annoying to do, so I might take care of it myself after this merges.

PR #6080, opened 2021-05-07
[1.13] Try just using a context with a timeout

actgardner:


actgardner:
Confirmed this doesn't resolve the issue, requests time out with a different error but the connection remains in a bad state for 30+ minutes.

PR #6085, opened 2021-05-10
Add role for k8s secrets management

actgardner:
In 1.x we let any authenticated user create, modify, list and delete k8s secrets that was created by Pachyderm. In 2.x make these specific permissions with a new role (`secretAdmin`) so customers can control who can manage secrets.

@nadegepepin  - Sorry this is a new role
@seslattery - We'll want to give this role to hub users as well


acohen4 (approved):
LGTM!

PR #6092, opened 2021-05-11
Deprecate vault

chainlink:
As per 1.12.5 release notes, vault support has been marked deprecated


actgardner:
FWIW this is a different use of Vault than the one I deprecated:
- we used to support a Vault plugin for getting pachd auth tokens on behalf of users, which is deprecated now
- this is about getting S3 credentials from Vault, which still seems like a valid use case?

actgardner (approved):
🔥

PR #6093, opened 2021-05-11
Update default image to ubuntu 20.04 and golang to 1.16.4

nitinjainsj:


chainlink:
Also need https://github.com/pachyderm/pachyderm/blob/master/etc/proto/Dockerfile (Looks like it got missed in the last round..)

nitinjainsj:
I had to revert the change because the protos don't build with newer version. There is most likely a fix for it, but it should be dealt with in another PR. I think that's also the reason why it was skipped last time around.

etc/testing/pipeline-build/go/go.mod
```diff
@@ -1,5 +1,5 @@
 module github.com/pachyderm/pachyderm/etc/testing/pipeline-build/go
 
-go 1.13
+go 1.16
```
brendoncarroll, line 3: We haven't updated this in the past, but I don't see any reason not to.

etc/testing/spout/go.mod
```diff
@@ -0,0 +1,3 @@
+module etc/testing/spout/main
```
brendoncarroll, line 1: `github.com/pachyderm/pachyderm/v2/etc/testing/spout/main` ?

I'm not sure what the rules are for nested modules; they are definitely allowed, but I'm unsure of the semantics.

src/server/auth/server/testing/admin_test.go
```diff
@@ -458,7 +458,7 @@ func TestPreActivationPipelinesKeepRunningAfterActivation(t *testing.T) {
 	require.NoError(t, aliceClient.CreateRepo(repo))
 	require.NoError(t, aliceClient.CreatePipeline(
 		pipeline,
-		"", // default image: ubuntu:14.04
+		"", // default image: DefaultUserImage
```
brendoncarroll, line 461: 👍

src/testing/saml-idp/Dockerfile.build
```diff
@@ -11,7 +11,7 @@
 # image (Dockerfile.idp) because we need to mount ../vendor into the container
 # as part of the build process, and there's no way to mount an external
 # directory into a docker container as part of 'docker build'
-FROM golang:1.10.2-alpine AS build
+FROM golang:1.16.4-alpine AS build
```
brendoncarroll, line 14: 1.10 seems almost intentionally old.  Are we trying to ensure backwards compatibility with this test?

brendoncarroll (approved):
Some of the images we test with might be old to ensure backwards compatibility, but if we only care about the latest Go version it LGTM.

etc/testing/pipeline-build/go/go.mod
nitinjainsj: I know less about go.mod than you. But reading about it this sets the min version if nothing else has been specified

etc/testing/spout/go.mod
nitinjainsj: I am not entirely sure what you mean here. This was generated by go mod init. Where does v2 come in?

src/testing/saml-idp/Dockerfile.build
nitinjainsj: @actgardner Is this intentional?

etc/testing/pipeline-build/go/go.mod
brendoncarroll: I just meant in the build pipeline.

etc/testing/spout/go.mod
brendoncarroll: That is the path for the root module.  If this works, it's probably not worth changing.  It would only really matter if you were trying to import it from elsewhere.

src/testing/saml-idp/Dockerfile.build
nitinjainsj: @actgardner Can we delete the whole test/saml-idp directory?

actgardner: I believe so, yes.

PR #6096, opened 2021-05-11
[WIP] Revoke User Tokens in all Pachd clusters registered with an enterprise server

acohen4:
This is a work in progress. Even if the stuff here seems to be in the right direction, I still have to figure out the best way to test this.

Below are assumptions and trade-offs that have come out of my brain storming. Hopefully these can serve as a conversation starter.

**Assumptions**

- Partial Commits ok, as long as we surface partial results for users to retry
- Username (email) is a user's "global identifier" 
- Revoke calls are effectively idempotent - repeated calls expect the same end state
- 3 Callout outcomes to note
	- Success
	- Failure w/ message
	- Timeout

**Server-side vs Client-side Fanout Tradeoffs**

Benefits of Server-side Fanout
- Permission only needs to be checked on the enterprise server. The permission doesn't have to be duplicated across all clusters
- A nice place to define such a pattern
	- In the future we may want to support other fan out operations that don't make sense in client
- Coordination between Pachds is the enterprise server's raison d'etre
- Performance?
	- In theory the enterprise server may live close by to pachds so the network calls shouldn't vary as much depending on the executing user's locale/network

Benefits of Client-side Fanout 

- One less network hop


src/server/auth/cmds/cmds.go
```diff
@@ -154,24 +154,40 @@ Activate Pachyderm's auth system, and restrict access to existing data to the ro
...
 				}
+				// when activating auth for the enterprise server, also set a Configuration for it to redirect to a privileged oauth flow
+				if _, err := c.SetConfiguration(c.Ctx(),
+					&auth.SetConfigurationRequest{
+						LoginType: auth.OIDCLoginType_PRIVILEDGED,
+						Configuration: &auth.OIDCConfig{
+							Issuer:          issuer,
+							ClientID:        clientId,
+							ClientSecret:    oidcClient.Client.Secret,
+							RedirectURI:     redirectPrivileged, // who configures the privileged redirect? Extra param? Should we instead just get a privileged Domain?
```
actgardner, line 184: What's the goal of the privileged redirect versus the normal one? We have all the state server-side, do we need two different URIs?

src/auth/auth.proto
```diff
@@ -437,6 +454,8 @@ service API {
 
   rpc GetConfiguration(GetConfigurationRequest) returns (GetConfigurationResponse) {}
   rpc SetConfiguration(SetConfigurationRequest) returns (SetConfigurationResponse) {}
+  rpc AddScopesToConfiguration(AddScopesToConfigurationRequest) returns (AddScopesToConfigurationResponse) {}
+  rpc RemoveScopesFromConfiguration(RemoveScopesFromConfigurationRequest) returns (RemoveScopesFromConfigurationResponse) {}
```
actgardner, line 458: These RPCs seem to overlap with the functionality in `GetConfiguration`?

src/server/auth/cmds/cmds.go
acohen4: I see what you mean. I guess we could use the state to achieve this. 

My thought was that we didn't have a way to parameterize the HTTP call to Callback to pass the information of whether to mint the privileged ID Token. I see that the state could be that parameterization

src/auth/auth.proto
acohen4: Yeah they use similar implementation logic to GetConfiguration and SetConfiguration.

Does having multiple ways to modify Configuration values make for an unclear interface? I could also make these helper functions in the License module

src/server/license/server/api_server.go
```diff
@@ -377,3 +379,140 @@ func (a *apiServer) ListUserClusters(ctx context.Context, req *lc.ListUserCluste
...
+	}
+	return &fanoutReport{
+		successes: successes,
+		failures:  failures,
+		timeouts:  timeouts,
+	}
+}
+
+// is there an already existing pattern to use instead of these functions?
+func submitToStrChan(done <-chan struct{}, failures chan<- string, s string) {
```
acohen4, line 498: is there some standard mechanism that does this?

src/auth/auth.proto
```diff
@@ -152,10 +156,11 @@ enum Permission {
...
 
-  CLUSTER_ENTERPRISE_ACTIVATE            = 114;
-  CLUSTER_ENTERPRISE_HEARTBEAT           = 115;
-  CLUSTER_ENTERPRISE_GET_CODE            = 116;
-  CLUSTER_ENTERPRISE_DEACTIVATE          = 117;
+  CLUSTER_ENTERPRISE_ACTIVATE                        = 114;
+  CLUSTER_ENTERPRISE_HEARTBEAT                       = 115;
+  CLUSTER_ENTERPRISE_GET_CODE                        = 116;
+  CLUSTER_ENTERPRISE_DEACTIVATE                      = 117;
+  CLUSTER_ENTERPRISE_REVOKE_USER_TOKENS_ALL_CLUSTERS = 148;
```
actgardner, line 163: I usually structure these permissions as `SCOPE_SERVICE_PERMISSION` so I wonder if these ones (that are evaluated at the enterprise level) should look like `ENTERPRISE_ LICENSE_REVOKE_TOKENS`?

src/auth/auth.proto
```diff
@@ -328,11 +333,17 @@ message SessionInfo {
   // information to a user who has network access to Pachyderm but not an
   // account in the OIDC provider.
   bool conversion_err = 3;
+  // rawIdToken is the ID token returned by the OAuth2 Flow when exchanged with the user's auth_code
+  string rawIdToken = 4;
```
actgardner, line 337: nit: for consistency can we use snake-case (`raw_id_token`)?

src/auth/auth.proto
```diff
@@ -328,11 +333,17 @@ message SessionInfo {
...
 }
 
 //// OIDC API
-
+enum OIDCLoginType {
+  STANDARD = 0;
+  PRIVILEDGED = 1;
+}
 message GetOIDCLoginRequest {
+  OIDCLoginType login_type = 1;
```
actgardner, line 346: +1, I think this is a good API for this.

src/auth/auth.proto
acohen4: whoops absolutely

src/server/auth/server/oidc.go
```diff
@@ -138,12 +138,24 @@ func (a *apiServer) getOIDCConfig(ctx context.Context) (*oidcConfig, error) {
...
+	if loginType == auth.OIDCLoginType_PRIVILEDGED {
+		// Add all Registered Clients as Audience of the resulting ID Token
+		// - query all CLIENT IDs from license.clusters table and add them to the config before generating the AuthCodeURL
+
+		// sort of what we'll do:
+		config.oauthConfig.Scopes = append(config.oauthConfig.Scopes, auth.TrustedPeerAudienceScopePrefix+"clientID")
+
+		// Make sure ALL Clients have Enterprise-Server Client as Trusted Peer
+		// - List Clients
+		// - Update each Client that doesn't have ES-Client as Trusted Peer
```
actgardner, line 157: 👍

src/server/license/server/api_server.go
```diff
@@ -390,3 +392,140 @@ func (a *apiServer) ListUserClusters(ctx context.Context, req *lc.ListUserCluste
...
+	failureChan := make(chan *fanoutError)
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	for _, cluster := range clusters {
+		waitingOnClusters[cluster.Id] = true
+
+		// kickoff goroutine to make request to a pachd cluster
+		go func(ctx context.Context, address string, id string) {
```
actgardner, line 455: I think this could be simplified with an error group (https://pkg.go.dev/golang.org/x/sync/errgroup). Since we propagate the context thorugh all the RPCs we could set a timeout on the context and then we don't have to worry about enforcing those timeouts ourselves either.

PR #6098, opened 2021-05-12
Remove the requirement for bucket details when deploying the enterprise server

actgardner:
In the hackathon when deploying the enterprise server to GKE we noticed it was impossible not to supply a bucket name, even though the enterprise server doesn't use a GCS bucket. This PR re-orders the arguments so GCS/ABS/S3 are the last arguments and can be skipped if the enterprise server flag is set.

/cc @nadegepepin - this will be relevant for the enterprise server docs, but we also need to update all the deploy docs since the arguments change order.

### DIFF
--- a/src/internal/deploy/cmds/cmds.go
+++ b/src/internal/deploy/cmds/cmds.go
@@ -573,14 +573,14 @@ func standardDeployCmds() []*cobra.Command {
 	commands = append(commands, cmdutil.CreateAlias(deployLocal, "deploy local"))
 
 	deployGoogle := &cobra.Command{
-		Use:   "{{alias}} <bucket-name> <disk-size> [<credentials-file>]",
+		Use:   "{{alias}} <disk-size> [<bucket-name>] [<credentials-file>]",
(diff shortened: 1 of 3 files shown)


src/internal/deploy/cmds/cmds.go
```diff
@@ -831,21 +850,33 @@ If <object store backend> is \"s3\", then the arguments are:
 	commands = append(commands, cmdutil.CreateAlias(deployAmazon, "deploy aws"))
 
 	deployMicrosoft := &cobra.Command{
-		Use:   "{{alias}} <container> <account-name> <account-key> <disk-size>",
+		Use:   "{{alias}} <disk-size> <container> <account-name> <account-key>",
```
msteffen, line 853: Nit: should `container`, `account-name` and `account-key` be in square braces (just based on the aws/google help text)?

msteffen (approved):
LGTM, with one minor nit

actgardner:
NB: the golden manifest tests are currently disabled - I ran them locally and these changes are fine, but our manifests have drifted in production quite a bit. We should re-enable them prior to 2.0.

PR #6099, opened 2021-05-12
Use correct context for enterprise calls in auth.Activate

actgardner:
In the hackathon `auth.Activate` failed because we were using the pachd auth token against the enterprise server. This fixes the code and updates the tests to use a unique auth token - the tests didn't catch this bug because we used the same token for both servers.


acohen4 (approved):
LGTM!

PR #6103, opened 2021-05-12
Explicitly log when enterprise context is used

actgardner:
In the hackathon we've had confusion about what context the `pachctl idp` and `pachctl license` run against. Print the context name when running commands against the enterprise server to make it clear to end users.


acohen4 (approved):
LGTM!

PR #6104, opened 2021-05-12
Remove update dash command (old dash)

chainlink:
This should be fairly non controversial, as old dash has been code locked for a while (and new dash will use helm)


msteffen (approved):
LGTM

PR #6105, opened 2021-05-12
Compact at read time

brendoncarroll:
- Removes compaction from `finishCommit`.
- Adds compaction to `getFileset` if the total fileset for the commit does not exist


src/server/pfs/server/driver.go
```diff
@@ -1306,6 +1279,37 @@ func (d *driver) resolveCommit(stm col.STM, userCommit *pfs.Commit) (*pfs.Commit
 	return commitInfo, nil
 }
 
+// getCommit is like inspectCommit, without the blocking.
+// It does not the size to the CommitInfo
```
brycemcanally, line 1283: comment wording

src/server/pfs/server/driver_file.go
```diff
@@ -450,3 +447,49 @@ func (d *driver) getFileset(pachClient *client.APIClient, commit *pfs.Commit) (*
...
+	ctx := pachClient.Ctx()
+	commitInfo, err := d.getCommit(pachClient, commit)
+	if err != nil {
+		return nil, err
+	}
+	if commitInfo.Finished == nil {
+		return nil, errors.Errorf("attempted to compute total of unfinished commit")
+	}
+	commit = commitInfo.Commit
+	id, err := d.commitStore.GetTotalFileset(ctx, commit)
```
brycemcanally, line 461: I would prefer this to return an error that indicates it doesn't exist, then we check for that to determine if we need to compute it.

src/server/pfs/server/driver_file.go
```diff
@@ -450,3 +447,49 @@ func (d *driver) getFileset(pachClient *client.APIClient, commit *pfs.Commit) (*
...
+		return nil, err
+	}
+	if id != nil {
+		return id, nil
+	}
+	id, err = d.commitStore.GetDiffFileset(ctx, commit)
+	if err != nil {
+		return nil, err
+	}
+	ids, err := d.storage.Flatten(ctx, []fileset.ID{*id})
```
brycemcanally, line 472: Why do we need to flatten here? That should be handled by the compaction.

src/server/pfs/server/driver_file.go
```diff
@@ -450,3 +447,49 @@ func (d *driver) getFileset(pachClient *client.APIClient, commit *pfs.Commit) (*
...
+	if err != nil {
+		return nil, err
+	}
+	var inputs []fileset.ID
+	if commitInfo.ParentCommit != nil {
```
brycemcanally, line 477: You don't need the if / else, just append the ids after the if.

src/server/pfs/server/commit_store.go
```diff
@@ -192,6 +163,39 @@ func getTotal(tx *sqlx.Tx, commit *pfs.Commit) (*fileset.ID, error) {
...
+		if err == sql.ErrNoRows {
+			return nil
+		}
+		return err
+	}
+	trackID := commitTotalTrackerID(commit, *id)
+	if err := tr.DeleteTx(tx, trackID); err != nil {
+		return err
+	}
+	if _, err := tx.Exec(`DELETE FROM pfs.commit_totals WHERE commit_id = $1`, commit.ID); err != nil {
```
brycemcanally, line 178: No if needed, just return err.

src/server/pfs/server/compactor.go
```diff
@@ -68,11 +96,9 @@ func (d *driver) compact(ctx context.Context, ids []fileset.ID) (*fileset.ID, er
 	})
 }
 
-func (d *driver) compactionWorker() {
-	ctx := context.Background()
-	w := work.NewWorker(d.etcdClient, d.prefix, storageTaskNamespace)
+func (c *compactor) compactionWorker(ctx context.Context) {
```
brycemcanally, line 99: Maybe this should be internal for the compactor, instead of having to run it in a goroutine outside of this abstraction?

brendoncarroll: At some point we should develop a pattern around supervising long running processes.  But for now everything is spawned in either the constructor or the master.

src/server/pfs/server/driver_file.go
brendoncarroll: I was thinking about that, but returning the not exist error from the fileset package is ambiguous, so we would need an error type specifically for this case.  And it would just complicate the error check every time it is used.  No fileset is also a valid result, as opposed to in the fileset package where asking for a fileset that doesn't exist is always an error.

brycemcanally: I don't think it is really that ambiguous. If the total fileset doesn't exist, then it doesn't exist. You need a check in both cases, either for a certain type of error or whether the ID is nil. The certain type of error approach is going to be much less prone to bugs, than returning a nil ID and nil error.

src/server/pfs/server/compactor.go
brycemcanally: I think this should just be a goroutine setup in the compactor constructor, and also pipe the env context through and use it for the task queue.

brycemcanally (approved):
lgtm

PR #6107, opened 2021-05-13
[1.13.x]Local loki

chainlink:
Copy stripped down version of loki client into pach to remove loki client dependency. Unblocks upgrading kube client


src/server/pkg/lokiutil/lokiutil.go
```diff
@@ -70,16 +65,10 @@ func QueryRange(ctx context.Context, c *loki.Client, queryStr string, from, thro
 			return nil
 		}
 		from = from.Add(time.Nanosecond)
-		// check if the context has been cancelled
```
chainlink, line 73: I think this is no longer required with being able to pass the context through to the request?

PR #6108, opened 2021-05-13
Update OIDC commands to accept YAML configs

actgardner:
The `pachctl idp create-connector`, `pachctl idp update-connector` and `pachctl idp get-connector` commands all operate on a YAML config document. However, the commands to work with OIDC clients and set the issuer in Dex weren't updated to act the same.

This makes all the `pachctl idp` commands work the same - they accept YAML from a file or stdin and print YAML. This makes it easy to grab the config and tweak it, back it up, etc.

@nadegepepin I think we had only documented `pachctl idp set-config` for setting the issuer? LMK if you need more context to update the docs for this.

### DIFF
--- a/src/server/identity/cmds/cmds_test.go
+++ b/src/server/identity/cmds/cmds_test.go
@@ -44,28 +44,24 @@ func TestClientCRUD(t *testing.T) {
 		t.Skip("Skipping integration tests in short mode")
 	}
 	tu.ActivateAuth(t)
-	defer tu.DeleteAll(t)
+	//defer tu.DeleteAll(t)
 	require.NoError(t, tu.BashCmd(`
-		pachctl idp create-client --id {{.id}} --name 'testclient' --secret 'a secret' --redirectUris https://localhost:1234 \
+		echo '{"id": "{{.id}}", "name": "testclient", "secret": "a secret", "redirect_uris": ["https://localhost:1234"]}' | pachctl idp create-client  \
 		  | match 'secret: "a secret"'
 		pachctl idp list-client | match '{{.id}}'
 		pachctl idp get-client {{.id}} \
 		  | match 'name: testclient' \
 		  | match 'secret: a secret' \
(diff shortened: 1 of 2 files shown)


src/server/identity/cmds/cmds_test.go
```diff
@@ -44,28 +44,24 @@ func TestClientCRUD(t *testing.T) {
 		t.Skip("Skipping integration tests in short mode")
 	}
 	tu.ActivateAuth(t)
-	defer tu.DeleteAll(t)
+	//defer tu.DeleteAll(t)
```
acohen4, line 47: did you mean to delete this line?

src/server/identity/cmds/cmds_test.go
```diff
@@ -79,8 +75,8 @@ func TestGetSetConfig(t *testing.T) {
 	tu.ActivateAuth(t)
 	defer tu.DeleteAll(t)
 	require.NoError(t, tu.BashCmd(`
-		pachctl idp set-config --issuer 'http://example.com:1234'
-		pachctl idp get-config | match 'issuer: "http://example.com:1234"' 
+		echo '{"issuer": "http://example.com:1234"}' | pachctl idp set-config
```
acohen4, line 78: does `deserializeYAML()` also parse JSON? I'm confused by what code path is deserializing this json.

src/server/identity/cmds/cmds_test.go
actgardner: oof, good catch

src/server/identity/cmds/cmds_test.go
actgardner: JSON is a strict subset of YAML, so if you support YAML you also suppot JSON. I used JSON in the tests because the whitespace is hard to format correctly.

acohen4: wow I can't believe I didn't know that! that's cool

acohen4 (approved):
LGTM

PR #6111, opened 2021-05-13
Disable enterprise heartbeat in sidecars

actgardner:
Disable the enterprise heartbeat goroutine in sidecars and the internal pachd services. We don't do any kind of master election ATM so multiple clustered pachds will all heartbeat redundantly, but this way they only heartbeat once each, and their workers don't heartbeat at all.

Also removes the `enterprise license heartbeat process failed` log message when a pachd isn't registered with an enterprise server, which was confusing to users in the hackathon.


acohen4 (approved):
LGTM!

PR #6113, opened 2021-05-13
Update Go for proto build image

actgardner:
In more recent Go versions `go get` installs modules to `$GOPATH/pkg/mod/` with the version,  instead of `$GOPATH/src`. Update the include flag for protoc to point to the new path.


chainlink:
Sweet! If this runs successfully I see no reason not to do it. 

Gogo lives another day 😅

chainlink (approved):

PR #6115, opened 2021-05-13
add gke optimized json logging as an option

seslattery:
This adds an additional json formatter to logrus that works well with stackdriver.  Enabled by setting env var LOG_FORMAT = JSON.


chainlink:
We might also want to get this on the 1.13.x train

seslattery:
> We might also want to get this on the 1.13.x train

Yeah I agree with that, I'd for this to hit Hub sooner rather than later

msteffen:
Closing this PR as obsolete

PR #6117, opened 2021-05-14
Make CI PPS test bucketing regex more specific

Tryneus:
The current regex does not account for test names which are substrings of other test names, so some tests may end up running multiple times in different buckets.  Fix this by anchoring the test name regexes with `^testname$`.


actgardner (approved):
Good call 👍

PR #6118, opened 2021-05-14
Enterprise heartbeat includes oauth client_id

actgardner:
On it's face this is a pretty simple change: we have a model in the license server for each registered pachd, and we want to tie it to the model in dex of the OIDC clients, so we have each pachd report it's oauth client_id (the unique identifier in dex) as part of the heartbeat message.

The "how" is likely to provoke some discusson: adding support for `auth.GetConfiguration` to `transactionenv.AuthTransactionServer`, so we can bypass the auth interceptor and call a privileged API in a goroutine where there are no user credentials. This problem also comes up in cluster metrics and the PPS master - we work around it in the PPS master by having a "superuser" token, and we don't currently work around it for cluster metrics.

In discussions with the team, there seems to be a consensus that making RPCs on the loopback interface makes our codebase harder to understand and adds a performance penalty. So it seems like we need some place in pachd which can hold a reference to the various services (auth, pfs, pps), so they can be called by other services. `TransactionEnv` provides this, but it has the historical baggage that we have only ever used it to propagate database transactions. 

My proposal here is: let's just put all the internal APIs in `TransactionEnv`, instead of trying to bifurcate this API into transaction and non-transaction, and we can iterate on naming in the future? This will allow us to make progress towards eliminating `GetPachClient` from `ServiceEnv` and no longer doing loopback RPCs.


acohen4 (approved):
Looks good!

PR #6128, opened 2021-05-14
[Proposal] Split out auth interface from txnenv

actgardner:


src/internal/serviceenv/service_env.go
```diff
@@ -32,6 +33,8 @@ const clusterIDKey = "cluster-id"
 // create more, if they want to create multiple pachyderm "clusters" served in
 // separate goroutines.
 type ServiceEnv interface {
+	AuthServer() auth_server.APIServer
```
actgardner, line 36: Add accessors for services to ServiceEnv - this replaces the cross-service references in TransactionEnv.

src/internal/transactionenv/env.go
```diff
@@ -39,16 +38,6 @@ type PpsWrites interface {
...
-// AuthWrites is an interface providing a wrapper for each operation that
-// may be appended to a transaction through the Auth server.  Each call may
-// either directly run the request through Auth or append it to the active
-// transaction, depending on if there is an active transaction in the client
-// context.
-type AuthWrites interface {
-	ModifyRoleBinding(*auth.ModifyRoleBindingRequest) (*auth.ModifyRoleBindingResponse, error)
-	DeleteRoleBinding(*auth.Resource) error
-}
-
```
actgardner, line 51: This wasn't used AFAICT.

src/internal/transactionenv/env.go
```diff
@@ -84,13 +73,6 @@ type TransactionContext struct {
 	txnEnv         *TransactionEnv
 }
 
-// Auth returns a reference to the Auth API Server so that transactionally-
-// supported methods can be called across the API boundary without using RPCs
-// (which will not maintain transactional guarantees)
-func (t *TransactionContext) Auth() AuthTransactionServer {
-	return t.txnEnv.authServer
-}
-
```
actgardner, line 93: Moved this to serviceenv since the interface contains transaction + non-transaction methods.

src/internal/transactionenv/env.go
```diff
@@ -136,29 +118,6 @@ type TransactionServer interface {
...
-	RemovePipelineReaderFromRepoInTransaction(*TransactionContext, string, string) error
-
-	// Create and Delete are internal-only APIs used by other services when creating/destroying resources.
-	CreateRoleBindingInTransaction(*TransactionContext, string, []string, *auth.Resource) error
-	DeleteRoleBindingInTransaction(*TransactionContext, *auth.Resource) error
-
-	// GetPipelineAuthTokenInTransaction is an internal API used by PPS to generate tokens for pipelines
-	GetPipelineAuthTokenInTransaction(*TransactionContext, string) (string, error)
-	RevokeAuthTokenInTransaction(*TransactionContext, *auth.RevokeAuthTokenRequest) (*auth.RevokeAuthTokenResponse, error)
-}
```
actgardner, line 160: Moved this to `src/server/auth` as part of the definition of `APIServer`.

src/internal/transactionenv/env.go
```diff
@@ -191,24 +150,18 @@ type PpsTransactionServer interface {
...
-	pfsServer  PfsTransactionServer
-	ppsServer  PpsTransactionServer
+	txnServer TransactionServer
+	pfsServer PfsTransactionServer
+	ppsServer PpsTransactionServer
 }
 
 // Initialize stores the references to APIServer instances in the TransactionEnv
 func (env *TransactionEnv) Initialize(
-	serviceEnv serviceenv.ServiceEnv,
```
actgardner, line 203: Removing the serviceenv here is necessary to resolve a circular dependency, instead we pass in the etcd client (soon to be postgres client) for each new transaction being created.

src/internal/transactionenv/env.go
```diff
@@ -281,21 +233,11 @@ func (t *directTransaction) UpdateJobState(original *pps.UpdateJobStateRequest)
 	return t.txnCtx.txnEnv.ppsServer.UpdateJobStateInTransaction(t.txnCtx, req)
 }
 
-func (t *directTransaction) ModifyRoleBinding(original *auth.ModifyRoleBindingRequest) (*auth.ModifyRoleBindingResponse, error) {
```
actgardner, line 284: These weren't commonly used and you can just call the InTransaction methods directly.

src/internal/transactionenv/env.go
```diff
@@ -358,21 +300,13 @@ func (t *appendTransaction) CreatePipeline(req *pps.CreatePipelineRequest, _ **p
...
-}
-
 // WithTransaction will call the given callback with a txnenv.Transaction
 // object, which is instantiated differently based on if an active
 // transaction is present in the RPC context.  If an active transaction is
 // present, any calls into the Transaction are first dry-run then appended
 // to the transaction.  If there is no active transaction, the request will be
 // run directly through the selected server.
-func (env *TransactionEnv) WithTransaction(ctx context.Context, cb func(Transaction) error) error {
+func (env *TransactionEnv) WithTransaction(ctx context.Context, etcdClient *etcd.Client, cb func(Transaction) error) error {
```
actgardner, line 309: Necessary change to decouple serviceenv and transactionenv.

src/server/auth/server/api_server.go
```diff
@@ -62,12 +63,6 @@ const (
 // DefaultOIDCConfig is the default config for the auth API server
 var DefaultOIDCConfig = auth.OIDCConfig{}
 
-// APIServer represents an auth api server
-type APIServer interface {
-	auth.APIServer
-	txnenv.AuthTransactionServer
-}
```
actgardner, line 69: This has to live in `src/server/auth` for circular reference reasons.

src/server/auth/server/api_server.go
```diff
@@ -139,7 +134,7 @@ func NewAuthServer(
 	public bool,
 	requireNoncriticalServers bool,
 	watchesEnabled bool,
-) (APIServer, error) {
+) (auth_server.APIServer, error) {
```
actgardner, line 137: I'm assuming we return an interface here because returning a private struct sucks. Could we just make the APIServer type public with all-private members? It's awkward to have `src/server/auth/server` reference `src/server/auth` like this.

src/server/auth/server/util.go
```diff
@@ -10,14 +10,15 @@ import (
 
 // CheckClusterIsAuthorizedInTransaction returns an error if the current user doesn't have
 // the permissions in `p` on the cluster
-func CheckClusterIsAuthorizedInTransaction(txnCtx *txnenv.TransactionContext, p ...auth.Permission) error {
-	me, err := txnCtx.Client.WhoAmI(txnCtx.ClientContext, &auth.WhoAmIRequest{})
+func (a *apiServer) CheckClusterIsAuthorizedInTransaction(txnCtx *txnenv.TransactionContext, p ...auth.Permission) error {
+	// TODO (actgardner): WhoAmIInTransaction
+	me, err := a.WhoAmI(txnCtx.ClientContext, &auth.WhoAmIRequest{})
```
actgardner, line 15: WhoAmI goes from an RPC to a method call, so it will benefit from the metadata being cached in the client context and avoid a postgres round-trip.

src/server/cmd/pachd/main.go
```diff
@@ -168,6 +167,7 @@ func doEnterpriseMode(config interface{}) (retErr error) {
 				return err
 			}
 			authclient.RegisterAPIServer(externalServer.Server, authAPIServer)
+			env.SetAuthServer(authAPIServer)
```
actgardner, line 170: I find it slightly tidier to give these servers to the serviceenv as we create them, instead of having one big initialize call at the end, but ymmv.

PR #6129, opened 2021-05-14
Update error message checks for new collections.

PFedak:
This was causing issues with the S3 tests, possibly others.

Also fix an error message copy/paste error and fix the branch key check
to work as intended.


PFedak:
I didn't go through and check if we could just check for the structured error that the collection returns, but that's probably a better fix

Tryneus:
LGTM - I think we'll revisit the keycheck later, but this is great for now.

PR #6130, opened 2021-05-14
Fix pq parsing error during nested queries in PostgresReadWriteCollection.GetByIndex

Tryneus:
In a transaction, we can't start the next query until the previous one has completed.  This changes `GetByIndex` to greedily read out the full result set on read-write collections before calling the user callback.  Also added `GetByIndex` tests for `PostgresReadWriteCollection`, including a test that reproduces this situation.  The symptom of this bug is an error that looks like `pq: unexpected Parse response 'C'`.

### DIFF
--- a/src/internal/collection/postgres_collection_test.go
+++ b/src/internal/collection/postgres_collection_test.go
@@ -8,6 +8,7 @@ import (
 
 	col "github.com/pachyderm/pachyderm/v2/src/internal/collection"
 	"github.com/pachyderm/pachyderm/v2/src/internal/dbutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/require"
 	"github.com/pachyderm/pachyderm/v2/src/internal/serviceenv"
 	"github.com/pachyderm/pachyderm/v2/src/internal/testutil"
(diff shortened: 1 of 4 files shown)


src/internal/collection/postgres_collection_test.go
```diff
@@ -64,6 +66,109 @@ func TestPostgresCollections(suite *testing.T) {
...
+			t.Parallel()
+			outerKeys := []string{}
+			innerKeys := []string{}
+			innerID := makeID(3)
+			err := defaultWriter(context.Background(), func(rw col.ReadWriteCollection) error {
+				testProto := &col.TestItem{}
+				pgrw := rw.(col.PostgresReadWriteCollection)
+				return pgrw.GetByIndex(TestSecondaryIndex, originalValue, testProto, col.DefaultOptions(), func(key string) error {
+					outerKeys = append(outerKeys, testProto.ID)
+					if err := pgrw.Get(innerID, testProto); err != nil {
```
PFedak, line 149: Maybe it's worth doing a write here, just to enforce what that behavior means.
Probably doesn't matter though

PFedak (approved):
Wow, that error is unimaginably obtuse. I guess I should have though about what happens if you try to modify the collection in the GetByIndex.

LGTM

PR #6134, opened 2021-05-17
2.0: Remove reserved protobuf fields, collapse unused field numbers

Tryneus:
Mentioned by @brycemcanally in the review for #6066 - since 2.x is breaking compatibility, now is a good opportunity to clean up our protobufs.  The only change here I'm not sure about is the `config.Config` protobufs, as the comments were fairly against doing this.  It seems to be fine to me as these protobufs aren't actually serialized as far as I can find - the config files are stored as JSON and parsed into an in-memory proto struct from there.


src/internal/config/config.proto
```diff
@@ -13,8 +13,8 @@ import "gogoproto/gogo.proto";
 // message (this allows us to make significant changes to the config structure
 // without breaking existing users by defining a new config version).
 //
-// DO NOT change or remove field numbers from this proto, otherwise ALL
-// pachyderm user configs will fail to parse.
+// These structures are stored in a JSON format, so it should be safe to modify
```
brycemcanally, line 16: Yeah, this seems correct.

brycemcanally (approved):
lgtm

PR #6139, opened 2021-05-17
Work around postgres error of too large of a notify payload on large collection rows

Tryneus:
This error was causing `TestPFS/FuzzProvenance` to fail because it generated very large protobufs due to deep DAGs with many provenances/subvenances in a single `CommitInfo`.  Added a new collection test that passed on `EtcdCollection` and failed on `PostgresCollection` with the correct error, then fixed it for postgres.

```
=== CONT  TestPFS/FuzzProvenance
    server_test.go:5304: No error is expected but got pq: payload string too long
```

Postgres has an upper limit of 8000 bytes in a notify payload.  The suggested workaround for this is to write the payload to a table and let the client read it from there, so this PR implements that.  It will still inline the data in the payload if it fits, so the notify trigger function was refactored to allow both paths.  Reorganized the order of fields in the payload to make the notify trigger function a bit more straightforward.

There are still a couple limitations we need to be cognizant of - we are only writing the protobuf data to the table and the other fields are still inline in the payload, so the collection's primary key + any index key must not be too long as well.  We can enforce the primary key length via the keycheck functions, but I haven't decided on the best way to do it for secondary keys yet.


src/internal/collection/postgres_listener.go
```diff
@@ -169,62 +174,75 @@ func parsePostgresEpoch(s string) (time.Time, error) {
...
 	if pe.err != nil {
 		return &watch.Event{Err: pe.err, Type: watch.EventError}
 	}
 	if pe.eventType == watch.EventDelete {
 		// Etcd doesn't return deleted row values - we could, but let's maintain parity
 		return &watch.Event{Key: []byte(pe.key), Type: pe.eventType, Template: template}
 	}
+	if pe.protoData == nil && pe.storedID != "" {
+		// The proto data was too large to fit in the payload, read it from a temporary location.
+		if err := db.QueryRowContext(ctx, "select proto from collections.large_notifications where id = $1", pe.storedID).Scan(&pe.protoData); err != nil {
```
Tryneus, line 240: Here's where we read the proto data from the `large_notifications` table - when we're delivering the event to the channel the user is reading from.  We could possibly make this more efficient in the future by caching these in case multiple listeners are receiving the same event, but it isn't worth the complexity at this point.

src/internal/collection/postgres_schema.go
```diff
@@ -102,6 +124,48 @@ $$ language 'plpgsql';
...
+	}
+
+	notificationsIndex := "create index on collections.large_notifications (createdat);"
+	if _, err := sqlTx.Exec(notificationsIndex); err != nil {
+		return errors.EnsureStack(err)
+	}
+
+	// The sweep function makes sure that the size of the large_notifications
+	// table is bounded by deleting notifications over a certain age when a new
+	// notification is written.
```
Tryneus, line 147: Also add a function to delete old notification proto data any time we write to the `large_notifications` table because unbounded growth bad.

src/internal/collection/postgres_schema.go
```diff
@@ -129,6 +193,13 @@ func SetupPostgresCollections(ctx context.Context, sqlTx *sqlx.Tx, collections .
 			return errors.EnsureStack(err)
 		}
 
+		for _, idx := range col.indexes {
```
Tryneus, line 196: Unrelated: I noticed we weren't actually creating secondary indexes on the collections tables, so non-pkey queries were not efficient - fixed here.

src/internal/collection/watch_test.go
```diff
@@ -53,7 +54,7 @@ func newTestEvent(t *testing.T, ev *watch.Event) TestEvent {
 }
 
 type WatchTester interface {
-	Write(id string, value ...string)     // writes a row with the given ID and value
+	Write(item *col.TestItem)             // writes the given row
```
Tryneus, line 57: refactored some of the test utility functions because they encouraged repetition of magic values

brendoncarroll (approved):

PR #6140, opened 2021-05-17
Remove pattern of passing pach client in PFS

actgardner:
In several places in the codebase we pass a pach client around, to propagate a context for calls within a single service. This eliminates this pattern in PFS. This is a first step to make it clearer where we're actually making cross-service calls (and consequently creating nested transactions) so we can migrate those to method calls that propagate a transaction.


Tryneus (approved):
LGTM

PR #6141, opened 2021-05-17
Fix concurrent transaction usage when authorizing in pps

Tryneus:
Found another place where we use a sqlx.Tx in an unsafe manner.  We may need some sort of locking in the future to ensure this doesn't crop up again.


PFedak (approved):
LGTM

PR #6143, opened 2021-05-17
Refactor PPS to not pass pach clients around

actgardner:
Follow up to https://github.com/pachyderm/pachyderm/pull/6140 but for PPS.

Eliminates internal methods that accept pach client in favour of passing around a ctx and creating a client only where needed. The next step is to replace these RPCs with internal method calls.


Tryneus (approved):
LGTM

PR #6146, opened 2021-05-18
Delete GarbageCollectRequest/Response

msteffen:
AFAICT these messages are no longer used, as GarbageCollect has been removed from the API


brycemcanally (approved):
lgtm

PR #6147, opened 2021-05-18
Fix broken worker pachClient auth in postgres integration branch

Tryneus:
I had changed the worker driver to get its PachClient straight from the `ServiceEnv`.  This has made a lot of tests very angry and has been widely regarded as a bad move.  Revert it so that the credentialed client in `worker/main.go` is passed down and reused in the driver.


acohen4 (approved):
LGTM!

PR #6148, opened 2021-05-18
Fix worker killJob to handle errors better

Tryneus:
`killJob` attempts to transactionally finish the output commits and job, but will fail if any of those operations fail (such as an output commit being closed).  Moved this to use the new and improved `StopJob`, which does the same thing (transactionally) except that it will continue if an output commit is closed or job is missing.


brycemcanally (approved):
lgtm

PR #6150, opened 2021-05-18
Define internal service interfaces

actgardner:
To move away from doing RPCs between services, define internal interfaces for the services and register them with the ServiceEnv. This replaces the old transaction-only interfaces defined in the transactionenv package.

This PR also removes the pachClient field from TransactionContext. In some cases where it would be complicated to eliminate an RPC I've opted to use GetPachClient on the serviceenv for now, but in a follow-on PR we should be able to remove any use of GetPachClient in pachd.


src/internal/serviceenv/service_env.go
```diff
@@ -32,6 +35,10 @@ const clusterIDKey = "cluster-id"
 // create more, if they want to create multiple pachyderm "clusters" served in
 // separate goroutines.
 type ServiceEnv interface {
+	AuthServer() auth_server.APIServer
```
actgardner, line 38: Add accessors to ServiceEnv so we can do method calls instead of RPCs between services.

src/internal/transactionenv/context/context.go
```diff
@@ -0,0 +1,69 @@
...
+// objects, which will be threaded through to every API call:
+//   ctx: the client context which initiated the operations being performed
+//   pachClient: the APIClient associated with the client context ctx
+//   stm: the object that controls transactionality with etcd.  This is to ensure
+//     that all reads and writes are consistent until changes are committed.
+//   txnEnv: a struct containing references to each API server, it can be used
+//     to make calls to other API servers (e.g. checking auth permissions)
+//   pfsDefer: an interface for ensuring certain PFS cleanup tasks are performed
+//     properly (and deduped) at the end of the transaction.
+type TransactionContext struct {
```
actgardner, line 22: Split `TransactionContext` out of `transactionenv` to avoid a circular dependency between the API definitions and the transactionenv package.

src/internal/transactionenv/env.go
```diff
@@ -136,81 +60,22 @@ type TransactionServer interface {
 	) (*transaction.TransactionResponse, error)
 }
 
-// AuthTransactionServer is an interface for the transactionally-supported
```
actgardner, line 139: Move these interfaces into the packages that implement them.

src/server/auth/server/util.go
```diff
@@ -1,23 +1,24 @@
...
 	"github.com/pachyderm/pachyderm/v2/src/internal/grpcutil"
-	txnenv "github.com/pachyderm/pachyderm/v2/src/internal/transactionenv"
+	"github.com/pachyderm/pachyderm/v2/src/internal/transactionenv/context"
 )
 
 // CheckClusterIsAuthorizedInTransaction returns an error if the current user doesn't have
 // the permissions in `p` on the cluster
-func CheckClusterIsAuthorizedInTransaction(txnCtx *txnenv.TransactionContext, p ...auth.Permission) error {
-	me, err := txnCtx.Client.WhoAmI(txnCtx.ClientContext, &auth.WhoAmIRequest{})
+func (a *apiServer) CheckClusterIsAuthorizedInTransaction(txnCtx *txncontext.TransactionContext, p ...auth.Permission) error {
```
actgardner, line 14: Switch these utility methods to be part of the internal API, before they were implemented in a strange way.

src/server/auth/server/util.go
```diff
@@ -48,15 +49,14 @@ func CheckRepoIsAuthorizedInTransaction(txnCtx *txnenv.TransactionContext, r str
 
 // CheckRepoIsAuthorized returns an error if the current user doesn't have
 // the permissions in `p` on the repo `r`
-func CheckRepoIsAuthorized(pachClient *client.APIClient, r string, p ...auth.Permission) error {
-	ctx := pachClient.Ctx()
-	me, err := pachClient.WhoAmI(ctx, &auth.WhoAmIRequest{})
+func (a *apiServer) CheckRepoIsAuthorized(ctx context.Context, r string, p ...auth.Permission) error {
```
actgardner, line 52: Switch this non-transaction utility method to be part of the internal API.

src/internal/testpachd/mock_transaction.go
```diff
@@ -31,6 +31,7 @@ func (mock *mockCreatePipelineInTransaction) Use(cb createPipelineInTransactionF
 }
 
 type ppsTransactionAPI struct {
+	ppsServerAPI
```
actgardner, line 34: The PPS server we give to the serviceenv has to expose all the API methods and the transaction methods, this seemed like the simplest way to achieve that (and the PFS tests pass) but I don't know this part of the codebase very well.

src/internal/transactionenv/context/context.go
```diff
@@ -0,0 +1,67 @@
+package txncontext
```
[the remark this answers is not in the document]
brendoncarroll: It's not incorrect from Go's perspective. But it does break auto-imports. I would say rename the directory to `txncontext`

PR #6151, opened 2021-05-18
Merge postgres-integration branch to master

Tryneus:
Most of the code in here has already been reviewed as part of other PRs, but I figured it would be good to do a PR for tracking the merge of `postgres-integration` into `master` now that CI is passing.

A few flaky tests were skipped to get to green CI, but as far as I am aware, these tests are also flaky on `master` (due to slow pipelines, for the most part):
* `TestLokiLogs`
* `TestStandby/ChainOf10`
* `TestStandby/ManyCommits`


src/internal/collection/postgres_collection.go
```diff
@@ -96,61 +90,68 @@ func (c *postgresCollection) ReadWrite(tx *sqlx.Tx) PostgresReadWriteCollection
 // NewSQLTx starts a transaction on the given DB, passes it to the callback, and
 // finishes the transaction afterwards. If the callback was successful, the
 // transaction is committed. If any errors occur, the transaction is rolled
-// back.  This will reattempt the transaction up to three times.
+// back.  This will reattempt the transaction forever.
 func NewSQLTx(ctx context.Context, db *sqlx.DB, apply func(*sqlx.Tx) error) error {
```
brendoncarroll, line 94: We have something like this in `dbutil`.

src/internal/pfsdb/pfsdb.go
```diff
@@ -14,85 +14,133 @@ import (
...
 
+var ReposTypeIndex = &col.Index{
+	Name: "type",
+	Extract: func(val proto.Message) string {
+		return val.(*pfs.RepoInfo).Repo.Type
+	},
+}
+
+var ReposNameIndex = &col.Index{
+	Name: "name",
```
brendoncarroll, line 31: Do either of these names work?  Aren't they both reserved words?

src/server/pfs/server/driver.go
```diff
@@ -380,8 +413,7 @@ func (d *driver) deleteRepo(txnCtx *txnenv.TransactionContext, repo *pfs.Repo, f
 				return errors.Wrapf(err, "err fixing subvenance of upstream commit %s@%s", prov.Commit.Branch.Repo.Name, prov.Commit.ID)
 			}
 		}
-		// TODO: use DropFilesetsTx
-		if err := d.commitStore.DropFilesets(txnCtx.ClientContext, ci.Commit); err != nil {
+		if err := d.commitStore.DropFilesetsTx(txnCtx.SqlTx, ci.Commit); err != nil {
```
brendoncarroll, line 416: 👍

src/server/pfs/server/driver.go
```diff
@@ -325,18 +345,31 @@ func (d *driver) deleteRepo(txnCtx *txnenv.TransactionContext, repo *pfs.Repo, f
 		return err
 	}
 
-	repoInfo := new(pfs.RepoInfo)
-	if err := repos.Get(repo.Name, repoInfo); err != nil {
-		if !col.IsErrNotFound(err) {
-			return errors.Wrapf(err, "repos.Get")
+	// if this is a user repo, delete any dependent repos
+	if repo.Type == pfs.UserRepoType {
```
brendoncarroll, line 349: Where is the current design doc for System Repos? After the last meeting we had about this, I was under the impression that this operation could be done without a read.

brendoncarroll: I'm mostly curious about the table structure here.

src/server/pfs/server/driver_file.go
```diff
@@ -416,25 +416,27 @@ func (d *driver) renewFileset(ctx context.Context, id fileset.ID, ttl time.Durat
...
-	return d.commitStore.AddFileset(ctx, commitInfo.Commit, filesetID)
+	return d.commitStore.AddFilesetTx(txnCtx.SqlTx, commitInfo.Commit, filesetID)
 }
 
 func (d *driver) getFileset(ctx context.Context, commit *pfs.Commit) (*fileset.ID, error) {
 	commitInfo, err := d.getCommit(ctx, commit)
 	if err != nil {
 		return nil, err
 	}
+	/* TODO: brendon: re-enable compaction
```
brendoncarroll, line 435: This needs to be removed.  Compaction is enabled in master.

src/internal/pfsdb/pfsdb.go
Tryneus: The index name is mapped by the `PostgresCollection` to the field `idx_name`

src/server/pfs/server/testing/server_test.go
```diff
@@ -571,6 +572,8 @@ func TestPFS(suite *testing.T) {
 	})
 
 	suite.Run("InspectCommit", func(t *testing.T) {
+		// TODO(2.0 required): Fix commit size with lazy compaction
+		t.Skip("Need to fix commit size on FinishCommit with lazy compaction in V2")
```
brendoncarroll, line 576: Remove. We skip just the size check in master.

src/server/pfs/server/testing/server_test.go
```diff
@@ -2230,6 +2236,8 @@ func TestPFS(suite *testing.T) {
 	})
 
 	suite.Run("InspectRepoComplex", func(t *testing.T) {
+		// TODO(2.0 required): Fix commit size with lazy compaction
+		t.Skip("Need to fix commit size on FinishCommit with lazy compaction in V2")
```
brendoncarroll, line 2240: Remove

src/server/pachyderm_test.go
```diff
@@ -105,6 +105,8 @@ func TestSimplePipeline(t *testing.T) {
 }
 
 func TestRepoSize(t *testing.T) {
+	// TODO(2.0 required): Fix commit size with lazy compaction
+	t.Skip("Need to fix commit size on FinishCommit with lazy compaction in V2")
```
brendoncarroll, line 109: Remove.  See line 155

src/internal/collection/postgres_collection.go
Tryneus: The implementations are different enough that we can't just use the `dbutil` one - `NewSQLTx`:
 * recognizes a custom `TransactionConflictErr` type as a reattempt indicator (used by `CreatePipelineInTransaction` to retry once the fileset is created)
 * recognizes all postgres transaction rollback errors instead of just serialization errors
 * doesn't do backoff for reattempts

In additon, we also have `NewDryrunSQLTx` which is different from doing `ReadOnly=true` because we actually do want writes to go through, just that they will be rolled back at the end (there may be better ways of doing this, but I suspect it will be complicated).

src/server/pfs/server/driver.go
PFedak: doc is [here](https://docs.google.com/document/d/1eq4_wIbh7-M4r3hcZWGFEPYU9FN8eGQIp0b4dQaORxA/edit#heading=h.3ivaro2ikgbw), and [this](https://github.com/pachyderm/pachyderm/pull/6070/files#diff-9c8c0ec115494d42aeabac6cddc20e58b1bc9fd5d44f05b879007998d08de8fb) is the specific PR.

I'm not exactly sure what you mean by "this operation", we can definitely delete the repo entries directly, but I don't see how the other components of `DeleteRepo` could happen as easily.

brendoncarroll (approved):
I only covered the PFS changes, but they look fine.  There were some regressions to fixes in master (from our pairing session) that we discussed offline.

My curiousness about the System Repos should not be a blocker since that has already been reviewed.

src/internal/collection/postgres_collection.go
brendoncarroll: It seems like these would all be improvements to the existing dbutil.WithTx, except for maybe the custom error. I don't think I have enough context to know why the custom error is needed.  Is it for a weaker isolation level?

Tryneus: No, it's that `CreatePipelineInTransaction` has to do some reads before it can produce the new spec file, which it then writes out to a fileset.  The fileset is necessarily outside of the transaction, so it can't add that fileset to the spec commit unless we start the transaction over again.  So, really, the custom error is useful if the apply function needs to restart the transaction to pull in the latest state.

PR #6152, opened 2021-05-19
Change file-related functions to take commit arguments.

PFedak:


PFedak:
The main motivation for this change is to facilitate doing basic operations on generic system repos. It does leave the common pfs functions in a weird state where some common ones still take the string arguments (notably `FinishCommit`), which could also be switched over.

Tryneus (approved):
LGTM - I think we can extend on this, but it seems like a step in the right direction.  I'd like to give the `*Commit` and possibly the `*Repo` RPC wrapper functions the same treatment in the future.

PR #6153, opened 2021-05-20
Add commit sizes to InspectCommit

brendoncarroll:
- Commit sizes were temporarily removed when compaction was moved to read time.  This adds them back and re-enables the checks in the tests.
- Repo size is retrieved from the commit at read time rather than stored with the repo info in the repo collection.

### DIFF
--- a/src/server/pfs/server/driver.go
+++ b/src/server/pfs/server/driver.go
@@ -261,6 +261,11 @@ func (d *driver) inspectRepo(txnCtx *txnenv.TransactionContext, repo *pfs.Repo, 
 	if err := d.repos.ReadWrite(txnCtx.SqlTx).Get(pfsdb.RepoKey(repo), result); err != nil {
 		return nil, err
 	}
+	size, err := d.getRepoSize(txnCtx.ClientContext, repo)
+	if err != nil {
+		return nil, err
+	}
+	result.SizeBytes = uint64(size)
 	if includeAuth {
 		permissions, roles, err := d.getPermissions(txnCtx.ClientContext, repo)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


src/server/pfs/server/driver_file.go
```diff
@@ -490,3 +490,12 @@ func (d *driver) getOrComputeTotal(ctx context.Context, commit *pfs.Commit) (*fi
...
 	return d.commitStore.GetTotalFileset(ctx, commit)
 }
+
+// sizeOfCommit gets the size of a commit.
+func (d *driver) sizeOfCommit(ctx context.Context, commit *pfs.Commit) (int64, error) {
+	fsid, err := d.getFileset(ctx, commit)
+	if err != nil {
+		return 0, err
+	}
+	return d.storage.SizeOf(ctx, *fsid)
```
brycemcanally, line 500: Just to make sure we are on the same page, this might not be the actual size of the commit since deletions may not have been applied yet. I think we need to actually do a merge read to compute this.

src/server/pfs/server/driver.go
```diff
@@ -848,28 +858,33 @@ func (d *driver) writeFinishedCommit(sqlTx *sqlx.Tx, commit *pfs.Commit, commitI
...
-	// update the repo size if this is the head of master
 	repoInfo := new(pfs.RepoInfo)
 	if err := d.repos.ReadWrite(sqlTx).Get(pfsdb.RepoKey(commit.Branch.Repo), repoInfo); err != nil {
 		return err
 	}
+	return nil
+}
+
+func (d *driver) getRepoSize(ctx context.Context, repo *pfs.Repo) (int64, error) {
+	// update the repo size if this is the head of master
```
brycemcanally, line 869: I don't think this comment is relevant anymore since we are not storing this.

brycemcanally (approved):
lgtm, just remove a comment that is no longer applicable.

src/server/pfs/server/driver_file.go
brendoncarroll: Yeah.  I think we can make that change in SizeOf.  Maybe we should have a method that returns a range?  And we can just use the sizes of the additive and deletive set?  Then you can get approximate sizes without too much work.

PR #6157, opened 2021-05-20
Allow VisitInput to bubble up an error.

PFedak:
This caused a problem in create pipeline where continuing past an error
in a transaction led to an expected postgres error code. This
functionality is hacked on in variously awkward ways across our
codebase, anyway, so probably worth doing.

### DIFF
--- a/src/internal/ppsutil/util.go
+++ b/src/internal/ppsutil/util.go
@@ -31,6 +31,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/client"
 	col "github.com/pachyderm/pachyderm/v2/src/internal/collection"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/errutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/ppsconsts"
 	"github.com/pachyderm/pachyderm/v2/src/internal/tracing"
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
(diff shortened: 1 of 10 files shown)


src/internal/ppsutil/util.go
```diff
@@ -435,16 +437,13 @@ func GetStatsCommit(commitInfo *pfs.CommitInfo) *pfs.Commit {
...
-		}
+	foundErr := pps.VisitInput(in, func(in *pps.Input) error {
 		if in.Pfs != nil && in.Pfs.S3 {
-			found = true
+			return errutil.ErrBreak
 		}
+		return nil
 	})
-	return found
+	return foundErr != nil
```
PFedak, line 446: Happy to revert this and the other cases that are looking for an input type if this is too "clever"/not what `ErrBreak` is expected to do.

src/server/worker/pipeline/transform/registry.go
```diff
@@ -601,31 +601,36 @@ func (reg *registry) processJobEgressing(ppj *pendingPipelineJob) error {
...
+				return err
+			}
 		}
+		return nil
 	})
-	return failed, vistErr
+	if visitErr != nil {
+		return nil, visitErr
+	}
+	return failed, nil
```
PFedak, line 635: I'm explicitly changing behavior here (to match the usual conventions), but this shouldn't affect anything since the only call site ignores `failed` on error anyway

src/pps/util.go
```diff
@@ -32,28 +32,26 @@ func init() {
...
-			VisitInput(input, f)
+		source = input.Union
+	}
+	for _, input := range source {
+		if err := VisitInput(input, f); err != nil {
+			return err
 		}
 	}
-	f(input)
+	return f(input)
```
Tryneus, line 54: I think the common way we recognize `ErrBreak` is to catch it and return `nil`, in which case I think we'd need that logic to be in a wrapper function.

Tryneus (changes requested):
A couple nits, but really happy with this change

src/internal/ppsutil/util.go
PFedak: That's a good point, I'll put it back and make `ErrBreak` behave the way it does in other callback iterators

Tryneus (approved):

PR #6158, opened 2021-05-20
Run 1.13.x tests on CircleCI VMs

actgardner:
Run 1.13.x tests on CircleCI VMs (Ubuntu 20.04). 

- This involved switching from the none driver to docker driver, which changed the IP when referring to pods in k8s (from localhost to VM_IP). 

- I also had to add the `-CIbuild` prefix to the golden manifests, because we set the `ADDITIONAL_VERSION` only when building in Circle (not sure of the reason for this? it affects master as well).

- Changed the paths for the old version extracts in the admin tests to be relative - go tests always run in a predictable location, this was simpler than figuring out the new absolute path.

- Switch the tests that run in a separate postgres to use the postgres pod in k8s. Previously we ran a separate docker image, but the networking is complicated by switching to the docker driver.


etc/testing/deploy-manifests/golden/amazon-deploy-manifest.json
```diff
@@ -515,7 +515,7 @@
         "containers": [
           {
             "name": "pachd",
-            "image": "pachyderm/pachd:1.13.2",
+            "image": "pachyderm/pachd:1.13.2-CIbuild",
```
nitinjainsj, line 518: The -CIbuild may be fine here, but when we do a release this will change to 1.13.3. I wonder how that would break. IIRC, -CIbuild was needed for testfaster to tag/pull a known image in its layers of cache.

Can you try one more thing... remove the -CIbuild from the top level Makefile? Let it build without VERSION_ADDITIONAL. If that works, that might be the right answer instead of changing the golden manifest.

nitinjainsj (commented):
Left one comment. Otherwise looks good to me.

nitinjainsj (approved):

PR #6164, opened 2021-05-21
Eliminate most uses of Repo.Name

PFedak:


src/server/pfs/pfs.go
```diff
@@ -73,23 +74,23 @@ type ErrCommitNotFinished struct {
 }
 
 func (e ErrFileNotFound) Error() string {
-	return fmt.Sprintf("file %v not found in repo %v at commit %v", e.File.Path, e.File.Commit.Branch.Repo.Name, e.File.Commit.ID)
+	return fmt.Sprintf("file %v not found in repo %v at commit %v", e.File.Path, pretty.CompactPrintRepo(e.File.Commit.Branch.Repo), e.File.Commit.ID)
```
PFedak, line 77: I tried to use `CompactPrint*` for more user-facing messages, and the various db key functions for where we were essentially creating a key anyway. I would have just implemented `.String()` for `Repo` etc but the proto already has that. It seems like `.Format()` might be higher priority, so that could be an alternative.

src/server/pfs/server/driver_fsck.go
```diff
@@ -338,10 +339,7 @@ func (d *driver) fsck(ctx context.Context, fix bool, cb func(*pfs.FsckResponse)
 					}
 					contains := false
 					for _, headProv := range headCommitInfo.Provenance {
-						if provBranchInfo.Head.Branch.Repo.Name == headProv.Commit.Branch.Repo.Name &&
-							provBranchInfo.Branch.Repo.Name == headProv.Commit.Branch.Repo.Name &&
-							provBranchInfo.Branch.Name == headProv.Commit.Branch.Name &&
-							provBranchInfo.Head.ID == headProv.Commit.ID {
+						if proto.Equal(provBranchInfo.Head, headProv.Commit) {
```
PFedak, line 342: I wasn't sure if we tend to avoid `proto.Equal`, I can definitely understand not wanting to use `reflect` more than necessary

src/server/worker/pipeline/transform/registry.go
```diff
@@ -331,18 +337,38 @@ func (reg *registry) startPipelineJob(commitInfo *pfs.CommitInfo) error {
 					ppj.logger.Logf("error incrementing restart count for job (%s): %v", ppj.pji.PipelineJob.ID, err)
 				}
 				// Reload the job's commitInfo(s) as they may have changed and clear the state of the commit(s).
-				ppj.commitInfo, err = reg.driver.PachClient().InspectCommit(ppj.commitInfo.Commit.Branch.Repo.Name, ppj.commitInfo.Commit.Branch.Name, ppj.commitInfo.Commit.ID)
+				pachClient := reg.driver.PachClient()
+				ppj.commitInfo, err = pachClient.PfsAPIClient.InspectCommit(
```
PFedak, line 341: these and others are kind of awkward since I decided to only alter functions which did logic beyond constructing the request. Just using protos everywhere would definitely be more consistent, just even more call sites to change

Tryneus (approved):
Like we've discussed in-person, in general I think this is not super attractive in its current state, but it is a step in the right direction (having explicit types on interfaces is clearer than string after string parameters).  I think we'll have other changes to make before 2.0 to clean this up, but for the purpose of unblocking system repos, I think we should merge this now.

src/internal/ppsutil/util.go
```diff
@@ -274,22 +274,22 @@ func PipelineJobInput(pipelineInfo *pps.PipelineInfo, outputCommitInfo *pfs.Comm
 	// for a given branch, the commit assigned to it will be the latest commit on that branch
 	// this is ensured by the way we sort the commit provenance when creating the outputCommit
 	for _, prov := range outputCommitInfo.Provenance {
-		branchToCommit[key(prov.Commit.Branch.Repo.Name, prov.Commit.Branch.Name)] = prov.Commit
+		branchToCommit[key(prov.Commit.Branch.Repo.Name, prov.Commit.Branch.Repo.Type, prov.Commit.Branch.Name)] = prov.Commit
```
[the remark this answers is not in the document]
PFedak: Makes sense, done

src/client/pfs.go
```diff
@@ -444,18 +439,18 @@ func (c APIClient) FlushCommitAll(commits []*pfs.Commit, toRepos []*pfs.Repo) ([
 
 // SubscribeCommit is like ListCommit but it keeps listening for commits as
 // they come in.
-func (c APIClient) SubscribeCommit(repoName string, branchName string, prov *pfs.CommitProvenance, from string, state pfs.CommitState, cb func(*pfs.CommitInfo) error) (retErr error) {
+func (c APIClient) SubscribeCommit(repo *pfs.Repo, branchName string, prov *pfs.CommitProvenance, from string, state pfs.CommitState, cb func(*pfs.CommitInfo) error) (retErr error) {
```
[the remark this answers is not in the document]
PFedak: Agreed, I'll plan on another pass soon

PR #6166, opened 2021-05-23
Files disappear from a commit when squashing a commit with an empty child

Tryneus:
This PR adds a test that reproduces a failure on 2.0 where squashing a commit with an empty commit as a child will result in the child having an empty fileset.  If we were to read the child commit before the squash, it does not lose the files after the squash.


PFedak:
Fixed by #6712

PR #6167, opened 2021-05-23
Fix TestDatumStatusRestart flakiness

brycemcanally:
This PR fixes the issue that was causing TestDatumStatusRestart to be flaky. There was a race in the test such that it would get a worker status while the datum was not being actively processed, which would result with the test checking against the zeroed out started field (thanks for pointing this out @PFedak). I restructured the worker status a bit such that we now have a datum status field which will be nil when a datum is not currently being processed. The test now checks if this is nil, and retries until it is not. Also, I removed some fields from the worker status proto that are not in use.


PFedak (approved):
The fix LGTM, is `QueueSize` just not relevant in 2.0?

brycemcanally:
Yeah, we might want to readd it later, but for 2.0 GA it will not exist.

PR #6169, opened 2021-05-24
Expose PFS load testing

brycemcanally:
This PR refactors and exposes the PFS load testing. The load testing has been simplified a bit by removing the required nesting of the components (operations, files, sizes) in a `fuzz*` component. The PFS load testing has been exposed through gRPC (`RunLoadTest`) and has a corresponding wrapper in the Go client (`RunPFSLoadTest`) / pachctl (`pachctl run pfs-load-test`). RunLoadTest takes a load test spec file (the specification and an example are viewable through `pachctl run pfs-load-test -h`) and a seed. RunLoadTest returns the branch that was generated for running the load test (the state is still viewable after the test completes, so additional validation can be run), the seed that was used, and any errors that occurred while running the load test.

Miscellaneous:
- Fixed a goroutine leak in fileset iterator (not respecting context in goroutine).
- Fixed an issue with fileset sharding (path ranges should be inclusive).
- ~~Updated and reenabled `LargeDeleteRepo` test.~~ (removed the test for now, will get the same coverage through a load test)
- ~~Reenabled standby tests (may be addressed by bug fix).~~ (still flaky)


src/client/pfs.go
```diff
@@ -45,11 +45,15 @@ func NewCommitProvenance(repoName string, branchName string, commitID string) *p
 }
 
 // NewFile creates a pfs.File.
-func NewFile(repoName string, branchName string, commitID string, path string) *pfs.File {
-	return &pfs.File{
+func NewFile(repoName string, branchName string, commitID string, path string, tag ...string) *pfs.File {
```
brendoncarroll, line 48: This snuck back in.

src/internal/fileutil/file.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+)
+
+func WithTmpFile(prefix string, cb func(*os.File) error) (retErr error) {
+	if err := os.MkdirAll(os.TempDir(), 0700); err != nil {
```
brendoncarroll, line 11: I don't think we should be initializing the operating systems tmp directory in a library.  That check should happen during setup.

src/internal/fileutil/file.go
```diff
@@ -0,0 +1,27 @@
+package fileutil
```
brendoncarroll, line 1: Maybe call this fsutil, so that we can put other things in here?

src/internal/pfsload/env.go
```diff
@@ -1,30 +1,37 @@
-package load
+package pfsload
+
+import (
+	"math/rand"
```
brendoncarroll, line 4: :+1:

src/internal/pfsload/file.go
```diff
@@ -1,16 +1,26 @@
...
-	"github.com/pachyderm/pachyderm/v2/src/internal/storage/chunk"
-	"github.com/pachyderm/pachyderm/v2/src/internal/uuid"
 )
 
+const pathSize = 32
+
+var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+// randSeq generates a random sequence of data (n is number of bytes)
+func randSeq(n int, random *rand.Rand) string {
```
brendoncarroll, line 16: I think we do something like this in the chunk layer too? Can we put these in `randutil`.  I think right now just UniqueString lives in there.

src/server/pfs/cmds/cmds.go
```diff
@@ -1311,6 +1312,36 @@ Objects are a low-level resource and should not be accessed directly by most use
...
+			c, err := client.NewOnUserMachine("user")
+			if err != nil {
+				return err
+			}
+			defer func() {
+				if err := c.Close(); retErr == nil {
+					retErr = err
+				}
+			}()
+			spec, err := ioutil.ReadFile(args[0])
```
brendoncarroll, line 1331: Is `len(args) > 0` ?  If the users can touch this then it shouldn't be able to panic.

src/server/pfs/server/testing/server_test.go
```diff
@@ -5876,16 +5865,18 @@ func TestPFS(suite *testing.T) {
...
-		t.Parallel()
-		env := testpachd.NewRealEnv(t, tu.NewTestDBConfig(t))
-		spec := &load.CommitsSpec{}
-		require.NoError(t, yaml.UnmarshalStrict([]byte(testLoad), spec))
-		msg := random.SeedRand()
-		c := env.PachClient
-		repo := "test"
-		require.NoError(t, c.CreateRepo(repo))
-		require.NoError(t, load.Commits(c, repo, "master", spec), msg)
+	suite.Run("TestLoad", func(subsuite *testing.T) {
```
brendoncarroll, line 5840: This shouldn't run in the suite with the behavioral tests.  Can we move this to a separate function `TestLoad(t *testing.T)`

brendoncarroll:
I think I mentioned this in an earlier P.R, but it's worth mentioning again.  Representing probabilities as floats and parsing them is going to surprise the user at some point.

1. If they don't sum to 1, then the user will get something unexpected. Unless you check they sum to 1, but...
2. There is a good chance they won't actually sum to 1 because of how floating point arithmetic works.  I don't see where that validation happens in here, but I'll just mention that: `floatA == floatB` is rarely what you want.  You need to check that they are withing some epsilon of each other.

Weights/ratios on the other hand are the same math, but the configuration is done with integers, and the only validation you need to do is check they aren't unset (zero).

src/server/pfs/cmds/cmds.go
brycemcanally: The wrapping function (`RunFixedArgs`) checks that exactly one argument is provided, and errors otherwise.

brycemcanally:
We discussed how we are representing the probabilities offline. We are moving to integer based probabilities in the range of [0, 100]. In the case where we are selecting from a set of components based on a probability per component, the probabilities must add up to 100.

PR #6170, opened 2021-05-24
Fix TestConfig in auth commands

acohen4:
There are two pieces to this test fix.
- Configuring OIDC Provider so that data passed to the `set-config` command is validated
-  Fixing JSON/YAML syntax in string comparisons

### DIFF
--- a/src/server/auth/cmds/cmds_test.go
+++ b/src/server/auth/cmds/cmds_test.go
@@ -276,26 +276,25 @@ func TestGetAndUseRobotToken(t *testing.T) {
 }
 
 func TestConfig(t *testing.T) {
-	// TODO(2.0 required): Investigate test not passing.
-	t.Skip("Test not passing")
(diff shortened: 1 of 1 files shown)


src/server/auth/cmds/cmds_test.go
```diff
@@ -241,37 +241,37 @@ func TestGetAndUseRobotToken(t *testing.T) {
...
 	if testing.Short() {
 		t.Skip("Skipping integration tests in short mode")
 	}
 	tu.ActivateAuth(t)
+	tu.ConfigureOIDCProvider(t)
+
 	defer tu.DeleteAll(t)
 
 	require.NoError(t, tu.BashCmd(`
 		pachctl auth set-config <<EOF
```
msteffen, line 253: Should this line be de-dented  too?

msteffen (approved):
LGTM, with one style question

PR #6171, opened 2021-05-24
Fixing server_test.go#TestPanicOnNilArgs and removing Redundant auth validations on DeleteRepoInTransaction()

acohen4:
Included in this PR:
- Fixing Skipped test, `server_test.go#TestPanicOnNilArgs`
- Removing redundant auth validation in `val_server.go#DeleteRepoInTransaction`
- Adding Auth test to verify the Delete All Repos functionality

### DIFF
--- a/src/server/pfs/server/testing/server_test.go
+++ b/src/server/pfs/server/testing/server_test.go
@@ -5936,7 +5936,6 @@ func TestPFS(suite *testing.T) {
 
 	suite.Run("TestPanicOnNilArgs", func(t *testing.T) {
 		// TODO(2.0 required): Add validation to all PFS endpoints.
-		t.Skip("PFS endpoints are not fully validated in V2")
 		t.Parallel()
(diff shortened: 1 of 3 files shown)


src/server/pfs/server/testing/server_test.go
```diff
@@ -5936,7 +5936,6 @@ func TestPFS(suite *testing.T) {
 
 	suite.Run("TestPanicOnNilArgs", func(t *testing.T) {
 		// TODO(2.0 required): Add validation to all PFS endpoints.
```
brycemcanally, line 5938: We can probably get rid of this TODO if this test passes now. We still need to do a full pass over the PFS endpoints to move the validation into the validation server and to make sure we are validating all of the fields that should not be unset (or for certain conditions similar to the repo or all flag being set in the delete repo case). We have a card for this in the 2.0 PFS / PPS project, so we should be fine to remove the TODO here.

brycemcanally (approved):
lgtm, just remove the TODO for the test that has been enabled.

PR #6175, opened 2021-05-24
Update TestListDatumDuringJob() with V2 Semantics

acohen4:
In V1: the behavior we verified is that in progress datums are represented in the ListDatumAll response

In V2: the datums are represented in the ListDatumAll response once their processing is complete.

In this test we now create several batches of datums to process, and verify that they incrementally become represented in the List View as the Job progresses

### DIFF
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -7692,8 +7692,6 @@ func TestPipelineWithDatumTimeout(t *testing.T) {
 }
 
 func TestListDatumDuringJob(t *testing.T) {
-	// TODO(2.0 required): Change semantics of test to list datums that have been processed while the job is running.
-	t.Skip("Semantics need to be changed for V2")
 	if testing.Short() {
 		t.Skip("Skipping integration tests in short mode")
 	}
@@ -7706,7 +7704,13 @@ func TestListDatumDuringJob(t *testing.T) {
 
 	commit1, err := c.StartCommit(dataRepo, "master")
 	require.NoError(t, err)
-	require.NoError(t, c.PutFile(commit1, "file", strings.NewReader("foo"), client.WithAppendPutFile()))
+
+	fileCount := 10
+	for i := 0; i < fileCount; i++ {
+		fileName := "file" + strconv.Itoa(i)
+		require.NoError(t, c.PutFile(commit1, fileName, strings.NewReader("foo")))
+	}
+
 	require.NoError(t, c.FinishCommit(dataRepo, commit1.Branch.Name, commit1.ID))
 	timeout := 20
 	pipeline := tu.UniqueString("TestListDatumDuringJob_pipeline")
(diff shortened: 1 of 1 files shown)


src/server/pachyderm_test.go
```diff
@@ -7708,7 +7706,13 @@ func TestListDatumDuringJob(t *testing.T) {
 
 	commit1, err := c.StartCommit(dataRepo, "master")
 	require.NoError(t, err)
-	require.NoError(t, c.PutFile(commit1, "file", strings.NewReader("foo"), client.WithAppendPutFile()))
+
+	fileCount := 10
+	for i := 0; i < fileCount; i++ {
+		fileName := "file" + strconv.Itoa(i)
+		require.NoError(t, c.PutFile(commit1, fileName, strings.NewReader("foo"), client.WithAppendPutFile()))
```
brycemcanally, line 7713: Something that is worth noting is that we want to over time remove the `WithAppendPutFile` functional option from calls when it is not needed. What it does is turn the put file into an append operation rather than an overwrite. If the overwrite semantics make sense for the use case then that functional option is unnecessary. The reason why it is in a lot of the test cases now is because we recently changed the semantics of `PutFile` from default append to overwrite, so to avoid figuring out which tests need the append vs not, we defaulted to just including the append option.

acohen4: Makes sense. I'll remove this option here. Thanks!

src/server/pachyderm_test.go
```diff
@@ -7745,9 +7752,39 @@ func TestListDatumDuringJob(t *testing.T) {
...
+				return err
+			}
+			if len(dis) == 0 {
+				return errors.Errorf("expected some datums to be listed")
+			}
+			return nil
+		}, backoff.NewTestingBackOff())
+	})
+
+	// see that some datums have been processed
```
brycemcanally, line 7774: I think the number of datums is necessarily >= 1 if we reach here since the retry errors until that is the case. It seems like we just need a check here to ensure that the number of datums recorded by the retry above are less than the total expected, and that the number of datums after the flush are equal to the total expected.

src/server/pachyderm_test.go
```diff
@@ -7745,9 +7752,39 @@ func TestListDatumDuringJob(t *testing.T) {
 		}, backoff.NewTestingBackOff())
 	})
 
+	// initially since no datum chunks have been processed, we receive 0 datums
 	dis, err := c.ListDatumAll(pipelineJobInfo.PipelineJob.ID)
 	require.NoError(t, err)
-	require.Equal(t, 1, len(dis))
+	require.Equal(t, len(dis), 0)
```
brycemcanally, line 7758: Minor nit, but the argument order should be (t, expected, actual), rather than (t, actual, expected).

brycemcanally (approved):
lgtm

PR #6177, opened 2021-05-25
'pachctl enterprise register' cleans up state changes when it fails

acohen4:
`pachctl enterprise register` first creates a cluster record in the license server, and then activates the enterprise service. As part of this change, we revert the creation of the cluster record in case the enterprise service activation fails. This way the user can rerun the command without receiving errors relating to the cluster id not being unique.


msteffen (approved):
LGTM!

PR #6178, opened 2021-05-25
Global IDs PFS support

Tryneus:
This PR adds the PFS components of the global IDs feature:

 * In a given pachyderm transaction, any new commits are created with the same commit ID - aka the `Job.ID` (but they are still unique with the (repo name, repo type, branch name, commit id) tuple.
 * In `propagateCommits`, if there are new commits in the transaction a `pfs.Job` is created
 * Any upstream commits will be aliased via an empty commit (using the same `Job.ID` as the other commits in the transaction)
 * These RPCs may result in a new job:
   * `StartCommit`
     * a new commit is generated for the job, and aliases will be created for any explicit provenance
   * `CreateBranch`
     * when moving a branch head to an existing commit, a commit alias is generated for the new branch head
     * when changing branch provenance, a new commit may be generated in `propagateCommits`
   * `SquashJob` - (renamed from `SquashCommit`), downstream branches may require a new commit if their set of inputs is no longer consistent
 * `CommitInfo.Provenance` and related fields have been removed - the same information can be determined via the `JobInfo` structure

There are a number of user-facing behavior and semantics changes:
 * `SquashCommit` -> `SquashJob` - now takes a job ID instead of a commit ID
   * No longer squashes a tree of downstream commits as 
   * Squashing the head of a branch can be difficult as it is likely that a new job will replace it
   * Squashing the head of a branch may leave the branch head `nil` rather than reverting it to the parent commit of the branch (if that commit is on a different branch)
 * `FlushCommit` -> `FlushJob` - now takes a job ID and a set of branches instead of a set of commits and a set of repos
 * A branch is no longer allowed to be provenant on other branches within the same repo
 * An open commit from an input repo can no longer be used as explicit provenance when starting a commit


src/client/pfs.go
```diff
@@ -362,37 +366,51 @@ func (c APIClient) DeleteBranch(repoName string, branchName string, force bool)
...
+// branches will be considered, otherwise all branches affected by the job are
+// considered.
 //
-// Note that it's never necessary to call FlushCommit to run jobs, they'll run
-// no matter what, FlushCommitF just allows you to wait for them to complete and
+// Note that it's never necessary to call FlushJob to run PipelineJobs, they'll
+// run no matter what, FlushJob just allows you to wait for them to complete and
 // see their output once they do.
-func (c APIClient) FlushCommit(commits []*pfs.Commit, toRepos []*pfs.Repo, cb func(*pfs.CommitInfo) error) (retErr error) {
+func (c APIClient) FlushJob(job *pfs.Job, toBranches []*pfs.Branch, cb func(*pfs.CommitInfo) error) (retErr error) {
```
jdoliner, line 405: Maybe this is a good opportunity to rename Flush. I think it's one of the more confusing verbs in our API, people often don't find it when they need it and some people assume they need it to make jobs run. Strawman proposal: `WaitJob`?

src/internal/pfsdb/pfsdb.go
```diff
@@ -68,9 +67,20 @@ var CommitsRepoIndex = &col.Index{
 	},
 }
 
-var commitsIndexes = []*col.Index{CommitsRepoIndex}
+var CommitsBranchlessIndex = &col.Index{
+	Name: "branchless",
+	Extract: func(val proto.Message) string {
+		return CommitBranchlessKey(val.(*pfs.CommitInfo).Commit)
+	},
+}
```
jdoliner, line 75: What is this indexing? My understanding based on the name is that it's commits that aren't on any branches, but it doesn't seem like there should be any commits like that so I feel like I'm misunderstanding something.

src/pfs/pfs.proto
```diff
@@ -131,36 +131,25 @@ message CommitInfo {
...
 
 message Job {
   string id = 1 [(gogoproto.customname) = "ID"];
 }
 
+message JobCommitInfo {
+  Branch branch = 1;
+  repeated Branch provenance = 2;
+}
+
```
jdoliner, line 144: So the name for this seems kind of weird, my understanding is that this structure is being used to record the link between a branch and its provenance for a specific job. With that in mind I feel like there's a few options for renaming this:

1. We could just use the `BranchInfo` msg instead, it has a few extra fields we don't want so we could just blank those. Not sure how ugly you feel that is.
2. We could just use an embedded msg in the msgs below, it seems like that's the only place it's used.
3. We could call it something like `BranchProvenance` since it seems like its main job is recording that.

src/pfs/pfs.proto
```diff
@@ -131,36 +131,25 @@ message CommitInfo {
...
-  repeated Commit commits = 3;
+  repeated JobCommitInfo job_commits = 3;
 }
 
 message JobInfo {
   Job job = 1;
-  CommitOrigin origin = 2;
-  repeated CommitInfo commits = 3;
+  repeated JobCommitInfo job_commits = 3;
 }
```
jdoliner, line 153: Do we need these as separate messages now? It looks like they have identical definitions.

src/pfs/pfs.proto
```diff
@@ -252,23 +241,26 @@ message CommitInfos {
...
 
-message FlushCommitRequest {
-  repeated Commit commits = 1;
-  repeated Repo to_repos = 2;
+message SquashJobRequest {
+  Job job = 1;
+}
+
+message FlushJobRequest {
```
jdoliner, line 252: We discussed rolling `FlushJob`s semantics in `InspectJob` on the call. Just recording that here I think we were all on board with it.

src/server/pfs/cmds/cmds.go
```diff
@@ -455,46 +454,49 @@ $ {{alias}} foo@XXX bar@YYY
...
-				return err
-			}
-			defer c.Close()
-
-			var toRepos []*pfsclient.Repo
-			for _, repoName := range repos {
-				toRepos = append(toRepos, client.NewRepo(repoName))
-			}
+			// TODO(global ids): implement flush job command
+			return errors.New("unimplemented")
```
jdoliner, line 458: Just adding to a todo there because I don't think we have tests for most of pachctl

src/server/pfs/pfs.go
```diff
@@ -121,6 +127,10 @@ func (e ErrCommitNotFinished) Error() string {
 	return fmt.Sprintf("commit %v not finished", e.Commit.ID)
 }
 
+func (e ErrAmbiguousCommit) Error() string {
+	return fmt.Sprintf("commit %v is ambiguous (specify the commit branch to resolve)", e.Commit.ID)
```
jdoliner, line 131:
```suggestion
	return fmt.Sprintf("commit %v is ambiguous (specify the branch to resolve)", e.Commit.ID)
```
I think the word "commit" is just going to confuse users since the thing they need to specify is just, for example, "master" right which they just thing of as a branch.

src/server/pfs/pretty/pretty.go
```diff
@@ -19,7 +19,7 @@ const (
 	// RepoAuthHeader is the header for repos with auth information attached.
 	RepoAuthHeader = "NAME\tCREATED\tSIZE (MASTER)\tACCESS LEVEL\t\n"
 	// CommitHeader is the header for commits.
-	CommitHeader = "REPO\tBRANCH\tCOMMIT\tFINISHED\tSIZE\tPROGRESS\tDESCRIPTION\n"
+	CommitHeader = "REPO\tBRANCH\tCOMMIT\tFINISHED\tSIZE\tDESCRIPTION\n"
```
jdoliner, line 22: Is it going to be possible to get this progress information from somewhere? It seems like it would just be the sums of the status of the commits within a particular job but I might be oversimplifying.

src/client/pfs.go
Tryneus: `WaitJob` sounds good to me :+1:

src/internal/pfsdb/pfsdb.go
Tryneus: It's an index on the commit without specifying the branch, I'm open to renaming this (it isn't used very often, basically just when resolving a commit from a user-provided `pfs.Commit` that doesn't specify a branch).  Maybe `CommitAnyBranchKey`?  Open to suggestions.

src/pfs/pfs.proto
Tryneus: `BranchProvenance` sounds good.  Like we discussed, this will be changed to specify `DirectProvenance` for each commit instead of the full provenance.  Also, this structure shouldn't really be returned directly to users, we'll return a hydrated structure with `CommitInfos` in it as well (and possibly an expanded `Provenance` graph).

src/pfs/pfs.proto
Tryneus: Like we discussed, the hydrated one will be filled out in `InspectJob` so its fields will be expanded to account for that.  I don't think there is much of an internal use-case for an un-hydrated `JobInfo`.

src/server/pfs/pretty/pretty.go
Tryneus: Yes - the CLI updates will happen in a separate PR, though.

src/server/pfs/server/driver.go
```diff
@@ -898,30 +718,46 @@ func (d *driver) getRepoSize(ctx context.Context, repo *pfs.Repo) (int64, error)
 // 'branches' are newly created (i.e. in CreatePipeline).
 //
 // The isNewCommit flag indicates whether propagateCommits was called during the creation of a new commit.
```
brycemcanally, line 720: This part of the comment should be deleted.

src/pfs/pfs.proto
Tryneus: I ended up using `BranchInfo` and clarifying that we only set the `Branch` and `DirectProvenance` fields - it seems like proto3 is pretty good about making this efficient

src/server/pps/server/api_server.go
```diff
@@ -874,41 +863,46 @@ func (a *apiServer) FlushPipelineJob(request *pps.FlushPipelineJobRequest, resp
...
-		// Even though the commit has been finished the job isn't necessarily
-		// finished yet, so we block on its state as well.
-		pji, err := a.InspectPipelineJob(resp.Context(), &pps.InspectPipelineJobRequest{PipelineJob: pjis[0].PipelineJob, BlockState: true})
-		if err != nil {
-			return err
-		}
-		return resp.Send(pji)
-	})
+
+	// TODO(global ids): implement whatever use case this is trying to solve (maybe pfs.FlushJob can do most of the work?)
```
jdoliner, line 867: With global ids this should be mostly doable with pfs.FlushJob or whatever we call it. The purpose of this function is basically to get back the jobs a commit caused so you can get logs from them / see why they failed.

jdoliner (approved):
A few ongoing conversations about naming going on but other than that this looks good to me.

src/client/pfs.go
brycemcanally: Are we going with Block or Wait? It looks like both verbs are being using currently (BlockCommitset vs the field being Wait).

src/server/pfs/server/driver.go
```diff
@@ -896,32 +716,46 @@ func (d *driver) getRepoSize(ctx context.Context, repo *pfs.Repo) (int64, error)
 // starts downstream output commits (which trigger PPS jobs) when new input
 // commits arrive on 'branch', when 'branches's HEAD is deleted, or when
 // 'branches' are newly created (i.e. in CreatePipeline).
-//
-// The isNewCommit flag indicates whether propagateCommits was called during the creation of a new commit.
-func (d *driver) propagateCommits(sqlTx *sqlx.Tx, job *pfs.Job, branches []*pfs.Branch, isNewCommit bool) error {
-	jobInfo := &pfs.StoredJobInfo{Job: job}
-	jobProvMap := make(map[string]*pfs.Commit)
+func (d *driver) propagateCommits(txnCtx *txncontext.TransactionContext, branches []*pfs.Branch) error {
```
brycemcanally, line 719: It seems like this should be called `propagateBranches`.

src/server/pfs/server/api_server.go
```diff
@@ -226,36 +231,40 @@ func (a *apiServer) ListCommit(request *pfs.ListCommitRequest, respServer pfs.AP
 	})
 }
 
-// SquashCommitInTransaction is identical to SquashCommit except that it can run
+// InspectCommitset implements the protobuf pfs.InspectCommitset RPC
+func (a *apiServer) InspectCommitset(ctx context.Context, request *pfs.InspectCommitsetRequest) (response *pfs.Commitset, retErr error) {
+	func() { a.Log(request, nil, nil, 0) }()
+	defer func(start time.Time) { a.Log(request, response, retErr, time.Since(start)) }(time.Now())
+	commitset, err := a.driver.inspectCommitset(ctx, request.ID, request.Wait)
```
brycemcanally, line 238: You could just return the call here.

src/server/pfs/server/driver.go
```diff
@@ -939,191 +773,122 @@ func (d *driver) propagateCommits(sqlTx *sqlx.Tx, job *pfs.Job, branches []*pfs.
...
+		if subvBI.Head == nil || subvBI.Head.ID != txnCtx.CommitsetID {
+			// This branch has no commit for this Commitset, start a new output commit in 'subvBI.Branch'
+			newCommit := &pfs.Commit{
+				Branch: subvBI.Branch,
+				ID:     txnCtx.CommitsetID,
+			}
+			newCommitInfo := &pfs.CommitInfo{
+				Commit:  newCommit,
+				Origin:  &pfs.CommitOrigin{Kind: pfs.OriginKind_AUTO},
+				Started: types.TimestampNow(),
```
brycemcanally, line 827: I kind of wonder if it would make more sense for the timestamp to be the same for all of the commits in the transaction / commitset?

Tryneus: That sounds reasonable to me, it's something we can set in the transaction context.  I'll do that.

src/client/pfs.go
Tryneus: Good point, renamed the field to `Block`.

PR #6179, opened 2021-05-25
Ensure /tmp exists in the pachd image.

PFedak:
2.x version of #6065, though cache root no longer exists

### DIFF
--- a/etc/pachd/Dockerfile
+++ b/etc/pachd/Dockerfile
@@ -6,6 +6,8 @@ LABEL name="Pachyderm" \
       vendor="Pachyderm"
 
 COPY --from=pachyderm_build /licenses /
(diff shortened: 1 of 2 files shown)


etc/pachd/Dockerfile
```diff
@@ -6,6 +6,7 @@ LABEL name="Pachyderm" \
       vendor="Pachyderm"
 
 COPY --from=pachyderm_build /licenses /
+COPY --from=pachyderm_build /tmp/to-copy /
```
brycemcanally, line 9: It would be good to have a TODO here to clean this up when we are no longer using scratch.

brycemcanally (approved):
lgtm, just add a comment about cleaning this up when we are no longer using scratch.

PR #6181, opened 2021-05-25
Rollback 'pachctl auth activate' partially committed state

acohen4:
`pachctl auth activate` creates a n OIDC Client record against the IDP service, and then creates a OIDC Configuration record that's later used to create the OIDC Login URL. 

Previously, if the OIDC Configuration creation failed, the OIDC Client record would remain in existence, which would break consecutive executions of this command.

As part of this change, we revert the creation of the OIDC Client record in case the OIDC Configuration creation fails. This way the user can rerun the command without receiving errors relating to the OIDC Client ID not being unique.

### DIFF
--- a/src/server/auth/cmds/cmds.go
+++ b/src/server/auth/cmds/cmds.go
@@ -175,7 +175,15 @@ Activate Pachyderm's auth system, and restrict access to existing data to the ro
 						LocalhostIssuer: true,
 						Scopes:          scopes,
 					}}); err != nil {
-					return errors.Wrapf(grpcutil.ScrubGRPC(err), "failed to configure OIDC in pachd")
+					err = errors.Wrapf(grpcutil.ScrubGRPC(err), "failed to configure OIDC in pachd")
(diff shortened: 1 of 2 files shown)


actgardner (approved):

msteffen:
QQ that I had stuck in a comment: Does DeleteOIDCClient() do an auth check?

I wanted to understand how this command works from an auth perspective, and got stuck on: on the one hand, if DeleteOIDCClient() auth checks, then how does that work in a partially-activated state (or do all auth checks succeed if no OIDC config is present)? OTOH, if DeleteOIDCClient() doesn't auth check, can an anonymous malicious actor disable a cluster by deleting its OIDC client (or is the client ID not guessable)?

I'll look through some of the surrounding code, but the question came to mind so I thought I'd post it.

actgardner:
Auth being enabled doesn't depend on OIDC being configured at all in 2.0 - activating auth generates the root token, and then auth is fully enabled. Configuring OIDC and creating/deleting clients does an auth check, and the user should have the root token (because they just ran Activate).

msteffen:
> user should have the root token (because they just ran Activate to generate it)

:facepalm: Got it, thanks! I always forget that there's always a root token, but it simplifies so much

Issue #6161, opened 2021-05-21
config list context should show active-enterprise-context


pappasilenus:
**What is the goal / desired outcome?**

As a Pachyderm user, I would like to see all the information about my active contexts when I list my context so I know to set them correctly.

**If there is a way to accomplish this today via workaround, what does that require?**

The user must explicitly call pachctl config get active-enterprise-context.

**(Optional) What is your proposal for a feature to solve this?**

Add an "E" to indicate the currently active enterprise context:

When they're different:
```
$  pc config list context
ACTIVE	NAME
	aws
	cluster-metrics
	demo
	docker-desktop
	enterprise
E	enterprise-2-16-alpha
*	jk-2-16-alpha
```
When they're the same:
```
$  pc config list context
ACTIVE	NAME
	aws
	cluster-metrics
	demo
	docker-desktop
	enterprise
	enterprise-2-16-alpha
E*	jk-2-16-alpha
```

acohen4:
Hey @pappasilenus , this feature request sounds good to me! 

I haven't gotten others' feedback on it yet, but in the meantime made a corresponding PR to implement this: https://github.com/pachyderm/pachyderm/pull/6184

actgardner:
This should be resolved in the next alpha, closing

PR #6184, opened 2021-05-25
Adding 'E' marker to active enterprise context in 'pachctl config list context'

acohen4:
This PR is a response to feature request: https://github.com/pachyderm/pachyderm/issues/6161

### DIFF
--- a/src/server/config/cmds.go
+++ b/src/server/config/cmds.go
@@ -2,13 +2,17 @@ package cmds
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"log"
 	"os"
 	"sort"
+	"time"
 
+	"github.com/pachyderm/pachyderm/v2/src/client"
+	"github.com/pachyderm/pachyderm/v2/src/enterprise"
 	"github.com/pachyderm/pachyderm/v2/src/internal/cmdutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/config"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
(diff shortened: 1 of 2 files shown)


src/server/config/cmds.go
```diff
@@ -374,13 +374,22 @@ func Cmds() []*cobra.Command {
 				return err
 			}
 
+			activeEnterpriseContext, _, err := cfg.ActiveEnterpriseContext(true)
+			if err != nil {
+				return err
```
msteffen, line 379: UX question: if `errorOnNoActive` is set to `true`, does returning an error here make it impossible to list your contexts if you have no enterprise context set? (does that apply to e.g. open-source users?)

msteffen (commented):
Quick UX question

src/server/config/cmds.go
acohen4: I was also curious about this. `ActiveEnterpriseContext`[ is implemented](https://github.com/pachyderm/pachyderm/blob/master/src/internal/config/config.go#L88) so that if no active-enterprise-context is listed, it will return the active-context. 

The `errorOnNoActive` argument actually gets used when checking for the active-context. So when `errorOnNoActive` is set to `true`, an error will only be returned if neither an `active-enterprise-context` nor an `active-context` are set

msteffen: I see. I mean, would it make sense to still print the overall list of contexts with no `E`, even in the case where there's no active context selected? My worry is that right now, if your active context becomes unset, there would be no way to list the contexts for purposes of selecting a new one

acohen4:
On second look, I actually think I also need to clarify some expected behavior around whether we should present the 'E' marker if enterprise isn't activated. Would it be bad or undesired to show an open-source user the 'E*' markers next to their active context?

Internally, when making enterprise calls without an explicit `active_enterprise_context` in a config, the `active_context` is used instead. This means that the absence of `active_enterprise_context` doesn't mean that enterprise isn't being used. I need to look more through the uses of `active_enterprise_context` to get an idea of whether there may be a simple semantic solution here.

msteffen:
> Would it be bad or undesired to show an open-source user the 'E*' markers next to their active context?

My inuition is that it would be better not to, but I see how that introduces extra complexity, and potentially requires `pachctl config list context` to make an RPC, when I would otherwise expect it to be a client-only command.

My intuition for a UX:
-  If `active_enterprise_context` is set, print an `E` next to it no matter what
-  Otherwise, make an RPC to get the current enterprise status. Print the list of contexts, and if the RPC succeeds *and* indicates that enterprise is active, print `E` next to the `active_context`. Otherwise, don't print an E. In the specific case that the RPC fails/times out, also print an error (e.g. `could not reach cluster`) prior to the list of contexts.

Alternatively, if that's too complicated to be worth implementing:
- simply print an `E` if and only if `active_enterprise_context` is set?

WDYT?

acohen4:
> My intuition:
> 
> * If `active_enterprise_context` is set, print an `E` next to it no matter what
> * Otherwise, make an RPC to get the current enterprise status. Print the list of contexts, and if the RPC succeeds _and_ indicates that enterprise is active, print `E*` next to the `active_context`. Otherwise, don't print an `E` (only `*`). In the specific case that the RPC fails/times out, also print an error (e.g. `could not reach cluster`) prior to the list of contexts.

I actually like this approach a lot. To me, it yielding the most accurate results seems to be worth the small extra cost of complexity and RPC. 

I think my thinking was too boxed into reflecting the enterprise status with the information stored on the local config.

src/server/config/cmds.go
```diff
@@ -374,13 +374,22 @@ func Cmds() []*cobra.Command {
...
+				var activeMarker string
 				if key == activeContext {
-					fmt.Printf("*\t%s\n", key)
-				} else {
-					fmt.Printf("\t%s\n", key)
+					activeMarker = "*"
+				}
+				var activeEnterpriseMarker string
+				if key == activeEnterpriseContext {
+					activeEnterpriseMarker = "E"
```
actgardner, line 390: IMO if the enterprise context == the active context, we should _only_ print the `*`. Otherwise I think it's going to be confusing for users who aren't using enterprise.

actgardner (approved):
LGTM!

PR #6186, opened 2021-05-26
Resume running all PPS tests.

PFedak:
We should probably have a safeguard on this, but I'm not sure what layer it should live at


.circleci/config.yml
```diff
@@ -171,6 +171,8 @@ workflows:
               - PPS2
               - PPS3
               - PPS4
+              - PPS5
```
nitinjainsj, line 174: Did this get removed?

nitinjainsj (approved):

.circleci/config.yml
PFedak: No, the opposite, the bucket number was increased.

I'm attempting to add an automated check, but it might benefit from someone with better bash/sed/awk foo

PR #6196, opened 2021-05-26
Rename GetFile -> GetFileTAR

brendoncarroll:
This is to make room for a backwards compatible `GetFile` which will make client implementations easier (evidenced by python, and javascript clients).
A TAR stream is more efficient, and is used by most of the v2 system.


brycemcanally (approved):
lgtm after `GetTAR` -> `GetFileTAR`

PR #6185, opened 2021-05-25
Put stats in a separate "meta" system repo

PFedak:

PR #6198, opened 2021-05-26
Create per-pipeline spec repos

PFedak:


src/server/pps/server/api_server.go
```diff
@@ -1623,17 +1622,6 @@ func (a *apiServer) sudo(ctx context.Context, f func(*client.APIClient) error) e
 	return f(superUserClient)
 }
 
-// sudoTransaction is a convenience wrapper around sudo for api calls in a transaction
-func (a *apiServer) sudoTransaction(txnCtx *txncontext.TransactionContext, f func(*txncontext.TransactionContext) error) error {
```
PFedak, line 1627: there's still a use of `sudo` in the pipeline controller for `getPipelineInfo`, I haven't actually checked if that can be removed, too

src/internal/ppsutil/util.go
```diff
@@ -117,6 +117,9 @@ func GetLimitsResourceList(limits *pps.ResourceSpec) (*v1.ResourceList, error) {
 func GetPipelineInfoAllowIncomplete(pachClient *client.APIClient, ptr *pps.StoredPipelineInfo) (*pps.PipelineInfo, error) {
 	result := &pps.PipelineInfo{}
 	buf := bytes.Buffer{}
+	// ensure we are authorized to read the pipeline's spec commit, but don't propagate that back out
+	pachClient = pachClient.WithCtx(pachClient.Ctx())
+	pachClient.SetAuthToken(ptr.AuthToken)
```
PFedak, line 122: previously, the (singular) spec repo gave everyone read permissions
While this grants more permissions, it seems like limiting the scope to this function call eliminates any issues (especially considering the caller has the pipeline info with the auth token, too)

src/server/pps/server/api_server.go
```diff
@@ -2119,6 +2059,24 @@ func (a *apiServer) CreatePipelineInTransaction(
 				"running, you can run 'pachctl update pipeline --clean' which will " +
 				"delete this open commit")
 		}
+	} else {
+		// Create output and spec repos
+		if err := a.env.PfsServer().CreateRepoInTransaction(txnCtx,
+			&pfs.CreateRepoRequest{
+				Repo:        client.NewRepo(pipelineName),
+				Description: fmt.Sprintf("Output repo for pipeline %s.", request.Pipeline.Name),
+				Update:      true,
```
PFedak, line 2068: the "Update" in these is identical to what was done before by ignoring AlreadyExists errors, but perhaps the former makes the intent more clear?

src/server/pps/server/api_server.go
```diff
@@ -2641,10 +2590,18 @@ func (a *apiServer) listPipelinePtr(ctx context.Context,
...
-			ci, err := pachClient.InspectCommit(ppsconsts.SpecRepo, p.SpecCommit.Branch.Name, p.SpecCommit.ID)
+			// allow reading spec commit regardless of calling user
+			pachClient.SetAuthToken(p.AuthToken)
+			ci, err := pachClient.PfsAPIClient.InspectCommit(
+				pachClient.Ctx(),
+				&pfs.InspectCommitRequest{
+					Commit:     p.SpecCommit,
+					BlockState: pfs.CommitState_STARTED,
+				},
+			)
```
PFedak, line 2601: I plan to switch `InspectCommit` and some others to just take a commit in a future PR, to eliminate both the verbosity and the potential for error by not scrubbing the grpc error (which also adds a stack trace) or losing auth info from the context

src/auth/auth.go
```diff
@@ -47,12 +47,6 @@ const (
 	// be removed from the set of cluster super-admins.
 	RootUser = "pach:root"
 
-	// PpsUser is a special, unrevokable cluster administrator account used by PPS
-	// to create pipeline tokens, close commits, and do other necessary PPS work.
-	// It's not possible to authenticate as ppsUser (pps reads the auth token for
-	// this user directly from etcd).
-	PpsUser = `pach:pps`
-
```
PFedak, line 55: My understanding is we wanted to move away from this, but I haven't kept up to date

src/server/pps/server/api_server.go
```diff
@@ -2641,9 +2590,21 @@ func (a *apiServer) listPipelinePtr(ctx context.Context,
...
 			switch {
 			case err != nil:
+				err = grpcutil.ScrubGRPC(err)
+				if auth.IsErrNoRoleBinding(err) {
+					return nil // a missing role binding suggests the spec repo is gone
+				}
```
PFedak, line 2607: This doesn't seem like a great way to address this, I guess I'd like `InspectCommit` to give `ErrRepoNotFound` in this case, rather than failing in auth, but I'm not sure if that should be a change in `Authorize` itself or special case logic for commits.

src/auth/auth.go
actgardner: Correct. There are three situations where we use the pps superuser today:
- When a user makes an RPC, we use the superuser to write to the spec repo - this should fix that case.
- Background tasks in pachd use the super-user token because there's no user identity. They can use internal APIs or use the pipeline's token from etcd.
- Pipelines making calls to the sidecar use the pipeline's token.

src/internal/ppsutil/util.go
actgardner: This is a fine solution for now. AFAICT these methods are used in pachd and the worker, but we really want two different implementations - the worker version should accept a client with the PPS token, but the PPS version shouldn't. We can clean that up in the future.

src/internal/serviceenv/config.go
```diff
@@ -30,12 +30,13 @@ type GlobalConfiguration struct {
 	PostgresServiceSSL  string `env:"POSTGRES_SERVICE_SSL,default=disable"`
 	PostgresDBName      string `env:"POSTGRES_DATABASE_NAME"`
 
-	// PPSSpecCommitID is only set for workers and sidecar pachd instances.
-	// Because both pachd and worker need to know the spec commit (the worker so
-	// that it can avoid jobs for other versions of the same pipelines and the
-	// sidecar so that it can serve the S3 gateway) it's stored in the
+	// PPSSpecCommitID and PPSPipelineName are only set for workers and sidecar
```
actgardner, line 33: I'd love to do a follow-up to clean up these structs - I think this comment correctly captures the categories of config (global, pachd, worker, sidecar) but I don't think we've actually organized things that way?

src/server/auth/server/api_server.go
```diff
@@ -81,10 +79,6 @@ type apiServer struct {
 	// oidcStates  contains the set of OIDC nonces for requests that are in progress
 	oidcStates col.EtcdCollection
 
-	// This is a cache of the PPS master token. It's set once on startup and then
-	// never updated
-	ppsToken string
```
actgardner, line 86: 👏 👏 👏

src/server/pachyderm_test.go
```diff
@@ -10437,50 +10456,6 @@ func TestInterruptedUpdatePipelineInTransaction(t *testing.T) {
 	require.Matches(t, "outside of transaction", err.Error())
 }
 
-func TestPipelineSpecCommitCleanup(t *testing.T) {
```
actgardner, line 10440: Is there a new corresponding test that makes sure we don't leave dangling spec repos? I might have missed it.

src/server/pfs/server/driver.go
```diff
@@ -334,6 +319,31 @@ func (d *driver) listRepo(ctx context.Context, includeAuth bool, repoType string
...
+				return errors.Wrapf(err, "error inspecting branch %s", pretty.CompactPrintBranch(branch))
+			}
+			branchInfos = append(branchInfos, bi)
+		}
+	}
+	// sort ascending provenance
+	sort.Slice(branchInfos, func(i, j int) bool { return len(branchInfos[i].Provenance) < len(branchInfos[j].Provenance) })
+	for i := range branchInfos {
+		// delete branches from most provenance to least, that way if one
+		// branch is provenant on another (such as with stats branches) we
```
actgardner, line 337: This is just my lack of understanding but we're moving stats to meta repos too right? 🤞

src/server/pps/server/api_server.go
actgardner: FWIW we also want to move away from nested GRPC calls, so maybe we can make this a method call and not an RPC?

src/server/pps/server/api_server.go
actgardner: I'll make an issue to clean this up - it's not a good user experience anywhere.

actgardner (approved):
I'm +1 on this but I don't have the most context on the PFS portion.

src/server/pps/server/api_server.go
PFedak: ... except that this clobbers any new description that a user might have set, so I'll revert this

src/server/pps/server/api_server.go
PFedak: Oh, of course, whoops

src/internal/ppsutil/util.go
PFedak: I'm not sure I fully understand the distinction you're making, but sounds good

src/internal/serviceenv/config.go
PFedak: Given how spare the worker-specific struct is at this point, I wonder if it should just be called "PipelineSpecific", reclaim these variables, and get included in the sidecar and worker

src/server/pachyderm_test.go
PFedak: Ah, that's a good point. I should come up with a good test for the system repo "cluster" deletion behavior, which I think is the equivalent guarantee now (no spec repo without its pipeline repo)

src/server/pfs/server/driver.go
PFedak: already done in #6185 😏, that was definitely a concern. That change definitely gets rid of the common instances of inter-branch provenance, but I'm not confident there are others (also depends if we nix build pipelines for now)

src/server/pps/server/api_server.go
PFedak: nevermind

src/server/pfs/cmds/cmds.go
```diff
@@ -634,6 +633,13 @@ Any pachctl command that can take a Commit ID, can take a branch name instead.`,
 				trigger.Branch == "" {
 				return errors.Errorf("trigger condition specified without a branch to trigger on, specify a branch with --trigger")
 			}
+			if trigger.Branch == "" {
+				trigger = nil
+			}
+			var headCommit *pfs.Commit
+			if head != "" {
+				headCommit = branch.Repo.NewCommit("", head)
+			}
```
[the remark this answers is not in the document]
PFedak: I guess I should have added a pfs/cmds test for it, surprised that doesn't include `create branch`

src/server/pfs/server/driver.go
PFedak: I've amended the comment to better reflect why it's needed

PR #6211, opened 2021-05-27
Fixes TestLokiLogs.

jdoliner:
This fixes the Loki Logs test, the issue seemed to be that after receiving all of the logs in follow mode not all of the logs were returned from a non-follow request. I'm guessing that there's a bit of a lag between when Loki returns them and when they get indexed. Unfortunately the best way to fix it seemed to be a sleep, I tried doing it with a `NoErrorUntil` and that proved to be much less stable. In local testing this version of the code succeeded every time I ran it.

### DIFF
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -4016,7 +4016,6 @@ func TestManyLogs(t *testing.T) {
 
 func TestLokiLogs(t *testing.T) {
 	// TODO(2.0 required): This test is taking too long to complete due to slow pipelines
-	t.Skip("Skipping flaky test")
 	if testing.Short() {
 		t.Skip("Skipping integration tests in short mode")
 	}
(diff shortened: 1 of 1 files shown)


src/server/pachyderm_test.go
```diff
@@ -4016,7 +4016,6 @@ func TestManyLogs(t *testing.T) {
 
 func TestLokiLogs(t *testing.T) {
 	// TODO(2.0 required): This test is taking too long to complete due to slow pipelines
```
brycemcanally, line 4018: We can probably remove this if it is no longer flaky and we suspect the issue was a delay in Loki.

brycemcanally (approved):
lgtm, just a minor suggestion.

src/server/pachyderm_test.go
jdoliner: Removed, the test took an average of ~14.5s to run on my machine which seems pretty reasonable. I suspect a big part of the slowness was the fact that the test had a 30 second sleep at the end for no particular reason.

PR #6213, opened 2021-05-28
Add file path validation

brycemcanally:
This PR adds validation for the characters we will support in file paths for 2.0. The characters should be within the printable ASCII range, excluding glob characters (as well as some extended glob characters). We will likely loosen the restrictions over time when we have more time to think through the implications of extending the valid character set.


brendoncarroll (approved):
We discussed WithValidator being an option on Writer vs UnorderedWriter offline.  It's fine how it is.  Possibility of removing Writer from public API, but no consensus there yet.

LGTM.

PR #6220, opened 2021-05-31
Remove self-RPCs from auth inteceptor

actgardner:
Remove the self-RPCs in the auth interceptor and when checking enterprise license status. This should reduce latency.

### DIFF
--- a/src/server/pfs/server/driver_fsck.go
+++ b/src/server/pfs/server/driver_fsck.go
@@ -9,10 +9,7 @@ import (
 	"github.com/gogo/protobuf/proto"
 	"github.com/jmoiron/sqlx"
 
-	"github.com/pachyderm/pachyderm/v2/src/auth"
 	col "github.com/pachyderm/pachyderm/v2/src/internal/collection"
-	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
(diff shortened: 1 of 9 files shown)


src/server/transaction/server/api_server.go
```diff
@@ -31,7 +31,6 @@ func newAPIServer(
 		Logger: log.NewLogger("transaction.API", env.Logger()),
 		driver: d,
 	}
-	go func() { env.GetPachClient(context.Background()) }() // Begin dialing connection on startup
```
actgardner, line 34: Remove a pointless eager initialization of the pach client.

src/server/pfs/server/driver_fsck.go
```diff
@@ -235,15 +232,6 @@ func (e ErrProvenanceOfSubvenance) Error() string {
 // 4. Commit provenance and commit subvenance are dual relations
 // If fix is true it will attempt to fix as many of these issues as it can.
 func (d *driver) fsck(ctx context.Context, fix bool, cb func(*pfs.FsckResponse) error) error {
-	// Check that the user is logged in (user doesn't need any access level to
```
actgardner, line 238: This auth check was redundant, we check that the user is authorized in the inteceptor.

acohen4 (approved):
Awesome!

PR #6223, opened 2021-06-01
Filtering out localhost context in sync-contexts command

acohen4:
It's a little unclear to me whether we want to provide some of the info in ListUserClusters in the ListClusters call. 

Does it make sense for us to filter the localhost record out of ListUserClusters not to confused the user synching contexts, but leave that record in `license list-clusters` for admins to see?

### DIFF
--- a/src/internal/auth/interceptor.go
+++ b/src/internal/auth/interceptor.go
@@ -112,7 +112,7 @@ var authHandlers = map[string]authHandler{
 	"/license_v2.API/DeleteAll":         authDisabledOr(clusterPermissions(auth.Permission_CLUSTER_DELETE_ALL)),
 	// Heartbeat relies on the shared secret generated at cluster registration-time
 	"/license_v2.API/Heartbeat":        unauthenticated,
-	"/license_v2.API/ListUserClusters": authenticated,
+	"/license_v2.API/ListUserClusters": authDisabledOr(authenticated),
 
 	//
 	// PFS API
--- a/src/server/enterprise/testing/cmd_test.go
+++ b/src/server/enterprise/testing/cmd_test.go
@@ -311,6 +311,10 @@ func TestSyncContexts(t *testing.T) {
 		"clusterId", newClusterId,
 		"userAddress", "grpc://pachd.default:700",
 	).Run())
+
+	// make sure that the cluster with id = 'localhost' does not get synched.
+	// it's used by the License Server to identify itself
+	require.YesError(t, tu.BashCmd(`pachctl config list context | match localhost`).Run())
 }
 
 // Tests RegisterCluster command's derived argument values if not provided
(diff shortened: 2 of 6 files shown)


src/internal/auth/interceptor.go
```diff
@@ -112,7 +112,7 @@ var authHandlers = map[string]authHandler{
 	"/license.API/DeleteAll":         authDisabledOr(clusterPermissions(auth.Permission_CLUSTER_DELETE_ALL)),
 	// Heartbeat relies on the shared secret generated at cluster registration-time
 	"/license.API/Heartbeat":        unauthenticated,
-	"/license.API/ListUserClusters": authenticated,
+	"/license.API/ListUserClusters": authDisabledOr(authenticated),
```
actgardner, line 115: 👍

src/server/enterprise/testing/cmd_test.go
```diff
@@ -311,6 +311,10 @@ func TestSyncContexts(t *testing.T) {
 		"clusterId", newClusterId,
 		"userAddress", "grpc://pachd.default:700",
 	).Run())
+
+	// make sure that the cluster with id = 'localhost' does not get synched.
```
actgardner, line 315: I think the metric we care about for filtering is actually `EnterpriseServer` being true? The idea is that an org has one enterprise server, and that's where we're getting the info from?

acohen4: thanks! that makes sense. I wasn't on the same page for the semantics of that field

src/server/license/server/api_server.go
```diff
@@ -383,7 +383,7 @@ func (a *apiServer) ListUserClusters(ctx context.Context, req *lc.ListUserCluste
 	a.LogReq(req)
 	defer func(start time.Time) { a.pachLogger.Log(req, resp, retErr, time.Since(start)) }(time.Now())
 	clusters := make([]*lc.UserClusterInfo, 0)
-	if err := a.env.GetDBClient().SelectContext(ctx, &clusters, `SELECT id, cluster_deployment_id, user_address, is_enterprise_server FROM license.clusters`); err != nil {
+	if err := a.env.GetDBClient().SelectContext(ctx, &clusters, `SELECT id, cluster_deployment_id, user_address, is_enterprise_server FROM license.clusters WHERE id != 'localhost'`); err != nil {
```
actgardner, line 386: NB: this is also affecting `ListCluster`, which is why the test is failing. I think we only want this to affect syncing user contexts.

src/server/license/cmds/cmds.go
```diff
@@ -77,20 +64,14 @@ func ActivateCmd() *cobra.Command {
...
-
 			// Register the localhost as a cluster
 			resp, err := c.License.AddCluster(c.Ctx(),
 				&license.AddClusterRequest{
 					Id:                  "localhost",
 					Address:             "grpc://localhost:653",
 					UserAddress:         "grpc://localhost:653",
 					ClusterDeploymentId: clusterInfo.DeploymentID,
-					EnterpriseServer:    enterpriseServer,
+					EnterpriseServer:    true,
```
acohen4, line 74: note: whatever cluster the license is activated on should note itself as an enterprise server

actgardner (approved):

PR #6226, opened 2021-06-01
Create default empty branch head commits

Tryneus:
This simplifies a lot of handling in PFS if we can assume that there is always a branch head.  In CreateBranch and SquashCommit, we ensure that a branch head always exists.

### DIFF
--- a/src/server/pfs/cmds/cmds.go
+++ b/src/server/pfs/cmds/cmds.go
@@ -15,7 +15,11 @@ import (
 
 	prompt "github.com/c-bata/go-prompt"
 	"github.com/gogo/protobuf/jsonpb"
+	"github.com/gogo/protobuf/proto"
 	"github.com/mattn/go-isatty"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
 	"github.com/pachyderm/pachyderm/v2/src/client"
 	"github.com/pachyderm/pachyderm/v2/src/internal/cmdutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
@@ -30,9 +34,6 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/server/cmd/pachctl/shell"
 	"github.com/pachyderm/pachyderm/v2/src/server/pfs/pretty"
 	txncmds "github.com/pachyderm/pachyderm/v2/src/server/transaction/cmds"
-
-	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
 )
 
 const (
@@ -669,9 +670,11 @@ Any pachctl command that can take a Commit ID, can take a branch name instead.`,
 			if len(provenance) != 0 && trigger.Branch != "" {
 				return errors.Errorf("cannot use provenance and triggers on the same branch")
 			}
-			if (trigger.CronSpec != "" || trigger.Size_ != "" || trigger.Commits != 0) &&
-				trigger.Branch == "" {
+			if (trigger.CronSpec != "" || trigger.Size_ != "" || trigger.Commits != 0) && trigger.Branch == "" {
 				return errors.Errorf("trigger condition specified without a branch to trigger on, specify a branch with --trigger")
+			}
+			if proto.Equal(trigger, &pfs.Trigger{}) {
+				trigger = nil
 			}
 			c, err := client.NewOnUserMachine("user")
 			if err != nil {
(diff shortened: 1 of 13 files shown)


src/server/pfs/cmds/cmds.go
```diff
@@ -669,21 +670,35 @@ Any pachctl command that can take a Commit ID, can take a branch name instead.`,
...
+						return err
+					}
+				}
+			}
+
 			return txncmds.WithActiveTransaction(c, func(c *client.APIClient) error {
 				_, err := c.PfsAPIClient.CreateBranch(
 					c.Ctx(),
 					&pfs.CreateBranchRequest{
-						Head:       branch.Repo.NewCommit("", head),
```
Tryneus, line 686: This was broken - it was specifying a head commit even if we weren't updating the branch head.

src/server/pfs/server/driver.go
```diff
@@ -671,23 +684,6 @@ func (d *driver) aliasCommit(txnCtx *txncontext.TransactionContext, parent *pfs.
 	return commitInfo, nil
 }
 
-// writeFinishedCommit writes these changes to etcd:
-// 1) it closes the input commit (i.e., it writes any changes made to it and
-//    removes it from the open commits)
-// 2) if the commit is the new HEAD of master, it updates the repo size
-func (d *driver) writeFinishedCommit(sqlTx *sqlx.Tx, commitInfo *pfs.CommitInfo) error {
```
Tryneus, line 678: this function was obsolete (we no longer track the repo size in the repo info)

src/server/pfs/server/driver.go
```diff
@@ -580,8 +580,21 @@ func (d *driver) finishCommit(txnCtx *txncontext.TransactionContext, commit *pfs
 	if description != "" {
 		commitInfo.Description = description
 	}
+	// Check that the parent commit is not still open (which can happen in output branches)
+	if commitInfo.ParentCommit != nil {
```
Tryneus, line 584: Unrelated fix for finishing a commit on an unfinished parent which should never happen.

src/server/pfs/pfs.go
```diff
@@ -289,3 +283,12 @@ func IsCommitOnOutputBranchErr(err error) bool {
 	}
 	return commitOnOutputBranchRe.MatchString(err.Error())
 }
+
+// IsParentNotFinishedErr returns true if the err is due to an attempt to
```
brycemcanally, line 287: I don't think this comment is correct.

src/server/pfs/server/driver.go
```diff
@@ -791,17 +787,14 @@ func (d *driver) propagateBranches(txnCtx *txncontext.TransactionContext, branch
 			if err != nil {
 				return err
 			}
-			if provOfSubvBI.Head == nil || provOfSubvBI.Branch.Repo.Type == pfs.SpecRepoType {
```
brycemcanally, line 794: What was the reason for filtering out the spec repo originally?

src/server/pfs/server/driver.go
```diff
@@ -791,17 +787,14 @@ func (d *driver) propagateBranches(txnCtx *txncontext.TransactionContext, branch
 			if err != nil {
```
brycemcanally, line 787: This comment is intended for the `propagateBranches` function comment (I guess you cannot comment on collapsed lines). We should probably remove the reference to a nil HEAD.

src/server/pfs/server/driver.go
```diff
@@ -791,17 +787,14 @@ func (d *driver) propagateBranches(txnCtx *txncontext.TransactionContext, branch
...
 				needsCommit = true
 				break
 			}
 		}
 
 		// If there are no upstream commits for this Commitset and no commit in this branch, we can skip
-		if !needsCommit && (subvBI.Head == nil || subvBI.Head.ID != txnCtx.CommitsetID) {
+		if !needsCommit && subvBI.Head.ID != txnCtx.CommitsetID {
 			continue
 		}
```
brycemcanally, line 800: There is a nil HEAD check in the provenance iteration below.

src/server/pfs/server/driver.go
```diff
@@ -1915,26 +1911,30 @@ func isNotFoundErr(err error) bool {
...
 	if pfsdb.BranchKey(provBranch) == pfsdb.BranchKey(branchInfo.Branch) {
 		return errors.Errorf("provenance loop, branch %s cannot be provenant on itself", pfsdb.BranchKey(provBranch))
 	}
 	add(&branchInfo.Provenance, provBranch)
 	provBranchInfo := &pfs.BranchInfo{}
-	if err := d.branches.ReadWrite(sqlTx).Upsert(pfsdb.BranchKey(provBranch), provBranchInfo, func() error {
-		// Set provBranch, we may be creating this branch for the first time
-		provBranchInfo.Branch = provBranch
+	if err := d.branches.ReadWrite(txnCtx.SqlTx).Upsert(pfsdb.BranchKey(provBranch), provBranchInfo, func() error {
+		if provBranchInfo.Branch == nil {
```
brycemcanally, line 1921: Does this mean that branches can be created through the provenance parameter in create branch?

src/server/pfs/server/driver.go
Tryneus: The idea is that it doesn't contribute datums, so we would only create an output commit (and thus a job) based on input commits.  Currently in global IDs, we will create an initial empty job (unless the pipeline has input datums on creation).

src/server/pfs/server/driver.go
Tryneus: Yes - it looks like before it would lazily create provenance on a branch that didn't yet exist, but now we eagerly instantiate it.

src/server/pfs/server/driver.go
brycemcanally: Wouldn't you need a new output commit for the reprocessing case?

src/server/pfs/server/driver.go
brycemcanally: I guess this is fine, but it seems a bit sketchy to create the provenant branches if they don't already exist. If the prior code was not creating the branches and not erroring, then this is certainly preferable.

brycemcanally (approved):
lgtm

PR #6227, opened 2021-06-01
PPS updates for global IDs

Tryneus:
This PR fixes PPS following the PFS changes for global IDs such that a simple pipeline can run (tested with the OpenCV example).  The main part is adding the `pps.PropagateJobs` call which follows a new commitset, creating any needed PPS jobs.  On the worker side, we subscribe to jobs rather than output commits, so a new RPC was needed for that.


src/client/pps.go
```diff
@@ -202,21 +202,6 @@ func NewPipeline(pipelineName string) *pps.Pipeline {
 	return &pps.Pipeline{Name: pipelineName}
 }
 
-// CreatePipelineJob creates and runs a job in PPS.
-// This function is mostly useful internally, users should generally run work
-// by creating pipelines as well.
-func (c APIClient) CreatePipelineJob(pipeline string, outputCommit, statsCommit *pfs.Commit) (*pps.PipelineJob, error) {
```
Tryneus, line 208: CreatePipelineJob has completely gone away since jobs are now solely created in the transaction-deferred `pps.PropagateJobs`.

src/internal/collection/postgres_collection.go
```diff
@@ -214,6 +214,31 @@ func (c *postgresReadWriteCollection) GetByIndex(index *Index, indexVal string,
 	return c.getByIndex(context.Background(), c.tx, index, indexVal, val, opts, true, f)
 }
 
+func (c *postgresCollection) getUniqueByIndex(ctx context.Context, q sqlx.ExtContext, index *Index, indexVal string, val proto.Message) error {
```
Tryneus, line 217: Added a convenience function to collections to get a unique row by a secondary index.  If this finds more than one row, an error is returned.

src/internal/ppsdb/ppsdb.go
```diff
@@ -45,7 +48,20 @@ var PipelineJobsOutputIndex = &col.Index{
 	},
 }
 
-var pipelineJobsIndexes = []*col.Index{PipelineJobsPipelineIndex, PipelineJobsOutputIndex}
+func PipelineJobTerminalKey(pipeline *pps.Pipeline, isTerminal bool) string {
+	return fmt.Sprintf("%s_%v", pipeline.Name, isTerminal)
+}
+
+// PipelineJobsOutputIndex maps job outputs to the PipelineJob that create them.
+var PipelineJobsTerminalIndex = &col.Index{
```
Tryneus, line 56: Add a secondary index for pipeline jobs which sorts them by `[pipeline, isTerminal]`, so that workers can easily subscribe for unfinished jobs.

src/pps/pps.proto
```diff
@@ -59,12 +59,13 @@ message PipelineJob {
 }
 
 enum PipelineJobState {
-  JOB_STARTING = 0;
-  JOB_RUNNING = 1;
-  JOB_FAILURE = 2;
-  JOB_SUCCESS = 3;
-  JOB_KILLED = 4;
-  JOB_EGRESSING = 5;
+  JOB_CREATED = 0;
```
Tryneus, line 62: Added a new `JOB_CREATED` job state for when a job has been created by the originating transaction, but before it has been started by a worker.

src/pfs/pfs.proto
```diff
@@ -93,6 +93,7 @@ enum OriginKind {
   USER = 0;
   AUTO = 1;
   FSCK = 2;
+  ALIAS = 3;
```
Tryneus, line 96: Added a separate commit type for alias commits so that we can avoid creating jobs on alias commits.

src/pps/pps.proto
```diff
@@ -529,13 +510,14 @@ message UpdatePipelineJobStateRequest {
...
   PipelineJobState state = 2;
   string reason = 3;
-  uint64 restart = 4;
-  int64 data_processed = 5;
-  int64 data_skipped = 6;
-  int64 data_failed = 7;
-  int64 data_recovered = 8;
-  int64 data_total = 9;
-  ProcessStats stats = 10;
+  google.protobuf.Timestamp started = 4;
```
Tryneus, line 513: Workers now need to update the job start time when it moves from `CREATED` -> `STARTED`.

src/server/pfs/cmds/cmds.go
```diff
@@ -431,13 +431,18 @@ $ {{alias}} foo@master --from XXX`,
 				fromCommit = branch.Repo.NewCommit("", from)
 			}
 
+			var toCommit *pfs.Commit
```
Tryneus, line 434: This was wrong (probably from when we changed the pfs client interface).

src/server/worker/pipeline/transform/transform.go
```diff
@@ -16,25 +15,13 @@ func Run(driver driver.Driver, logger logs.TaggedLogger) error {
...
-			// we are not guaranteed to hit this code path after a deletion.
-			if pfsserver.IsCommitFinishedErr(err) || pfsserver.IsCommitNotFoundErr(err) || pfsserver.IsCommitDeletedErr(err) || ppsserver.IsPipelineJobFinishedErr(err) {
-				return nil
-			}
-			return err
+	return driver.PachClient().SubscribePipelineJob(
+		driver.PipelineInfo().Pipeline.Name,
+		true,
+		func(pipelineJobInfo *pps.PipelineJobInfo) error {
+			// TODO: check that this is for the right version of the pipeline
```
Tryneus, line 23: Pipeline version updates will be in a follow-up PR.

src/internal/ppsutil/util.go
```diff
@@ -410,7 +411,7 @@ func WriteJobInfo(pachClient *client.APIClient, pipelineJobInfo *pps.PipelineJob
 	return err
 }
 
-func StatsCommit(commit *pfs.Commit) *pfs.Commit {
+func MetaCommit(commit *pfs.Commit) *pfs.Commit {
```
PFedak, line 414: I've been meaning to ask, are we intending to just drop the word "stats" everywhere in favor of "meta"?

src/server/pps/server/transaction_defer.go
```diff
@@ -0,0 +1,47 @@
+package server
+
+import (
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/transactionenv/txncontext"
+	"github.com/pachyderm/pachyderm/v2/src/pfs"
+)
+
+// Propagater is an object that is used to propagate PFS branches at the end of
```
PFedak, line 9: update comment (maybe struct name, too - JobStarter or something?)

src/server/worker/pipeline/service/service.go
```diff
@@ -74,60 +70,30 @@ func Run(driver driver.Driver, logger logs.TaggedLogger) error {
...
-		func(ci *pfs.CommitInfo) error {
-			if cancel != nil {
-				logger.Logf("canceling previous service, new commit ready")
-				cancel()
-				if err := eg.Wait(); err != nil && !errors.Is(err, context.Canceled) {
-					return err
-				}
+	return pachClient.SubscribePipelineJob(pipelineInfo.Pipeline.Name, true, func(pji *pps.PipelineJobInfo) error {
+		if cancel != nil {
+			logger.Logf("canceling previous service, new commit ready")
```
PFedak, line 83: s/commit/job, maybe?

src/server/worker/pipeline/service/service.go
```diff
@@ -74,60 +70,30 @@ func Run(driver driver.Driver, logger logs.TaggedLogger) error {
...
-		// get PipelineJobInfo to look up spec commit, pipeline version, etc (if this
-		// worker is stale and about to be killed, the new job may have a newer
-		// pipeline version than the master. Or if the commit is stale, it may
-		// have an older pipeline version than the master)
-		return pachClient.InspectPipelineJob(pipelineJob.ID, false)
-	}
-	// Get latest pipeline job state.
-	logger.Logf("found existing pipeline job %q for output commit %q", pipelineJobInfos[0].PipelineJob.ID, commit.ID)
-	return pachClient.InspectPipelineJob(pipelineJobInfos[0].PipelineJob.ID, false)
+		logger.Logf("starting new service, commit: %s", pji.OutputCommit.ID)
```
PFedak, line 89: not exactly sure who would be looking for this log line - would including the job ID be useful?

src/server/worker/pipeline/transform/registry.go
```diff
@@ -632,14 +616,9 @@ func (reg *registry) processJobEgressing(ppj *pendingPipelineJob) error {
 func failedInputs(pachClient *client.APIClient, pipelineJobInfo *pps.PipelineJobInfo) ([]string, error) {
 	var failed []string
 	blockCommit := func(name string, commit *pfs.Commit) error {
-		ci, err := pachClient.PfsAPIClient.InspectCommit(pachClient.Ctx(),
-			&pfs.InspectCommitRequest{
-				Commit:     commit,
-				BlockState: pfs.CommitState_FINISHED,
-			})
+		ci, err := pachClient.BlockCommit(commit.Branch.Repo.Name, commit.Branch.Name, commit.ID)
```
PFedak, line 619: I guess you could make `blockCommit` take strings, since that's what the input has, anyway (unless you're anticipating me changing `BlockCommit` to take a commit proto)

src/server/pachyderm_test.go
```diff
@@ -95,7 +95,9 @@ func TestSimplePipeline(t *testing.T) {
 		false,
 	))
 
-	commitset, err := c.BlockCommitset(commit1.ID)
+	commitInfo, err := c.InspectCommit(pipeline, "master", "")
+	require.NoError(t, err)
+	commitset, err := c.BlockCommitset(commitInfo.Commit.ID)
```
PFedak, line 100: A little late now, I guess, but do you think it's worth making a convenience wrapper for this pattern?
Not sure how representative our tests are of usage of our API

PFedak (approved):
LGTM, a couple of suggested string changes. I would like a way to use multicolumn indices in the collections library if it would be useful elsewhere, but not worth it yet.

I haven't fully familiarized myself with global IDs/aliases yet, so I'm kind of assuming that part works (based on tests) and just looking at the new stuff

src/server/pachyderm_test.go
```diff
@@ -3514,15 +3517,18 @@ func TestChainedPipelines(t *testing.T) {
 		"",
 		false,
 	))
-	commitInfos, err := c.FlushJobAll([]*pfs.Commit{aCommit, dCommit}, nil)
+
+	commitInfo, err := c.InspectCommit(cPipeline, "master", "")
```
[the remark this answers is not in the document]
Tryneus: Yes, the old code was intersecting the subvenance of those two commits, which isn't an operation we support anymore.

src/internal/ppsutil/util.go
Tryneus: Yes, it's in the low-priority column of the [global ids project](https://github.com/pachyderm/pachyderm/projects/14).

src/server/pps/server/transaction_defer.go
Tryneus: The action has been renamed to `PropagateJobs` - `JobStarter` doesn't really work because it just creates the jobs rather than starting them (it's the worker that ends up doing that), so maybe `JobCreator` could work...

src/server/worker/pipeline/service/service.go
Tryneus: I mean, the job ID is the same as the output commit ID, but mentioning that it's the job ID is probably more relevant to this code so I'll update the message.

src/server/worker/pipeline/transform/registry.go
Tryneus: I am planning to do a pass over all the methods in `client/pfs.go` and `client.pps.go` to make them all take `Repo`, `Branch`, and `Commit` protos, but after this lands in `master`.

src/server/pachyderm_test.go
Tryneus: I think there are a few options here that we could discuss as a team - if you just want the blocking behavior then `BlockCommit` is pretty straightforwardly what you want.  If you want the full commitset, then you need to know the ID of the commitset (but I suppose we could allow an optional `pfs.Commit` in the request to resolve to get the commitset ID).  Another thing I've been thinking about is having any RPC which can change the DAG return a commitset ID (like `StartCommit` does) - this would affect things like `CreatePipeline`, `CreateBranch`, and `SquashCommit` so you would already have the commitset ID in the client.

src/internal/ppsdb/ppsdb.go
Tryneus: So this isn't actually the same as the `multi` option we previously had in collections (which was the ability to have a row represented in an index multiple times based on an array field), this is more of a 'compound' or 'composite' index, which we are already using for the primary keys of commits and jobs.

PFedak: Oh, sorry, I meant compound indices, I wasn't aware of `multi` at all. I forgot that a bunch of our primary keys could also work that way, which could probably save us an index on those

PR #6228, opened 2021-06-01
Remove `Commitset` collection in global IDs implementation

Tryneus:
Turns out we don't really need a commitset collection if we store the `DirectProvenance` in the `CommitInfo` protobuf.  This PR adds that and reworks commitset code accordingly.  Blocking on a commitset now just returns a stream of `CommitInfo`s since all the data is inlined into that structure.


src/internal/testpachd/mock_transaction.go
```diff
@@ -9,6 +9,16 @@ import (
 )
 
 // This code can all go away if we ever get the ability to run a PPS server without external dependencies
+type newPropagaterFunc func(*txncontext.TransactionContext) txncontext.PpsPropagater
```
Tryneus, line 12: Some test code that wasn't updated in #6227

src/internal/transactionenv/txncontext/context.go
```diff
@@ -30,11 +30,11 @@ type TransactionContext struct {
...
 
-// PropagateCommitset saves a commitset to be propagated at the end of the
-// transaction (if all operations complete successfully).  This is used to tell
-// PPS that new PipelineJobs may need to be created.
-func (t *TransactionContext) PropagateCommitset(commitset *pfs.StoredCommitset) error {
-	return t.PpsPropagater.PropagateCommitset(commitset)
+// PropagateJobs notifies PPS that there are new commits in the transaction's
+// commitset that need jobs to be created at the end of the transaction
+// transaction (if all operations complete successfully).
+func (t *TransactionContext) PropagateJobs() {
```
Tryneus, line 36: Changed `PropagateCommitset` -> `PropagateJobs` since there's really only one commitset for the transaction.

src/pfs/pfs.proto
```diff
@@ -227,17 +216,13 @@ message ListCommitRequest {
   bool reverse = 5;  // Return commits oldest to newest
 }
 
-message CommitInfos {
```
Tryneus, line 230: This was unused.

src/server/pfs/server/api_server.go
```diff
@@ -231,17 +231,24 @@ func (a *apiServer) ListCommit(request *pfs.ListCommitRequest, respServer pfs.AP
 	})
 }
 
+// InspectCommitsetInTransaction performs the same job as InspectCommitset
+// without the option of blocking for commits to finish so that it can run
+// inside an existing postgres transaction.  This is not an RPC.
+func (a *apiServer) InspectCommitsetInTransaction(txnCtx *txncontext.TransactionContext, commitset *pfs.Commitset) ([]*pfs.CommitInfo, error) {
```
Tryneus, line 237: It is useful to inspect a commitset in a transaction, but we can't block so this provides everything but the blocking functionality of `InspectCommitset`.

src/server/pfs/server/driver.go
```diff
@@ -684,16 +667,20 @@ func (d *driver) aliasCommit(txnCtx *txncontext.TransactionContext, parent *pfs.
 			parentCommitInfo.ChildCommits = append(parentCommitInfo.ChildCommits, commit)
 			return nil
 		}); err != nil {
+			if col.IsErrNotFound(err) {
+				return nil, pfsserver.ErrCommitNotFound{parent}
```
jdoliner, line 671: Will this error have a stack?

src/server/pfs/server/driver.go
```diff
@@ -1280,191 +1207,147 @@ func (d *driver) listCommit(ctx context.Context, repo *pfs.Repo, to *pfs.Commit,
...
-				Info: commitInfo,
-			})
-		}
-	} else {
-		commitset = &pfs.StoredCommitset{}
-		if err := d.commitsets.ReadOnly(ctx).Get(commitsetID, commitset); err != nil {
-			return nil, err
-		}
+	if len(commitMap) == 0 {
+		return nil, pfsserver.ErrCommitsetNotFound{Commitset: commitset}
```
jdoliner, line 1221: Same here, I think this won't have a stack.

jdoliner (approved):
Few questions around returning errors that might not have stacks.

src/server/pfs/server/driver.go
Tryneus: Yeah, this won't (but this is how PFS returns all `pfsserver` errors) - I can do a pass on these in a separate PR if that's ok.

PR #6229, opened 2021-06-01
Change jobs to use the same ID as the commitset which spawned them

Tryneus:
This changes `pps.PropagateJobs` to generate jobs with the same ID as the `txnCtx.CommitsetID`.  This required a few changes elsewhere in PPS, as the job ID needed to be joined with the pipeline name to keep it unique across all pipelines.


src/pps/pps.proto
```diff
@@ -261,67 +262,65 @@ message GPUSpec {
...
-  int64 data_recovered = 22;
-  int64 data_total = 23;
-  ProcessStats stats = 24;
-  repeated WorkerStatus worker_status = 25;
-  ResourceSpec resource_requests = 26;         // requires ListPipelineJobRequest.Full
-  ResourceSpec resource_limits = 27;           // requires ListPipelineJobRequest.Full
-  ResourceSpec sidecar_resource_limits = 28;   // requires ListPipelineJobRequest.Full
-  Input input = 29;                            // requires ListPipelineJobRequest.Full
-  pfs.BranchInfo new_branch = 30;
-  bool enable_stats = 32;                      // requires ListPipelineJobRequest.Full
```
PFedak, line 316: we just skipped 31 before?

src/server/worker/pipeline/transform/chain/chain.go
```diff
@@ -130,7 +132,7 @@ func (jdi *JobDatumIterator) Iterate(cb func(*datum.Meta) error) error {
 			if outputFilesetID, err = jdi.withDatumFileset(pachClient, func(outputSet *datum.Set) error {
 				return datum.Merge([]datum.Iterator{parentFilesetIterator, filesetIterator}, func(metas []*datum.Meta) error {
 					if len(metas) == 1 {
-						if metas[0].PipelineJobID != jdi.pipelineJobID {
+						if !proto.Equal(metas[0].PipelineJob, jdi.pipelineJob) {
```
PFedak, line 135: will this ever actually be considering datum metadata from another pipeline?
happy with comparing the protos regardless

PFedak (approved):
LGTM

src/pps/pps.proto
Tryneus: yeah, probably something I overlooked in a previous PR

src/server/worker/pipeline/transform/chain/chain.go
Tryneus: no, the chain is only run in the worker, which is for a single pipeline

PR #6230, opened 2021-06-01
Update PPS to use pipeline version instead of spec commit ID

Tryneus:
This PR adds a 'version' field to `StoredPipelineInfo` and `StoredPipelineJobInfo`.  Pipelines still track their original spec commit (because aliases will be created for each job).  Now, instead of checking logic by the spec commit ID, we determine if an operation is valid based on the pipeline version.


src/server/pps/server/pipeline_controller.go
```diff
@@ -359,9 +359,10 @@ func (op *pipelineOp) rcIsFresh() bool {
 		log.Errorf("PPS master: auth token in %q is stale %s != %s",
 			op.ptr.Pipeline.Name, rcAuthTokenHash, hashAuthToken(op.ptr.AuthToken))
 		return false
-	case rcSpecCommit != op.ptr.SpecCommit.ID:
+	// TODO(global ids): this won't trigger properly on start/stop pipeline which no longer updates the pipeline's OriginalSpecCommit
```
msteffen, line 362: Still reading through api_server.go, but does `CreatePipeline` update the pipeline version? It may be all that's needed to fix this is to change the tag in `worker_rc.go`  from `specCommitAnnotation` to `pipelineVersionAnnotation` and then compare that here

msteffen (commented):
Just publishing the one comment I have so far. Just need to finish understanding pps api_server.go

src/server/pps/server/pipeline_controller.go
Tryneus: Yes, but the thing I am worried about is start and stop pipeline, since those would create new spec commits (although I'm not entirely sure that's necessary) - I'm working on fixing those tests right now in the `grey/global-ids-pps-tests` branch.

msteffen (commented):
One small question about OriginalSpecCommit, but otherwise LGTM

PR #6232, opened 2021-06-01
Updating Skipped tests for 2.0 

acohen4:
Tests being addressed 
- TestCancelPipelineJob
- TestCancelManyPipelineJobs
- TestSquashCommitRunsJob
- TestEntryPoint
- TestUserWorkingDir
- TestSpoutPachctl/SpoutAuth, SpoutAuthEnabledAfter, SpoutProvenance

### DIFF
--- a/etc/testing/circle/run_tests.sh
+++ b/etc/testing/circle/run_tests.sh
@@ -26,6 +26,9 @@ export TESTFLAGS
 minikube status
 kubectl version
 
+# any tests that build images will do it directly in minikube's docker registry
+eval $(minikube docker-env)
+
 echo "Running test suite based on BUCKET=$BUCKET"
 
 function test_bucket {
--- a/src/internal/ppsutil/util.go
+++ b/src/internal/ppsutil/util.go
@@ -428,16 +428,17 @@ func WriteJobInfo(pachClient *client.APIClient, jobInfo *pps.JobInfo) error {
 	return err
 }
 
-func GetStatsCommit(commitInfo *pfs.CommitInfo) *pfs.Commit {
+func GetStatsCommit(commitInfo *pfs.CommitInfo) (*pfs.Commit, error) {
 	outputRepo := commitInfo.Commit.Branch.Repo.Name
 	for _, commitRange := range commitInfo.Subvenance {
 		repo := commitRange.Lower.Branch.Repo
 		if repo.Type == pfs.MetaRepoType && repo.Name == outputRepo {
-			return commitRange.Lower
+			return commitRange.Lower, nil
 		}
 	}
-	// TODO: Getting here would be a bug in 2.0, log?
-	return nil
+	// This may happen in the case the commitInfo represents an output commit for a Spout or Service pipeline,
+	// since stats aren't enabled in those cases
+	return nil, errors.Errorf("could not find a stats commit for commit %v", commitInfo.Commit.ID)
 }
 
 // ContainsS3Inputs returns 'true' if 'in' is or contains any PFS inputs with
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -8070,8 +8070,6 @@ func TestListJobInputCommits(t *testing.T) {
 // TestCancelJob creates a long-running job and then kills it, testing
 // that the user process is killed.
 func TestCancelJob(t *testing.T) {
-	// TODO(2.0 required): Investigate hang.
-	t.Skip("Investigate hang")
 	if testing.Short() {
 		t.Skip("Skipping integration tests in short mode")
 	}
@@ -8168,8 +8166,6 @@ func TestCancelJob(t *testing.T) {
 // running (which tests that only one job can run at a time), and then is
 // cancelled.
 func TestCancelManyJobs(t *testing.T) {
-	// TODO(2.0 required): Investigate hang.
-	t.Skip("Investigate hang")
 	if testing.Short() {
 		t.Skip("Skipping integration tests in short mode")
 	}
@@ -8201,7 +8197,7 @@ func TestCancelManyJobs(t *testing.T) {
 	var err error
 	for i := 0; i < 10; i++ {
 		commits[i], err = c.StartCommit(repo, "master")
-		require.NoError(t, c.PutFile(commits[i], "file", strings.NewReader("foo"), client.WithAppendPutFile()))
+		require.NoError(t, c.PutFile(commits[i], "file", strings.NewReader("foo")))
 		require.NoError(t, err)
 		require.NoError(t, c.FinishCommit(repo, commits[0].Branch.Name, commits[i].ID))
 	}
@@ -8321,9 +8317,6 @@ func TestSquashCommitPropagation(t *testing.T) {
 // which deletes the job's output commit and cancels the job. This should start
 // another job that processes the original input HEAD commit's parent.
 func TestSquashCommitRunsJob(t *testing.T) {
-	// TODO(2.0 required): Investigate transaction error,
-	// keeps using transaction after a conflict
-	t.Skip("Investigate transaction error")
 	if testing.Short() {
 		t.Skip("Skipping integration tests in short mode")
 	}
@@ -8454,12 +8447,9 @@ func TestSquashCommitRunsJob(t *testing.T) {
 }
 
 func TestEntryPoint(t *testing.T) {
-	// TODO(2.0 required): Investigate hang.
-	t.Skip("Investigate hang")
 	if testing.Short() {
 		t.Skip("Skipping integration tests in short mode")
 	}
-
 	c := tu.GetPachClient(t)
 	require.NoError(t, c.DeleteAll())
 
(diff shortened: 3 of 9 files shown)


src/server/pachyderm_test.go
```diff
@@ -8088,7 +8086,7 @@ func TestCancelPipelineJob(t *testing.T) {
 	// Create an input commit
 	commit, err := c.StartCommit(repo, "master")
 	require.NoError(t, err)
-	require.NoError(t, c.PutFile(commit, "/time", strings.NewReader("600"), client.WithAppendPutFile()))
+	require.NoError(t, c.PutFile(commit, "/time", strings.NewReader("15"), client.WithAppendPutFile()))
```
acohen4, line 8089: this timeout just appeared to be too long

src/server/pps/server/s3g_sidecar_test.go
```diff
@@ -667,7 +665,7 @@ func TestS3SkippedDatums(t *testing.T) {
 		pipelineCommit := client.NewCommit(pipeline, "master", "")
 		// Add files to 'repo'. Old files in 'repo' should be reprocessed in every
 		// job, changing the 'background' field in the output
-		for i := 0; i < 10; i++ {
+		for i := 0; i < 5; i++ {
```
acohen4, line 668: this just takes a long time. Reducing the number of iterations seem to make it more manageable

src/server/pfs/server/driver.go
```diff
@@ -1461,6 +1461,15 @@ func (d *driver) squashCommit(txnCtx *txncontext.TransactionContext, userCommit
 			if commit == nil {
 				return errors.Errorf("encountered nil parent commit in %s@%s...%s", lowerKey, lower.ID, upper.ID)
 			}
+			// are we leaking commits if we squash a commit with subvenance that forks and then merges?
```
acohen4, line 1464: @brycemcanally I may be misunderstanding the code, but this comment I made looks to me like it could be an issue.

In the diagram below it looks to me that within a subvenance commit range we would trace back from one upper commit to the lower, but would have deleted other commits along the path that needs to be taken for another deletion.

Does that make sense? I would expect the algorithm to queue and process multiple parent commits

src/internal/ppsutil/util.go
```diff
@@ -433,7 +433,7 @@ func GetStatsCommit(commitInfo *pfs.CommitInfo) *pfs.Commit {
 			return commitRange.Lower
 		}
 	}
-	// TODO: Getting here would be a bug in 2.0, log?
+	// TODO: Getting here would be a bug in 2.0, log? -- worth discussion?
```
PFedak, line 436: Ah, this is just false because of spouts, right?

src/server/pfs/server/driver.go
acohen4: @brycemcanally after our conversation earlier, this looks to me like a non-issue. Since the commit path is a tree, if this delete closure is called on all HEAD leaves, then there shouldn't be any undeleted commits from most descendant to to ascendant commits.

src/internal/ppsutil/util.go
acohen4: it seems like we do sometimes hit this. I'm trying to understand when that would happen

src/server/pps/server/api_server.go
```diff
@@ -976,12 +976,13 @@ func (a *apiServer) stopJob(txnCtx *txncontext.TransactionContext, job *pps.Job,
 		}); err != nil && !pfsServer.IsCommitNotFoundErr(err) && !pfsServer.IsCommitDeletedErr(err) && !pfsServer.IsCommitFinishedErr(err) {
 			return err
 		}
-
-		if err := a.env.PfsServer().FinishCommitInTransaction(txnCtx, &pfs.FinishCommitRequest{
-			Commit: ppsutil.GetStatsCommit(commitInfo),
-			Empty:  true,
-		}); err != nil && !pfsServer.IsCommitNotFoundErr(err) && !pfsServer.IsCommitDeletedErr(err) && !pfsServer.IsCommitFinishedErr(err) {
-			return err
+		if statsCommit := ppsutil.GetStatsCommit(commitInfo); statsCommit != nil {
```
acohen4, line 979: this check is basically handles: `if commit_type != spout`. Is there a clearer way to express this?

PFedak (commented):
LGTM except I'm a little concerned about removing the append in `TestCancelManyJobs` without looking into it further

src/internal/ppsutil/util.go
PFedak: I think we can change this comment to better reflect reality (that this might sometimes be called for spouts or services)

src/server/pps/server/api_server.go
```diff
@@ -976,12 +976,13 @@ func (a *apiServer) stopJob(txnCtx *txncontext.TransactionContext, job *pps.Job,
...
+			if err := a.env.PfsServer().FinishCommitInTransaction(txnCtx, &pfs.FinishCommitRequest{
+				Commit: ppsutil.GetStatsCommit(commitInfo),
```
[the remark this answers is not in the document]
acohen4: good catch :)

etc/testing/circle/run_tests.sh
```diff
@@ -26,6 +26,9 @@ export TESTFLAGS
 minikube status
 kubectl version
 
+# any tests that build images will do it directly in minikube's docker registry
+eval $(minikube docker-env)
+
```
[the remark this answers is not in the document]
acohen4: yep, it is also a fix for an image that we were building for PPS tests but weren't explicitly pushing. This seems like a double win for performance, and not needing to do a push for images built in test

src/server/pachyderm_test.go
```diff
@@ -8202,7 +8198,7 @@ func TestCancelManyJobs(t *testing.T) {
 	var err error
 	for i := 0; i < 10; i++ {
 		commits[i], err = c.StartCommit(repo, "master")
-		require.NoError(t, c.PutFile(commits[i], "file", strings.NewReader("foo"), client.WithAppendPutFile()))
+		require.NoError(t, c.PutFile(commits[i], "file", strings.NewReader("foo")))
```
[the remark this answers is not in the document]
acohen4: ah should have beat you to a github comment. Three of these tests (TestCancelJob, TestCancelManyJobs, TestSquashCommitRunsJob) are actually fixed by this change: https://github.com/pachyderm/pachyderm/pull/6232/files#diff-a84d233b0c21b3063c531e93ebb48872b3bb468f62b59476d459d5540fc6d7b3R298

Canceled Jobs' goroutines weren't being cancelled which would make the system hang


I just edited this line because of a discussion i had with Bryce where we can clean up some instances of "append" which could be addressed with normal PUT operations. I'm also happy to remove the change from this line

etc/testing/circle/run_tests.sh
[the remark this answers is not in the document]
acohen4: specifically `TestEntryPoint`, and `TestUserWorkingDir` were relying on the `pachyderm_entrypoint` testing image that wasn't being explicitly pushed

src/server/pps/server/api_server.go
acohen4: that's good to know. I think i'll leave it as a null check and just add a comment explaining the cases

src/server/pachyderm_test.go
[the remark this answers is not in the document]
PFedak: Ah okay, then I'm fine with it either way

PFedak (approved):
LGTM

src/internal/ppsutil/util.go
```diff
@@ -436,7 +436,7 @@ func GetStatsCommit(commitInfo *pfs.CommitInfo) *pfs.Commit {
 			return commitRange.Lower
 		}
 	}
-	// TODO: Getting here would be a bug in 2.0, log?
+	// Returns nil in the case our Commit is in Spout or Service repo
```
acohen4, line 439: @PFedak before I merge, does this comment use the right terminology? 

Is it that the commit is being processed by Spout or Service pipeline? I'm not entirely sure of the commit semantics around those cases

PFedak: I'd say in the case it's an output commit for a spout or service pipeline (those are really properties of the pipeline rather than the repo), specifically because we don't enable stats in those cases.

src/internal/ppsutil/util.go
brycemcanally: We shouldn't be calling this function for services / spouts. I think we should either return an error or panic when it isn't found so we don't end up using a nil commit when we shouldn't.

src/server/pfs/server/driver.go
brycemcanally: Yeah, we can revisit this after global IDs are in because this will probably change.

src/server/pps/server/s3g_sidecar_test.go
brycemcanally: This test is probably exposing a real performance issue we need to address. I think we should keep this as skipped with the original test. I am wrapping up a few performance fixes that may address this. One of the more significant ones is in master, so maybe you should merge that into this PR and see if that helps.

src/server/pps/server/api_server.go
brycemcanally: Like I mentioned above, I think we should follow the more standard convention of returning an error rather than a nil value. I do kind of prefer not actually calling that function when there is no expectation of a stats commit existing, so we could just put a panic in that function and hard fail when anything attempts to get a stats commit when it shouldn't.

brycemcanally (approved):
lgtm

PR #6236, opened 2021-06-02
Support getting cluster metrics in 2.0 with auth enabled

actgardner:
Switch metric collection to use the internal API, so we don't need an auth token to list pipelines and repos.

This also fixes a subtle bug in `ListRepo` which would have prevented users from listing repos if there was any repo they didn't have read access to. I added a test to make sure it won't regress, but this really highlights the need to pull all the auth checks in PFS into a common place - ideally high in the call stack so it's very clear what permission is required for each RPC.

### DIFF
--- a/src/server/auth/server/testing/auth_test.go
+++ b/src/server/auth/server/testing/auth_test.go
@@ -1103,6 +1103,10 @@ func TestListAndInspectRepo(t *testing.T) {
 	require.Equal(t,
 		buildBindings(alice, auth.RepoOwnerRole), getRepoRoleBinding(t, aliceClient, repoNone))
 
+	// put a file in the repo Bob can't access - we need to be able to get the size of the commits
+	err := aliceClient.PutFile(client.NewCommit(repoNone, "master", ""), "/test", strings.NewReader("test"))
+	require.NoError(t, err)
+
 	// bob creates a repo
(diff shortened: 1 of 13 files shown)


src/server/auth/server/testing/auth_test.go
```diff
@@ -1103,6 +1103,10 @@ func TestListAndInspectRepo(t *testing.T) {
 	require.Equal(t,
 		buildBindings(alice, auth.RepoOwnerRole), getRepoRoleBinding(t, aliceClient, repoNone))
 
+	// put a file in the repo Bob can't access - we need to be able to get the size of the commits
+	err := aliceClient.PutFile(client.NewCommit(repoNone, "master", ""), "/test", strings.NewReader("test"))
```
acohen4, line 1107: should we assert that there's a certain amount of bytes in the repo?

acohen4 (approved):
Just one question on the test. Otherwise LGTM

src/server/auth/server/testing/auth_test.go
actgardner: I'm not too concerned about asserting that here, PFS has it's own test that the results are accurate, here we're just testing the permissions model.

chainlink:
Niice! cc @mindthevirt

mindthevirt:
Love it! Thank you @actgardner

PR #6241, opened 2021-06-03
Autoscaling for 2.0

jdoliner:
This PR is mostly the same code as #5857 with a few additional / different things.

- coefficient parallelism was removed as part of this PR
- the datum task chunking had to be rewritten since the code it was based on had completely changed
- scaling is no longer based on unclaimed tasks but on total tasks, this removes any risk of overscaling due to slow pod spin up or pod crashes.


jdoliner:
@nadegepepin this has a few implications for docs:
- coefficient was removed, I removed it from the docs as well but some sections may be a bit awkward since they're written like there are two alternatives for parallelism and now there's only 1
- a new field was added to the chunk_spec called `datums_per_worker` I documented it in line but haven't added it to the pipeline spec yet.

src/internal/work/work.go
```diff
@@ -388,3 +391,16 @@ func (w *Worker) subtaskFunc(subtaskKey string, processFunc ProcessFunc) subtask
 		}
 	}
 }
+
+// TaskCount returns how many subtasks are in the queue and how many are claimed.
+func (w *Worker) TaskCount(ctx context.Context) (int64, int64, error) {
```
brendoncarroll, line 396: There should probably be names for the 2 int64s, so they can be distinguished by looking at the signature.  Use `_` for the error.

src/server/pps/server/monitor.go
```diff
@@ -325,6 +329,53 @@ func (m *ppsMaster) monitorPipeline(ctx context.Context, pipelineInfo *pps.Pipel
 			}, backoff.NewInfiniteBackOff(),
 				backoff.NotifyCtx(ctx, "monitorPipeline for "+pipeline))
 		})
+		if pipelineInfo.ParallelismSpec != nil && pipelineInfo.ParallelismSpec.Constant > 1 && pipelineInfo.Autoscaling {
+			eg.Go(func() error {
+				pachClient := m.a.env.GetPachClient(ctx)
+				return backoff.RetryNotify(func() error {
```
brendoncarroll, line 335: should this be `RetryUntilCancel` so it can respect the context?

brendoncarroll (approved):
A couple small suggestions, otherwise LGTM.

src/internal/deploy/assets/assets.go
```diff
@@ -66,7 +66,7 @@ var (
 	}, {
 		APIGroups: []string{""},
 		Verbs:     []string{"get", "list", "watch", "create", "update", "delete"},
-		Resources: []string{"replicationcontrollers", "services"},
+		Resources: []string{"replicationcontrollers", "services", "replicationcontrollers/scale"},
```
jdoliner, line 69: @chainlink this is the added resources for autoscaling.

PR #6245, opened 2021-06-03
Allow modifications to transactions while finishing.

PFedak:
Creating pipelines in a transaction has the dubious distinction that we
expect it to fail and update its saved response when it encounters
modifications outside of the transaction. To accommodate this logic,
appending a CreatePipelineRequest to a transaction involves updating the
stored transaction info to reflect the most recent state of the
pipeline's spec repo.

However, another source of inside/outside transaction conflicts is
during FinishTransaction, after the append logic has finished.
Previously, the finishing code was not able to modify the stored
transaction, ended up looping in a futile attempt to resolve the
conflict.

The new logic may be overengineered, but provides a straightforward way
to modify in a transaction in a way that should be persisted in some
cases.

Also, modified the error for a transaction being modified to have a name
better reflecting that scenario, as opposed to one almost identical to
an unrelated SQL-level error.


src/server/pachyderm_test.go
```diff
@@ -10449,8 +10447,7 @@ func TestInterruptedUpdatePipelineInTransaction(t *testing.T) {
 	require.NoError(t, createPipeline(c, inputC, true))
 
 	_, err = c.FinishTransaction(txn)
-	require.YesError(t, err)
-	require.Matches(t, "outside of transaction", err.Error())
+	require.NoError(t, err)
```
Tryneus, line 10450: Since we're expecting this to succeed now, we should probably check that the latest pipeline info is the right version and has the right input.

src/server/transaction/server/driver.go
```diff
@@ -197,26 +198,34 @@ func (d *driver) runTransaction(txnCtx *txncontext.TransactionContext, info *tra
...
+	return d.updateTransaction(ctx, func() (*transaction.TransactionInfo, error) {
+		if err := d.txnEnv.WithWriteContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
+			err := d.transactions.ReadWrite(txnCtx.SqlTx).Get(txn.ID, info)
+			if err != nil {
+				return err
+			}
+			info, err = d.runTransaction(txnCtx, info)
+			if err != nil {
+				if col.IsErrTransactionConflict(err) {
+					// a pipeline changed, and we need to update the transaction data
```
Tryneus, line 210: This _could_ happen for other reasons in the future, maybe make this comment more general or just list that as an example.

src/server/transaction/server/driver.go
```diff
@@ -241,44 +247,60 @@ func (d *driver) appendTransaction(
...
-			return info, nil
-		} else if !errors.As(err, &transactionConflictError{}) {
+		if err != nil && errors.Is(err, errutil.ErrBreak) {
+			return newInfo, nil // no need to update
+		}
+
+		var oldInfo transaction.TransactionInfo
+		// a transactionModifiedError from the update function should be understood
+		// as the active transaction needing to change in a way that should be retried after
+		if err == nil || err != nil && errors.Is(err, &transactionModifiedError{}) {
```
Tryneus, line 283: This boolean logic can be simplified (I think `err != nil` is redundant)

src/server/transaction/server/driver.go
```diff
@@ -241,44 +247,60 @@ func (d *driver) appendTransaction(
...
-				info.Responses = newResponses
+			if newInfo.Responses != nil {
+				info.Responses = newInfo.Responses
 			}
 
 			// This may error with a postgres-level transaction conflict, in which
 			// case we want to save the generated responses and reattempt.  The saved
 			// responses will contain any references to objects generated outside of
 			// the transaction and need to be reused on the next attempt.
 			info, err = d.runTransaction(txnCtx, info)
```
Tryneus, line 259: I think this function needs the same handling as `FinishTransaction` to map a collection error to a `transactionModifiedError` (for example if the collection transaction error occurs when doing the dryrun when appending a CreatePipeline to an existing transaction) so maybe that logic can just be moved into `updateTransaction`?

Tryneus (changes requested):

src/server/transaction/server/driver.go
PFedak: Hmm, I think the old logic is correct as it was, since it is careful to make sure the SQL transaction is retried with the updated `TransactionInfo`. Regardless, I moved things around so it doesn't have to do this dance.

src/server/transaction/server/driver.go
```diff
@@ -225,60 +221,81 @@ func (d *driver) appendTransaction(
...
-
-			// Save the length so that we can check that nothing else modifies the
-			// transaction in the meantime
-			numRequests = len(info.Requests)
-			numResponses = len(info.Responses)
-			info.Requests = append(info.Requests, items...)
-			if newResponses != nil {
-				info.Responses = newResponses
-			}
+		info.Requests = append(info.Requests, items...)
```
PFedak, line 228: This will have different behavior from before, since now a `CreatePipeline` retry will count against the 10 attempts, rather than happening entirely inside of the SQL transaction. That case *should* be rare - honestly I'm not sure the real world cases where we expect most of this code to be exercised, anyway

src/server/transaction/server/driver.go
```diff
@@ -225,60 +221,81 @@ func (d *driver) appendTransaction(
...
-			// Update the existing transaction with the new requests/responses
-			return d.transactions.ReadWrite(sqlTx).Update(txn.ID, info, func() error {
-				if len(info.Requests) != numRequests || len(info.Responses) != numResponses {
-					// Someone else modified the transaction while we did the dry run
-					return &transactionConflictError{}
-				}
+	for i := 0; i < 10; i++ {
+		var err error
+		if writeTxn {
+			err = d.txnEnv.WithWriteContext(ctx, attempt)
```
PFedak, line 262: could also pass in one of these functions but the signatures are getting kind of cumbersome. I guess I could have defined some types for them

Tryneus (approved):
LGTM once CI passes

src/server/transaction/server/driver.go
```diff
@@ -225,60 +221,81 @@ func (d *driver) appendTransaction(
...
-				info.Requests = append(info.Requests, items...)
-				info.Responses = newResponses
-				return nil
-			})
-		}); err == nil {
-			return info, nil
-		} else if !errors.As(err, &transactionConflictError{}) {
+		// a transactionModifiedError from the update function should be understood
+		// as the active transaction needing to change in a way that should be retried after
+		if err == nil || errors.Is(err, &transactionModifiedError{}) {
```
acohen4, line 272: why go through update logic when `err == nil `? isn't that condition also implying success

runTransaction called [here](https://github.com/pachyderm/pachyderm/pull/6245/files#diff-4f64d228babb3598928440a7b5e19713aadadc53c352918ee0c12293640ead7eR229) looks like it would return err == nil in a success case

PR #6246, opened 2021-06-03
Rename all client proto packages to _v2

actgardner:
Rename the proto packages in 2.x so both 1.x and 2.x can be imported in the same codebase, like so:

```
package main

import (
        "fmt"
        v1 "github.com/pachyderm/pachyderm/src/client"
        v2 "github.com/pachyderm/pachyderm/v2/src/client"
)

func main() {
        _, err := v1.NewOnUserMachine("client")
       fmt.Printf("v1 error: %v\n", err)

        v2Client, err := v2.NewOnUserMachine("client")
        if err != nil {
                fmt.Printf("v2 error: %v\n", err)
        }

        v2Resp, err := v2Client.Version()
        if err != nil {
                fmt.Printf("v2 error: %v\n", err)
        }

        fmt.Printf("V2: %v\n", v2Resp)
}
```

Output (running against a v2 cluster):

```
v1 error: could not get cluster ID: unknown service admin.API
V2: 2.0.0-8981a76b1d340071b4b2392a5914f761c6df65d2
```


jrockway (approved):
This is really nice.  I like that it makes pachctl 1.x fail against a 2.x cluster (and vice-versa); this error message is much easier to debug than the alternative (which would be field tag numbers referring to different things, or missing data in requests/responses).

chainlink:
Looking forward to this proto change in all the clients! 👍 
cc @albscui

PR #6248, opened 2021-06-04
Support accessing meta repos via the s3 gateway.

PFedak:


src/server/pfs/s3/driver.go
```diff
@@ -72,28 +81,31 @@ func (d *MasterDriver) listBuckets(pc *client.APIClient, r *http.Request, bucket
 }
 
 func (d *MasterDriver) bucket(pc *client.APIClient, r *http.Request, name string) (*Bucket, error) {
+	var repo *pfs.Repo
 	branch := "master"
-	var repo string
-	parts := strings.SplitN(name, ".", 2)
-	if len(parts) == 2 {
+	parts := strings.SplitN(name, ".", 3)
```
PFedak, line 86: The resolution behavior is the thing that seems most important, since it impacts how users interact with it. I'd like to require as little change as possible, but anyone who used S3 to access the `stats` branch before will need to do something different regardless.

src/server/pfs/s3/driver.go
```diff
@@ -50,19 +50,28 @@ func NewMasterDriver() *MasterDriver {
...
 
 func (d *MasterDriver) listBuckets(pc *client.APIClient, r *http.Request, buckets *[]*s2.Bucket) error {
-	repos, err := pc.ListRepo()
+	repos, err := pc.ListRepoByType("") // get repos of all types
 	if err != nil {
 		return err
 	}
 
 	for _, repo := range repos {
+		if repo.Repo.Type == pfs.SpecRepoType {
```
PFedak, line 59: My understanding of the code was that leaving a bucket out of the response here doesn't actually prevent it from being accessed directly by name, but maybe that's incorrect. Also not sure on whether to include stats/meta repos by default

src/server/pfs/s3/master_test.go
```diff
@@ -416,6 +417,76 @@ func masterListObjectsRecursive(t *testing.T, pachClient *client.APIClient, mini
...
+func masterResolveSystemRepoBucket(t *testing.T, pachClient *client.APIClient, minioClient *minio.Client) {
+	repo := tu.UniqueString("testsystemrepo")
+	require.NoError(t, pachClient.CreateRepo(repo))
+	specRepo := client.NewSystemRepo(repo, pfs.SpecRepoType)
+	// create a branch named "spec" in the repo
+	branch := pfs.SpecRepoType
+	require.NoError(t, pachClient.CreateBranch(repo, branch, "", "", nil))
+
+	// as well as a branch named "master" on an associated repo of type "spec"
+	_, err := pachClient.PfsAPIClient.CreateRepo(pachClient.Ctx(), &pfs.CreateRepoRequest{Repo: specRepo})
```
msteffen, line 464: very minor suggestion: move `specRepo` definition down to where `CreateRepo` is called:
```suggestion
	// create a branch named "spec" in the repo
	branch := pfs.SpecRepoType
	require.NoError(t, pachClient.CreateBranch(repo, branch, "", "", nil))

	// as well as a branch named "master" on an associated repo of type "spec"
	specRepo := client.NewSystemRepo(repo, pfs.SpecRepoType)
	_, err := pachClient.PfsAPIClient.CreateRepo(pachClient.Ctx(), &pfs.CreateRepoRequest{Repo: specRepo})
```

src/server/pfs/s3/multipart.go
```diff
@@ -59,11 +59,11 @@ func multipartKeepArgs(path string) (repo string, branch string, key string, upl
 }
 
 func parentDirPath(bucket *Bucket, key string, uploadID string) string {
```
msteffen, line 61: Out of scope, but this could be massively improved by using the new filesets API

msteffen (approved):
Few suggestions, but overall LGTM

src/server/pfs/s3/multipart.go
PFedak: I didn't entirely understand the use case from reading the code, not surprised there's a better way now

src/server/pfs/s3/driver.go
```diff
@@ -72,28 +81,31 @@ func (d *MasterDriver) listBuckets(pc *client.APIClient, r *http.Request, bucket
 }
 
 func (d *MasterDriver) bucket(pc *client.APIClient, r *http.Request, name string) (*Bucket, error) {
+	var repo *pfs.Repo
 	branch := "master"
-	var repo string
-	parts := strings.SplitN(name, ".", 2)
-	if len(parts) == 2 {
+	parts := strings.SplitN(name, ".", 3)
+	if len(parts) == 3 {
```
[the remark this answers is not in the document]
PFedak: yeah, "user" is just the default

PR #6250, opened 2021-06-04
Change 2.0 default storage configuration to avoid thrashing

brycemcanally:
This PR changes the default storage configuration for 2.0 to avoid thrashing. The thrashing occurs because the max fan in is higher than the max number of chunks that we can cache in memory. In certain workloads (small files that were not added in lexicographical order) this results with downloading, decompressing, and decrypting chunks many more times than we need to. The new default storage configuration avoids this by having the number of chunks we can cache in memory larger than the max fan in (this should probably be multiple times larger in general to account for parallel operations).

Miscellaneous:
- Fixes a bug in the load tester that would not reset the run for directories.
- Fixes a race condition that could cause compaction subtasks to be collected in the incorrect order.


jdoliner (approved):
LGTM

PR #6251, opened 2021-06-04
Update job status (datums processed, skipped, etc...) while a job is running

brycemcanally:
This PR adds job status updates while a job is running in 2.0 and logging for the skipping / datum deletion step in a job.


jdoliner (approved):
LGTM

PR #6252, opened 2021-06-04
PPS worker performance improvements

brycemcanally:
This PR contains a couple small changes to improve the performance of the PPS workers:
- Improves the performance of skipping by writing meta files directly to the fileset stream, rather than going through the local filesystem (the per file overhead needs to be kept really low for this to be acceptable).
- We now convert the output to a tar stream on the fly, rather than buffering the output into a tar file on the local filesystem.
- Refactors the handling of symlinks out of the tarutil package.


acohen4 (approved):
LGTM

PR #6253, opened 2021-06-04
Rewrite stable hash computation to improve performance

brycemcanally:
This PR rewrites the stable hash computation to improve the performance. The current stable hash computation, based on the index resolver, has bad performance on workloads that cannot take advantage of the cheap copy functionality provided by the chunk storage layer (avoids reading chunks that have a full set of data references copied into the writer). Workloads that have a lot of small files with the paths randomly distributed across multiple filesets will essentially need to read all / a lot of the data to compute the hashes. This approach doesn't make as much sense with the notion of file parts removed since we don't need to determine the hash of a file by combining multiple file parts. So, the new approach basically fast paths the most common case which is that the hash that we store with the file is the resolved hash. For the case where we have a file spread across multiple filesets (a large file or using the append functionality), we setup a chunk writer for just the edges of the filesets / appended content (the part that needs to be resolved).


src/internal/storage/chunk/storage.go
```diff
@@ -76,3 +77,49 @@ func (s *Storage) List(ctx context.Context, cb func(id ID) error) error {
 func (s *Storage) NewDeleter() track.Deleter {
 	return &deleter{}
 }
+
+// StableHash returns a stable hash for a set of data refs.
+func (s *Storage) StableHash(ctx context.Context, dataRefs []*DataRef) ([]byte, error) {
```
brendoncarroll, line 82: I don't think this really belongs in the chunk package.  Can we move this to the fileset package?  It's a little weird that we have to punch through all the way to the chunk level to do this in the pfs server.

src/internal/storage/chunk/chunk.proto
```diff
@@ -8,8 +8,7 @@ message DataRef {
   // The chunk the referenced data is located in.
   Ref ref = 1;
   // The hash of the data being referenced.
-  // This field is empty when it is equal to the chunk hash (the ref is the whole chunk).
-  string hash = 2;
+  bytes hash = 2;
```
brendoncarroll, line 11: 👍

src/server/pfs/server/source.go
```diff
@@ -20,6 +20,7 @@ type Source interface {
 }
 
 type source struct {
+	storage    *fileset.Storage
```
brendoncarroll, line 23: Can we not depend on the whole fileset.Storage?  We seem to just grab the ChunkStorage() below.  I would prefer to pass that in in the constructor.

src/internal/storage/chunk/storage.go
brycemcanally: I think this functionality makes sense in the chunk layer because it requires knowledge about how the data refs are generated by the writers and how they correspond to chunks, but I agree that we shouldn't need to punch into the chunk layer at the PFS layer. How about if I add a wrapper for this at the fileset layer that operates on the indexes?

src/server/pfs/server/source.go
brycemcanally: If you agree with the change above, we would want to keep this.

brendoncarroll (approved):
We talked offline about keeping the functionality in the chunk package and passing the chunk.Storage directly to the pfs Source.

PR #6254, opened 2021-06-04
Add cache to fileset index layer that enables binary searching on index chunks

brycemcanally:
This PR adds a cache to the fileset index layer that enables binary searching on index chunks. Index chunks store serialized index protos and are subject to the same min / avg / max configuration of normal file content chunks. This means that index chunks can have a lot of index protos packed into them (i.e. the current default chunk avg is 8MB, which would be a lot of index protos). A linear scan through these protos is a significant amount of overhead for point queries (we had a similar problem in 1.x with output repos that we "addressed" by just making the chunks smaller). To address this, I added a caching layer to the indexing layer that stores sorted path offsets in memory for a set of cached index chunks, which allows us to perform a binary search on the indexes in the chunk when a cache hit occurs. There might be some opportunity for improving this later by either exposing arbitrary metadata to associate with chunks cached in the chunk storage layer so that we don't need a separate cache for this and potentially just storing the index entries rather than offsets that are used to select the right byte offset to feed into the normal index reader.

### DIFF
--- a/src/internal/storage/fileset/index/cache.go
+++ b/src/internal/storage/fileset/index/cache.go
@@ -1,0 +1,106 @@
+package index
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"sort"
+	"sync"
+
+	"github.com/hashicorp/golang-lru/simplelru"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pbutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/chunk"
+)
+
+type Cache interface {
+	Get(ctx context.Context, chunkRef *chunk.DataRef, filter *pathFilter, w io.Writer) error
+}
+
+type cache struct {
+	storage *chunk.Storage
+	cache   *simplelru.LRU
+	mu      sync.Mutex
+}
+
+func NewCache(storage *chunk.Storage, size int) Cache {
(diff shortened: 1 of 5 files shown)


src/internal/storage/fileset/index/cache.go
```diff
@@ -0,0 +1,106 @@
...
+	"sort"
+	"sync"
+
+	"github.com/hashicorp/golang-lru/simplelru"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pbutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/chunk"
+)
+
+type Cache interface {
+	Get(ctx context.Context, chunkRef *chunk.DataRef, filter *pathFilter, w io.Writer) error
```
brendoncarroll, line 16: `chunkRef` -> `dataRef`

brendoncarroll (commented):
The interface Cache doesn't seem like it will realistically get much reuse.  It would be one thing if it had an Add method, but it doesn't, the cache miss logic is implied in the Get method.

I think the `Cache` interface should either be
1.  A concrete type.
2. Also have an Add method.  The package can have a function for looking up from the cache and falling back to the chunk storage.  At that point, I could expect there to be another implementation of Cache.

brendoncarroll:
Let's also avoid exposing a knob for this, at least for a little while.

PR #6256, opened 2021-06-04
Remove build pipelines.

PFedak:


src/internal/metrics/metrics.proto
```diff
@@ -22,41 +22,40 @@ message Metrics {
...
-    float resource_cpu_limit       = 45; // Total CPU limits set across all pipelines
-    float resource_cpu_limit_max   = 46; // Max CPU limit set
-    string resource_mem_limit      = 47; // String of memory limits set
-    int64 resource_gpu_limit       = 48; // Number of pipelines with
-    int64 resource_gpu_limit_max   = 49; // Max GPU limit set
-    string resource_disk_limit     = 50; // String of disk limits set across all pipelines
-    uint64 max_parallelism         = 51; // Max parallelism set
-    uint64 min_parallelism         = 52; // Min parallelism set
-    uint64 num_parallelism         = 53; // Number of pipelines with parallelism set
+    int64 cfg_egress               = 17; // Number of pipelines with Egress configured
```
nitinjainsj, line 25: I would not change the number. Just mark the field unused. An old interpreter will read "17" as data for pps_build.

nitinjainsj (commented):
In a separate PR, we should remove build pipeline from top level make file, the corresponding directories in etc/pipeline-build, and goreleaser/docker-build-pipeline.yml

src/internal/metrics/metrics.proto
PFedak: Hmm, I thought the plan was to make protos explicitly incompatible between 1.x and 2.x, but I can preserve the numbers

chainlink:
Follow on PR: https://github.com/pachyderm/pachyderm/pull/6289

nitinjainsj (approved):

PR #6257, opened 2021-06-04
Add golanglint-ci

actgardner:
Re-enable linting in CI and clean up a few pieces of unused code that are detected. golanglint-ci is a meta-linter which runs govet a few other checks by default (https://golangci-lint.run/usage/linters/). This is a mostly stock config, except `errcheck` is currently disabled because we have an alarming number of unchecked errors to work through. 

This removes golint - which is deprecated and "is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process" (https://github.com/golang/lint#purpose). It is no longer necessary to write a godoc comment for every public field, struct and method.

### DIFF
--- a/src/internal/serviceenv/config.go
+++ b/src/internal/serviceenv/config.go
@@ -68,8 +68,8 @@ type PachdSpecificConfiguration struct {
 	MemoryRequest              string `env:"PACHD_MEMORY_REQUEST,default=1T"`
 	WorkerUsesRoot             bool   `env:"WORKER_USES_ROOT,default=true"`
 	DeploymentID               string `env:"CLUSTER_DEPLOYMENT_ID,default="`
-	RequireCriticalServersOnly bool   `env:"REQUIRE_CRITICAL_SERVERS_ONLY",default="false"`
-	MetricsEndpoint            string `env:"METRICS_ENDPOINT",default=""`
+	RequireCriticalServersOnly bool   `env:"REQUIRE_CRITICAL_SERVERS_ONLY,default=false"`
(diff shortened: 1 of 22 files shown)


src/internal/serviceenv/config.go
```diff
@@ -68,8 +68,8 @@ type PachdSpecificConfiguration struct {
 	MemoryRequest              string `env:"PACHD_MEMORY_REQUEST,default=1T"`
 	WorkerUsesRoot             bool   `env:"WORKER_USES_ROOT,default=true"`
 	DeploymentID               string `env:"CLUSTER_DEPLOYMENT_ID,default="`
-	RequireCriticalServersOnly bool   `env:"REQUIRE_CRITICAL_SERVERS_ONLY",default=false"`
-	MetricsEndpoint            string `env:"METRICS_ENDPOINT",default="`
```
msteffen, line 72: :exploding_head:

src/internal/sql/sql.go
```diff
@@ -81,5 +81,4 @@ func (r *PGDumpReader) readFooter() error {
 			return err
 		}
 	}
-	return nil
```
msteffen, line 84: :joy:

msteffen (approved):
LGTM. Two comments, both emoji.

Also, for posterity (I had to look it up): golangci-lint runs staticcheck, so removing staticcheck from `etc/testing/lint.sh` should be a no-op.

src/internal/serviceenv/config.go
actgardner: struct tags are evil and nothing will ever convince me otherwise 🤷

PR #6259, opened 2021-06-04
Remove Githooks from 2.0

actgardner:
As discussed at the engineering sync, Githooks are deprecated and we're removing them from 2.0.


nitinjainsj (approved):
LGTM.

I would check one more thing... are there any examples to that use githook. Customer team will probably know this as well.

PR #6264, opened 2021-06-07
Merge global IDs to master

Tryneus:
This PR is for the same branch as #6263, but is targeted at `master` instead of the integration branch, so it contains the full diff of code for the global IDs feature, including a lot of code which has already been reviewed in #6178, #6216, #6226, #6227, #6228, #6229, and #6230.


src/server/pps/server/transaction_defer.go
```diff
@@ -0,0 +1,36 @@
+package server
+
+import (
+	"github.com/pachyderm/pachyderm/v2/src/internal/transactionenv/txncontext"
+)
+
+// Propagater is an object that is used to create jobs in response to a new
```
brycemcanally, line 7: :+1:

src/server/pps/server/api_server.go
```diff
@@ -3084,6 +2852,60 @@ func (a *apiServer) RunCron(ctx context.Context, request *pps.RunCronRequest) (r
 	return &types.Empty{}, nil
 }
 
+func (a *apiServer) propagateJobs(txnCtx *txncontext.TransactionContext) error {
```
brycemcanally, line 2855: :+1:

src/client/pfs.go
```diff
@@ -410,43 +388,58 @@ func (c APIClient) FlushCommit(commits []*pfs.Commit, toRepos []*pfs.Repo, cb fu
...
-	if len(cis) > 0 {
-		sort.Slice(cis, func(i, j int) bool {
-			return len(cis[i].Provenance) < len(cis[j].Provenance)
-		})
+	return result, nil
+}
+
+// BlockCommitset blocks until all of a Commitset's commits are finished.  To
+// wait for an individual commit, use BlockCommit instead.
+func (c APIClient) BlockCommitsetAll(id string) ([]*pfs.CommitInfo, error) {
```
brycemcanally, line 405: It would probably be better to have `BlockCommitsetAll` call `BlockCommitset` rather than inspect commit. That would be a safer pattern for ensuring the same semantics. Also, you can just return the `inspectCommitset` call in `BlockCommitset`.

brycemcanally: Also, in regards to the gRPC scrubbing, it would probably be safer to handle it right at the API boundary (set up the defer immediately in an exported method) rather than in the internal methods. It seems like it would be easier to ensure that we are doing it.

src/client/pps.go
```diff
@@ -202,28 +207,12 @@ func NewPipeline(pipelineName string) *pps.Pipeline {
...
 // InspectJob returns info about a specific job.
-// blockState will cause the call to block until the job reaches a terminal state (failure or success).
 // full indicates that the full job info should be returned.
-func (c APIClient) InspectJob(jobID string, blockState bool, full ...bool) (*pps.JobInfo, error) {
+func (c APIClient) InspectJob(pipelineName string, jobID string, full ...bool) (*pps.JobInfo, error) {
 	req := &pps.InspectJobRequest{
-		Job:        NewJob(jobID),
-		BlockState: blockState,
+		Job:        NewJob(pipelineName, jobID),
+		BlockState: false,
```
brycemcanally, line 215: You could just use the default here. Also, I guess there was a decision to expose the block functionality as two separate methods rather than an option on one. What was the thinking behind this decision?

src/client/pps.go
```diff
@@ -232,18 +221,84 @@ func (c APIClient) InspectJob(jobID string, blockState bool, full ...bool) (*pps
...
+	return result, nil
+}
+
+func (c APIClient) BlockJobsetAll(id string, full ...bool) ([]*pps.JobInfo, error) {
+	isFull := false
+	if len(full) > 0 {
+		isFull = full[0]
+	}
+	result := []*pps.JobInfo{}
+	if err := c.inspectJobset(id, true, isFull, func(ji *pps.JobInfo) error {
```
brycemcanally, line 289: Same as what I mentioned above, it would be better for this to call `BlockJobset`. Also, I wouldn't be against making full required. I'm on board with moving away from variadic parameters for single optional parameters. Having the pattern of required parameters or functional options makes sense to me.

src/client/pps.go
```diff
@@ -329,116 +383,88 @@ func (c APIClient) ListJobFilterF(pipelineName string, inputCommit []*pfs.Commit
...
-	req := &pps.FlushJobRequest{
-		Commits: commits,
-	}
-	for _, pipeline := range toPipelines {
-		req.ToPipelines = append(req.ToPipelines, NewPipeline(pipeline))
-	}
-	client, err := c.PpsAPIClient.FlushJob(c.Ctx(), req)
+// SubscribeJob calls the given callback with each open job in the given
+// pipeline until canceled.
+func (c APIClient) SubscribeJob(pipelineName string, includePipelineInfo bool, cb func(*pps.JobInfo) error) error {
```
brycemcanally, line 388: It seems like `includePipelineInfo` should just be called `full`. Also, is "open job" what we are calling a job that isn't terminal?

src/client/pps.go
```diff
@@ -329,116 +383,88 @@ func (c APIClient) ListJobFilterF(pipelineName string, inputCommit []*pfs.Commit
...
-		return nil
-	}); err != nil {
-		return nil, err
-	}
-	return result, nil
-}
-
 // DeleteJob deletes a job.
-func (c APIClient) DeleteJob(jobID string) error {
+func (c APIClient) DeleteJob(pipelineName string, jobID string) error {
```
brycemcanally, line 415: This should probably just take a `pps.Job`.

src/internal/auth/interceptor.go
```diff
@@ -119,49 +119,50 @@ var authHandlers = map[string]authHandler{
...
+	"/pfs_v2.API/ActivateAuth":     clusterPermissions(auth.Permission_CLUSTER_AUTH_ACTIVATE),
+	"/pfs_v2.API/CreateRepo":       authDisabledOr(authenticated),
+	"/pfs_v2.API/InspectRepo":      authDisabledOr(authenticated),
+	"/pfs_v2.API/ListRepo":         authDisabledOr(authenticated),
+	"/pfs_v2.API/DeleteRepo":       authDisabledOr(authenticated),
+	"/pfs_v2.API/StartCommit":      authDisabledOr(authenticated),
+	"/pfs_v2.API/FinishCommit":     authDisabledOr(authenticated),
+	"/pfs_v2.API/InspectCommit":    authDisabledOr(authenticated),
+	"/pfs_v2.API/ListCommit":       authDisabledOr(authenticated),
+	"/pfs_v2.API/SquashCommit":     authDisabledOr(authenticated),
```
brycemcanally, line 131: This got removed right?

src/internal/cmdutil/cobra.go
```diff
@@ -205,6 +206,19 @@ func ParseBranch(arg string) (*pfs.Branch, error) {
 	return commit.Branch, nil
 }
 
+// ParseJob takes an argument of the form "pipeline@job-id" and returns
+// the corresponding *pfs.Job.
```
brycemcanally, line 210: `pps.Job`

src/internal/collection/errors.go
```diff
@@ -66,3 +66,25 @@ func (err ErrTransactionConflict) Error() string {
 func IsErrTransactionConflict(err error) bool {
 	return errors.Is(err, ErrTransactionConflict{})
 }
+
+// ErrNotUnique indicates that an indexed query expected to have exactly one
```
brycemcanally, line 70: This reads weird.

src/internal/collection/types.go
```diff
@@ -93,6 +93,10 @@ type PostgresReadWriteCollection interface {
 	// a large number of rows. Consider using a read-only collection if possible
 	GetByIndex(index *Index, indexVal string, val proto.Message, opts *Options, f func(string) error) error
 
+	// GetUniqueByIndex is identical to GetByIndex except it is an error if
```
brycemcanally, line 96: This seems like something that would be better as an option on `GetByIndex` or a wrapper, rather than an additional function on the interface.

src/internal/ppsutil/util.go
```diff
@@ -31,14 +31,18 @@ import (
...
-	"github.com/pachyderm/pachyderm/v2/src/internal/pfsdb"
 	"github.com/pachyderm/pachyderm/v2/src/internal/ppsconsts"
 	"github.com/pachyderm/pachyderm/v2/src/internal/tracing"
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
 	"github.com/pachyderm/pachyderm/v2/src/pps"
 	ppsServer "github.com/pachyderm/pachyderm/v2/src/server/pps"
 )
 
+// JobKey is the string representation of a Job suitable for use as an indexing key
+func JobKey(job *pps.Job) string {
```
brycemcanally, line 42: Is there a dependency reason for why this cannot just be in ppsdb?

src/internal/ppsutil/util.go
```diff
@@ -120,7 +124,7 @@ func GetPipelineInfoAllowIncomplete(pachClient *client.APIClient, ptr *pps.Store
 	// ensure we are authorized to read the pipeline's spec commit, but don't propagate that back out
 	pachClient = pachClient.WithCtx(pachClient.Ctx())
 	pachClient.SetAuthToken(ptr.AuthToken)
-	if err := pachClient.GetFile(ptr.SpecCommit, ppsconsts.SpecFile, &buf); err != nil {
+	if err := pachClient.GetFile(ptr.OriginalSpecCommit, ppsconsts.SpecFile, &buf); err != nil {
```
brycemcanally, line 127: As a side note, I don't like the idea of attempting to proceed if we error when getting the spec file. I feel like this should be considered a critical error, and we shouldn't need to account for an incomplete pipeline info after attempting to get it. Not really suggesting any changes for this PR, just wanted to draw attention to it.

src/internal/ppsutil/util.go
```diff
@@ -363,23 +355,32 @@ func UpdateJobState(pipelines col.ReadWriteCollection, jobs col.ReadWriteCollect
...
 	pipelinePtr.JobCounts[int32(state)]++
 	pipelinePtr.LastJobState = state
-	if err := pipelines.Put(jobPtr.Pipeline.Name, pipelinePtr); err != nil {
+	if err := pipelines.Put(jobPtr.Job.Pipeline.Name, pipelinePtr); err != nil {
 		return err
 	}
 
 	// Update job info
 	var err error
 	if state == pps.JobState_JOB_STARTING {
```
brycemcanally, line 364: Where do we actually update the job state to starting? I see where we are setting a timestamp in `startJob` in the registry, which should probably just go through this code path instead. Also, I'm not sure if it makes sense to set the started if it is nil in the terminal case. If we are depending on either both or none being then set, then I guess it is fine.

src/internal/testpachd/mock_transaction.go
```diff
@@ -4,11 +4,20 @@ import (
 	"fmt"
 
 	"github.com/pachyderm/pachyderm/v2/src/internal/transactionenv/txncontext"
-	"github.com/pachyderm/pachyderm/v2/src/pfs"
 	"github.com/pachyderm/pachyderm/v2/src/pps"
 )
 
 // This code can all go away if we ever get the ability to run a PPS server without external dependencies
```
brycemcanally, line 10: I usually add a `TODO` for stuff like this.

src/pfs/pfs.proto
```diff
@@ -106,13 +107,6 @@ message Commit {
...
 }
 
-// CommitRange represents chain of commits with Lower being an ancestor of
-// Upper or, in the case of a range of size 1, the same commit.
-message CommitRange {
-  Commit lower = 1;
-  Commit upper = 2;
-}
-
 // CommitProvenance keeps track of where (i.e. which branch) a certain commit
```
brycemcanally, line 110: Do we still need this?

brycemcanally: Also, some of the comments in this file need to be updated.

src/pps/pps.proto
```diff
@@ -56,15 +56,17 @@ message Egress {
 
 message Job {
   string id = 1 [(gogoproto.customname) = "ID"];
+  Pipeline pipeline = 2;
```
brycemcanally, line 59: Really minor nit, but the pipeline should probably be the first field to match how we nest things in PFS.

src/pps/pps.proto
```diff
@@ -268,54 +270,50 @@ message StoredJobInfo {
...
-  string reason = 13;
-  google.protobuf.Timestamp started = 14;
-  google.protobuf.Timestamp finished = 15;
+  JobState state = 11;
+  string reason = 12;
+  google.protobuf.Timestamp started = 13;
+  google.protobuf.Timestamp finished = 14;
 }
 
 message JobInfo {
```
brycemcanally, line 279: Could we just nest the `StoredJobInfo` in this rather than having to copy the fields over?

brycemcanally, line 279: Would it make sense to just nest the `StoredJobInfo` rather than copy the fields over?

src/pps/pps.proto
```diff
@@ -378,8 +379,7 @@ message StoredPipelineInfo {
   // pachd). This allows the worker master to shard work correctly without
   // k8s privileges and without knowing the number of cluster nodes in the
   // Coefficient case.
-  uint64 parallelism = 7;
-  Pipeline pipeline = 8;
+  uint64 parallelism = 10;
 }
 
 message PipelineInfo {
```
brycemcanally, line 385: Same as the above for job info, nest rather than copy the fields.

src/pps/pps.proto
```diff
@@ -430,57 +430,40 @@ message PipelineInfo {
...
+  bool full = 3;
 }
 
 message InspectJobRequest {
   // Callers should set either Job or OutputCommit, not both.
   Job job = 1;
-  pfs_v2.Commit output_commit = 2;
-  bool block_state = 3; // block until state is either FAILURE or SUCCESS
-  bool full = 4;
+  bool block_state = 2; // block until state is either FAILURE or SUCCESS
```
brycemcanally, line 460: Probably should just call this `block` for consistency.

src/pps/pps.proto
```diff
@@ -502,9 +485,10 @@ message ListJobRequest {
   string jqFilter = 6;
 }
 
-message FlushJobRequest {
-  repeated pfs_v2.Commit commits = 1;
-  repeated Pipeline to_pipelines = 2;
+// Streams active jobs until canceled
```
brycemcanally, line 488: We should probably get the naming down for terminal vs non-terminal jobs. Open / active seem fine for the non-terminal jobs.

src/pps/pps.proto
```diff
@@ -513,21 +497,21 @@ message DeleteJobRequest {
 
 message StopJobRequest {
   Job job = 1;
-  pfs_v2.Commit output_commit = 2;
   string reason = 3;
 }
 
 message UpdateJobStateRequest {
```
brycemcanally, line 503: Yeah, I think there is an opportunity to decrease the number of fields in a lot of these protos by refactoring out some of the common fields in the Stored* protos.

src/server/pfs/cmds/cmds.go
```diff
@@ -453,21 +459,38 @@ $ {{alias}} foo@master --from XXX`,
 	shell.RegisterCompletionFunc(listCommit, shell.RepoCompletion)
 	commands = append(commands, cmdutil.CreateAlias(listCommit, "list commit"))
 
-	var repos cmdutil.RepeatedStringArg
-	flushCommit := &cobra.Command{
-		Use:   "{{alias}} <repo>@<branch-or-commit> ...",
-		Short: "Wait for all commits caused by the specified commits to finish and return them.",
-		Long:  "Wait for all commits caused by the specified commits to finish and return them.",
+	var branches cmdutil.RepeatedStringArg
+	waitCommit := &cobra.Command{
```
brycemcanally, line 463: The naming should be consistent between pachctl and the Go client. I kind of prefer wait because it seems a bit clearer, but I'm not strongly against block.

src/server/debug/server/server.go
```diff
@@ -369,7 +369,7 @@ func (s *debugServer) collectInputRepos(tw *tar.Writer, pachClient *client.APICl
 	for _, repoInfo := range repoInfos {
 		if _, err := pachClient.InspectPipeline(repoInfo.Repo.Name); err != nil {
 			// TODO: It would be better for this to be a structured error.
```
brycemcanally, line 371: This TODO can be removed.

src/server/pfs/cmds/cmds.go
brycemcanally: Also, it doesn't seem like a good idea to merge the commitset wait and commit wait into the same command. This code would be simpler if they were separate and I feel like it would be better to have a more explicit API.

src/server/pfs/cmds/cmds.go
```diff
@@ -562,41 +618,31 @@ $ {{alias}} test@master --new`,
...
 	subscribeCommit.Flags().AddFlagSet(rawFlags)
 	subscribeCommit.Flags().AddFlagSet(fullTimestampsFlags)
 	shell.RegisterCompletionFunc(subscribeCommit, shell.BranchCompletion)
 	commands = append(commands, cmdutil.CreateAlias(subscribeCommit, "subscribe commit"))
 
-	deleteCommit := &cobra.Command{
-		Use:   "{{alias}} <repo>@<branch-or-commit>",
-		Short: "Delete an input commit.",
-		Long:  "Delete an input commit. An input is a commit which is not the output of a pipeline.",
+	deleteCommitset := &cobra.Command{
```
brycemcanally, line 628: We should probably rename this to squash.

src/client/pps.go
Tryneus: I'm planning to do a pass soon to change this across all our functions

src/internal/collection/errors.go
Tryneus: accidentally a word - thanks!

src/internal/collection/types.go
Tryneus: That's basically how it's implemented internally, but I'd prefer not to have bool parameters in a function, they are very hard to read.  A wrapper could work, but it sacrifices the ergonomics of using the collection in a similar way.

src/internal/ppsutil/util.go
Tryneus: Yes, I ran into a circular dependency here =(

src/internal/testpachd/mock_transaction.go
Tryneus: When I wrote this, I was not optimistic that this would ever happen, and I still doubt it.

src/pfs/pfs.proto
Tryneus: We can totally get rid of this and it's on my list, but I didn't want to bloat this PR (hah).

src/server/pfs/pfs.go
```diff
@@ -224,3 +274,40 @@ func IsCommitNotFinishedErr(err error) bool {
...
+// IsCommitOnOutputBranchErr returns true if the err is due to an attempt to
+// start a commit on an output branch.
+func IsCommitOnOutputBranchErr(err error) bool {
+	if err == nil {
+		return false
+	}
+	return commitOnOutputBranchRe.MatchString(err.Error())
+}
+
+// IsParentNotFinishedErr returns true if the err is due to an attempt to
```
brycemcanally, line 306: It seems like the description of this error is different above (finishing commit on output with unfinished parent).

src/client/pps.go
Tryneus: Trying to be consistent with out API between PFS and PPS, and also a preference for avoiding boolean parameters if possible because they make code very hard to read.

src/client/pps.go
Tryneus: I think 'open' job works, but I'm open to using a different term.

src/client/pps.go
Tryneus: I don't think we were actually using `Full=true` anyway - removed and if you really need it you can use the lower-level interface (or we can go ahead with a more powerful functional interface).

src/pps/pps.proto
Tryneus: Ideally we could inline all the fields, but I don't think that's possible, and I don't think we want to sacrifice ergonomics for this.

src/server/pfs/cmds/cmds.go
Tryneus: Yeah, I'm planning to rename all the `Block.*` stuff to `Wait.*` after this (assuming everyone is on board) - definitely got some comments that 'block' isn't very clear about what it means.

src/pps/pps.proto
Tryneus: I think `active` sounds ok, but it implies that something is currently happening to it so I don't think it captures 'enqueued' jobs that are just waiting for parallelism or inputs.  'Open' kind of gets it, but maybe not as clear to other people?

src/internal/ppsutil/util.go
Tryneus: So we have some bits of code which expect that `Started` is set if `Finished` is set (for example, calculating the duration of a job).  I think it's reasonable to ensure that in the data structure.

As for where the job state is set to starting, I think that's actually a bug and it's skipping over the STARTED state in `registry.go` - I'll fix that.

src/server/pfs/server/api_server.go
```diff
@@ -194,14 +187,27 @@ func (a *apiServer) FinishCommit(ctx context.Context, request *pfs.FinishCommitR
 	defer func(start time.Time) { a.Log(request, response, retErr, time.Since(start)) }(time.Now())
 	if err := a.txnEnv.WithTransaction(ctx, func(txn txnenv.Transaction) error {
 		return txn.FinishCommit(request)
+	}, func(txnCtx *txncontext.TransactionContext) (string, error) {
+		// FinishCommit in a transaction by itself has special handling with regards
```
brycemcanally, line 191: How is this handled if you finish a commit in a transaction with other operations? It is not clear to me what the commitset ID should be in that case. Also, would it be possible to handle this through an alias for the finished commit so that we don't need to do this override thing, which seems to exist because we haven't found a good way to incorporate triggers / deferred processing into the global ID model.

src/server/pfs/server/trigger.go
```diff
@@ -16,86 +15,73 @@ import (
...
 					}
-					result = append(result, proto.Clone(commit.Branch).(*pfs.Branch))
-					headBranches[bi.Branch.Name] = true
 				}
 			}
 		}
-		return nil
+		return nil, nil
 	}
 	for _, b := range repoInfo.Branches {
```
brycemcanally, line 79: Not related to this PR, but this seems like a scalability problem for branches (noteworthy if we want to go the branch route for run pipeline).

src/server/pfs/server/trigger.go
```diff
@@ -16,86 +15,73 @@ import (
...
 		}
-		triggeredBranches[branch.Name] = true
+
 		bi := &pfs.BranchInfo{}
 		if err := d.branches.ReadWrite(txnCtx.SqlTx).Get(pfsdb.BranchKey(branch), bi); err != nil {
-			return err
+			return nil, err
 		}
+
+		triggeredBranches[branch.Name] = nil
```
brycemcanally, line 47: Isn't this necessarily nil if you reach here?

src/server/pps/server/api_server.go
```diff
@@ -474,13 +473,14 @@ func (a *apiServer) UpdateJobState(ctx context.Context, request *pps.UpdateJobSt
...
 	jobPtr := &pps.StoredJobInfo{}
-	if err := jobs.Get(request.Job.ID, jobPtr); err != nil {
+	if err := jobs.Get(ppsdb.JobKey(request.Job), jobPtr); err != nil {
 		return err
 	}
 	if ppsutil.IsTerminal(jobPtr.State) {
 		return ppsServer.ErrJobFinished{Job: jobPtr.Job}
 	}
 
+	jobPtr.Started = request.Started
```
brycemcanally, line 483: Related to what I mentioned above, this should probably not be updateable.

brycemcanally: Actually, maybe it should in the restart case?

src/server/pps/server/api_server.go
```diff
@@ -492,81 +492,23 @@ func (a *apiServer) UpdateJobStateInTransaction(txnCtx *txncontext.TransactionCo
...
-			request.Job = ji.Job
-			return nil
-		}); err != nil {
-			return nil, err
-		}
-		if request.Job == nil {
-			return nil, errors.Errorf("job with output commit %s not found", request.OutputCommit.ID)
-		}
+	// Make sure the job exists
+	// TODO: there's a race condition between this check and the watch below where
```
brycemcanally, line 504: We could potentially address this by having a timeout on the first event, which should return the job info as is when the watch is setup. Also, we could probably handle this by having an option on `WatchOne` that requires that the key exists.

src/server/pps/server/monitor.go
```diff
@@ -197,12 +197,10 @@ func (m *ppsMaster) monitorPipeline(ctx context.Context, pipelineInfo *pps.Pipel
 			defer close(ciChan)
 			return backoff.RetryNotify(func() error {
 				pachClient := m.a.env.GetPachClient(ctx)
-				return pachClient.SubscribeCommit(client.NewRepo(pipeline), "",
-					pipelineInfo.SpecCommit.NewProvenance(),
-					"", pfs.CommitState_READY, func(ci *pfs.CommitInfo) error {
-						ciChan <- ci
-						return nil
-					})
+				return pachClient.SubscribeCommit(client.NewRepo(pipeline), "", "", pfs.CommitState_READY, func(ci *pfs.CommitInfo) error {
```
brycemcanally, line 200: Was this changed just to play it safe and go through all of the commits? Also, this seems like it might hang if the context gets canceled while it is blocked on sending to `ciChan`?

src/server/worker/pipeline/spout/spout.go
```diff
@@ -7,6 +7,13 @@ import (
 
 // Run will run a spout pipeline until the driver is canceled.
 func Run(driver driver.Driver, logger logs.TaggedLogger) error {
+	// If the output branch head commit is open, close it
```
brycemcanally, line 10: Why are we doing this? The user should manage the state of the branches after a restart (they have full access to the repo through the pachclient with spouts 2.0). This would impose semantics on them that may not line up with their use case. Also, the branches should have semantics similar to source branches (I've started using this term rather than input because output branches can be inputs for downstream pipelines) rather than the output branches of transform pipelines.

src/server/pps/server/api_server.go
```diff
@@ -1598,16 +1504,36 @@ func (a *apiServer) writePipelineInfoToFileset(ctx context.Context, pipelineInfo
...
 func (a *apiServer) commitPipelineInfoFromFileset(
 	txnCtx *txncontext.TransactionContext,
 	pipelineName string,
 	filesetID string,
-	prevSpecCommit *pfs.Commit,
+	prevPipelineVersion uint64,
 ) (*pfs.Commit, error) {
+	// Make sure the StoredPipelineInfo is the same version we expect so we don't clobber another pipeline update
+	if prevPipelineVersion != 0 {
+		latestPipelineInfo, err := a.latestPipelineInfo(txnCtx, pipelineName)
```
brycemcanally, line 1523: We should restructure this such that the not found error propagates to here, rather than returning nil, nil in that case.

src/server/pps/server/api_server.go
```diff
@@ -1598,16 +1504,36 @@ func (a *apiServer) writePipelineInfoToFileset(ctx context.Context, pipelineInfo
...
+		if err != nil {
+			return nil, err
+		} else if latestPipelineInfo == nil {
+			return nil, newErrPipelineNotFound(pipelineName)
+		}
+		if prevPipelineVersion != latestPipelineInfo.Version {
```
brycemcanally, line 1529: :+1:

src/pps/pps.proto
```diff
@@ -364,12 +362,15 @@ enum PipelineState {
...
 message StoredPipelineInfo {
-  PipelineState state = 1;
-  string reason = 2;
-  pfs_v2.Commit spec_commit = 3;
-  map<int32, int32> job_counts = 4;
-  string auth_token = 5;
-  JobState last_job_state = 6;
+  Pipeline pipeline = 1;
+  uint64 version = 2;
+  pfs_v2.Commit original_spec_commit = 3; // The first spec commit for this version of the pipeline
```
brycemcanally, line 367: Can a pipeline version have multiple spec commits? I thought the purpose of using the fileset functionality for the spec commit was to make it transactional such that there is only one spec commit for a version. If that is the case, then we could probably just have this be `spec_commit`. Maybe this is needed for aliases?

src/server/pps/server/api_server.go
```diff
@@ -737,70 +703,50 @@ func (a *apiServer) listJob(
...
+		Finished:        jobPtr.Finished,
 	}
-	result.SpecCommit = specCommit
+
 	pipelinePtr := &pps.StoredPipelineInfo{}
-	if err := a.pipelines.ReadOnly(ctx).Get(jobPtr.Pipeline.Name, pipelinePtr); err != nil {
+	if err := a.pipelines.ReadOnly(ctx).Get(result.Job.Pipeline.Name, pipelinePtr); err != nil {
 		return nil, err
 	}
 	// Override the SpecCommit for the pipeline to be what it was when this job
```
brycemcanally, line 743: :+1:

src/server/pps/server/api_server.go
```diff
@@ -926,66 +866,50 @@ func (a *apiServer) StopJobInTransaction(txnCtx *txncontext.TransactionContext,
...
 	}
+
 	if commitInfo != nil {
 		if err := a.env.PfsServer().FinishCommitInTransaction(txnCtx, &pfs.FinishCommitRequest{
 			Commit: commitInfo.Commit,
 			Empty:  true,
 		}); err != nil && !pfsServer.IsCommitNotFoundErr(err) && !pfsServer.IsCommitDeletedErr(err) && !pfsServer.IsCommitFinishedErr(err) {
 			return err
 		}
-		statsCommit, err := ppsutil.GetStatsCommit(commitInfo)
```
brycemcanally, line 959: Just a note, be sure to account for the case where the job is for a service pipeline which doesn't have a meta commit.

brycemcanally: This PR has the change / more info: https://github.com/pachyderm/pachyderm/pull/6232

src/server/pps/server/api_server.go
```diff
@@ -1092,17 +1016,18 @@ func convertDatumState(state datum.State) pps.DatumState {
...
 	jobInfo, err := a.InspectJob(ctx, &pps.InspectJobRequest{
-		Job: client.NewJob(job.ID),
+		Job: job,
 	})
 	if err != nil {
 		return err
 	}
 	pachClient := a.env.GetPachClient(ctx)
-	fsi := datum.NewCommitIterator(pachClient, jobInfo.StatsCommit)
+	metaCommit := ppsutil.MetaCommit(jobInfo.OutputCommit)
```
brycemcanally, line 1025: Probably should handle a service pipeline job the same way as the PR referenced above. So, just return an error that the job does not have datums associated with it (even though it does, this probably needs more thought later).

src/server/pps/server/api_server.go
```diff
@@ -2313,6 +2225,18 @@ func setInputDefaults(pipelineName string, input *pps.Input) {
 	})
 }
 
+func (a *apiServer) stopAllJobsInPipeline(txnCtx *txncontext.TransactionContext, pipeline *pps.Pipeline) error {
+	// Using ReadWrite here may load a large number of jobs inline in the
```
brycemcanally, line 2229: This seems like something that should be handled by the pipeline controller (I think it already does in certain cases). Including all of the jobs in this transaction seems problematic for fast processing jobs, and I don't think it is necessary to stop all of the jobs in a single transaction. I may be missing something, but the way I am thinking about it is that we have one transaction that handles the state transition for the pipeline info, then the pipe controller sees that state transition and spins down the workers and stops the jobs, then spins up new workers.

src/server/pps/server/api_server.go
```diff
@@ -2194,37 +2103,41 @@ func (a *apiServer) CreatePipelineInTransaction(
 
 	// Create/update output branch (creating new output commit for the pipeline
 	// and restarting the pipeline)
-	if err := a.env.PfsServer().CreateBranchInTransaction(txnCtx, &pfs.CreateBranchRequest{
-		Branch:     outputBranch,
-		Provenance: provenance,
-		Head:       outputBranchHead,
-	}); err != nil {
-		return errors.Wrapf(err, "could not create/update output branch")
+	if update && request.Reprocess {
```
brycemcanally, line 2106: This condition should be negated and just be a single block with what is currently in the else block.

brycemcanally: How are we breaking the commit chain if we don't update the branch in the update && reprocess case?

src/server/pps/server/api_server.go
```diff
@@ -2348,7 +2272,8 @@ func (a *apiServer) inspectPipelineInTransaction(txnCtx *txncontext.TransactionC
 		}
 		return nil, err
 	}
-	pipelinePtr.SpecCommit.ID = ancestry.Add(pipelinePtr.SpecCommit.ID, ancestors)
+	// TODO(global ids): how should we treat ancestry here?  Alias commits are going to gum it up.
```
brycemcanally, line 2275: I guess this is a case where having the alias commits not be in the commit chain would be helpful.

src/server/pps/server/api_server.go
```diff
@@ -2758,193 +2659,57 @@ func (a *apiServer) StopPipeline(ctx context.Context, request *pps.StopPipelineR
...
-				Branch:     client.NewSystemRepo(request.Pipeline.Name, pfs.MetaRepoType).NewBranch("master"),
-				Provenance: append(provenance, newCommit.NewProvenance()),
-			})
-			if err != nil {
-				return err
-			}
-		}
-		return nil
+		// Kill any remaining jobs
+		return a.stopAllJobsInPipeline(txnCtx, request.Pipeline)
```
brycemcanally, line 2699: Similar to above, I think the pipeline controller should manage this.

src/server/pps/server/api_server.go
```diff
@@ -3008,6 +2773,60 @@ func (a *apiServer) RunCron(ctx context.Context, request *pps.RunCronRequest) (r
...
+		// Check if there is an existing job for the output commit
+		job := client.NewJob(pipelineInfo.Pipeline.Name, txnCtx.CommitsetID)
+		jobInfo := &pps.StoredJobInfo{}
+		if err := a.jobs.ReadWrite(txnCtx.SqlTx).Get(ppsdb.JobKey(job), jobInfo); err == nil {
+			continue // Job already exists, skip it
+		} else if !col.IsErrNotFound(err) {
+			return err
+		}
+
+		if commitInfo.Origin.Kind == pfs.OriginKind_ALIAS || commitInfo.Finished != nil {
```
brycemcanally, line 2811: This check can be a part of the initial check in the loop.

src/server/pps/server/api_server.go
```diff
@@ -3008,6 +2773,60 @@ func (a *apiServer) RunCron(ctx context.Context, request *pps.RunCronRequest) (r
...
+		// Skip commits from repos that have no associated pipeline
+		pipelineInfo := &pps.StoredPipelineInfo{}
+		if err := a.pipelines.ReadWrite(txnCtx.SqlTx).Get(commitInfo.Commit.Branch.Repo.Name, pipelineInfo); err != nil {
+			if col.IsErrNotFound(err) {
+				continue
+			}
+			return err
+		}
+
+		// TODO(global ids): don't create the jobs for certain states or for spouts (can't detect spouts without loading pipeline spec?)
```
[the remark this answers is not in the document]
brycemcanally: Would we get new commits for stopped pipelines? It seems like we wouldn't see any commits for stopped pipelines here?

src/server/pfs/server/api_server.go
Tryneus: > How is this handled if you finish a commit in a transaction with other operations?
If there is a user-level transaction, this will use the transaction ID.

> would it be possible to handle this through an alias for the finished commit
Yes, that is what this does, it's just that aliases are written to always use the transaction ID, so this overrides the transaction ID.

src/server/pfs/server/trigger.go
Tryneus: It would be `nil` if you did a read, but it isn't present in the map - this adds an entry to mark it as processed.

src/server/pps/server/api_server.go
Tryneus: In general, we shouldn't have code that is changing `Started` from a value to another value, this RPC just allows you to sync the state in the worker with the state in the collection transactionally with other calls.

src/server/pps/server/api_server.go
Tryneus: Yes, we've had some discussions about how to fix this, but this was an existing problem before so I consider it out of scope right now - this TODO was just to make the problem more visible.  I was thinking of maybe just having a new empty event type to indicate when the initial values are done which may be more generally useful?

src/server/worker/pipeline/spout/spout.go
Tryneus: Since branches now have default head commits, PFS generates an open commit for the spout's output branch.  There might be other ways to solve this, but PFS has no idea if the output branch is for a spout or other type of pipeline.  Open to suggestions.

src/server/pps/server/monitor.go
Tryneus: This was changed because we can no longer subscribe to commits by their provenance.  I think your concerns about this hanging are valid but out-of-scope here - that was an existing problem.

src/pps/pps.proto
Tryneus: So I removed the part in `StartPipeline` and `StopPipeline` where they would create new spec commits, but alias commits are technically commits and we should have plenty of those in the spec repo - so this records the original spec commit from when the pipeline was updated or created.

src/server/pps/server/api_server.go
Tryneus: So this function definitely gracefully handles commits not being found (it biases towards just getting everything to a closed state rather than erroring if something isn't found).

src/server/pps/server/api_server.go
Tryneus: This is what the 'pipeline finisher' used to do at the end of transactions in PFS and I agree it's not great, but it's better than it was before.  Before, it would query all of the output commits in the output repo (_outside of the transaction_), then attempt to stop the job associated with each output commit inside the transaction (leading to false positives and possible race conditions).  I'm up for rethinking this, but I'd prefer to tackle that separately.

src/server/pps/server/api_server.go
Tryneus: We do not break the commit chain, is that necessary?

src/server/pps/server/api_server.go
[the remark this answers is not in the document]
Tryneus: Yeah, I was thinking of adding a pipeline type field - I'll go ahead with that.

src/server/pfs/server/driver.go
```diff
@@ -811,66 +565,131 @@ func (d *driver) finishCommit(txnCtx *txncontext.TransactionContext, commit *pfs
...
+			return nil, err
+		}
+		// No commit already exists, create a new one
+		// First load the parent commit and update it to point to the child
+		parentCommitInfo := &pfs.CommitInfo{}
+		if err := d.commits.ReadWrite(txnCtx.SqlTx).Update(pfsdb.CommitKey(parent), parentCommitInfo, func() error {
+			if parentCommitInfo.Finished == nil {
+				// We allow aliases on the same branch as the original
+				if !proto.Equal(branch, parent.Branch) {
+					return errors.Errorf("cannot create an alias for an open commit from a different branch: %s -> %s", pfsdb.CommitKey(parent), pfsdb.BranchKey(branch))
```
brycemcanally, line 650: Why do we not allow this?

src/server/pfs/server/driver.go
```diff
@@ -954,169 +771,91 @@ func (d *driver) propagateCommits(sqlTx *sqlx.Tx, commitsetID string, branches [
...
-		// If the only branches in the hypothetical output commit's provenance are
-		// in the 'spec' repo, creating it would mean creating a confusing "dummy"
-		// commitset with no non-spec input data. If this is the case, don't create a new
-		// output commit
-		allSpec := true
-		for _, p := range newCommitProvMap {
-			if p.Commit.Branch.Repo.Type != pfs.SpecRepoType {
-				allSpec = false
-				break
+		if allSameString(ids) {
```
brycemcanally, line 793: :+1:

src/server/pfs/server/driver.go
```diff
@@ -1149,35 +892,39 @@ func (d *driver) inspectCommit(ctx context.Context, commit *pfs.Commit, blockSta
...
-			d.inspectCommit(ctx, p.Commit, pfs.CommitState_FINISHED)
-		}
-	}
-	if blockState == pfs.CommitState_FINISHED {
-		// Watch the CommitInfo until the commit has been finished
-		if err := d.commits.ReadOnly(ctx).WatchOneF(pfsdb.CommitKey(commit), func(ev *watch.Event) error {
-			if ev.Type == watch.EventDelete {
-				return pfsserver.ErrCommitDeleted{
-					Commit: commit,
+	if commitInfo.Finished == nil {
```
brycemcanally, line 895: :+1: for the refactor of the blocking functionality.

src/internal/collection/types.go
brycemcanally: This could also just be an option on `GetByIndex`. My concern with functions like this is interface bloat, but I'm not really strongly against this approach.

src/internal/testpachd/mock_transaction.go
brycemcanally: It seems like it should either be a TODO or removed. I don't think it is really adding value otherwise.

src/pps/pps.proto
brycemcanally: It doesn't seem like a bad idea to have consistent terminology between commits and jobs, particularly because they are mapped 1:1. So open commit and open job seem like good terms.

src/server/pps/server/api_server.go
brycemcanally: If we don't want to change the started field after it is first set, then we should probably remove this and lean on `ppsutil.UpdateJobState` to set it when the job is transitioned to the starting state.

src/server/pps/server/api_server.go
brycemcanally: Okay, yeah we can leave this as a TODO for now, and circle back later.

src/server/pps/server/monitor.go
brycemcanally: Ohh right, forgot that was removed. I'll add a card for not respecting the context.

src/server/worker/pipeline/spout/spout.go
brycemcanally: It looks like you already made this change, but what makes sense to me is for the empty commit to be finished at creation time. It looks like you made a change to finish the empty commit when there is no provenance or when the only provenance is a branch in the spec repo.

src/pps/pps.proto
brycemcanally: Okay, it does seem a bit confusing though. I'm not sure what the benefit of the "original" part is. Do we ever need to consider the original vs alias commits, or do we just grab the spec commit when we need a commit that has the correct spec file?

src/server/pps/server/api_server.go
brycemcanally: So this will just error out based on what was discussed in the thread above, which should be fine.

src/server/pps/server/api_server.go
brycemcanally: Yeah, this is one that we should get a handle on before 2.0 GA, but not this PR. I'll add a card.

src/server/pps/server/api_server.go
brycemcanally: Ohh, that is right. We don't need to break the chain anymore because we will perform the correct deletions on the parent to remove the old state.

PR #6266, opened 2021-06-07
[1.13.x] Fix collecting metrics with auth enabled

actgardner:
Make metrics work in 1.x by adding an unauthenticated internal interface for listing repos and pipelines. This isn't an RPC.

To add support for cross-service method calls that aren't RPCs, this PR applies the same refactor to 1.13 as we did in 2.x - create a new interface for each service in `src/server/<service>` which combines the RPC interface, transaction interface and a new internal interface for calls within pachd.

To avoid circular dependencies we need to split out the TransactionContext struct into it's own module. That makes up the bulk of the changes, just find-and-replacing.

### DIFF
--- a/src/server/auth/iface.go
+++ b/src/server/auth/iface.go
@@ -1,0 +1,30 @@
+package auth
+
+import (
+	"github.com/pachyderm/pachyderm/src/client/auth"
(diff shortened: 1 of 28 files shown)


Tryneus (approved):
LGTM

PR #6271, opened 2021-06-07
[S3G] Parse commit_id from bucket name

### LINKED ISSUE #6235: S3G: Need a way to list files at a certain commit id

As part of an integration, I need to be able to see which files exist in a pachyderm repo as of a specific commit id via the S3 Gateway. You can fetch specific versions by passing the S3 object version id with an "object get" request, but there's no way to pass this version for an "object list" request today.

(There's a bit of an API impedence mismatch here because S3 assigns version ids to individual objects, whereas Pachyderm versions whole repos by commit id.)

Ref: https://docs.pachyderm.com/latest/deploy-manage/manage/s3gateway/#versioning

Not that I know of.

** What is your proposal for a feature to solve this?**

Currently bucket names of the form `<branch>.<repo>` are supported. I propose adding a new `<commit>.<branch>.<repo>` bucket name. The S3G logic would be fairly straightforward: if there are two `.`s in the bucket name, use the new addressing format. Otherwise, fall back to the old one.

albscui:
This PR allows S3 gateway clients to list objects by referencing a commit_id.

Resolves #6235

To-Do:

- [x] with respect to Global IDs, how should we disambiguate commits within the same repo that have the same Id but different branches? Something like `commit.branch.repo`?
- [ ] the 63 char limit imposed by S3 clients is out of our control, so we should think about supporting short commit_ids as part of the query?

### DIFF
--- a/src/server/pfs/s3/master_test.go
+++ b/src/server/pfs/s3/master_test.go
@@ -320,7 +320,8 @@ func masterListObjectsPaginated(t *testing.T, pachClient *client.APIClient, mini
 	// `LastModified` date is correct. A few minutes are subtracted/added to
 	// each to tolerate the node time not being the same as the host time.
 	startTime := time.Now().Add(time.Duration(-5) * time.Minute)
-	repo := tu.UniqueString("testlistobjectspaginated")
+	// bucket name cannot exceed 63 characters long
+	repo := tu.UniqueString("testLOP")
 	require.NoError(t, pachClient.CreateRepo(repo))
 	commit, err := pachClient.StartCommit(repo, "master")
 	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


src/server/pfs/s3/master_test.go
```diff
@@ -320,7 +320,8 @@ func masterListObjectsPaginated(t *testing.T, pachClient *client.APIClient, mini
 	// `LastModified` date is correct. A few minutes are subtracted/added to
 	// each to tolerate the node time not being the same as the host time.
 	startTime := time.Now().Add(time.Duration(-5) * time.Minute)
-	repo := tu.UniqueString("testlistobjectspaginated")
+	// bucket name cannot exceed 63 characters long
+	repo := tu.UniqueString("testLOP")
```
jdoliner, line 324: This change was necessary because s3 puts a hard limit of 63 chars on the length of a bucket and the repo name was too long.

jdoliner (approved):
Looks like this is all passing and the code all LGTM. @lukemarsden if you want to make a pass feel free but I think one approval should be plenty for this.

jdoliner:
Wrt global ids I think the syntax of `commit.branch.repo` makes sense to me. That shouldn't be too hard to add to the parsing. @Tryneus might have an opinion on this.

Wrt the shortened commit names that seems like a reasonable idea but probably better to do it later so this can be merged. I think there's some interesting issues with it because there's a number of places where we depend on being able to identify commit ids which is a lot harder if you can abbreviated them. I think the 63 char limit should be possible for people to work within for now.

PR #6276, opened 2021-06-08
[1.13.x] Pipeline operations ignore errors in revoking auth tokens

actgardner:
Jimmy reported this issue with a hub cluster:

```
$ pachctl delete pipeline raw_data
error revoking old auth token:  pachyderm/1.7.0/pachyderm_auth/tokens/faa0c5df1d27cedaa87fa55e417e295c019f1ba6603afda83b7832fa8d59b3f6 not found
```

In 1.13 revoking auth tokens and updating the pipeline isn't atomic, so it's possible for a pipeline to end up with an invalid token in etcd. If the pipeline's token is invalid it throws an error when updating or deleting.

In 2.x we do this all in one DB transaction (revoke the token,  generate the new one, add it to the pipeline spec) so it's not an issue.

### DIFF
--- a/src/server/auth/server/api_server.go
+++ b/src/server/auth/server/api_server.go
@@ -2233,9 +2233,10 @@ func (a *apiServer) RevokeAuthTokenInTransaction(txnCtx *txnenv.TransactionConte
 	tokens := a.tokens.ReadWrite(txnCtx.Stm)
 	var tokenInfo auth.TokenInfo
 	if err := tokens.Get(auth.HashToken(req.Token), &tokenInfo); err != nil {
-		if !col.IsErrNotFound(err) {
-			return nil, err
+		if col.IsErrNotFound(err) {
+			return nil, auth.ErrRevokeUnknownToken
 		}
+		return nil, err
 	}
 	if !isAdmin && tokenInfo.Subject != callerInfo.Subject {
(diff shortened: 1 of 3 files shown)


src/server/auth/server/api_server.go
```diff
@@ -2233,9 +2233,10 @@ func (a *apiServer) RevokeAuthTokenInTransaction(txnCtx *txnenv.TransactionConte
 	tokens := a.tokens.ReadWrite(txnCtx.Stm)
 	var tokenInfo auth.TokenInfo
 	if err := tokens.Get(auth.HashToken(req.Token), &tokenInfo); err != nil {
-		if !col.IsErrNotFound(err) {
-			return nil, err
+		if col.IsErrNotFound(err) {
+			return nil, auth.ErrRevokeUnknownToken
 		}
```
msteffen, line 2238: I almost wonder if the old code was just supposed to return here in the case where the token isn't found (making `DeleteToken` idempotent); that it doesn't might be a bug. The fix in this PR makes sense to me, though.

msteffen (approved):
LGTM

src/server/auth/server/api_server.go
actgardner: Yeah I could see this originally just silently dropping `col.IsErrNotFound`.

PR #6278, opened 2021-06-08
Remove hub as target from Reset.py

chainlink:
(Hub endpoints removed 6+ months ago, so likely no one is using this)


actgardner (approved):
A good start 🔥

PR #6279, opened 2021-06-08
Deflake TestListDatumDuringJob 

acohen4:
Increasing Datum task length to assert 0 datums are retrieved before they complete. 

The current timeout taken in each datum is short enough that sometimes it is retrieved in queries immediately after the job starts. Here we are just making the datums take longer to complete to assert that unfinished datums are not returned in the list datums query.


brycemcanally (approved):
lgtm

PR #6281, opened 2021-06-08
Prevent crashing when canceling a put file request

Tryneus:
If a `PutFile` attempts to acquire from the memory limiter with a canceled or otherwise errored context, this code was not checking the return value of the `Acquire` and it would continue as if successful.  This could then result in releasing more than was acquired in the memory limiter semaphore, which would crash `pachd` with the error: `panic: semaphore: released more than held`.


msteffen (approved):
Nice catch!

I think alysha's change to add golanglint-ci could start warning on unchecked errors, but we'd have to do a pass over the codebase to deal with them all

actgardner:
Yeah, we have a lot of errors that we don't currently check, I need to do a follow-on to enable that linter. This is also 1.13 and golanglint-ci is only running on 2.0.

PR #6282, opened 2021-06-08
Don't install TLS certs into the worker container

actgardner:
The way our build process works today, we grab all the TLS certs from our Go build image and bundle them into a Go file, compile them into the worker binary, then extract them into the user-supplied container for the worker process.

It's hard to figure out why we started doing this - it looks like the PR that added bindata was just replicating older functionality. Since the worker container is supplied by the user, it's surprising to me that they wouldn't just use a container that has the necessary certificates?


msteffen (approved):
LGTM

Recording convo from slack, for posterity:
> 2.0 seems like a good opportunity to drop this and see if anyone notices. Also...if we are going to build the certs into the binary, calling AppendCertsFromPEM() (or whatever) directly seems better (and definitely more non-root-container-friendly) than writing them to disk and then reading them back

PR #6292, opened 2021-06-09
Use goreleaser to build binaries outside of Docker

actgardner:
Use goreleaser to build all the binaries outside of Docker. This should be easy to extend to supporting ARM versions of our images and binaries. It could also be extended to replace our homebrew release script, uploading binaries to Github, etc. This removes the pachyderm_build docker image.

The downside is that users have to have Go installed locally to build the binaries (they probably already do, because `make install` uses their local Go). This version is used to build the binaries, which introduces a potentially inconsistency in releases, but the solution to this is to stop doing releases from developer laptops ;)

This shaves up to 5 minutes off the build step in CI, depending on how warm the cache is.


chainlink:
This cleans things up nicely! I'm a little nervous about not compiling within a consistent environment, but I don't think that really matters with go? (The only thing I can really think of it adding a go version check to make sure we're consistent there)

chainlink:
Also, I'm happy with alpine, but that may create a problem with RH container cert

chainlink:
A potential follow on here is to make a release GH action (which would get releases off dev laptops and allow for a consistent env) Happy to work on that after this lands if that seems like a good direction.

actgardner:
> A potential follow on here is to make a release GH action (which would get releases off dev laptops and allow for a consistent env) Happy to work on that after this lands if that seems like a good direction.

Yeah this would be great, this is definitely the end goal :)

> Also, I'm happy with alpine, but that may create a problem with RH container cert

I pulled alpine out of this so we can deal with one thing at a time.

chainlink (approved):

PR #6293, opened 2021-06-09
[S3G] Support commit.repo bucket name

### LINKED ISSUE #6235: S3G: Need a way to list files at a certain commit id

albscui:
Previously merged a similar update in `master` https://github.com/pachyderm/pachyderm/pull/6271
This is just to achieve the same functionality in the 1.13.x branch.

Resolves #6235

### DIFF
--- a/src/server/pfs/s3/driver.go
+++ b/src/server/pfs/s3/driver.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/pachyderm/pachyderm/src/client"
 	"github.com/pachyderm/pachyderm/src/client/pkg/errors"
+	"github.com/pachyderm/pachyderm/src/server/pkg/uuid"
 
 	"github.com/gogo/protobuf/types"
 	"github.com/pachyderm/s2"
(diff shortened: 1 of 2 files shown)


lukemarsden (changes requested):
Let's add support for `commit.branch.repo` syntax. This way rslex won't have to change how it addresses the S3 gateway.

src/server/pfs/s3/driver.go
```diff
@@ -70,25 +73,29 @@ func (d *MasterDriver) listBuckets(pc *client.APIClient, r *http.Request, bucket
...
 		Repo:   repo,
-		Commit: branch,
+		Branch: branch,
+		Commit: commit,
 		Name:   name,
 	}, nil
 }
 
 func (d *MasterDriver) bucketCapabilities(pc *client.APIClient, r *http.Request, bucket *Bucket) (bucketCapabilities, error) {
-	branchInfo, err := pc.InspectBranch(bucket.Repo, bucket.Commit)
```
albscui, line 91: This broke the `<commit>.<repo>` syntax, because it was trying to inspect a branch but given commit_id instead.

lukemarsden (approved):
I guess it would be nice to have a test that this works for non-master branches but don't let that block the merge. LGTM, thanks so much!

PR #6300, opened 2021-06-09
Reorganize GlobalConfiguration to make enterprise config cleaner

actgardner:
`PACHD_POD_NAME` and `STORAGE_BACKEND` were required properties for the enterprise server, but we don't actually use them. This pulls the configs that affect the enterprise server and pachd into GlobalConfiguration - none of them are required so this doesn't change the behaviour for other pods that use the global configuration.


chainlink (approved):
LGTM

PR #6302, opened 2021-06-09
Remove old compatibility system

chainlink:
The current compatibility system is brittle, as it relies on parsing and inspecting go code (which is being removed anyway) It's also mainly targeted at dash 1.x and IDE for pach 1.x

Dash 2.x compatibility will be maintained and tested separately (still tbd, but will likely rely on the forthcoming proto npm packages) 

For IDE : with the move to Pach 2.x we're removing much of the bespoke code (ie. pachyderm authenticator) in favour of built in functionality (oauth) meaning we don't need to tightly control the compatibility as much. Extension compatibility will also be maintained separately (when we get there)


nitinjainsj (approved):
Can you also update the docs/release_instrcutions.md to remove the step to build compatiibility.

Issue #5799, opened 2021-03-10
Allow limiting session duration when using OIDC authentication


RaananHadar:
**What is the goal / desired outcome?**
OIDC session duration in Pachyderm is hard coded to 30 days. This means that even if a user logged out, they would be allowed to instantly log in without needing authentication. This is also applicable the Pachyderm IDE. From a security perspective, this is problematic if another person were to use a previously authenticated computer or if the computer was lost.

**If there is a way to accomplish this today via workaround, what does that require?**
Use SAML.

**(Optional) What is your proposal for a feature to solve this?**
Add support to limiting the session duration. It would be beneficial if this was configurable via the command line similar to the current `--ttl` flag which limits the token lifetime. For example, by using a command line flag or optionally if this was supplied by the OIDC provider.

mindthevirt:
@RaananHadar this should be doable once 2.0 will be released.

mindthevirt:
<div>Agent Dale Georg linked Freshdesk ticket <a href="https://pachyderm-support.freshdesk.com/a/tickets/266">266</a> for this issue.</div>

PR #6308, opened 2021-06-10
Make auth token duration configurable

actgardner:
Users have requested the ability to set the auth token duration shorter than 30 days for compliance reasons.

This is half of the solution to https://github.com/pachyderm/pachyderm/issues/5799, the other half is reducing the expiration on access tokens in dex.


acohen4 (approved):
Looks good!

PR #6310, opened 2021-06-10
Support shortening the id token lifespan in dex

actgardner:
Support configuring the expiration time for dex id tokens, which applies to other apps besides pachd.


src/server/identity/server/api_server_test.go
```diff
@@ -274,3 +275,39 @@ func TestIDPConnectorCRUD(t *testing.T) {
...
+	tu.DeleteAll(t)
+	defer tu.DeleteAll(t)
+
+	tu.ConfigureOIDCProvider(t)
+
+	adminClient := tu.GetAuthenticatedPachClient(t, auth.RootUser)
+	_, err := adminClient.SetIdentityServerConfig(adminClient.Ctx(), &identity.SetIdentityServerConfigRequest{
+		Config: &identity.IdentityServerConfig{
+			Issuer:        "http://localhost:30658/",
+			IdTokenExpiry: "1h",
```
acohen4, line 294: maybe this is a trivial question, but is it worth calling out somewhere the format of the Expiry string?

src/server/identity/db.go
```diff
@@ -32,3 +32,11 @@ CREATE TABLE IF NOT EXISTS identity.config (
 `)
 	return err
 }
+
+// AddTokenExpiryConfig adds expiry fields for token lifespan to the server config
+func AddTokenExpiryConfig(ctx context.Context, tx *sqlx.Tx) error {
+	_, err := tx.ExecContext(ctx, `
+ALTER TABLE identity.config ADD COLUMN
+	id_token_expiry VARCHAR(4096)`)
```
acohen4, line 40: maybe this is a trivial question, but is it worth calling out somewhere the format of the Expiry string?

acohen4 (approved):
Left a small doc related question, but looks good!

src/server/identity/db.go
actgardner: @nadegepepin This does add a new configuration option for the identity server we'll need to document - it takes a Go duration string (https://golang.org/pkg/time/#ParseDuration)

PR #6332, opened 2021-06-14
ModifyFile protocol change

brendoncarroll:
Changes ModfiyFile so that paths can be set in the same message as any source.
Removes TARSource because it does not fit into the source model well.
Set commit for the stream in separate message.  This also simplifies the read loop.


src/pfs/pfs.proto
```diff
@@ -267,57 +267,13 @@ enum Delimiter {
...
 
-message PutFile {
-  bool append = 1;
+message AddFile {
+  string path = 1;
   string tag = 2;
   oneof source {
-    RawFileSource raw_file_source = 3;
-    TarFileSource tar_file_source = 4;
+    google.protobuf.BytesValue raw_file_source = 4;
```
brycemcanally, line 274: Start at 3 here. It might also make sense to nest the URLFileSource type and just have it be URL, then you could just name the fields raw / data and url.

src/pfs/pfs.proto
```diff
@@ -327,7 +283,7 @@ message URLFileSource {
 }
 
 message DeleteFile {
-  string file = 1; 
+  string path = 1;
```
brycemcanally, line 286: :+1:

src/pfs/pfs.proto
```diff
@@ -339,11 +295,13 @@ message CopyFile {
...
-  Commit commit = 1;
-  oneof modification {
-    PutFile put_file = 2;
+  oneof body {
+    Commit set_commit = 1;
+    AddFile add_file = 2;
     DeleteFile delete_file = 3;
     CopyFile copy_file = 4;
+
+    google.protobuf.BytesValue tar_data = 5;
```
brycemcanally, line 304: Lets just get rid of this. It doesn't make sense in this model and I don't think it is too much of a lift to handle this in the client (going from tar to add files).

src/client/option.go
```diff
@@ -2,20 +2,25 @@ package client
 
 import "github.com/pachyderm/pachyderm/v2/src/pfs"
 
+type putFileConfig struct {
```
brycemcanally, line 5: Maybe we should just expose add file and have that be the append path. That would let us get rid of this, and put would always be delete followed by add.

src/client/pfs_file.go
```diff
@@ -22,10 +24,10 @@ func (c APIClient) PutFile(commit *pfs.Commit, path string, r io.Reader, opts ..
 	})
 }
 
-// PutFileTar puts a set of files into PFS from a tar stream.
-func (c APIClient) PutFileTar(commit *pfs.Commit, r io.Reader, opts ...PutFileOption) error {
+// PutFileTAR puts a set of files into PFS from a tar stream.
```
brycemcanally, line 27: :+1:

src/client/pfs_file.go
```diff
@@ -123,38 +125,35 @@ type modifyFileCore struct {
...
+			}
 		}
-		if err := mfc.sendPutFile(pf); err != nil {
+		if err := mfc.sendPutFile(&pfs.AddFile{
+			Path: path,
+			Tag:  config.tag,
+		}); err != nil {
 			return err
 		}
 		if _, err := grpcutil.ChunkReader(r, func(data []byte) error {
```
brycemcanally, line 147: Nit, you can just return the err.

src/client/pfs_file.go
```diff
@@ -123,38 +125,35 @@ type modifyFileCore struct {
...
+		if !config.append {
+			if err := mfc.sendDeleteFile(&pfs.DeleteFile{
+				Path: path,
+				Tag:  config.tag,
+			}); err != nil {
+				return err
+			}
 		}
-		if err := mfc.sendPutFile(pf); err != nil {
+		if err := mfc.sendPutFile(&pfs.AddFile{
```
brycemcanally, line 141: This is still operating on a stream. I think you should have an add file message that is repeatedly updated and sent with the data for each chunk. We are not operating on the messages independently with this approach.

src/client/pfs_file.go
```diff
@@ -694,3 +722,22 @@ func (c APIClient) WalkFile(commit *pfs.Commit, path string, cb func(*pfs.FileIn
 		}
 	}
 }
+
+// This is duplicated from the object package to remove a circular dependency.
+func withPipe(wcb func(w io.Writer) error, rcb func(r io.Reader) error) error {
```
brycemcanally, line 727: We should refactor this into a utility package.

src/server/pfs/server/api_server.go
```diff
@@ -384,97 +387,87 @@ type modifyFileSource interface {
...
-				}
-			case *pfs.ModifyFileRequest_CopyFile:
-				cf := mod.CopyFile
-				if err := a.driver.copyFile(ctx, uw, cf.Dst, cf.Src, cf.Append, cf.Tag); err != nil {
-					return bytesRead, err
-				}
-			}
+// modifyFile reads from a modifyFileSource until io.EOF and writes changes to an UnorderedWriter.
+// SetCommit messages will result in an error.
+func (a *apiServer) modifyFile(ctx context.Context, uw *fileset.UnorderedWriter, server modifyFileSource) (int64, error) {
```
brycemcanally, line 392: We should rewrite this to just apply each message as they come in to the buffering layer. We can get rid of the majority of this logic if we do that.

src/client/pfs_file.go
brendoncarroll: Really the object package shouldn't depend on the client.

src/server/pfs/server/api_server.go
brendoncarroll: You want to force the client to provide the path and tag for each message?

src/client/pfs_file.go
brycemcanally: It shouldn't be in the client. It should be in a separate utility package that both the client and obj package can import and use.

src/server/pfs/server/api_server.go
brycemcanally: Yeah, that is what I thought we agreed on? We make each message an independent operation so we don't need to manage streams on the client or server side.

src/client/option.go
brendoncarroll: I think the client should mirror the CLI.  If we have put file in the CLI then we should have it in here too.

brycemcanally: I agree with that. I didn't mean get rid of put file, just this extra config layer that is used to handle the append case. If we exposed add, delete, and put then we can get rid of the append option for put.

src/client/pfs_file.go
brendoncarroll: I think that it's probably a good idea to assume the last path so we don't have to keep resending it.  This will come in hand especially for large files.  The cost in server implementation complexity is 2 if statements.

src/pfs/pfs.proto
brendoncarroll: What about the "recursive" url flag?

src/client/pfs_file.go
brendoncarroll: What should we call the utililty package?

src/server/pfs/server/api_server.go
brendoncarroll: I guess we can always add the other behavior later.

src/client/pfs_file.go
brycemcanally: Resending the path is negligible when we can fit ~20MB of data in each message, and it is simple on the client side and simpler on the server side. I'm surprised you're not in favor of this because it will make it much easier to ensure that each message in the stream is setup properly since we can validate them in isolation.

src/client/option.go
brendoncarroll: The CLI has append as an option on put though.

brycemcanally: Yeah, we can change that.

src/pfs/pfs.proto
brycemcanally: It should be captured in the nested URL type.

brendoncarroll: I don't really understand why you would want to nest anything.  There is no type nesting in Go, so it just obscures the name.  It seems like it's a feature that might work well in Java.

src/client/pfs_file.go
brendoncarroll: Okay, will remove.

src/pfs/pfs.proto
brycemcanally: It would give us an easier way to scope the type rather than having to suffix with `FileSource`. It would just be a field with type AddFile_URL and name url in the source oneof. Take a look at `PhoneNumber` here: https://developers.google.com/protocol-buffers/docs/gotutorial.

src/client/option.go
brycemcanally: We discussed this offline. We are going to circle back to this later.

brendoncarroll: We spoke offline about this.  CLI and client facade changes are out of scope of this P.R.

src/client/pfs_file.go
brycemcanally: Hmm, none of the packages we currently have really jump out as the right one to put it in. We could just put it in a package called pipeutil or maybe a package called goutil to meme on Go stuff we need utilities for lol.

src/client/pfs_file.go
```diff
@@ -171,61 +174,86 @@ func (mfc *modifyFileCore) maybeError(f func() error) (retErr error) {
...
-			return err
-		}
-		_, err := grpcutil.ChunkReader(r, func(data []byte) error {
-			return mfc.sendPutFile(&pfs.PutFile{
-				Source: &pfs.PutFile_TarFileSource{
-					TarFileSource: &pfs.TarFileSource{
-						Data: data,
-					},
-				},
+		return tarutil.Iterate(r, func(f tarutil.File) error {
```
brycemcanally, line 191: We don't need to use tarutil for this. It is useful when we need to convert to a write, but the standard tar package has good support for tar readers, which is what we need here. Also, as a side note, you need to set the stream flag when using tarutil.Iterate or else it is going to buffer the files in memory (and yes, we should make that a required parameter).

src/server/pfs/server/api_server.go
```diff
@@ -384,97 +388,63 @@ type modifyFileSource interface {
...
+			if err := deleteFile(uw, mod.DeleteFile); err != nil {
+				return bytesRead, err
+			}
+		case *pfs.ModifyFileRequest_CopyFile:
+			cf := mod.CopyFile
+			if err := a.driver.copyFile(ctx, uw, cf.Dst, cf.Src, cf.Append, cf.Tag); err != nil {
+				return bytesRead, err
+			}
+		case *pfs.ModifyFileRequest_SetCommit:
+			return bytesRead, errors.Errorf("cannot set commit")
```
brycemcanally, line 432: It might be clearer to change this to "cannot set commit more than once".

src/server/pfs/server/api_server.go
```diff
@@ -740,16 +676,12 @@ func (a *apiServer) Fsck(request *pfs.FsckRequest, fsckServer pfs.API_FsckServer
 	return nil
 }
 
-// CreateFileSet implements the pfs.CreateFileSet RPC
+// CreateFileset implements the pfs.CreateFileset RPC
```
brycemcanally, line 679: CreateFileset -> CreateFileSet

src/server/pfs/server/api_server.go
brendoncarroll: It is also used in CreateFileset, which does not allow a commit to be set at any point.

brycemcanally (approved):
lgtm

PR #6333, opened 2021-06-14
Remove `CommitProvenance` protobuf

Tryneus:
This protobuf was pretty much unused and redundant with the `Commit` protobuf now that commits require branches.


PFedak (approved):
LGTM

PR #6334, opened 2021-06-14
Rename Block operations to Wait

Tryneus:
After some discussion with other team members, we came to the conclusion that this would be easier to understand.


brycemcanally (approved):
lgtm

PR #6341, opened 2021-06-14
Support postgres between namespaces or outside of cluster

chainlink:
In 1.x we rely on environment variables like `POSTGRES_SERVICE_HOST` which are injected by kubernetes based on services that exist in the namespace. This makes it difficult to use an external Postgres (ex. CloudSQL), or to use a postgres service in another namespace. 

This PR changes the names of postgres host and port env vars, and sets the default host to `postgres`,  which will resolve to the postgres service IP in the current namespace by default.


chainlink:
TODO: Wire up worker

src/server/pps/server/worker_rc.go
```diff
@@ -96,17 +96,36 @@ func (a *apiServer) workerPodSpec(options *workerOptions, pipelineInfo *pps.Pipe
 		Name:  "STORAGE_BACKEND",
 		Value: a.storageBackend,
 	}, {
-		Name:  "PORT",
-		Value: strconv.FormatUint(uint64(a.port), 10),
+		Name:  "POSTGRES_USER",
+		Value: a.env.Config().PostgresUser,
 	}, {
-		Name:  "PEER_PORT",
+		Name:  "POSTGRES_PASSWORD",
```
actgardner, line 102: NB: both the worker and the sidecar need postgres access, I guess? so all of these env vars have to be set in both.

chainlink: Ah! Nice catch. Common Env Vars is nice addition

actgardner: It came out to -1 LOC

seslattery (approved):

PR #6342, opened 2021-06-14
Fall back to no-browser case on error in auth login.

### LINKED ISSUE #6318: 'pachctl auth login' should print URL if browser can't be exec-ed
Currently when running `pachctl auth login` in e.g. WSL (or any VM), pachctl tries to open a login page in your browser, but the VM may not have a browser installed.

In this case, we should ideally make the `-b/--no-browser` flag discoverable, or simply default to the "no browser" behavior in this case

(We ran into this during the 2.0 gameday because @PFedak uses Windows)

PFedak:
Fixes #6318


acohen4 (approved):
LGTM

PR #6343, opened 2021-06-15
Remove enable stats and rename set types to follow *Set naming pattern (rather than *set)

brycemcanally:
This PR will finish removing enabling / disabling stats commits for pipelines in 2.0 (they are always enabled for transform pipelines and called meta commits now). This PR will also rename our set types (Fileset, Commitset, and Datumset) to follow the *Set naming pattern. We discussed the naming pattern we currently have (*set) internally and decided that it reads weird.


jdoliner (approved):
Lgtm

PR #6346, opened 2021-06-15
Pachd uses unprivileged ports

actgardner:
Update all the default ports in pachd to be > 1000, so the pod doesn't need to run as root. Originally this was also going to run pachd as a non-root user by default, but accessing the hostpath requires root access, so local deploys don't work :/

There's also a bigger project here to reduce the number of places where we define these ports.

### DIFF
--- a/src/server/cmd/pachctl/cmd/cmd.go
+++ b/src/server/cmd/pachctl/cmd/cmd.go
@@ -544,7 +544,9 @@ This resets the cluster to its initial state.`,
 	var uiPort uint16
 	var uiWebsocketPort uint16
 	var pfsPort uint16
+	var remotePfsPort uint16
 	var s3gatewayPort uint16
+	var remoteS3gatewayPort uint16
 	var dexPort uint16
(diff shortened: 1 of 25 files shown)


src/server/cmd/pachctl/cmd/cmd.go
```diff
@@ -693,15 +695,17 @@ This resets the cluster to its initial state.`,
 		}),
 	}
 	portForward.Flags().Uint16VarP(&port, "port", "p", 30650, "The local port to bind pachd to.")
-	portForward.Flags().Uint16Var(&remotePort, "remote-port", 650, "The remote port that pachd is bound to in the cluster.")
+	portForward.Flags().Uint16Var(&remotePort, "remote-port", 1650, "The remote port that pachd is bound to in the cluster.")
```
actgardner, line 698: Support configuring the S3 and PFS HTTP container ports in port-forwarding, these were hard-coded.

chainlink (approved):
LGTM

PR #6347, opened 2021-06-15
Rewrite stable hash computation again

brycemcanally:
This PR further simplifies the stable hash computation and fixes an issue with the previous implementation. The stable hash algorithm is now:
- If the file is fully contained in one file set, then we can just use the data ref hashes we store in the index for the file.
- If the file is spread across multiple file sets, then we need to run the content through a chunk writer to get the stable split points for the content (this will take advantage of the cheap copy in the common case where the file is spread across multiple file sets because it is large).


src/internal/storage/fileset/util.go
```diff
@@ -144,3 +145,14 @@ func (i *Iterator) Next() (File, error) {
 		return nil, err
 	}
 }
+
+func hashDataRefs(dataRefs []*chunk.DataRef) ([]byte, error) {
+	h := pachhash.New()
+	for _, dataRef := range dataRefs {
+		_, err := h.Write(dataRef.Hash)
```
brendoncarroll, line 152: This is 32 bytes raw, not hex right?

src/internal/storage/fileset/util.go
```diff
@@ -144,3 +145,14 @@ func (i *Iterator) Next() (File, error) {
 		return nil, err
 	}
 }
+
+func hashDataRefs(dataRefs []*chunk.DataRef) ([]byte, error) {
```
brendoncarroll, line 149: I did like having the public function `StableHash` is that gone now?

src/internal/storage/fileset/util.go
brycemcanally: Yeah.

src/internal/storage/fileset/util.go
brycemcanally: Yeah, it is replaced by the `Hash` function on the fileset.File interface which provides a stable hash.

Issue #6344, opened 2021-06-15
2.0 Put file/get file issues


nitinjainsj:
```
$ pc version
COMPONENT           VERSION             
pachctl             2.0.0-alpha.21      
pachd               2.0.0-alpha.21      
$
```

A few put file issues...

- [x] Error message shows "test1.user@master" instead of "test1@master"
```
$ pc create repo test1
$ pc list file test1@master:
branches test1.user@master not found
```

- [x] Can not delete dir and all the files under the dir. Delete dir does not have equivalent "-r"
```
$ pc put file -r test1@master -f emptydir/
$ pc list file test1@master:/emptydir
NAME                TYPE SIZE 
/emptydir/emptyfile file 0B   
$ pc delete file test1@master:/emptydir
$ pc list file test1@master:/emptydir
NAME                TYPE SIZE 
/emptydir/emptyfile file 0B   
$ pc delete file test1@master:/emptydir/emptyfile
$ pc list file test1@master:/emptydir
NAME TYPE SIZE 
$
```

- [x] spurious dir. Seems like we are creating some sort of dir entry when the first command failed due to lack of "-f"
```
$ pc put file -r test@master: ./emptydir/
expected 1 arguments, got 2

Usage:
  pachctl put file <repo>@<branch-or-commit>[:<path/to/file>] [flags]

Examples:
....
$ pc put file -r test@master: -f ./emptydir/
emptydir/emptyfile 0.00 b / 0.00 b [---------------------------------------------] 0s 0.00 b/s
$ pc list file test@master:/emptydir
NAME                TYPE SIZE 
/emptydir/emptydir/ dir  0B   
$ pc list file test@master:/emptydir/emptydir
NAME                         TYPE SIZE 
/emptydir/emptydir/emptyfile file 0B   
$
$ ls -R emptydir/
emptyfile
$
```

- [x] Incorrect files created. After deleting the empty file, creating it again somehow remembered the last path which created /emptydir/emptydir
```
$ pc list file test@master:/emptydir
NAME                TYPE SIZE 
/emptydir/emptydir/ dir  0B   
$ pc delete file test@master:/emptydir/emptydir/emptyfile
$ pc list file test@master
NAME TYPE SIZE 
$ pc put file -r test@master: -f ./emptydir/
emptydir/emptyfile 0.00 b / 0.00 b [----------------------------------] 0s 0.00 b/s
$ pc list file test@master
NAME       TYPE SIZE 
[...]

PR #6353, opened 2021-06-15
Add custom String() implementations to protos.

PFedak:
Give pfs protos commonly mentioned in error strings and other
user-facing messages use new custom String implementations which match
our preferred representations so that they are handled consistently.


PFedak:
This should make working with our protos a little easier, and address the first part of #6344, independent of whether we eliminate the user suffix entirely

src/pfs/pfs.proto
```diff
@@ -14,11 +14,13 @@ import "auth/auth.proto";
 ////  PFS Data structures (stored in etcd)
 
 message Repo {
+  option (gogoproto.goproto_stringer) = false;
```
Tryneus, line 17: the other options we use put these in `[]` - do we know why that is?  maybe just do the same thing here for consistency?

Tryneus (approved):
LGTM

src/pfs/pfs.proto
PFedak: I think that might be special syntax for the ones that apply to specific fields, but I can try to figure that out

PFedak: Yeah, seems to just be for tags

Issue #6322, opened 2021-06-11
`pachctl list datum -f` doesn't work in pachyderm 2.0


msteffen:
From the [Joins example](https://github.com/pachyderm/pachyderm/tree/master/examples/joins):
![image](https://user-images.githubusercontent.com/235674/121741253-1f6b1400-cab3-11eb-9cd5-cc39742be772.png)

However, when I run that command locally, I get this error:
```
$ pc list datum -f inner_join.json
Error: unknown shorthand flag: 'f' in -f
Usage:
  pachctl list datum <pipeline>@<job> [flags]

Flags:
  -h, --help            help for datum
  -o, --output string   Output format when --raw is set: "json" or "yaml" (default "json")
      --raw             Disable pretty printing; serialize data structures to an encoding such as json or yaml

Global Flags:
      --no-color   Turn off colors.
  -v, --verbose    Output verbose logs

unknown shorthand flag: 'f' in -f
```

jdoliner:
@acohen4 have you got bandwidth to take a look at this. I think the behavior was never hooked up in 2.0. Feel free to ping me if you want a run down of how this should work and how to approach this.

acohen4:
@jdoliner happy to work on this :)

acohen4:
Submitted PR to hook this back up in 2.0 : https://github.com/pachyderm/pachyderm/pull/6355
Closing this issue.

cc @jdoliner @msteffen

PR #6355, opened 2021-06-15
Restoring CLI option for Listing Datums based on uninitialized Pipeline (pachctl list datum -f <pipeline>.json)

acohen4:
Enabling `pachctl list datum -f <pipeline>.json`
Addressing https://github.com/pachyderm/pachyderm/issues/6322


src/server/pps/server/api_server.go
```diff
@@ -999,7 +999,10 @@ func (a *apiServer) ListDatum(request *pps.ListDatumRequest, server pps.API_List
 	// TODO: Auth?
 	if request.Input != nil {
 		return a.listDatumInput(server.Context(), request.Input, func(meta *datum.Meta) error {
-			return server.Send(convertDatumMetaToInfo(meta))
+			meta.State = datum.State_UNPROCESSED
+			di := convertDatumMetaToInfo(meta)
+			di.Datum.ID = ""
```
acohen4, line 1004: leaving the Datum ID blank restores the behavior in 1.x. This seems like reasonable behavior to me. I could see value exposing it but in fact this ID is different from that which will get created when live datums are instantiated. I'm still figuring out why that is

Tryneus (approved):
LGTM

PR #6360, opened 2021-06-16
Adding an Auth Perm for getting Pipeline Logs

acohen4:


src/internal/auth/interceptor.go
```diff
@@ -176,7 +176,7 @@ var authHandlers = map[string]authHandler{
 	"/pps_v2.API/StopPipeline":    authDisabledOr(authenticated),
 	"/pps_v2.API/RunPipeline":     authDisabledOr(authenticated),
 	"/pps_v2.API/RunCron":         authDisabledOr(authenticated),
-	"/pps_v2.API/GetLogs":         authDisabledOr(authenticated),
+	"/pps_v2.API/GetLogs":         authDisabledOr(clusterPermissions(auth.Permission_PIPELINE_GET_LOGS)),
```
actgardner, line 179: `GetLogs` is actually a weird RPC because depending on the fields that are set it can be "get the logs for pachd" or "get the logs for the pipelines", which have different access controls. I think it would be reasonable to split those into two different RPCs, and use the interceptor for the cluster logs one, or you could add the auth check in the PPS api server itself.

acohen4: @actgardner to clarify, for now we just want to guard the "get logs for pachd" with perm. is that right?

actgardner: Correct

src/server/auth/server/roles.go
```diff
@@ -183,6 +183,7 @@ func init() {
 				auth.Permission_CLUSTER_ENTERPRISE_GET_CODE,
 				auth.Permission_CLUSTER_ENTERPRISE_DEACTIVATE,
 				auth.Permission_CLUSTER_DELETE_ALL,
+				auth.Permission_PIPELINE_GET_LOGS,
```
actgardner, line 186: Also can we make a role to go with this permission so users don't have to be all-or-nothing admins?

acohen4:
@actgardner I decided to go with the less invasive approach of checking auth within the GetLogs RPC instead of splitting it up into two RPCs. I don't love it but I think the less invasive option makes sense in this case. Let me know if you think differently

src/server/pps/server/api_server.go
```diff
@@ -3111,3 +3116,21 @@ func labels(app string) map[string]string {
 		"component": "worker",
 	}
 }
+
+func (a *apiServer) authCheckPachdLogs(ctx context.Context) error {
```
actgardner, line 3120: This would probably make sense as a non-transaction verson of `CheckClusterIsAuthorizedInTransaction`, similar to `CheckRepoIsAuthorized` (https://github.com/pachyderm/pachyderm/blob/master/src/server/auth/server/util.go#L12). But we could do that in a follow-up, up to you.

src/server/auth/server/testing/auth_test.go
```diff
@@ -2780,3 +2780,61 @@ func TestExpiredClusterLocksOutUsers(t *testing.T) {
...
+	// wait for the pipeline pod to come up
+	time.Sleep(time.Second * 10)
+
+	// must be authorized to access pachd logs
+	pachdLogsIter := aliceClient.GetLogs("", "", nil, "", false, false, 0)
+	pachdLogsIter.Next()
+	require.YesError(t, pachdLogsIter.Err())
+	require.True(t, strings.Contains(pachdLogsIter.Err().Error(), "is not authorized to perform this operation"))
+
+	// only need to be authenticated to get pipeline logs
```
actgardner, line 2824: nit: you need some permissions on the repo to get the logs, I think? in this instance alice owns the pipeline so they can get the logs, but the logs shouldn't be acessible to a random authenticated user.

src/server/pps/server/api_server.go
acohen4: good point. I'll update it here

actgardner (approved):

PR #6361, opened 2021-06-17
Remove `Stored` variations of protobufs, now use a `Details` nested message

Tryneus:
`PipelineInfo` and `JobInfo` previously had a `StoredPipelineInfo` and `StoredJobInfo` protobuf representing what was stored in etcd.  Now, we use `PipelineInfo` and `JobInfo` everywhere, but they have an extra `Details` nested message which represents the expensive-to-load information which is not present in the database.  For these two structures, we get most of it by reading out the relevant pipeline spec commit from PFS.

Applied the same changes to `CommitInfo` and `FileInfo`, which have similar expensive-to-load fields.

Any RPC which reads this can pass in `Details: true` to request the additional data.  Renamed any such existing fields from `Full` to `Details` where applicable for consistency.


src/pps/pps.proto
```diff
@@ -339,95 +323,88 @@ enum PipelineState {
...
-  bool s3_out = 36;
-  Metadata metadata = 37;
-  string reprocess_spec = 38;
-  // Set at call time, not stored in etcd
-  int64 unclaimed_tasks = 39;
-  string worker_rc = 40;
-  bool autoscaling = 41;
+  PipelineType type = 10;
+
+  string auth_token = 11;
```
Tryneus, line 362: `AuthToken` is a little special in that it's stored in the database, but we _don't_ want to return it to users.  I added a test that `ListPipeline` and `InspectPipeline` don't pass it through, but it would be nice if we could have something more bulletproof.

src/server/pfs/server/api_server.go
```diff
@@ -609,7 +608,11 @@ func getFileURL(ctx context.Context, URL string, src Source) (int64, error) {
 		}); err != nil {
 			return err
 		}
-		bytesWritten += int64(fi.SizeBytes)
+		// TODO(2.0 required) - SizeBytes requires constructing the src with
+		// `WithDetails` which we don't do here, so we can't get the size from here
+		// that easily.  One option is to always calculate the size of files in the
+		// source.
+		// bytesWritten += int64(fi.Details.SizeBytes)
```
Tryneus, line 615: Discussed this case with @brycemcanally, will be handled in a followup PR - the nice thing about this change is it made the places we were using these uninitialized zero-values more apparent, as they'll now segfault.

src/server/pfs/server/trigger.go
```diff
@@ -96,16 +96,18 @@ func (d *driver) isTriggered(txnCtx *txncontext.TransactionContext, t *pfs.Trigg
...
 		if err != nil {
 			// Shouldn't be possible to error here since we validate on ingress
 			return false, errors.EnsureStack(err)
 		}
-		var oldSize uint64
-		if oldHead != nil {
-			oldSize = oldHead.SizeBytes
-		}
-		merge(int64(newHead.SizeBytes-oldSize) >= size)
+		// TODO(2.0 required): the size of a commit isn't known when finishing the
```
Tryneus, line 104: This also turned into a segfault because we weren't getting the `CommitInfo.Details` here (the commit has just finished and we don't want to wait for compaction).

src/server/pps/server/pipeline_controller.go
```diff
@@ -234,19 +237,6 @@ func (op *pipelineOp) run() error {
 	return nil
 }
 
-// getPipelineInfo reads the pipelineInfo associated with 'op's pipeline. This
-// should be one of the first calls made on 'op', as most other methods (e.g.
-// getRC, though not failPipeline) assume that op.pipelineInfo is set.
-func (op *pipelineOp) getPipelineInfo() error {
```
Tryneus, line 240: This was only called from one place, so I just inlined it up above.

src/server/pps/server/poller.go
```diff
@@ -50,20 +50,19 @@ func (m *ppsMaster) cancelPipelinePodsPoller() {
 	}
 }
 
-// startPipelineEtcdPoller starts a new goroutine running
-// pollPipelinesEtcd
-func (m *ppsMaster) startPipelineEtcdPoller() {
+// startPipelineDBPoller starts a new goroutine running watchPipelines
+func (m *ppsMaster) startPipelineWatcher() {
```
Tryneus, line 54: Scrubbed mentions of etcd from this file (because it's postgres now), and did a slight renaming pass on stuff that got touched (this poller wasn't actually polling, it was watching).

src/server/pps/server/s3g_sidecar.go
```diff
@@ -63,27 +59,18 @@ func (a *apiServer) ServeSidecarS3G() {
...
+		s.pachClient.SetAuthToken(s.pipelineInfo.AuthToken)
 
-		buf := bytes.Buffer{}
-		commit := client.NewSystemRepo(pipelineName, pfs.SpecRepoType).NewCommit("", specCommit)
-		if err := s.pachClient.GetFile(commit, ppsconsts.SpecFile, &buf); err != nil {
-			return errors.Wrapf(err, "sidecar s3 gateway: could not read pipeline spec commit: could not read existing PipelineInfo from PFS")
-		}
-		if err := proto.Unmarshal(buf.Bytes(), s.pipelineInfo); err != nil {
-			return errors.Wrapf(err, "sidecar s3 gateway: could not read pipeline spec commit: could not unmarshal PipelineInfo bytes from PFS")
+		if err := ppsutil.GetPipelineDetails(s.pachClient, s.pipelineInfo); err != nil {
```
Tryneus, line 71: The old code was just a janky reimplementation of `GetPipelineDetails`.

src/server/pps/server/s3g_sidecar.go
```diff
@@ -63,27 +59,18 @@ func (a *apiServer) ServeSidecarS3G() {
...
+			return errors.Wrapf(err, "sidecar s3 gateway: could not get pipeline details")
 		}
-		if !ppsutil.ContainsS3Inputs(s.pipelineInfo.Input) && !s.pipelineInfo.S3Out {
```
Tryneus, line 83: why

src/server/pps/server/s3g_sidecar.go
```diff
@@ -376,14 +363,19 @@ func (h *handleJobsCtx) start() {
...
 			}
 
 			// create new ctx for this job
 			jobCtx, jobCancel := context.WithCancel(context.Background())
 			h.processJobEvent(jobCtx, e.Type, jobInfo.Job)
+			// TODO(2.0 required): this is not true - this will continue to get PUT
+			// notifications on every job update - this needs to be refactored
+			// (preferrably with just one watcher, will likely require a way to know
+			// when initial PUTs are complete so we can remove dead jobs after
+			// recovering from a failed watcher):
```
Tryneus, line 378: Unrelated, but came across this in debugging - it looks like we're spawning a _lot_ of watches here (one every time the job is updated).  In addition, it doesn't look like the `.end` function ever returns?  Will file an issue tomorrow after talking to someone.

src/internal/ppsutil/util.go
```diff
@@ -114,43 +112,27 @@ func GetLimitsResourceList(limits *pps.ResourceSpec) (*v1.ResourceList, error) {
 	return getResourceListFromSpec(limits)
 }
 
-// GetPipelineInfoAllowIncomplete retrieves and returns a PipelineInfo from PFS,
-// or a sparsely-populated PipelineInfo if the spec data cannot be found in PPS
-// (e.g. due to corruption or a missing block). It does the PFS
-// read/unmarshalling of bytes as well as filling in missing fields
-func GetPipelineInfoAllowIncomplete(pachClient *client.APIClient, ptr *pps.StoredPipelineInfo) (*pps.PipelineInfo, error) {
-	result := &pps.PipelineInfo{}
+// GetPipelineDetails etrieves and returns a PipelineInfo from PFS,
```
brycemcanally, line 115: etrieves

brycemcanally: I think this comment needs to be updated if it now either returns the details or errors.

src/internal/ppsutil/util.go
```diff
@@ -61,7 +59,7 @@ func PipelineRcName(name string, version uint64) string {
 // GetRequestsResourceListFromPipeline returns a list of resources that the pipeline,
 // minimally requires.
 func GetRequestsResourceListFromPipeline(pipelineInfo *pps.PipelineInfo) (*v1.ResourceList, error) {
-	return getResourceListFromSpec(pipelineInfo.ResourceRequests)
+	return getResourceListFromSpec(pipelineInfo.Details.ResourceRequests)
```
brycemcanally, line 62: It might be a good idea to check for nil details here.

src/internal/ppsutil/util.go
```diff
@@ -114,43 +112,27 @@ func GetLimitsResourceList(limits *pps.ResourceSpec) (*v1.ResourceList, error) {
...
+func GetPipelineDetails(pachClient *client.APIClient, pipelineInfo *pps.PipelineInfo) error {
+	loadedPipelineInfo := &pps.PipelineInfo{}
 	buf := bytes.Buffer{}
 	// ensure we are authorized to read the pipeline's spec commit, but don't propagate that back out
 	pachClient = pachClient.WithCtx(pachClient.Ctx())
-	pachClient.SetAuthToken(ptr.AuthToken)
-	if err := pachClient.GetFile(ptr.SpecCommit, ppsconsts.SpecFile, &buf); err != nil {
-		log.Error(errors.Wrapf(err, "could not read existing PipelineInfo from PFS"))
+	pachClient.SetAuthToken(pipelineInfo.AuthToken)
+	if err := pachClient.GetFile(pipelineInfo.SpecCommit, ppsconsts.SpecFile, &buf); err != nil {
```
brycemcanally, line 125: We don't need an else here.

src/internal/ppsutil/util.go
```diff
@@ -246,37 +228,37 @@ func SetPipelineState(ctx context.Context, db *sqlx.DB, pipelinesCollection col.
...
+		return pipelines.Put(pipeline, pipelineInfo)
 	})
 	return err
 }
 
 // JobInput fills in the commits for an Input
 func JobInput(pipelineInfo *pps.PipelineInfo, outputCommit *pfs.Commit) *pps.Input {
 	commitsetID := outputCommit.ID
-	jobInput := proto.Clone(pipelineInfo.Input).(*pps.Input)
+	jobInput := proto.Clone(pipelineInfo.Details.Input).(*pps.Input)
```
brycemcanally, line 261: Nil check?

src/pps/pps.proto
```diff
@@ -233,10 +233,11 @@ message GPUSpec {
   int64 number = 2;
 }
 
-// StoredJobInfo is the portion of the JobInfo that gets stored
-// in the database during job execution. It contains fields which change over
-// the lifetime of the job but aren't used in the execution of the job.
-message StoredJobInfo {
+// JobInfo is the data stored in the database regarding a given job.  The
```
brycemcanally, line 236: I would probably change "from object storage" -> "from PFS".

src/pps/pps.proto
```diff
@@ -254,48 +255,31 @@ message StoredJobInfo {
...
-  string salt = 30;                            // requires ListJobRequest.Full
-  DatumSetSpec datum_set_spec = 31;                   // requires ListJobRequest.Full
-  google.protobuf.Duration datum_timeout = 32; // requires ListJobRequest.Full
-  google.protobuf.Duration job_timeout = 33;   // requires ListJobRequest.Full
-  int64 datum_tries = 34;                      // requires ListJobRequest.Full
-  SchedulingSpec scheduling_spec = 35;         // requires ListJobRequest.Full
-  string pod_spec = 36;                        // requires ListJobRequest.Full
-  string pod_patch = 37;                       // requires ListJobRequest.Full
+  message Details {
+    Transform transform = 2;
```
brycemcanally, line 263: 1

src/pps/pps.proto
```diff
@@ -339,95 +323,88 @@ enum PipelineState {
...
 }
 
-// StoredPipelineInfo is proto for each pipeline that Pachd stores in the
+// PipelineInfo is proto for each pipeline that Pachd stores in the
 // database. It tracks the state of the pipeline, and points to its metadata in
 // PFS (and, by pointing to a PFS commit, de facto tracks the pipeline's
-// version)
-message StoredPipelineInfo {
+// version).  Any information about the pipeline _not_ stored in the database is
+// in the Details object, which requires fetching the spec from object storage
```
brycemcanally, line 330: "from object storage" -> "from PFS"

src/pps/pps.proto
```diff
@@ -651,6 +629,9 @@ message CreatePipelineRequest {
 
 message InspectPipelineRequest {
   Pipeline pipeline = 1;
+  // When true, return PipelineInfos with the details field, which requires
+  // loading the pipeline spec from object storage.
```
brycemcanally, line 633: "from object storage" -> "from PFS"

src/server/pfs/pretty/pretty.go
```diff
@@ -148,10 +148,10 @@ func PrintCommitInfo(w io.Writer, commitInfo *pfs.CommitInfo, fullTimestamps boo
 			fmt.Fprintf(w, "%s\t", pretty.Ago(commitInfo.Finished))
 		}
 	}
-	if commitInfo.Finished == nil {
+	if commitInfo.Finished == nil || commitInfo.Details == nil {
```
brycemcanally, line 151: This should probably just be based on details being nil.

src/server/pfs/server/driver.go
```diff
@@ -924,7 +923,7 @@ func (d *driver) inspectCommit(ctx context.Context, commit *pfs.Commit, wait pfs
 		if err != nil {
 			return nil, err
 		}
-		commitInfo.SizeBytes = uint64(size)
+		commitInfo.Details = &pfs.CommitInfo_Details{SizeBytes: uint64(size)}
```
brycemcanally, line 926: We should probably check if details is not nil here before setting it. If it is not nil, we should just set the size. I don't think this is a problem as is, but could be later if we fill out more details elsewhere.

src/server/worker/datum/testing.go
```diff
@@ -76,10 +76,12 @@ func (mi *MockIterator) DatumN(index int) []*common.Input {
 		// Warning: this might break some assumptions about the format of this data
 		result = append(result, &common.Input{
 			FileInfo: &pfs.FileInfo{
-				File:      client.NewFile("dummy-repo", "dummy-branch", "dummy-commit", "/path"),
-				FileType:  pfs.FileType_FILE,
-				SizeBytes: uint64(index),
-				Hash:      []byte(fmt.Sprintf("%d", index)),
+				File:     client.NewFile("dummy-repo", "dummy-branch", "dummy-commit", "/path"),
```
brycemcanally, line 79: This file can be deleted. The `MockIterator` isn't used anywhere and doesn't have the same interface as the 2.0 datum iterators.

src/server/pfs/server/trigger.go
brycemcanally: How would triggers work with deletions? It looks like it only triggers if the commit size gets larger (in the size case), rather than the size of new data since the last commit. It might make sense to have a notion of total referenced size in a commit chain, which might make more sense for trigger semantics and would give us a low cost way to provide users with feedback that their data is going into a commit without computing the commit size.

src/server/pps/server/s3g_sidecar.go
brycemcanally: @msteffen probably has the most context for this.

src/server/pps/server/api_server.go
```diff
@@ -2353,17 +2346,18 @@ func (a *apiServer) inspectPipelineInTransaction(txnCtx *txncontext.TransactionC
 				return nil, err
 			}
 		} else {
-			pipelineInfo.Service.IP = service.Spec.ClusterIP
+			pipelineInfo.Details.Service.IP = service.Spec.ClusterIP
 		}
 	}
 
+	// TODO: move this into
```
brycemcanally, line 2353: helper function?

src/server/pps/server/api_server.go
```diff
@@ -2479,21 +2475,15 @@ func (a *apiServer) listPipeline(ctx context.Context, request *pps.ListPipelineR
...
-// resolvePipelineInfo looks up additional pipeline info in PFS needed to turn a StoredPipelineInfo into a PipelineInfo
-func (a *apiServer) resolvePipelineInfo(ctx context.Context, allowIncomplete bool, ptr *pps.StoredPipelineInfo) (*pps.PipelineInfo, error) {
-	pachClient := a.env.GetPachClient(ctx)
-	if allowIncomplete {
-		return ppsutil.GetPipelineInfoAllowIncomplete(pachClient, ptr)
-	}
-	return ppsutil.GetPipelineInfo(pachClient, ptr)
-}
-
 // listPipelinePtr enumerates all PPS pipelines in etcd, filters them based on
```
brycemcanally, line 2478: Might as well update this comment and change Ptr -> Info.

src/server/pps/server/api_server.go
```diff
@@ -2550,28 +2540,22 @@ func (a *apiServer) DeletePipeline(ctx context.Context, request *pps.DeletePipel
...
-	// - kubernetes services (for service pipelines, etc)
-	pipelineInfo, err := a.inspectPipeline(ctx, request.Pipeline.Name)
-	if err != nil {
-		logrus.Errorf("error inspecting pipeline: %v", err)
-		pipelineInfo = &pps.PipelineInfo{Pipeline: request.Pipeline, OutputBranch: "master"}
-	}
-
+	// Attempt to load pipeline details
 	pachClient := a.env.GetPachClient(ctx)
+	if err := ppsutil.GetPipelineDetails(pachClient, pipelineInfo); err != nil {
```
brycemcanally, line 2555: This seems sketchy. If we need the pipeline details to delete, then we should just error. If we don't and just default to master, then we would have an incorrect pipeline info when the output branch isn't master.

src/server/pps/server/api_server.go
Tryneus: The location that this code should be moved to is left as an exercise for the reader.

src/server/pfs/server/trigger.go
Tryneus: I think we should look into storing an accumulator on the branch trigger that we update after each change so we don't have to guess at the differences at this point in the code.  For example, it walks backwards from the head of the source branch to find a common ancestor, but that stops working if you don't make linear commits in the source branch.  Maybe we could do a similar thing with the diff size?

brycemcanally (approved):
lgtm

jdoliner (approved):
This all looks reasonable to me. Do we want to add a way to pachctl to request that details be included?

Tryneus:
@jdoliner - I think we can add that in the future if we have a solid use case, but for now it seems fine to do the expensive thing for CLI access

jdoliner:
@Tryneus ahh, so you'll always see the details field for commands like `list commit`?

Tryneus:
I think `list commit` might be a bit of an outlier because compaction is asynchronous at the moment, but it depends on how @brycemcanally wants to take care of it in the follow-up PR.

PR #6363, opened 2021-06-17
Run pachd as non-root user by default and make local deploy use PSP

actgardner:
Under normal circumstances pachd doesn't need to run as root. This PR makes the pachd docker image rootless by setting the UID to 1000. 

`pachctl deploy local` is a special case, because we use a host path mounted at `/pach` to store the data. We can't trust that the path will exist (with the right permissions) on the host machine, so we need to run as root in that specific case. However, all our testing is done with local deploys, and we want to run most of our tests as a non-root user.

The solution I'm proposing here is:
- the dockerfile runs as UID 1000 by default
- `pachctl deploy local` uses a pod security context to run as root by default, with a `--rootless` option to not use the PSC.
- in local development we'll use the PSC
- in CI we'll use the `--rootless` flag to make sure the entire test suite works when we don't have root (which should be the typical case in 99% of deployments - nobody should be using a hostpath in production!). This requires a little extra work to set up the directory on the host with the right permissions.

We'll need to add the PSC to the helm chart to maintain parity for local deploys as well.

### DIFF
--- a/Dockerfile.pachd
+++ b/Dockerfile.pachd
@@ -16,4 +16,6 @@ COPY dex-assets /dex-assets
 COPY --from=pachyderm_build /tmp/to-copy /
 COPY --from=pachyderm_build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 
+USER 1000
+
 ENTRYPOINT ["/pachd"]
--- a/etc/testing/circle/launch.sh
+++ b/etc/testing/circle/launch.sh
@@ -6,11 +6,14 @@ source "$(dirname "$0")/env.sh"
 
 make launch-loki
 
-for i in $(seq 3); do
-    make clean-launch-dev || true # may be nothing to delete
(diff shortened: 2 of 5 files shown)


chainlink:
Looks good! We could also maybe build in setting up that directory to `reset.py` 

Interesting failure in the object bucket it's trying to write to`/pach/chunk` We might want to check in with cust team to find out how they changed the various directories pach writes to for rootless envs

actgardner:
> Interesting failure in the object bucket it's trying to write to/pach/chunk We might want to check in with cust team to find out how they changed the various directories pach writes to for rootless envs

That was my mistake, that's just test that does a `pachctl deploy local` into a separate namespace and tries to make RPCs to it. I used UID 1 instead of 0 which works on my machine but isn't actually root :/ Should be fixed now.

chainlink (approved):

PR #6368, opened 2021-06-17
Prom port fix

chainlink:
Update prom port in one more spot 

(Really looking forward to refactoring/deprecating assets :D )


actgardner (approved):
Ahhhh, I can't believe I missed it!

PR #6373, opened 2021-06-17
Updating rabbit mq spout example for v2

avigil:


PFedak (approved):
If you're feeling up to it, it would great to add an integration test that just runs the writeFile function with as little setup as possible and make it run in the examples test bucket. 
Otherwise, LGTM

PR #6378, opened 2021-06-17
Add the ability to send profiles to Google Cloud Profiler

jrockway:
This would allow GCP customers (like Hub) to continuously monitor the performance of pachyderm and identify hotspots in actual production loads.  We use it for Hub itself and it's found a lot of issues (like a goroutine leak in the sentry client library).

Capture

Doc note: if a customer were to enable this, they'd need to also enable "cloud profiling" through GCP at: https://console.cloud.google.com/apis/library/cloudprofiler.googleapis.com?project=<their gcp project name>

### DIFF
--- a/src/internal/profileutil/profileutil.go
+++ b/src/internal/profileutil/profileutil.go
@@ -14,21 +14,26 @@ import (
 // https://cloud.google.com/profiler/docs
 //
 // Service is the name of this binary (pachd, worker, etc.).
-func StartCloudProfiler(service string, config *serviceenv.Configuration) error {
+//
+// If there is a problem starting the cloud profiler, it logs a message but we continue.
+func StartCloudProfiler(service string, config *serviceenv.Configuration) {
 	if config == nil {
 		log.Warn("nil configuration passed to StartCloudProfiler; profiling not enabled")
-		return nil
+		return
 	}
 	p := config.GoogleCloudProfilerProject
 	if p == "" {
-		return nil
+		return
 	}
-	log.Debugf("enabling google cloud profiler; sending profiles to project %q", p)
-	return profiler.Start(profiler.Config{
+	log.Infof("enabling google cloud profiler; sending profiles to project %q", p)
+	err := profiler.Start(profiler.Config{
 		Service:        service,
 		ServiceVersion: version.PrettyPrintVersion(version.Version),
 		MutexProfiling: true,
 		DebugLogging:   false,
 		ProjectID:      p,
 	})
+	if err != nil {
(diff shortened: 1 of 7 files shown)


src/server/cmd/pachd/main.go
```diff
@@ -118,6 +119,9 @@ func doEnterpriseMode(config interface{}) (retErr error) {
 		log.Printf("no Jaeger collector found (JAEGER_COLLECTOR_SERVICE_HOST not set)")
 	}
 	env := serviceenv.InitWithKube(serviceenv.NewConfiguration(config))
+	if err := profileutil.StartCloudProfiler("pachyderm-pachd-enterprise", env.Config()); err != nil {
+		return errors.Wrapf(err, "error starting cloud profiler")
```
brycemcanally, line 123: Should startup fail if we cannot setup the profiler?

src/internal/profileutil/profileutil.go
```diff
@@ -0,0 +1,34 @@
...
+func StartCloudProfiler(service string, config *serviceenv.Configuration) error {
+	if config == nil {
+		log.Warn("nil configuration passed to StartCloudProfiler; profiling not enabled")
+		return nil
+	}
+	p := config.GoogleCloudProfilerProject
+	if p == "" {
+		return nil
+	}
+	log.Debugf("enabling google cloud profiler; sending profiles to project %q", p)
```
brycemcanally, line 26: This seems like something that should be at the info level.

src/server/cmd/pachd/main.go
jrockway: My thought is if you specify the wrong project and it fails, you can just edit the environment variable and restart, or empty it if you can't figure it out and just want to get started without it.  The alternative is to wait a while and then never see the profiles, and wonder why.

But I'm very OK with just a log.Error here if you prefer that.

brycemcanally: I think I'm fine with the just set it to empty if it doesn't come up.

brycemcanally (approved):
lgtm, just one minor change in regards to the log level when we setup profiling. I definitely will take advantage of this with the load testing.

src/server/cmd/pachd/main.go
jrockway: I made it log an error and continue on.

src/internal/profileutil/profileutil.go
jrockway: Done.

jrockway:
I fixed some documentation typos and made the environment variable propagate to the worker and sidecars (which can also submit profiles).  If CI passes, I'm going to merge this so that it can be in the next beta.

PR #6380, opened 2021-06-17
Deprecate old dash from core pach

chainlink:


src/server/cmd/pachctl/cmd/cmd.go
```diff
@@ -601,26 +599,6 @@ This resets the cluster to its initial state.`,
 				successCount++
 			}
 
-			fmt.Printf("Forwarding the dash (Pachyderm dashboard) UI port to http://localhost:%v...\n", uiPort)
```
nitinjainsj, line 604: Make sure to address doc changes since this will no longer be visible to users.

nitinjainsj (approved):

PR #6386, opened 2021-06-17
Change some inspect pipelines to collection checks.

PFedak:
Both of these seem like cases where we don't really care about the extra
information and work inspect pipeline does, and which might give
confusing errors with broken cluster states.

One is also a gRPC unnecessarily.


jdoliner (approved):
Lgtm

PR #6394, opened 2021-06-18
Add back GetFile

brendoncarroll:
- GetFile returns the file matched by the path.  If more than one file mathces, it errors.  If it matches a directory it errors.
- GetFileTAR can be used to get the contents of multiple files in one call.


src/client/pfs_file.go
```diff
@@ -416,34 +415,48 @@ func (c APIClient) RenewFileSet(ID string, ttl time.Duration) (retErr error) {
 // size limits the total amount of data returned, note you will get fewer bytes
 // than size if you pass a value larger than the size of the file.
 // If size is set to 0 then all of the data will be returned.
-// TODO: Should we error if multiple files are matched?
-func (c APIClient) GetFile(commit *pfs.Commit, path string, w io.Writer) error {
-	r, err := c.getFileTar(commit, path)
+func (c APIClient) GetFile(commit *pfs.Commit, path string, w io.Writer) (retErr error) {
+	ctx, cf := context.WithCancel(c.Ctx())
```
Tryneus, line 419: I think this should defer a `grpcutil.ScrupGRPC`

src/server/pfs/server/val_server.go
```diff
@@ -147,6 +147,13 @@ func (a *validatedAPIServer) SquashCommitSet(ctx context.Context, request *pfs.S
 	return a.apiServer.SquashCommitSet(ctx, request)
 }
 
+func (a *validatedAPIServer) GetFile(request *pfs.GetFileRequest, server pfs.API_GetFileServer) error {
```
Tryneus, line 150: This should probably also check that `File.Commit`, `File.Commit.Branch`, and `File.Commit.Branch.Repo` are not `nil`.

src/server/pfs/server/testing/server_test.go
```diff
@@ -3135,41 +3135,56 @@ func TestPFS(suite *testing.T) {
 			require.Equal(t, numFiles+1, len(fileInfos))
 
 			var output strings.Builder
-			err = env.PachClient.GetFile(commit, "*", &output)
+			rc, err := env.PachClient.GetFileTAR(commit, "*")
```
Tryneus, line 3138: seems like this would be worth putting in a helper function

src/server/pfs/server/api_server.go
```diff
@@ -530,6 +530,32 @@ func (a *apiServer) GetFileTAR(request *pfs.GetFileRequest, server pfs.API_GetFi
 	})
 }
 
+// GetFile implements the protobuf pfs.GetFile RPC
+func (a *apiServer) GetFile(request *pfs.GetFileRequest, server pfs.API_GetFileServer) (retErr error) {
+	func() { a.Log(request, nil, nil, 0) }()
+	defer func(start time.Time) { a.Log(request, nil, retErr, time.Since(start)) }(time.Now())
+	return metrics.ReportRequestWithThroughput(func() (int64, error) {
+		ctx := server.Context()
+		src, err := a.driver.getFile(ctx, request.File)
```
Tryneus, line 539: This seems kind of heavyweight for `api_server.go` - maybe it's fine because `GetFile` can't be done in an existing transaction anyway so an RPC boundary doesn't hurt.

src/server/pfs/server/source.go
```diff
@@ -187,3 +187,26 @@ type emptySource struct{}
...
+
+// singleFile iterates through the source and returns exactly one file or an error.
+// If the source contains a directory, then singleFile errors.
+// If the source contains more than one file of any type, then singleFile errors
+func singleFile(ctx context.Context, src Source) (*pfs.FileInfo, fileset.File, error) {
+	var retFinfo *pfs.FileInfo
+	var retFile fileset.File
+	err := src.Iterate(ctx, func(finfo *pfs.FileInfo, fsFile fileset.File) error {
+		if retFinfo != nil {
+			return errors.Errorf("matched multiple files")
```
Tryneus, line 199: should this also recommend using `GetFileTAR`?

Tryneus (approved):
A few nits but overall looks good

PR #6396, opened 2021-06-18
License service stores enterprise license in postgres collection

actgardner:
In the move from etcd to postgres it looks like the license server's record of the user's current enterprise license didn't get moved to postgres.

### DIFF
--- a/src/internal/clusterstate/current.go
+++ b/src/internal/clusterstate/current.go
@@ -16,6 +16,7 @@ import (
 	authserver "github.com/pachyderm/pachyderm/v2/src/server/auth/server"
 	"github.com/pachyderm/pachyderm/v2/src/server/identity"
 	"github.com/pachyderm/pachyderm/v2/src/server/license"
+	licenseserver "github.com/pachyderm/pachyderm/v2/src/server/license/server"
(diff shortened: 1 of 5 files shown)


acohen4 (approved):
LGTM, although I thought we intentionally kept the license record in etcd to keep it easily cached/watched.

Although this makes sense since we'd like it to be backed up

actgardner:
There's two places the license is stored:
- In the license service, the user uploads a license and it's served to the enterprise services asynchronously in heartbeats. This needs to be durable.
- In the enterprise service, we get the most recent license from the license server in the heartbeat and cache it in etcd. This is stored in etcd and cached in memory because this is how we enforce whether enterprise features are active.

acohen4:
> There's two places the license is stored:
> 
> * In the license service, the user uploads a license and it's served to the enterprise services asynchronously in heartbeats. This needs to be durable.
> * In the enterprise service, we get the most recent license from the license server in the heartbeat and cache it in etcd. This is stored in etcd and cached in memory because this is how we enforce whether enterprise features are active.

Yep thanks for clarifying. It was just lost on me reviewing the code here. Good catch fixing this!

PR #6397, opened 2021-06-18
Remove docker socket from worker containers

actgardner:
Switch `NoExposeDockerSocket` to `ExposeDockerSocket` everywhere, and make it false by default.

This also adds an integration test that checks whether we can run an un-privileged pipeline. It only runs in CI, until we get rid of the hostpath for local deploys in favour of minio.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,9 +112,6 @@ docker-gpu: docker-build-gpu docker-push-gpu
 
 docker-gpu-dev: docker-build-gpu docker-push-gpu-dev
 
-docker-build-test-entrypoint:
-	docker build $(DOCKER_BUILD_FLAGS) -t pachyderm_entrypoint etc/testing/entrypoint
-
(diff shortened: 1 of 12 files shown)


chainlink:
Definitely approve of the logic switch! I am wondering though if we should just nuke docker entirely. It was used in a couple of deploy commands, but that's gone now. The only place it's really used is to do a couple of quality of life checks in the worker. Docker deps are a real mess, so it has the potential to nicely clean up the go modules file too.
 https://github.com/pachyderm/pachyderm/blob/320ed850f01b0e01c82c26af861547a599dbcb4d/src/server/worker/worker.go#L71-L94

chainlink (approved):

msteffen (approved):
Love the test!

actgardner:
Yeah, I agree we should just remove docker altogether, I'll update this ;)

PR #6398, opened 2021-06-18
Give meta repos the same provenance as output repos.

PFedak:
This is motivated at least in part by our use of meta repos as a place
to store information during a job.


Tryneus (approved):
I'm almost surprised it was this simple.

LGTM!

Issue #6365, opened 2021-06-17
Pachctl create branch related issues


brycemcanally:
For pachctl create branch, when using the head flag, you have to specify the head as `repo@branch` instead of just `branch`. We should allow just specifying the branch and filling in the repo based on the first parameter. Also, create branch currently allows setting the head to a branch in a different repo which we should disallow.

brendoncarroll:
Also ran into this in the market-sentiment example.

@JimmyWhitaker depending on the resolution here this may need to be updated
```pachctl create branch financial_phrase_bank@v1 --head master
```

acohen4:
the updated behavior in https://github.com/pachyderm/pachyderm/pull/6399 allows passing a branch / commit-id as the `head` value, or passing the fully qualified commit ID: <repo>@ <branch>=<commit-id>. 

The fully qualified commit format is only necessary in the edge case where there may be multiple commits in a repo for the commit-set-id (up to one per repo branch). For example, commit-set-id `id-2` may be in both data@master and data@test, in which case we need the user to provide that repo@branch info to determine which exact commit to branch off of. 

@nitinjainsj @Tryneus @brycemcanally @brendoncarroll

acohen4:
This issue is fixed as mentioned in the comment above

PR #6399, opened 2021-06-18
`pachctl create branch --head` takes a branch name & Do  backend validation

acohen4:
Addressing Issue: https://github.com/pachyderm/pachyderm/issues/6365

### DIFF
--- a/src/server/pfs/cmds/cmds.go
+++ b/src/server/pfs/cmds/cmds.go
@@ -687,13 +687,14 @@ Any pachctl command that can take a Commit ID, can take a branch name instead.`,
 			}
 			var headCommit *pfs.Commit
 			if head != "" {
-				if uuid.IsUUIDWithoutDashes(head) {
-					headCommit = branch.Repo.NewCommit("", head)
-				} else {
+				if strings.Contains(head, "@") {
 					headCommit, err = cmdutil.ParseCommit(head)
 					if err != nil {
 						return err
 					}
+				} else {
+					// treat head as the commitID or branch name
+					headCommit = branch.Repo.NewCommit("", head)
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


src/server/pfs/cmds/cmds.go
```diff
@@ -687,14 +687,7 @@ Any pachctl command that can take a Commit ID, can take a branch name instead.`,
 			}
 			var headCommit *pfs.Commit
 			if head != "" {
-				if uuid.IsUUIDWithoutDashes(head) {
-					headCommit = branch.Repo.NewCommit("", head)
-				} else {
-					headCommit, err = cmdutil.ParseCommit(head)
```
Tryneus, line 693: I think we still need something to handle this case, because you may have multiple commits matching the given commitset in a repo (one per branch).  In that case, the RPC would fail unless the user provides a fully-qualified commit (i.e. `repo@branch=id`).  I realize the repo is redundant in this case, but the branch is important and I don't think it's worth inventing a new syntax.

Tryneus (changes requested):
LGTM except for one corner case

src/server/pfs/cmds/cmds.go
```diff
@@ -711,7 +719,7 @@ Any pachctl command that can take a Commit ID, can take a branch name instead.`,
 	}
 	createBranch.Flags().VarP(&branchProvenance, "provenance", "p", "The provenance for the branch. format: <repo>@<branch-or-commit>")
 	createBranch.MarkFlagCustom("provenance", "__pachctl_get_repo_commit")
-	createBranch.Flags().StringVarP(&head, "head", "", "", "The head of the newly created branch.")
+	createBranch.Flags().StringVarP(&head, "head", "", "", "The head of the newly created branch. Either pass the commit with format: <branch-or-commit>, or fully-qualified as <repo>@<branch-or-commit>=<id>")
```
Tryneus, line 722: I think the fully-qualified example should be `<repo>@<branch>=<id>`, since `<repo>@<commit>=<id>` doesn't work.

Tryneus (approved):
LGTM, one nit

PR #6402, opened 2021-06-21
Renumber protobuf enums so that zero is no longer a valid value

Tryneus:
My main purpose for this is that when doing `pachctl --raw`, an enum field will be omitted if the underlying value is zero.  Several of our protobufs already had placeholder 'UNKNOWN' values, so I tried to standardize on that.  There were a few places where we depended on initializing a protobuf to a default state by not specifying the enum value, so I fixed those places as well.


brendoncarroll (approved):
We talked a little about this offline.

I don't really see these values ever being referenced.  On the sender, assigning the unknown value is the same as not setting the field.  So just don't set the field.
On the receiver, all the unknown values (not just 0, but any value not known to the application) are usually handled in a switch default clause.

The encryption enum's unknown value seems especially unnecessary since you won't be getting the data back if you don't know the algorithm.

The reasoning for this seems to be partly cosmetic for the CLI, and for uniformity with other parts of the codebase that already do this.

PR #6403, opened 2021-06-21
Proxy postgres listener

brycemcanally:
This PR implements and sets up a proxy postgres listener. The reason for implementing a proxy postgres listener is to limit the number of dedicated postgres connections needed in a cluster to the number of full pachd instances. Each worker pod runs a pachd sidecar instance which will go through the proxy postgres listener for listen operations (primarily watch operations in the collection package). The changes / additions that were required to implement the proxy postgres listener were:
- Refactoring the normal listener such that it and the proxy listener could be behind the same interface (`PostgresListener`).
- Adding a new `Notifier` interface to define the common abstraction to receive the listen notifications (for normal listens this is implemented by the postgres watcher and for proxy listens this is implemented by the proxy server notifier).
- Adding a new gRPC client / server for providing the listen operations over the network, which includes client side channel multiplexing (multiple listens on the same client and same channel will be multiplexed to the same listen gRPC stream).
- Refactoring the collection package tests such that the test suite could be run with the proxy postgres listener.
- Adding a service pach client to the service environment (proxy requests need to go through the pachd service to get to the full pachd instances).


src/internal/serviceenv/service_env.go
```diff
@@ -77,6 +78,14 @@ type NonblockingServiceEnv struct {
 	// returned, then pachd would be unable to start)
 	pachEg errgroup.Group
 
+	// servicePachClient is the "template" service pach client (client connected through the pachd service)
+	// other clients returned by this library are based on. It contains the original GRPC client
+	// connection and has no ctx and therefore no auth credentials or cancellation.
+	servicePachClient *client.APIClient
+	// servicePachEg coordinates the initialization of servicePachClient (see pachdEg)
+	servicePachEg   errgroup.Group
+	servicePachOnce sync.Once
```
Tryneus, line 87: Either move these into local variables when initializing the proxy listener or expose them with `ServiceEnv` getters (if you have other plans for these) - otherwise I don't see the point of having them here since they're only used in one place.

src/internal/serviceenv/service_env.go
```diff
@@ -281,6 +290,11 @@ func (env *NonblockingServiceEnv) initDBClient() error {
 }
 
 func (env *NonblockingServiceEnv) initListener() error {
+	// TODO: Change this to be based on whether a direct connection to postgres is available.
+	// A direct connection will not be available in the workers when the PG bouncer changes are in.
+	if env.Config().PPSPipelineName != "" {
+		return env.initProxyListener()
```
Tryneus, line 296: for cleanliness, I'd rather have a `initProxyListener` and `initDirectListener` that return a `(col.PostgresListener, error)` and do the setting of it in one place here.

src/internal/collection/postgres_listener.go
```diff
@@ -15,52 +15,80 @@ import (
...
+	closer   sync.Once
 
 	// Filtering variables:
 	opts  watch.WatchOptions // may filter by the operation type (put or delete)
 	index *string            // only set if the watch filters by an index (for dealing with hash collisions)
 	value *string            // only set if 'index' is set
 }
 
-func newPostgresWatch(
+func NewPostgresWatcher(
```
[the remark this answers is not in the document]
brycemcanally: No, we can make this unexported.

Tryneus (approved):
LGTM

PR #6404, opened 2021-06-21
Simplify Storage Secrets 

chainlink:
This simplifies our manifests by using an `envFrom` block to pull in secret env vars instead of needing to define them manually in the deployment / worker RC. Keys in the secret have been renamed to all caps for env var conformity


actgardner (approved):
Looks good, modulo the S3 test that's failing in MISC.

chainlink:
Docs: Create/Edit the `pachyderm-storage-secret`

Issue #6330, opened 2021-06-11
Edit pipeline errors in 2.0


JimmyWhitaker:
In the Iris example, when running `pachctl edit pipeline inference` the command errors with: 
`Subcommand required`

Tryneus:
It appears that `pachctl edit pipeline` was removed (but the top-level `edit` verb was left in) - do we need to add it back or should we just remove `pachctl edit`?

jdoliner:
I think `pachctl edit pipeline` is a nice to have and shouldn't be too hard to keep. I think `extract pipeline` still works (it kind of has to because `extract` still works in general) so this shouldn't be too hard to hook up.

PR #6406, opened 2021-06-21
Avigil/edit pipeline

avigil:
This fixes issue #6330


PFedak (approved):
LGTM, is this just identical to the old version? I didn't notice anything different

src/server/pps/cmds/cmds.go
```diff
@@ -674,6 +675,80 @@ All jobs created by a pipeline will create commits in the pipeline's output repo
...
+		Short: "Edit the manifest for a pipeline in your text editor.",
+		Long:  "Edit the manifest for a pipeline in your text editor.",
+		Run: cmdutil.RunFixedArgs(1, func(args []string) (retErr error) {
+			client, err := pachdclient.NewOnUserMachine("user")
+			if err != nil {
+				return err
+			}
+			defer client.Close()
+		
+			pipelineInfo, err := client.InspectPipeline(args[0])
```
Tryneus, line 691: Heads up that this will break when merged with `master` - `InspectPipeline` now takes a `details bool` arg (which this should set to true).

avigil:
> LGTM, is this just identical to the old version? I didn't notice anything different

@PFedak there are changes. The previous version used ExtractPipeline. So, I had to switch to using InspectPipeline and change some things related to that

src/server/pps/cmds/cmds.go
avigil: @Tryneus I added the second parameter

avigil:
I removed MaxQueueSize and CacheSize since they are not being used and were causing errors when editing and saving a pipeline. But I want to make sure that doesn't cause any issues that I don't know about

src/pps/pps.proto
```diff
@@ -612,11 +612,9 @@ message CreatePipelineRequest {
   ResourceSpec sidecar_resource_limits = 11;
   Input input = 12;
   string description = 13;
-  string cache_size = 14;
```
Tryneus, line 615: Also need to remove these earlier in the file (lines 396 and 619) and that will reveal some other places that are still using these fields.

Tryneus (changes requested):

src/pps/pps.proto
avigil: Thank you! Working on those now

src/server/pachyderm_test.go
```diff
@@ -6480,7 +6480,6 @@ func TestMaxQueueSize(t *testing.T) {
 			ParallelismSpec: &pps.ParallelismSpec{
 				Constant: 2,
 			},
-			MaxQueueSize: 1,
```
brycemcanally, line 6483: I would remove this whole test since it is just testing this configuration.

src/server/pps/server/api_server.go
```diff
@@ -1423,12 +1423,6 @@ func (a *apiServer) validatePipelineRequest(request *pps.CreatePipelineRequest)
 
 // TODO: Implement the appropriate features.
 func (a *apiServer) validateV2Features(request *pps.CreatePipelineRequest) (*pps.CreatePipelineRequest, error) {
```
brycemcanally, line 1425: You can just remove this function since we do not need to block any features once these two are removed.

Tryneus: @avigil - should probably still do this ^

Tryneus (approved):
LGTM after taking care of an earlier comment

src/server/pps/server/api_server.go
avigil: This method (and references to it are removed now)

PR #6407, opened 2021-06-21
add support for passing through user and password to connect to postgres

seslattery:


chainlink:
Nice! Can you add this to `assets.go` as well?

seslattery:
> Nice! Can you add this to `assets.go` as well?

Sure, but what should the default value for "POSTGRES_PASSWORD" be?

chainlink:
Oh, good point, I guess empty? (In keeping with `TRUST` for now)

chainlink (approved):

PR #6409, opened 2021-06-21
`pachctl put file` strips directory from target path

actgardner:
One common frustration with `pachctl put file` is that by default the target file name is PFS is the entire relative path.

For example, in 1.x:
```
> pachctl put file training@master -f examples/ml/iris/data/iris.csv
> pachctl list file training@master
NAME                            TAG     TYPE SIZE
/examples/ml/iris/data/iris.csv default file 4.444KiB
```

This makes our examples clumsy because users have to cd to the correct directory to run commands (or provide the target path, which our examples don't do).

This PR proposes to make `pachctl put file` consider only the base of the path provided by the user. Examples:

Putting a single or multiple files uses their names, but not the path:
```
> pachctl put file training@example1 -f examples/ml/iris/data/iris.csv
> pachctl list file training@example1
NAME      TAG     TYPE SIZE
/iris.csv default file 4.444KiB
```

Putting multiple files at a specified path puts them in a subdirectory:
```
> pachctl put file training@example2:/test -f examples/ml/iris/data/test/1.csv -f examples/ml/iris/data/test/2.csv
> pachctl list file training@example2:/test
NAME        TAG     TYPE SIZE
/test/1.csv default file 16B
/test/2.csv default file 96B
```

Recursive copies the containing directory:
```
> pachctl put file training@example3: -r -f examples/ml/iris/data
> pachctl list file training@example3:/data
NAME           TAG     TYPE SIZE
/data/iris.csv default file 4.444KiB
/data/test/            dir  112B
```

### DIFF
--- a/src/server/pfs/cmds/cmds.go
+++ b/src/server/pfs/cmds/cmds.go
@@ -952,7 +952,8 @@ $ {{alias}} repo@branch -i http://host/path`,
 						if source == "-" {
 							return errors.Errorf("must specify filename when reading data from stdin")
 						}
-						if err := putFileHelper(mf, joinPaths("", source), source, recursive, appendFile); err != nil {
+						target := filepath.Base(source)
+						if err := putFileHelper(mf, joinPaths("", target), source, recursive, appendFile); err != nil {
 							return err
 						}
 					} else if len(sources) == 1 {
@@ -964,7 +965,8 @@ $ {{alias}} repo@branch -i http://host/path`,
 					} else {
 						// We have multiple sources and the user has specified a path,
 						// we use that path as a prefix for the filepaths.
-						if err := putFileHelper(mf, joinPaths(file.Path, source), source, recursive, appendFile); err != nil {
+						target := filepath.Base(source)
+						if err := putFileHelper(mf, joinPaths(file.Path, target), source, recursive, appendFile); err != nil {
 							return err
 						}
 					}
--- a/src/server/pfs/cmds/cmds.go
+++ b/src/server/pfs/cmds/cmds.go
@@ -838,6 +838,7 @@ from commits with 'get file'.`,
 	var appendFile bool
 	var compress bool
 	var enableProgress bool
+	var fullPath bool
 	putFile := &cobra.Command{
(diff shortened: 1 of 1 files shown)


jdoliner:
I'm in favor of doing this, however I think we need to limit the cases in which we do it. I'm thinking that we should only do it for non-recursive cases. The reason for this is file trees that contain duplicate base filenames which I think is fairly common. For example if you have a directory per day and within in directory you have a database dump the database dumps might be called `01-01-2021/dump`, `01-02-2021/dump` etc. and those will all wind up going to a file called `dump` if you put them recursively.

Also likely we'll want a flag to disable this behavior and get the old behavior. But the new behavior seems like a better default to me for single file uploads.

actgardner:
I'm not sure I follow the example, if I have a structure like:

```
> find database-dumps
database-dumps
database-dumps/01-01-2021
database-dumps/01-01-2021/dump
database-dumps/01-02-2021
database-dumps/01-02-2021/dump
```

and I do a recursive put:
```
> pachctl put file training@recursive: -r -f database-dumps
> pachctl list file training@recursive:/database-dumps
NAME                        TAG TYPE SIZE
/database-dumps/01-01-2021/     dir  0B
/database-dumps/01-02-2021/     dir  0B
```

The stripping of the leading directories applies to the containing directory (so `some/long/path/database-dumps/` would just be truncated to `database-dumps/`), but it doesn't flatten the structure inside the directory.

jdoliner:
I see, I think that sounds good then. I thought it was taking the base of the full path in the recursive case, but it's only taking the base of the path you pass in. Another case to try it with is with `-i` but I think that should work fine as well.

brendoncarroll (approved):

smalyala:
@actgardner In the `database-dumps` example you gave, what would happen if we ran
`pachctl put file training@master -f database-dumps/01-01-2021/dump -f database-dumps/01-02-2021/dump`?

Wondering if specifying individual files for `put file` has a different behavior than using the recursive flag.

actgardner:
@smalyala Those file names would collide, for sure. That's actually the same behaviour as cp. We have the `--full-path` option which would preserve the old behaviour and put them at two different paths.

PR #6412, opened 2021-06-22
Deflake test for non-root pipelines

actgardner:
The non-root pipeline test is just a copy of the simple pipeline test, apply the same fix from #6410


PFedak (approved):
LGTM

PR #6413, opened 2021-06-22
Split `pachctl config set-context` into two commands

actgardner:
`pachctl config set-context` can accept a kube context name, but the usage is kind of obscure. When we eliminate `pachctl deploy local` we'll need a simple one-liner that creates a context based on the current kube context.

This PR introduces `pachctl config import-kube <pach context name>`, which creates a new Pach context from the current kube context. This also adds a `--namespace` flag, to support the case where we deploy into a context that isn't the default in the kube context, and `--enterprise` for deploying an enterprise server.

This should make the minikube helm deploy process as simple as `helm install && pachctl config import-kube`.

### DIFF
--- a/src/server/config/cmds.go
+++ b/src/server/config/cmds.go
@@ -209,11 +209,10 @@ func Cmds() []*cobra.Command {
 	commands = append(commands, cmdutil.CreateAlias(getContext, "config get context"))
 
 	var overwrite bool
-	var kubeContextName string
 	setContext := &cobra.Command{
 		Use:   "{{alias}} <context>",
 		Short: "Set a context.",
-		Long:  "Set a context config from a given name and either JSON stdin, or a given kubernetes context.",
+		Long:  "Set a context config from a given name and a JSON configuration file on stdin",
 		Run: cmdutil.RunFixedArgs(1, func(args []string) (retErr error) {
 			name := args[0]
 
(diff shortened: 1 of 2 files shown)


chainlink:
Relatedly, I noticed I can install the helm chart and immediately use pachctl` (it seems to bootstrap a pach config with local connection for you?)

actgardner:
I guess this is because if you have no config it makes a new empty context, and the default behaviour for an empty pach context is to use the current kube context?

chainlink:
Starting from no `~./pachyderm/config.json` here's what it looks like after running `pachctl list repo` directly after a helm install

```
{
  "user_id": "ecd09fe6-f1e5-4556-8e9a-db5fc1b38904",
  "v2": {
    "active_context": "default",
    "contexts": {
      "default": {}
    },
    "metrics": true
  }
}
```

actgardner:
Oh, it's definitely using port forwarding - the pachctl config supports either "implicitly" port-forwarding given a kube context, or a grpc address. I think what's surprising is that if there's no kube context specified in the pachctl config that it just uses the active context.

chainlink:
We chatted about this offline, but have an explicit option will be good

chainlink:
A test might be good here if it's not too much of a pain to add

chainlink (approved):

chainlink:
FYI Was just going through some docs stuff with @nadegepepin  and found this https://docs.pachyderm.com/latest/deploy-manage/deploy/import-kubernetes-context/

PR #6416, opened 2021-06-22
Clean up portforward methods

actgardner:


chainlink:
So this strips the ports down to the bare minimum? (ie, no s3g, etc)

actgardner:
Nope, we have the exact same behaviour, just without 4 copies of the same method with different names

chainlink (approved):

PR #6417, opened 2021-06-22
Use UID 1000 in the worker binary

msteffen:
I really have not done much investigation into what was preventing this from working before; I just tried adding the relevant line in the dockerfile.

In particular:
- Volume mounts? AFAICT the only way to change the owner/permissions associated with mounts in kubernetes is with security contexts, and I know the init container needs to mount at least `/pach-bin` to work (to copy the worker binary there, so the user container can run it). Doesn't this mean that user 1000 shouldn't be able to copy the worker binary into `/pach-bin` and therefore that nothing works?
- Worker service port? There were issues setting this previously, and I didn't change anything in this PR
- This just changes the UID in the init container, but wouldn't it be necessary to change the default UID of user pipelines? That sounds completely insane though, and I'd expect this to break every pipeline.

I guess CI runs though?

### DIFF
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -9,3 +9,5 @@ WORKDIR /app
 COPY worker . 
 COPY worker_init init
 COPY pachctl .
+
+USER 1000


chainlink:
Nice!

chainlink:
cc @actgardner

actgardner:
> I really have not done much investigation into what was preventing this from working before; I just tried adding the relevant line in the dockerfile.

My hesitation to do this was basically because the UID of the init container and the user container may not match, and I wasn't 100% sure what the behaviour would be if they were both unprivileged. This prompted me to do some experimentation:

> Doesn't this mean that user 1000 shouldn't be able to copy the worker binary into /pach-bin and therefore that nothing works?

AFAICT k8s changes the volume permissions such that the UID of the container has appropriate access. Otherwise it would be complicated to have volumes mounted in non-root containers.

> Worker service port? There were issues setting this previously, and I didn't change anything in this PR

I changed the default in 2.0, we know this works because we have a test that runs both the whole pod with a non-root security context.

> This just changes the UID in the init container, but wouldn't it be necessary to change the default UID of user pipelines? That sounds completely insane though, and I'd expect this to break every pipeline.

What happens is that, absent a security context, the init container runs as UID 1000, and the user container runs as whatever UID is baked in. I did some testing, and the files that the init container installs end up having 777 permissions, so they're executable even if the user container UID is a different, unprivilleged container. For instance, I ran a pipeline where the init container is UID 1000 and the user container is UID 1001 with no problem.

We should add an integration test which builds a non-root, non UID 1000 container and confirms that a pipeline runs end-to-end with that setup, in addition to the existing test that runs both containers with UID 1000 using a pod security context.

actgardner (approved):

msteffen:
Since CI is passing, and non-root is better than root, I'll merge :shrug: 

@actgardner thanks for the super-thorough explanation! Part of where my uncertainty was coming from was that, in the limited experimentation I did, I had some issues running a `local` pachd deployment as a non-root user due to the `pach-disk` HostPath volume we use instead of object storage; the pachd process couldn't access it (I was setting the UID in the Dockerfile instead of the pod security policy too, which might have mattered), and I was worried `pach-bin` would have the same problems. The fact that it's an EmptyDir resolves the main potential issue with this specific PR for me, though, and `pach-disk` clearly now works in master too, so even the related issue I've observed must be solveable; I think this is just an area where I need to improve my understanding of k8s volumes.

actgardner:
> Part of where my uncertainty was coming from was that, in the limited experimentation I did, I had some issues running a local pachd deployment as a non-root user due to the pach-disk HostPath volume we use instead of object storage: the pachd process couldn't access the mount at /pach (I was setting the UID in the Dockerfile instead of the pod security policy too, which might have mattered).

This issue affects pachd and the sidecar container when running as non-root users, because they don't usually have the right permissions on the host filesystem. The way we've worked around that is to run as non-root by default, but add a pod security context when users do a minikube deployment that runs as root. In CI we do some prep work to set the permissions on the host so the host volume works without the pod security context. Hopefully that clarifies a bit!

PR #6418, opened 2021-06-22
Remove old IAM role annotation

chainlink:
Best practice now is to set the role ARN on the service account instead of the deployment

https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html

The new functionality is supported in the helm chart


chainlink:
cc @tybritten

PR #6420, opened 2021-06-23
Add pachctl inspect jobset/commitset

Tryneus:
This PR adds the commands:
 * `pachctl inspect jobset`
 * `pachctl wait jobset`
 * `pachctl inspect commitset`
 * `pachctl wait commitset` - broken out from `pachctl wait commit` so each one takes a more specific input format
 * `pachctl wait job` - for parity with commit commands


src/server/cmd/pachctl/cmd/cmd.go
```diff
@@ -375,14 +371,15 @@ Environment variables:
 		Short: "Print Pachyderm version information.",
 		Long:  "Print Pachyderm version information.",
 		Run: cmdutil.RunFixedArgs(0, func(args []string) (retErr error) {
+			if !raw && output != "" {
```
PFedak, line 374: I see that this is consistent with what we have elsewhere, but this flag structure has never made much sense to me. I think two or three flags (`--raw`, `--json`, `--yaml` where the first two are equivalent?) would represent things a bit better.

src/server/pfs/cmds/cmds.go
```diff
@@ -458,43 +462,26 @@ $ {{alias}} foo@master --from XXX`,
...
 
 	var branches cmdutil.RepeatedStringArg
 	waitCommit := &cobra.Command{
-		Use:   "{{alias}} ( <commitset-id> | <repo>@<branch-or-commit> )",
-		Short: "Wait for the specified commit(s) to finish and return them.",
-		Long:  "Wait for the specified commit(s) to finish and return them.",
+		Use:   "{{alias}} <repo>@<branch-or-commit>",
+		Short: "Wait for the specified commit to finish and return it.",
+		Long:  "Wait for the specified commit to finish and return it.",
 		Example: `
```
PFedak, line 475: move these examples down to wait commitset

src/server/pfs/cmds/cmds.go
```diff
@@ -503,74 +490,26 @@ $ {{alias}} foo@XXX -b bar@baz`,
...
-			return nil
+			return pretty.PrintDetailedCommitInfo(os.Stdout, ci)
 		}),
 	}
-	waitCommit.Flags().VarP(&branches, "branch", "b", "Wait only for commits in the specified set of branches")
-	waitCommit.MarkFlagCustom("branch", "__pachctl_get_branch")
-	waitCommit.Flags().AddFlagSet(rawFlags)
-	waitCommit.Flags().AddFlagSet(fullTimestampsFlags)
+	waitCommit.Flags().AddFlagSet(outputFlags)
+	waitCommit.Flags().AddFlagSet(timestampFlags)
```
PFedak, line 512: should this register a completion function?

src/server/pfs/cmds/cmds.go
```diff
@@ -624,11 +569,101 @@ $ {{alias}} test@master --new`,
...
+			for _, commitInfo := range commitInfos {
+				pretty.PrintCommitInfo(writer, commitInfo, fullTimestamps)
+			}
+			return writer.Flush()
+		})
+	}
+
+	waitCommitSet := &cobra.Command{
+		Use:   "{{alias}} <commitset-id>",
+		Short: "Return info about the commits in a commitset.",
```
PFedak, line 601: presumably this should have "wait" somewhere

src/server/pfs/cmds/cmds.go
```diff
@@ -624,11 +569,101 @@ $ {{alias}} test@master --new`,
...
+			client, err := client.NewOnUserMachine("user")
+			if err != nil {
+				return err
+			}
+			defer client.Close()
+
+			var commitInfos []*pfs.CommitInfo
+			if len(toBranches) != 0 {
+				for _, branch := range toBranches {
+					ci, err := client.WaitCommit(branch.Repo.Name, branch.Name, args[0])
```
PFedak, line 622: Oh huh, didn't realize this wasn't part of the gRPC

src/server/pfs/cmds/cmds.go
```diff
@@ -624,11 +569,101 @@ $ {{alias}} test@master --new`,
...
+					if err != nil {
+						return errors.Wrap(err, "error from InspectCommit")
```
PFedak, line 624: I'm wondering if it would make more sense to someone using this to say WaitCommit here, rather than the name of the gRPC

src/server/pps/cmds/cmds.go
```diff
@@ -141,12 +96,95 @@ If the job fails, the output commit will not be populated with data.`,
...
+			pji := &pretty.PrintableJobInfo{
+				JobInfo:        jobInfo,
+				FullTimestamps: fullTimestamps,
+			}
+			return pretty.PrintDetailedJobInfo(os.Stdout, pji)
+		}),
+	}
+	waitJob.Flags().AddFlagSet(outputFlags)
+	waitJob.Flags().AddFlagSet(timestampFlags)
+	shell.RegisterCompletionFunc(waitJob, shell.JobCompletion)
```
PFedak, line 136: I think this name is misleading now; this just completes IDs rather than pipeline@id

src/server/pps/cmds/cmds.go
```diff
@@ -141,12 +96,95 @@ If the job fails, the output commit will not be populated with data.`,
...
+	commands = append(commands, cmdutil.CreateAlias(waitJob, "wait job"))
+
+	inspectJobSet := &cobra.Command{
```
PFedak, line 139: Did I miss some discussion about why `inspect jobset` should take a wait flag rather than splitting that out?

src/server/pps/cmds/cmds.go
```diff
@@ -192,15 +230,16 @@ $ {{alias}} -p foo -i bar@YYY`,
...
+					return e.EncodeProto(ji)
+				})
+			} else if output != "" {
+				return errors.New("cannot set --output (-o) without --raw")
+			}
+
 			return pager.Page(noPager, os.Stdout, func(w io.Writer) error {
-				if raw {
-					e := encoder(output)
-					return client.ListJobFilterF(pipelineName, commits, history, true, filter, func(ji *ppsclient.JobInfo) error {
```
PFedak, line 198: was this just broken with the pager before?

PFedak (approved):
LGTM, just a couple things with help strings and completion that would be good to fix

src/server/pachyderm_test.go
```diff
@@ -2691,10 +2691,14 @@ func TestUpdateFailedPipeline(t *testing.T) {
 	require.NoError(t, c.FinishCommit(dataRepo, "master", ""))
 
 	// Wait for pod to try and pull the bad image
-	time.Sleep(10 * time.Second)
-	pipelineInfo, err := c.InspectPipeline(pipelineName, false)
-	require.NoError(t, err)
-	require.Equal(t, pps.PipelineState_PIPELINE_CRASHING.String(), pipelineInfo.State.String())
+	require.NoErrorWithinTRetry(t, time.Second*30, func() error {
```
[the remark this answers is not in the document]
Tryneus: Yes - I ran into a flake on this test and wanted to prevent it from happening again.

src/server/cmd/pachctl/cmd/cmd.go
Tryneus: We'd still have similar situations where we have to deal with people passing both `--raw` and `--yaml` (or `--json` and `--yaml`), so I don't think it would be more concise.  I'm open to rethinking how we do this, but a little out-of-scope of this PR since this just maintains the status quo in this regard.

src/server/pfs/cmds/cmds.go
Tryneus: ah, good catch

src/server/pfs/cmds/cmds.go
Tryneus: It's just a porcelain function around `InspectCommit` with `wait=true`

src/server/cmd/pachctl/cmd/cmd.go
PFedak: I was thinking `--raw` would just be there for legacy support, but that's a good point about checking for `--json` and `--yaml`. It just seems annoying to have to do `--raw --yaml` to get yaml now. Agreed it's out of scope

src/server/pfs/cmds/cmds.go
Tryneus: I _think_ `InspectCommit` is better because it is what will show up in the pachd logs.

src/server/pps/cmds/cmds.go
Tryneus: For all of the `Wait` operations, it's just a `wait=true` flag on the inspect.

src/server/pps/cmds/cmds.go
Tryneus: I am not sure, I just wanted to make this more consistent - I think when outputting encoded data, we shouldn't use the pager (and I believe that is how other places in the codebase do it).

src/server/cmd/pachctl/cmd/cmd.go
Tryneus: In that case I'd rather just do a breaking change and drop `--raw` entirely, so you would just do `--output=json` or `--output=yaml` and the error case goes away entirely and we don't have to check the consistency of flags.

src/server/pfs/cmds/cmds.go
Tryneus: I think we need to do a completion update in a separate pass, it's definitely out-of-sync with a lot of the changes in 2.0.

PR #6421, opened 2021-06-23
Re-enable spout auth test

actgardner:
In 1.x we had a test that created a spout without auth, and then activated auth, and the pipeline would get stuck until you updated the spec. This seems like a bug we turned into a test. 

Update the test to activate auth and expect the spout to stay in a running state without raising an error, since activating auth should restart the pipeline.

### DIFF
--- a/src/server/spout_test.go
+++ b/src/server/spout_test.go
@@ -37,9 +37,6 @@ func TestSpoutPachctl(t *testing.T) {
 		defer tu.DeleteAll(t)
 		c := tu.GetAuthenticatedPachClient(t, auth.RootUser)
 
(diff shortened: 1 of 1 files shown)


src/server/spout_test.go
```diff
@@ -37,9 +37,6 @@ func TestSpoutPachctl(t *testing.T) {
 		defer tu.DeleteAll(t)
 		c := tu.GetAuthenticatedPachClient(t, auth.RootUser)
 
-		dataRepo := tu.UniqueString("TestSpoutAuth_data")
```
actgardner, line 40: `dataRepo` gets created in a bunch of these tests but it's not used, remove it.

brycemcanally (approved):
lgtm

PR #6422, opened 2021-06-23
Fix double proto import

actgardner:


jrockway (approved):

PFedak (approved):
Huh, wonder how that happened. Is goimports not smart enough to check all the imports?

actgardner:
Both branches passed CI, and there wasn't a Git merge conflict because it was added in different blocks. But the merge failed CI.

PR #6424, opened 2021-06-23
Re-enable PutFileAtomic test

actgardner:


brycemcanally:
Did we decide to change the semantics of empty one-off requests in 2.0? In 1.x, a commit wouldn't be created in this case, but it looks like the test was updated to check for a new commit in this case.

brycemcanally (approved):
lgtm

Issue #6425, opened 2021-06-23
Pipeline in inner_join example crashes on 2.0.0-alpha.24


nadegepepin:
in the joins examples https://github.com/pachyderm/pachyderm/tree/master/examples/joins:

- `make setup`
- `make create`
- `pachctl create pipeline -f inner_join.json` (make sure that you have updated the tag of the image to 2.0)
->>`"transform": {
  "cmd": [ "python", "inner/main.py" ],
  "image": "pachyderm/example-joins-inner-outer:2.0"
}`

`pachctl list pipeline`
>>NAME       VERSION INPUT CREATED STATE / LAST JOB  DESCRIPTION
outer_join -       -     -       running / failure pipeline details unavailable

`pachctl list job`
>>ID                               PIPELINE   STARTED           DURATION  RESTART PROGRESS  DL   UL STATE
883a018b28004ec0a8852c85b5d2fa99 outer_join About an hour ago 7 seconds 0       0 + 0 / 7 760B 0B failure: datum 8c2ba186bd32e96600a...

`pachctl logs --job=outer_join@883a018b28004ec0a8852c85b5d2fa99`
>>python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
python: can't open file '//outer/main.py': [Errno 2] No such file or directory
[...]

PR #6426, opened 2021-06-24
Don't override workdir when none is supplied

actgardner:
Resolves https://github.com/pachyderm/pachyderm/issues/6425

Docker images have a WORKDIR, which is PWD when the container starts.  The Pachyderm pipeline spec also has an optional `working_dir` field which overrides the WORKDIR. Previously if the working_dir was empty, we would use the docker socket to inspect the image and set the working_dir to the image WORKDIR, and then set PWD for the user-specified command to that value.

Since we removed the docker socket, we no longer inspect the image and set `working_dir` when it's empty. This is fine, because PWD is already the WORKDIR for the image. We just need to not override it when `working_dir` is empty.

### DIFF
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/docker/go-units"
+	"github.com/gogo/protobuf/proto"
 	"github.com/pachyderm/pachyderm/v2/src/auth"
 	"github.com/pachyderm/pachyderm/v2/src/client"
 	"github.com/pachyderm/pachyderm/v2/src/internal/ancestry"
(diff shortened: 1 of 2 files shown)


src/server/pachyderm_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/docker/go-units"
+	"github.com/gogo/protobuf/proto"
```
PFedak, line 25: hmm, did someone just mess up a merge before?

PFedak (approved):
LGTM

src/server/pachyderm_test.go
actgardner: Yeah master is broken again, i'll ask everyone to rebase

PR #6428, opened 2021-06-24
Revert "DirInserter always inserts root (#6277)"

brendoncarroll:
This reverts commit b58ffbdcb66dd946b87dd5abd5f9173501e296e0.

### DIFF
--- a/src/client/pfs_file.go
+++ b/src/client/pfs_file.go
@@ -701,9 +701,6 @@ func (c APIClient) DiffFileAll(newCommit *pfs.Commit, newPath string, oldCommit 
 
 // WalkFile walks the files under path.
 func (c APIClient) WalkFile(commit *pfs.Commit, path string, cb func(*pfs.FileInfo) error) (retErr error) {
-	defer func() {
-		retErr = grpcutil.ScrubGRPC(retErr)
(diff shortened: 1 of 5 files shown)


src/client/pfs_file.go
```diff
@@ -701,9 +701,6 @@ func (c APIClient) DiffFileAll(newCommit *pfs.Commit, newPath string, oldCommit
 
 // WalkFile walks the files under path.
 func (c APIClient) WalkFile(commit *pfs.Commit, path string, cb func(*pfs.FileInfo) error) (retErr error) {
-	defer func() {
-		retErr = grpcutil.ScrubGRPC(retErr)
-	}()
```
Tryneus, line 706: Why was this removed?  I think we still want to scrub the error here, unless we're going to do a pass across all the client code and make this consistent.

Tryneus (approved):
One nit, otherwise LGTM

src/client/pfs_file.go
brendoncarroll: This was removed so that the gRPC status code can be inspected.  The test now checks for that.  I don't think we need to move all the error handling over right now.

PR #6429, opened 2021-06-24
CI performance improvements

actgardner:
This has two fixes to improve CI performance:

Some latency hiding for the Loki deploy. We used to deploy Loki and then sit idly while k8s pulls the images, etc. This deploys the helm chart and then goes away to do the pachd build, and comes back to check on the progress later.

Recompute the cache more often. Right now the cache key doesn't change between runs, so the cache would get staler and staler for 30 days until it expired. This PR keys the module cache on the go.sum checksum, and accumulates the build cache for a week before rotating it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,15 +43,22 @@ jobs:
       - run: etc/testing/circle/launch-loki.sh
       - restore_cache:
          keys:
-         - pach-go-cache-{{ .Branch }}
-         - pach-go-cache-
+         - pach-go-build-cache-{{ .Branch }}
+         - pach-go-build-cache
+      - restore_cache:
+         keys:
+         - pach-go-mod-cache-{{ .Branch }}
+         - pach-go-mod-cache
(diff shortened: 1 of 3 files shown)


jrockway (approved):
Very good cleanup.  Do you have a good way of seeing if the cache helps?  I went through many iterations of caching for Hub, and found that `npm install` and <restore the result of npm install> took almost exactly the same amount of time -- the bottleneck seemed like the container -> network, and not network -> internet, sadly.

(Caching golangci-lint helped a lot though, because it precomputes a lot of stuff and it's very slow the first time, and pretty fast the next time.)

actgardner:
> Do you have a good way of seeing if the cache helps?

In the perfect case (no code changes), the build takes 1 minute, saving the cache takes 1.5 minutes, restoring the cache takes ~40s. By comparison, master today building with a stale cache takes ~5.5 minutes. About 3 minute savings (although I wish the cache performance was better). I'll keep an eye on how master and new branches perform and potentially do some tweaking.

PR #6430, opened 2021-06-24
ListPipeline omits details when a user doesn't have access

actgardner:
When a user does `ListPipeline` and requests details, we attempt to deserialize the spec commit for every pipeline into the response. A user may not have permissions to view every spec commit - previously this caused ListCommit to fail. Instead, this PR omits the details from a pipeline entry when the user doesn't have access. This shouldn't have much user-facing impact since `pachctl list pipeline` doesn't request the details or display any information from them.

This also removes the `--spec` flag for `pachctl list pipeline`, because (presumably for performance reasons) `pachctl list pipeline` no longer requests the spec commits for every pipeline, which meant the command was useless and stack-traced.

### DIFF
--- a/src/internal/ppsutil/util.go
+++ b/src/internal/ppsutil/util.go
@@ -269,39 +269,6 @@ func JobInput(pipelineInfo *pps.PipelineInfo, outputCommit *pfs.Commit) *pps.Inp
 	return jobInput
 }
 
-// PipelineReqFromInfo converts a PipelineInfo into a CreatePipelineRequest.
-func PipelineReqFromInfo(pipelineInfo *pps.PipelineInfo) *pps.CreatePipelineRequest {
-	return &pps.CreatePipelineRequest{
-		Pipeline:              pipelineInfo.Pipeline,
-		Transform:             pipelineInfo.Details.Transform,
-		ParallelismSpec:       pipelineInfo.Details.ParallelismSpec,
-		Egress:                pipelineInfo.Details.Egress,
-		OutputBranch:          pipelineInfo.Details.OutputBranch,
-		ResourceRequests:      pipelineInfo.Details.ResourceRequests,
-		ResourceLimits:        pipelineInfo.Details.ResourceLimits,
-		SidecarResourceLimits: pipelineInfo.Details.SidecarResourceLimits,
-		Input:                 pipelineInfo.Details.Input,
-		Description:           pipelineInfo.Details.Description,
-		CacheSize:             pipelineInfo.Details.CacheSize,
-		MaxQueueSize:          pipelineInfo.Details.MaxQueueSize,
-		Service:               pipelineInfo.Details.Service,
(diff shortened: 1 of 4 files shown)


src/internal/ppsutil/util.go
```diff
@@ -269,39 +269,6 @@ func JobInput(pipelineInfo *pps.PipelineInfo, outputCommit *pfs.Commit) *pps.Inp
 	return jobInput
 }
 
-// PipelineReqFromInfo converts a PipelineInfo into a CreatePipelineRequest.
```
actgardner, line 272: This was only used in `pachctl list pipeline --spec`, so I'm removing it.

Tryneus: Ooh, this is actually being used in #6406, so maybe leave it in.

src/server/auth/server/testing/auth_test.go
```diff
@@ -544,19 +544,34 @@ func TestCreateAndUpdatePipeline(t *testing.T) {
...
-	// TODO(2.0 required): this call errors if it uses aliceClient (maybe skip
-	// pipelines we don't have read access on?)
-	// pipelineInfos, err = aliceClient.ListPipeline(true)
+
+	// Check that an unprivileged user can list pipelines with details,
+	// but the details are skipped
+	pipelineInfos, err = aliceClient.ListPipeline(true)
+	require.NoError(t, err)
+	require.Equal(t, 2, len(pipelineInfos))
+	for _, pipelineInfo := range pipelineInfos {
```
Tryneus, line 560: I think it's actually bad to make the results kind of polymorphic like this, but I suppose we don't have any prior art in the codebase.  I feel like if we request details we should always return details and the correct behavior might be to filter out those pipelines?

For example, there is a function in `completions.go` - `PipelineCompletion` that will crash now.  Also, I believe the `PPS.ActivateAuth` RPC could crash `pachd` if called on a server with auth already active by an unprivileged user.

Tryneus (changes requested):
Some concerns about the overall direction here

 > This also removes the --spec flag for pachctl list pipeline, because (presumably for performance reasons) pachctl list pipeline no longer requests the spec commits for every pipeline, which meant the command was useless and stack-traced.

That may have been an oversight by me when adding the `PipelineInfo.Details` field, and it might be better to leave it in unless we know people don't want to use that flag.

src/server/auth/server/testing/auth_test.go
actgardner: I think from a product perspective omitting pipelines when a user doesn't have access is weird - it'll probably break the DAG in dash, for instance. I think we want to limit listing repos/pipelines at the project level, not the individual repo/pipeline level.

That said, dash also appears to rely on being able to access the details of the pipelines. So it kind of seems like the option here is to make spec commits world-readable, which would be consistent with the behaviour of 1.x.

actgardner:
Dropping this in favour of making spec repos world-readable, since they're an implementation detail.

PR #6432, opened 2021-06-25
Spec commits are world-readable

actgardner:
When we added spec repos, we decided that spec repos could just inherit the same permissions as the associated user repo. This mostly matches the permission model in 1.x, except that spec commits were world-readable in 1.x to make `ListPipeline` work. The fact that pipeline specs live in PFS repos is an implementation detail end-users are unlikely to care about - the user-facing functionality we're aiming for is "users can list pipelines". When we add projects we'll support listing pipelines per project, and we'll need to extend this, but for now it suffices to make all spec pipelines readable by all authenticated users.


Tryneus (approved):
LGTM

I have some vague doubts about how we treat the filtering of things based on permissions.  For example, PPS will skip over jobs from pipelines the user doesn't have access to, but we don't apply the same behavior to pipelines.  But this appears to maintain the status quo so it's probably fine.

actgardner:
Yeah I'm not sure how I feel about that filtering behaviour for ListJobs? Maybe I'll ask customer eng if they want to revisit that in 2.0.

PR #6435, opened 2021-06-25
Don't transition a pipeline to failed during auth activation

actgardner:
During auth activation a pipeline won't have an auth token associated briefly, which can race with the PPS master recreating the RC. When this happens the pipeline would transition into the failure state, but we actually should retry.


Tryneus (approved):
LGTM, good catch!

PR #6437, opened 2021-06-25
Switch S3 tests to use MFC instead of RPC per file

actgardner:
`TestMasterDriver/master/ListObjectsPaginated` was one of our slowest tests, according to Circle, at 2.5 minutes. This PR switches it from doing 1000 serial RPCs to put files, to using a ModifyFileClient and streaming the updates. Testing locally this whole package takes 5 seconds with this change.


brendoncarroll (approved):
LGTM

brycemcanally:
Be sure to check the error returned from `WithModifyFileClient`.

PR #6442, opened 2021-06-28
[WIP] List commit changes

brycemcanally:
This PR contains a few changes to list commit:
- Expose the details configuration through the list commit Go client wrapper and pachctl.
- Change the list commit Go client wrapper to an interface we can commit to longer term. The current design is to take the request proto in the wrapper and provide a builder for the request proto.
- Compute the correct commit size (after applying deletions).


src/pfs/pfs.go
```diff
@@ -78,3 +78,32 @@ func (b *Branch) NewCommit(id string) *Commit {
 func (b *Branch) String() string {
 	return b.Repo.String() + "@" + b.Name
 }
+
+func NewListCommitRequest(repo *Repo) *ListCommitRequest {
+	return &ListCommitRequest{Repo: repo}
+}
+
+func (lcr *ListCommitRequest) WithTo(to *Commit) *ListCommitRequest {
```
brendoncarroll, line 86: This API is a little misleading with the pointers.  The requests aren't immutable, they are being modified and then returned.  I would expect the receiver not to be a pointer, but maybe there are performance issues with copying like that.

brycemcanally: Yeah, I had concerns about the performance, but those may be unfounded. I would like to use pointers because that is what gRPC takes, and we could change this to a copy / clone under the hood later.

jdoliner: I have a feeling for almost all requests it'd be fine to clone. It maybe gets to be a bit of overhead if it's heavily chained. That being said I don't find the api that confusing, with does imply that it's not mutating anything but I can't really think of a better name, you could maybe do `SetTo` but that somehow doesn't seem as good to me.

PR #6445, opened 2021-06-28
[WIP] Removed nested watcher in S3 sidecar

actgardner:
This seems too simple - as the TODO comment says, the original comment is incorrect:

> 'watcher' will not see any job state updates after the first because job state updates don't update the pipelines index, so this establishes a watcher that will.

The pipelines index does see the writes to the the job state, so this seems to be a straightforward change to watch for the job to start, and then to wait for it to reach a terminal state. One invariant is that there's one job per pipeline, which I'm not 100% confident about?

### DIFF
--- a/src/server/pps/server/s3g_sidecar.go
+++ b/src/server/pps/server/s3g_sidecar.go
@@ -119,6 +119,8 @@ func (s *sidecarS3G) serveS3Instances() {
 
 func (s *sidecarS3G) createK8sServices() {
 	logrus.Infof("Launching sidecar s3 gateway master process")
+	// TODO: This is one gateway per pipeline, but can multiple jobs run per pipeline at once?
+	// Is there a risk of a new job seeing old data if the S3 gateway is slow to process events?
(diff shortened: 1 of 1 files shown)


msteffen:
I'm on my phone at the moment, but I wanted to send a quick high-level response

> TODO: This is one gateway per pipeline, but can multiple jobs run per pipeline at once?

In 1.x, there can be multiple concurrent jobs running for the same pipeline. I think in 2.x, right now, there can't, but I think ( @Tryneus will know for sure) that the hope is to bring that back in 2.x as well.

> // Is there a risk of a new job seeing old data if the S3 gateway is slow to process events?

The goal (bugs are always possible) is that there shouldn't be—each job (and soon, each in-process datum, so # up to the parallelism of the pipeline) should have a unique k8s service created for it that only exposes that job's (and soon, that datum's) input data

As far as the index seeing every write—I could be super wrong but I think it's possible that this may be new with postgres, since indexes in 1.x were implemented in the collections library. If so, other uses of WatchByIndex may be affected too

Issue #6369, opened 2021-06-17
Diff File empty. Prints "pattern contains path separator"


brendoncarroll:
**What happened?**:
This is in the market-sentiment example.  Same repo different files (note the different sizes)
```
➜  market-sentiment git:(master) ✗ pachctl inspect file financial_phrase_bank@v2
Path: /
Tag:
Type: dir
Size: 292.6KiB
➜  market-sentiment git:(master) ✗ pachctl inspect file financial_phrase_bank@v1
Path: /
Tag:
Type: dir
Size: 655.6KiB
➜  market-sentiment git:(master) ✗ pachctl diff file financial_phrase_bank@v1 financial_phrase_bank@v2
pattern contains path separator
```

How to reproduce
Create two branches in the same repo, put different files at the same path, diff them.

**Environment?**:
- Pachyderm CLI and pachd server version (use `pachctl version`):
- Alpha 23
- Cloud provider (e.g. aws, azure, gke) or local deployment (e.g. minikube vs dockerized k8s):
Hub
- OS (e.g. from /etc/os-release):
macOS
- Others:

brycemcanally:
Similar issue here: https://github.com/pachyderm/pachyderm/issues/6364

Issue #6364, opened 2021-06-17
Potential issue with pachctl diff file wrapper


brycemcanally:
We are seeing an error when attempting basic usage of pachctl diff file: `pattern contains path separator`. Need to take a pass over this wrapper to ensure it is functioning correctly with 2.0. In general, we need more comprehensive automated testing of pachctl because several issues that have popped up through game days are related to issues with pachctl rather than underlying issues with the Go client or backend.

acohen4:
Closing with merge of fix: #6446

PR #6446, opened 2021-06-28
Fix `pachctl diff file` error: 'pattern contains path separator'

acohen4:
Addressing issue: https://github.com/pachyderm/pachyderm/issues/6364

I assume that this error was introduced when we updated to a more recent version of the `filepath` module


brendoncarroll (approved):
LGTM

brendoncarroll:
Closes #6369

PR #6447, opened 2021-06-29
Fix some problems with multipart uploads.

PFedak:
This fixes two large issues with our s3 multipart upload support. The
first impacts correctness: previously, the file parts were put into
destination directory piece by piece, in separate commits, which could
trigger downstream jobs with partial states. Instead, we now build the
file incrementally in a uniquely named file in the multipart repo and
copy it into the destination in one go.

The second is a bit more nebulous. I have observed the aws cli client
making multiple concurrent CompleteMultipartUpload requests. The AWS
documentation isn't clear on this case, but the client seems like it
expects subsequent calls to succeed until one of them has finished. In
our system, this will just cause identical jobs, so I have attempted to
accommodate this behavior.

### DIFF
--- a/src/server/pfs/s3/multipart.go
+++ b/src/server/pfs/s3/multipart.go
@@ -8,6 +8,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/gogo/protobuf/types"
 	"github.com/pachyderm/pachyderm/src/client"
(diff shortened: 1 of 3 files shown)


src/server/pfs/s3/multipart.go
```diff
@@ -280,17 +275,20 @@ func (c *controller) CompleteMultipart(r *http.Request, bucketName, key, uploadI
 			return nil, s2.EntityTooSmallError(r)
 		}
 
-		err = pc.CopyFile(c.repo, "master", srcPath, bucket.Repo, bucket.Commit, key, false)
-		if err != nil {
-			if errutil.IsWriteToOutputBranchError(err) {
-				return nil, writeToOutputBranchError(r)
-			}
+		if err := pc.CopyFile(c.repo, "master", srcPath, c.repo, "master", dstPath, false); err != nil {
```
acohen4, line 278: @PFedak why does writing to the random path not create additional commits?

acohen4: nevermind i just noticed this is in another repo

PFedak: It does, but only in the special `_s3gateway_multipart_` repo, rather than the destination. It might have made sense to use different branches or even different repos for separate uploads so that they could use some of the put file client methods rather than all creating new commits, but this does make it easy to not worry about concurrency.

acohen4 (approved):
Looks good!

mindthevirt:
<div>Agent Dale Georg linked Freshdesk ticket <a href="https://pachyderm-support.freshdesk.com/a/tickets/292">292</a> for this issue.</div>

Issue #5743, opened 2021-02-25
Issues deploying the Pachyderm IDE


actgardner:
**What happened?**:

Ran `pachctl deploy ide` on a cluster where enterprise and auth are enabled. We ran into several errors:

The Helm package in pachctl is unable to create resources in the k8s cluster, even though running the Helm CLI works:
```
failed to deploy Pachyderm IDE: install failed: create: failed to create: an empty namespace may not be set during creation
```

Once we deployed manually with Helm, the `1.1.0` images are missing from DockerHub, so we had to manually specify the images to use:

```
pachctl deploy ide --dry-run --hub-image pachyderm/ide-hub:1.0.0 --user-image pachyderm/ide-user:1.0.0 > values.json
```

The 1.0 Hub image appears to be missing the health check endpoint, so the pod would never become ready. logs:

```
[I 2021-02-25 14:38:50.241 JupyterHub proxy:301] Checking routes
[W 2021-02-25 14:38:54.117 JupyterHub log:158] 404 GET /hub/health (@172.17.0.1) 1.35ms
[W 2021-02-25 14:39:04.118 JupyterHub log:158] 404 GET /hub/health (@172.17.0.1) 1.24ms
[W 2021-02-25 14:39:14.117 JupyterHub log:158] 404 GET /hub/health (@172.17.0.1) 1.13ms
[W 2021-02-25 14:39:24.117 JupyterHub log:158] 404 GET /hub/health (@172.17.0.1) 1.26ms
[W 2021-02-25 14:39:34.118 JupyterHub log:158] 404 GET /hub/health (@172.17.0.1) 1.47ms
[W 2021-02-25 14:39:44.117 JupyterHub log:158] 404 GET /hub/health (@172.17.0.1) 1.34ms
```

I was able to work around this by disabling the readiness check in the Helm chart:

```
helm upgrade --install pachyderm-ide jupyterhub/jupyterhub --version 0.9.1 -f values.json --set hub.readinessProbe.enabled=false
```

**What you expected to happen?**:

The IDE should be deployed.

**How to reproduce it (as minimally and precisely as possible)?**:

We don't have a clean reproduction for the first Helm issue. The rest of the errors happen on the happy path.

**Anything else we need to know?**:

**Environment?**:
- Kubernetes version (use `kubectl version`): 1.19.3
- Pachyderm CLI and pachd server version (use `pachctl version`): 1.12.3
[...]

PR #6448, opened 2021-06-29
Ensure IDE deployment sets a namespace.

PFedak:
Also add --namespace analogous to other deploy commands.

Should fix at least the first problem in #5743


jdoliner (approved):
LGTM

PR #6449, opened 2021-06-29
Postgres: make tuneables configurable, monitor connection pool status, monitor transaction success

jrockway:
This adds some code that helps with tuning Postgres.  All configurable connection pool parameters can be controlled through environment variables.  The defaults for a production pachd are now set in the GlobalConfiguration, but a comment points from `dbutil`'s constants there, so it shouldn't be confusing if you ever need to change them.  I also did some testing to ensure that the configuration actually takes effect.

I also added some monitoring that exports connection pool stats to Prometheus (list of metrics here: https://github.com/dlmiddlecote/sqlstats/blob/main/collector_115.go#L28).  When collected and hooked into Grafana, it looks something like:

Capture

(I manually verified that these look reasonable during CI runs, and that we don't hit the error case where we try to add the metrics more than once.)

I also added some monitoring for transactions run through dbutil/WithTx.  The number of transactions started, finished (by outcome), and their latency is recorded.  The number of underlying database transactions started and finished are also recorded (exposing details such as whether the transaction committed or rolled back).  Note that we still have no visibility on queries like `env.WithDBClient().ExecContext(...)`; I think we would have to change the Postgres driver to `pgx` to have a good place to hook in there.

Finally, I made CI dump all prometheus metrics to the failure log if the tests fail -- may be helpful.  While developing this PR, I noticed a bug where I changed the connection limit to 0 instead of the default of 10, and that caused tests to fail.  I ran the job under SSH, probed the prometheus metrics, and noticed that the new DB metrics also said the limit was 0.  Suspicious!  I fixed the code that incorrectly changed the default, and all was well.  With the metrics dump, I would have saved 15 minutes ;)

This should let us keep an eye on our interactions with Postgres and easily tune it for optimal performance.

### DIFF
--- a/etc/testing/circle/kube_debug.sh
+++ b/etc/testing/circle/kube_debug.sh
@@ -25,6 +25,7 @@ cmds=(
   'kubectl get all --all-namespaces'
   'kubectl describe pod -l suite=pachyderm,app=pachd'
   'kubectl describe pod -l suite=pachyderm,app=etcd'
+  'curl -s http://$(minikube ip):30656/metrics'
   # Set --tail b/c by default 'kubectl logs' only outputs 10 lines if -l is set
   'kubectl logs --tail=1500 -l suite=pachyderm,app=pachd'
   'kubectl logs --tail=1500 -l suite=pachyderm,app=pachd --previous # if pachd restarted'
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/dexidp/dex v0.0.0-20201118094123-6ca0cbc85759
 	github.com/dexidp/dex/api/v2 v2.0.0
 	github.com/dlclark/regexp2 v1.2.0 // indirect
+	github.com/dlmiddlecote/sqlstats v1.0.2 // indirect
 	github.com/docker/go-units v0.4.0
 	github.com/docker/spdystream v0.0.0-20181023171402-6480d4af844c // indirect
 	github.com/elazarl/goproxy v0.0.0-20191011121108-aa519ddbe484 // indirect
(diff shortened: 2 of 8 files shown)


brendoncarroll:
This is great 😁.  The defaults in multiple places is not ideal, but I think your comments are a good enough solution for now.

Is there any magic going on with sqlstats?  I notice you give it "postgres", does it try to attach to the postgres driver in any way?

Just wanted to confirm we can get at the number of retries per transaction by subtracting the metrics you have added?

brendoncarroll (approved):

jrockway:
There isn't any magic going on with sqlstats; "postgres" just results in the metric being tagged with a certain "name" field.  (After we add pgbouncer and have two connections, we'll probably have "direct" and "pgbouncer" as the two names.)  The stats come from the `database/sql` library: https://pkg.go.dev/database/sql#DB.Stats.  The interesting one in there is the amount of time waited on for a connection -- so if we set the max connections to 10, and they start blocking, we have a sum of the time spent blocking (and can adjust that).  The count of actually-open connections is also useful (I was surprised on Hub on how the GraphQL planner will execute 50 SQL queries in parallel to answer one user query; I couldn't believe something like that was happening, but this metric led me to finding the code that did that).

The exact number of retries isn't directly exported.  Prometheus will collect these metrics every X seconds (usually something like 15), and we can take a delta between two samples to get the total number of transactions run in that time period, and the total number of database transactions in that time period, and from those two numbers, we can guess at it.  It's an average, so if one WithTx call runs 3 database transactions, and another one runs 1, we'll calculate 2 database transactions per WithTx call.  Probably not ideal; I'll just add an explicit metric.

jrockway:
I added a "tries per tx" metric that records a histogram of how many tries were required per transaction, along with the outcome.  This directly measures how many database transactions were required to resolve a WithTx, and will let us get data like "in the last 15 seconds, 1 failing transaction took 5 attempts, and 3 successful transactions took 2 attempts, and 1 successful transaction took 1 attempt".  (Some resolution is lost for large values; 999 attempts is indistinguishable from 998 attempts, with the bucketing that I defined.  There is resolution where we care, however.)

Another way to do this is to add an additional "count" label to tx_finished, and just put the count there.  That doesn't lose resolution, but tends to be slow (at prometheus querying/storage time) if the cardinality of "count" is high.

PR #6452, opened 2021-06-30
Unskip most stats-related tests.

PFedak:
TestPipelineWithStats and TestPipelineWithStatsFailedDatums didn't change much, TestPipelineWithStatsAcrossJobs required adding a simple check for whether datums were skipped, and the behavior in TestPipelineWithStatsSkippedEdgeCase no longer applies.


src/pps/pps.proto
```diff
@@ -158,13 +158,12 @@ message Datum {
 }
 
 enum DatumState {
-  DATUM_STATE_UNKNOWN = 0;
+  DATUM_STATE_UNKNOWN = 0; // or not part of a job
```
brycemcanally, line 161: Nit, but this could just be unknown. The datum state part is implicit in the type.

brycemcanally (approved):
lgtm, just a minor change to the unknown datum state name.

PR #6454, opened 2021-06-30
Don't save build cache from every bucket

actgardner:
RIght now we save the build cache from every bucket, but we really only need one per build. Saving the cache takes about 2 minutes, so removing it from the slowest bucket reduces end-to-end latency.


actgardner:
Already merged in #6455

PR #6455, opened 2021-06-30
Remove unused internal packages and run all internal tests

actgardner:
Also includes https://github.com/pachyderm/pachyderm/pull/6454

Looking at all the tests in the `internal` subpackages, we only seemed to be running tests for 13 out of 22. This reworks how we bucket tests to run all the internal subtests in a single bucket, which has a comparable runtime (14m) to the PPS buckets. The internal bucket replaces the object bucket, and runs some tests that were formerly covered by MISC and PFS/

The `bloom` package was failing because it depended on the size of the hash used in pachhash, but on further inspection it wasn't used anywhere in the codebase. `discovery` and `localcache` have also been removed because they're ununsed.

Long-term we should work on bucketing the entire set of tests for the codebase - our current bucketing code only works for a single package.

Affected packages and the number of missed tests:
```
  # We were running tests for these
   1 "github.com/pachyderm/pachyderm/v2/src/internal/migrations"
   1 "github.com/pachyderm/pachyderm/v2/src/internal/storage/track"
   2 "github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset/index"
   2 "github.com/pachyderm/pachyderm/v2/src/internal/obj"
   3 "github.com/pachyderm/pachyderm/v2/src/internal/storage/chunk"
   3 "github.com/pachyderm/pachyderm/v2/src/internal/cert"
   3 "github.com/pachyderm/pachyderm/v2/src/internal/deploy/cmds"
   3 "github.com/pachyderm/pachyderm/v2/src/internal/grpcutil"
   5 "github.com/pachyderm/pachyderm/v2/src/internal/obj/integrationtests"
   6 "github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset"
   7 "github.com/pachyderm/pachyderm/v2/src/internal/work"
   11 "github.com/pachyderm/pachyderm/v2/src/internal/collection"
   
   # We weren't running tests for these, but they're still in use
   1 "github.com/pachyderm/pachyderm/v2/src/internal/uuid"
   2 "github.com/pachyderm/pachyderm/v2/src/internal/ancestry"
   2 "github.com/pachyderm/pachyderm/v2/src/internal/pacherr"
   4 "github.com/pachyderm/pachyderm/v2/src/internal/deploy/assets"
   4 "github.com/pachyderm/pachyderm/v2/src/internal/errors/testing"
   6 "github.com/pachyderm/pachyderm/v2/src/internal/serde"
  15 "github.com/pachyderm/pachyderm/v2/src/internal/backoff"
  
  # Removed because they're unused
   2 "github.com/pachyderm/pachyderm/v2/src/internal/discovery"
   4 "github.com/pachyderm/pachyderm/v2/src/internal/localcache"
   6 "github.com/pachyderm/pachyderm/v2/src/internal/bloom"
```


etc/testing/circle/run_tests.sh
```diff
@@ -83,13 +83,16 @@ case "${BUCKET}" in
         make test-tls
     fi
     ;;
+  INTERNAL)
+    go install -v ./src/testing/match
```
chainlink, line 87: Maybe wrap this in a make task so it can be run from make test? (Not sure how common running tests locally is)

chainlink (approved):
One small thing, but otherwise looks good!

.circleci/config.yml
```diff
@@ -58,14 +58,18 @@ jobs:
...
       - run: etc/testing/circle/run_tests.sh 
-      - save_cache:
-         key: pach-go-mod-cache-v2-{{ checksum "go.sum" }}
-         paths:
-         - /home/circleci/.go_workspace/pkg/mod
-      - save_cache:
-         key: pach-go-build-cache-v1-{{ .Branch }}-{{ checksum "current_week" }}-{{ .BuildNum }}
-         paths:
-         - /home/circleci/.gocache
+      - when:
```
chainlink, line 61: pick one bucket to save cache from?

jrockway (approved):
Good cleanup.  I wonder if we can maybe print somewhere what tests have been selected (`go list -json ./src/example/...  | jq .TestGoFiles` before running `go test ./src/example/...`) and then see which _test files in the repo don't show up there?  Out of scope for this PR, but it would be good to have some way of avoiding this problem.

Maybe something else to consider... maybe have a TestMain function in every test that requires pachd running in k8s, and skip the whole test if some environment variable is unset.  That way "go test ./..." will do something useful, even if it doesn't run every test we have.

etc/testing/circle/run_tests.sh
actgardner: I don't think that running the entire test suite locally is very common, and I'm trying to move away from adding Make targets in favour of straight `go test` targets. If people feel strongly we can re-add the Make target but I think it just abstracts and causes these kinds of issues.

.circleci/config.yml
actgardner: Not sure I follow - this only does these steps when the bucket is MISC. The syntax is wild.

.circleci/config.yml
```diff
@@ -58,14 +58,18 @@ jobs:
...
+          condition: 
+            equal: [MISC, <<parameters.bucket>> ]
```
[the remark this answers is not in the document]
actgardner: irony

actgardner:
> Good cleanup. I wonder if we can maybe print somewhere what tests have been selected (go list -json ./src/example/... | jq .TestGoFiles before running go test ./src/example/...) and then see which _test files in the repo don't show up there? Out of scope for this PR, but it would be good to have some way of avoiding this problem.

I think the dream is being able to partition the tests in `./...` such that we run a single `go test ./...` in each bucket and it runs a subset of all the tests for the codebase, such that all the buckets have roughly equal runtime and cover the entire codebase. The only solution I've seen has been to add a hook into every individual test that condiitonally skips, which feels very onerous.

etc/testing/circle/run_tests.sh
chainlink: Makes sense, mostly proposing for consistency, but agree with the move away from make targets. I'd love to be able to run `go test ./...` on the whole repo

PR #6458, opened 2021-06-30
Fix load testing with auth enabled

brycemcanally:
This PR will include the changes necessary to make the load testing work with auth enabled.


acohen4 (approved):
LGTM

PR #6463, opened 2021-07-01
roles: add GET_PACHD_LOGS permission to debugger role

jrockway:
A debug dump returns the logs anyway, so this doesn't escalate the capabilities the user has.


acohen4 (approved):
LGTM

PR #6464, opened 2021-07-02
Fixing TestStandby tests

acohen4:


src/server/pachyderm_test.go
```diff
@@ -3114,7 +3112,7 @@ func TestStandby(t *testing.T) {
 			require.NoError(t, err)
 		}
 
-		require.NoErrorWithinTRetry(t, time.Second*30, func() error {
+		require.NoErrorWithinTRetry(t, time.Second*60, func() error {
```
acohen4, line 3115: it seems that  getting all pipelines up in standby mode is taking longer than it did in the past. Should we investigate this as a performance regression, or does this seem acceptable? @brycemcanally @Tryneus

src/server/pachyderm_test.go
```diff
@@ -3079,8 +3079,6 @@ func TestStopPipeline(t *testing.T) {
 }
 
 func TestStandby(t *testing.T) {
```
Tryneus, line 3081: It seems like `Standby` is now just `Autoscaling` - should probably rename the test as well.

src/server/pachyderm_test.go
```diff
@@ -3183,11 +3184,22 @@ func TestStandby(t *testing.T) {
...
 		commitInfos, err := c.WaitCommitSetAll(commitInfo.Commit.ID)
 		require.NoError(t, err)
-		require.Equal(t, 2, len(commitInfos))
+		nonAliasCommits := 0
+		for _, v := range commitInfos {
+			if v.Origin.Kind != pfs.OriginKind_ALIAS {
+				nonAliasCommits++
+			}
+		}
+		require.Equal(t, 3, nonAliasCommits) // we got an extra meta commit here
```
Tryneus, line 3193: The comment here is probably unnecessary unless you want to just list all three commits - otherwise this is just an explanation of the change from the previous version which is no longer present in the code to give the necessary context.

src/server/pachyderm_test.go
Tryneus: This seems like something we should file an issue for and investigate - I think we're going to be putting a bigger emphasis on tracking down performance regressions in the next few weeks.

Tryneus (approved):
A couple nits, but otherwise LGTM.

src/server/pachyderm_test.go
acohen4: good point. Will remove this! I also just wanted you to see this since you're aware of any more "syntactic sugar" like changes that may come in to deal with explore the different commit types

PR #6469, opened 2021-07-02
Run PFS, Collections, and storage tests on PG Bouncer

brendoncarroll:
The goal of this P.R. is to work out incompatibilities between our driver and pg bouncer, and get the worker code paths (chunk uploading) working.  There is a minimal 2 container test environment to iterate on.

So far there the errors encountered are:
- Increase "query cancelled due to user request"  It was possible to hit this before, but there was an increase after adding pg bouncer.
- Unnamed prepared statement error, caused by a known bug in pq, and fixed with the `binary_parameters=yes` workaround


chainlink:
For "query cancelled due to user request" I think there might be contention with pach trying to connect to PG while PG Bouncer does as well.

src/internal/deploy/cmds/cmds.go
```diff
@@ -443,7 +442,7 @@ func standardDeployCmds() []*cobra.Command {
 
 				// Set the postgres and etcd nodeports explicitly for developers
 				if !enterpriseServer {
```
chainlink, line 444: This won't work anymore as pach isn't install through pachctl deploy

etc/helm/pachyderm/templates/pgbouncer/service.yaml
```diff
@@ -14,10 +14,11 @@ metadata:
...
   - name: client-port
+    {{- if eq .Values.pgbouncer.service.type "NodePort" }}
     nodePort: 32229
+    {{- end }}
     port: 5432
   selector:
     app: pg-bouncer
     suite: pachyderm
-  type: NodePort
-{{- end }}
```
chainlink: Need to keep this end statement

etc/helm/examples/enterprise-dev.yaml
```diff
@@ -9,3 +9,6 @@ enterpriseServer:
     tag: local
 pachd:
   enabled: false
+pgbouncer:
+  enabled: true
```
chainlink, line 13: This should be the default

chainlink: I believe this should be false now, as we're not using PG bouncer for enterprise?

chainlink (approved):
I think we're g2g on the helm side, tested locally and install works. Should we get a review on the pachd side? I'm a little concerned by the random "cancelled by user" errors we're seeing sporadically in the tests

etc/helm/examples/enterprise-dev.yaml
brendoncarroll: Yeah, we have to pass credentials for PG_BOUNCER though since the enterprise server is now equivalent to a full pachd.  They can also point to postgres, but they do have to be set.

src/server/pps/server/worker_rc.go
```diff
@@ -118,6 +118,12 @@ func (a *apiServer) workerPodSpec(options *workerOptions, pipelineInfo *pps.Pipe
 	}, {
 		Name:  client.PPSPipelineNameEnv,
 		Value: pipelineInfo.Pipeline.Name,
+	}, {
+		Name:  "POSTGRES_PORT",
```
chainlink, line 122: Maybe we should add a note as to why we're zero'ing these out (to override the kube service name env vars)

src/internal/collection/postgres_collection.go
```diff
@@ -236,12 +236,14 @@ func (c *postgresCollection) list(
 ) error {
 	query := fmt.Sprintf("select key, createdat, updatedat, proto from collections.%s", c.table)
 
-	params := map[string]interface{}{}
+	var args []interface{}
 	if len(withFields) > 0 {
 		fields := []string{}
+		i := 1
```
PFedak, line 242: good old one indexing :)

brendoncarroll:
This P.R. removes the deployment tests in `./src/internal/obj/integrationtest`.  It also skips the `RunCron` tests.  Those tests have passed locally, so I am confident it is not a config issue.  They are just too flaky to keep enabled.  @PFedak has a possible fix for them.  They appear to be misusing transactions in a way that works some of the time, but works only rarely with pg_bouncer.

src/internal/serviceenv/service_env.go
```diff
@@ -296,21 +304,46 @@ func (env *NonblockingServiceEnv) initDBClient() error {
...
 			return err
 		}
+		env.directDBClient = db
+		if err := prometheus.Register(sqlstats.NewStatsCollector("pg_bouncer", db.DB)); err != nil {
+			// This is not a retryable error.  Rather it will always happen for the
+			// second (and subsequent) service environment because of a naming conflict.
+			// If you see this message in production, it's a bug.  In tests, it's OK.
+			log.WithError(err).Warn("problem registering database statistics collector")
+		}
+		return db.Ping()
```
PFedak, line 314: does this prevent/reduce the pachd crashes at startup while postgres comes up? Or is etcd the limiting factor

PFedak (approved):
LGTM

src/internal/serviceenv/service_env.go
brendoncarroll: I didn't even think about that actually.  I added it to make sure the bad connection errors weren't happening during startup.

PR #6474, opened 2021-07-06
Ruhl/add dash client secret

robert-uhl:


armaanv (approved):
LGTM

etc/helm/pachyderm/templates/_helpers.tpl
```diff
@@ -19,3 +19,11 @@ LOCAL
 {{ fail "pachd.storage.backend required when no matching deploy target found" }}
 {{- end -}}
 {{- end -}}
+
+{{- define "pachyderm.dashSecret" -}}
```
chainlink, line 23: can simplify with `default`

chainlink (changes requested):
I think we can eliminate the need for the template by using `default` instead

etc/helm/pachyderm/templates/_helpers.tpl
robert-uhl: And done.

chainlink (approved):

PR #6479, opened 2021-07-06
Expose size as an upper bound in list operations

brycemcanally:
The primary goal of this PR is to expose the repo and commit size as an upper bound in list operations. Computing the actual size of a repo or commit can be an expensive operation and we don't want to incur this cost in a list operation which can potentially return a lot of repos or commits. The sizes in list operations are now an upper bound, which is much cheaper to collect. The general structure of the protos is a field called `SizeBytesUpperBound` which is filled out in list operations and a field called `SizeBytes` in the `Details` field with is filled out by inspect operations. An example of what this would look like in pachctl is:
```
bryce@thinkpad:~/go/src/github.com/pachyderm/pachyderm$ pachctl create repo test
bryce@thinkpad:~/go/src/github.com/pachyderm/pachyderm$ echo "data" | pachctl put file test@master:/dir/file1
stdin 5.00 b / 5.00 b [===================================================================================================================================================================================================================] 0s 0.00 b/s
bryce@thinkpad:~/go/src/github.com/pachyderm/pachyderm$ echo "data" | pachctl put file test@master:/dir/file2
stdin 5.00 b / 5.00 b [===================================================================================================================================================================================================================] 0s 0.00 b/s
bryce@thinkpad:~/go/src/github.com/pachyderm/pachyderm$ pachctl delete file -r test@master:/dir
bryce@thinkpad:~/go/src/github.com/pachyderm/pachyderm$ pachctl list repo
NAME CREATED        SIZE (MASTER) DESCRIPTION 
test 19 seconds ago <= 10B                    
bryce@thinkpad:~/go/src/github.com/pachyderm/pachyderm$ pachctl list commit test
REPO BRANCH COMMIT                           FINISHED       SIZE   ORIGIN DESCRIPTION
test master 38254d7af72f448f9f6a75016f0b1894 7 seconds ago  <= 10B USER    
test master 2e024fcd2b914f7b8d7f439a8ca8e4b7 16 seconds ago <= 10B USER    
test master 199ab74930704afeb37a111149325b91 18 seconds ago <= 5B  USER    
bryce@thinkpad:~/go/src/github.com/pachyderm/pachyderm$ pachctl inspect repo test
Name: test
Created: 28 seconds ago
Size of HEAD on master: 0B
bryce@thinkpad:~/go/src/github.com/pachyderm/pachyderm$ pachctl inspect commit test@master
Commit: test@38254d7af72f448f9f6a75016f0b1894
Original Branch: master
Parent: 2e024fcd2b914f7b8d7f439a8ca8e4b7
Started: 24 seconds ago
Finished: 24 seconds ago
Size: 0B
```

This PR also:
- Removes the `Details` field from `FileInfo`. The file sizes and hashes will always be available. The sizes are readily available and the hashes will be cheap to compute in the majority of cases. There is a followup change needed to make this true for open commits (they currently have different performance characteristics than finished commits).
- Adds a recursive flag to pachctl delete file for indicating a directory deletion.
- Changes the PFS protos to use int64 rather than uint64.


src/internal/metrics/metrics.go
```diff
@@ -376,11 +376,7 @@ func (r *Reporter) internalMetrics(metrics *Metrics) {
 		}
 		var sz, mbranch uint64 = 0, 0
 		for _, ri := range repoInfos {
-			if (sz + ri.SizeBytes) < sz {
-				sz = 0xFFFFFFFFFFFFFFFF
-			} else {
-				sz += ri.SizeBytes
-			}
+			sz += uint64(ri.SizeBytesUpperBound)
```
jdoliner, line 379: Do we want to move this off of uint64 as well?

jdoliner (approved):
One suggestion about switching off unit64 in the metrics protos. Other than that lgtm.

src/internal/metrics/metrics.go
brycemcanally: Yeah, there are several other places where we are using uint64 that I would like switched over, but that would have to be in a followup or else this PR is going to get large. I focused on getting the PFS protos moved off of uint64 first.

PR #6481, opened 2021-07-06
Monitor more things

jrockway:
This adds many more Prometheus metrics.  As the server runs, you can really get a feel for what it's doing.

"Many more" includes:

gRPC client monitoring keeps an eye on all outgoing RPCs, including self-RPCs and etcd.  `grpc_client_handled_total`, `grpc_client_msg_received_total`, `grpc_msg_sent_total`, and `grpc_client_started_total`.  Each is split by the service ("pfs_v2.API"), method ("InspectCommit"), and response code ("OK").

gRPC server monitoring does the same for incoming RPCs.  To some extent, we already have that for some methods that we're suspicious about; collecting a histogram of response times.  (I did update the buckets on those metrics to capture more "fast" calls, and removed a redundant "total number of seconds" metric for each; a histogram `foo_seconds` will automatically create a `foo_seconds_sum`, and that's what that is.)  I didn't enable the generic histograms (they are expensive, and we have them where we want them), but we now have `grpc_server_handled_total`, `grpc_server_msg_received_total`, `grpc_server_msg_sent_total`, `grpc_server_started_total`.  They work like the client ones; split by service, method, and response code.  They also match the client ones (client_msg_sent = server_msg_recv for internal-only methods).

I added HTTP client monitoring for everything I could think of that makes HTTP requests -- Kubernetes API server, S3, Google Cloud Storage, Azure Storage, Minio, etc.  This will show response status ("200") `http_client_requests_total`, the number of requests in-flight to a certain service (good for debugging "I think we're overloading the k8s API") `http_client_in_flight_requests`, and a histogram of the time spent waiting for the response. `http_client_request_duration_seconds`.

I added a ton of metrics around storage in general.  Counts of the high-level operation are recorded as `pachyderm_object_storage_operation_count_total`, grouped by storage provider ("amazon") and the operation ("delete").  I also injected an io.Reader/io.Writer to the actual file IO, so that we can record the total bytes we sent/received to each provider (`pachyderm_object_storage_written_bytes_total`, `pachyderm_object_storage_read_bytes_total`).  This is mostly interesting for the derivative; how much of your network card (or disk IO for "Local" / "DiskCache" storage types) are you using? The wrapper has the side-effect of removing ReadAt/WriteTo from the underlying io.Reader/io.Writer objects, theoretically reducing performance in some cases by preventing the client from seeking or using copy-less copies. I checked what concrete values of io.Reader/io.Writer we use, and it's a mix of io.Pipe and bytes.Buffer, mostly.  So there would be some benefit in writing a wrapper that can expose  the underlying methods to the storage client correctly, but I haven't done that yet.  (ok, pipes can't seek, but bytes.Buffer can; and the S3 client appears to use this ability behind the scenes).

I also added metrics around the Cache client (cache hits `pachyderm_object_storage_cache_hits_total`, cache misses `_misses_total`, object evicted `_evictions_total`), and the Limited client (time spent blocking on the semaphore, by operation type (get/put), `pachyderm_object_storage_limited_seconds_total`).

Still to do are adding byte counters to put-file/get-file; that would let us observe bottlenecks during file upload (but it's also easy to get that information from `sar` on the machine that is running the upload).

I also fixed the TracingClient; at some point Reader was replaced with Get, it seems, but the tracing client wasn't updated.  Now it's updated.  Finally, I added a gitignore for build artifacts like `go build ./src/server/cmd/pachd`, which is how I tested all this ;)


src/internal/grpcutil/server.go
```diff
@@ -44,6 +46,8 @@ func NewServer(ctx context.Context, publicPortTLSAllowed bool, options ...grpc.S
 			MinTime:             5 * time.Second,
 			PermitWithoutStream: true,
 		}),
+		grpc.ChainUnaryInterceptor(grpc_prometheus.UnaryServerInterceptor),
+		grpc.ChainStreamInterceptor(grpc_prometheus.StreamServerInterceptor),
```
jrockway, line 50: Multiple `grpc.Chain*Interceptor` options passed to the same NewServer call compose, so it's safe for calling code to add interceptors in addition to these.

src/internal/promutil/promutil.go
```diff
@@ -0,0 +1,79 @@
...
+		Name: "http_client_in_flight_requests",
+		Help: "A gauge of in-flight requests being made against an HTTP API, by client.",
+	}, []string{"client"})
+
+	requestCountMetric = promauto.NewCounterVec(prometheus.CounterOpts{
+		Name: "http_client_requests_total",
+		Help: "A summary of requests made against an HTTP API, by client, status code, and request method.",
+	}, []string{"client", "code", "method"})
+
+	requestTimeMetric = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
chainlink, line 24: We might want to weight these buckets down to smaller increments, or eliminate them entirely (At least, I sure hope we're not seeing 60s+ requests very often)

jrockway: I think we have k8s watches that last for ~30 minutes.

src/internal/obj/limited_client.go
```diff
@@ -4,10 +4,22 @@ import (
...
 	"math"
+	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"golang.org/x/sync/semaphore"
 )
 
+var (
+	blockedSecondsMetric = promauto.NewCounterVec(prometheus.CounterOpts{
```
brycemcanally, line 15: This seems like something that would make more sense as a histogram. Also, it might be good to ensure we have a consistent naming scheme for the metrics. The `object_storage` subsystem seems fine for what we will be collecting after this PR is merged, but we probably should have the subsystem either be or be prefixed by the top level subsystem (i.e. pfs / pps).

jrockway: Done.  We need a wide range here, disk is fast, the network is slow, so we have everything from ~10us to 60 seconds.  In testing, ends up looking something like this:

```
pachyderm_pfs_object_storage_limited_seconds_bucket{op="get",le="1e-05"} 8387
pachyderm_pfs_object_storage_limited_seconds_bucket{op="get",le="5e-05"} 8940
pachyderm_pfs_object_storage_limited_seconds_bucket{op="get",le="0.0001"} 8946
pachyderm_pfs_object_storage_limited_seconds_bucket{op="get",le="0.0005"} 8982
pachyderm_pfs_object_storage_limited_seconds_bucket{op="get",le="0.001"} 9008
pachyderm_pfs_object_storage_limited_seconds_bucket{op="get",le="0.005"} 9009
```

I looked to see if we could easily change the code to distinguish between "waited 10us" and "it just took 10us to get to that line of code", but the semaphore API doesn't make that particularly easy.  So I'm guessing that the smallest bucket pretty much equals "didn't block", the second smallest bucket being some noise, and the rest being meaningful.  Probably totally fine.

jrockway: Oh, also added pfs_ to the object storage-related metrics, and audited other uses of subsystem for consistency.  (The subsystems we have now are postgres, pfs, pps, pachd, and worker.  There might be some value in calling postgres pachd_postgres and worker_postgres, but it's not a super easy change to make.)

jrockway:
I added Dex metrics as well.  Please take another look.

src/internal/obj/limited_client.go
```diff
@@ -39,17 +52,21 @@ func NewLimitedClient(client Client, maxReaders, maxWriters int) Client {
 }
 
 func (loc *limitedClient) Put(ctx context.Context, name string, r io.Reader) error {
+	t := time.Now()
```
brendoncarroll, line 55: This might be more of a general monitoring question, but it's applicable here.  Is it better (or is there a cost) to marking start and end events?  If we run into a problem where everything gets hung up here, the start end delta would make it more obvious.

brendoncarroll (approved):
LGTM

src/internal/obj/limited_client.go
jrockway: You're right that we should have _started and _finished for this, because waiting until it's done to emit data means that we don't know how many waits are in-flight (or indefinitely blocked).  I'll add those.

I worry a little about the cost around things that aren't IO, but this wraps IO so it doesn't trigger my "obviously a bad idea" heuristic.  And, we have cross-fleet continuous profiling with Cloud Profiler, so if we see that we spend a lot of time in `(*prometheus.Counter).Inc()` then we can just remove or rework the slow metrics.

jrockway: Done.  I didn't add a _finished because the histogram computes that for us.  The `HELP` message explains what that's called, so someone browsing `/metrics` knows that `pachyderm_pfs_object_storage_limited_seconds_count` is the end of `pachyderm_pfs_object_storage_block_started_total`.  I also tested this to be sure it was correct.

PR #6482, opened 2021-07-07
local.sh: update ports, add correct RBAC rules, allow "kind" 

jrockway:
I remembered that this existed today, and it's been very helpful.  I tweaked a few things to make it potentially work for more people, mostly me.  (The base64 command burned me again; MacOS's version wants different flags than Linux's -- so I just did the transform in jq instead.)

I also added the RBAC rules; the default service account in kube-system didn't have the necessary permissions.  I extracted a minimal manifest out of the currently-checked-in helm chart and that works for me.  Putting a heredoc in the script is distasteful... we can replace this with something that convinces helm to apply the correct manifests, but that is a bigger change and will do it later ;)

### DIFF
--- a/etc/testing/local.sh
+++ b/etc/testing/local.sh
@@ -1,6 +1,19 @@
 #!/bin/bash
 
-set -euo pipefail
+set -xeuo pipefail
+
+# Run pachd locally like:
+# CGO_ENABLED=0 go build ./src/server/cmd/pachd
+# ./etc/testing/local.sh ./pachd
+#
+# If you use kind instead of minikube, you will have to override some configuration:
+#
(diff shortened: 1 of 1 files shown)


actgardner (approved):
This seems fine. The service account wasn't part of this script because the way that I usually use this script is to deploy pachyderm and then spin down the pachd deployment. If we had an option to the helm chart for pachd replicas we could probably just set it to 0 during the deploy and get postgres, etcd, SA, etc. without a pachd deployment.

I haven't used the local setup recently or spent a lot of time on it, I don't expect it will Just Work since the 2.0 changes that require workers to connect back to the pachd pod to listen for new work (the workaround for postgres watches).

jrockway:
Good point on the connect-back-to-pachd stuff.  I agree that this method of testing PPS is sub-optimal, but it's good for everything else.

PR #6484, opened 2021-07-07
Updating S3 Gateway to work with V2

acohen4:
Changes:

1) Now that it's possible for multiple Pipeline Jobs to exist at the same time, we cannot represent each S3G endpoint as its own Server with its own port. Instead we now keep a mapping of mux router objects that are used by the server to serve the corresponding s3 endpoints

2) S3_ENDPOINT usr container environment variable now resolves to a Global-ID compatible Job ID (pipeline@id).
- ex. http://s3-pipeline-id:1600

### DIFF
--- a/src/internal/ppsutil/util.go
+++ b/src/internal/ppsutil/util.go
@@ -17,6 +17,8 @@ import (
 	"strings"
 	"time"
 
+	"crypto/md5"
+
 	"github.com/gogo/protobuf/proto"
 	"github.com/gogo/protobuf/types"
 	"github.com/jmoiron/sqlx"
@@ -433,7 +435,10 @@ func ContainsS3Inputs(in *pps.Input) bool {
 // sidecar server) and the worker (which passes the endpoint to the user code)
 // need to know it.
 func SidecarS3GatewayService(pipeline, commitSetId string) string {
-	return fmt.Sprintf("s3-%v-%v", strings.ToLower(pipeline), commitSetId)
+	hash := md5.New()
+	hash.Write([]byte(pipeline))
+	hash.Write([]byte(commitSetId))
+	return "s3-" + pfs.EncodeHash(hash.Sum(nil))
 }
 
 // ErrorState returns true if s is an error state for a pipeline, that is, a
--- a/src/internal/ppsutil/util.go
+++ b/src/internal/ppsutil/util.go
@@ -433,7 +433,7 @@ func ContainsS3Inputs(in *pps.Input) bool {
 // sidecar server) and the worker (which passes the endpoint to the user code)
 // need to know it.
 func SidecarS3GatewayService(jobID string) string {
-	return "s3-" + jobID
+	return "s3-" + strings.Replace(strings.ToLower(jobID), "@", "-", 1)
 }
 
 // ErrorState returns true if s is an error state for a pipeline, that is, a
--- a/src/internal/ppsutil/util.go
+++ b/src/internal/ppsutil/util.go
@@ -432,8 +432,8 @@ func ContainsS3Inputs(in *pps.Input) bool {
 // is in ppsutil because both PPS (which creates the service, in the s3 gateway
 // sidecar server) and the worker (which passes the endpoint to the user code)
 // need to know it.
-func SidecarS3GatewayService(jobID string) string {
-	return "s3-" + strings.Replace(strings.ToLower(jobID), "@", "-", 1)
+func SidecarS3GatewayService(pipeline, commitSetId string) string {
+	return fmt.Sprintf("s3-%v-%v", strings.ToLower(pipeline), commitSetId)
 }
 
 // ErrorState returns true if s is an error state for a pipeline, that is, a
--- a/src/server/pfs/s3/s3.go
+++ b/src/server/pfs/s3/s3.go
@@ -113,15 +113,45 @@ func Router(driver Driver, clientFactory ClientFactory) *mux.Router {
 	return s3Server.Router()
 }
 
-// Server runs an HTTP server with an S3-like API for PFS. This allows you to
+// S3Server wraps an HTTP server with an S3-like API for PFS. This allows you to
 // use s3 clients to access PFS contents.
 
-// Modifications to `routerMap` allow dynamic changes for the mapping between s3 endpoints and API configurations
-func Server(port uint16, defaultRouter *mux.Router, routerMap map[string]*mux.Router, routersLock *sync.Mutex) *http.Server {
+// In addition to providing the http server itself, S3Server exposes methods
+// to configure handlers (mux.Routers) corresponding to different request URI hostnames.
+// This way one http Server can respond differently and accordingly to each specific job.
+type S3Server struct {
+	HttpServer  *http.Server
+	routerMap   map[string]*mux.Router
+	routersLock sync.RWMutex
+}
+
+func (s *S3Server) ContainsRouter(k string) bool {
+	s.routersLock.RLock()
+	defer s.routersLock.RUnlock()
+	_, ok := s.routerMap[k]
+	return ok
+}
+
+func (s *S3Server) AddRouter(k string, r *mux.Router) {
+	s.routersLock.Lock()
+	defer s.routersLock.Unlock()
+	s.routerMap[k] = r
+}
+
+func (s *S3Server) RemoveRouter(k string) {
+	s.routersLock.Lock()
+	defer s.routersLock.Unlock()
+	delete(s.routerMap, k)
+}
+
+// Server runs an HTTP server with an S3-like API for PFS. This allows you to
+// use s3 clients to access PFS contents.
+func Server(port uint16, defaultRouter *mux.Router) *S3Server {
 	logger := logrus.WithFields(logrus.Fields{
 		"source": "s3gateway",
 	})
-	server := &http.Server{
+	s3Server := S3Server{routerMap: make(map[string]*mux.Router)}
+	s3Server.HttpServer = &http.Server{
 		Addr:         fmt.Sprintf(":%d", port),
 		ReadTimeout:  requestTimeout,
 		WriteTimeout: requestTimeout,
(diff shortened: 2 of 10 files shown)


src/server/pps/server/s3g_sidecar.go
```diff
@@ -193,69 +207,16 @@ func (s *s3InstanceCreatingJobHandler) OnCreate(ctx context.Context, jobInfo *pp
 	// more than one job in s.servers). When parallel jobs are implemented, the
 	// servers in s.servers won't actually serve anymore, and instead parent
 	// server will forward requests based on the request hostname
-	port := s.s.apiServer.env.Config().S3GatewayPort
```
actgardner, line 196: ^ The TODO above can probably go away now?

actgardner (approved):

src/server/pps/server/s3g_sidecar_test.go
```diff
@@ -666,14 +664,14 @@ func TestS3SkippedDatums(t *testing.T) {
 		pipelineCommit := client.NewCommit(pipeline, "master", "")
 		// Add files to 'repo'. Old files in 'repo' should be reprocessed in every
 		// job, changing the 'background' field in the output
-		for i := 0; i < 10; i++ {
+		for i := 1; i < 7; i++ {
```
acohen4, line 667: Going to file an issue here. It looks like by the later iterations it take about 50 seconds for a job to run. This seems like a lot to me. Maybe there's some big latency each time we query the S3 endpoint, given that there are 3 of these that get run per Datum, and there are approx 10 datums at the end?

acohen4, line 667: I also set `i` to start at 1 to make the counting a little more readable now that we start with 1 ALIAS commit

src/server/pps/server/s3g_sidecar.go
acohen4: ah yes!

src/internal/ppsutil/util.go
```diff
@@ -431,7 +431,7 @@ func ContainsS3Inputs(in *pps.Input) bool {
 // sidecar server) and the worker (which passes the endpoint to the user code)
 // need to know it.
 func SidecarS3GatewayService(jobID string) string {
-	return "s3-" + jobID
+	return "s3-" + strings.Replace(strings.ToLower(jobID), "@", "-", 1)
```
PFedak, line 434: feels kind of unnecessary to build a string then modify it. Also jobID is kind of ambiguous now, I would prefer either passing in a job proto or two strings

src/server/pps/server/s3g_sidecar.go
```diff
@@ -309,18 +266,17 @@ func (s *k8sServiceCreatingJobHandler) OnCreate(ctx context.Context, jobInfo *pp
 		return nil
 	})
 	if err != nil {
-		logrus.Errorf("could not create service for job %q: %v", jobInfo.Job.ID, err)
+		logrus.Errorf("could not create service for job %q: %v", jobInfo.Job.String(), err)
```
PFedak, line 269: you don't actually have to call String() here, `Printf` etc will do that automatically

src/server/pfs/s3/s3.go
```diff
@@ -109,20 +110,46 @@ func Server(port uint16, driver Driver, clientFactory ClientFactory) (*http.Serv
...
 		WriteTimeout: requestTimeout,
 		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			// Log that a request was made
 			logger.Infof("http request: %s %s", r.Method, r.RequestURI)
-			router.ServeHTTP(w, r)
+			if strings.HasPrefix(r.Host, "s3-") {
+				routersLock.Lock()
+				defer routersLock.Unlock()
+				host := r.Host
+				if sep := strings.Index(r.Host, "."); sep != -1 {
```
PFedak, line 135: is there a specific case this is handling?

src/server/pfs/s3/s3.go
```diff
@@ -109,20 +110,46 @@ func Server(port uint16, driver Driver, clientFactory ClientFactory) (*http.Serv
...
 	s3Server.Multipart = c
-	router := s3Server.Router()
+	return s3Server.Router()
+}
+
+// Server runs an HTTP server with an S3-like API for PFS. This allows you to
+// use s3 clients to access PFS contents.
 
+// Modifications to `routerMap` allow dynamic changes for the mapping between s3 endpoints and API configurations
+func Server(port uint16, defaultRouter *mux.Router, routerMap map[string]*mux.Router, routersLock *sync.Mutex) *http.Server {
```
PFedak, line 120: Is there a particular reason to pass the map and mutex in, rather than having the Server own them itself

src/server/pfs/s3/s3.go
```diff
@@ -109,20 +110,46 @@ func Server(port uint16, driver Driver, clientFactory ClientFactory) (*http.Serv
...
 	server := &http.Server{
 		Addr:         fmt.Sprintf(":%d", port),
 		ReadTimeout:  requestTimeout,
 		WriteTimeout: requestTimeout,
 		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			// Log that a request was made
 			logger.Infof("http request: %s %s", r.Method, r.RequestURI)
-			router.ServeHTTP(w, r)
+			if strings.HasPrefix(r.Host, "s3-") {
+				routersLock.Lock()
```
PFedak, line 132: a RWMutex could make sense here, but I doubt it makes much difference

src/server/pfs/s3/s3.go
```diff
@@ -109,20 +110,46 @@ func Server(port uint16, driver Driver, clientFactory ClientFactory) (*http.Serv
...
+				router, ok := routerMap[host]
+				if ok {
+					router.ServeHTTP(w, r)
+				} else {
+					w.WriteHeader(http.StatusInternalServerError)
+				}
+			} else if defaultRouter != nil {
+				defaultRouter.ServeHTTP(w, r)
+			} else {
+				w.WriteHeader(http.StatusInternalServerError)
```
PFedak, line 147: These feel more like 400 errors, although I guess there's also the possibility that the worker asks for it before the sidecar has set up the router

src/server/pps/server/s3g_sidecar.go
```diff
@@ -85,6 +91,16 @@ func (a *apiServer) ServeSidecarS3G() {
 		return // break early (nothing to serve via S3 gateway)
 	}
 
+	go func() {
+		for i := 0; i < 2; i++ { // If too many errors, the worker will fail the job
+			err := s.server.ListenAndServe()
+			if err == nil || errors.Is(err, http.ErrServerClosed) {
+				break // server was shutdown/closed
+			}
+			logrus.Errorf("error serving sidecar s3 gateway: %v; strike %d/3", err, i+1)
```
PFedak, line 100: I know it was already written this way, but doesn't it only get two strikes?

PFedak (changes requested):
Mostly LGTM, a couple of small changes I'd like to see, but overall this is awesome!

src/server/pps/server/s3g_sidecar.go
acohen4: that's right! good catch

src/server/pps/server/s3g_sidecar.go
acohen4: yep that's true. Will update the comment

src/internal/ppsutil/util.go
acohen4: that makes sense. Is the GUID in the ID the commit set ID?

src/server/pfs/s3/s3.go
acohen4: the reason is that ServeSidecarS3G is modifying that map concurrently to add/remove endpoint router objects. We could definitely push another layer of abstraction and have another Server wrapper object that manages this. 
What do you think?

src/server/pfs/s3/s3.go
acohen4: i agree with both statements :)

i'll switch to it since it seems logical

src/server/pfs/s3/s3.go
acohen4: my thinking here was to the user, who's calling our S3 gateway this is probably more of an internal error of our system not being totally set up yet. 

Specifically, if there's an S3 DNS that hits us but we don't have a backend to support it up yet, that's our bad which makes me think it should be an internal error.

PFedak: Ah okay, that seems reasonable

src/server/pfs/s3/s3.go
acohen4: yes and I'm not exactly sure when this case comes up, but sometimes we get routed to a host with the k8s namespace parent domain: ex. http://s3-gateway.default:1600 instead of http://s3-gateway:1600

PFedak: ahh, I see

src/server/pfs/s3/s3.go
```diff
@@ -113,29 +113,59 @@ func Router(driver Driver, clientFactory ClientFactory) *mux.Router {
...
+// S3Server wraps an HTTP server with an S3-like API for PFS. This allows you to
 // use s3 clients to access PFS contents.
 
-// Modifications to `routerMap` allow dynamic changes for the mapping between s3 endpoints and API configurations
-func Server(port uint16, defaultRouter *mux.Router, routerMap map[string]*mux.Router, routersLock *sync.Mutex) *http.Server {
+// In addition to providing the http server itself, S3Server exposes methods
+// to configure handlers (mux.Routers) corresponding to different request URI hostnames.
+// This way one http Server can respond differently and accordingly to each specific job.
+type S3Server struct {
+	HttpServer  *http.Server
```
PFedak, line 123: making this an embedded struct would make some of the http-related call sites simpler

PFedak (approved):
LGTM, this is awesome! Just one potential change, totally optional

src/server/pfs/s3/s3.go
acohen4: I like it!

acohen4:
@PFedak I used an embedded struct with S3Server and actually implemented the Hashing here.

One issue I'm seeing is that adding the TestS3Chain test seems to push us over the timeout for the entire Test Suite. How do we typically handle this? I would probably prefer for each Unit test to be time bound as opposed to the suite

PR #6486, opened 2021-07-07
Raise ingress object to top-level

robert-uhl:


chainlink (approved):
Looks good, can you also move the ingress and tls secret files to their own `ingress` folder (instead of being in dash)

PR #6491, opened 2021-07-07
Request pipeline details for metrics.

### LINKED ISSUE #6485: internal metrics can dereference a nil pointer
After having a local pachd running for an hour, it crashed here:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1dddcd7]

goroutine 225 [running]:
github.com/pachyderm/pachyderm/v2/src/internal/metrics.(*Reporter).internalMetrics.func1(0xc0004f75f0, 0x2962e98, 0xc000da4800, 0xc000113dd8, 0xc000da4980, 0x0)
        /home/jrockway/pach/pachyderm/src/internal/metrics/metrics.go:264 +0x157
github.com/pachyderm/pachyderm/v2/src/internal/metrics.(*Reporter).internalMetrics(0xc0004f75f0, 0xc000113dd8)
        /home/jrockway/pach/pachyderm/src/internal/metrics/metrics.go:364 +0x148
github.com/pachyderm/pachyderm/v2/src/internal/metrics.(*Reporter).reportClusterMetrics(0xc0004f75f0)
        /home/jrockway/pach/pachyderm/src/internal/metrics/metrics.go:149 +0x8e
created by github.com/pachyderm/pachyderm/v2/src/internal/metrics.NewReporter
        /home/jrockway/pach/pachyderm/src/internal/metrics/metrics.go:46 +0x15b
```

The relevant code:

```go
               lpClient, err := r.env.GetPachClient(ctx).PpsAPIClient.ListPipeline(ctx, &pps.ListPipelineRequest{Details: false})
                if err != nil {
                        return err
                }
                pipelineInfos, err := clientsdk.ListPipelineInfo(lpClient)
                if err != nil {
                        return err
                }
                metrics.Pipelines = int64(len(pipelineInfos)) // Number of pipelines
                for _, pi := range pipelineInfos {
                        if pi.Details.ParallelismSpec != nil { # <-------- line 264
                                if metrics.MaxParallelism < pi.Details.ParallelismSpec.Constant {
                                        metrics.MaxParallelism = pi.Details.ParallelismSpec.Constant
                                }
                                if metrics.MinParallelism > pi.Details.ParallelismSpec.Constant {
                                        metrics.MinParallelism = pi.Details.ParallelismSpec.Constant
                                }
                                metrics.NumParallelism++
                        }

```

The only pipeline submitted:

```
$ pachctl list pipeline --raw -o yaml
job_counts:
  "1": 548
last_job_state: JOB_CREATED
parallelism: "1"
pipeline:
  name: a
spec_commit:
  branch:
    name: master
    repo:
      name: a
      type: spec
  id: 3502257bb71545b7afc7541f0f6799d9
state: PIPELINE_RUNNING
type: PIPELINE_TYPE_TRANSFORM
version: "1"
```

This never started successfully because I don't have a worker image that can be pulled, so it's been in a "running" (but not running) state for a really long time.

Anyway, I think the fix is to check that Details isn't nil.  Or make `clientsdk.ListPipelineInfo` never return nil details, even when stuff is broken ;)

PFedak:
Fix #6485 

I guess we don't support the old "allow incomplete" behavior from before, which would have allowed retrieving whatever details we could rather than all or nothing.

### DIFF
--- a/src/internal/metrics/metrics.go
+++ b/src/internal/metrics/metrics.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal"
 	"github.com/pachyderm/pachyderm/v2/src/internal/config"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/middleware/auth"
 	"github.com/pachyderm/pachyderm/v2/src/internal/serviceenv"
 	"github.com/pachyderm/pachyderm/v2/src/internal/uuid"
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
@@ -28,6 +29,8 @@ type Reporter struct {
 	clusterID string
 	env       serviceenv.ServiceEnv
 }
+
+const metricsUsername = "metrics"
 
 // NewReporter creates a new reporter and kicks off the loop to report cluster
 // metrics
@@ -243,6 +246,9 @@ func (r *Reporter) internalMetrics(metrics *Metrics) {
 	// Activation code
 	ctx, cf := context.WithCancel(context.Background())
 	defer cf()
+
+	// add permission-less user to requests
+	ctx = auth.AsInternalUser(ctx, metricsUsername)
 	enterpriseState, err := r.env.EnterpriseServer().GetState(ctx, &enterprise.GetStateRequest{})
 	if err == nil {
 		metrics.ActivationCode = enterpriseState.ActivationCode
--- a/src/internal/metrics/metrics.go
+++ b/src/internal/metrics/metrics.go
@@ -5,7 +5,7 @@ import (
 	"time"
 
 	"github.com/pachyderm/pachyderm/v2/src/enterprise"
-	"github.com/pachyderm/pachyderm/v2/src/internal/clientsdk"
+	"github.com/pachyderm/pachyderm/v2/src/internal"
 	"github.com/pachyderm/pachyderm/v2/src/internal/config"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/serviceenv"
@@ -251,11 +251,7 @@ func (r *Reporter) internalMetrics(metrics *Metrics) {
 
 	// Pipeline info
 	if err := func() error {
-		lpClient, err := r.env.GetPachClient(ctx).PpsAPIClient.ListPipeline(ctx, &pps.ListPipelineRequest{Details: false})
-		if err != nil {
-			return err
-		}
-		pipelineInfos, err := clientsdk.ListPipelineInfo(lpClient)
+		pipelineInfos, err := internal.ListPipeline(ctx, r.env.PpsServer(), &pps.ListPipelineRequest{Details: true})
 		if err != nil {
 			return err
 		}
@@ -354,11 +350,7 @@ func (r *Reporter) internalMetrics(metrics *Metrics) {
 	}
 
 	if err := func() error {
-		rClient, err := r.env.GetPachClient(ctx).PfsAPIClient.ListRepo(ctx, &pfs.ListRepoRequest{})
-		if err != nil {
-			return err
-		}
-		repoInfos, err := clientsdk.ListRepoInfo(rClient)
+		repoInfos, err := internal.ListRepo(ctx, r.env.PfsServer(), &pfs.ListRepoRequest{Type: pfs.UserRepoType})
 		if err != nil {
 			return err
 		}
--- a/src/internal/middleware/auth/handler.go
+++ b/src/internal/middleware/auth/handler.go
@@ -79,11 +79,11 @@ func setWhoAmI(ctx context.Context, username string) context.Context {
 	return context.WithValue(ctx, whoAmIResultKey, username)
 }
 
-// AsInternalUser gives a context's a cached whoami username of form internal:<name>. It also
-// strips away existing incoming metadata to add an empty auth token. As a result, this context will
-// not be able to make additional gRPCs
+// AsInternalUser should never be used during user requests, only internal background jobs.
+// It gives a context a cached whoami username of form internal:<name>. It also overwrites
+// any existing metadata. As a result, this context may not be able to make additional gRPCs.
 func AsInternalUser(ctx context.Context, username string) context.Context {
-	emptyToken := metadata.New(map[string]string{auth.ContextTokenKey: ""})
-	ctx = metadata.NewIncomingContext(ctx, emptyToken)
+	ctx = metadata.NewIncomingContext(ctx, metadata.MD{})
+	ctx = metadata.NewOutgoingContext(ctx, metadata.MD{})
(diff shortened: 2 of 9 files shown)


src/internal/metrics/metrics.go
```diff
@@ -261,18 +261,6 @@ func (r *Reporter) internalMetrics(metrics *Metrics) {
 		}
 		metrics.Pipelines = int64(len(pipelineInfos)) // Number of pipelines
 		for _, pi := range pipelineInfos {
-			if pi.Details.ParallelismSpec != nil {
```
nitinjainsj, line 264: If not here in the loop, where do we update the metrics for max/min parallelism?

PFedak: It's actually lower down in the file, I think this is only here because of an incorrect merge

actgardner: Are we ever going to be able to get the pipeline details with auth enabled? They're populated by reading the spec commit from PFS, but this code runs with no credentials - I don't think we have an internal code path that can read spec commits without credentials ATM.

PFedak: Ah, `GetPipelineDetails` attempts to use the auth token from the pipeline info that gets passed in, but `listPipelineInfo` strips it. I'm confused right now - I don't see any code path that tries to only return the subset of pipelines which the caller is authorized to access, so it seems like if auth is enabled and you don't have access to all repos, `pachctl list pipeline` will error.

PFedak: Now I'm not sure how this is a problem - it seems like exactly this behavior is tested in the auth test `TestCreateAndUpdatePipeline`, enabled by #6432, so I don't think we need auth?

acohen4 (approved):
Overall Looks good. I'm also a little surprised by the GetFile behavior without credentials. I'll try to get a better grip on it and follow up with @actgardner on monday

src/internal/metrics/metrics.go
PFedak: Ok, to summarize the problem: we enforce authorization on internal gRPCs, which meant that even the ListRepo here would fail without credentials (even though it doesn't actually need permissions). The ListPipeline additionally needs to pass an auth check to read the spec repo (which again will work with any non-empty credentials). 

To solve these problems, I added more convenient wrappers for the non-gRPC versions of relevant functions (since I think we're still trying to avoid unnecessary gRPCs?) and made metrics try to request a robot token with no permissions.

src/internal/ppsutil/util.go
```diff
@@ -117,8 +117,9 @@ func GetLimitsResourceList(limits *pps.ResourceSpec) (*v1.ResourceList, error) {
 func GetPipelineDetails(pachClient *client.APIClient, pipelineInfo *pps.PipelineInfo) error {
 	// ensure we are authorized to read the pipeline's spec commit, but don't propagate that back out
 	pachClient = pachClient.WithCtx(pachClient.Ctx())
-	pachClient.SetAuthToken(pipelineInfo.AuthToken)
-
+	if pipelineInfo.AuthToken != "" {
+		pachClient.SetAuthToken(pipelineInfo.AuthToken)
```
actgardner, line 121: I don't think this code is necessary for metrics to work anymore?

src/internal/metrics/metrics.go
```diff
@@ -238,11 +242,33 @@ func inputMetrics(input *pps.Input, metrics *Metrics) {
 	}
 }
 
+func (r *Reporter) attemptGetToken() bool {
+	if r.token != "" {
+		return true
+	}
+	// retry asynchronously here?
+	if resp, _ := r.env.AuthServer().GetRobotToken(context.Background(),
```
actgardner, line 250: I like the idea of generating a token for background jobs, rather than trying to create unauthenticated interfaces across services. Instead of a robot user for this, my recommendation would be to create a new internal API for creating tokens with a unique namespace, like `internal:`, so they're easily identifiable compared to user-created robots.

PFedak: Both of those sound like great ideas, thanks!

src/internal/ppsutil/util.go
PFedak: You're correct, I think this is necessary for the pipeline controller. Happy to just set it outside of this function in that case, though.

src/server/pps/server/pipeline_controller.go
```diff
@@ -96,10 +100,7 @@ func (m *ppsMaster) step(pipeline string, keyVer, keyRev int64) (retErr error) {
 		// Don't put the pipeline in a failing state if we're in the middle
 		// of activating auth, retry in a bit
 		if auth.IsErrNotAuthorized(err) || auth.IsErrNotSignedIn(err) {
-			return stepError{
-				error:        errors.Wrap(err, "couldn't initialize pipeline op"),
-				failPipeline: false,
-			}
+			return newRetriableError(err, "couldn't initialize pipeline op")
```
PFedak, line 103: this wasn't retrying before

src/internal/middleware/auth/handler.go
```diff
@@ -75,3 +78,12 @@ func GetWhoAmI(ctx context.Context) string {
 func setWhoAmI(ctx context.Context, username string) context.Context {
 	return context.WithValue(ctx, whoAmIResultKey, username)
 }
+
+// AsInternalUser gives a context's a cached whoami username of form internal:<name>. It also
+// strips away existing incoming metadata to add an empty auth token. As a result, this context will
+// not be able to make additional gRPCs
+func AsInternalUser(ctx context.Context, username string) context.Context {
+	emptyToken := metadata.New(map[string]string{auth.ContextTokenKey: ""})
```
PFedak, line 86: Hopefully this doesn't create any glaring issues

actgardner: I think based on the implementation of grpc metadata we could just do 
```
emptyToken := metadata.New(map[string]string{})
```

to have zero grpc metadata. We could also reset the outgoing context, just to make sure (if the user uses client.Ctx() it'll have outgoing auth metadata copied from the incoming request).

actgardner (approved):
+1, I think this could be a good general purpose approach to authorizing background tasks. We'll need to be careful that we don't use the internal users to fulfil user-controlled requests, because that can lead to a confused deputy problem.

src/internal/middleware/auth/handler.go
PFedak: I specifically added that key because `getAuthenticatedUser` checks that before the cached value and errors if it's not present. Could also switch those checks but I thought there might be a good reason. Clearing outgoing metadata makes sense to me. Agreed about comment - do you think it could make sense to add logic that prevents "internal:" users from being granted permissions? That would limit the damage it could do (though would complicate things if spec repos go back to requiring a role).

actgardner: We can double-check with @acohen4 but checking the cached value before the grpc metadata makes sense to me. 

Long-term I'd like to be able to grant internal users permissions. I actually don't think it's possible for a user to modify an internal user's permissions today because we validate that the users in a role binding have a valid prefix.

acohen4: makes sense to me too 👍

PR #6492, opened 2021-07-08
Fix helm release

chainlink:


doc/release_instructions.md
```diff
@@ -88,6 +90,8 @@ annotations:
   artifacthub.io/prerelease: "true"
 ```
 
+This will ensure the release is marked as a pre-release on artifact hub
+
 git commit -am "Update Pachyderm version in helm for $(pachctl version --client-only) release"
```
nitinjainsj, line 95: Change this to `git commit -am "Update Helm version to support <VERSION> release`

During an alpha/beta/custom release we don't do the step above (i.e. make VERSION_ADDITIONAL= install) so local `pachctl` may be pointing to the wrong version.

Yesterday, while doing the release, it is possible that we may have to rollback a failed release. If it makes sense add HELM release instructions after doing the point release. We know at that point all the artifacts and images have been published to docker/github.

chainlink:
TODO: Change to `--push` in cr index

doc/release_instructions.md
chainlink: We chatted about this offline. I've added rollback instructions for the helmchart and updated the commit message above to be safer for all situations.

W.r.t building helm on the tag push, we need a way to trigger CI and with circle, a tag is the most straightforward way.

PR #6493, opened 2021-07-08
Set WORKER_USES_ROOT to false by default

actgardner:
By default in 1.x we set a security context on all worker pods to force them to run as root, which breaks OpenShift and seems like overkill generally - we only _need_ to run as root for local deploys where we're using a hostpath. 

We'll need to expose this in the helm chart as well.


chainlink:
@actgardner Do you mind taking a crack at adding this to the helmchart? (it's in the pach repo now) Happy to help if you have questions

actgardner:
@chainlink should be ready to go now :)

chainlink (approved):
Looks good! We do have `requireRoot` under pachd, but I think it makes sense to keep the worker separate

PR #6500, opened 2021-07-08
Add CLI command to explicitly sync with license server

actgardner:
Adds `pachctl enterprise heartbeat` command to explicitly sync the enterprise service immediately. This is only available for cluster admins.


acohen4 (approved):
LGTM!

PR #6508, opened 2021-07-09
Add command to get roles for permission

actgardner:
Users have reported that the permissions errors in 2.0 are less clear and harder to action because they report missing permissions, and the remedy is to grant a role that has the missing permission.

This attempts to make permissions errors easier to respond to, without recommending the user any one role which may not be appropriate for their use case:

```
> pachctl get file test@master:
rpc error: code = Unknown desc = robot:test is not authorized to perform this operation - needs permissions [REPO_READ] on REPO test. Run `pachctl auth roles-for-permission` to find roles that grant a given permission.

> pachctl auth roles-for-permission repo_read
clusterAdmin
repoOwner
repoReader
repoWriter
```


acohen4 (approved):
yay!

PR #6516, opened 2021-07-12
Deflake: TestCronPipeline/RunCronOverwrite

acohen4:
The flaky part of this test is no longer necessary since we are already checking that the cron input repo receives the 4 User Commits

### DIFF
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -27,7 +27,6 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/client"
 	"github.com/pachyderm/pachyderm/v2/src/internal/ancestry"
 	"github.com/pachyderm/pachyderm/v2/src/internal/backoff"
-	"github.com/pachyderm/pachyderm/v2/src/internal/clientsdk"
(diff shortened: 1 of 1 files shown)


chainlink (approved):
Lgtm

src/server/pachyderm_test.go
```diff
@@ -6097,13 +6097,20 @@ func TestCronPipeline(t *testing.T) {
 					})
 				}))
 				listCommitClient, err := c.PfsAPIClient.ListCommit(c.Ctx(), &pfs.ListCommitRequest{
-					Repo:       client.NewRepo(repo),
-					OriginKind: pfs.OriginKind_USER,
+					Repo:       client.NewRepo(pipeline8),
```
PFedak, line 6100: after some archaeology, I think this whole list commit part can be deleted, and we should go back to checking the file count in the downstream repo (could just wait on that commit instead of `WaitCommitSetAll`)

PFedak (changes requested):
Also, as mentioned, could you change the cron spec to `*/1` for minute, so we don't randomly wait an extra minute sometimes?

PFedak (approved):
LGTM, thanks!

PR #6520, opened 2021-07-12
Load testing PPS

brycemcanally:
This PR updates the load testing to work with PPS and adds a PPS load test to CI. Also, the following changes were made to address performance issues:
- The load testing now generates the files in a streaming fashion rather than buffering them in memory.
- The storage layer now respects the memory threshold correctly. Previously, the actual memory usage could be double the threshold because it did not account for buffering in the chunk layer.
- Slight change to file merging such that a file fully contained in one additive set (the common path) is read with a file reader rather than a merge file reader. This has a significant impact on the performance of the hash computation since the hash can be computed by just looking at the metadata in this case.
- Datums in jobs are now laid out with respect to the lexicographical ordering of the input files when no parent job exists. The no parent job code path was skipping the step where we prefix the paths in the datum file set with the index from the datum iterator.


src/internal/storage/fileset/unordered_writer.go
```diff
@@ -29,6 +29,9 @@ func newUnorderedWriter(ctx context.Context, storage *Storage, memThreshold int6
 	if err := storage.filesetSem.Acquire(ctx, 1); err != nil {
 		return nil, err
 	}
+	// Half of the memory will be for buffering in the unordered writer.
+	// The other half will be for buffering in the chunk writer.
+	memThreshold /= 2
```
acohen4, line 34: 👍

src/internal/storage/fileset/merge.go
```diff
@@ -49,23 +49,21 @@ func (mr *MergeReader) iterate(ctx context.Context, cb func(File) error) error {
...
+			fss = append(fss, fs)
+		}
+		if len(fss) == 0 {
+			return nil
 		}
 		if len(fss) == 1 {
-			if fss[0].deletive {
-				return nil
-			}
 			return cb(newFileReader(ctx, mr.chunks, fss[0].file.Index()))
```
acohen4, line 63: I'm confused to why this case wasn't being hit previously

acohen4 (approved):
I left one question but otherwise it looks great!

src/internal/storage/fileset/merge.go
brycemcanally: Files normally get into the system through put file which creates a delete and add operation for the files. When we are merging here, this will result with two file streams (one for the delete and one for the add). In this case, and really any case where there is only one additive stream, we don't need to use a merge file reader since we are not merging content from multiple file sets.

PR #6522, opened 2021-07-13
Stop pipeline during deletion.

PFedak:
In some of our tests, we delete a pipline which has commits being
created very frequently. Deleting a pipeline output repo can end up
being a large transaction, which can lead to DeleteAll hanging.

### DIFF
--- a/src/server/auth/server/testing/auth_test.go
+++ b/src/server/auth/server/testing/auth_test.go
@@ -2428,7 +2428,7 @@ func TestDeletePipelineMissingInput(t *testing.T) {
 		false,
 	))
 
-	// force-delete input and output repos
+	// force-delete input repo
 	require.NoError(t, aliceClient.DeleteRepo(repo, true))
 
 	// Attempt to delete the pipeline--must succeed
--- a/src/server/auth/server/testing/auth_test.go
+++ b/src/server/auth/server/testing/auth_test.go
@@ -983,10 +983,14 @@ func TestStopAndDeletePipeline(t *testing.T) {
 		buildBindings(alice, auth.RepoOwnerRole, bob, auth.RepoWriterRole, pl(pipeline), auth.RepoWriterRole),
 		getRepoRoleBinding(t, aliceClient, pipeline))
 
-	// bob still can't stop or delete alice's pipeline
-	require.NoError(t,  bobClient.StopPipeline(pipeline))
-	require.NoError(t,  bobClient.DeletePipeline(pipeline, false))
-
+	// bob can now stop the pipeline, but can't start or delete it
+	require.NoError(t, bobClient.StopPipeline(pipeline))
+	err = bobClient.StartPipeline(pipeline)
+	require.YesError(t, err)
+	require.Matches(t, "not authorized", err.Error())
+	err = bobClient.DeletePipeline(pipeline, false)
+	require.YesError(t, err)
+	require.Matches(t, "not authorized", err.Error())
 	// alice re-adds bob as a reader of the input repo
 	require.NoError(t, aliceClient.ModifyRepoRoleBinding(repo, bob, []string{auth.RepoReaderRole}))
 	require.Equal(t,
--- a/src/server/auth/server/testing/auth_test.go
+++ b/src/server/auth/server/testing/auth_test.go
@@ -984,12 +984,8 @@ func TestStopAndDeletePipeline(t *testing.T) {
 		getRepoRoleBinding(t, aliceClient, pipeline))
 
 	// bob still can't stop or delete alice's pipeline
-	err = bobClient.StopPipeline(pipeline)
-	require.YesError(t, err)
-	require.Matches(t, "not authorized", err.Error())
-	err = bobClient.DeletePipeline(pipeline, false)
-	require.YesError(t, err)
-	require.Matches(t, "not authorized", err.Error())
+	require.NoError(t,  bobClient.StopPipeline(pipeline))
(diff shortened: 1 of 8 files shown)


src/server/pps/server/api_server.go
```diff
@@ -2733,38 +2737,48 @@ func (a *apiServer) StopPipeline(ctx context.Context, request *pps.StopPipelineR
...
 			}
-			newPipelineInfo.Stopped = true
-			return nil
-		}); err != nil {
+		} else if !col.IsErrNotFound(err) {
 			return err
 		}
 
 		// Kill any remaining jobs
+		// if the pipeline output repo doesn't exist, we technically run this without authorization,
```
msteffen, line 2780: Does `StopPipeline` do an auth check?

src/server/pps/server/api_server.go
```diff
@@ -2733,38 +2737,48 @@ func (a *apiServer) StopPipeline(ctx context.Context, request *pps.StopPipelineR
...
-		}); err != nil {
-			return err
-		}
-		if err := a.env.PfsServer().CreateBranchInTransaction(txnCtx, &pfs.CreateBranchRequest{
-			Branch:     client.NewSystemRepo(pipelineInfo.Pipeline.Name, pfs.MetaRepoType).NewBranch(pipelineInfo.Details.OutputBranch),
-			Provenance: nil,
-		}); err != nil {
-			return err
-		}
+			// Remove branch provenance to prevent new output and meta commits from being created
```
msteffen, line 2749: Does `StopPipeline` not remove this provenance? Does that mean output and meta commits might be created for stopped pipelines?

src/server/pps/server/api_server.go
```diff
@@ -2733,38 +2737,48 @@ func (a *apiServer) StopPipeline(ctx context.Context, request *pps.StopPipelineR
...
-			if newPipelineInfo.Version != pipelineInfo.Version {
-				// If the pipeline has changed, restart the transaction and try again
-				return col.ErrTransactionConflict{}
+			newPipelineInfo := &pps.PipelineInfo{}
+			if err := a.pipelines.ReadWrite(txnCtx.SqlTx).Update(pipelineInfo.Pipeline.Name, newPipelineInfo, func() error {
+				if newPipelineInfo.Version != pipelineInfo.Version {
+					// If the pipeline has changed, restart the transaction and try again
+					return col.ErrTransactionConflict{}
+				}
+				newPipelineInfo.Stopped = true
```
msteffen, line 2770: Does `StopPipeline` not set `pipelineInfo.Stopped`?

msteffen (commented):
A few high-level questions, though these may just be gaps in my understanding

src/server/pps/server/api_server.go
msteffen: My mistake, this change is to the body of `StopPipeline`

src/server/pps/server/api_server.go
msteffen: As above, this change is to the body of `StopPipeline`

src/server/pps/server/api_server.go
msteffen: As above, this change is to the body of `StopPipeline`

src/server/auth/server/testing/auth_test.go
```diff
@@ -983,14 +983,14 @@ func TestStopAndDeletePipeline(t *testing.T) {
 		buildBindings(alice, auth.RepoOwnerRole, bob, auth.RepoWriterRole, pl(pipeline), auth.RepoWriterRole),
 		getRepoRoleBinding(t, aliceClient, pipeline))
 
-	// bob still can't stop or delete alice's pipeline
-	err = bobClient.StopPipeline(pipeline)
+	// bob can now stop the pipeline, but can't start or delete it
```
actgardner, line 986: Is this comment accurate? I thought Bob could delete this pipeline, and we're adding the ability to stop the pipeline so it can be stopped before deleting? 

For symmetry, should bob also be able to restart the stopped pipeline? It kind of feels like the only time we should care about the input repos is at pipeline creation/update time.

PFedak: My goal was to not change the permissions for deleting a pipeline at all. I realize now that I was still thinking about the old auth model, where having `delete` permission implied `write` permission, but in principle someone could create role bindings now which make that not the case.

The main intended auth change was to not require input repo read access to stop a pipeline. I considered your symmetry point - we currently have the same issue where being able to delete a pipeline does not mean you can recreate it. I agree with you that input repo read access seems like a more strict requirement than makes sense in many of these cases.

The two solutions I'm thinking of are just ignoring auth errors on the stop attempt during delete, which puts aside the issue of the proper permissions for `StopPipeline`, or add a new explicit stop/start pipeline operation type with the desired semantics. Should probably do both, as it would be bad for a delete to fail for lack of write access, as bizarre as the perm combo seems to me.

actgardner: Yeah I'm thinking let's get away from reusing the operation types and be explicit about each one - it might be a little more verbose but then this kind of issue will be immediately apparent :)

actgardner: I think this comment is still inaccurate but otherwise the behaviour looks good 👍

actgardner (approved):
LGTM, one nit about the comment in the tests

src/server/auth/server/testing/auth_test.go
PFedak: Ah whoops, thanks! fixed things up

PR #6525, opened 2021-07-13
Change worker_uses_root env var to follow pachd.storage.local.requireRoot

chainlink:
Since root is required locally to write to `hostPath` synchronize both pach and workers to use the same value.

Fixes issue where local workloads would not work by default and required `usesRoot` to be set to true manually

### DIFF
--- a/etc/helm/pachyderm/templates/pachd/deployment.yaml
+++ b/etc/helm/pachyderm/templates/pachd/deployment.yaml
@@ -73,8 +73,10 @@ spec:
           {{- end }}
         - name: WORKER_IMAGE
           value: "{{ .Values.worker.image.repository }}:{{ default .Chart.AppVersion .Values.pachd.image.tag }}"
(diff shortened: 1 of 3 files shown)


etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -73,8 +73,10 @@ spec:
           {{- end }}
         - name: WORKER_IMAGE
           value: "{{ .Values.worker.image.repository }}:{{ default .Chart.AppVersion .Values.pachd.image.tag }}"
+        {{- if and (eq .Values.deployTarget "LOCAL") .Values.pachd.storage.local.requireRoot }}
         - name: WORKER_USES_ROOT
-          value: {{ .Values.worker.usesRoot | quote }}
+          value: "True"
```
robert-uhl, line 78: `"True"`, not `"true"` or `"TRUE"`?

I see `"true"` a couple other places in the templates, FWIW.

robert-uhl (approved):
LGTM — casing is not at all a showstopper.

PR #6534, opened 2021-07-14
Configure prom port in Service env instead of assets

chainlink:
Prep for the removal of assets.go files


msteffen (approved):
LGTM

jrockway (approved):
In PRs like this I'd like to see assets.PrometheusPort get deleted, so that CI can verify that it's not secretly used somewhere else.  But, you say you're doing that in a follow-up and I can live with that :)

PR #6535, opened 2021-07-14
Initial commit for pg bouncer

chainlink:


etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+      containers:
+      - env:
+        - name: DB_USER
+          value: pachyderm
+        - name: DB_PASSWORD
+          value: elephantastic
+        - name: DB_HOST
+          value: postgres.default
+        - name: AUTH_TYPE
+          value: trust
```
brendoncarroll, line 38: I think this needs to be `any` ?

etc/helm/pachyderm/templates/pgbouncer/service.yaml
```diff
@@ -0,0 +1,24 @@
...
+  creationTimestamp: null
+  labels:
+    app: pg-bouncer
+    suite: pachyderm
+  name: pg-bouncer
+  namespace: {{ .Release.Namespace }}
+spec:
+  ports:
+  - name: client-port
+    #nodePort: 32228
```
brendoncarroll, line 18: Can you make this 32229?

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
chainlink: Makes sense, I need to plumb through alot of the options

PR #6541, opened 2021-07-15
Kill final pieces of http api from helm chart

chainlink:


etc/helm/pachyderm/values.schema.json
```diff
@@ -566,9 +563,6 @@
                             "type": "string"
                         }
                     }
-                },
-                "usesRoot": {
```
jrockway, line 570: Oh, what was this?

jrockway (approved):
I wondered what the HTTP API port was.  Now I know -- a blackhole ;)

etc/helm/pachyderm/values.schema.json
chainlink: that was killed in another PR, but looks like the schema for it got missed, I can break that out into another cleanup PR

jrockway: Nah, doing it here is fine.  api port won't be coming back ;)

jrockway (approved):

Issue #6543, opened 2021-07-16
Segmentation Faults in pachd and Pipeline storage Containers


dgeorg42:
**What happened?**:

I have a single pipeline deployed for performance testing. The glob pattern is "/", so the entire input repo is treated as a single datum.  The code in the pipeline simply copies all of the files from the input repo to the output repo. My performance test puts 10,000 files at a time into the input repo using "pachctl put file", waits for the pipeline to complete, then puts another 10,000 files into the input repo. It does this until a total of 200,000 files exist in the repo, at which point it stops. Therefore, the pipeline will run a total of 20 times, each time with more files than the previous run.

I started a run of these tests last night, and checked the results this morning. I discovered that both pachd and my pipeline's storage container had triggered restarts due to segmentation faults, not just once, but multiple times:

```
{pach 2.0.0-beta.1 : aws} / {k8s 1.16.13-dispatcher : aws} / {git 2.24.3 : pachyderm/master}
~/Documents/Work/Repos/pachyderm -->> kubectl get pods
NAME                           READY   STATUS    RESTARTS   AGE
etcd-0                         1/1     Running   0          12h
pachd-6ffcddcfbc-7cm6v         1/1     Running   13         12h
pipeline-benchtests-v3-bd945   2/2     Running   11         11h
postgres-0                     1/1     Running   0          12h
```

From the end of the previous pachd container log:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1dd8c77]

goroutine 162 [running]:
github.com/pachyderm/pachyderm/v2/src/internal/metrics.(*Reporter).internalMetrics.func1(0xc0003ad680, 0x295b178, 0xc001972b00, 0xc000713dd8, 0xc001972cc0, 0x0)
	/Users/nitin/Code/Pachyderm/pachyderm/src/internal/metrics/metrics.go:264 +0x157
github.com/pachyderm/pachyderm/v2/src/internal/metrics.(*Reporter).internalMetrics(0xc0003ad680, 0xc000713dd8)
	/Users/nitin/Code/Pachyderm/pachyderm/src/internal/metrics/metrics.go:364 +0x148
github.com/pachyderm/pachyderm/v2/src/internal/metrics.(*Reporter).reportClusterMetrics(0xc0003ad680)
	/Users/nitin/Code/Pachyderm/pachyderm/src/internal/metrics/metrics.go:149 +0x8e
[...]

PR #6547, opened 2021-07-16
Fix metrics pipeline details

brycemcanally:
The list pipeline call in the metrics collection was not collecting the details which was causing a panic when attempting to access the details field. Also, the place where the details field was being accessed is a duplicate of what is captured in a block that protects against the details field being nil. We need to invest some time in making metrics collection unit testable, or at least add some tests that will hit the metrics collection code path to avoid these issues.

Fixes: #6543

### DIFF
--- a/src/internal/metrics/metrics.go
+++ b/src/internal/metrics/metrics.go
@@ -251,7 +251,7 @@ func (r *Reporter) internalMetrics(metrics *Metrics) {
 
 	// Pipeline info
 	if err := func() error {
-		lpClient, err := r.env.GetPachClient(ctx).PpsAPIClient.ListPipeline(ctx, &pps.ListPipelineRequest{Details: true})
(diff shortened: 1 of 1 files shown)


jrockway:
@brycemcanally @PFedak there is another PR open to fix this issue: https://github.com/pachyderm/pachyderm/pull/6491

I mention it because there was a lot of discussion on that PR, so maybe it's not as simple a fix as it appears.

brycemcanally:
Ohh, interesting. Well, we can at least fix the panic. Whether or not we can collect the details can be figured out later I guess.

PFedak (approved):
LGTM, my change was mainly about making metrics work with auth (again). I'd be happy if you took a look at it, @brycemcanally , as I might have gone overboard, but I'll probably ask for @actgardner to review, too.

PR #6548, opened 2021-07-16
For Localhost Enterprise Deployments, store the in-cluster pachd hostname as issuer instead of 'localhost'

acohen4:
The Change:
To make the identity-server accessible + usable for pods other than pachd that are deployed in the same k8s cluster (in this case Dash), we must set the Issuer (a.k.a Identity-Server / Dex) as having the in-cluster URL so that the other in-cluster clients can run through the OIDC flows (Client establishment + Token endpoint).
Since the browser gets redirected to the auth callback endpoints, we overwrite those endpoints' Host to be `localhost:30658`  

Dependencies:

~~1. For Minikube deployments, one may need to update the networking config so that pods can connect to themeselves via service, 
  i.e. pachd pod can reach itself by using its service DNS host as `pachd:1658`.
with: 
`minikube ssh -- sudo ip link set docker0 promisc on`~~

2. Dex's Library needs to be updated so that it doesn't enforce that the it's Callback URL (localhost:30650/...) doesn't equal its Issuer URL: (pachd:1658/...)


src/server/identity/server/connector.go
```diff
@@ -16,19 +16,26 @@ var (
...
+}
 
-func (*placeholder) LoginURL(s connector.Scopes, callbackURL, state string) (string, error) {
+func (p *placeholder) LoginURL(s connector.Scopes, callbackURL, state string) (string, error) {
 	u, err := url.Parse(callbackURL)
 	if err != nil {
 		return "", fmt.Errorf("failed to parse callbackURL %q: %v", callbackURL, err)
 	}
+	if p.LocalhostIssuer {
+		u.Host = "localhost:30658"
```
acohen4, line 37: move this to global constant

etc/helm/examples/local-dev-values.yaml
```diff
@@ -15,7 +15,18 @@ pachd:
...
+  enabled: true
+  config:
+    issuerURI: "http://pachd:1658"
+    oauthRedirectURI: "http://localhost:4000/oauth/callback/?inline=true"
+    oauthClientID: "dash"
+    oauthClientSecret: "pgJG75eecrhJEbpWDjvi4q1zw6gFRX1o"
+    graphqlPort: 4000
+    oauthPachdClientID: "pachd"
+    pachdAddress: "pachd-peer.default.svc.cluster.local:30653"
+  image:
```
acohen4, line 27: will remove this image section

src/internal/testutil/identity.go
```diff
@@ -23,7 +23,7 @@ const DexMockConnectorEmail = `kilgore@kilgore.trout`
 // OIDCOIDCConfig is an auth config which can be used to connect to the identity service in tests
 func OIDCOIDCConfig() *auth.OIDCConfig {
 	return &auth.OIDCConfig{
-		Issuer:          "http://localhost:30658/",
+		Issuer:          "http://pachd:1658/",
```
actgardner, line 26: This doesn't work in minikube by default because hairpin network routes (where a pod connects to a service it's a part of) don't work. So we may need both the re-write for localhost _and_ the issuer change for this to be effective :/

src/server/auth/server/oidc.go
```diff
@@ -84,31 +86,19 @@ func half(state string) string {
...
-	oauthConfig   oauth2.Config
-	rewriteClient *http.Client
+	oidcProvider *oidc.Provider
+	oauthConfig  oauth2.Config
 }
 
 func newOIDCConfig(ctx context.Context, config *auth.OIDCConfig) (*oidcConfig, error) {
-	var rewriteClient *http.Client
-	var err error
-	if config.LocalhostIssuer {
```
actgardner, line 95: Some version of this code is still going to be necessary because by default minikube doesn't support hairpin network routes, so a pod in the pachd service can't access itself (https://kubernetes.io/docs/tasks/debug-application-cluster/debug-service/#a-pod-fails-to-reach-itself-via-the-service-ip)

src/server/auth/server/http_client.go
```diff
@@ -0,0 +1,51 @@
+package server
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
```
chainlink, line 8: Perhaps we should add a note here about hairpin networks

cpepin:
@acohen4 can we add the `REACT_APP_RUNTIME_ISSUER_URI` to the values.schema.json? We don't need to set it for local deployments, but hub generates some code based on that file to set the console config.

etc/helm/examples/local-dev-values.yaml
```diff
@@ -15,7 +15,15 @@ pachd:
     enabled: false
 
 dash:
-  enabled: false
+  enabled: true
+  config:
+    issuerURI: "http://pachd:1658"
+    oauthRedirectURI: "http://localhost:4000/oauth/callback/?inline=true"
+    oauthClientID: "dash"
+    oauthClientSecret: "pgJG75eecrhJEbpWDjvi4q1zw6gFRX1o"
```
actgardner, line 23: We probably shouldn't hard-code a secret here.

src/server/identity/db.go
```diff
@@ -40,3 +40,11 @@ ALTER TABLE identity.config ADD COLUMN
 	id_token_expiry VARCHAR(4096)`)
 	return err
 }
+
+// AddTokenExpiryConfig adds expiry fields for token lifespan to the server config
```
actgardner, line 44: This comment is not accurate :)

src/server/identity/server/connector.go
```diff
@@ -10,25 +10,34 @@ import (
...
 
-func (*placeholder) LoginURL(s connector.Scopes, callbackURL, state string) (string, error) {
+func (p *placeholder) LoginURL(s connector.Scopes, callbackURL, state string) (string, error) {
 	u, err := url.Parse(callbackURL)
 	if err != nil {
 		return "", fmt.Errorf("failed to parse callbackURL %q: %v", callbackURL, err)
 	}
+	if p.LocalhostIssuer {
+		u.Host = localhostIdentityServer
+	}
```
actgardner, line 40: I'm wondering if we could get away with a relative redirect to avoid this? Not a big deal either way.

src/server/identity/db.go
acohen4: 🤦 ... thanks

actgardner (approved):
One comment about the placeholder connector redirect - if we just used a relative redirect (`"/static/not-configured.html"`) I wonder if we could avoid having an extra config for the identity server. Otherwise LGTM

PR #6550, opened 2021-07-17
helm: mark "enabled" fields required in the schema

jrockway:
This improves the results of utilities that generate code from the schema, and it doesn't seem to break any values files that we have in the repo.

### DIFF
--- a/etc/helm/pachyderm/values.schema.json
+++ b/etc/helm/pachyderm/values.schema.json
@@ -566,6 +566,7 @@
         },
         "pgbouncer": {
             "type": "object",
+            "required": ["enabled"],
             "properties": {
                 "enabled": {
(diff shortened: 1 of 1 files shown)


chainlink:
This is a nice optimisation! Currently we generate the schema with an automated tool which would overwrite this though. Schema customizations seem quite valuable, but not sure how to reconcile with the convenience and stability of the auto gen

jrockway:
Oh shit, the codegen goes deep.  Honestly, this is just to work around a bug in `schematyper`, so we could probably fix it there instead.

chainlink:
We currently use this to produce the json schema from the values.yaml, but there may be better options https://github.com/karuppiah7890/helm-schema-gen

PR #6551, opened 2021-07-19
Re-enable obj client integration tests.

brendoncarroll:
These were accidentally disabled because the files did not end in `_test.go`.  Some of what they cover is redundant with the deployment tests, but this is definitely the preferred way to test the clients; it only requires credentials for a bucket.


src/internal/obj/minio_client.go
```diff
@@ -46,11 +47,19 @@ func newMinioClientV2(endpoint, bucket, id, secret string, secure bool) (*minioC
...
-	if _, err := io.Copy(wc, r); err != nil {
+	opts := minio.PutObjectOptions{
+		ContentType: "application/octet-stream",
+		PartSize:    uint64(8 * 1024 * 1024),
+	}
+	n, err := c.Client.PutObjectWithContext(ctx, c.bucket, name, r, -1, opts)
+	if err != nil {
+		return err
+	}
+	if n == 0 {
```
Tryneus, line 58: what's going on here?  could probably use a comment

src/internal/obj/amazon_client.go
```diff
@@ -301,10 +301,14 @@ func (c *amazonClient) transformError(err error, objectPath string) error {
 	if strings.Contains(err.Error(), "Not Found") {
 		return pacherr.NewNotExist(c.bucket, objectPath)
 	}
-	var awsErr awserr.Error
-	if !errors.As(err, &awsErr) {
+	awsErr, ok := err.(awserr.Error)
```
Tryneus, line 304: pretty sure we should be using `errors.As` instead of casting as a general rule since we have wrapped errors around

Tryneus (commented):
A couple nits but overall LGTM

src/internal/obj/minio_client.go
brendoncarroll: I was trying to fix the minio client to support 0 length objects.  This didn't fix it, so I will just remove it.

src/internal/obj/amazon_client.go
brendoncarroll: I think errors.Is and errors.As are for concrete types.  For interfaces (which `awserr.Error` is), you are supposed to use a normal interface cast.

brendoncarroll: The request canceled case should break if they start returning awserr wrapped in anything.

Tryneus (approved):

PR #6554, opened 2021-07-19
Detect incompatible pachd version in pachctl

actgardner:
A different approach to https://github.com/pachyderm/pachyderm/pull/6552, with the advantage of not breaking old beta clients.


jrockway (approved):
I think this makes it clear that there is a problem, and doesn't have a performance impact.

I think we should also make pachd detect v1 clients on the server side and return a useful error message.

PR #6562, opened 2021-07-20
Add middleware to throw a user-friendly error when using an old client

actgardner:
When a user tries to use a 1.x client against a 2.x server, raise a clearer error. Also moves the auth interceptor to group them both into `src/internal/middleware/`.

The implementation is complicated by the way GRPC handles missing services:
- if there's no `UnknownServiceHandler`, GRPC will immediately throw an error without running any interceptors
- if there is an `UnknownServiceHandler`, GRPC will run the interceptors, which prompts the auth interceptor to fail (because this is an unknown method)

This PR installs a useless `UnknownServiceHandler`, which only has the effect of making the interceptors run. The version interceptor runs first, and checks whether the service is a known V1 service. If the service isn't blacklisted, the auth interceptor checks against a specific list of all methods. This should give us the most precise possible errors for both cases (an end-user using the wrong client versus a developer adding a new method and forgetting to add it to the auth interceptor).


jrockway (approved):
Very nice!  Good call to only apply the version check to the public port; less code between the question and the answer when we know the client is legit ;)

PR #6563, opened 2021-07-20
Branding dex assets

cpepin:
This takes another pass at branding the Dex assets. Basically just uses all of our global variables, and makes things look like they are view in hub/dash.

A couple thoughts I had while doing this:
- It might be nice to eventually add a `dev_server.go` to the `/dex_assets` directory that just compiles the templates using the active variables. I ended up just running a basic web server out of that director and manually injected the template variables.
- Eventually, the FE team should host some of these assets (particularly the global.css stuff) in a CDN. Maybe even the logos as well.


actgardner (approved):
Looks great! Re: using a CDN, some security conscious customers might not want to pull assets from the web, it might be worth checking with the customer team.

PR #6564, opened 2021-07-20
Update Dex to v2.29.0

actgardner:
Bump the version of Dex to the most recent release. This requires pinning the grpc and prometheus modules because we have transitive dependencies that rely on the older versions.


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/coreos/etcd v3.3.13+incompatible
 	github.com/coreos/go-oidc v2.2.1+incompatible
 	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f
-	github.com/dexidp/dex v0.0.0-20201118094123-6ca0cbc85759
+	github.com/dexidp/dex v0.0.0-20210628213014-ceb4324c183d
```
acohen4, line 18: where is this version from? I don't see it on https://openbase.com/go/github.com/dexidp/dex/@/versions

actgardner: hmm, I thought that SHA was tagged at v2.29.0 but it turns out it's one of the parents of that merge commit - I'll fix this to be the v2.29.0 commit instead.

go.mod
```diff
@@ -56,49 +54,58 @@ require (
...
+	sigs.k8s.io/yaml v1.2.0
 )
 
 replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible
 
 replace k8s.io/client-go => k8s.io/client-go v0.0.0-20190718183610-8e956561bbf5
 
 replace github.com/sercand/kuberesolver => github.com/sercand/kuberesolver v1.0.1-0.20200204133151-f60278fd3dac
+
+replace google.golang.org/grpc => google.golang.org/grpc v1.29.1
```
actgardner, line 99: This is the full module surgery required to keep our etcd client happy - we can't let the protobuf library move past 1.3.5 or there's a runtime error. That requires us to pin everything that uses protobuf to a sufficiently old version as well. Long-term we should pay off some technical debt by upgrading the etcd library.

src/internal/testutil/identity.go
```diff
@@ -109,6 +109,10 @@ func DoOAuthExchange(t testing.TB, pachClient, enterpriseClient *client.APIClien
 	resp, err := c.Get(RewriteURL(t, loginURL, DexHost(enterpriseClient)))
 	require.NoError(t, err)
 
+	// Dex login redirects to the provider page, which will generate it's own state
```
actgardner, line 112: Dex made a change where the incoming state is stored in the DB and a new nonce is created for the login method, which adds a redirect to the login flow. We just need to follow the extra redirect and everything works.

src/server/identity/server/dex_web_test.go
```diff
@@ -119,18 +120,19 @@ func TestUpdateIDP(t *testing.T) {
 	err := env.GetDexDB().CreateConnector(dex_storage.Connector{ID: "conn", Type: "github", Config: []byte(`{"clientID": "test1", "redirectURI": "/callback"}`)})
 	require.NoError(t, err)
 
-	// Create an auth request that hasn't done the flow
-	require.NoError(t, env.GetDexDB().CreateAuthRequest(dex_storage.AuthRequest{
-		ID:       "testreq",
-		ClientID: "testclient",
+	require.NoError(t, env.GetDexDB().CreateClient(dex_storage.Client{
```
actgardner, line 123: The client for the request actually has to exist now.

src/server/identity/server/dex_web_test.go
```diff
@@ -119,18 +120,19 @@ func TestUpdateIDP(t *testing.T) {
...
+		ID:           "testclient",
+		RedirectURIs: []string{"http://example.com/callback"},
 	}))
 
 	// make a request to authenticate
 	recorder := httptest.NewRecorder()
-	req := httptest.NewRequest("GET", "/auth/conn?req=testreq", nil)
+	req := httptest.NewRequest("GET", "/auth/conn?req=testreq&client_id=testclient&redirect_uri=http%3A%2F%2Fexample.com%2Fcallback&scope=openid&response_type=code", nil)
```
actgardner, line 130: Dex requires all these parameters to be passed to the connector, previously I think they were just stashed in the auth request. I haven't dug into the architectural reason for the change but it should be transparent to end users, we just need to handle it when we're trying to programmaticaly drive the OAuth login flow.

go.mod
acohen4: should we capture this hairy detail in a comment here?

acohen4 (approved):
LGTM!

PR #6566, opened 2021-07-20
Rewrites Local Object Client

brendoncarroll:
Rewrites the local object client to fix a few issues:
- It was impossible to have an object at a path containing a slash, and any other object at a path having the first path as a prefix.  This is now fixed by base64 encoding the paths.
- Concurrent access to the store had the possibility to return partially written files.  This is fixed by writing to a staging area under a UUID and then renaming the completed file into it's final place.

The staging area is cleared during client initialization.  So "turn it off and on again" will be an available strategy for troubleshooting.

### DIFF
--- a/src/internal/obj/local_client.go
+++ b/src/internal/obj/local_client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -15,11 +16,25 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// NewLocalClient returns a Client that stores data on the local file system
+func NewLocalClient(rootDir string) (Client, error) {
+	var c Client
+	var err error
+	c, err = newFSClient(rootDir)
+	if err != nil {
+		return nil, err
+	}
+	if monkeyTest {
+		c = &monkeyClient{c}
+	}
+	return newUniformClient(c), nil
+}
+
 type fsClient struct {
 	rootDir string
 }
 
-func NewLocalClient(rootDir string) (Client, error) {
+func newFSClient(rootDir string) (Client, error) {
 	c := &fsClient{rootDir: filepath.Clean(rootDir)}
 	if c.rootDir == "" || c.rootDir == "/" || c.rootDir == "." {
 		panic("you probably didn't want to set the local client's root path to " + c.rootDir)
@@ -94,6 +109,10 @@ func (c *fsClient) Walk(ctx context.Context, prefix string, cb func(string) erro
 		}
 	}
 	return nil
+}
+
+func (c *fsClient) BucketURL() string {
+	return fmt.Sprintf("local://%s/", strings.ReplaceAll(filepath.ToSlash(c.rootDir), "/", "."))
 }
 
 func (c *fsClient) stagingPathFor(name string) string {
--- a/src/internal/obj/local_client.go
+++ b/src/internal/obj/local_client.go
@@ -18,9 +18,7 @@ import (
 
 // NewLocalClient returns a Client that stores data on the local file system
 func NewLocalClient(rootDir string) (Client, error) {
-	var c Client
-	var err error
-	c, err = newFSClient(rootDir)
+	c, err := newFSClient(rootDir)
 	if err != nil {
 		return nil, err
 	}
@@ -35,21 +33,26 @@ type fsClient struct {
 }
 
 func newFSClient(rootDir string) (Client, error) {
-	c := &fsClient{rootDir: filepath.Clean(rootDir)}
+	c := &fsClient{
+		rootDir: filepath.Clean(rootDir),
+	}
 	if c.rootDir == "" || c.rootDir == "/" || c.rootDir == "." {
 		panic("you probably didn't want to set the local client's root path to " + c.rootDir)
 	}
-	return c, c.init()
+	if err := c.init(); err != nil {
+		return nil, err
+	}
+	return c, nil
 }
 
-func (c *fsClient) Put(ctx context.Context, name string, r io.Reader) error {
+func (c *fsClient) Put(ctx context.Context, name string, r io.Reader) (retErr error) {
 	staging := c.stagingPathFor(name)
 	final := c.finalPathFor(name)
 	f, err := os.Create(staging)
 	if err != nil {
 		return err
 	}
-	defer f.Close()
+	defer c.closeFile(&retErr, f)
 	if _, err := io.Copy(f, r); err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


brendoncarroll:
`os.ReadDir` was added in 1.16... trying to figure out where else we set a Go version.

src/internal/obj/local_client.go
```diff
@@ -1,126 +1,149 @@
...
 // NewLocalClient returns a Client that stores data on the local file system
-func NewLocalClient(root string) (Client, error) {
-	if err := os.MkdirAll(root, 0755); err != nil {
-		return nil, errors.EnsureStack(err)
-	}
+func NewLocalClient(rootDir string) (Client, error) {
 	var c Client
-	c = &localClient{filepath.Clean(root)}
+	var err error
+	c, err = newFSClient(rootDir)
```
brycemcanally, line 23: This could just be `c, err := newFSClient(rootDir)`.

src/internal/obj/local_client.go
```diff
@@ -1,126 +1,149 @@
...
 
-func (c *localClient) normPath(path string) string {
-	path = filepath.Clean(path)
-	return filepath.Join(c.root, path)
+func newFSClient(rootDir string) (Client, error) {
+	c := &fsClient{rootDir: filepath.Clean(rootDir)}
+	if c.rootDir == "" || c.rootDir == "/" || c.rootDir == "." {
+		panic("you probably didn't want to set the local client's root path to " + c.rootDir)
+	}
+	return c, c.init()
```
brycemcanally, line 42: I don't like potentially returning a non-nil client with a non-nil error.

src/internal/obj/local_client.go
```diff
@@ -1,126 +1,149 @@
...
+func (c *fsClient) Put(ctx context.Context, name string, r io.Reader) error {
+	staging := c.stagingPathFor(name)
+	final := c.finalPathFor(name)
+	f, err := os.Create(staging)
+	if err != nil {
 		return err
 	}
-	file, err := os.Create(fullPath)
-	if err != nil {
+	defer f.Close()
```
brycemcanally, line 52: I think we should remove then close if retErr != nil in this defer.

src/internal/obj/local_client.go
```diff
@@ -1,126 +1,149 @@
...
 	if err != nil {
 		return err
 	}
-	defer func() {
-		if err := file.Close(); retErr == nil {
-			retErr = err
-		}
-	}()
-	_, err = io.Copy(w, file)
+	defer f.Close()
```
brycemcanally, line 68: Probably shouldn't discard the error.

src/internal/obj/local_client.go
```diff
@@ -1,126 +1,149 @@
...
+	_, err = io.Copy(w, f)
 	return err
 }
 
-func (c *localClient) Delete(_ context.Context, path string) (retErr error) {
-	defer func() { retErr = c.transformError(retErr, path) }()
-	return os.Remove(c.normPath(path))
+func (c *fsClient) Delete(ctx context.Context, name string) error {
+	err := os.Remove(c.finalPathFor(name))
+	if os.IsNotExist(err) {
```
brycemcanally, line 75: Is this normal object storage semantics?

src/internal/obj/local_client.go
```diff
@@ -1,126 +1,149 @@
...
+	return filepath.Join(c.rootDir, "objects", enc.EncodeToString([]byte(name)))
+}
+
+func (c *fsClient) init() error {
+	if err := os.RemoveAll(filepath.Join(c.rootDir, "staging")); err != nil {
+		return err
 	}
-	tracing.TagAnySpan(ctx, "err", err)
-	return true, nil
+	if err := os.MkdirAll(filepath.Join(c.rootDir, "staging"), 0o755); err != nil {
```
brycemcanally, line 131: Why is the extra `o` necessary in the file mode?

src/internal/obj/local_client.go
brendoncarroll: The behavior varies in terms of whether Delete is idempotent.  We enforce this behavior again in the uniform client.

src/internal/obj/local_client.go
brendoncarroll: The convention that numbers starting with 0 are interpreted as octal is ancient tribal knowledge that I don't expect everyone to have.  The `o` makes it obvious.

src/internal/obj/local_client.go
```diff
@@ -1,126 +1,163 @@
...
+func (c *fsClient) Put(ctx context.Context, name string, r io.Reader) (retErr error) {
+	staging := c.stagingPathFor(name)
+	final := c.finalPathFor(name)
+	f, err := os.Create(staging)
+	if err != nil {
 		return err
 	}
-	file, err := os.Create(fullPath)
-	if err != nil {
+	defer c.closeFile(&retErr, f)
```
brycemcanally, line 55: I think we want to close and remove it upon error.

src/internal/obj/local_client.go
brycemcanally: Yeah, but I don't think this is what people normally do in Go for setting the file mode. I normally see `0...`.

src/internal/obj/local_client.go
brendoncarroll: This will error if the Put can't be completed.  If the staging file can't be cleaned up, it will log the error, but the Put still succeeded so we shouldn't return an error.  We clean up the staging area during initialization, so we won't be leaking staging files indefinitely.

src/internal/obj/local_client.go
```diff
@@ -1,126 +1,163 @@
...
 
-func (c *localClient) Walk(_ context.Context, dir string, walkFn func(name string) error) error {
-	dir = c.normPath(dir)
-	fi, _ := os.Stat(dir)
-	prefix := ""
-	if fi == nil || !fi.IsDir() {
-		dir, prefix = filepath.Split(dir)
+func (c *fsClient) Exists(ctx context.Context, name string) (bool, error) {
+	_, err := os.Stat(c.finalPathFor(name))
+	if os.IsNotExist(err) {
```
brycemcanally, line 86: nit, we normally structure this with an outer if err != nil, then handle specific cases of that within the block (the `IsNotExist` in this case).

src/internal/obj/local_client.go
brycemcanally: I think we should make an attempt at removing the file from the staging area if an error occurs, rather than relying on a restart and having a leak in between.

brendoncarroll: Added `removeFile` similar to `closeFile` to handle this.

brycemcanally (approved):
lgtm

brendoncarroll:
Going to merge this.  CI is being flaky, but the failing test (`TestPFS/DeleteDir`) is passing locally.

PR #6567, opened 2021-07-20
Combine jobset and commitset pachctl commands into 'list job' and 'list commit'

Tryneus:
After internal discussions, this PR removes references to "commitset" and "jobset" from pachctl and instead adds several new modes for `pachctl list job` and `pachctl list commit`:

* `pachctl list job` will now group all "sub-jobs" by their ID to have one row per "job", similar to how `pachctl list jobset` worked previously
* `pachctl list job --expand` will break out the "jobs" into one line per "sub-job", similar to how `pachctl list job` worked previously, with various filters available
* `pachctl list job <job-id>` will list the individual "sub-jobs" in a "job" using the same format as `pachctl list job --expand`

* `pachctl list commit` will now group all "sub-commits" by their ID to have one row per "commit", similar to how `pachctl list commitset` worked previously.  Before this, the positional argument for `pachctl list commit` was required, so it is now optional.
* `pachctl list commit --expand` will break out the "commits" into one line per "sub-commit", grouped by commit-id and ordered by the commitset changed timestamp
* `pachctl list commit <commit-id>` will list the individual "sub-commits" in a "commit" using the same format as `pachctl list commit --expand`
* `pachctl list commit <repo>` will have the same behavior as before, listing the individual "sub-commits" in a repo with various filters available.


jdoliner (approved):
This all lgtm

PR #6569, opened 2021-07-20
Improve PFS collection error messages.

PFedak:
Suppress `.user` suffix by changing the underlying keys to match the
pretty-printed string forms, and rephrase so the plural/collection name
is more natural.

### DIFF
--- a/src/internal/collection/postgres_collection.go
+++ b/src/internal/collection/postgres_collection.go
@@ -27,7 +27,10 @@ type postgresCollection struct {
 	listener PostgresListener
 	template proto.Message
 	indexes  []*Index
+	keyGen   func(interface{}) (string, error)
 	keyCheck func(string) error
+	notFound func(interface{}) string
+	exists   func(interface{}) string
 }
 
 func indexFieldName(idx *Index) string {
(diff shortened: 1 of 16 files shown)


Tryneus (approved):
LGTM

src/internal/collection/postgres_collection.go
```diff
@@ -42,16 +45,45 @@ type model struct {
...
+	}
+}
+
+func WithKeyGen(gen func(interface{}) (string, error)) Option {
+	return func(c *postgresCollection) {
+		c.keyGen = gen
+	}
+}
+
+func WithExistsMessage(convert func(interface{}) string) Option {
```
Tryneus, line 62: nit: `convert` is probably better named `format` now

src/internal/collection/postgres_collection.go
```diff
@@ -151,14 +183,44 @@ func (c *postgresCollection) get(ctx context.Context, key string, q sqlx.Queryer
...
+	}
+	rawKey, err := c.keyGen(key)
+	if err != nil {
+		return err
+	}
+	err = query(rawKey)
+	if err != nil {
+		var notFound ErrNotFound
+		var exists ErrExists
+		if c.notFound != nil && errors.As(err, &notFound) {
```
Tryneus, line 211: I think you could get rid of this bit of complexity if the error lazily does the formatting (like instead of storing a custom message, store a reference to the formatter, then do this logic in the `ErrNotFound.Error` call).  And then always set the formatter to `c.notFound` or `c.exists` whenever those error types are constructed.

PFedak: hmm, I'm not sure where to pass the proto in, then

Tryneus: I see, this is ~probably fine then.

Tryneus (approved):

PR #6578, opened 2021-07-21
Bump go version for CI to 1.16

actgardner:
The CircleCI VM comes with Go 1.15, but we want to upgrade dependencies and write code that depends on a newer version. Install 1.16.6 explicitly in CI instead.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,7 @@
 run:
   timeout: 5m
+  skip-dirs:
+    - cached-deps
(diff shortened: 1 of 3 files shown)


brendoncarroll (approved):
I have this in my object client branch too.

actgardner:
@brendoncarroll cool, yeah not trying to duplicate work here, it was a simple change to get in and unblock the dex upgrade.

PR #6580, opened 2021-07-21
Make deploy target required

chainlink:
change to storageBackend for local backend related values

### DIFF
--- a/etc/helm/pachyderm/templates/pachd/deployment.yaml
+++ b/etc/helm/pachyderm/templates/pachd/deployment.yaml
@@ -71,13 +71,13 @@ spec:
           #value:
         - name: STORAGE_BACKEND
           value: {{ include "pachyderm.storageBackend" . | quote }}
-          {{- if eq .Values.deployTarget "LOCAL" }}
+          {{- if eq (include "pachyderm.storageBackend" . ) "LOCAL" }}
         - name: STORAGE_HOST_PATH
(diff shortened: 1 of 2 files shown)


etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -71,13 +71,13 @@ spec:
...
+          {{- if eq (include "pachyderm.storageBackend" . ) "LOCAL" }}
         - name: STORAGE_HOST_PATH
           value: {{ .Values.pachd.storage.local.hostPath | default $randHostPath }}pachd
           {{- end }}
         - name: WORKER_IMAGE
           value: "{{ .Values.worker.image.repository }}:{{ default .Chart.AppVersion .Values.pachd.image.tag }}"
-        {{- if and (eq .Values.deployTarget "LOCAL") .Values.pachd.storage.local.requireRoot }}
+        {{- if and (eq ( include "pachyderm.storageBackend" . ) "LOCAL") .Values.pachd.storage.local.requireRoot }}
         - name: WORKER_USES_ROOT
           value: "True"
```
jrockway, line 82: Never really thought about it but it's weird that "True" is the value, and not "true".

jrockway (approved):
WORKER_USES_ROOT probably shouldn't even be an option.  We should look into the reasoning behind that flag... you don't need to be root to read files on a disk, have read many files while not being root ;)

PR #6586, opened 2021-07-22
[Doc] [ci-skip]  Removal of datum merge

nadegepepin:


doc/docs/master/concepts/pipeline-concepts/datum/glob-pattern.md
```diff
@@ -4,29 +4,21 @@ Defining how your data is spread among workers is one of
...
-The files and directories that match the
-glob pattern constitute the [datums](https://docs.pachyderm.com/latest/concepts/pipeline-concepts/datum/)
-that will be processed by the worker(s) that run your pipeline code.
-
-!!! Important
-        You must **configure a glob pattern for each PFS input** of a [pipeline specification](). 
+define data distribution. 
 
 !!! Note
      The Pachyderm's concept of glob patterns is similar to the Unix glob patterns.
```
brycemcanally, line 10: I would change `The Pachyderm's` to just `Pachyderm's`. Also `the Unix` to `Unix`.

doc/docs/master/concepts/pipeline-concepts/datum/glob-pattern.md
```diff
@@ -4,29 +4,21 @@ Defining how your data is spread among workers is one of
...
      For example, the `ls *.md` command matches all files with the
      `.md` file extension.
 
-TODO: I would just remove this line, the examples below seem to be what we want. Also, I would be very clear that the below are just some examples, globs have much more flexibility than what is shown in the examples below.
-In Pachyderm, the `/` and `*` indicators are most
-commonly used globs.
+The glob is applied to given branches of all the directories / files of each pipeline's [PFS inputs](../../../../reference/pipeline_spec/#pfs-input).
```
brycemcanally, line 14: I would probably reword this. There aren't branches of directories / files, the branches have commits which can have directories / files. You could say `The glob pattern is applied to all of the directories / files in the branch specified by the PFS input. The directories / files that match are the datums that will be processed...`.

doc/docs/master/concepts/pipeline-concepts/datum/glob-pattern.md
```diff
@@ -179,25 +172,24 @@ You can use the `pachctl list datum -f <my_pipeline_spec.json>` command to previ
     ```
 
 ### Running list datum on a past job 
-You can use the `pachctl list datum <job_number>` command to check the datums processed by a given job.
+You can use the `pachctl list datum <pipeline>@<job_number>` command to check the datums processed by a given job.
```
brycemcanally, line 175: `job_number` -> `job_id`

doc/docs/master/concepts/pipeline-concepts/datum/index.md
```diff
@@ -5,63 +5,64 @@
...
-A datum is the smallest indivisible unit of computation within a
-job.
+## Definition
+A datum is **the smallest indivisible unit of computation within a
+job**.
 A job can have one, many or no datums. Each datum is processed
-independently with a single execution of the user code and
-then the results of all the datums are merged together to
+independently with a single execution of the user code on one of
+the pipeline worker nodes.
```
brycemcanally, line 13: `nodes` -> `pods`. This distinction is important because there can be multiple worker pods (kubernetes pods) on a single kubernetes node.

doc/docs/master/concepts/pipeline-concepts/datum/index.md
```diff
@@ -5,63 +5,64 @@
...
+The results of all the datums are then combined together to
```
brycemcanally, line 14: `The results of all the datums` -> `The files output by all of the datums`. I'm being kind of nitpicky, but I think it is good to be very clear about it being the files output when processing the datum.

doc/docs/master/concepts/pipeline-concepts/datum/index.md
```diff
@@ -5,63 +5,64 @@
...
+## Glob Pattern and Data distribution
+Think of datums as a way to **divide your input data** 
+and **distribute processing workloads**.
+They are instrumental in optimizing your pipeline performance.
 
-A datum defines the input data. An input can take one or multiple
-repositories. Pachyderm has the following types of inputs that
-combine multiple repositories:
+You define how your data is spread among workers by
+**specifying a [glob pattern](glob-pattern.md) for each input** of your pipeline.
```
brycemcanally, line 23: `input` -> `PFS input`

doc/docs/master/concepts/pipeline-concepts/datum/index.md
```diff
@@ -5,63 +5,64 @@
...
+Think of datums as a way to **divide your input data** 
+and **distribute processing workloads**.
+They are instrumental in optimizing your pipeline performance.
 
-A datum defines the input data. An input can take one or multiple
-repositories. Pachyderm has the following types of inputs that
-combine multiple repositories:
+You define how your data is spread among workers by
+**specifying a [glob pattern](glob-pattern.md) for each input** of your pipeline. 
+The data in the `input` of your pipeline is "glob-ed" or "broken down" in datums,
```
brycemcanally, line 24: I think it would be better if we had one section for PFS inputs that included all of the glob information / examples. Glob patterns only apply to PFS inputs, and should only really be discussed in that context imo.

brycemcanally:
A couple higher level changes I would make is being careful about re-defining concepts in multiple places instead of referencing one definition (I see multiple definitions of datums and globs) and also appropriately nesting the concepts. The most obvious example of the latter, which I mentioned in a comment, is the frequent references to glob patterns as determining the datums for a job. Although it is true that it is pretty common for people to just use PFS inputs which use a glob pattern to determine the datums, it is not correct in general to say that the datums in a job are based on glob patterns. The datums that a PFS input will produce are based on a glob, but if the input for a pipline is a cross, join, union, etc, that is not the case. The datums will be determined based on the composition of inputs, which in general will be PFS inputs at the lowest level. Having a PFS input section that covers the globs then describing datums for a job as being based on the pipeline input would be a more correct description.

doc/docs/master/concepts/pipeline-concepts/datum/relationship-between-datums.md
```diff
@@ -38,127 +29,84 @@ and data generation workflows.
 For more information about glob patterns, see
 [Glob Pattern](glob-pattern.md).
 
-Disregarding of how many datums you define and how many
+### 2- **Transformation**
+The pipeline **uses your code to process the datums**.
+
+!!! Note "Reminder"
+        The output produced by a pipeline's job is written to an output repo of the same name. However, you do not need to specify this name explicitly; you write your output to `/pfs/out`. 
+### 3- **Creation of output datums**
```
brycemcanally, line 37: I'm not sure I understand this section. There isn't really a notion of input / output datums, just the datums for the job. The datums have input files associated with them, and the user code may output files after processing those input files. Also, just to be clear, the number of input files for a datum may not be the same as the number of files output by the user code.

doc/docs/master/concepts/pipeline-concepts/datum/relationship-between-datums.md
```diff
@@ -38,127 +29,84 @@ and data generation workflows.
...
 
-In the diagram above, Pachyderm appends the file `1` from the `datum 2`
-to the file `1` in the final commit, deletes the file `2` from `datum 2`,
-overwrites the old part from `datum 2` in file `3`  with a new version,
-and creates a new output file `4`.
-
-Similarly, if you have multiple files in your input datum, Pachyderm might
-write them into multiple files in output datums that are later merged into
-files with the same file path.
+## Overwriting files
```
brycemcanally, line 96: This seems like it belongs in a different doc, probably with other PFS file operations.

doc/docs/master/concepts/pipeline-concepts/datum/relationship-between-datums.md
```diff
@@ -38,127 +29,84 @@ and data generation workflows.
...
 
+Pachyderm enables you to appends files as
+well by using the `--appends` flag. The order of processing
+is not guaranteed, and all datums are processed randomly.
+For more information, see [File](../../data-concepts/file.md).
+ 
 ## Note: Data persistence between datums
-
-Pachyderm only controls and wipes the /pfs directories between datums. If scratch/temp space is used during execution, the user needs to be careful to clean that up. Not cleaning temporary directories may cause unexpected bugs where one datum accesses temporary files that were previously used by another datum!
+Pachyderm only controls and wipes the `/pfs` directories between datums. If scratch/temp space is used during execution, the user needs to be careful to clean that up. Not cleaning temporary directories may cause unexpected bugs where one datum accesses temporary files that were previously used by another datum!
```
brycemcanally, line 112: :+1:

doc/docs/master/concepts/pipeline-concepts/datum/relationship-between-datums.md
nadegepepin: I removed the term `output datum` entirely since it does not relate to any existing concept in Pachyderm. Changed for `output files`.

doc/docs/master/concepts/pipeline-concepts/datum/relationship-between-datums.md
nadegepepin: Noted - This belongs to the file.md page. I still want to mention the default 'overwrite' behavior but then point to the full info on the appropriate page.

nadegepepin:
> A couple higher level changes I would make is being careful about re-defining concepts in multiple places instead of referencing one definition (I see multiple definitions of datums and globs) and also appropriately nesting the concepts. The most obvious example of the latter, which I mentioned in a comment, is the frequent references to glob patterns as determining the datums for a job. Although it is true that it is pretty common for people to just use PFS inputs which use a glob pattern to determine the datums, it is not correct in general to say that the datums in a job are based on glob patterns. The datums that a PFS input will produce are based on a glob, but if the input for a pipline is a cross, join, union, etc, that is not the case. The datums will be determined based on the composition of inputs, which in general will be PFS inputs at the lowest level. Having a PFS input section that covers the globs then describing datums for a job as being based on the pipeline input would be a more correct description.



> A couple higher level changes I would make is being careful about re-defining concepts in multiple places instead of referencing one definition (I see multiple definitions of datums and globs) and also appropriately nesting the concepts. The most obvious example of the latter, which I mentioned in a comment, is the frequent references to glob patterns as determining the datums for a job. Although it is true that it is pretty common for people to just use PFS inputs which use a glob pattern to determine the datums, it is not correct in general to say that the datums in a job are based on glob patterns. The datums that a PFS input will produce are based on a glob, but if the input for a pipline is a cross, join, union, etc, that is not the case. The datums will be determined based on the composition of inputs, which in general will be PFS inputs at the lowest level. Having a PFS input section that covers the globs then describing datums for a job as being based on the pipeline input would be a more correct description.

I have made the following changed that should address the pbm of redundancy as well as the need of a Pipeline Input section containing the PFS input and Glob Pattern info as well as Group, Join, etc...

1- The **Datum** Menu on the left now has its own **Pipeline Input** section containing the Glob Pattern page as well as the group, cross, union, and joins.
2- Additionnally, I have removed the mention of Glob Pattern everywhere else to point to the Pipeline Input section instead.

doc/docs/master/concepts/pipeline-concepts/datum/glob-pattern.md
```diff
@@ -35,14 +27,15 @@ Let's list the glob patterns at your disposal. We will later illustrate their us
 | `/*/*`| Pachyderm processes **each filesystem object in each subdirectory as a separate datum**.|
 | `/**` | Pachyderm processes **each filesystem object in all directories and subdirectories as a separate datum**.|
 
-TODO: I would just say subset of the files / directories.
```
brycemcanally, line 38: This TODO wasn't addressed.

doc/docs/master/concepts/pipeline-concepts/datum/index.md
```diff
@@ -5,63 +5,77 @@
...
-combine multiple repositories:
+You define how your data is spread among workers by
+**specifying [pipeline inputs](#pipeline-inputs)** for your pipeline 
+in its pipeline specification file. 
 
-TODO: I would focus more on these inputs being a composition of other inputs rather than
-referring to branches being combined in a certain way. It seems like the PFS input should be mentioned first,
-then the below inputs could be described as a way of composing inputs with a PFS input usually being the most
-primitive input. 
+Based on this specification file, the data in the `input`
```
brycemcanally, line 25: We should remove this or replace it with just saying the input describes how to turn the input data into datums. Describing the process is leaking details that are not generally correct (the globbing only applies to PFS inputs). Also, I would not consider globbing in a PFS input as breaking down the data.

doc/docs/master/concepts/pipeline-concepts/datum/relationship-between-datums.md
```diff
@@ -1,164 +1,86 @@
...
-
-If you decide to create a new commit and overwrite the file in `datum 2`,
-Pachyderm detects three datums. Because `datum 1` and `datum 3` are
-unchanged, it skips processing these datums. Then, Pachyderm detects
-that something has changed in `datum 2`. Pachyderm is unaware of any
-details of the change. Therefore, it processes the whole `datum 2`
-and outputs the files `1`, `3`, and `4`. Then, Pachyderm merges
-these datums to create the following final result:
+### 1- **Creation of input datums** 
+In this stage, Pachyderm **breaks
```
brycemcanally, line 13: Calling it breaking isn't really correct. I would just say `Pachyderm creates datums based on the input data according to the...`

doc/docs/master/concepts/pipeline-concepts/datum/relationship-between-datums.md
```diff
@@ -1,164 +1,86 @@
...
+unchanged, it skips processing these datums. Then, Pachyderm detects
+that something has changed in `datum 2`. Pachyderm is unaware of any
+details of the change; therefore, it processes the whole `datum 2`
+and outputs 3 files. Then, the following pipeline aggregates
+these data to create the following final result:
+
+![Map Reduce](../../../images/parallel_data_processing_following_commit.png)
+
+## Incrementality 
+In Pachyderm, glob patterns are applied to the entire input,
```
brycemcanally, line 68: Shouldn't mention glob patterns here.

nadegepepin: Removed `glob patterns are applied to the entire input,`

doc/docs/master/concepts/pipeline-concepts/datum/glob-pattern.md
brycemcanally: I was referring to the section below, not above, with this TODO. The change you made above probably makes sense though. File / directory is more precise than filesystem object.

brycemcanally (approved):
one minor change, then lgtm.

PR #6587, opened 2021-07-22
Revert "Save cache after build and before tests (#6540)"

nitinjainsj:
This reverts commit 28a586e0a216da418f2b7ea52dd5d674c9f1919b.

I am not sure why this change caused a higher failure rate on CI for PPS4 bucket. For now I am going to revert this.


nitinjainsj:
This change made no difference -- removing the revert.

PR #6590, opened 2021-07-22
Change pachctl stop job and wait job to accept job or subjob id

Tryneus:
Following #6567, the 'jobset' concept has been renamed to a 'job', so it would make sense for `stop job` and `wait job` to accept a top-level job object (not constrained by a pipeline).  This changes those commands to accept both formats and do the right thing in either case.  To have `wait job` not have a polymorphic output, it now always outputs a table (instead of the details view it had previously).  There are a few other tangentially-related bug fixes (see comments).

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -16,7 +16,7 @@
     {
       "label": "launch-dev",
       "type": "shell",
-      "command": "eval $(minikube docker-env --shell bash); yes | pachctl undeploy; LAUNCH_DEV_ARGS='--cluster-deployment-id=dev' make launch-dev",
+      "command": "eval $(minikube docker-env --shell bash); make clean-launch; LAUNCH_DEV_ARGS='pachd.clusterDeploymentID=dev' make launch-dev",
       "problemMatcher": []
     },
     {
--- a/src/internal/uuid/uuid.go
+++ b/src/internal/uuid/uuid.go
@@ -42,4 +42,4 @@ func IsUUIDWithoutDashes(s string) bool {
 // Because we use UUIDv4, the 13th character is a '4'.
 // Moreover, a UUID can only contain "hexadecimal" characters,
 // lowercase here.
-var uuidWithoutDashesRegexp = regexp.MustCompile("[0-9a-f]{12}4[0-9a-f]{19}")
+var uuidWithoutDashesRegexp = regexp.MustCompile("^[0-9a-f]{12}4[0-9a-f]{19}$")
--- a/src/server/pps/cmds/cmds.go
+++ b/src/server/pps/cmds/cmds.go
@@ -102,34 +102,58 @@ If the job fails, the output commit will not be populated with data.`,
 	shell.RegisterCompletionFunc(inspectJob, shell.JobCompletion)
 	commands = append(commands, cmdutil.CreateAlias(inspectJob, "inspect job"))
 
+	writeJobInfos := func(out io.Writer, jobInfos []*pps.JobInfo) error {
+		if raw {
+			e := cmdutil.Encoder(output, out)
+			for _, jobInfo := range jobInfos {
+				if err := e.EncodeProto(jobInfo); err != nil {
+					return err
+				}
+			}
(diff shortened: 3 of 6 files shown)


.vscode/tasks.json
```diff
@@ -16,7 +16,7 @@
     {
       "label": "launch-dev",
       "type": "shell",
-      "command": "eval $(minikube docker-env --shell bash); yes | pachctl undeploy; LAUNCH_DEV_ARGS='--cluster-deployment-id=dev' make launch-dev",
+      "command": "eval $(minikube docker-env --shell bash); make clean-launch; LAUNCH_DEV_ARGS='pachd.clusterDeploymentID=dev' make launch-dev",
```
Tryneus, line 19: Moved the windows launch fixes from #6556 because that won't be merged for a while and this is impacting my development flow.  cc @chainlink in case you have opinions on these changes because it's related to the helm change.

src/internal/uuid/uuid.go
```diff
@@ -42,4 +42,4 @@ func IsUUIDWithoutDashes(s string) bool {
 // Because we use UUIDv4, the 13th character is a '4'.
 // Moreover, a UUID can only contain "hexadecimal" characters,
 // lowercase here.
-var uuidWithoutDashesRegexp = regexp.MustCompile("[0-9a-f]{12}4[0-9a-f]{19}")
+var uuidWithoutDashesRegexp = regexp.MustCompile("^[0-9a-f]{12}4[0-9a-f]{19}$")
```
Tryneus, line 45: This regex wasn't anchored, so it would return `true` for something like `repo@<uuid>`, which is obviously wrong.  This _shouldn't_ impact any existing code, but it's possible someone is relying on this broken behavior.

src/server/pps/cmds/cmds.go
```diff
@@ -190,7 +214,7 @@ $ {{alias}} -p foo -i bar@YYY`,
 			}
 			defer client.Close()
 
-			if raw && output != "" {
+			if !raw && output != "" {
```
Tryneus, line 217: This was totally wrong before (my fault).

src/server/pps/pretty/pretty.go
```diff
@@ -80,10 +81,9 @@ func PrintJobSetInfo(w io.Writer, jobSetInfo *ppsclient.JobSetInfo, fullTimestam
 	var created *types.Timestamp
 	var modified *types.Timestamp
 	for _, job := range jobSetInfo.Jobs {
-		switch job.State {
-		case ppsclient.JobState_JOB_SUCCESS:
+		if job.State == ppsclient.JobState_JOB_SUCCESS {
 			success++
-		case ppsclient.JobState_JOB_FAILURE:
+		} else if ppsutil.IsTerminal(job.State) {
```
Tryneus, line 86: This wasn't handling the `KILLED` state correctly, so tried to future-proof it a bit for arbitrary failed states.

src/server/transaction/server/driver.go
```diff
@@ -168,6 +168,8 @@ func (d *driver) runTransaction(txnCtx *txncontext.TransactionContext, info *tra
 			err = directTxn.DeleteBranch(request.DeleteBranch)
 		} else if request.UpdateJobState != nil {
 			err = directTxn.UpdateJobState(request.UpdateJobState)
+		} else if request.StopJob != nil {
+			err = directTxn.StopJob(request.StopJob)
```
Tryneus, line 172: `StopJob` wasn't useable in a transaction because this was missing.

PFedak: Oh hmm, do we want `stop job` in general to use the current transaction?

PFedak (approved):
LGTM

src/server/transaction/server/driver.go
Tryneus: Oh, that's a good point, but it comes up against the same thing that makes `pachctl start commit` with multiple branches tough to do in a transaction - I'll handle that in the PR for that feature.

PR #6593, opened 2021-07-22
Make cron ticks atomic

PFedak:
RunCron was using modify file clients to add cron ticks to all input
repos, but modify file doesn't attempt to use the active transaction
since it cannot by its nature be performed entirely in a transaction.
Instead, we create filesets for all cron inputs, and then add them all
to commits in the same transaction.

In addition, it's now easier to create a fileset internally without a
gRPC. The fileset creation in CreatePipeline has also moved over to use
this implementation.


brendoncarroll:
I think this duplicates a lot of the functionality of `AddFileset` and `CreateFileset`.  AddFileset has an in-transaction counterpart.

brendoncarroll:
AddFilesetInTransaction
https://github.com/pachyderm/pachyderm/blob/master/src/server/pfs/server/api_server.go#L718
CreateFileSet
https://github.com/pachyderm/pachyderm/blob/master/src/server/pfs/server/api_server.go#L683

src/server/pps/server/monitor.go
```diff
@@ -420,15 +437,9 @@ func (m *ppsMaster) makeCronCommits(ctx context.Context, in *pps.Input) error {
...
-	commitInfo, err := pachClient.InspectCommit(in.Cron.Repo, "master", "")
-	if err != nil {
+	// finish any open commit on the branch
+	if err := pachClient.FinishCommit(in.Cron.Repo, "master", ""); err != nil && !pfs_server.IsCommitFinishedErr(err) {
 		return err
-	} else if commitInfo != nil && commitInfo.Finished == nil {
-		// and if there is, delete it
-		if err = pachClient.SquashCommitSet(commitInfo.Commit.ID); err != nil {
-			return err
-		}
```
PFedak, line 431: some part of this block is wrong - the if condition isn't sufficient to avoid losing data, but it didn't seem clear to me that we should squash the entire commitset, potentially deleting unrelated data, just because a user (accidentally?) opened a commit on the cron repo. This could use broader discussion, as maybe `squash` should have more safeguards even after we fix the data loss bug.

src/server/pps/server/api_server.go
```diff
@@ -2745,6 +2710,21 @@ func (a *apiServer) RunPipeline(ctx context.Context, request *pps.RunPipelineReq
 	return nil, errors.New("unimplemented")
 }
 
+func commitFilesetInTransaction(a pfsServer.APIServer, txnCtx *txncontext.TransactionContext, branch *pfs.Branch, filesetID string) (*pfs.Commit, error) {
```
brycemcanally, line 2713: Functions, variables, etc that reference file sets should be ...FileSet or fileSet... since file set is two words rather than one now.

src/server/pps/server/api_server.go
```diff
@@ -2745,6 +2710,21 @@ func (a *apiServer) RunPipeline(ctx context.Context, request *pps.RunPipelineReq
...
+	commit, err := a.StartCommitInTransaction(txnCtx, &pfs.StartCommitRequest{Branch: branch})
+	if err != nil {
+		return nil, err
+	}
+	if err := a.AddFileSetInTransaction(txnCtx, &pfs.AddFileSetRequest{Commit: commit, FileSetId: filesetID}); err != nil {
+		return nil, err
+	}
+	err = a.FinishCommitInTransaction(txnCtx, &pfs.FinishCommitRequest{Commit: commit})
```
brycemcanally, line 2721: The if could just be here (`if err := a.Finish...`).

src/server/pps/server/monitor.go
```diff
@@ -420,15 +433,9 @@ func (m *ppsMaster) makeCronCommits(ctx context.Context, in *pps.Input) error {
 		return err // Shouldn't happen, as the input is validated in CreatePipeline
 	}
 	pachClient := m.a.env.GetPachClient(ctx)
-	// make sure there isn't an unfinished commit on the branch
-	commitInfo, err := pachClient.InspectCommit(in.Cron.Repo, "master", "")
-	if err != nil {
+	// finish any open commit on the branch
+	if err := pachClient.FinishCommit(in.Cron.Repo, "master", ""); err != nil && !pfs_server.IsCommitFinishedErr(err) {
```
brycemcanally, line 437: We should probably remove this.

brycemcanally (approved):
lgtm, just a few minor changes.

PR #6595, opened 2021-07-23
Add monitoring to the database connection after we're certain it works

jrockway:
If db.Ping fails, we end up calling `prometheus.Register(sqlstats.NewStatsCollector(db))` more than once, which fails the second time.  This prevents the case where we call it the second time; we add the monitoring after Ping succeeds, so no error can cause the code to run again and try to monitor a second db object with the same stats collector name.

### DIFF
--- a/src/internal/serviceenv/service_env.go
+++ b/src/internal/serviceenv/service_env.go
@@ -291,7 +291,7 @@ func (env *NonblockingServiceEnv) initKubeClient() error {
 }
 
 func (env *NonblockingServiceEnv) initDirectDBClient() error {
-	return backoff.Retry(func() error {
+	if err := backoff.Retry(func() error {
 		db, err := dbutil.NewDB(
 			dbutil.WithHostPort(env.config.PostgresHost, env.config.PostgresPort),
 			dbutil.WithDBName(env.config.PostgresDBName),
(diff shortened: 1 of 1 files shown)


brendoncarroll (approved):
LGTM, but this may also be a good place to panic.

jrockway:
I looked at this more carefully this afternoon, and moved the monitoring initialization to outside of the retry loop.  Dunno how good you feel about `if err := ...` that extends over multiple lines, but other parts of the codebase do it and I personally like it, but realize it's one of my quirks and maybe not enjoyable for others.  Happy to change it up.

I think we also had the naming inverted; the DB connection that connected to PostgresHost was called "pg_bouncer" and the DB connection that connected to PGBouncerHost was called "postgres".  I renamed the Postgres one to "direct" and the pgbouncer one to "pg_bouncer".

Finally, I made the monitoring registration just return an error.  I don't think we need to panic there; the number of codepaths that lead to that point are minimal and the error should be clearly identifiable.  I don't feel great about crashing because we couldn't init monitoring, but it would be annoying to have a bunch of deployed code and only notice that monitoring is broken when we really needed the data.  If CI passes without issue, I'm happy with this new arrangement.

jrockway:
On Slack, @brendoncarroll and I discussed possibly making the Prometheus registry a per-serviceenv object.  That is a good idea, but we need to do some additional plumbing to support that.  I'm submitting this with a plan to clean that up in the future, post 2.0; so for now, we just warn when a second database connection is created and not monitored, rather than panicking.

PR #6600, opened 2021-07-23
Remove non-greedy postgres list behavior.

PFedak:
Even in the read only case, row iteration takes up a connection as long
as the row is open. Any nested query will require another, making it
possible to deadlock if all connections are waiting on outer queries.
List pipeline was particularly egregious here because it opened up to
twenty subqueries.


Tryneus:
I don't think this is feasible for stuff like listing commits, which could be hundreds of thousands of rows.

PR #6602, opened 2021-07-23
Don't bother renewing the spec fileset

PFedak:


brendoncarroll:
There may be more involved to fix the RunCron test, but dropping the `RenewFileset` can go in on it's own.

PR #6605, opened 2021-07-25
Fix mount for postgres

chainlink:


robert-uhl (approved):
LGTM. I see this in [the image docs](https://hub.docker.com/_/postgres):

> The default is `/var/lib/postgresql/data`.

I don’t see any reason to change the default `PGDATA`.

PR #6613, opened 2021-07-26
Asynchronous finish commit / path validation and job management refactor

brycemcanally:
This PR implements asynchronous finish commit / path validation. Asynchronous finish commit / path validation is described here: https://docs.google.com/document/d/1rA1YUJF2S5hPJksosoa9pZZ_DOA-4xOWqtMfFJkYzDk. Also, the job management code was refactored to simplify the implementation and to start making the changes necessary to improve job resiliency (transient sidecar failures / etcd crashing). There is still more to do to improve job resiliency which has been mentioned in a few TODOs I added, but a safe fallback for now is to kill the master which will rebuild the unfinished jobs from scratch.

Miscellaneous:
- Commit size is now readily available after a commit is finished.
- All of the trigger tests are now working since we now collect size info when the commit is finished. The semantics of triggers are still a bit questionable, but they will have the same behavior as 1.x, which is probably fine for 2.0.
- Lazily get parent file set ID in modify file. Previously, we would get the parent file set ID for any modify file operation, which would potentially kick off a compaction that may be unnecessary if a directory deletion is not involved in the modify file operation.
- The finishing state is not considered terminal in PPS. A job will be marked successful after the commit has been finished by the background process. The background process will only transition the job state if it is in finishing, which means killing / stopping a job will result with the job state being terminal first, then the commit state eventually becoming finished.
- The job stopper run call was missing from the transaction finish, so that was added.
- Fixed a few tests with race conditions that became more obvious once the performance characteristics of finish commit changed.
- Fixed a bug with job timeouts which wasn't being caught by the job timeout test (We were referencing the started field of a job info that was not updated to reflect the state after it was started. Jobs enter the registry in the created state, then are moved to started by the registry)

TODO:
- Errored commits should have a reason field explaining the error. It might make sense to just have an error field that is a string.
- TestListDeletedDatums and TestRewindCrossPipeline need to be updated since duplicate file paths from different datums are no longer allowed.
- Distribute the background processing of the finished commits. Although the compaction is distributed, I still think we will hit a commit finishing bottleneck as is since there is only one goroutine handling the validation and management of a compaction when a compaction is needed.
- Client side ergonomics for finishing state / invalid paths and commit sizes:
  - Remove tag from list datum and expose datum id through environment variable in user code.
  - Make pachctl finish commit synchronous?
  - Ensure that details are being handled correctly (no details until commit is finished) and only use upper bound when size is not available. Also, update <= symbol to something more readable.


src/client/pfs.go
```diff
@@ -181,14 +181,19 @@ func (c APIClient) StartCommitParent(repoName string, branchName string, parentB
 // FinishCommit ends the process of committing data to a Repo and persists the
 // Commit. Once a Commit is finished the data becomes immutable and future
 // attempts to write to it with PutFile will error.
-func (c APIClient) FinishCommit(repoName string, branchName string, commitID string) error {
+func (c APIClient) FinishCommit(repoName string, branchName string, commitID string) (retErr error) {
+	defer func() { retErr = grpcutil.ScrubGRPC(retErr) }()
```
brendoncarroll, line 185: ScrubGRPC prevents inspecting structured errors.  I think we should stop using it.

src/internal/storage/fileset/option.go
```diff
@@ -58,9 +58,9 @@ func WithRenewal(ttl time.Duration, r *renew.StringSet) UnorderedWriterOption {
 // WithParentID sets the parent fileset ID for the unordered writer.
 // This is used for converting directory deletions into a set of point
 // deletes for the files contained within the directory.
-func WithParentID(parentID *ID) UnorderedWriterOption {
+func WithParentID(getParentID func() (*ID, error)) UnorderedWriterOption {
```
brendoncarroll, line 61: The comment needs to be updated.  Maybe have a type `type FilesetThunk =  func() (*FilesetID, error)` which would be more self explanatory.

src/internal/storage/fileset/unordered_writer.go
```diff
@@ -136,8 +136,12 @@ func (uw *UnorderedWriter) Delete(p, tag string) error {
 	if IsDir(p) {
 		uw.buffer.Delete(p, tag)
 		var ids []ID
-		if uw.parentID != nil {
-			ids = []ID{*uw.parentID}
+		if uw.getParentID != nil {
```
brendoncarroll, line 139: So no getParentID means no parent?

src/server/pachyderm_test.go
```diff
@@ -9347,6 +9346,8 @@ func TestPipelineAutoscaling(t *testing.T) {
 }
 
 func TestListDeletedDatums(t *testing.T) {
+	// TODO(2.0 optional): Duplicate file paths from different datums no longer allowed.
+	t.Skip("Duplicate file paths from different datums no longer allowed.")
```
brendoncarroll, line 9350: It seems like this test will never pass if we changed the behavior?  Should we delete it?

src/server/pfs/server/api_server.go
```diff
@@ -184,19 +184,7 @@ func (a *apiServer) FinishCommit(ctx context.Context, request *pfs.FinishCommitR
 	defer func(start time.Time) { a.Log(request, response, retErr, time.Since(start)) }(time.Now())
 	if err := a.txnEnv.WithTransaction(ctx, func(txn txnenv.Transaction) error {
 		return txn.FinishCommit(request)
-	}, func(txnCtx *txncontext.TransactionContext) (string, error) {
-		// FinishCommit in a transaction by itself has special handling with regards
```
brendoncarroll, line 188: Is this comment just totally unnecessary now?

src/server/pfs/server/driver.go
```diff
@@ -523,7 +523,7 @@ func (d *driver) startCommit(
 			return nil, errors.Wrapf(err, "parent commit not found")
 		}
 		// fail if the parent commit has not been finished
-		if parentCommitInfo.Finished == nil {
+		if parentCommitInfo.Finishing == nil {
```
brendoncarroll, line 526: Why did this need to change?  Also the error message is not in sync.

src/server/pfs/server/driver_file.go
```diff
@@ -129,6 +133,12 @@ func (d *driver) openCommit(ctx context.Context, commit *pfs.Commit, opts ...ind
 	if err != nil {
 		return nil, nil, err
 	}
+	if commitInfo.Finishing != nil && commitInfo.Finished == nil {
+		_, err := d.inspectCommit(ctx, commit, pfs.CommitState_FINISHED)
```
brendoncarroll, line 137: So this is going to block, while the parent compacts.  I guess that's okay.

src/server/pfs/server/master.go
```diff
@@ -34,9 +42,122 @@ func (d *driver) master(ctx context.Context) {
...
+		if err := ev.Unmarshal(&key, commitInfo); err != nil {
+			return err
+		}
+		if commitInfo.Finishing == nil || commitInfo.Finished != nil {
+			return nil
+		}
+		return backoff.RetryUntilCancel(ctx, func() error {
+			id, err := d.getFileSet(ctx, commitInfo.Commit)
+			if err != nil {
+				if pfsserver.IsCommitNotFoundErr(err) {
```
brendoncarroll, line 71: I think we should check for non-transient errors in the other callback, below.  Everything in here returns errors, the second callback checks them for things we can't hope will disappear.

src/server/pfs/server/master.go
```diff
@@ -34,9 +42,122 @@ func (d *driver) master(ctx context.Context) {
...
+			}
+			// Compact the commit.
+			compactedId, err := d.compactor.Compact(ctx, []fileset.ID{*id}, defaultTTL)
+			if err != nil {
+				return err
+			}
+			if err := d.commitStore.SetTotalFileSet(ctx, commitInfo.Commit, *compactedId); err != nil {
+				return err
+			}
+			// Validate the commit.
```
brendoncarroll, line 84: I really think this should be split out as a separate method/function.

src/server/pfs/server/master.go
```diff
@@ -34,9 +42,122 @@ func (d *driver) master(ctx context.Context) {
...
+				}
+				prev = idx.Path
+				size += index.SizeBytes(idx)
+				return nil
+			}); err != nil {
+				return err
+			}
+			// Finish the commit.
+			return d.txnEnv.WithWriteContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
+				commitInfo.Finished = txnCtx.Timestamp
```
brendoncarroll, line 106: We should double check that this comes from postgres's `CURRENT_TIMESTAMP` and is not an artisanally crafted pachd originating timestamp.

src/server/pfs/server/master.go
```diff
@@ -34,9 +42,122 @@ func (d *driver) master(ctx context.Context) {
...
+				commitInfo.SizeBytesUpperBound = size
+				if commitInfo.Details == nil {
+					commitInfo.Details = &pfs.CommitInfo_Details{}
+				}
+				commitInfo.Details.SizeBytes = size
+				if err := d.commits.ReadWrite(txnCtx.SqlTx).Put(pfsdb.CommitKey(commitInfo.Commit), commitInfo); err != nil {
```
brendoncarroll, line 112: Do we have an update method anywhere, or do we always do a Get then Put?

src/server/pfs/server/master.go
```diff
@@ -34,9 +42,122 @@ func (d *driver) master(ctx context.Context) {
...
+				return d.triggerCommit(txnCtx, commitInfo.Commit)
+			})
+		}, backoff.NewInfiniteBackOff(), func(err error, _ time.Duration) error {
+			log.Errorf("error finishing commit %v: %v", commitInfo.Commit.ID, err)
+			return nil
+		})
+	}, watch.IgnoreDelete)
+}
+
+// finishAliasChildren will traverse the given commit's children, finding all
```
brendoncarroll, line 134: comment doesn't match

src/server/pfs/server/master.go
brendoncarroll: Especially since we will want a well defined begging and end in the code to point engineers to when they ask "what do we do in validation"

src/server/pfs/server/master.go
```diff
@@ -34,9 +42,122 @@ func (d *driver) master(ctx context.Context) {
...
+			fs, err := d.storage.Open(ctx, []fileset.ID{*compactedId})
+			if err != nil {
+				return err
+			}
+			var prev string
+			var size int64
+			if err := fs.Iterate(ctx, func(f fileset.File) error {
+				idx := f.Index()
+				if prev != "" && (idx.Path == prev || strings.HasPrefix(idx.Path, prev+"/")) {
+					commitInfo.Error = true
```
brendoncarroll, line 96: Is there any place to put all of the other useful information about what failed that is at our fingertips here?

src/server/pfs/server/testing/server_test.go
```diff
@@ -2611,6 +2639,7 @@ func TestPFS(suite *testing.T) {
 		require.Equal(t, BCommit, commitInfos[1].Commit)
 		require.Equal(t, CCommit, commitInfos[2].Commit)
 		require.Equal(t, DCommit, commitInfos[3].Commit)
+		<-done
```
brendoncarroll, line 2642: This seems like it is going to timeout in CI.  Maybe use a select with a context.

src/server/pfs/server/master.go
```diff
@@ -34,9 +42,122 @@ func (d *driver) master(ctx context.Context) {
...
+		})
 		return eg.Wait()
 	}, backoff.NewInfiniteBackOff(), func(err error, _ time.Duration) error {
 		log.Errorf("error in pfs master: %v", err)
 		return nil
 	})
 }
+
+func (d *driver) finishCommits(ctx context.Context) error {
+	return d.commits.ReadOnly(ctx).WatchF(func(ev *watch.Event) error {
```
brendoncarroll, line 56: Does this spit out everything at the start, and then diffs?  We need to make sure that if we miss an event, everything still gets compacted.

src/client/pfs.go
brycemcanally: Sure, but we should do that separate from this PR.

src/internal/storage/fileset/option.go
brycemcanally: I went ahead and updated the comment to setting a factory for the parent file set ID. We can tweak this more later in a separate PR if you would like.

src/internal/storage/fileset/unordered_writer.go
brycemcanally: Right.

src/server/pachyderm_test.go
brycemcanally: It just needs to be updated. It seems like a decent enough test for joins.

src/server/pfs/server/api_server.go
brycemcanally: The special case still applies, but it occurs in the background process that finishes the commit rather than here. There is a note about it where the commit is finished.

src/server/pfs/server/driver.go
brycemcanally: If you call finish commit, then start quickly, it would error if we did not allow you to start a commit until the parent was finished (as compared to finishing). Any reads on the commit would block until the parent is finished. I was thinking about keeping the error as finished commit for both finishing / finished in the interest of time.

src/server/pfs/server/driver_file.go
brycemcanally: Yeah, the alternative is what we have now which is that any call that is made before a separate call compacts will duplicate the work. This seems like the ideal model.

src/server/pfs/server/master.go
brycemcanally: Hmm, that seems a bit weird because we would have to bubble the error up and handle it outside of the function rather than just considering it a terminal condition local to the computation. Having the commit disappear is expected behavior for a squash commit.

src/server/pfs/server/master.go
brycemcanally: It is coming from pachd, I made a card for this change.

src/server/pfs/server/master.go
brycemcanally: I will make this an update.

src/server/pfs/server/master.go
brycemcanally: We need to. It was something I mentioned in the comment above. I'm not sure if we can get that in this PR by today though.

src/server/pfs/server/master.go
brycemcanally: Watch sets up the listen, does a get, then syncs the get with the notifications to ensure that events are not lost.

src/server/pfs/server/master.go
brycemcanally: I didn't for now because we would need to disambiguate errors reading the file set from errors associated with validation. The ergonomics for errored commits need work.

src/server/pfs/server/testing/server_test.go
brycemcanally: I put the close in a defer, so we should be good to go if we reach this point.

brendoncarroll (approved):

PR #6614, opened 2021-07-26
[Doc] [skip-ci] pachchtl deploy storage removal

nadegepepin:


chainlink:
I believe we can just remove this whole file, as it refers to commands that have been deprecated

PR #6524, opened 2021-07-13
(Breaking change) Rename dash to console

chainlink:

PR #6616, opened 2021-07-26
Config Pod with Helm

acohen4:
This is the based off this branch, followed by a rebase on master. @chainlink let me know if you would instead prefer if i just updated your branch directly.

### DIFF
--- a/etc/helm/pachyderm/templates/pachd/config-secret.yaml
+++ b/etc/helm/pachyderm/templates/pachd/config-secret.yaml
@@ -1,0 +1,79 @@
+{{- /*
+SPDX-FileCopyrightText: Pachyderm, Inc. <info@pachyderm.com>
+SPDX-License-Identifier: Apache-2.0
+*/ -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: pachyderm-config
+  namespace: {{ .Release.Namespace }}
+stringData:
+{{- if eq .Values.deployTarget "LOCAL" }}
+  # authConfig is the configuration for pachd as an OIDC client
+  authConfig: |
+    client_id: pachd 
+    client_secret: notsecret
+    issuer: http://localhost:30658/
+    localhost_issuer: true
+    require_email_verified: false
+    redirect_uri: http://pachd:1657/authorization-code/callback
+    scopes:
+    - email
+    - profile
+    - groups
+    - openid 
+{{- end }}
+
+  # clusterRoleBindings is a set of user -> role mappings to apply
+  clusterRoleBindings: |
+    robot:test:
+    - repoReader
+
+  # enterpriseClusters is the set of pachds covered by license service 
+  enterpriseClusters: |
+    - address: grpc://localhost:1650
+      id: localhost
+      secret: secret
+      user_address: grpc://localhost:30650
+      cluster_deployment_id: {{ include "pachyderm.dashSecret" . }}
+
+  # enterpiseConfig points the pachd to a license service (in this case itself)
+{{- if eq .Values.deployTarget "LOCAL" }}
+  enterpriseConfig: |
+    id: localhost
+    license_server: grpc://localhost:1650
+    secret: secret
+{{- end }}
+
+  # identityServiceConfig configures the OIDC provider
+{{- if eq .Values.deployTarget "LOCAL" }}
+  identityServiceConfig: |  
+    issuer: http://localhost:30658/
+    id_token_expiry: 1d
+{{- end }}
+
+{{- if eq .Values.deployTarget "LOCAL" }}
+  # idps is the set of Identity Providers to support for logging in
+  idps: |
+    - id: test
+      jsonConfig: '{"username": "admin", "password": "password"}'
+      name: test
+      type: mockPassword
+{{- end }}
+
+  license: {{ required "Enterprise license key required" .Values.pachd.enterpriseLicenseKey | quote }}
+
+{{- if eq .Values.deployTarget "LOCAL" }}
+  # oidcClients is the set of OIDC clients registered with the OIDC provider
+  oidcClients: |
+    - id: pachyderm
+      name: pachd
+      secret: oidcsecret
+      redirect_uris:
+      - http://pachd:1657/authorization-code/callback
+      trusted_peers:
+      - dash
+{{- end }}
+
+  # rootToken is the auth token used to communicate with the cluster as the root user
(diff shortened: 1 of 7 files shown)


etc/helm/pachyderm/values.yaml
```diff
@@ -402,3 +403,6 @@ cloudsqlAuthProxy:
     labels: {}
     # type specifies the Kubernetes type of the cloudsql auth proxy service.
     type: ClusterIP
+
+config:
```
chainlink, line 407: I think this can be removed?

etc/helm/pachyderm/values.yaml
```diff
@@ -195,6 +195,7 @@ pachd:
     #apiGrpcPort:
     #  expose: true
     #  port: 30650
+  enterpriseLicenseKey: ""
```
chainlink, line 198: You'll want to run `make schema` in `etc/helm` to bring the json schema up to date

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -0,0 +1,104 @@
...
+  #           "insecureSkipEmailVerified": true,
+  #           "insecureSkipIssuerCallbackDomainCheck": true
+  #       }
+  #     name: okta
+  #     type: oidc
+{{- end }}
+
+  license: {{ required "Enterprise license key required" .Values.pachd.enterpriseLicenseKey | quote }}
+
+  enterpriseSecret: "secret" # {{ randAlphaNum 32 | quote }}
```
chainlink, line 82: We'll probably want to treat this similarly to clusterDeploymentId (option to pass in or generate value?)

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -0,0 +1,104 @@
...
+    - openid 
+{{- end }}
+
+  # clusterRoleBindings is a set of user -> role mappings to apply
+  clusterRoleBindings: |
+    robot:test:
+    - repoReader
+
+  # enterpriseClusters is the set of pachds covered by license service 
+  enterpriseClusters: |
```
chainlink, line 33: we'll probably want to surround this with an 'is local' check

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -0,0 +1,104 @@
...
+  #           "redirectURI": "http://localhost:30658/callback",
+  #           "insecureEnableGroups": true,
+  #           "insecureSkipEmailVerified": true,
+  #           "insecureSkipIssuerCallbackDomainCheck": true
+  #       }
+  #     name: okta
+  #     type: oidc
+{{- end }}
+
+  license: {{ required "Enterprise license key required" .Values.pachd.enterpriseLicenseKey | quote }}
```
chainlink, line 80: As this stands, this will block creation of the helm chart, which I'm not sure is what we want. I think perhaps instead we should only run the config pod if the license is specified? In the future, we may automatically activate a free trial key, but we'll need a larger product discussion around this.

chainlink (changes requested):
a few notes, but otherwise looking good!

etc/helm/pachyderm/templates/pachd/config-secret.yaml
chainlink: (This is also why CI is currently failing)

acohen4: so our "local checks" here would become: `"is local" + "has enterprise license"`?

chainlink: Good question! I think a simpler path would be to surround the entire configs for the job and configmap with the check for enterprise license. (See how we use dash.enabled for example) That way, the config pod is not run at all if the ent key is not specified.

chainlink:
Overall, i think we're ok with the first cut of this targeting local and we can expand / parameterize to other deployTargets from there

chainlink:
Also, should this bump the dash sha as I believe it requires the new version?

etc/helm/pachyderm/values.yaml
```diff
@@ -38,7 +38,8 @@ dash:
   # podLabels specifies labels to add to the dash pod.
```
chainlink, line 38: Seems like perhaps a formatter got run on this file?

chainlink:
Looks like a small lint error being reported in the helm tests (looks like I originally used an old syntax, feel free to update!)

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -0,0 +1,110 @@
...
+      name: pachd
+      secret: oidcsecret
+      redirect_uris:
+      - http://localhost:30657/authorization-code/callback
+      trusted_peers:
+      - dash
+    {{- if .Values.dash.enabled }}
+    - id: dash
+      name: dash
+      secret: notsecret
```
chainlink, line 102: This should be `{{ include "pachyderm.dashSecret" . }}`

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -0,0 +1,110 @@
...
+kind: Secret
+metadata:
+  name: pachyderm-config
+  namespace: {{ .Release.Namespace }}
+stringData:
+{{ if eq .Values.deployTarget "LOCAL" }}
+  # authConfig is the configuration for pachd as an OIDC client
+  authConfig: |
+    client_id: pachd 
+    client_secret: oidcsecret
```
chainlink, line 16: Should probably have a note saying this should match the secret below under `oidcClients`

chainlink:
Also note https://github.com/pachyderm/pachyderm/pull/6524 has shipped, so dash references are now `console`

etc/helm/pachyderm/values.yaml
acohen4: > Also note #6524 has shipped, so dash references are now `console`

ah right, I'll pull in those changes

acohen4:
> Also note #6524 has shipped, so dash references are now `console`

ah right, I'll pull in those changes

etc/helm/pachyderm/values.yaml
acohen4: oof yes, will revert that

acohen4:
@chainlink I think we're getting there! could you give this another pass 🚀 ?

PR #6618, opened 2021-07-26
Fix additional help topics in base `pachctl` command

### LINKED ISSUE #6599: pachctl with --help or no args generates error in beta3
**What happened?**:
```
$pachctl
Access the Pachyderm API.

Environment variables:
  PACH_CONFIG=<path>, the path where pachctl will attempt to load your config.
  JAEGER_ENDPOINT=<host>:<port>, the Jaeger server to connect to, if PACH_TRACE
    is set
  PACH_TRACE={true,false}, if true, and JAEGER_ENDPOINT is set, attach a Jaeger
    trace to any outgoing RPCs.
  PACH_TRACE_DURATION=<duration>, the amount of time for which PPS should trace
    a pipeline after 'pachctl create-pipeline' (PACH_TRACE must also be set).

Error: template: top:13:47: executing "top" at <.IsHelpCommand>: can't evaluate field IsHelpCommand in type *cobra.Command
```

run `pachctl` in beta3

**Environment?**:
- If you deployed with helm, the values you used (`helm get values pachyderm`):
- Others:

Tryneus:
It looks like this was written against an old API that no longer works.  I will also backport this to 1.13.x when merged, as it looks like it was broken there as well (with slightly different outcomes).

Fixes #6599

### DIFF
--- a/src/server/cmd/pachctl/cmd/cmd.go
+++ b/src/server/cmd/pachctl/cmd/cmd.go
@@ -931,7 +931,7 @@ Commands by Action:{{range actions}}{{if .IsAvailableCommand}}
 Other Commands:{{range other}}{{if .IsAvailableCommand}}
   {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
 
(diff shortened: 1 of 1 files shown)


acohen4 (approved):
LGTM

PR #6642, opened 2021-07-29
Fix miscutil.WithPipe not returning errors.

brendoncarroll:
miscutil.WithPipe was not returning errors because CloseWithError always returns nil, and that was being passed to the errgroup.  So errgroup.Wait always returned nil, even in the case of failure.

### DIFF
--- a/src/internal/miscutil/miscutil.go
+++ b/src/internal/miscutil/miscutil.go
@@ -19,7 +19,7 @@ func WithPipe(wcb func(w io.Writer) error, rcb func(r io.Reader) error) error {
 	eg.Go(func() error {
 		err := rcb(pr)
 		pr.CloseWithError(err)
-		return pr.Close()
+		return err
 	})
 	return eg.Wait()
 }
--- a/src/internal/miscutil/miscutil.go
+++ b/src/internal/miscutil/miscutil.go
@@ -12,15 +12,13 @@ func WithPipe(wcb func(w io.Writer) error, rcb func(r io.Reader) error) error {
 	pr, pw := io.Pipe()
 	eg := errgroup.Group{}
 	eg.Go(func() error {
-		if err := wcb(pw); err != nil {
-			return pw.CloseWithError(err)
-		}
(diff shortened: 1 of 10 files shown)


src/internal/storage/chunk/transform.go
```diff
@@ -190,7 +190,7 @@ func cryptoXOR(key, dst, src []byte) {
 func verifyData(id ID, x []byte) error {
 	actualHash := Hash(x)
 	if !bytes.Equal(actualHash[:], id) {
-		return errors.Errorf("bad chunk. HAVE: %x WANT: %x", actualHash, id)
```
brendoncarroll, line 193: This is the most important change in here.  This will prevent the double hex encoding.

src/internal/miscutil/miscutil.go
```diff
@@ -12,15 +12,13 @@ func WithPipe(wcb func(w io.Writer) error, rcb func(r io.Reader) error) error {
...
+		pw.CloseWithError(err)
+		return err
 	})
 	eg.Go(func() error {
-		if err := rcb(pr); err != nil {
-			return pr.CloseWithError(err)
-		}
+		err := rcb(pr)
+		pr.CloseWithError(err)
 		return pr.Close()
```
brycemcanally, line 22: This seems like it should be `return err`. The pipe reader is already closed at this point and we want the error.

src/server/pfs/server/driver.go
```diff
@@ -91,6 +91,14 @@ func newDriver(env serviceenv.ServiceEnv, txnEnv *txnenv.TransactionEnv, etcdPre
 	if err != nil {
 		return nil, err
 	}
+	// test object storage.
+	if err := func() error {
+		ctx, cf := context.WithTimeout(context.Background(), 10*time.Second)
```
brycemcanally, line 96: We might want to bump this up slightly, maybe 30s.

brycemcanally (approved):
lgtm

Issue #6536, opened 2021-07-15
Resumable downloads in pachctl get file


lukemarsden:
**What is the goal / desired outcome?**

If there's a network glitch or pachd restarts while a large file is downloaded via `pachctl get file` currently the experience is a frustrating failure. This is especially annoying for long slow downloads.

**If there is a way to accomplish this today via workaround, what does that require?**

Currently it can be achieved by using the S3 gateway in conjunction with `wget -c --timeout=1 --tries=5 --retry-connrefused`:

```
luke@mind:~$ (sleep 5; kubectl delete po -l app=pachd) &
[1] 189910
luke@mind:~$ wget -c --timeout=1 --tries=5 --retry-connrefused 192.168.39.50:30600/master.foo/ubuntu-20.04.2-live-server-amd64.iso
--2021-07-15 13:11:07--  http://192.168.39.50:30600/master.foo/ubuntu-20.04.2-live-server-amd64.iso
Connecting to 192.168.39.50:30600... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1215168512 (1.1G) [application/octet-stream]
Saving to: 'ubuntu-20.04.2-live-server-amd64.iso'
ubuntu-20.04.2-live-server-amd64.iso                  27%[===============================>                                                                                      ] 324.01M   105MB/s    eta 8s     pod "pachd-6b9f84b767-rr5sv" deleted
ubuntu-20.04.2-live-server-amd64.iso                  29%[==================================>                                                                                   ] 346.00M   107MB/s    in 3.2s    
2021-07-15 13:11:11 (107 MB/s) - Connection closed at byte 362807296. Retrying.
--2021-07-15 13:11:12--  (try: 2)  http://192.168.39.50:30600/master.foo/ubuntu-20.04.2-live-server-amd64.iso
Connecting to 192.168.39.50:30600... failed: Connection refused.
Retrying.
--2021-07-15 13:11:14--  (try: 3)  http://192.168.39.50:30600/master.foo/ubuntu-20.04.2-live-server-amd64.iso
Connecting to 192.168.39.50:30600... failed: Connection refused.
Retrying.
--2021-07-15 13:11:17--  (try: 4)  http://192.168.39.50:30600/master.foo/ubuntu-20.04.2-live-server-amd64.iso
Connecting to 192.168.39.50:30600... failed: Connection refused.
Retrying.
[...]

PR #6646, opened 2021-07-29
GetFile -  allow file read to begin at a certain offset number of bytes

acohen4:
Story: 
A user is downloading a big file, and that file download is interrupted/fails while it's successfully written a fraction of the total bytes.  The user would like to retry the download without having to re-load all of the completed bytes.

Proposed UX:
Add a `--retry` flag to `pachctl get file <FILE> -o <OUTPUT>` that is interpreted to only read the file's bytes beginning at the length of the current OUTPUT file.

As a part of this change, we also expose a `--offset` flag that can be used to specify getting file starting a certain offset number of bytes within the file.

See:
https://github.com/pachyderm/pachyderm/issues/6536


src/internal/progress/progress.go
```diff
@@ -86,6 +86,18 @@ func Open(path string) (*File, error) {
 	return newFile(file, path, fi.Size()), nil
 }
 
+func OpenAppend(path string, appendSize int64) (*File, error) {
+	file, err := os.OpenFile(path, os.O_APPEND|os.O_RDWR, 0666)
```
brendoncarroll, line 90: Is read permission required for the progress bar?

src/pfs/pfs.proto
```diff
@@ -320,8 +320,8 @@ message ModifyFileRequest {
 message GetFileRequest {
   File file = 1;
   string URL = 2;
+  int64 offset_bytes = 3;
```
brendoncarroll, line 323: "_bytes" might be unnecessary.  There's also a bytes type in protobuf's which is a little confusing.

src/server/pfs/server/testing/server_test.go
```diff
@@ -890,6 +890,33 @@ func TestPFS(suite *testing.T) {
 		require.Equal(t, data, b.String())
 	})
 
+	suite.Run("BasicFileWithOffset", func(t *testing.T) {
```
brendoncarroll, line 893: I see "SimpleFile" below, we should be consistent with "Basic" vs. "Simple"

src/server/pfs/cmds/cmds.go
```diff
@@ -1201,18 +1203,32 @@ $ {{alias}} 'foo@master:/test\[\].txt'`,
 				if err != nil {
 					return err
 				}
-				f, err := progress.Create(outputPath, int64(fi.SizeBytes))
-				if err != nil {
-					return err
+				var f *progress.File
+				if ofi, err := os.Stat(outputPath); retry && err == nil {
```
brendoncarroll, line 1207: The right way to do this if not exists pattern is to use `os.O_EXCL` but it's not a big deal because Create can just error and the user will probably retry.
https://pkg.go.dev/os#pkg-constants

src/server/pfs/server/testing/server_test.go
brendoncarroll: One thing to consider here is that when working on changes to the read path you might want to run all the GetFile tests using `go test -run GetFile`.  Calling this `GetFileWithOffset` would let us easily match this with the other GetFile tests.

brendoncarroll (approved):
I'm trying to think of another option we would pass to `Content` that wasn't to control the chunk.Reader.  But it seems like that method will only ever deal with chunk data.  So I think what you have makes sense.

I think we should drop "Bytes" from the end of the protobuf field.  The private variables don't need to be changed.

src/internal/storage/chunk/reader.go
```diff
@@ -11,24 +11,44 @@ import (
...
 // Iterate iterates over the data readers for the data references.
 func (r *Reader) Iterate(cb func(*DataReader) error) error {
 	for _, dataRef := range r.dataRefs {
+		if dataRef.SizeBytes < r.offsetBytes {
+			r.offsetBytes -= dataRef.SizeBytes
+			continue
+		}
+		dataRef.OffsetBytes += r.offsetBytes
+		dataRef.SizeBytes -= r.offsetBytes
+		r.offsetBytes = 0
```
brendoncarroll, line 51: I think there might be an issue with this.  We shouldn't be modifying the datarefs, or writing to the Reader variables.  We need to be able to call iterate more than once.  The Reader is supposed to be like a specification for the iteration.

I think we need to change `chunk.DataReader` to take an offset, and use that offset in `Get`.  Then in this loop we skip the refs that are before, use the DataReader with the offset for the overlap, and then just iterate as usual for the rest.

brendoncarroll (changes requested):
Found an issue in Iterate

src/internal/storage/chunk/reader.go
acohen4: good catch! that makes sense 👍

src/internal/progress/progress.go
acohen4: it actually isn't necessary -  changed it to write only

src/server/pfs/cmds/cmds.go
acohen4: I can restructure it but the reason I use os.Stat s to check file size if it exists

acohen4:
On Friday, @brycemcanally had mentioned that we will ultimately want to have resumable downloads/uploads implemented by potentially running the storage layer client-side, and have it be able to coordinate to see what data is stored in pach. 

Is the behavior here any different than that of resumable downloads? My understanding is that this "retry" path is the more manual of the two, since it's initiated by the user, whereas the "resume" path could potentially automatically handle more failure scenarios.

Are there other future considerations we should be making as part of this story?

What do you think @brendoncarroll & @brycemcanally ?

src/internal/storage/chunk/reader.go
```diff
@@ -7,29 +7,49 @@ import (
...
 func (r *Reader) Iterate(cb func(*DataReader) error) error {
+	offset := r.offsetBytes
 	for _, dataRef := range r.dataRefs {
-		dr := newDataReader(r.ctx, r.client, r.memCache, dataRef)
+		if dataRef.SizeBytes <= offset {
+			offset -= dataRef.SizeBytes
+			continue
+		}
+		dr := newDataReader(r.ctx, r.client, r.memCache, dataRef, offset)
+		offset = 0
```
brendoncarroll, line 52: Is this assignment necessary?

src/internal/storage/chunk/reader.go
```diff
@@ -71,8 +93,9 @@ func (dr *DataReader) DataRef() *DataRef {
 
 // Get writes the data referenced by the data reference.
 func (dr *DataReader) Get(w io.Writer) error {
+	logrus.Errorf("Offset:::: %v", dr.offset)
 	return Get(dr.ctx, dr.client, dr.memCache, dr.dataRef.Ref, func(chunk []byte) error {
-		data := chunk[dr.dataRef.OffsetBytes : dr.dataRef.OffsetBytes+dr.dataRef.SizeBytes]
+		data := chunk[dr.dataRef.OffsetBytes+dr.offset : dr.dataRef.OffsetBytes+dr.dataRef.SizeBytes]
```
brendoncarroll, line 98: We need a check so we don't panic here.

src/internal/storage/chunk/reader.go
acohen4: yep! if it's omitted and there are more dataRefs to iterate through, we will apply that small offset to them too

src/client/option.go
```diff
@@ -72,3 +72,9 @@ func WithDatumGetFile(datum string) GetFileOption {
 		gf.File.Datum = datum
 	}
 }
+
+func WithOffsetBytes(offset int64) GetFileOption {
```
brendoncarroll, line 76: The public API should also be `WithOffset` to match the gRPC message.  Private variables don't matter, but anything user facing does.

PR #6647, opened 2021-07-29
Remove context from transaction context.

PFedak:
Nested transactions have the potential to cause deadlocks if they
conflict - the database might allow the inner transaction, not realizing
that when the outer one fails, the inner will be performed again. It's
also not a good idea in general to call a gRPC inside a transaction, in
case it involves waiting on some other process.

To make these operations hard to perform, the transaction shouldn't have
easy access to a context. Most existing usages were trivial to avoid by
only allowing them in the parent call, which had a context. Inspect
pipeline with details and checking enterprise limits were two such cases
where we never actually needed the context part in a transaction.

The remaining case is pipeline spec files. While we've discussed some
workarounds, creating and updating a pipeline still requires both
creating a fileset and reading a file, which cannot currently be done
transactionally. These operations are provided through an interface which
essentially just fills in the original context.


src/internal/transactionenv/txncontext/context.go
```diff
@@ -29,6 +30,36 @@ type TransactionContext struct {
...
+func New(ctx context.Context, sqlTx *sqlx.Tx, authServer identifier) (*TransactionContext, error) {
+	var username string
+	// check auth once now so that we can refer to it later
+	if authServer != nil {
+		if me, err := authServer.WhoAmI(ctx, &auth.WhoAmIRequest{}); err != nil && !auth.IsErrNotActivated(err) {
+			return nil, err
+		} else if err == nil {
+			username = me.Username
+		}
+	}
```
PFedak, line 49: Moving the auth check here might be unexpected for requests that don't ever require auth, but this isn't a problem for user requests, and internal ones should either be given an auth token or maybe using an internal name with the right permissions. It would obviously be easier to just pass in the username string, but I wanted to make it a little harder to accidentally escalate permissions.

src/server/pfs/server/api_server.go
```diff
@@ -64,8 +67,18 @@ func newAPIServer(env serviceenv.ServiceEnv, txnEnv *txnenv.TransactionEnv, etcd
 func (a *apiServer) ActivateAuth(ctx context.Context, request *pfs.ActivateAuthRequest) (response *pfs.ActivateAuthResponse, retErr error) {
 	func() { a.Log(request, nil, nil, 0) }()
 	defer func(start time.Time) { a.Log(request, response, retErr, time.Since(start)) }(time.Now())
+	var repoInfo pfs.RepoInfo
 	if err := a.txnEnv.WithWriteContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
-		return a.driver.activateAuth(txnCtx)
+		return a.driver.repos.ReadOnly(ctx).List(&repoInfo, col.DefaultOptions(), func(string) error {
```
PFedak, line 72: I just inlined the function to make it clearer that this is still nested

src/server/pfs/server/api_server.go
```diff
@@ -336,7 +349,12 @@ func (a *apiServer) InspectBranchInTransaction(txnCtx *txncontext.TransactionCon
 func (a *apiServer) ListBranch(request *pfs.ListBranchRequest, srv pfs.API_ListBranchServer) (retErr error) {
 	func() { a.Log(request, nil, nil, 0) }()
 	defer func(start time.Time) { a.Log(request, nil, retErr, time.Since(start)) }(time.Now())
-	return a.driver.listBranch(srv.Context(), request.Repo, request.Reverse, srv.Send)
+	if request.Repo == nil {
+		return a.driver.listBranch(srv.Context(), request.Reverse, srv.Send)
+	}
+	return a.txnEnv.WithReadContext(srv.Context(), func(txnCtx *txncontext.TransactionContext) error {
+		return a.driver.listBranchInTransaction(txnCtx, request.Repo, request.Reverse, srv.Send)
+	})
```
PFedak, line 357: I separated these two explicitly based on transaction compatibility, but I am happier to call out that this is a different path with potentially different auth requirements

src/server/pfs/server/driver.go
```diff
@@ -1738,24 +1726,11 @@ func (d *driver) inspectBranch(txnCtx *txncontext.TransactionContext, branch *pf
...
-	// Make sure that the repo exists
-	if repo.Name != "" {
-		if err := d.repos.ReadOnly(ctx).Get(repo, &pfs.RepoInfo{}); err != nil {
-			if col.IsErrNotFound(err) {
-				return pfsserver.ErrRepoNotFound{Repo: repo}
-			}
-			return err
-		}
+	} else if err == nil {
+		return errors.New("Cannot list branches from all repos with auth activated")
```
PFedak, line 1733: this is effectively what happened before, since we were checking for a role binding for `REPO:`

src/server/pfs/server/driver.go
```diff
@@ -1789,17 +1764,41 @@ func (d *driver) listBranch(ctx context.Context, repo *pfs.Repo, reverse bool, c
...
 	}
 
-	return sendBis()
+	opts := &col.Options{Target: col.SortByCreateRevision, Order: col.SortDescend}
+	if reverse {
+		opts.Order = col.SortAscend
+	}
+	branchInfo := &pfs.BranchInfo{}
+	return d.branches.ReadWrite(txnCtx.SqlTx).GetByIndex(pfsdb.BranchesRepoIndex, pfsdb.RepoKey(repo), branchInfo, opts, func(_ string) error {
+		return cb(branchInfo)
```
PFedak, line 1800: this skips the provenance sorting logic, which does nothing here because branches in a repo can't be provenant on each other

src/server/pfs/server/master.go
```diff
@@ -27,6 +28,7 @@ const (
 )
 
 func (d *driver) master(ctx context.Context) {
+	ctx = auth.AsInternalUser(ctx, "pfs-master")
```
PFedak, line 31: this is the approach I've taken for metrics and pps master, this makes it appear authorized but gives no permissions

src/server/pps/server/api_server.go
```diff
@@ -2366,74 +2355,72 @@ func (a *apiServer) InspectPipeline(ctx context.Context, request *pps.InspectPip
...
-	}
-	// Erase any AuthToken - this shouldn't be returned to anyone (the workers
-	// won't use this function to get their auth token)
-	pipelineInfo.AuthToken = ""
-	// TODO(global ids): how should we treat ancestry here?  Alias commits are going to gum it up.
-	pipelineInfo.SpecCommit.ID = ancestry.Add(pipelineInfo.SpecCommit.ID, ancestors)
 	if details {
-		// the spec commit must already exist outside of the transaction, so we can retrieve it normally
-		if err := internal.GetPipelineDetails(txnCtx.ClientContext, a.env.PfsServer(), pipelineInfo); err != nil {
+		if err := internal.GetPipelineDetails(ctx, a.env.PfsServer(), info); err != nil {
```
PFedak, line 2368: we were never requesting the details in transactions, anyway

src/server/transaction/server/driver.go
```diff
@@ -170,12 +170,6 @@ func (d *driver) runTransaction(txnCtx *txncontext.TransactionContext, info *tra
 			err = directTxn.UpdateJobState(request.UpdateJobState)
 		} else if request.StopJob != nil {
 			err = directTxn.StopJob(request.StopJob)
-		} else if request.DeleteAll != nil {
```
PFedak, line 173: seemed unused, makes things easier

src/internal/storage/fileset/metadata_store.go
```diff
@@ -2,21 +2,21 @@ package fileset
...
 	"github.com/jmoiron/sqlx"
 	"github.com/pachyderm/pachyderm/v2/src/internal/dbutil"
-	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/require"
 )
 
 var (
 	// ErrFileSetExists path already exists
-	ErrFileSetExists = errors.Errorf("path already exists")
+	ErrFileSetExists = fmt.Errorf("path already exists")
```
brendoncarroll, line 15: This prevents us from checking for theses errors with `==`.  It's generally better to have each error be a separate type so that the instances can carry stack information, but we aren't ready to do that yet.

src/internal/internal.go
```diff
@@ -75,21 +74,30 @@ func (g *fileGetter) Send(bytes *types.BytesValue) error {
 	return err // always nil
 }
 
-func GetPipelineDetails(ctx context.Context, apiServer pfs.APIServer, info *pps.PipelineInfo) error {
+type fileReader interface {
```
brendoncarroll, line 77: All of these helper methods can go in `src/internal/clientsdk`

src/internal/transactionenv/txncontext/context.go
```diff
@@ -5,17 +5,18 @@ import (
...
 )
 
 // TransactionContext is a helper type to encapsulate the state for a given
 // set of operations being performed in the Pachyderm API.  When a new
 // transaction is started, a context will be created for it containing these
 // objects, which will be threaded through to every API call:
 type TransactionContext struct {
-	// ClientContext is the incoming context.Context for the request.
-	ClientContext context.Context
+	username string
```
brendoncarroll, line 19: Why not export this?  This is great, we should use this when we are in a transaction.

src/server/pachyderm_test.go
```diff
@@ -9648,6 +9647,56 @@ func TestMoveBranchTrigger(t *testing.T) {
...
+
+	info, err := c.InspectTransaction(txn)
+	require.NoError(t, err)
+	require.Equal(t, 1, len(info.Responses))
+	require.NotNil(t, info.Responses[0].CreatePipelineResponse)
+	id := info.Responses[0].CreatePipelineResponse.FileSetId
+
+	// make sure the fileset is gone
+	// TODO: do this without a sleep (delete from database directly?)
+	require.NoError(t, c.RenewFileSet(id, time.Second))
```
brendoncarroll, line 9690: We support negative durations internally for this exact use-case.  Check if they work through the API, if not, this is fine for now.

src/server/pfs/server/transaction_access.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal"
+	"github.com/pachyderm/pachyderm/v2/src/internal/transactionenv/txncontext"
+	"github.com/pachyderm/pachyderm/v2/src/pfs"
+)
+
+type FileAccessor struct {
+	ctx       context.Context
```
brendoncarroll, line 17: This seems like it is just moving the problem. Any fileset returned from this shouldn't even be readable with snapshot or serializable

brendoncarroll (changes requested):
I don't think the FileAccessor approach is going to work.  All the filesets need to be created before the call to BeginTx.

We either need to remove Create and Update pipeline as in-transaction operations.  Or special case the pipeline requests so that we create the filesets for them in the transaction context constructor, and then maybe store them in a `map[string]string` and refer to them in the transaction using the map.

src/server/pachyderm_test.go
PFedak: The API has an explicit check disallowing durations < 1 second, but even if I set it negative, it seems like reads of filesets succeed regardless of TTL, so I thought I'd need to wait for GC anyway.

src/internal/internal.go
PFedak: I can definitely do that, I worried it would be confusing since they aren't really wrappers around client functions as these are. I also thought we might want to make the existing clientsdk functions non-internal at some point, since it would help anyone writing code against our API.

src/internal/transactionenv/txncontext/context.go
PFedak: I didn't want to make it easy to modify, but maybe that's not an actual concern. The `WhoAmI` method I exposed seemed like a simple dropin, anyway.

src/server/pfs/server/transaction_access.go
PFedak: I guess I wasn't clear about intention - I planned on explicitly sidestepping that issue for now. This accomplishes the goal of removing other nested transactions and making future ones harder to create. 

This interface makes it easier to implement the compromise you (iirc) mentioned earlier, where filesets are handled by a layer outside of the transaction. We can make the `CreateFileset` method here always return a `TransactionConflict` error itself, and have it add the ID to a list that the txnCtx renews on failure, since any fileset must be created through this object.

src/server/pfs/server/api_server.go
brendoncarroll: I think this has the potential to work correctly.  We just need a way to construct a read only collection from a transaction.

src/server/pfs/server/transaction_access.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (a *apiServer) NewFileAccessor(ctx context.Context) txncontext.FileAccessor {
+	return &FileAccessor{
+		ctx:       ctx,
+		apiServer: a,
+		d:         a.driver,
+	}
+}
+
+// CreateFileset creates a fileset with a single file
```
brendoncarroll, line 30: If we are going to fix this in a later P.R. We should have a TODO here indicating that this is still a problem.

brendoncarroll: We also don't need the PFS driver in here.  We can just use the APIServer.

src/internal/internal.go
brendoncarroll: Maybe ppsutil then?  I'm not really a fan of just dumping things into the root of `./src/internal`.

src/internal/transactionenv/txncontext/context.go
brendoncarroll: Okay, makes sense to me.

src/server/pfs/server/transaction_access.go
PFedak: I was hoping to avoid writing another grpc streaming server wrapper

brendoncarroll: You need to send one message over the CreateFileset stream, then close it.  The generated client should have the same API as ModifyFile.  You may be able to factor something out into clientsdk that client and this can reference.  But it shouldn't be much code either way to create a fileset with one file.

PFedak: The requests aren't the issue, but the client is only generated automatically when you actually call the gRPC, if I'm not mistaken. I had written the fake server logic in #6593 but you didn't seem to like that idea.

src/server/pfs/server/transaction_access.go
brendoncarroll: Talked about this offline. I think the direction here is really cool.

src/server/pfs/server/transaction_access.go
brendoncarroll: In the case of that other PR: If you want a utility function that takes a slice of file specs or maybe a `map[string][]byte` and turns it into a fileset, that's fine.  I just think it should live somewhere other than the root of internal.

In the case of this P.R: It looks like we are just calling createFileset on the driver in a way we could also call it on the APIServer.  If there is a parameter or behavior that would be different between the two, then lets keep using the driver.

The common concern is that we put helper code in a discoverable place, only write it once, and only write it if the API doesn't provide an easy way to do it, and write it to use the gRPC API, so it will see the most use.

src/server/pachyderm_test.go
brendoncarroll: This is fine for now.  We can probably expose a way to force the GC to run.

src/internal/ppsutil/util.go
```diff
@@ -117,12 +117,17 @@ func GetLimitsResourceList(limits *pps.ResourceSpec) (*v1.ResourceList, error) {
 
 // GetPipelineDetails retrieves the pipeline spec for the given PipelineInfo from PFS,
 // then fills in the Details field in the given PipelineInfo from the spec.
-func GetPipelineDetails(pachClient *client.APIClient, pipelineInfo *pps.PipelineInfo) error {
-	// ensure we are authorized to read the pipeline's spec commit, but don't propagate that back out
-	pachClient = pachClient.WithCtx(pachClient.Ctx())
-	buf := bytes.Buffer{}
-	if err := pachClient.GetFile(pipelineInfo.SpecCommit, ppsconsts.SpecFile, &buf); err != nil {
-		return errors.Wrapf(err, "could not retrieve pipeline spec file from PFS for pipeline '%s'", pipelineInfo.Pipeline.Name)
+func GetPipelineDetails(ctx context.Context, env serviceenv.ServiceEnv, pipelineInfo *pps.PipelineInfo) error {
```
brendoncarroll, line 120: 👍

src/server/pps/iface.go
```diff
@@ -18,5 +20,7 @@ type APIServer interface {
 	StopJobInTransaction(*txncontext.TransactionContext, *pps_client.StopJobRequest) error
 	UpdateJobStateInTransaction(*txncontext.TransactionContext, *pps_client.UpdateJobStateRequest) error
 	CreatePipelineInTransaction(*txncontext.TransactionContext, *pps_client.CreatePipelineRequest, *string, *uint64) error
-	InspectPipelineInTransaction(*txncontext.TransactionContext, string, bool) (*pps_client.PipelineInfo, error)
+	InspectPipelineInTransaction(*txncontext.TransactionContext, string) (*pps_client.PipelineInfo, error)
+
+	ListPipelineCallback(context.Context, *pps_client.ListPipelineRequest, func(*pps_client.PipelineInfo) error) error
```
brendoncarroll, line 25: I think we use `ForEachX` for this elsewhere.  It's internal, so we can change it later.

src/internal/storage/option.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/obj"
+	"github.com/pachyderm/pachyderm/v2/src/internal/serviceenv"
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/chunk"
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset"
+	"github.com/pachyderm/pachyderm/v2/src/internal/uuid"
+)
+
+// FilesetOptions returns the fileset storage options for the config.
+func FilesetOptions(conf *serviceenv.Configuration) []fileset.StorageOption {
```
brendoncarroll, line 15: Is there a reason why these needed to move? Like a dependency cycle?

I'd prefer to keep the storage directory empty.  We can call these `MakeOptions` or `OptionsFromConfig` or whatever name you want.

PFedak: I was fighting an import cycle from trying to allow `CreateFilesetCallback(context.Context, func(*fileset.UnorderedWriter) error)` to exist in the pfs api server interface, but I think it's a lost cause. It just takes a slice of requests now.

src/server/pfs/iface.go
```diff
@@ -29,4 +33,8 @@ type APIServer interface {
 	DeleteBranchInTransaction(*txncontext.TransactionContext, *pfs_client.DeleteBranchRequest) error
 
 	AddFileSetInTransaction(*txncontext.TransactionContext, *pfs_client.AddFileSetRequest) error
+
+	ListRepoCallback(context.Context, *pfs_client.ListRepoRequest, func(*pfs_client.RepoInfo) error) error
+	CreateFileSetCallback(context.Context, []*pfs_client.ModifyFileRequest) (*fileset.ID, error)
```
brendoncarroll, line 38: This doesn't take a callback.

src/server/pfs/iface.go
```diff
@@ -29,4 +33,8 @@ type APIServer interface {
...
+	GetFileWithWriter(context.Context, *pfs_client.GetFileRequest, io.Writer) (int64, error)
```
brendoncarroll, line 39: This doesn't need to be on the interface we can implement this in terms of the `pfs.APIClient`.

src/internal/transactionenv/refresh.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset"
+	"github.com/pachyderm/pachyderm/v2/src/internal/transactionenv/txncontext"
+	"github.com/pachyderm/pachyderm/v2/src/pfs"
+	"github.com/pachyderm/pachyderm/v2/src/pps"
+	"github.com/pachyderm/pachyderm/v2/src/transaction"
+)
+
+const defaultTTLSeconds = 60 * 10
+
+type errTransactionConflict struct{}
```
brendoncarroll, line 21: 👍

src/server/pfs/iface.go
```diff
@@ -29,4 +33,8 @@ type APIServer interface {
 	DeleteBranchInTransaction(*txncontext.TransactionContext, *pfs_client.DeleteBranchRequest) error
 
 	AddFileSetInTransaction(*txncontext.TransactionContext, *pfs_client.AddFileSetRequest) error
+
```
brendoncarroll, line 36: Talked offline, these extra methods can come out.

brendoncarroll (approved):
:shipit:

Issue #6626, opened 2021-07-27
"No such file or directory" Trying to List Contents of Pipeline's Input Repo


dgeorg42:
I am now working on testing Beta 3 with some real-life pipelines and data that one of our customers was kind enough to share with us. The first pipeline is pretty simple, as it just lists the files in the input repo and send the listing to a file in the output repo:
```json
{
  "pipeline": {
    "name": "data_source_dualfan_list_years"
  },
  "transform": {
    "image": "registry.access.redhat.com/ubi8/ubi-minimal:8.3",
    "cmd": [
      "/bin/bash"
    ],
    "stdin": [
      "#!/bin/bash",
      "ls $REPO_IN > /pfs/out/data_years.txt"
    ]
  },
  "input": {
    "pfs": {
      "name": "REPO_IN",
      "repo": "data_source_dualfan_linkmerge",
      "glob": "/dualfan",
      "empty_files": true
    }
  },
  "resource_requests": {
    "memory": "1K",
    "cpu": 0.01
  },
  "autoscaling": false
}
```

When the job runs, it immediately fails with the following message:
```
"message":"ls: cannot access '/pfs/data_source_dualfan_linkmerge': No such file or directory"
```

The repo does, in fact, exist, and it does have files in it:
```
-->> pachctl list file data_source_dualfan_linkmerge@master
NAME      TAG TYPE SIZE     
/dualfan/     dir  11.12MiB 
```

Has something changed in the file system semantics for accessing the input repo?

dgeorg42:
OK, I see what the problem is now. 

In the pipeline's `input` section, they have the repo named as follows:
```
"pfs": {
      "name": "REPO_IN",
      "repo": "data_source_dualfan_linkmerge",
```

As a result, the directory for the input repo is actually `/pfs/REPO_IN`

In their `transform` section, the command they execute is:
```
ls $REPO_IN > /pfs/out/data_years.txt
```

This is failing because the $REPO_IN environment variable is being set to `/pfs/data_source_dualfan_linkmerge` and that directory does not exist.  Since this pipeline works for them in 1.13.2, I'm assuming that 1.13.2 must be setting the environment variable to `/pfs/REPO_IN` so that it matches that actual directory name that was created.

Is this an intentional change in 2.0?

dgeorg42:
Confirmed that this is fixed in Beta 5.

PR #6648, opened 2021-07-29
A few ergonomic changes and a bug fix

brycemcanally:
This PR contains the following ergonomic changes and bug fix:
- Rename the file tag field to datum. Everything that references this field as a tag (functions, variables, comments, etc.) have been moved over to datum. This should make it more clear to users, and internally, what the purpose of this field is, which is to map a file back to the datum that output it.
- Removes the datum column from the list file output in pachctl.
- Adds an environment variable called `PACH_DATUM_ID` that is available to user code which stores the datum ID. This gives users an easy way to ensure that each datum outputs distinct file paths.
- Changes the commit error field to a string so we can record the reason for the error and propagate this to the job reason if it causes a job failure.
- Changes the commit size upper bound symbol to ≤ rather than <=.
- Fixes https://github.com/pachyderm/pachyderm/issues/6626.


acohen4 (approved):
LGTM

PR #6649, opened 2021-07-30
Bitnami pg

chainlink:
This also gets rid of trust auth in PG

Fixes: https://github.com/pachyderm/pachyderm/issues/6659


etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -55,6 +55,11 @@ spec:
         - name: POSTGRES_SSL
           value: {{ .Values.global.postgresql.ssl | quote }}
         {{- end }}
+        - name: POSTGRES_PASSWORD
+          valueFrom:
+            secretKeyRef:
```
seslattery, line 60: Very Minor: Could maybe do a `envFrom: - secretRef:` here so if additional secrets are ever added, don't need to add additional wiring.

etc/helm/pachyderm/templates/enterprise-server/postgres-secret.yaml
```diff
@@ -0,0 +1,16 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/ -}}
+{{- if and .Values.enterpriseServer.enabled (not .Values.postgresql.enabled) -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    app: postgresql
+    suite: pachyderm
+  name: postgres # Must match secretname specified for postgres password in pachd deployment, enterprise and pg bouncer
```
seslattery, line 12: Doesn't this conflict with `pachd/postgres-secret.yaml`?

seslattery (approved):

etc/helm/pachyderm/templates/enterprise-server/postgres-secret.yaml
chainlink: They'll only ever be deployed separately

PR #6650, opened 2021-07-30
Switch s3 sidecar pipeline detail fetch back to gRPC

PFedak:
Also run s3 tests with auth enabled by default.

### DIFF
--- a/etc/testing/circle/run_tests.sh
+++ b/etc/testing/circle/run_tests.sh
@@ -103,6 +103,7 @@ case "${BUCKET}" in
     bucket_num="${BUCKET#PPS}"
     test_bucket "./src/server" test-pps "${bucket_num}" "${PPS_BUCKETS}"
     if [[ "${bucket_num}" -eq "${PPS_BUCKETS}" ]]; then
+      export PACH_TEST_WITH_AUTH=1
       go test -v -count=1 ./src/server/pps/server -timeout 420s
     fi
     ;;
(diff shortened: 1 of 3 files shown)


etc/testing/circle/run_tests.sh
```diff
@@ -115,7 +116,7 @@ case "${BUCKET}" in
     make test-enterprise
     # Launch a stand-alone enterprise server in a separate namespace
     make launch-enterprise
-    echo "{\"pachd_address\": \"grpc://${VM_IP}:${ENTERPRISE_PORT}\", \"source\": 2}" | pachctl config set context "enterprise" --overwrite
```
acohen4, line 118: this is just formatting right?

acohen4 (approved):
Change looks good! I just don't understand why S3G needs to communicate over gRPC. Could you explain?

PFedak:
It doesn't actually need to, I should be a bit clearer. It's just a silly thing with context metadata. We put the auth token in the outgoing metadata for the grpc to read and place in the new context's incoming metadata. Our checks look at the incoming metadata, so if you remove the gRPC from the process they don't find auth info. It could also be fixed by copying the metadata over.

etc/testing/circle/run_tests.sh
PFedak: yes, I guess my editor did that automatically.

PR #6651, opened 2021-07-30
Fix context cancel race in new version of GetFile

brycemcanally:
This PR fixes a race condition in the new version of GetFile. The basic problem was that an abstraction (`fileset.File`) was being used outside of the callback based iteration that it was created in. This was a problem because the context used by `fileset.File` was the one associated with the iteration. Once the iteration is completed, the context is canceled and attempting to get the content would return context canceled. The change I made was to remove the internal context from the implementations of `fileset.File` and require the appropriate context to be passed into the functions that need a context.


acohen4 (approved):
LGTM!

PR #6654, opened 2021-08-02
Remove alias to dbutil.WithTx

brendoncarroll:
Makes it easier to identify places that perform a transaction.  This was already an alias, so it made sense to just do the rename.


PFedak (approved):
LGTM, figured this was probably just a vestige of the previous consolidation

PR #6655, opened 2021-08-02
Fix launch dev

chainlink:
Currently `make launch-dev` with no args fails with 

```
Error: flag needs an argument: --set
```

Looks like from the changes here: https://github.com/pachyderm/pachyderm/pull/6590/files 

Setting `clusterDeploymentId: dev` is now done in the values file referenced, so we don't need to set it via the command line. 

I've also removed it from the vscode tasks file. @Tryneus `make launch-dev` doesn't do a docker build, should we add a `make docker-build` to the `launch-dev` vs code task?

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -16,7 +16,7 @@
     {
       "label": "launch-dev",
       "type": "shell",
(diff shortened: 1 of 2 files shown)


Tryneus (approved):
Ah, didn't know it wouldn't handle this case, but the config values file seems much better to me.  I prefer not having `launch-dev` do a build in my workflow, tbh.

LGTM!

PR #6656, opened 2021-08-03
Allow for multiple pach clusters in different namespaces.

PFedak:
While our deploy logic in general works within the provided namespace,
we also create a ClusterRoleBinding which spans namespaces. This gets
overwritten during kubectl apply, meaning only one pachyderm
ServiceAccount will have the appropriate role at a time.

Switching to (non-Cluster) RoleBindings would also solve this, but I
didn't want to assume those permissions were never relied upon.


PFedak:
Other solution would be to use `kubectl patch` to append in this case, but I didn't like the special logic that would involve

mindthevirt:
<div>Agent Dale Georg linked Freshdesk ticket <a href="https://pachyderm-support.freshdesk.com/a/tickets/300">300</a> for this issue.</div>

jdoliner (approved):
LGTM

PR #6657, opened 2021-08-03
Add PFS & PPS default load test RPCs

brycemcanally:
This PR refactors the PFS & PPS load tests we run in CI into RPCs called `pfs.RunLoadTestDefault` and `pps.RunLoadTestDefault`. Calling these RPCs result with a default set of load tests being run against the deployment, and will return the results from the most recent load test. The pachctl wrappers for these RPCs are `pachctl run pfs-load-test` and `pachctl run pps-load-test` without any arguments. If one of the load tests fail, its spec, branch, seed, and the error will be returned by the RPC. If all of the load tests pass, the last load test that was run will be returned. We should eventually change this model to setup the load tests in a collection and run them asynchronously, but for now this gives us a straightforward way to run a default set of load tests in Hub.


src/server/pfs/server/api_server.go
```diff
@@ -788,14 +789,110 @@ func (a *apiServer) RunLoadTest(ctx context.Context, req *pfs.RunLoadTestRequest
...
+			return nil, err
+		}
+		if resp.Error != "" {
+			return resp, nil
+		}
+	}
+	return resp, nil
+}
+
+var defaultLoadSpecs = []string{`
```
brendoncarroll, line 818: If we are going to hard code these I think we should store them structured. Like call must parse on them or something.

src/server/pps/server/api_server.go
```diff
@@ -3135,6 +3136,83 @@ func (a *apiServer) ActivateAuth(ctx context.Context, req *pps.ActivateAuthReque
...
+	pachClient := a.env.GetPachClient(ctx)
+	repo := "load_test"
+	if err := pachClient.CreateRepo(repo); err != nil && !pfsServer.IsRepoExistsErr(err) {
+		return nil, err
+	}
+	branch := uuid.New()
+	if err := pachClient.CreateBranch(repo, branch, "", "", nil); err != nil {
+		return nil, err
+	}
+	pipeline := tu.UniqueString("TestLoadPipeline")
```
brendoncarroll, line 3153: Should we incorporate the time into this somehow, maybe in the description for the pipeline?  If there will only ever be one in a cluster then this is fine.  But if we are reusing a cluster we should have them timestamped.

brendoncarroll (approved):
LGTM.  The huge string constants aren't great.  We should look into a way to store those as separate files.

src/server/pfs/server/api_server.go
brendoncarroll: Unless they are exercised in tests.  e.g. All the SQL we have is going to be valid if it's sent to the database at all during tests.

brycemcanally: Would that be fine in a followup? I would like to get this in ASAP so we can have regression testing in Hub.

src/server/pps/server/api_server.go
brycemcanally: Yeah, there are definitely some changes like this that we should make over time. The main goal with this is an MVP for default tests that we can periodically run in Hub.

PR #6658, opened 2021-08-03
Allow multiple pachyderms in different namespaces.

PFedak:
2.0/helm version of #6656


chainlink:
I think we can close this? (discussed built in namespaced rbac offline)

PFedak:
`--local-roles` already works around this option, so it should me mentioned where we describe that flag (https://github.com/pachyderm/pachyderm/blob/master/doc/docs/master/how-tos/advanced-data-operations/use-transactions-to-run-multiple-commands.md)

PR #6662, opened 2021-08-03
Don't overwrite the default working dir unnecessarily.

PFedak:
Backport of part of #6426, which is needed when 1.x is run without
docker.


acohen4 (approved):
Looks good!

PR #6668, opened 2021-08-04
Fix error messages

nitinjainsj:
Minor changes.

In case of failure we still end up creating the output file. Is that expected?

```
$ pc get file images@master:/ -o /tmp/abc2.jpg
/tmp/abc2.jpg 0.00 b / 0.00 b [---------------------------------------------------------------------------------------] 0s 0.00 b/s
File / not found. Command only supports file paths
```


brendoncarroll (approved):
LGTM

Issue #6615, opened 2021-07-26
Job Enters Infinite Retry Loop with 120,000 Files in the Input Repo


dgeorg42:
I'm running one of my standard performance tests against Beta 3, where I load 20,000 files at a time into a repo, and the pipeline which runs simply copies each file over to the output repo. (The glob pattern is `/` so all files are part of a single datum.)

Once I reach 120,000 files, the job goes into a seemingly infinite loop of restarts. 
![image](https://user-images.githubusercontent.com/15801149/127038544-6aaf9bca-719c-4be2-9bf6-ca573d5673f8.png)

It's worth pointing out that it's only at the job level that the restarts are occurring.  The pods are not restarting:
![image](https://user-images.githubusercontent.com/15801149/127027416-4dc5997c-8091-4b4d-b0ac-030133ab74c2.png)

Nor are the containers:
![image](https://user-images.githubusercontent.com/15801149/127028032-d6a37dbd-b1f0-4ee8-b128-da60155daf26.png)

I'll attach the full logs, but here are some of the highlights:

Grepping for "error" in the storage container log:
![image](https://user-images.githubusercontent.com/15801149/127034911-2481a9bc-0dc4-47ee-84de-f068cfb195bf.png)

Grepping for "error" in the user container log:
![image](https://user-images.githubusercontent.com/15801149/127035306-74dcaebb-abb9-4d86-91db-57ec9ccef949.png)

While I saw occasional behavior like this in Beta 2, I couldn't reliably reproduce it.  With Beta 3, it doesn't always necessarily happen at the 120,000 file mark, but it always happens before I reach 200,000, which means that 
I'm unable to run the performance test to completion.

[user.log](https://github.com/pachyderm/pachyderm/files/6880382/user.log)
[storage.log](https://github.com/pachyderm/pachyderm/files/6880383/storage.log)

[AB#28](https://dev.azure.com/PachydermCompany/92b98a32-88e5-46cb-a28f-5d064712e03b/_workitems/edit/28)

jdoliner:
@Tryneus would you mind take a look at this if you have bandwidth? Seems like you've probably got the most context on it.

jdoliner:
It's possible that this is just looping because it's hitting the linux limit on files in a directory. In which case this would work if you broke up the paths a little bit.

dgeorg42:
[...]

PR #6673, opened 2021-08-04
Fix an issue with the file set renewer usage in the unordered writer

brycemcanally:
This PR fixes an issue with the file set renewer usage in the unordered writer. The identifier being used in the renewer add call was the tracker id rather than the file set id which resulted with renewal failing and canceling the renewal context. This error was obscured by the fact that the error is not within the main control flow, so the error that is seen is a context cancel. Logging will be added in another PR (https://github.com/pachyderm/pachyderm/pull/6664) to make these types of issues more obvious. Also, this PR removes the tag header column from the file info pretty printing (this was missed in a prior PR) and makes a change to ensure we don't attempt to succeed a job after a datum failure.

Fixes https://github.com/pachyderm/pachyderm/issues/6615

### DIFF
--- a/src/internal/storage/fileset/unordered_writer.go
+++ b/src/internal/storage/fileset/unordered_writer.go
@@ -115,7 +115,7 @@ func (uw *UnorderedWriter) withWriter(cb func(*Writer) error) error {
 	}
 	uw.ids = append(uw.ids, *id)
 	if uw.renewer != nil {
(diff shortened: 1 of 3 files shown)


brendoncarroll (approved):
We talked offline. TrackerID -> HexString is prone to getting messed up again.  Better fix in follow up P.R.

Issue #6671, opened 2021-08-04
[Helm] Use kubernetes secret for database password in PG Bouncer


chainlink:
The database password referenced in PG Bouncer env vars should be stored in a kubernetes secret 

[AB#40](https://dev.azure.com/PachydermCompany/92b98a32-88e5-46cb-a28f-5d064712e03b/_workitems/edit/40)

PR #6677, opened 2021-08-04
Store postgres password in k8s secret for pg-bouncer

acohen4:
closes https://github.com/pachyderm/pachyderm/issues/6671


chainlink:
The DB password is also in the storage secret (and will be in another secret in the upcoming bitnami PR) but i think this is good for now (we can clean up later)

chainlink (approved):

PR #6680, opened 2021-08-05
Run load tests against a Hub workspace

jrockway:
This adds a CI job that builds pachd, starts that on Hub, and runs the pfs and pps load tests.

I worry a little about pushing to pachyderm/pachd (etc.) from CI, since it seems easy for a bug to overwrite an existing tag and cause a massive outage across the fleet.  So we should do something about that.

### DIFF
--- a/etc/testing/circle/run_hub_tests.sh
+++ b/etc/testing/circle/run_hub_tests.sh
@@ -32,7 +32,7 @@ make install
 export VERSION=$(pachctl version --client-only)
 
 # build docker images
-DOCKER_BUILDKIT=1 goreleaser release -p 1 --snapshot --skip-publish --rm-dist -f goreleaser/docker.yml
+goreleaser release -p 1 --snapshot --skip-publish --rm-dist -f goreleaser/docker.yml
 docker tag pachyderm/pachd pachyderm/pachd:$VERSION
 docker tag pachyderm/worker pachyderm/worker:$VERSION
 docker tag pachyderm/pachctl pachyderm/pachctl:$VERSION
--- a/etc/testing/circle/run_hub_tests.sh
+++ b/etc/testing/circle/run_hub_tests.sh
@@ -4,6 +4,7 @@ set -euxo pipefail
 
 mkdir -p $HOME/go/bin
 export PATH=$PATH:/usr/local/go/bin:$HOME/go/bin
+export GOPATH=$HOME/go
 
 # Install go.
 sudo rm -rf /usr/local/go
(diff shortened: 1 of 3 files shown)


etc/testing/circle/run_hub_tests.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# install pachctl and friends
+make install
+VERSION=$(pachctl version --client-only)
+git config user.email "donotreply@pachyderm.com"
+git config user.name "anonymous"
+git tag -f -am "Circle CI test v$VERSION" v"$VERSION"
+
+# push a docker image
+make docker-build
```
brycemcanally, line 24: I think you are going to need a `make docker-push` here to actually push the images.

etc/testing/circle/run_hub_tests.sh
```diff
@@ -0,0 +1,54 @@
...
+popd
+
+# install goreleaser
+GORELEASER_VERSION=0.169.0
+curl -L https://github.com/goreleaser/goreleaser/releases/download/v${GORELEASER_VERSION}/goreleaser_Linux_x86_64.tar.gz \
+    | tar xzf - -C $HOME/go/bin goreleaser
+
+# TODO: upgrade go here (as per install.sh)
+
+# setup environment for build
```
brycemcanally, line 22: Why can't the build steps be:
```
make install
export VERSION=$(pachctl version --client-only)
make docker-build
make docker-push
```

jrockway: Yeah, I think I can clean some of that up.  The Makefile hard-codes a couple of things that don't work, `DOCKER_BUILDKIT=1` doesn't work on Circle's `docker` executor, but I can switch to the machine executor for consistency with the rest of CI.  (I think it's slower and more expensive, but I guess I don't really care.)

`make docker-push` pushes etcd, which we don't actually build.  The version is also hard-coded there, so if there is a pachyderm/etcd built, that overwrites what users get in production, which is super worrying to me.  I expect it to flat-out fail, but I'll try it and see.

The thing that I most wonder about is why I copied all the version-defining variables out of the Makefile.  That can definitely go away 😂 

(In general I think we should get away from pushing to pachyderm/* from CI, except for actual releases, which I don't think we actually do on CI.  Other people do it, but the risk of changing what people are running in production is high.  Even having it in the Makefile feels pretty risky to me, local developers are already "docker login"'d to Docker Hub and have permission to write to the pachyderm repo.

Something that would mitigate this is forcing the helm charts to identify docker images by the immutable sha256 instead of a mutable tag, but that requires new tooling to some extent.  Users will want to map the sha256 to a release number to see what they're actually running, and the Helm chart would probably have to translate a human-readable release to the sha256.  What a mess!)

brycemcanally: Yeah, if we could clean it up a bit, that would be nice. Also, I agree with your concerns, although the etcd push issue seems like it would be unlikely to happen by accident since it requires a build with that tag that isn't the right build for that version. Either way, we should have a safeguard to prevent that from happening.

jrockway: I've adjusted this to use a `machine` executor, and as a result `make docker-build` and `make docker-push` work.   The machine executor is slower, adding about 4 minutes to the CI cycle.  

The only thing this script duplicates that exists elsewhere is installing go and installing goreleaser.  (I could switch to using install.sh and adding the appropriate caching.  It would hurt runtime, though, installing a bunch of stuff we don't need.  The caching doesn't really help that -- downloading from CircleCI's cache is as slow as downloading from the Internet.  Each executor seems to be very network-constrained, so unless your caching saves CPU time, it doesn't seem to be worth it except at odd hours when the network isn't contended.  Even then, it's a maybe.)

brycemcanally (approved):
lgtm

Issue #6660, opened 2021-08-03
Random number handling in Helm Chart


chainlink:
Random number generation in helm has some real pitfalls:
- `randAlphaNum` will get regenerated on update, causing generated values to become out of sync (ie, if they've already been pulled into an app) - this can be overcome somewhat with isInstall
- When used in `_helpers` it generates a new value everytime a named template is called, leading to different values being used throughout the chart

[AB#30](https://dev.azure.com/PachydermCompany/92b98a32-88e5-46cb-a28f-5d064712e03b/_workitems/edit/30)

PR #6691, opened 2021-08-05
Create a helper template for stable random values

acohen4:
In this solution, the current timestamp is used to compute a hash. 

There's an obvious, though probabilistically unlikely problem of having the time change between invocations of this template within the script, where the resulting hash wouldn't be stable.


acohen4:
closes https://github.com/pachyderm/pachyderm/issues/6660

etc/helm/pachyderm/templates/_helpers.tpl
```diff
@@ -24,13 +24,28 @@ LOCAL
...
+## truncate the date to the minute and use 'genPrefix' to produce a different hash per useage
+## expects a context containing 'defaultValue' and 'hashSalt' keys
+{{- define "defaultOrStableHash" -}}
+{{- if .defaultValue }}
+    {{- .defaultValue }}
+{{- else }}
+    {{- if .OnlyDefault }}
+        {{- fail "must provide a default value during release upgrade." }}
+    {{- else }}
+        {{- derivePassword 1 "long" (now | toString | trunc 16) .hashSalt "pachyderm" -}}
```
chainlink, line 48: Can you add some docs here? (ie, talking about the date truncation, perhaps referencing the article we found this in.

etc/helm/pachyderm/templates/_helpers.tpl
```diff
@@ -24,13 +24,44 @@ LOCAL
...
+## is because we want to this template to produce the same hash each 
+## time it's executed in a release. In this case, the current timestamp 
+## gives us a unique enough value to use as a hash seed. We truncate it 
+## to the minute to reduce the odds that it will produce different timestamp 
+## values between renderings of this template.
+## Reference https://github.com/helm/helm/issues/6456
+##
+## 2.) 'onlyDefault' is used to force this template to fail if a default value is not provided
+{{- define "defaultOrStableHash" -}}
+{{- if .defaultValue }}
```
chainlink, line 58: This method might be a little "too smart" for helm (specifically passing in a hash as an arg) Generally with helm it's ok to break DRY to make things more clear / use values directly.

chainlink (changes requested):
This looks good, but we need a couple of changes on the testing side.

I think we should ship this PR, but the more I think about it, I think we should fix the root issue in the config pod before RC (being able to reference the secret values from Env vars/ secrets instead of embedding them in the config)

PR #6694, opened 2021-08-05
Adds completions for jobsets.

### LINKED ISSUE #6661: `pachctl shell` needs an update now that some pachctl commands have been modified
The changes in some of our pachctl commands are not reflected in `pachctl shell` or `pachctl completion`:

For example, in the case of `pachctl shell`: `inspect job` suggests jobIDs when it now requires pipeline@jobID -> results in an ```invalid format "fd9454d06d8e4fa38a75c8cd20b39538": expected pipeline@job-id```
Screen Shot 2021-08-03 at 4 18 47 PM

jdoliner:
Fixes #6661

### DIFF
--- a/src/server/cmd/pachctl/shell/completions.go
+++ b/src/server/cmd/pachctl/shell/completions.go
@@ -9,10 +9,12 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/gogo/protobuf/types"
 	"github.com/pachyderm/pachyderm/v2/src/client"
 	"github.com/pachyderm/pachyderm/v2/src/internal/clientsdk"
 	"github.com/pachyderm/pachyderm/v2/src/internal/cmdutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/ppsutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/pretty"
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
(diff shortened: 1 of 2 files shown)


src/server/cmd/pachctl/shell/completions.go
```diff
@@ -215,6 +217,50 @@ func PipelineCompletion(_, _ string, maxCompletions int64) ([]prompt.Suggest, Ca
...
+		}
+	}
+	return fmt.Sprintf("%s, %d subjobs %d failure(s)", pretty.Ago(created), len(jsi.Jobs), failure)
+}
+
+// JobCompletion completes job parameters of the form <job-set>
+func JobSetCompletion(_, text string, maxCompletions int64) ([]prompt.Suggest, CacheFunc) {
+	c := getPachClient()
+	var result []prompt.Suggest
+	listJobSetClient, err := c.PpsAPIClient.ListJobSet(c.Ctx(), &pps.ListJobSetRequest{})
```
brendoncarroll, line 248: If this is for autocomplete, should this context have a timeout? 3-5 seconds maybe?

brendoncarroll (approved):
LGTM. One suggestion about the context, but that might be set elsewhere.

src/server/cmd/pachctl/shell/completions.go
jdoliner: I believe that's typically set by the caller. I could maybe see the argument that the completion layer should set its own lower timeout so as not to block things for too long. Because if the completion function blocks then the shell just locks up. But it's a little hard to know what the right timeout would be and I think what we have now has been working reasonably well so I'm inclined to just leave it.

PR #6703, opened 2021-08-07
Prevent ListPipeline from holding an open query.

PFedak:
ListPipeline sent pipeline infos on a channel to have their details
fetched and possibly be sent to the grpc caller. This kept the list
pipeline SQL query open for a potentially long time, and had issues
around managing the channel when errors were encountered. A smaller
issue was that pipelines were returned in a random order.

Instead, fetch all pipeline infos at once from the database. Pipeline
info objects should be relatively small, so it's unlikely that this
will cause memory issues anytime soon. In addition, we canprocess all
infos in order after their details are fetched.

### DIFF
--- a/src/server/pps/server/api_server.go
+++ b/src/server/pps/server/api_server.go
@@ -45,6 +45,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/ppsutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/serde"
 	"github.com/pachyderm/pachyderm/v2/src/internal/serviceenv"
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/chunk"
 	tu "github.com/pachyderm/pachyderm/v2/src/internal/testutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/tracing"
 	"github.com/pachyderm/pachyderm/v2/src/internal/tracing/extended"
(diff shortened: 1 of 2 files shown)


brycemcanally:
I think this is a pretty good candidate for an abstraction we have in the chunk layer called the `TaskChain`. The basic pattern is a set of tasks you want to run in parallel, but complete in the order they were created. This is managed by running a goroutine per task and using internal channels to ensure a callback, that is provided with the task, is executed in order with respect to the task creation order. So, in this case you could create a task for each pipeline info that fetches the details, and the callback could just be `f`. Also, we should probably filter before fetching details. We can chat offline if that would help.

PFedak:
TaskChain sounds good, I was surprised there wasn't just a standard way of doing this. The only issue with the order of filtering vs details is in principle the filtering could rely on details being fetched. Right now it's only used through the CLI for state, but arbitrary `jq` filters are exposed through the API and I don't feel great about limiting it just for performance.

src/server/pps/server/api_server.go
```diff
@@ -2391,55 +2403,30 @@ func (a *apiServer) listPipeline(ctx context.Context, request *pps.ListPipelineR
...
-				}(); err != nil {
+
+	// fetch details and filter via a task chain so that order is preserved
+	chain := chunk.NewTaskChain(ctx)
+
+	// spin up goroutines to get the PFS info
+	limiter := limit.New(5)
+	for i := range infos {
+		limiter.Acquire()
+		chain.CreateTask(func(ctx context.Context, serial func(func() error) error) error {
```
brycemcanally, line 2414: We need to check the error here. It can return an error if the context is canceled.

src/server/pps/server/api_server.go
```diff
@@ -2391,55 +2403,30 @@ func (a *apiServer) listPipeline(ctx context.Context, request *pps.ListPipelineR
...
-					}
-					return nil
-				}(); err != nil {
+
+	// fetch details and filter via a task chain so that order is preserved
+	chain := chunk.NewTaskChain(ctx)
+
+	// spin up goroutines to get the PFS info
+	limiter := limit.New(5)
+	for i := range infos {
```
brycemcanally, line 2412: You need to lift the scope here. This should do it:
```
for i := range infos {
    i := i
    ...
```

brycemcanally (approved):
lgtm, just some minor changes needed.

src/server/pps/server/api_server.go
PFedak: ah whoops, had that in the previous version

PR #6704, opened 2021-08-09
Using exponential backoff instead of busy-loop in NoErrorWithinTRetry()

acohen4:


src/internal/require/require.go
```diff
@@ -419,21 +421,11 @@ func NoErrorWithinT(tb testing.TB, t time.Duration, f func() error, msgAndArgs .
...
-				close(doneCh)
-				break
-			}
-		}
-	}()
-	select {
-	case <-doneCh:
-	case <-time.After(t):
-		timeout = true
+	ctx, _ := context.WithTimeout(context.Background(), t)
```
brycemcanally, line 424: We should still have the cancel in a defer in the case of an error before the timeout.

src/internal/require/require.go
```diff
@@ -419,21 +421,11 @@ func NoErrorWithinT(tb testing.TB, t time.Duration, f func() error, msgAndArgs .
...
+	err := backoff.RetryUntilCancel(ctx, f, backoff.NewExponentialBackOff(), func(err error, d time.Duration) error {
```
brycemcanally, line 425: You can just use `nil` for the notify function.

brycemcanally (approved):
Few minor changes, then lgtm. Also, it looks like there is a circular dependency between the backoff and require packages. You might be able to address that by splitting the require package testing into a separate package (you should be able to do this by just adding `_test` to the package name for the test files https://pkg.go.dev/cmd/go/internal/test).

PR #6705, opened 2021-08-09
Fix a few places that are referencing the wrong commit state

brycemcanally:
This PR fixes a few places that are referencing the wrong commit state (finishing vs finished) and changes lazy / empty download to use full paths (the semantics of 1.x) rather than relative.


acohen4 (approved):
LGTM

Issue #6713, opened 2021-08-10
Pipeline Specs Get Mixed Up When Creating Inside of a Transaction


dgeorg42:
Ran the following to create 5 customer pipelines:
```
pachctl start transaction
pachctl create pipeline -f ../pipe-2.0/dualfan/data_source_dualfan_list_years.json
pachctl create pipeline -f ../pipe-2.0/dualfan/dualfan_location_asset_assignment.json
pachctl create pipeline -f ../pipe-2.0/dualfan/dualfan_data_location_group.yaml
pachctl create pipeline -f ../pipe-2.0/dualfan/dualfan_structure_repo_by_location.json
pachctl create pipeline -f ../pipe-2.0/dualfan/dualfan_merge_data_by_location.json
pachctl finish transaction
```

After it ran, I inspected the `dualfan_data_location_group` pipeline, and the spec was wrong.  Specifically, it was actually the spec for one of the other pipelines created as part of the transaction. This was reproducible many times.  If I run the `pachctl create` commands _without_ the transaction, though, the pipelines are all created correctly.

[AB#82](https://dev.azure.com/PachydermCompany/92b98a32-88e5-46cb-a28f-5d064712e03b/_workitems/edit/82)

PR #6714, opened 2021-08-10
Fix an issue with transactional create pipeline

brycemcanally:
This PR fixes an issue that occurs when creating pipelines in a transaction where the pipelines all have the same spec file (https://github.com/pachyderm/pachyderm/issues/6713). The issue appears to be because a hash is computed for the spec file sets created in the transaction, but this hash doesn't actually use the disambiguating data (path + data) because the sum isn't actually called on the hasher, but rather the sum exposed by the hashing package (md5). This would cause all of the file sets to have the same hash, so the first spec file set would get used for the rest of the pipelines that get created.

### DIFF
--- a/src/internal/transactionenv/env.go
+++ b/src/internal/transactionenv/env.go
@@ -2,6 +2,7 @@ package transactionenv
 
 import (
 	"context"
+	"time"
 
 	"github.com/gogo/protobuf/proto"
 	"github.com/jmoiron/sqlx"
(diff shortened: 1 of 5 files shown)


PFedak (approved):
Ah, that's embarrassing. I'll send you a PR fixing the test to catch this unless you feel like doing that

PFedak:
We've decided to solve this in a different way, see #6735

PR #6721, opened 2021-08-12
Remove branches from propagater on deletion.

PFedak:
Deleting a branch will modify the provenance of all downstream branches.
This in turn marks those branches for propagation. Depending on the
order that delete all uses, this can end up attempting to propagate an
already-deleted branch. While we could simply ignore not found errors,
it seems safer to ensure the branch set is accurate.


Tryneus (approved):
LGTM, nice catch!

PR #6722, opened 2021-08-12
Fix #6582: Reword PPS master messages & log @debug

### LINKED ISSUE #6582: pachd Looking to Scale Up Workers Every 2 Seconds
After creating a pipeline called "benchmarking", when I look at the `pachd` logs, I see the following messages logged every 2 seconds:
```
2021-07-21T21:45:47Z INFO PPS master: processing event for "benchmarking"
2021-07-21T21:45:47Z INFO pfs.API.GetFileTAR {"request":{"file":{"commit":{"branch":{"repo":{"name":"benchmarking","type":"spec"},"name":"master"},"id":"7a732661c11e426c9dbe81ac7b0154d8"},"path":"spec"}}} 
2021-07-21T21:45:47Z INFO pfs.API.GetFileTAR {"duration":0.013665573,"request":{"file":{"commit":{"branch":{"repo":{"name":"benchmarking","type":"spec"},"name":"master"},"id":"7a732661c11e426c9dbe81ac7b0154d8"},"path":"spec"}}} 
2021-07-21T21:45:47Z INFO PPS master: scaling up workers for "benchmarking"
```
If I'm interpreting these messages correctly, there's an event every 2 seconds which causes `pachd` to query the `spec` repo for the pipeline and then determine if it needs to scale up workers.  Which would make sense, except that there are no jobs running, there aren't even any files in the input repo, the pipeline doesn't have autoscaling enabled, and there's already a single worker available, so it's not clear to me why this event would need to be triggered at all, let alone be triggered every 2 seconds.

msteffen:


PFedak (approved):
LGTM, excited for cleaner logs :)

PR #6723, opened 2021-08-12
Make DeletePipeline handle missing repos again.

PFedak:


PFedak:
Seems like some more parts of the delete could/should be in a transaction

src/server/pps/server/api_server.go
```diff
@@ -2535,20 +2540,23 @@ func (a *apiServer) deletePipeline(ctx context.Context, request *pps.DeletePipel
...
-					return pachClient.DeleteRepo(input.Cron.Repo, request.Force)
-				})
-			}
-			return nil
-		})
-	}
-	if err := eg.Wait(); err != nil {
-		return err
+	// only do this if we were able to read the spec file
+	if pipelineInfo.Details != nil {
```
acohen4, line 2544: I think having this makes logical sense. I'm only confused about the pickle the user may be in if we fail to get the pipeline details, but there are in fact cron repos to delete. I guess in that case they would need to manually debug anyway

src/server/pps/server/api_server.go
```diff
@@ -2558,28 +2566,30 @@ func (a *apiServer) deletePipeline(ctx context.Context, request *pps.DeletePipel
...
+	if !missingRepo {
+		if !request.KeepRepo {
+			// delete the pipeline's output repo
+			if _, err := a.env.PfsServer().DeleteRepo(ctx, &pfs.DeleteRepoRequest{
+				Repo:  client.NewRepo(request.Pipeline.Name),
+				Force: request.Force,
+			}); err != nil && !errutil.IsNotFoundError(err) {
+				return errors.Wrap(err, "error deleting pipeline repo")
+			}
+		} else if pipelineInfo.Details != nil {
```
acohen4, line 2578: should we add a final else that throws an error?

acohen4 (approved):
Left a few comments, but overall looks good!

src/server/pps/server/api_server.go
PFedak: Yeah, I think if the spec really is gone for whatever reason we're stuck. It is the case that we create cron repos with a certain name pattern, but that doesn't seem safe enough. If I get my cron repos -> system repos change in this goes away, anyway

src/server/pps/server/api_server.go
PFedak: Good idea

msteffen:
All tests are green, so I'm merging this PR so I can merge https://github.com/pachyderm/pachyderm/pull/6722

PR #6724, opened 2021-08-12
Don't Filter PFS Logs

brendoncarroll:
We were previously filtering any timestamped logs because travis couldn't handle all of the logs output during a test run.
Circle doesn't seem to have a problem.  It won't display them all in the browser, but you can download the full output.
This will let use see which test is hanging better than a mysterious timeout.


PFedak (approved):
LGTM

PR #6732, opened 2021-08-16
Fix fuzz provenance

brendoncarroll:


chainlink (approved):
LGTM (pending passing tests)

PFedak:
I'm pretty sure this is working as is, I added the loop below to finish everything, not just the commit we create explicitly. Squash commit would almost certainly fail if something was left open, as it was before my change

PFedak (approved):
LGTM, I tried fixing the problem we discussed in #6741 but that wasn't quite enough, so I might dig into it a bit more once I have everything else in

PR #6734, opened 2021-08-16
Define inter-file random values using K8s Secrets

acohen4:
closes https://github.com/pachyderm/pachyderm/issues/6660


etc/helm/pachyderm/templates/pachd/config-job.yaml
```diff
@@ -21,6 +21,17 @@ spec:
         volumeMounts:
         - name: config
           mountPath: "/pachConfig"
+        env:
+        - name: CONSOLE_OAUTH_CLIENT_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: console-secret
+              key: OAUTH_CLIENT_SECRET
+        - name: CLUSTER_DEPLOYMENT_ID
```
chainlink, line 30: If the env var is named the same, an `envFrom` block would be more terse here.

etc/helm/pachyderm/templates/pachd/deployment-id-secret.yaml
```diff
@@ -0,0 +1,14 @@
+{{- /*
+SPDX-FileCopyrightText: Pachyderm, Inc. <info@pachyderm.com>
+SPDX-License-Identifier: Apache-2.0
+*/ -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    app: deployment-id-secret
```
chainlink, line 9: app should be `pachd` (Should match the pachd deployment)

etc/helm/pachyderm/templates/pachd/deployment-id-secret.yaml
```diff
@@ -0,0 +1,14 @@
...
+    suite: pachyderm
+  name: deployment-id-secret
```
chainlink, line 11: I'd prefix this with pachyderm to make it clear who the secret belongs to when getting secrets from the namespace

etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -111,7 +111,10 @@ spec:
               resource: requests.memory
         {{- end }}
         - name: CLUSTER_DEPLOYMENT_ID
-          value: {{ include "pachyderm.clusterDeploymentId" . | quote }}
```
chainlink, line 114: This should use `envFrom` instead as it's more terse

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -3,6 +3,8 @@ SPDX-FileCopyrightText: Pachyderm, Inc. <info@pachyderm.com>
 SPDX-License-Identifier: Apache-2.0
 */ -}}
 {{ if .Values.pachd.enterpriseLicenseKey }}
+{{- $enterpriseSecret := include "pachyderm.enterpriseSecret" . -}}
```
chainlink, line 6: I'd declare enterprise secret directly here to ensure the template isn't misused in multiple places

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -3,6 +3,8 @@ SPDX-FileCopyrightText: Pachyderm, Inc. <info@pachyderm.com>
...
+{{- $oidcSecret := include "pachyderm.pachClientSecret" . -}}
```
chainlink, line 7: Same here (declare directly here vs in helpers)

etc/helm/pachyderm/values.yaml
```diff
@@ -198,6 +198,7 @@ pachd:
   enterpriseLicenseKey: ""
   rootToken: ""
   enterpriseSecret: ""
+  oauthClientSecret: ""
```
chainlink, line 201: Need to run `make schema` to add this to the json schema

etc/helm/test/local_test.go
```diff
@@ -71,6 +73,17 @@ func TestLocal(t *testing.T) {
 				continue
 			}
 			checks["headless service"] = true
+
+		case *v1.Secret:
+			if object.Name == "pachyderm-config" {
```
chainlink, line 78: We might want to consider renaming the secret to `pachyderm-auth-config` to make way for a global `pachyderm-config` secret object

chainlink (changes requested):
Looks good! I've preposed a few changes proposed to tightening things up a bit

etc/helm/pachyderm/values.yaml
acohen4: It should be included in this PR

etc/helm/pachyderm/templates/pachd/config-job.yaml
acohen4: fair point. My thinking was that this would make it easier to see all the Variables that are used by the config-pod

etc/helm/test/local_test.go
acohen4: this is an interesting one. Although most of the config-pod data is auth related (IDP connectors, etc...), it doesn't have to be strictly so. So I am wondering whether a future `pachyderm-config` may still be related to the config-pod. In the case that it's not, maybe we could call this one `pachyderm-config-job-config` ?

acohen4: based on conversation in slack, we'll go with `pachyderm-bootstrap-config`

etc/helm/pachyderm/templates/pachd/deployment-id-secret.yaml
```diff
@@ -0,0 +1,14 @@
+{{- /*
```
chainlink, line 1: Do we need this under enterprise server only deployments as well? (`templates/enterprise-server`)

etc/helm/pachyderm/templates/console/secret.yaml
```diff
@@ -12,5 +12,5 @@ metadata:
   name: console-secret
```
chainlink, line 12: Future TODO: rename to `pachyderm-console-secret`

chainlink (approved):
Looking good! (with one small note on enterprise only installs) 

If you haven't done so, can you test locally to make sure everything comes up correctly? (Console, etc)

etc/helm/pachyderm/templates/console/secret.yaml
```diff
@@ -12,5 +12,9 @@ metadata:
...
   namespace: {{ .Release.Namespace }}
 data:
+{{- if .Release.IsUpgrade }}
+  OAUTH_CLIENT_SECRET: {{ .Values.console.config.oauthClientSecret | toString | b64enc | quote }}
```
chainlink, line 16: Can we add a required to the IsUpgrade code path to make sure it's set?

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -3,8 +3,15 @@ SPDX-FileCopyrightText: Pachyderm, Inc. <info@pachyderm.com>
 SPDX-License-Identifier: Apache-2.0
 */ -}}
 {{ if .Values.pachd.enterpriseLicenseKey }}
-{{- $enterpriseSecret := default (randAlphaNum 32) .Values.pachd.enterpriseSecret -}}
-{{- $oidcSecret := default (randAlphaNum 32) .Values.pachd.oauthClientSecret -}}
+{{- $oidcSecret := "" -}}
+{{- $enterpriseSecret := "" -}}
+{{- if .Release.IsUpgrade -}}
+{{- $oidcSecret = .Values.pachd.oauthClientSecret -}}
+{{- $enterpriseSecret = .Values.pachd.enterpriseSecret -}}
```
chainlink, line 10: Same here (add required to make sure the value is set in the upgrade case

chainlink (approved):
Couple small changes, pre approving :)

PR #6735, opened 2021-08-16
Store pipeline specs in postgres.

PFedak:


src/internal/ppsdb/ppsdb.go
```diff
@@ -17,7 +19,18 @@ const (
 	jobsCollectionName      = "jobs"
 )
 
-var pipelinesIndexes = []*col.Index{}
+// PipelinesVersionIndex records the version numbers of pipelines
+var PipelinesVersionIndex = &col.Index{
+	Name: "version",
+	Extract: func(val proto.Message) string {
+		info := val.(*pps.PipelineInfo)
+		return fmt.Sprintf("%s@%d", info.Pipeline.Name, info.Version)
```
brendoncarroll, line 27: The number needs to 0 padded for it to be lexicographically sorted.

src/server/cmd/worker/main.go
```diff
@@ -46,18 +47,16 @@ func getPipelineInfo(pachClient *client.APIClient, env serviceenv.ServiceEnv) (*
...
-	pachClient.SetAuthToken(pipelineInfo.AuthToken)
-	// Notice we use the SpecCommitID from our env, not from etcd. This is
-	// because the value in etcd might get updated while the worker pod is
+	// Notice we use the SpecCommitID from our env, not from postgres. This is
+	// because the value in postgres might get updated while the worker pod is
 	// being created and we don't want to run the transform of one version of
 	// the pipeline in the image of a different verison.
-	pipelineInfo.SpecCommit.ID = env.Config().PPSSpecCommitID
-	if err := ppsutil.GetPipelineDetails(pachClient, pipelineInfo); err != nil {
+	fmt.Sprintf("%s@%s", env.Config().PPSPipelineName, env.Config().PPSSpecCommitID)
```
acohen4, line 54: unintentional addition?

src/server/cmd/worker/main.go
```diff
@@ -46,18 +47,16 @@ func getPipelineInfo(pachClient *client.APIClient, env serviceenv.ServiceEnv) (*
...
-		return nil, err
-	}
-	pachClient.SetAuthToken(pipelineInfo.AuthToken)
-	// Notice we use the SpecCommitID from our env, not from etcd. This is
-	// because the value in etcd might get updated while the worker pod is
+	// Notice we use the SpecCommitID from our env, not from postgres. This is
+	// because the value in postgres might get updated while the worker pod is
 	// being created and we don't want to run the transform of one version of
 	// the pipeline in the image of a different verison.
-	pipelineInfo.SpecCommit.ID = env.Config().PPSSpecCommitID
```
acohen4, line 57: did you mean to keep this line's functionality? The comment suggests that it should still be here

src/server/pps/server/s3g_sidecar.go
```diff
@@ -58,17 +59,19 @@ func (a *apiServer) ServeSidecarS3G() {
...
+				return err
+			}
+			return a.pipelines.ReadWrite(txnCtx.SqlTx).Get(specCommit, s.pipelineInfo)
+		}); err != nil {
 			return errors.Wrapf(err, "sidecar s3 gateway: could not find pipeline")
 		}
 
 		// Set auth token for s.pachClient (pipelinePtr.AuthToken will be empty if
 		// auth is off)
 		s.pachClient.SetAuthToken(s.pipelineInfo.AuthToken)
```
acohen4, line 74: do we still need to set the AuthToken here?

src/server/pps/server/api_server.go
```diff
@@ -2566,9 +2502,15 @@ func (a *apiServer) deletePipeline(ctx context.Context, request *pps.DeletePipel
 		}
 	}
 
-	// Delete PipelineInfo
+	// Delete all past PipelineInfos
 	if err := dbutil.WithTx(ctx, a.env.GetDBClient(), func(sqlTx *sqlx.Tx) error {
-		return a.pipelines.ReadWrite(sqlTx).Delete(request.Pipeline.Name)
+		for version := pipelineInfo.Version; version > 0; version-- {
```
acohen4, line 2507: So pipelines start at version 1? Will there ever be gaps between existing pipeline version numbers?

src/server/pps/server/api_server.go
```diff
@@ -2370,74 +2318,52 @@ func (a *apiServer) listPipeline(ctx context.Context, request *pps.ListPipelineR
...
-			return err
-		}
-	} else {
-		if err := a.pipelines.ReadOnly(ctx).Get(pipeline.Name, p); err != nil {
-			if col.IsErrNotFound(err) {
-				return errors.Errorf("pipeline \"%s\" not found", pipeline.Name)
+		} else {
+			// we haven't seen this pipeline yet, rely on sort order and assume this is latest
+			var lastVersionToSend uint64
+			if history < 0 || uint64(history) > p.Version {
```
acohen4, line 2350: I think the second predicate here should be ">=", right?

src/server/pps/server/s3g_sidecar.go
PFedak: I think you're a better person to answer that. We don't need it in this function, but I assumed the s3g sidecar would rely on auth elsewhere.

src/server/cmd/worker/main.go
PFedak: Ah, nice catch. I did, but the condition looks different now. The `Sprintf` I didn't hook up should be the collection key that's used

src/server/pps/server/api_server.go
PFedak: They start at 1, yes. This is definitely something that could be fragile, at the very least "not found" errors should be ignored here if they aren't already by the collection

src/server/pps/server/api_server.go
PFedak: Ah probably true, although I think it works as is because both 1 and 0 will send everything. In any case I should add a test for this to make sure.

PFedak: Ah probably true, although I think it works as is because both 1 and 0 will send everything. In any case I should add a test for this to make sure.

src/internal/ppsdb/ppsdb.go
PFedak: Ah true, I don't think I'm relying on that, but couldn't hurt. 9999 pipeline versions ought to be enough for anyone

src/server/pps/server/api_server.go
PFedak: This now deletes based on an index

src/internal/middleware/auth/handler.go
```diff
@@ -75,6 +75,10 @@ func GetWhoAmI(ctx context.Context) string {
 	return ""
 }
 
+func ClearWhoAmI(ctx context.Context) context.Context {
```
acohen4, line 78: 👍

src/server/pps/server/s3g_sidecar.go
acohen4: I'm not immediately sure whether s3g will end up using this value. I need to look into it. In the meantime leaving it is fine

src/internal/ppsdb/ppsdb.go
brycemcanally: I think we should bump the padding a bit higher. 9999 versions is probably fine in the majority of cases, but there isn't really a downside to bumping this higher. I think 8 would be high enough to make it never be an issue.

src/internal/ppsdb/ppsdb.go
```diff
@@ -26,7 +55,23 @@ func Pipelines(db *sqlx.DB, listener col.PostgresListener) col.PostgresCollectio
 		db,
 		listener,
 		&pps.PipelineInfo{},
-		nil,
+		pipelinesIndexes,
+		col.WithKeyGen(func(key interface{}) (string, error) {
+			if commit, ok := key.(*pfs.Commit); ok {
+				if commit.Branch.Repo.Type != pfs.SpecRepoType {
```
brycemcanally, line 61: This should be refactored into a helper function that takes a pipeline name and spec commit, and returns the key. Anything that accesses the pipeline collection through the name & commit key should use the helper function.

brycemcanally: Actually, since you are just relying on the spec repo name, then I guess it just needs to take a spec commit. I guess that explains why you create the key explicitly below since you don't have the spec commit, just the id. It would probably be better to get the spec commit and pass that to a key helper function rather than having the key defined in multiple places.

brycemcanally: Okay, so I was not aware that you could pass in either a string or a proto to the collection package, and it would attempt to handle both. That seems like something that could end up being too clever.

src/server/pps/server/api_server.go
```diff
@@ -812,24 +804,20 @@ func (a *apiServer) listJob(
...
 	// Override the SpecCommit for the pipeline to be what it was when this job
 	// was created, this prevents races between updating a pipeline and
 	// previous jobs running.
-	specCommit := client.NewSystemRepo(jobInfo.Job.Pipeline.Name, pfs.SpecRepoType).NewCommit("master", jobInfo.OutputCommit.ID)
-	pipelineInfo.SpecCommit = specCommit
-
-	// If the commits for the job have been squashed, the pipeline spec can no
-	// longer be read for this job
-	if err := ppsutil.GetPipelineDetails(a.env.GetPachClient(ctx), pipelineInfo); err != nil {
+	var pipelineInfo pps.PipelineInfo
```
brycemcanally, line 816: This should probably just be `pipelineInfo := &pps.PipelineInfo{}` simplify the code below.

src/server/pps/server/master.go
```diff
@@ -198,7 +203,15 @@ func (m *ppsMaster) attemptStep(ctx context.Context, e *pipelineEvent) error {
 	})
 	// we've given up on the step, check if the error indicated that the pipeline should fail
 	if err != nil && errors.As(err, &stepErr) && stepErr.failPipeline {
-		failError := m.a.setPipelineFailure(ctx, e.pipeline, fmt.Sprintf(
+		var specCommit *pfs.Commit
+		if err := m.a.txnEnv.WithReadContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
+			var err error
+			specCommit, err = m.a.findPipelineSpecCommit(txnCtx, e.pipeline, "")
```
brycemcanally, line 209: It seems like it would be better to have a transactional and non-transactional version of this so each call site that doesn't need the transaction won't have to set one up.

src/server/pps/server/pipeline_controller.go
```diff
@@ -132,17 +140,8 @@ func (m *ppsMaster) newPipelineOp(ctx context.Context, pipeline string) (*pipeli
 
 	// add pipeline auth
 	pachClient := m.a.env.GetPachClient(ctx)
+	op.ctx = auth_middleware.ClearWhoAmI(pachClient.Ctx())
```
brycemcanally, line 143: What is going on here?

src/server/pps/server/poller.go
```diff
@@ -258,18 +278,23 @@ func (m *ppsMaster) watchPipelines(ctx context.Context) {
 			if event.Err != nil {
 				return errors.Wrapf(event.Err, "event err")
 			}
+			parts := strings.SplitN(string(event.Key), "@", 2)
```
brycemcanally, line 281: This should be a call to a helper function that is also called by the pipeline key check.

src/server/pps/server/pipeline_controller.go
PFedak: This definitely deserves more of a comment, I'll put that in. We use the master context for pipeline operations but want it to be authorized as the pipeline itself. That's what the old code handles, but there's this extra layer where background tasks like the master don't have any auth of their own and might need to do operations requiring auth (but no permissions). To meet this need, they explicitly add the key WhoAmI uses to cache lookups and skip the token check. In order to switch to the pipeline auth, the old cached value should be removed.

src/server/pps/server/api_server.go
PFedak: Ah you're right, that does seem to be our prevailing style. I tend to default to making an empty struct and passing pointers but consistency is always better

src/server/pps/server/master.go
PFedak: Sure, I'll add that

src/server/pps/server/poller.go
```diff
@@ -205,20 +209,36 @@ func (m *ppsMaster) pollPipelinePods(ctx context.Context) {
...
+					}
+					return m.a.setPipelineCrashing(ctx, pipelineInfo.SpecCommit, reason)
+				}
 				for _, status := range pod.Status.ContainerStatuses {
 					if status.State.Waiting != nil && failures[status.State.Waiting.Reason] {
-						if err := m.a.setPipelineCrashing(ctx, pipelineName, status.State.Waiting.Message); err != nil {
+						if err := crashPipeline(status.State.Waiting.Message); err != nil {
 							return errors.Wrap(err, "error moving pipeline to CRASHING")
 						}
+						continue WatchLoop
```
brycemcanally, line 232: :+1:

src/internal/transactionenv/txncontext/context.go
```diff
@@ -50,11 +47,10 @@ func New(ctx context.Context, sqlTx *sqlx.Tx, authServer identifier, m FilesetMa
...
 	}
 	return &TransactionContext{
-		SqlTx:          sqlTx,
-		CommitSetID:    uuid.NewWithoutDashes(),
-		Timestamp:      types.TimestampNow(),
-		username:       username,
-		FilesetManager: m,
+		SqlTx:       sqlTx,
+		CommitSetID: uuid.NewWithoutDashes(),
+		Timestamp:   types.TimestampNow(),
```
brendoncarroll, line 52: We have the transaction, can we do `select CURRENT_TIMESTAMP` ?

src/server/pps/server/api_server.go
```diff
@@ -1909,44 +1890,27 @@ func (a *apiServer) CreatePipelineInTransaction(
...
-	update := false
-	if request.Update {
-		// check if the pipeline already exists, meaning this is a real update
-		if err := a.pipelines.ReadWrite(txnCtx.SqlTx).Get(request.Pipeline.Name, &pps.PipelineInfo{}); err == nil {
-			update = true
-		} else if !col.IsErrNotFound(err) {
-			return err
-		}
-	}
+	update := request.Update && oldPipelineInfo != nil
```
brycemcanally, line 1893: So I guess you can set the update flag without an existing pipeline info?

PFedak: This is meant to be equivalent to the old logic, I think the idea was setting "update" means "I don't care if it already exists" (vs erroring in that case), while this variable is whether we need to perform cleanup operations on the existing version.

src/internal/transactionenv/txncontext/context.go
PFedak: Ah, didn't realize SQL transactions had an associated timestamp, that makes sense to me

src/server/pps/server/api_server.go
```diff
@@ -1909,44 +1890,27 @@ func (a *apiServer) CreatePipelineInTransaction(
...
 	// Get the expected number of workers for this pipeline
-	parallelism, err := getExpectedNumWorkers(a.env.GetKubeClient(), newPipelineInfo)
-	if err != nil {
+	if parallelism, err := getExpectedNumWorkers(newPipelineInfo); err != nil {
 		return err
+	} else {
+		newPipelineInfo.Parallelism = uint64(parallelism)
 	}
 
-	if update {
```
brycemcanally, line 1936: :+1:

src/server/pps/server/api_server.go
```diff
@@ -1909,44 +1890,27 @@ func (a *apiServer) CreatePipelineInTransaction(
...
-		}
-	}
+	update := request.Update && oldPipelineInfo != nil
+
 	var (
 		// provenance for the pipeline's output branch (includes the spec branch)
 		provenance = append(branchProvenance(newPipelineInfo.Details.Input),
 			client.NewSystemRepo(pipelineName, pfs.SpecRepoType).NewBranch("master"))
 		outputBranch = client.NewBranch(pipelineName, newPipelineInfo.Details.OutputBranch)
 		statsBranch  = client.NewSystemRepo(pipelineName, pfs.MetaRepoType).NewBranch(newPipelineInfo.Details.OutputBranch)
```
brycemcanally, line 1900: Would you mind renaming this to metaBranch.

src/server/pps/server/parallelism_test.go
```diff
@@ -20,30 +19,28 @@ func wrap(t testing.TB, ps *pps.ParallelismSpec) *pps.PipelineInfo {
 }
 
 func TestGetExpectedNumWorkers(t *testing.T) {
-	kubeClient := tu.GetKubeClient(t)
-
 	// An empty parallelism spec should default to 1 worker
-	workers, err := getExpectedNumWorkers(kubeClient, wrap(t,
+	workers, err := getExpectedNumWorkers(wrap(t,
```
brendoncarroll, line 23: I can't find where this changed.  Why don't we need the kubeClient anymore?

src/server/pps/server/pipeline_controller.go
```diff
@@ -123,7 +125,13 @@ func (m *ppsMaster) newPipelineOp(ctx context.Context, pipeline string) (*pipeli
 	}
 	// get latest PipelineInfo (events can pile up, so that the current state
 	// doesn't match the event being processed)
-	if err := m.a.pipelines.ReadOnly(ctx).Get(pipeline, op.pipelineInfo); err != nil {
+	if err := m.a.txnEnv.WithReadContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
+		specCommit, err := m.a.findPipelineSpecCommit(txnCtx, pipeline, "")
+		if err != nil {
+			return err
+		}
+		return m.a.pipelines.ReadWrite(txnCtx.SqlTx).Get(specCommit, op.pipelineInfo)
```
brendoncarroll, line 133: Isn't this a Read only transaction?

src/server/pps/server/api_server.go
```diff
@@ -2664,7 +2639,7 @@ func (a *apiServer) StopPipeline(ctx context.Context, request *pps.StopPipelineR
 		if _, err := a.env.PfsServer().InspectRepoInTransaction(txnCtx, &pfs.InspectRepoRequest{
```
brycemcanally, line 2639: Couldn't the inspect pipeline operation below serve the same purpose as this inspect repo call?

src/server/pps/server/pipeline_controller.go
brendoncarroll: We should make sure the we are passing the read only option to `dbutil.WithTx` so that we get the full benefits of a read only transaction, and also so that writing causes an error.

src/internal/transactionenv/txncontext/context.go
brendoncarroll: It doesn't do anything fancy with transaction ordering/retrying.  It's just from the start of the transaction.  But I figure we should try to use the database clock where we can.

src/server/pps/server/api_server.go
PFedak: Ah, good point

src/server/pps/server/parallelism_test.go
PFedak: The function no longer uses it. I think we used to determine default parallelism by the number of nodes

src/server/pps/server/pipeline_controller.go
PFedak: handled these transactions in a more straightforward manner. Do you mean that for WithReadContext generally? We'd need to change how user-facing transactions work in that case since they rely on writes succeeding inside a "read only" transaction (just in the sense that it will not be committed). Probably not hard otherwise, though

src/server/pps/server/pipeline_controller.go
PFedak: Moved it to a separate function which hopefully makes things clearer

src/internal/ppsdb/ppsdb.go
brycemcanally: I still think we should have a key function that takes a spec commit, and that would be the only way to get the string encoding.

src/server/pps/server/pipeline_controller.go
brendoncarroll: Okay, maybe add a comment saying that it's not actually a read only transaction.  But this doesn't need to change in this P.R.

src/server/pps/server/api_server.go
```diff
@@ -2564,23 +2564,35 @@ func (a *apiServer) deletePipeline(ctx context.Context, request *pps.DeletePipel
...
-			// Remove branch provenance from output and meta
-			// we can only proceed if we have details
-			if _, err := a.env.PfsServer().CreateBranch(ctx, &pfs.CreateBranchRequest{
-				Branch: client.NewBranch(pipelineName, pipelineInfo.Details.OutputBranch),
-			}); err != nil && !errutil.IsNotFoundError(err) {
-				return err
+		} else {
+			// Remove branch provenance from output and then delete meta and spec repos
+			// this leaves the repo as a source repo, eliminating pipeline metadata
+			if err := a.txnEnv.WithWriteContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
```
brycemcanally, line 2570: We should refactor this such that all of the Postgres state change when deleting a pipeline happens in one transaction.

src/server/pps/server/pipeline_controller.go
PFedak: This bit is explicitly a ReadOnly access in the current version

brendoncarroll (approved):
I'm approving because I'm assigned to this P.R. and everything looks good.  All my feedback has been addressed.  Some of the other reviewers may still have concerns.

brycemcanally (approved):
lgtm, thanks for taking on this change.

PR #6739, opened 2021-08-17
Fix a couple symlink upload issues

brycemcanally:
This PR fixes the following issues with symlink uploading:
- Symlinks to directories that were not in PFS would not be uploaded. We now walk the resolved path when the symlink resolves to a path that is not under the PFS prefix (files downloaded from PFS).
- The meta commit would have incorrect state after a symlink upload that performed a copy file. This was due to a modification to the `*pfs.File` that was used for recording the input file. This was addressed by cloning the proto before performing the modification for the copy file.

This PR also adds logging to the PFS gRPC endpoints that do not have the standard gRPC logging we use.


acohen4 (approved):
LGTM

PR #6741, opened 2021-08-17
Close commits resulting from squash during fuzz test.

PFedak:


src/server/pfs/server/testing/server_test.go
```diff
@@ -5029,6 +5029,15 @@ func TestPFS(suite *testing.T) {
 				commit := commits[i]
 				commits = append(commits[:i], commits[i+1:]...)
 				require.NoError(t, env.PachClient.SquashCommitSet(commit.ID))
+				// some output branches might have had new commits added either because their old heads
+				// were squashed or because the head of a branch they are provenant on was squashed
+				for _, branch := range outputBranches {
```
brycemcanally, line 5034: This change isn't passing, but this change does:
```
ris, err := env.PachClient.ListRepo()
require.NoError(t, err)
for _, ri := range ris {
	bis, err := env.PachClient.ListBranch(ri.Repo.Name)
	require.NoError(t, err)
	for _, bi := range bis {
		branch := bi.Branch
		info, err := env.PachClient.InspectCommit(branch.Repo.Name, branch.Name, "")
		require.NoError(t, err)
		if info.Finishing == nil {
			require.NoError(t, finishCommit(env.PachClient, branch.Repo.Name, branch.Name, ""))
		}
	}
}
```
I'm not exactly sure why, but we should go ahead and get this in so we can have a green CI.

etc/testing/pfs_server.sh
```diff
@@ -5,3 +5,4 @@ export TIMEOUT=$1
 
 # grep out PFS server logs, as otherwise the test output is too verbose
 go test -v -count=1 ./src/server/pfs/server ./src/server/pfs/server/testing -timeout "$TIMEOUT" | grep -v "$(date +^%FT)"
+go test -v -count=10 ./src/server/pfs/server/testing -run FuzzProvenance
```
brycemcanally, line 8: Why was this added? Just to run it a few more times?

PFedak: Yeah, just wanted to be more sure

PR #6742, opened 2021-08-17
Add DropCommitSet RPC, disable squashing commits with no children

### LINKED ISSUE #6717: SquashCommitSet should not allow the head of a branch to be squashed
Because we define a squash as squashing the data in the child of a commit (as opposed to git which squashes with the parent commit), squashing a commit with no children implies data loss and is more like what we have been calling a 'drop' operation.  Therefore, we should make it an error to squash a commit with no children and add a new RPC `DropCommitSet` that implies data loss which can be used to remove such a commit.  This is also the RPC we would use to remove the commits in the middle of a commit chain, but implementing such behavior will likely take significant engineering effort, so it will be an error to call `DropCommitSet` on a commit with children until that is implemented.

[AB#87](https://dev.azure.com/PachydermCompany/92b98a32-88e5-46cb-a28f-5d064712e03b/_workitems/edit/87)

Tryneus:
This PR splits up the existing `SquashCommitSet` RPC into:
 * `DropCommitSet` - implies data loss, only currently supported on commits that have no children
 * `SquashCommitSet` - errors if there would be data loss, does not support squashing commits with no children

Fixes #6717.


src/server/pfs/server/driver.go
```diff
@@ -1336,16 +1336,10 @@ func (d *driver) listCommitSet(ctx context.Context, cb func(*pfs.CommitSetInfo)
 	})
 }
 
-func (d *driver) squashCommitSet(txnCtx *txncontext.TransactionContext, commitset *pfs.CommitSet) error {
+func (d *driver) squashCommitsInternal(txnCtx *txncontext.TransactionContext, commitInfos []*pfs.CommitInfo) error {
```
brycemcanally, line 1339: nit, `squashCommitsInternal` -> `squashCommitSetInternal`

src/server/pfs/server/driver.go
```diff
@@ -1486,10 +1480,59 @@ func (d *driver) squashCommitSet(txnCtx *txncontext.TransactionContext, commitse
...
+// commits in the commitSet have children the operation will fail.
+func (d *driver) dropCommitSet(txnCtx *txncontext.TransactionContext, commitset *pfs.CommitSet) error {
+	// Look up the commits in the CommitSet
+	commitInfos, err := d.inspectCommitSetImmediate(txnCtx, commitset)
+	if err != nil {
+		return err
+	}
+
+	for _, ci := range commitInfos {
+		if len(ci.ChildCommits) != 0 {
```
brycemcanally, line 1496: nit, I usually prefer > for checking against an empty slice since they cannot have negative lengths.

brycemcanally (approved):
lgtm, just some minor nits. This can be in a followup PR, but we need to move the delete job functionality into squash / drop commit set.

PR #6743, opened 2021-08-17
Only specify storageclass key if storage class specified

chainlink:
Blank value will disable dynamic provisioning


tybritten:
don't we need that here too? https://github.com/pachyderm/pachyderm/blob/master/etc/helm/pachyderm/templates/postgresql/statefulset.yaml#L67-L68

chainlink:
Good catch! I'll fix that up (but also, I'm nuking the built in postgres config shortly with bitnami lol)

tybritten:
Oh yeah haha

tybritten (approved):

PR #6753, opened 2021-08-18
Remove more references to assets.go

chainlink:


chainlink:
TODO
- Move secret volume mounts into worker RC from Assets Object storage client still pulls in via file in `src/internal/obj/factory.go`

src/server/pps/server/worker_rc.go
```diff
@@ -829,6 +830,23 @@ func (pc *pipelineController) createWorkerSvcAndRc(ctx context.Context, pipeline
 	return nil
 }
 
+// GetBackendSecretVolumeAndMount returns a properly configured Volume and
+// VolumeMount object given a backend.  The backend needs to be one of the
+// constants defined in pfs/server.
+func GetBackendSecretVolumeAndMount(backend string) (v1.Volume, v1.VolumeMount) {
```
acohen4, line 836: i think we can remove the backend parameter since it's unused

acohen4 (approved):
One Q but otherwise LGTM

Issue #6758, opened 2021-08-18
List commit doesn’t respect ancestry syntax


JimmyWhitaker:
**What happened?**:
List commit doesn’t respect ancestry syntax
```
$ pachctl list commit data
REPO BRANCH COMMIT                           FINISHED       SIZE ORIGIN DESCRIPTION
data master 55589fe63b3b46b19d4e8362a24e6d60 3 seconds ago  364B USER    
data master 2b8b243b5a654985a1ef6d22c6f6af72 8 seconds ago  0B   USER    
data master 619517fbf86f42a7a70c708679b89673 14 seconds ago 364B USER  
```
When you reference the previous commit, it still returns the same output. 
```
$ pachctl list commit data@master^
REPO BRANCH COMMIT                           FINISHED       SIZE ORIGIN DESCRIPTION
data master 55589fe63b3b46b19d4e8362a24e6d60 3 seconds ago  364B USER    
data master 2b8b243b5a654985a1ef6d22c6f6af72 8 seconds ago  0B   USER    
data master 619517fbf86f42a7a70c708679b89673 14 seconds ago 364B USER  
```

**What you expected to happen?**:
It should return the commits list for the commit referenced. 

**How to reproduce it (as minimally and precisely as possible)?**:

```
pachctl create repo data
touch file.txt
pachctl put file data@master -f file.txt
pachctl delete file data@master:/file.txt
pachctl put file data@master -f file.txt
pachctl list commit data@master
pachctl list commit data@master^
```

**Environment?**:
Pachyderm Hub 
```
COMPONENT           VERSION             
pachctl             2.0.0-beta.5        
pachd               2.0.0-beta.5 
```

[AB#110](https://dev.azure.com/PachydermCompany/92b98a32-88e5-46cb-a28f-5d064712e03b/_workitems/edit/110)

PFedak:
What do you expect the output of this command to be? It seems like you want to use the `--from` flag, as the help text doesn't list `pachctl list commit <fully-specified-commit>`  as an option, just `pachctl list commit <branch>` and `pachctl list commit <commit-id>`.

JimmyWhitaker:
I believe that's correct. I would expect the ID to resolve to the commit ID before passing it to the command. Otherwise we end up with some pretty gnarly pipe commands for examples. 

The behavior in 1.12.5 is: 

```
$ pachctl list commit data
REPO BRANCH COMMIT                           FINISHED       SIZE ORIGIN DESCRIPTION
[...]

PR #6761, opened 2021-08-19
Skip alias commits during traversal

brendoncarroll:
I'm not sure this is actually the best thing to do in all cases.
If the base commit is an alias commit, should that be skipped as well?
Is it worth adding a flag to InspectCommit to skip aliases?

### DIFF
--- a/src/server/pfs/server/driver.go
+++ b/src/server/pfs/server/driver.go
@@ -1000,6 +1000,9 @@ func (d *driver) resolveCommit(sqlTx *sqlx.Tx, userCommit *pfs.Commit) (*pfs.Com
 				}
 				return nil, err
 			}
+			if commitInfo.Origin.Kind == pfs.OriginKind_ALIAS {
+				i--
+			}
 			commit = commitInfo.ParentCommit
 		}
 	} else {


src/server/pfs/server/driver.go
```diff
@@ -1000,6 +1000,9 @@ func (d *driver) resolveCommit(sqlTx *sqlx.Tx, userCommit *pfs.Commit) (*pfs.Com
 				}
 				return nil, err
 			}
+			if commitInfo.Origin.Kind == pfs.OriginKind_ALIAS {
```
PFedak, line 1003: I assume we'd add this to the opposite direction, too

PFedak (commented):
We'd need special handling around the ancestry=0 case for this to work. I think we can assume all non-zero cases come from user input, meaning we want to do this. Internally, we expect this to give the head of a branch when passed a branch and otherwise be idempotent. I'm having trouble coming up with a command where someone expects it to resolve to the most recent non-alias, and where using the alias would give different results.

I think we need some concrete use cases to get a sense of what people expect. It makes sense to me that `diff file repo@branch repo@branch^` should always diff between two non-alias commits.  It's not clear to me from the description in #6758 what the interaction with alias commits would be. If it worked the way he wanted, you'd get different commits if you ran the list on different parts of the DAG, which seems reasonable to me.

In any case, maybe this just isn't a blocker anymore, assuming we can change the behavior after release.

PR #6766, opened 2021-08-19
Few miscellaneous changes

brycemcanally:
This PR contains a few miscellaneous changes:
- Remove the deadline for a gc run. It doesn't make sense to have a deadline, particularly a short one (5s), for a gc run that needs to run to completion.
- Call delete all before running the PPS load test in CI.
- Change symlink upload to only copy files from PFS when the empty file feature is used (this should be enabled for large files and disabled for small files). This provides an easy way to transition to the more performant small file symlink upload. It is more performant to upload small files from the local filesystem in a batch than a large number of small PFS copy file requests.

### DIFF
--- a/src/server/worker/datum/datum.go
+++ b/src/server/worker/datum/datum.go
@@ -405,36 +405,9 @@ func (d *Datum) handleSymlinks(mf client.ModifyFile, storageRoot string) error {
 		if fi.Mode()&os.ModeNamedPipe != 0 {
 			return nil
 		}
+		// Upload the local files if they are not from PFS.
 		if !strings.HasPrefix(file, d.PFSStorageRoot()) {
-			cb := func(dstPath, file string) (retErr error) {
-				f, err := os.Open(file)
-				if err != nil {
-					return err
-				}
-				defer func() {
-					if err := f.Close(); retErr == nil {
-						retErr = err
-					}
-				}()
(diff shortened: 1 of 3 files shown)


src/server/worker/datum/datum.go
```diff
@@ -447,12 +420,48 @@ func (d *Datum) handleSymlinks(mf client.ModifyFile, storageRoot string) error {
...
+		return mf.PutFile(dstPath, f, client.WithDatumPutFile(d.ID))
+	}
+	if fi.IsDir() {
+		dir := file
+		return filepath.Walk(dir, func(file string, fi os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if fi.IsDir() {
+				return nil
```
acohen4, line 453: @brycemcanally why are we returning nil here? my naive expectation would be that we search for files recursively

acohen4: also I realize this logic was here before, but figured it's a good time to ask

brycemcanally: We skip directories here because we cannot upload directories through PFS, they are created implicitly based on files. We upload the files, and if they have a directory prefix, the directories will be created. Empty directories will be skipped because PFS does not support empty directories.

acohen4 (approved):
LGTM

PR #6771, opened 2021-08-19
Don't silently throw away commit IDs or ancestry in CLI.

PFedak:
Commands that take an argument of form `repo[@branch]` parse their arguments using `ParseBranch`, which accepts commits such as `repo@ID`or `repo@branch~<ancestors>` without error, resulting in a "branch" with an empty name. While commands that require a branch name will error when the server attempts to validate them, other commands where a bare repo is valid may succeed, doing something unexpected.


PFedak:
This fixes a bug observed in #6758, making the indicated command error instead of doing something unintended. What remains is probably a feature request.

Tryneus (approved):
LGTM

PR #6772, opened 2021-08-20
Slight datum ID change

brycemcanally:
This PR changes the datum ID computation slightly to account for uses cases that involve performing a union of inputs with the same input name and file paths. The datum ID computation is now based on input name + repo name + branch name + file path. In a followup PR, we will need to prevent the creation of pipelines with multiple inputs that have the same input, repo, and branch name. Also, this PR changes the datum hash computation to include the datum ID and just the file hashes and pipeline salt. The reason why the datum ID is included is to leave the door open to global skipping (across arbitrary jobs for a pipeline) in the future, so we could just perform the lookup using the datum hash.

### DIFF
--- a/src/client/pps.go
+++ b/src/client/pps.go
@@ -2,8 +2,6 @@ package client
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"io"
 	"time"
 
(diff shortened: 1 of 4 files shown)


jdoliner (approved):
LGTM

PR #6773, opened 2021-08-20
Parallelize trigger tests

brycemcanally:
This PR parallelizes the trigger tests, which might be one of the main contributors to the more frequent PFS test timeouts recently. All of the trigger tests ran serially and several involve waiting that can add up to ~3mins in the runs I have looked at.

### DIFF
--- a/src/server/pfs/server/testing/server_test.go
+++ b/src/server/pfs/server/testing/server_test.go
@@ -5201,6 +5201,7 @@ func TestPFS(suite *testing.T) {
 		c := env.PachClient
 
 		t.Run("Simple", func(t *testing.T) {
+			t.Parallel()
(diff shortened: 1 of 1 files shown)


src/server/pfs/server/testing/server_test.go
```diff
@@ -5201,6 +5201,7 @@ func TestPFS(suite *testing.T) {
 		c := env.PachClient
```
brendoncarroll, line 5201: Should we set up a test environment per test?  I notice these occasionally break.  Is that because they share the same test environment?

brycemcanally: Yeah, I'll make that change.

brendoncarroll (approved):
LGTM

PR #6777, opened 2021-08-20
Change when job started time is set

brycemcanally:
This PR changes the job started time to be set when it transitions to the running state. Currently, we set it when it transitions to the starting state, which will cause the duration to include the time spent waiting on upstream jobs.


Tryneus (approved):
LGTM

PR #6778, opened 2021-08-20
Swap pipeline name and job id columns in job info pretty printing

brycemcanally:
This PR swaps the pipeline name and job id columns when pretty printing job info. This change was made because it is more consistent with the way we display commits and how you need to set up the parameters for requests that operate on a pipeline & job id.


Tryneus (approved):
LGTM

PR #6782, opened 2021-08-23
Fix job start time being reset during update

brycemcanally:
This PR fixes an issue where the job start time was reset by an update while the job is running. This happens because each update "transitions" the job state to running, which is when we set the started time. The fix is to check that we are transitioning from the starting state to the running state before setting the start time.


acohen4 (approved):
LGTM

acohen4 (approved):

PR #6784, opened 2021-08-23
Unskip TestAutoscalingStandby/ChainOf10

acohen4:


PFedak (approved):
LGTM, we discussed a bit about performance since the initial wait for pipelines to reach standby takes longer than expected, but that might require a more involved investigation

PR #6789, opened 2021-08-24
Fix fuzz provenance test

brycemcanally:
This PR is intended to fix the fuzz provenance test. The fuzz provenance test was consistently failing when trying to squash a commit that had an unfinished child. The way this can come about is the following (branches are numbers, commits are letters and are left to right from ancestor to descendant, alias commits have an apostrophe): 
```
Before drop c:
1: a, c
  \
    2: a, c
  /
3: a', b', c'
  \
    4: b
  /
5: b

After drop c, before propagation:
1: a
  \
    2: a
  /
3: a', b'
  \
    4: b
  /
5: b

After drop c, after propagation:
1: a, d'
  \
    2: a, d
  /
3: a', b', d'
  \
    4: b
  /
5: b
```

The fix is to ensure that all of the head commits of the branches in the system are finished after a squash / drop.


Tryneus (approved):
LGTM

Might be useful to add a comment to the new code to explain why it's finishing commits.

PR #6790, opened 2021-08-24
Alias commits need total filesets

brendoncarroll:
Prior to these changes, alias commits did not have any filesets associated with them.  When getting the fileset for an alias commit, the commit was skipped and the content was instead retrieved from the parent.  Now only open commits require reading the parent.  The read path for alias commits is the same as regular commits.

### DIFF
--- a/src/server/pfs/server/driver.go
+++ b/src/server/pfs/server/driver.go
@@ -2026,6 +2026,13 @@ func (d *driver) makeEmptyCommit(txnCtx *txncontext.TransactionContext, branchIn
 	if err := d.commits.ReadWrite(txnCtx.SqlTx).Create(commit, commitInfo); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


Tryneus (approved):
LGTM

src/server/pfs/server/driver.go
```diff
@@ -2018,6 +2026,13 @@ func (d *driver) makeEmptyCommit(txnCtx *txncontext.TransactionContext, branchIn
 	if err := d.commits.ReadWrite(txnCtx.SqlTx).Create(commit, commitInfo); err != nil {
 		return nil, err
 	}
+	total, err := d.storage.ComposeTx(txnCtx.SqlTx, nil, defaultTTL)
```
brycemcanally, line 2029: This should only be set when it is closed.

brycemcanally (approved):
lgtm

PR #6799, opened 2021-08-27
Periodic load tests

brycemcanally:
This PR sets up periodic load tests in CircleCI and adds visualizations for some of the steps we track.

There are 10 performance workloads that will each run in a separate bucket (to keep the environment somewhat hermetic) in CircleCI. The workloads each have a comment in their spec that briefly explains what is being tested. Each run will create a hub cluster, run a PFS and PPS workload with the same spec (and report times for each in the logs), then collect a debug dump and save it as a CircleCI artifact. The high level metric that we are tracking is the time for the CircleCI bucket and the debug dump has details for where the time was spent. The debug dump now has visualizations for time spent in steps that we track for each commit and job. These visualizations are stored similarly to the way commits and jobs are stored (in the associated repo / pipeline directories). The visualizations are charts created with go-chart and below is an example of each type.
image
image


etc/testing/circle/run_hub_tests.sh
```diff
@@ -35,14 +35,27 @@ make docker-build
 make docker-push
 
 # Create a workspace running the image we just built.
-hubcli --endpoint https://hub.pachyderm.com/api/graphql --apikey $HUB_API_KEY --op create-workspace-and-wait --orgid 2193 --loglevel trace --infofile workspace.json --version $VERSION --expiration 2h --description $CIRCLE_BUILD_URL --prefix "ci-"
+for i in {1..3}; do
+	hubcli --endpoint https://hub.pachyderm.com/api/graphql --apikey $HUB_API_KEY --op create-workspace-and-wait --orgid 2193 --loglevel trace --infofile workspace.json --version $VERSION --expiration 2h --description $CIRCLE_BUILD_URL --prefix "ci-" && break
```
brendoncarroll, line 39: Where is VERSION set.  Is this going to be the git hash?  Are the docker images already pushed at this point?

.circleci/config.yml
```diff
@@ -168,3 +190,25 @@ workflows:
   hub_live_tests:
     jobs:
       - hub
+  hub_load_tests:
+    #triggers:
```
brendoncarroll, line 194: Remember to re-enable before merging.

src/internal/pfsload/validator.go
```diff
@@ -138,7 +146,7 @@ func validate(client Client, commit *pfs.Commit, files []*file) (retErr error) {
 		}
 		files = files[1:]
 		return nil
-	})
+	}, true)
```
brendoncarroll, line 149: Is this the don't use all the memory flag?

src/server/pps/cmds/cmds.go
```diff
@@ -1094,11 +1095,12 @@ All jobs created by a pipeline will create commits in the pipeline's output repo
 	}
 	commands = append(commands, cmdutil.CreateAlias(listSecret, "list secret"))
 
+	var seed int64
 	runLoadTest := &cobra.Command{
-		Use:   "{{alias}}",
+		Use:   "{{alias}} <spec> ",
```
brendoncarroll, line 1100: maybe include "path" or "file" to indicate the spec is not literally passed as an argument, and that the command will read the filesystem.  There is a way to hide cobra commands with `Hidden`, maybe we should do that for the load testing stuff.

src/pfs/pfs.proto
```diff
@@ -133,6 +134,8 @@ message CommitInfo {
     int64 size_bytes = 1;
   }
   Details details = 12;
+  google.protobuf.Duration compacting_time = 13;
```
brendoncarroll, line 137: I think these should be in details.

src/server/auth/server/testing/auth_test.go
```diff
@@ -2603,10 +2603,13 @@ func TestDebug(t *testing.T) {
 		require.NoError(t, err)
 		expectedFiles[pattern] = g
 	}
-	pattern := path.Join("input-repos", dataRepo, "commits")
-	g, err := globlib.Compile(pattern, '/')
-	require.NoError(t, err)
-	expectedFiles[pattern] = g
+	// Record glob patterns for expected source repo files.
+	for _, file := range []string{"commits", "commits-chart**"} {
```
brendoncarroll, line 2607: Why was this added?

brendoncarroll:
I think we should include the points in a json/csv file with the plots.  No tools or scripts will realistically be able to consume the charts.  They are just for humans.

Some suggestions about user facing stuff and the CommitInfo API change.

etc/testing/circle/run_hub_tests.sh
brycemcanally: The commands immediately before this in this script set the VERSION and build / push the images.

src/internal/pfsload/validator.go
brycemcanally: Yeah, the file content is streamed when this is set.

src/server/pps/cmds/cmds.go
brycemcanally: Sure, I'll change it to `spec-file`. We don't want to hide the load testing. I could see users wanting to use it for a sanity check on their deployment, or the customer team for testing purposes.

src/server/auth/server/testing/auth_test.go
brycemcanally: We now collect commit charts for source repos, so we are checking for those existing in the debug dump output. The reason for the `**` at the end is because the expectation is to produce an error file when collecting the chart in this workload because there is only one commit.

brycemcanally:
The commit infos and job infos are the raw data from the which the chart is created, so we don't need additional data in the format understandable by the particular library I am using for charting. My thinking on this is that this isn't the long term solution, but it will provide a lot of value in the short term. We may end up moving the code for generating the charts elsewhere (maybe the debug dump service that was worked on during the hackathon) or using a more robust visualization library / program / service.

brendoncarroll (approved):

PR #6803, opened 2021-08-31
Vendoring loki client into pach2

chainlink:
Tested on hub and this successfully grabs the logs from loki through pachctl


go.mod
```diff
@@ -99,12 +98,18 @@ require (
 	modernc.org/mathutil v1.0.0
 )
 
+require (
```
acohen4, line 101: why the additional require block?

src/internal/lokiutil/client/client.go
```diff
@@ -0,0 +1,85 @@
...
+	var r QueryResponse
+
+	if err = c.doRequest(ctx, path, query, quiet, &r); err != nil {
+		return nil, err
+	}
+
+	return &r, nil
+}
+
+func (c *Client) doRequest(ctx context.Context, path, query string, quiet bool, out interface{}) error {
```
acohen4, line 57: looks like `quiet` doesn't get used so we can remove it from this and all the above signatures

src/internal/lokiutil/lokiutil.go
```diff
@@ -70,16 +65,10 @@ func QueryRange(ctx context.Context, c *loki.Client, queryStr string, from, thro
 			return nil
 		}
 		from = from.Add(time.Nanosecond)
-		// check if the context has been cancelled
```
acohen4, line 73: why remove this?

go.mod
chainlink: I was curious about that as well, I ran a `go mod tidy` (with go v17) and that's what it came up with.

chainlink: https://stackoverflow.com/questions/69432877/why-there-are-two-require-blocks-in-go-mod-since-go-1-17

src/internal/lokiutil/lokiutil.go
chainlink: We're can now pass context into `QueryRange` so i don't think it's needed anymore?

src/internal/lokiutil/client/client.go
chainlink: I was figuring we'd keep the code as close to the original lib as possible, but I'm also fine to remove.

acohen4: your call 👍

src/internal/lokiutil/lokiutil.go
acohen4: I see, so if ctx is cancelled QueryRange would return an error

acohen4 (approved):
LGTM

PR #6806, opened 2021-08-31
[Doc] [ci-skip] Added `echo <activation-code>|` in front of pachctl enterprise activate…

nadegepepin:
… - original pachctl was hanging b/c we can't copy/paste in the terminal


albscui (approved):
LGTM!

PR #6812, opened 2021-09-02
Move pipeline spec parsing to gopkg.in/yaml.v3 

msteffen:
The 2.0.x version of this PR is in https://github.com/pachyderm/pachyderm/pull/6813

This is necessary for jsonnet-based pipeline templates in Pachyderm. Jsonnet can output multiple JSON documents, but when it does so it separates them with a YAML document separator. Our current pipeline spec parser can't handle this case.

This change will have a very minor effect on the pipeline specs that we can parse. Previously, users could have multiple pipeline specs in the same file, like so:
```
pachctl create pipeline -f - << EOF
{ 
  // spec 1
}
{
  // spec 2
}
```
This isn't valid JSON (which doesn't allow multiple documents in the same file), but it is accepted by Go's 'encoding/json' parser. Now, however, we'll require that such configs include a YAML document separator between the specs (which *is* valid YAML):
```
pachctl create pipeline -f - << EOF
{ 
  // spec 1
}
--- # New
{
  // spec 2
}
```

Also, configs with leading whitespace will no longer be accepted (no one should be doing this anyway):
```
   {
     // spec
   }
```
will need to be
```
{
  // spec
}
```


brendoncarroll (approved):
LGTM.  I don't think we should advertise that you can write pipeline specs in YAML though.  Just that multiple JSON documents are now supported with a suspiciously YAML separator.

src/internal/ppsutil/decoder.go
```diff
@@ -52,16 +51,6 @@ func NewPipelineManifestReader(path string) (result *PipelineManifestReader, ret
 			return nil, err
 		}
 	}
-	// TODO(msteffen): if we can get the yaml decoder to handle leading tabs, as
```
msteffen, line 55: The point of this code block is to determine whether we need to use `encoding/json` or `gopkg.in/pachyderm/yaml` to parse pipeline specs. Since the overall point of this PR is get rid of `encoding/json` at the cost of slight backwards incompatibility, this code block is no longer necessary.

The comment references the fact that `gopkg.in/yaml` can't parse JSON-style documents with leading tabs. This is a bug in the yaml parser, but it's one that I don't think will affect our users. (At the time, I didn't want to break existing pipeline specs with leading tabs, thus the comment, but since we're about to GA 2.0 anyway, and since realistically I think this hedging was more motivated by anxiety than user data, I don't think the comment is really relevant anymore and probably never was)

src/internal/serde/interface.go
```diff
@@ -138,40 +138,3 @@ type Encoder interface {
...
 	EncodeProtoTransform(proto.Message, func(map[string]interface{}) error) error
 }
-
-// Decoder is the interface for decoding a particular serialization format.
-// Currently, the only implementation of Decoder is in serde/yaml/decoder.go
-// (the point of having a common interface is to make it possible to switch
-// between decoding different serialization formats at runtime, but we have no
-// need to switch between a YAML decoder and a JSON decoder, as our YAML decoder
-// can decode JSON).
-type Decoder interface {
```
msteffen, line 148: This interface unified JSON and YAML decoding for PipelineManifestReader, but since we're switching to YAML parsing only, there's only one implementation of the interface and it's no longer really necessary.

src/internal/serde/yaml_decoder.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/gogo/protobuf/jsonpb"
 	"github.com/gogo/protobuf/proto"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
-	"gopkg.in/pachyderm/yaml.v3"
```
msteffen, line 14: We had forked `gopkg.in/yaml.v3` and modified it to parse adjacent json documents:
```
{
  ...
}{
  ...
}
```
in keeping with `encoding/json`. Again, though, since we're about to GA 2.0 it probably makes sense to move off our fork and just ask users to tolerate this change by adding three dashes between their pipelines specs:
```
{ ... }
---
{ ... }
```

src/server/identity/cmds/cmds_test.go
```diff
@@ -60,7 +60,7 @@ func TestClientCRUD(t *testing.T) {
 		  | match '  - https://localhost:1234' \
                   | match '  - https://localhost:5678' \
 		  | match '  - x' \
-		  | match '  - y' \
+		  | match '  - "?y"?' \
```
msteffen, line 63: This is a fix that had been merged upstream in gopkg.in/yaml.v3, that was not in our fork (in yaml 1.1, "y" is a bool so it has to be quoted)

src/server/pfs/server/api_server.go
```diff
@@ -832,7 +832,7 @@ func (a *apiServer) RunLoadTest(ctx context.Context, req *pfs.RunLoadTestRequest
 
 func (a *apiServer) runLoadTest(pachClient *client.APIClient, branch *pfs.Branch, specStr string, seed int64) error {
 	spec := &pfsload.CommitsSpec{}
-	if err := yaml.UnmarshalStrict([]byte(specStr), spec); err != nil {
+	if err := serde.DecodeYAML([]byte(specStr), spec); err != nil {
```
msteffen, line 835: Our serde library uses gopkg.in/yaml.v3 under the hood, so this change gets all of our yaml parsing using the same library.

src/server/pps/cmds/cmds_test.go
```diff
@@ -232,59 +230,6 @@ func TestJSONStringifiedNumbers(t *testing.T) {
 	require.NoError(t, tu.BashCmd(`pachctl list pipeline`).Run())
 }
 
-// TestJSONMultiplePipelinesError tests that when creating multiple pipelines
-// (which only the encoding/json parser can parse) you get an error indicating
-// the problem in the JSON, rather than an error complaining about multiple
-// documents.
-func TestJSONMultiplePipelinesError(t *testing.T) {
```
msteffen, line 239: This behavior is changing. If you send these pipeline specs to Pachyderm, you should now get an error complaining about multiple documents.

Tryneus (approved):
LGTM

PR #6814, opened 2021-09-03
Support using an external database in helm chart

hobbsh:
This PR adds support for using an external database. 

- Adds value structure for `externalDatabase` to configure the connection parameters
- Adds a flag to disable `pgbouncer` because it should probably be disabled in the case of using an external database

Templated it locally using (did not spin it up in a local cluster):
```
helm template . --set deployTarget=AMAZON --set externalDatabase.enabled=true --set externalDatabase.host="some.host" --set externalDatabase.port=5432 --set externalDatabase.database="database" --set externalDatabase.user="user" --set externalDatabase.existingSecretName="existing" --set externalDatabase.existingSecretKey="existing-key" --set pgbouncer.enabled=false --set postgresql.enabled=false
```

### DIFF
--- a/etc/helm/pachyderm/Chart.yaml
+++ b/etc/helm/pachyderm/Chart.yaml
@@ -21,14 +21,14 @@ type: application
 # each time you make changes to the chart and its templates, including
 # the app version.  Versions are expected to follow Semantic
 # Versioning (https://semver.org/)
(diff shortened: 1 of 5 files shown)


chainlink:
This is supported using the global flag

PR #6818, opened 2021-09-07
Light config/cmd refactor

msteffen:
This PR:
- moves the src/internal/config package over to serde (instead of encoding/json) for consistency with other YAML serialization/deserialization=
- makes a few readability-motivated changes to src/internal/config/config.go

### DIFF
--- a/src/internal/config/config.go
+++ b/src/internal/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"bytes"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


src/internal/config/config.go
```diff
@@ -10,6 +9,7 @@ import (
 	"github.com/gogo/protobuf/proto"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/grpcutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/serde"
```
acohen4, line 12: why the move to serde?

acohen4 (approved):
Overall changes make sense and look good! My only question is - why move to serde?

Also I'm not immediately sure why CI is failing

PR #6819, opened 2021-09-08
Refactor pps/cmds and NewPipelineManifestReader

msteffen:
This change factors the 'read pipeline bytes' stage out of `PipelineManifestReader`. This significantly simplifies the serde library, and it lays the groundwork for pipeline templates (as a pipeline spec's raw bytes can either be provided by the `readPipelineBytes` function introduced in this PR, or, soon, by executing a Jsonnet template and passing its output).

Changes:
- Rename `serde.DecodeYAML` to `serde.Decode`. Allow `serde.Decode` to deserialize into both protobuf-generated and non-protobuf structs (with an internal type switch).
- Remove `serde.Decoder.DecodeTransform{,Proto}`. It's currently only used to support TFJobs in pipeline specs, which were never supported anyway, and which have now been removed from Kubeflow. Remove all code for parsing TFJobs from CreatePipeline, and remove all tests for TFJobs.
- Remove `serde.Decoder`, as the above changes render it redundant.
- Change `NewPipelineManifestReader` to accept raw pipeline specs (`[]byte`) rather than paths to files. Change all callers to read the pipeline spec from disk before initializing a `PipelineManifestReader`, instead of passing in the path. This will allow us to generate pipeline specs using templates, and then pass the generated specs to `NewPipelineManifestReader` in a followup PR.


Tryneus (approved):
LGTM - I like breaking up this logic into smaller parts!

PR #6820, opened 2021-09-08
Make OpenTrace setup easier

acohen4:
Each of the first 4 commits handle a different set of suggested changes. They are:

1. Setup Tracing in Test Client
  - setting PACH_TRACE and JAEGER_ENDPOINT should enable Extended Traces across all requests to pachd within a test
2. Adding Jaeger to helm
  - I pulled the k8s resources from the previously recommended URL into helm.
3. Instrument Workers with OpenTrace
  - I found it useful to have the workers create spans against their Pipeline's Trace. I added a utility to their driver
4. Adding a `--trace` argument to pachctl create/update pipeline
  - This way the user only needs to have JAEGER_ENDPOINT var set, and can ergonomically set requests to have traces (includes extended traces)

### DIFF
--- a/etc/helm/examples/local-dev-values.yaml
+++ b/etc/helm/examples/local-dev-values.yaml
@@ -38,3 +38,9 @@ etcd:
 postgresql:
   service:
     type: NodePort
+
+jaeger:
+  enabled: true
+  query:
+    service:
+      type: ClusterIP
--- a/etc/helm/pachyderm/templates/pachd/deployment.yaml
+++ b/etc/helm/pachyderm/templates/pachd/deployment.yaml
@@ -67,6 +67,10 @@ spec:
           value: "/pach"
         - name: ETCD_PREFIX
           #value:
+        {{- if .Values.jaeger.enabled }}
+        - name: JAEGER_ENDPOINT
+          value: jaeger-collector:14268
+        {{- end }}
         - name: STORAGE_BACKEND
           value: {{ include "pachyderm.storageBackend" . | quote }}
           {{- if eq (include "pachyderm.storageBackend" . ) "LOCAL" }}
--- a/src/internal/cmdutil/util.go
+++ b/src/internal/cmdutil/util.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/pachyderm/pachyderm/v2/src/internal/serde"
@@ -51,6 +52,19 @@ func PagerFlags(noPager *bool) *pflag.FlagSet {
 	pagerFlags := pflag.NewFlagSet("", pflag.ContinueOnError)
 	pagerFlags.BoolVar(noPager, "no-pager", false, "Don't pipe output into a pager (i.e. less).")
 	return pagerFlags
+}
+
+func TraceFlags(opentrace *bool) *pflag.FlagSet {
+	traceFlags := pflag.NewFlagSet("", pflag.ContinueOnError)
+	traceFlags.BoolVarP(opentrace, "trace", "t", false, "To tag the request as traced to the backend")
+	return traceFlags
+}
+
+func SetupTrace(opentrace bool) {
+	os.Setenv("PACH_TRACE", strconv.FormatBool(opentrace))
+	if _, ok := os.LookupEnv("PACH_TRACE_DURATION"); !ok {
+		os.Setenv("PACH_TRACE_DURATION", "5m")
+	}
 }
 
 func readLine(r io.Reader) (string, error) {
--- a/src/server/cmd/worker/main.go
+++ b/src/server/cmd/worker/main.go
@@ -37,6 +37,8 @@ func main() {
 func do(config interface{}) error {
 	// must run InstallJaegerTracer before InitWithKube/pach client initialization
 	tracing.InstallJaegerTracerFromEnv()
+	defer tracing.CloseAndReportTraces()
+
 	env := serviceenv.InitServiceEnv(serviceenv.NewConfiguration(config))
 
 	// Enable cloud profilers if the configuration allows.
--- a/src/server/pps/server/worker_rc.go
+++ b/src/server/pps/server/worker_rc.go
@@ -126,6 +126,9 @@ func (a *apiServer) workerPodSpec(options *workerOptions, pipelineInfo *pps.Pipe
 	}, {
 		Name:  client.PPSPipelineNameEnv,
 		Value: pipelineInfo.Pipeline.Name,
+	}, {
+		Name:  "JAEGER_ENDPOINT",
+		Value: os.Getenv("JAEGER_ENDPOINT"),
 	},
 		// These are set explicitly below to prevent kubernetes from setting them to the service host and port.
 		{
--- a/src/server/worker/pipeline/transform/common_test.go
+++ b/src/server/worker/pipeline/transform/common_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/jmoiron/sqlx"
+	"github.com/opentracing/opentracing-go"
 
 	"github.com/pachyderm/pachyderm/v2/src/client"
 	col "github.com/pachyderm/pachyderm/v2/src/internal/collection"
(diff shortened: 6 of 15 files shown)


etc/helm/examples/local-dev-values.yaml
```diff
@@ -38,3 +38,9 @@ etcd:
 postgresql:
   service:
     type: NodePort
+
+jaeger:
+  enabled: true
```
acohen4, line 43: I'll update this to false

src/server/worker/pipeline/transform/common_test.go
```diff
@@ -118,6 +119,10 @@ func (td *testDriver) NewSQLTx(cb func(*sqlx.Tx) error) error {
 	return td.inner.NewSQLTx(cb)
 }
 
+func (td *testDriver) AddSpanToAnyPipelineTrace(operation string, kvs ...interface{}) (opentracing.Span, context.Context) {
+	return td.inner.AddSpanToAnyPipelineTrace(operation, kvs...)
```
acohen4, line 123: I would like to add a tag to identify the worker but am not aware of any worker IDs we store

etc/helm/examples/local-dev-values.yaml
msteffen: This is perhaps a product of my unfamiliarity with Helm, but how would one deploy only Jaeger into an existing cluster? The way at least I've used this in the past is:
- User is having some kind of performance problem
- We walk them through deploying Jaeger into their cluster
- We walk them through collecting a trace

I'm basically hoping to check that this workflow still works/how it works?

etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -67,6 +67,10 @@ spec:
           value: "/pach"
         - name: ETCD_PREFIX
           #value:
+        {{- if .Values.jaeger.enabled }}
+        - name: JAEGER_ENDPOINT
+          value: jaeger-collector:14268
```
msteffen, line 72: IIRC, pachd shouldn't need this environment variable (i.e. this is a pachctl/client-side thing). I believe pachd uses `JAEGER_COLLECTOR_SERVICE_HOST` which is set by Kubernetes after the Jaeger service is created?

src/server/pps/server/worker_rc.go
```diff
@@ -126,6 +126,9 @@ func (a *apiServer) workerPodSpec(options *workerOptions, pipelineInfo *pps.Pipe
 	}, {
 		Name:  client.PPSPipelineNameEnv,
 		Value: pipelineInfo.Pipeline.Name,
+	}, {
+		Name:  "JAEGER_ENDPOINT",
```
msteffen, line 130: Could the workers rely on `JAEGER_COLLECTOR_SERVICE_HOST`? I think that would eliminate the need to set any env vars explicitly, and might make deploying Jaeger into an existing cluster easier (just deploy Jaeger and restart the worker pods, and they'll start talking to Jaeger automatically)

src/server/worker/pipeline/transform/common_test.go
msteffen: I believe we've used worker IP in the past. I'll have to look up how to get it but there's a recipe and we (used to?) use it.

src/server/cmd/worker/main.go
```diff
@@ -37,6 +37,8 @@ func main() {
 func do(config interface{}) error {
 	// must run InstallJaegerTracer before InitWithKube/pach client initialization
 	tracing.InstallJaegerTracerFromEnv()
+	defer tracing.CloseAndReportTraces()
```
msteffen, line 40: I thought `do` didn't exit—does it? Or is this important to error recovery?

etc/helm/examples/local-dev-values.yaml
acohen4: I believe the right way to handle this is to use `helm upgrade` and pass the updated `override-values.yaml`. I'll test to see that this works

etc/helm/pachyderm/templates/pachd/deployment.yaml
acohen4: ah yes that makes sense

src/server/cmd/worker/main.go
acohen4: I'm not sure I understand the question. What do you mean by doesn't exist?

msteffen: Doesn't exit. As in, my understanding is that this defer statement will never run

src/server/worker/pipeline/transform/common_test.go
acohen4: figured this one out: `env.Config().PPSWorkerIP`
https://github.com/pachyderm/pachyderm/blob/master/src/server/cmd/worker/main.go#L70

chainlink:
I think we might want to make jaeger a separate install, or make it a sub chart

src/server/cmd/worker/main.go
msteffen: (to clarify, I think `server.Wait()` on l. 100 doesn't return, typically, which means `return server.Wait()` just blocks. Is that not right?)

src/server/worker/worker.go
```diff
@@ -74,6 +75,9 @@ func NewWorker(
 
 	worker.APIServer = server.NewAPIServer(driver, worker.status, env.Config().PodName)
 
+	span, _ := driver.AddSpanToAnyPipelineTrace("/worker.Master/Startup")
+	tracing.FinishAnySpan(span)
```
msteffen, line 79: My guess is that the goal of this is to try and deduce whether a slow job is blocking on worker startup (i.e. datums aren't being claimed due to scheduling delays)—is that right? Having a length-0 span mark an event strikes me as a somewhat unusual (but very creative!) use of tracing, and I wonder if it would be possible to have a span measure "worker startup time" (i.e. time from startup to first datum) which I think would give at least as much information and seems more conventional.

WDYT? We could do this in a followup PR too.

msteffen:
> I think we might want to make jaeger a separate install, or make it a sub chart

TBH I am also feeling very anxious about this change to the deployment (though, I admit, I haven't tried it!). My main two worries, though, are:
- 99.99% of pachyderm clusters will never include Jaeger (and never should, as tracing does introduce overhead). Will someone see this and think they're supposed to run this?
    - Do you think clusters should be running Jaeger more often than they are? Collecting a trace is kind of a PITA right now, and since users can't generally interpret them they have to extract the trace as JSON and send that to us, and then we have to run Jaeger and load it which is an even bigger PITA.
- The 0.01% of the time that we do want to run Jaeger, we're usually adding it to an existing cluster that's experiencing performance problems. Will having this in the helm chart make it scary to add Jaeger to a production cluster, where we really don't want to change anything else that's there?

msteffen (commented):
A few more thoughts. Thanks for being patient with me on this PR; there are just some pieces of this that make me anxious so I wanted to resolve those. At this point I've read all the changes though so these are my last concerns.

src/internal/cmdutil/util.go
```diff
@@ -53,6 +54,19 @@ func PagerFlags(noPager *bool) *pflag.FlagSet {
...
 }
 
+func TraceFlags(opentrace *bool) *pflag.FlagSet {
+	traceFlags := pflag.NewFlagSet("", pflag.ContinueOnError)
+	traceFlags.BoolVarP(opentrace, "trace", "t", false, "To tag the request as traced to the backend")
+	return traceFlags
+}
+
+func SetupTrace(opentrace bool) {
+	os.Setenv("PACH_TRACE", strconv.FormatBool(opentrace))
```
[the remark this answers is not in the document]
acohen4: That's a very good point. Given the introduced weirdness you speak of, and not a vocal demand for this besides me finding this to be a slightly cleaner interface, I think we can scrap this part of the change.

acohen4:
> TBH I am also feeling very anxious about this change to the deployment (though, I admit, I haven't tried it!). My main two worries are:
> 
> * 99.99% of pachyderm clusters will never include Jaeger (and never should, as tracing does introduce overhead). Will someone see this and think they're supposed to run this?
>   
>   * Do you think clusters should be running Jaeger more often than they are? Collecting a trace is kind of a PITA right now, and since users can't generally interpret them they have to extract the trace as JSON and send that to us, and then we have to run Jaeger and load it which is an even bigger PITA.
> * The 0.01% of the time that we do want to run Jaeger, we're usually adding it to an existing cluster that's experiencing performance problems. Will having this in the helm chart make it scary to add Jaeger to a production cluster, where we really don't want to change anything else that's there?

I found jaeger super helpful in debugging PPS locally, and all of the commits in this PR are meant as ideas for making it slightly less of a PITA. That being said, I totally understand the argument for breaking this into a separate helm chart as we don't manage Jaeger. I think this is something that we can cut out of this PR. 

@chainlink I'm conflicted to how much benefit we then get from deploying Jaeger via a separate helm install or the old way (big Kubernetes manifest).

etc/helm/examples/local-dev-values.yaml
chainlink: In that case, I'd prefer to do it as a separate install

PR #6821, opened 2021-09-08
Remove `PACH_IN_WORKER` environment variable.

brendoncarroll:


src/client/client.go
```diff
@@ -647,28 +648,28 @@ func NewInCluster(options ...Option) (*APIClient, error) {
...
-			return nil, errors.Wrap(err, "could not create client")
-		}
-		if context.SessionToken != "" {
-			client.authenticationToken = context.SessionToken
-		}
-		return client, nil
-	}
-	return nil, errors.New("PEER_PORT not set")
-}
+// func NewInWorker(options ...Option) (*APIClient, error) {
```
brendoncarroll, line 651: Need to actually remove this.

src/server/pps/server/worker_rc.go
```diff
@@ -635,6 +628,12 @@ func (a *apiServer) getWorkerOptions(pipelineInfo *pps.PipelineInfo) (*workerOpt
 	}, nil
 }
 
+func makeWorkerPachctlSecretName(pipelineInfo *pps.PipelineInfo) string {
+	x := "pachctl-secret-" + pipelineInfo.Pipeline.Name
```
PFedak, line 632: still need to handle underscores, should resolve some of the test failures

also maybe make the "localhost:1653" below based on the env

PFedak (commented):
LGTM after the changes I mentioned, hopefully the name stuff is the only source of test failures

src/server/pps/server/worker_rc.go
brendoncarroll: I'm not sure what you mean. Do you mean use the env ValueFrom to get it from kubernetes?  Or should we set it explicitly when we create the worker, and then look there?

The localhost is something we can depend on.  Pods share a networking namespace.  Whatever creates this secret will be doing it with knowledge of how the pod is set up.  Maybe moving it to a constant would be a good idea.  I think trying to get it from k8s is going to be more brittle that this, but I'm not sure exactly what you had in mind.

PFedak: oh sorry, I just meant the port. `PeerPort` is configurable

PFedak (approved):
whoops, sorry, meant to approve before

PR #6825, opened 2021-09-09
Change to Bitnami rootless pg-bouncer image

tybritten:
Switches the pg-bouncer image over to Bitnami. Is dependant on [this PR](https://github.com/bitnami/bitnami-docker-pgbouncer/pull/15) to work.

### DIFF
--- a/etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
+++ b/etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
@@ -46,8 +46,6 @@ spec:
           value: "300"
         - name: PGBOUNCER_IGNORE_STARTUP_PARAMETERS
           value: "extra_float_digits"
-        - name: PGBOUNCER_UNIX_SOCKET_DIR
-          value: " "
         - name: PGBOUNCER_EXTRA_ARGS
           value: "--with-udns"
         PGBOUNCER_EXTRA_ARGS
--- a/etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
+++ b/etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
@@ -30,10 +30,14 @@ spec:
       - env:
         - name: POSTGRESQL_USERNAME
           value: {{ .Values.global.postgresql.postgresqlUsername }}
+        - name: PGBOUNCER_SET_USER
+          value: "True"
         - name: POSTGRESQL_HOST
           value: {{ .Values.global.postgresql.postgresqlHost }}      
         - name: PGBOUNCER_PORT
           value: "5432"
+        - name: PGBOUNCER_DATABASE
+          value: "*"
         - name: PGBOUNCER_MAX_CLIENT_CONN
           value: "1000"
         - name: PGBOUNCER_POOL_MODE
@@ -47,7 +51,7 @@ spec:
             secretKeyRef:
               name: postgres # Must match secret setup by postgres subchart, or postgres-secret.yaml 
               key: postgresql-password
-        image: bitnami/pgbouncer:1.16.0
+        image: vmtyler/pgbouncer:v1.1
         imagePullPolicy: IfNotPresent
         name: pg-bouncer
         ports:
--- a/etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
+++ b/etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
@@ -28,24 +28,26 @@ spec:
 {{- include "pachyderm.imagePullSecrets" . | indent 6 }}
       containers:
       - env:
-        - name: DB_USER
+        - name: POSTGRESQL_USERNAME
           value: {{ .Values.global.postgresql.postgresqlUsername }}
-        - name: DB_HOST
-          value: {{ .Values.global.postgresql.postgresqlHost }}
-        - name: AUTH_TYPE
-          value: md5
-        - name: MAX_CLIENT_CONN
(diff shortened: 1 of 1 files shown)


brendoncarroll:
It looks like there are problems connecting to the database.
```panic: failed to connect to `host=pg-bouncer user=pachyderm database=pachyderm`: server error (FATAL: no such database: pachyderm (SQLSTATE 08P01))```  Which I think is coming from the new pgbouncer, saying it doesn't know where to find pachyderm.

File: etc/helm/pachyderm/templates/pgbouncer/deployment.yaml [hunk 1]
```diff
@@ -28,24 +28,26 @@ spec:
...
           value: "1000"
-        - name: POOL_MODE
+        - name: PGBOUNCER_POOL_MODE
           value: "transaction"
-        - name: IDLE_TRANSACTION_TIMEOUT
+        - name: PGBOUNCER_IDLE_TRANSACTION_TIMEOUT
           value: "300"
-        - name: DB_PASSWORD
+        - name: PGBOUNCER_IGNORE_STARTUP_PARAMETERS
+          value: "extra_float_digits"
```
brendoncarroll, line 44: What does this do?

brendoncarroll:
Because we aren't setting the `POSTGRESQL_DATABASE` env variable, it is defaulting to `postgres`, which does not exist.  We need to have pools for more than one DB because we have `pachyderm` and `dex`.  I'm not sure how that works with this container.  In the previous one.  The previous container defaulted to something like "all databases are provided by the same host".

tybritten:
Yes, I have fixes in the bitnami chart PR to address these issues. I've temporarily built it in my own hub account for testing, so we need to wait for bitnami to merge the changes and then update the chart but this works with my image version for testing

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
tybritten: the other image had this set too. apparently its just to ignore a spurious error when pgb connect to pgsql

brendoncarroll:
We may just be able to pass an `*` in that environment variable.  And that should makes it's way into the config file.
https://github.com/bitnami/bitnami-docker-pgbouncer/blob/master/1/debian-10/rootfs/opt/bitnami/scripts/libpgbouncer.sh#L192
But if that doesn't work, then this might be more involved.

tybritten:
i tried that, but that chart still sets a backend db (which is no good), also it doesn't set the username on the backend which we shouldn't need to but it doesn't work unless we do

brendoncarroll:
This is what is generated by the image we currently use.
```
/ $ cat ./etc/pgbouncer/pgbouncer.ini
################## Auto generated ##################
[databases]
* = host=postgres port=5432 user=pachyderm

[pgbouncer]
listen_addr = 0.0.0.0
listen_port = 5432
unix_socket_dir =
user = postgres
auth_file = /etc/pgbouncer/userlist.txt
auth_type = md5
pool_mode = transaction
max_client_conn = 1000
ignore_startup_parameters = extra_float_digits

# Log settings
admin_users = postgres

# Connection sanity checks, timeouts

# TLS settings

# Dangerous timeouts
idle_transaction_timeout = 300
################## end file ##################```

If you already forked their image, you could hard code a line in their script that ignore the database env var, and generated the * = line.

**Comment by tybritten** [comment]
yup, i did that in a way that they will hopefully merge. The one my chart/image is creating looks like this now:
````
[databases]
*=host=postgres port=5432 user=pachyderm

[pgbouncer]
listen_port=5432
listen_addr=0.0.0.0
auth_file=/opt/bitnami/pgbouncer/conf/userlist.txt
auth_type=md5
pidfile=/opt/bitnami/pgbouncer/tmp/pgbouncer.pid
logfile=/opt/bitnami/pgbouncer/logs/pgbouncer.log
admin_users=pachyderm
client_tls_sslmode=disable
pool_mode=transaction
max_client_conn=1000
idle_transaction_timeout=300
ignore_startup_parameters=extra_float_digits```

But it's unable to connect to pgsql in CI, even though it works for me locally and on gke. Troubleshooting now

**Comment by brendoncarroll** [comment]
I working on debugging this with another branch, just posting an update here.   The DNS lookup is failing in CI.
```
2021-09-14 17:13:59.602 UTC [1] WARNING C-0x5574a52147b0: pachyderm/pachyderm@172.17.0.1:48351 pooler error: pgbouncer cannot connect to server
2021-09-14 17:14:00.604 UTC [1] LOG C-0x5574a52147b0: pachyderm/pachyderm@172.17.0.1:51643 login attempt: db=pachyderm user=pachyderm tls=no
2021-09-14 17:14:00.604 UTC [1] LOG C-0x5574a52147b0: pachyderm/pachyderm@172.17.0.1:51643 closing because: pgbouncer cannot connect to server (age=0s)
2021-09-14 17:14:00.604 UTC [1] WARNING C-0x5574a52147b0: pachyderm/pachyderm@172.17.0.1:51643 pooler error: pgbouncer cannot connect to server
2021-09-14 17:14:00.619 UTC [1] WARNING DNS lookup failed: postgres: result=0
```

tybritten:
i used the CI Minikube settings locally and still couldn't repro

tybritten:
Was able to fix by supplying the fdqn if you're using the included pgsql chart

chainlink (approved):
Thanks for tackling this!

brendoncarroll (approved):
LGTM

PR #6828, opened 2021-09-09
Update loki log docs for helm

PFedak:


PFedak:
Not sure the best way to communicate helm values, the `path.with.dots` version or a little YAML snippet like you added in that other doc.

nadegepepin (approved):
LGTM

PR #6829, opened 2021-09-10
Fix shellcheck in 'make lint'

msteffen:
Previously, `etc/testing/lint.sh` was failing to run `shellcheck` in CI because it was passing the output of a `find` command run with `-print0` (which instructs `find` to separate its output with null characters) to an `xargs` command that was not run with `-0` or `--null`. Thus, `xargs` was interpreting `find`'s output as one super-huge filename containing all of our shell files, and then exiting when that super-file didn't exist. It was logging a warning in CI, but not exiting with a non-zero exit code.

This PR passes the `-0` argument to xargs in lint.sh, so that it splits its input at null characters, effectively re-enabling shellcheck in CI. As this change revealed many previously-suppressed shellcheck errors, this PR fixes those errors as well.

### DIFF
--- a/etc/build/doc.sh
+++ b/etc/build/doc.sh
@@ -3,6 +3,7 @@
 set -e
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+# shellcheck disable=SC1090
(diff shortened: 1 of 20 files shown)


jrockway (approved):
I wondered about that.  I think I asked and was told "oh that's not a problem" but I'm glad someone didn't believe that and found the problem :)

PR #6833, opened 2021-09-13
Pramaterize the existing pachyderm-bootstrap-config k8s secret

acohen4:


etc/helm/examples/local-dev-values.yaml
```diff
@@ -16,10 +16,22 @@ pachd:
   clusterDeploymentID: dev
   enterpriseLicenseKey: ""
 
+  oidc:
+    mockIDP: false
+    clientId: "pachd"
+    issuer: "http://pachd:1658/"
```
tybritten, line 22:
```suggestion
    issuer: "http://pachd:30658/"
```

tybritten (changes requested):
LGTM besides those port changes

acohen4:
> LGTM besides those port changes

Ah right. So this is kinda weird but the reason pachd:1658 is used for LOCAL deploys is because in local deploys, we set the pachd k8s service to type "NodePort", which then forces us to use the in-cluster service address port as 1658 instead of 30658. 
We do that here: https://github.com/pachyderm/pachyderm/blob/master/etc/helm/pachyderm/templates/pachd/service.yaml#L40

I'm not sure what the best way to communicate it is for folks doing iterative deployments where they initially want to setup localhost in their OAuth redirect URI. One option could be adding values with pachd:30658 to a non-local helm values.yaml.

What do you think @tybritten ?

tybritten:
if we're doing nodeport and its 30658 (https://github.com/pachyderm/pachyderm/blob/master/etc/helm/pachyderm/templates/pachd/service.yaml#L41) why would't that work for local?

acohen4:
> if we're doing nodeport and its 30658 (https://github.com/pachyderm/pachyderm/blob/master/etc/helm/pachyderm/templates/pachd/service.yaml#L41) why would't that work for local?

because Console -> pachd can't go through Nodeport i.e. `localhost:30658`.  It would instead have to go through `pachd:1658`, and since the Issuer URL needs to be the same value on side of the issuer as on the side of the client, local pachd's identity service must also see itself as `pachd:1658`.

Related PR: https://github.com/pachyderm/pachyderm/pull/6548

tybritten:
Ah, can't we set both nodeport and port to 30658 there?

acohen4:
> Ah, can't we set both nodeport and port to 30658 there?

Good point! That's a viable (and simpler) alternative for that. 

I'm not sure what the exact reason is that we are careful to to have `nodeport != port`. That's something @chainlink could weigh in on when he's back, or perhaps @robert-uhl knows.

etc/helm/examples/local-values.yaml
```diff
@@ -3,12 +3,39 @@
 
 deployTarget: LOCAL
 
+pachd:
+  # registerWithEnterpriseOnly: false
```
acohen4, line 7: I'll delete this line

chainlink:
Can we access the nodeport from inside the cluster? (I don't think so, but am unsure)

tybritten:
no, but you can set both the nodeport and port to the same number and inside the cluster will hit `pachd:30658` and the local os outside the cluster will hit the nodeport at 20658

etc/helm/pachyderm/values.yaml
```diff
@@ -212,7 +212,18 @@ pachd:
   enterpriseLicenseKey: ""
   rootToken: ""
   enterpriseSecret: ""
-  oauthClientSecret: ""
+  oidc:
+    mockIDP: false
+    clientId: ""
```
chainlink, line 217: Set default to `pachd` to replicate prior behavior

acohen4:
@tybritten I'd like to handle the Port == NodePort change outside of this PR.

 I made an Issue to track this: https://github.com/pachyderm/pachyderm/issues/6901
I would also like to do one last round of thoughts on whether this change is actually desirable. My main concern is whether this would be unconventional, and if it is in fact unconventional, is it worth doing for a simpler configuration?

chainlink (changes requested):
Looking good! A couple of comments:
- Removing `pachd.oauthClientSecret` will technically break compatibility, we may want to consider having the new variables follow the old convention (ie. `oauthClientName`) instead of moving to a new stanza.
- idps / connectors should just be passed through as a `toYaml` (with the exception of the special case for mockIDP)

etc/helm/examples/local-dev-values.yaml
```diff
@@ -15,11 +15,21 @@ pachd:
...
+    issuer: ""
+    clientID: ""
+    clientSecret: ""
+    redirectURI: "http://localhost:30658/callback"
 
 console:
   enabled: false
   config:
-    issuerURI: "http://pachd:1658"
+    issuerURI: "http://pachd:1658/"
```
chainlink, line 32: Is a trailing slash always required here?

etc/helm/examples/local-values.yaml
```diff
@@ -3,12 +3,37 @@
...
     oauthClientID: "console"
     oauthClientSecret: "abc"
     graphqlPort: 4000
     oauthPachdClientID: "pachd"
     pachdAddress: "pachd-peer.default.svc.cluster.local:30653"
+
+enterpriseServer:
+  enabled: false
+
+etcd:
```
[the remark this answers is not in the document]
acohen4: will remove this

etc/helm/examples/local-values.yaml
```diff
@@ -3,12 +3,37 @@
 
 deployTarget: LOCAL
 
+pachd:
+  enterpriseLicenseKey: "license"
```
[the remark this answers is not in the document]
acohen4: will remove this

etc/helm/examples/local-dev-values.yaml
acohen4: verified that the backslash isn't essential. The issuer URI value just needs to be consistent between console and pachd

chainlink (approved):

PR #6834, opened 2021-09-13
fixes port reference for pachd console

tybritten:
Fixes references to port 1658 to 30658 in default config pod settings


chainlink:
Can we use the nodeport on cluster?

PR #6838, opened 2021-09-14
Helm Deploy Enterprise Member Clusters

acohen4:
depends on using https://github.com/pachyderm/config-pod/pull/3 to correctly interpret the new config in config-secret.yaml


etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -24,7 +24,7 @@ stringData:
     client_id: {{ required "For Authentication, an OIDC Client ID for this pachd must be set." .Values.pachd.oidc.clientId }}
     client_secret: {{ $oidcSecret }}
     issuer: {{ required "For Authentication, an OIDC Issuer for this pachd must be set." .Values.pachd.oidc.issuer }}
-    localhost_issuer: true
+    localhost_issuer: {{ if .Values.pachd.enterpriseLicenseKey }}true{{ else }}false{{ end }}
```
chainlink, line 27: I think you can use the `toString` function here

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -2,7 +2,7 @@
 SPDX-FileCopyrightText: Pachyderm, Inc. <info@pachyderm.com>
 SPDX-License-Identifier: Apache-2.0
 */ -}}
-{{ if .Values.pachd.enterpriseLicenseKey }}
+{{ if .Values.pachd.activateEnterprise }}
```
chainlink, line 5: I like changing this to be an explicit bool instead of the implicit value check on license key

etc/helm/examples/enterprise-dev.yaml
```diff
@@ -9,4 +9,17 @@ enterpriseServer:
     tag: local
 pachd:
   enabled: false
-
+  activateEnterprise: false
+  enterpriseLicenseKey: ""
```
chainlink, line 13: Won't this fail without a license key?

Makefile
```diff
@@ -193,12 +193,20 @@ launch-enterprise: check-kubectl check-kubectl-connection
...
+	kubectl wait --for=condition=ready pod -l app=pachd --timeout=5m
+	@echo "pachd launch took $$(($$(date +%s) - $(STARTTIME))) seconds"
+
 clean-launch: check-kubectl
 	helm delete pachyderm || true
 	helm delete enterprise || true
 	# These resources were not cleaned up by the old pachctl undeploy
 	kubectl delete roles.rbac.authorization.k8s.io,rolebindings.rbac.authorization.k8s.io -l suite=pachyderm
 	kubectl delete clusterroles.rbac.authorization.k8s.io,clusterrolebindings.rbac.authorization.k8s.io -l suite=pachyderm
+	kubectl delete job -l suite=pachyderm
```
chainlink, line 209: Nice catch!

chainlink:
What's the workflow here?
1. helm install enterprise server (with license key?)
2. helm install member cluster (No license key?)

Also, in general can we derive more values? ie. if they supply a DNS entry for the enterprise server, can we fill in callback URLs, etc using string interpolation?

etc/helm/examples/enterprise-member-values.yaml
```diff
@@ -0,0 +1,33 @@
...
+# SPDX-License-Identifier: Apache-2.0
+deployTarget: LOCAL
+
+pachd:
+  metrics:
+    enabled: false
+  clusterDeploymentID: dev
+  activateEnterprise: true
+  enterpriseLicenseKey: ""
+  enterpriseServerAddress: "grpc://<ENTERPRISE-SERVER-IP>:1650"
```
chainlink, line 11: If possible, the examples values files should be able to be run directly

etc/helm/examples/enterprise-values.yaml
```diff
@@ -1,7 +1,24 @@
...
+  activateEnterprise: true
+  enterpriseLicenseKey: "<ENTEPRRISE-LICENSE-KEY>"
+  oidc:
+    clientId: "enterprise-pach"
+    issuer: "http://<ENTERPRISE-SERVER-IP>:31658/"
+    redirectURI: "http://<ENTERPRISE-SERVER-IP>:31657/authorization-code/callback"
+    setupConnector: true
+    mockIDP: false
+    connector:
+      id: "idp"
```
chainlink, line 20: Is there ever a case where we'd have multiple connectors? (ie multiple social oauth providers?)

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -40,20 +40,26 @@ stringData:
     - repoReader
 {{- end }}
 
+# when registering a cluster with a separate enterprise server
+{{ if not .Values.pachd.enterpriseLicenseKey }}
```
chainlink, line 44: Is there ever a case where you're not specifying the enterprise license key and you wouldn't nee enterpriseServerAddress and enterpriseRootToken?

etc/helm/pachyderm/values.yaml
```diff
@@ -210,6 +210,9 @@ pachd:
     #  expose: true
     #  port: 30650
   enterpriseLicenseKey: ""
+  enterpriseRootToken: ""
```
chainlink, line 213: Need to run `make schema` to regen the the json schema

chainlink (changes requested):

etc/helm/pachyderm/templates/pachd/config-secret.yaml
acohen4: that sounds good to me. I was conflicted on this and made the initial decision, based on reducing parameters. I agree being explicit is something we should strive for in configuration

etc/helm/pachyderm/templates/pachd/config-secret.yaml
acohen4: Good question. The only enterprise activation case where you don't require an enterprise key is when you already have an enterprise license key activated on another cluster, and you want to register with that cluster to inherit it's enterprise capabilities. 

That's the reason `activate == true && licenseKey == ""` is implying registering with another cluster here

etc/helm/pachyderm/values.yaml
acohen4: right! thanks

etc/helm/examples/enterprise-values.yaml
acohen4: excellent question. based on the data structure it appears possible but I'm not sure what that would look like. Let me try that locally

etc/helm/examples/enterprise-dev.yaml
acohen4: as noted in another PR comment, `activateEnterprise == true && enterpriseLicenseKey == "" is implying registering with another cluster here`
https://github.com/pachyderm/pachyderm/pull/6838/files#r711054238

acohen4:
re-creating PR with modifications to master.  #6918

PR #6840, opened 2021-09-14
First draft of Enterprise Deployment with Helm Doc

acohen4:


doc/docs/master/enterprise/auth/enterprise-server/setup.md
```diff
@@ -20,8 +20,37 @@ The setup of an Enterprise Server requires to:
...
+To deploying a Pachyderm cluster with the embedded enterprise server, an additional set of helm parameter values should be populated. 
+
+#### Create your values.yaml 
+Update your values.yaml with your enterprise license key, and auth configurations ([see example of values.yaml here](https://github.com/pachyderm/pachyderm/blob/master/etc/helm/examples/local-values.yaml)) or use our minimal example below.
+
+=== "values.yaml with activation of an enterprise license and authentication"
+
+	```yaml
+	pachd:
+	  	activateEnterprise: true
```
chainlink, line 32: Couple of things here:
- I think it would be good to add an explanation for all/most of these values to clarify what they do. 
- I think perhaps we should have a pachdHost which is the `hostname:port` or `ip:port` and that gets templated through everywhere `<PACH-IP>` is mentioned (instead of specifying multiple values)

chainlink (changes requested):

nadegepepin (changes requested):
- It looks like some of the values YAML in this page are not referenced in https://docs.pachyderm.com/2.0.x-rc/reference/helm_values/
- If the registration of a cluster can be done while deploying the cluster (by setting specific values in the YAML), then we probably want to quickly mention it in step 2 (activate enterprise licensing and enable auth). Otherwise, the only solution left at the point we reach step 3 (register your cluster) is using pachctl. Do I make sense?

acohen4:
Effectively a duplicate of https://github.com/pachyderm/pachyderm/pull/6920

PR #6842, opened 2021-09-14
Configure a user accessible Issuer address for deployments when the OIDC issuer address isn't publicly accessible

acohen4:
Will also make a PR to 2.0.x

When this goes in the API client will need to be updated by config-pod project

For console, we currently assume the publicly addressable url is localhost in a local deployment, but this may not be the case. This allows the publicly addressable url (ie. the login URL the user uses) to be configured separately.


acohen4:
@jrockway this change adds a field to the auth.OIDCConfig proto. What's the process for making sure the version is updated in hub?

jrockway:
@acohen4 Hub pipes the JSON representation of the protos into `pachctl`.  If the operation will succeed without changing the JSON we produce, then Hub doesn't need to change.  (My view on this PR is that we don't need to do anything.  Hub's issuer is accessible outside the cluster, so it doesn't seem like we need to supply this field, except maybe for local development... and we'll add it there if it's needed.)

https://github.com/pachyderm/hub/blob/master/pachctl/pachctl.go#L73

https://github.com/pachyderm/hub/blob/master/pachctl/pachctl.go#L155

chainlink:
How does this weave together in (or what wasn't covered in ) https://github.com/pachyderm/pachyderm/pull/6548 ?

msteffen (approved):
LGTM

acohen4:
> How does this weave together in (or what wasn't covered in ) #6548

Good question. 

This is actually fixing a concept introduced in that change. That change made the assumption that when LocalhostIssuer was set, we always wanted to have users reach the OAuth Issuer (dex), at `localhost:30658`.

Localhost-Issuer actually indicates whether Dex is located on localhost, and was introduced to avoid hairpin mode (mostly for Minikube). It can still be set in non-local clusters. 

This change lets you separately configure the Dex Auth URL in case it's Issuer address is not accessible outside the cluster.

PR #6880, opened 2021-09-16
Chunk prefetcher & file set caching

brycemcanally:
This PR will be used for iterating on the chunk prefetching functionality.


.circleci/config.yml
```diff
@@ -191,12 +191,12 @@ workflows:
     jobs:
       - hub
   hub_load_tests:
-    triggers:
-      - schedule:
-          cron: "0 6 * * *"
-          filters:
-            branches:
-              only: master
+    #triggers:
```
brendoncarroll, line 194: Un-comment before merging.

etc/testing/loads/load-1.yaml
```diff
@@ -10,7 +10,9 @@ operations:
               - source: "random"
                 prob: 100
         prob: 100
-validator: {}
+validator:
+  frequency:
+    prob: 0
```
brendoncarroll, line 15: So the validator is enabled, but it has zero probability of checking anything?

etc/testing/loads/load-9.yaml
```diff
@@ -14,7 +14,9 @@ operations:
           count: 10 
           directoryProb: 50 
         prob: 20 
-validator: {}
+validator:
+  frequency:
+    prob: 0
```
brendoncarroll, line 19: Same as all the above.

src/client/pfs_file.go
```diff
@@ -380,6 +380,23 @@ func (ctfsc *CreateFileSetClient) Close() (*pfs.CreateFileSetResponse, error) {
 	return ret, nil
 }
 
+// GetFileSet gets a file set for a commit.
+func (c APIClient) GetFileSet(repo, branch, commit string) (_ string, retErr error) {
+	defer func() {
+		retErr = grpcutil.ScrubGRPC(retErr)
+	}()
```
brendoncarroll, line 387: This removes the structured error. I don't know what the benefit is, other than maybe formatting in the CLI.  It shouldn't be in the client, which can be used programmatically.

src/internal/pfssync/cache_client.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// TODO: Expose configuration for cache size?
+// TODO: Dedupe work?
+func NewCacheClient(pachClient *client.APIClient, renewer *renew.StringSet) *CacheClient {
+	cc := &CacheClient{
+		APIClient: pachClient,
+		renewer:   renewer,
+	}
+	cache, err := simplelru.NewLRU(100, cc.onEvicted)
```
brendoncarroll, line 28: 100*20MB = 2GB.  Is there a way to set this? Are we okay with saying the workers need this much RAM?

etc/testing/loads/load-1.yaml
brycemcanally: Yeah, it needs to be enabled for deletions, but we don't want to run the validator.

src/client/pfs_file.go
brycemcanally: Yeah, this is the common pattern we have been following, but the go client needs a full rewrite and consistent error handling.

src/internal/pfssync/cache_client.go
brycemcanally: This is caching the ID, not any chunks.

src/internal/pfssync/cache_client.go
```diff
@@ -0,0 +1,59 @@
...
+		// lru.NewWithEvict only errors for size < 1
+		panic(err)
+	}
+	cc.cache = cache
+	return cc
+}
+
+func (cc *CacheClient) GetFileTAR(commit *pfs.Commit, path string) (io.ReadCloser, error) {
+	key := pfsdb.CommitKey(commit)
+	cc.mu.Lock()
```
brendoncarroll, line 39: These lock/unlocks are probably going to get screwed up at some point.  They look correct at the moment.  Maybe pull them out into get and put methods.  Then you could add type safety around the cache, do the casting there.

src/internal/pfssync/sync_test.go
```diff
@@ -1,5 +1,54 @@
...
+
+	"github.com/pachyderm/pachyderm/v2/src/client"
+	"github.com/pachyderm/pachyderm/v2/src/internal/dockertestenv"
+	"github.com/pachyderm/pachyderm/v2/src/internal/randutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/require"
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/renew"
+	"github.com/pachyderm/pachyderm/v2/src/internal/testpachd"
+)
+
+func BenchmarkDownload(b *testing.B) {
```
brendoncarroll, line 17: Do we ever run these in CI?

src/internal/storage/chunk/reader.go
```diff
@@ -60,9 +66,36 @@ func (r *Reader) Iterate(cb func(*DataReader) error) error {
...
 
 // Get writes the concatenation of the data referenced by the data references.
-func (r *Reader) Get(w io.Writer) error {
+func (r *Reader) Get(w io.Writer) (retErr error) {
+	if len(r.dataRefs) <= 1 {
+		return r.Iterate(func(dr *DataReader) error {
+			return dr.Get(w)
+		})
+	}
+	ctx, cancel := context.WithCancel(r.ctx)
```
brendoncarroll, line 75: You should do a `defer cancel()` here.  I think it leaks below when retErr == nil.  It's safe to call more than once.

src/internal/storage/chunk/transform.go
```diff
@@ -47,33 +50,38 @@ func Create(ctx context.Context, opts CreateOptions, ptext []byte, createFunc fu
 
 // Get calls getFunc to retrieve a chunk, then verifies, decrypts, and decompresses the data.
 // cb is called with the uncompressed plaintext
-func Get(ctx context.Context, client Client, cache kv.GetPut, ref *Ref, cb kv.ValueCallback) error {
-	if err := getFromCache(ctx, cache, ref, cb); err == nil {
-		return nil
-	}
+func Get(ctx context.Context, client Client, cache kv.GetPut, deduper *miscutil.WorkDeduper, ref *Ref, cb kv.ValueCallback) error {
```
brendoncarroll, line 53: I thought we were going to move the cache out of here into the dataReader, so that this code just focused on the transformation, now that the caching is more complicated.

src/internal/pfssync/sync_test.go
brycemcanally: No.

src/internal/storage/chunk/transform.go
```diff
@@ -47,33 +50,38 @@ func Create(ctx context.Context, opts CreateOptions, ptext []byte, createFunc fu
...
-			return err
-		}
-		var r io.Reader = bytes.NewReader(ctext)
-		var err error
-		if r, err = decrypt(ref.Dek, r); err != nil {
+	b := backoff.NewExponentialBackOff()
+	b.InitialInterval = 1 * time.Millisecond
+	return backoff.RetryUntilCancel(ctx, func() error {
+		return getFromCache(ctx, cache, ref, cb)
+	}, b, func(err error, _ time.Duration) error {
```
brendoncarroll, line 61: This is definitely not what the author of backoff intended.  Literally everything is happening in a callback that's supposed to be for logging.

The logic should be (in a loop).
1. check the cache, if it's there then call the callback.
2. fill the cache.  This is done in the deduper.  And will call the transform Get with a callback.  The transform Get loses knowledge of the cache.

After like 10 tries fallback to just getting it manually.  Alternatively you can call cb inside the deduper, but that will block other calls, which could use it if it were just in the cache.

src/internal/storage/chunk/transform.go
brycemcanally: I'll add a TODO for this. I think we should focus on what is needed for now.

src/internal/storage/chunk/reader.go
```diff
@@ -60,9 +66,36 @@ func (r *Reader) Iterate(cb func(*DataReader) error) error {
...
+			cancel()
+		}
+		if err := taskChain.Wait(); retErr == nil {
+			retErr = err
+		}
+	}()
+	sem := semaphore.NewWeighted(int64(r.prefetchLimit))
 	return r.Iterate(func(dr *DataReader) error {
-		return dr.Get(w)
+		if err := sem.Acquire(ctx, 1); err != nil {
```
brendoncarroll, line 87: Why don't we Acquire inside CreateTask?

src/internal/storage/fileset/postgres_store.go
```diff
@@ -69,7 +81,54 @@ func (s *postgresStore) get(ctx context.Context, q sqlx.QueryerContext, id ID) (
...
+func (s *postgresStore) getFromCache(ctx context.Context, id ID) (*Metadata, error) {
+	var mdData []byte
+	if err := s.cache.Get(ctx, id[:], func(data []byte) error {
+		mdData = data
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	md := &Metadata{}
+	if err := proto.Unmarshal(mdData, md); err != nil {
```
brendoncarroll, line 120: I'm pretty sure `proto.Unmarshal` will copy the underlying bytes. json.Unmarshal does.  So you should be able to stick this in the callback, and we don't have to make another copy.

src/internal/storage/fileset/storage.go
```diff
@@ -69,7 +71,8 @@ func NewStorage(mds MetadataStore, tr track.Tracker, chunks *chunk.Storage, opts
 			FixedDelay:  DefaultCompactionFixedDelay,
 			LevelFactor: DefaultCompactionLevelFactor,
 		},
-		filesetSem: semaphore.NewWeighted(math.MaxInt64),
+		filesetSem:    semaphore.NewWeighted(math.MaxInt64),
```
brendoncarroll, line 74: `MaxInt64`? Apparently this isn't used?

src/server/pps/server/api_server.go
```diff
@@ -3008,7 +3008,7 @@ func (a *apiServer) RunLoadTest(ctx context.Context, req *pfs.RunLoadTestRequest
 			fmt.Sprintf("cp -r /pfs/%s/* /pfs/out/", repo),
 		},
 		&pps.ParallelismSpec{
-			Constant: 1,
+			Constant: 5,
```
brendoncarroll, line 3011: Did you mean to leave this changed?

src/internal/storage/chunk/transform.go
brycemcanally: Yeah, so this probably wasn't the intended use of the notify callback, but I think it is preferable to what you suggested because it is simpler and includes an exponential backoff. I don't think there is any reason the notify callback cannot be more general purpose than just logging.

src/internal/storage/chunk/reader.go
brycemcanally: The goroutine would exist at that point, we want to acquire before that.

src/server/worker/datum/option.go
```diff
@@ -60,6 +59,6 @@ func WithTimeout(timeout time.Duration) Option {
 // WithPrefixIndex prefixes the datum directory name (both locally and in PFS) with its index value.
 func WithPrefixIndex() Option {
 	return func(d *Datum) {
-		d.storageRoot = path.Join(d.set.storageRoot, fmt.Sprintf("%016d", d.meta.Index)+"-"+d.ID)
+		d.IDPrefix = fmt.Sprintf("%016d", d.meta.Index) + "-"
```
brendoncarroll, line 62: What's going on here?

src/internal/storage/fileset/storage.go
brycemcanally: I'm not following.

src/server/pps/server/api_server.go
brycemcanally: Yeah, I wanted to bump the load test parallelism a bit.

src/server/worker/datum/option.go
brycemcanally: No, it addresses a performance issue with the way work is laid out (we need to prefix a lexicographically ordered ID with respect to the order output by the datum iterators).

src/internal/storage/chunk/transform.go
brendoncarroll: You could just move this into the regular callback.  It would do the same thing.

src/internal/storage/fileset/storage.go
brendoncarroll: If it has the maximum size then we probably won't ever reach the limit.

brycemcanally: It is the default, we overwrite it with the functional option.

brendoncarroll (approved):

PR #6894, opened 2021-09-17
`dockertestenv`: Add Minio backed `obj.Client`

brendoncarroll:
Adds `NewTestObjectClient` to `dockertestenv` for creating `obj.Client`s backed by minio, cleaned up at the end of the test.

- Adds support for minio URLs in the object package.  These are special cased to only work in tests, so we are not adding support for these urls in production, they will continue to return errors there.
- The implementation uses the docker client to create the minio container programmatically, rather than using a shell script.
- Modifies PFS and PPS tests which use write to object storage to use this implementation
- `obj.NewTestClient` and `testutil.NewObjectClient` have been removed.


src/server/worker/pipeline/transform/transform_test.go
```diff
@@ -297,10 +296,10 @@ func TestTransformPipeline(suite *testing.T) {
 
 	suite.Run("TestJobSuccessEgressEmpty", func(t *testing.T) {
 		t.Parallel()
-		_, bucket := testutil.NewObjectClient(t)
+		objC := dockertestenv.NewTestObjClient(t)
 		pi := defaultPipelineInfo()
 		pi.Details.Input.Pfs.Glob = "/"
-		pi.Details.Egress = &pps.Egress{URL: fmt.Sprintf("local://%s/", bucket)}
+		pi.Details.Egress = &pps.Egress{URL: fmt.Sprintf("local://%s/", objC.BucketURL())}
```
Tryneus, line 302: Wait, is this setting it to `local://minio://<host>/<bucket>`?  That can't be right

src/internal/obj/api.go
```diff
@@ -28,4 +28,7 @@ type Client interface {
 
 	// Exists checks if a given object already exists
 	Exists(ctx context.Context, name string) (bool, error)
+
+	// BucketURL returns the URL of the bucket this client uses.
+	BucketURL() string
```
Tryneus, line 33: Should this maybe return the `ObjectStoreURL` that we already generate?

src/server/worker/pipeline/transform/transform_test.go
brendoncarroll: Yeah I'm not sure why this didn't break.  Interestingly, url.Parse doesn't error with that input.
https://play.golang.org/p/CIVsAPZW2NA

src/internal/dockertestenv/util.go
```diff
@@ -0,0 +1,125 @@
...
+// portMap is a mapping from host ports to container ports.  The image/values of this mapping are taken implicitly to be the exposed container ports.
+func ensureContainer(ctx context.Context, dclient docker.APIClient, containerName string, spec containerSpec) error {
+	imageName := spec.Image
+	portMap := spec.PortMap
+	if _, err := dclient.ContainerInspect(ctx, containerName); err != nil {
+		if !isErrNoSuchContainer(err) {
+			return err
+		}
+	} else {
+		logrus.Infof("container %s exists. skip creation.", containerName)
```
Tryneus, line 58: This may still need to ensure that the container is running - it may exist but be stopped.

src/internal/obj/amazon_client.go
```diff
@@ -284,6 +284,10 @@ func (c *amazonClient) Exists(ctx context.Context, name string) (bool, error) {
 	return true, nil
 }
 
+func (c *amazonClient) BucketURL() ObjectStoreURL {
+	panic("not implemented")
```
Tryneus, line 288: All the other bucket URLs were implemented - is there a reason this was left out?

src/internal/dockertestenv/minio.go
```diff
@@ -0,0 +1,87 @@
...
+		imageName     = "minio/minio"
+	)
+	minioLock.Lock()
+	defer minioLock.Unlock()
+	return ensureContainer(ctx, dclient, containerName, containerSpec{
+		Image: imageName,
+		Cmd:   []string{"server", "/data", `--console-address=:9001`},
+		PortMap: map[uint16]uint16{
+			9000: 9000,
+			9001: 9001,
```
Tryneus, line 84: should these be based off of the `minioPort` variable?

src/internal/obj/factory.go
```diff
@@ -429,6 +429,16 @@ func NewClientFromURLAndSecret(url *ObjectStoreURL, reverse ...bool) (c Client,
 	case "wasb":
 		// In Azure, the first part of the path is the container name.
 		c, err = NewMicrosoftClientFromSecret(url.Bucket)
+	case "minio":
+		// TODO(brendon): This is kind of janky
+		// This is to get the PFS tests working without the filesystem object store
+		// The alternative was writing to environment variables in dockertestenv.
+		if !strings.HasSuffix(os.Args[0], ".test") {
```
Tryneus, line 436: Yeah, this is terrifying but I guess there isn't really a better way supported by golang.

Tryneus (approved):
LGTM, a few nits

src/internal/obj/factory.go
brendoncarroll: The environment variables I'm referring to are to set the secret.  I'm not too familiar with how we do secrets for {Put,Get}FileURL.  But we need a way for PFS to figure out what secrets to use.  Maybe calling the scheme `test://` would be better?

src/internal/obj/amazon_client.go
brendoncarroll: There's definitely not a good reason why it was left out.  Will implement.

PR #6903, opened 2021-09-21
Eliminate no-op scaling updates

msteffen:


src/server/pps/server/pipeline_controller.go
```diff
@@ -512,23 +517,55 @@ func (op *pipelineOp) scaleUpPipeline() (retErr error) {
...
+				return maxParallelism
+			}
+			return nTasks
+		}()
+		if curReplicas == target {
+			return false // no changes necessary
 		}
+
+		// Scaling is being updated; schedule another check in scaleUpInterval
+		go func() {
```
msteffen, line 556: only run this if spec.Autoscaling == true

PR #6910, opened 2021-09-22
Readd get file recursive to pachctl

brycemcanally:
This PR readds the get file recursive functionality to pachctl get file.

### DIFF
--- a/src/server/pfs/cmds/cmds.go
+++ b/src/server/pfs/cmds/cmds.go
@@ -1,6 +1,7 @@
 package cmds
 
 import (
+	"archive/tar"
 	"bufio"
 	"fmt"
 	"io"
@@ -9,7 +10,6 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
-	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


src/server/pfs/cmds/cmds.go
```diff
@@ -1199,6 +1201,30 @@ $ {{alias}} 'foo@master:/test\[\].txt'`,
...
+				}
+				dir, err := ioutil.TempDir("", "pachyderm_get_file_recursive")
+				if err != nil {
+					return err
+				}
+				defer os.RemoveAll(dir)
+				if err := tarutil.Import(dir, r); err != nil {
+					return err
+				}
+				return os.Rename(path.Join(dir, file.Path), outputPath)
```
brendoncarroll, line 1226: Do we really just clobber everything in the entire directory? Overwriting the files we are exporting is expected, but a hard replace of the directory node itself is going to surprise someone and cause data loss.

brycemcanally: Okay, I was thinking in terms of syncing up the PFS state with the local state, but that seems like the incorrect way to think about this functionality (considering how recursive put file works). I'll change this to just create / overwrite the files that are returned.

brendoncarroll: Cool, everything else looks good. I will approve.  FWIW this is definitely more useful for certain use cases; I just don't think it can be a good default.

brendoncarroll (approved):

PR #6912, opened 2021-09-22
Make more storage helm fields required.

PFedak:


PFedak:
Noticed this while deploying on AWS, I noticed we marked the bucket as required in google so I figured these make sense as well.

chainlink (approved):

Issue #6811, opened 2021-09-02
Prometheus Pipeline worker stats missing


tybritten:
**What happened?**:
Using the prometheus endpoint on pipelines exposes pachd_pfs metrics (and go and grpc ones) but the expected pipeline worker ones aren't there (https://github.com/pachyderm/pachyderm/blob/master/src/server/worker/stats/stats.go)
**What you expected to happen?**:
the documented job metrics to be there (https://docs.pachyderm.com/2.0.x-rc/deploy-manage/deploy/prometheus/job_metrics/)
**How to reproduce it (as minimally and precisely as possible)?**:
curl the /metrics endpoint of any pipeline service
**Anything else we need to know?**:
broken in 1.13 and 2.0
**Environment?**:
- Kubernetes version (use `kubectl version`):
- Pachyderm CLI and pachd server version (use `pachctl version`):
- Cloud provider (e.g. aws, azure, gke) or local deployment (e.g. minikube vs dockerized k8s):
- If you deployed with helm, the values you used (`helm get values pachyderm`):
- OS (e.g. from /etc/os-release):
- Others:

[AB#130](https://dev.azure.com/PachydermCompany/92b98a32-88e5-46cb-a28f-5d064712e03b/_workitems/edit/130)

jdoliner:
Does this bug apply to all pipelines or is it specific to pipelines with autoscaling / standby enabled. I remember there being an issue with that because the pipeline gets spun down and thus the endpoint goes away.

tybritten:
All pipelines

avigil:
@Tryneus assigning to you since you made the last changes on the file listed in description

jdoliner:
Looking at release notes for 1.13 this PR looks a little bit suspicious: https://github.com/pachyderm/pachyderm/pull/5912

Tryneus:
Fixed by #6914.

tybritten:
Trying w RC2 and I see all the pachd/pfs stats, but still no worker stats?

JoeyZwicker:
@tybritten do we need to reopen this issue for worker stats too?

PR #6914, opened 2021-09-22
Fix worker metrics in 2.x

### LINKED ISSUE #6809: Updated Prometheus port name in pachd but not pipelines
**What happened?**:
service monitor set to connect to `prom-metrics` port which is what pachd exposes but no pipelines are watched.
all pipelines watched too
deploy a pipeline. they still deploy with the prometheus port named `prometheus-metrics` instead of `prom-metrics`
https://github.com/pachyderm/pachyderm/blob/master/src/internal/deploy/assets/assets.go#L740

**Environment?**:
- If you deployed with helm, the values you used (`helm get values pachyderm`):
- Others:

[AB#128](https://dev.azure.com/PachydermCompany/92b98a32-88e5-46cb-a28f-5d064712e03b/_workitems/edit/128)

Tryneus:
Metrics were broken in a couple ways - the identifier for the service port was incorrect due to an incomplete identifier change, and workers were not actually incrementing any of the metric counters.  Fixed by incrementing counters in `datum.go`.  Also changed the worker metrics to use `promauto` to remove a little boilerplate.

Fixed #6809 and #6811


src/server/worker/datum/datum.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/pps"
 	pfsserver "github.com/pachyderm/pachyderm/v2/src/server/pfs"
 	"github.com/pachyderm/pachyderm/v2/src/server/worker/common"
+	workerStats "github.com/pachyderm/pachyderm/v2/src/server/worker/stats"
```
PFedak, line 29: was this just for clarity, or is there another `stats`?

src/server/worker/datum/datum.go
```diff
@@ -161,6 +162,11 @@ func (s *Set) WithDatum(meta *Meta, cb func(*Datum) error, opts ...Option) error
 				if retErr == nil || i == d.numRetries {
 					retErr = d.finish(retErr)
 				}
+				duration := time.Duration(int64(d.meta.Stats.ProcessTime.GetNanos()) + d.meta.Stats.ProcessTime.GetSeconds()*1000000000)
```
PFedak, line 165: maybe use `time.Second` to avoid the zeros
```suggestion
				duration := time.Duration(d.meta.Stats.ProcessTime.GetNanos()) + time.Duration(d.meta.Stats.ProcessTime.GetSeconds())*time.Second
```

PFedak (approved):
LGTM, just one readability change

src/server/worker/datum/datum.go
Tryneus: It's not a direct naming conflict, but there are stats on the datum object that I didn't want this to be confused with.

src/server/worker/datum/datum.go
Tryneus: good point, thanks!

tybritten:
can we get this pulled into the next RC?

Tryneus:
@tybritten - this has been cherry-picked into `2.0.x`, it should be in the next RC

PR #6920, opened 2021-09-23
Embedded Enterprise Server w/ Helm doc

acohen4:


doc/docs/master/enterprise/auth/enterprise-server/setup.md
```diff
@@ -20,9 +20,58 @@ The setup of an Enterprise Server requires to:
...
+		oauthIssuer: "http://<PACHD-IP>:30658/"
+		oauthRedirectURI: "http://<PACHD-IP>:30657/authorization-code/callback"
+  		## if a secret is not provided, a secret will be autogenerated on install and stored in the k8s secret 'pachyderm-bootstrap-config.authConfig.clientSecret'
+  		oauthClientSecret: ""
+  		## if a secret is not provided, a secret will be autogenerated on install and stored in the k8s secret 'pachyderm-bootstrap-config.enterpriseSecret'
+  		enterpriseSecret: ""
+  		## if a token is not provided, a secret will be autogenerated on install and stored in the k8s secret 'pachyderm-bootstrap-config.rootToken'
+  		rootToken: ""
+  		## userAccessibleOauthIssuerHost is necessary in localhost settings or anytime the registered Issuer address isn't accessible outside the cluster
+  		# userAccessibleOauthIssuerHost: "localhost:30658"
```
acohen4, line 46: this section could technically be omitted since we handle these variables as optional on install - this would reduce the cognitive load going through this first pass.

On the other hand I think knowing these params exist is instructive

doc/docs/master/enterprise/auth/enterprise-server/setup.md
```diff
@@ -20,9 +20,58 @@ The setup of an Enterprise Server requires to:
...
+  		          "clientSecret": "<CLIENT-SECRET>",
+  		          "redirectURI": "http://<PACHD-IP>:30658/callback",
+  		          "insecureEnableGroups": true,
+  		          "insecureSkipEmailVerified": true,
+  		          "insecureSkipIssuerCallbackDomainCheck": true,
+  		          "forwardedLoginParams": ["login_hint"]
+  		      }
+  		    name: idpConnector
+  		    type: oidc
+	```
```
acohen4, line 67: it may be out of the scope of this doc, but we could add the `console` configuration options here, so that the user can just copy this directly into their values file and deploy our whole stack

doc/docs/master/enterprise/auth/enterprise-server/setup.md
nadegepepin: I agree that these are instructive -  I would leave them there.

chainlink (approved):

doc/docs/master/enterprise/auth/enterprise-server/setup.md
nadegepepin: Discussed this with Alon. The installation of the Console will be documented in a separate page

doc/docs/master/enterprise/auth/enterprise-server/setup.md
```diff
@@ -66,6 +66,12 @@ Update your values.yaml with your enterprise license key, and auth configuration
   		    type: oidc
 	```
 
+!!! Note
+     Update the following values as follow:
+
+	 - `PACHD-IP`: The address of your Pachyderm host. Check the [Expose your Enterprise Server with a Load Balancer](#3-register-your-clusters) note to retrieve Pachyderm external IP address if necessary.
```
acohen4, line 72: ❤️

nadegepepin (approved):
LGTM

PR #6926, opened 2021-09-23
Refactor Before Parallel PPS change

acohen4:
Moving some code around in preparation for another PR

### DIFF
--- a/src/server/pps/server/master.go
+++ b/src/server/pps/server/master.go
@@ -2,18 +2,15 @@ package server
 
 import (
 	"context"
-	"fmt"
 	"path"
 	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/pachyderm/pachyderm/v2/src/internal/backoff"
 	"github.com/pachyderm/pachyderm/v2/src/internal/dlock"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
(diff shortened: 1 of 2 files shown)


src/internal/ppsutil/util.go
```diff
@@ -425,3 +429,38 @@ func GetWorkerPipelineInfo(pachClient *client.APIClient, env serviceenv.ServiceE
 
 	return pipelineInfo, nil
 }
+
+func FindPipelineSpecCommit(ctx context.Context, pfsServer pfsServer.APIServer, txnEnv transactionenv.TransactionEnv, pipeline string) (*pfs.Commit, error) {
+	var commit *pfs.Commit
+	if err := txnEnv.WithReadContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
```
msteffen, line 435: Very minor style idea (and I know this is a pure refactoring PR, so maybe it's better not to change anything at all), but if you wanted I think you could use a named return value, omit the declaration, and save a line e.g.:
```
if err := txnEnv.WithReadContext(ctx, func(txnCtx *txncontext.TransactionContext) (err error) {
	commit, err = FindPipelineSpecCommitInTransaction(txnCtx, pfsServer, pipeline, "")
	return;
}); err != nil {...
```

(per https://tour.golang.org/basics/7)

msteffen (approved):
LGTM, thanks for being so patient!

msteffen:
Context for the PPS Master parallelization project, of which this is part: [Google doc](https://docs.google.com/document/d/1ICe3h45YdMM5x8F1bNBm77jIlnBw4ltuOIDmavSs_54/edit?usp=sharing)

PR #6900, opened 2021-09-20
[WIP] Parallelize PPS Master pipeline Ops w/ Bump()

acohen4:

PR #6927, opened 2021-09-23
PPS Master Parallelization with less noise than #6900

acohen4:


src/server/pps/server/pipeline_controller.go
```diff
@@ -627,25 +708,33 @@ func (op *pipelineOp) restartPipeline(reason string) error {
...
-		"/pps.Master/DeletePipelineResources", "pipeline", pipelineName)
+// deletePipelineResources deletes the RC and services associated with op's
+// pipeline. It doesn't return a stepError, leaving retry behavior to the caller
+func (op *pipelineOp) deletePipelineResources() (retErr error) {
+	if op.pipelineInfo == nil {
+		log.Infof("PPS master: NIL PIPELINE INFO pipeline %q", op.pipelineInfo.Pipeline.Name)
+	}
+	if op.pipelineInfo.Pipeline == nil {
+		log.Infof("PPS master: NIL PIPELINE pipeline %v", op.pipelineInfo.Pipeline)
+	}
```
acohen4, line 719: will remove

src/server/pps/server/master.go
```diff
@@ -73,28 +75,48 @@ type ppsMaster struct {
...
 	watchCancel     func() // protected by pollPipelinesMu
 
+	opsInProcessMu sync.Mutex
+	opsInProcess   map[string]*pipelineOp
+
 	// channel through which pipeline events are passed
 	eventCh chan *pipelineEvent
 }
 
+type monitorManager struct {
```
acohen4, line 93: the monitorManager is an attempt to breakup dependencies so that we can pass in the monitor logic to pipelineOp without keeping a back-reference to the API server. I think this helps us know that pipelineOp goros don't do anything unexpected by reaching into the API server

src/server/pps/server/pipeline_controller.go
```diff
@@ -159,21 +260,6 @@ func (m *ppsMaster) step(pipeline string, timestamp time.Time) (retErr error) {
 		tracing.FinishAnySpan(span, "err", retErr)
 	}()
 
-	// Retrieve pipelineInfo from the spec repo
-	op, err := m.newPipelineOp(opCtx, pipeline)
-	if err != nil {
-		// Don't put the pipeline in a failing state if we're in the middle
-		// of activating auth, retry in a bit
-		if auth.IsErrNotAuthorized(err) || auth.IsErrNotSignedIn(err) {
```
acohen4, line 167: This check was moved here: https://github.com/pachyderm/pachyderm/pull/6927/files#diff-d5f802277550bae9f4934102a2776841987d414328368ada3d674b4cae116435R131

src/server/pps/server/master.go
msteffen: Thinking about this a bit more, if we go with the model where the pipelineOp's lifetime matches the pipeline's lifetime, we maybe don't even need `monitorCancels` or `crashingMonitorCancels`--those could maybe just be fields of the op.

In other words, in this PR there are three maps keyed by pipeline:
- `monitorCancels`
- `crashingMonitorCancels`
- `opsInProcess`

I think, instead, we could maybe just have `opsInProcess` (we should still have a central data structure for these: `opManager` or some such). It would map the pipeline name to its op, which would contain `monitorCancel` (if any) and `crashingMonitorCancel` (if any). We might not need `monitorCancelsMu` anymore--if the ppsMaster wants to cancel these goros, it can just cancel the ops via the `opManager`, which has `opsInProcessMu`.

A cool Go trick worth knowing about: if you embed a mutex in a struct, the struct becomes a mutex:
```
type opManager struct {
  sync.Mutex
  ops map[string]*pipelineOp
}
```
then
```
func (m *opManager) bump(pipeline string) {
  m.Lock()  // just lock m itself. If every method starts this way, m is visually thread-safe
  defer m.Unlock()
  op, ok := m.ops[pipeline]
  if ok { op.opChan <- struct{}{}; }
}
```

PR #6931, opened 2021-09-26
Fix encoding of config files

msteffen:
I previously made a change to fix the way config files are encoded (so that they go through the serde library, which encodes them with jsonpb), and I forgot to add an option to indent the output, which this PR adds.

This also adds a small fix to the cmdutil testing library: `BashCmd` now warns if the given command text can't be dedented properly (which can affect the semantics of bash), in particular catching the case where a bash command in indented with a mix of spaces and tabs (which the go compiler doesn't catch because the bash commands in our tests are strings)

### DIFF
--- a/src/internal/config/config.go
+++ b/src/internal/config/config.go
@@ -231,7 +231,7 @@ func (c *Config) write() error {
 		panic("config V1 included (this is a bug)")
 	}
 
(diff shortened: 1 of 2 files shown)


PFedak (approved):
LGTM

PR #6935, opened 2021-09-27
Rename dash to console (Fixes port forward) 

chainlink:
Remove old dash port options from port forward command


acohen4 (approved):
LGTM. How does this fix port-forward?

chainlink:
targets pods named console instead of dash

PR #6948, opened 2021-09-28
Fix local install of console and update to latest version

chainlink:
To run console locally run the following: 

```
helm install pachyderm etc/helm/pachyderm --set console.enabled=true -f entkeyhelm.yaml -f etc/helm/examples/local-dev-values.yaml
```

Note: running with local-values.yaml does not work due to a bug with the port number assignment when not using node port

Questions 
- [x] Does `REACT_APP_RUNTIME_SUBSCRIPTIONS_PREFIX` now need to be set in all environments?
- [x] Is the current setup for  `REACT_APP_RUNTIME_ISSUER_URI`  best practice? (With trailing slash)


etc/helm/examples/local-dev-values.yaml
```diff
@@ -50,7 +50,7 @@ pachd:
 console:
   enabled: false
   config:
-    issuerURI: "http://pachd:1658/"
+    issuerURI: "http://pachd:1658"
```
acohen4, line 53: is this trailing slash removed on purpose? I think `console.issuerURI `must equal `pachd.oauthIssuer`

etc/helm/pachyderm/values.yaml
```diff
@@ -53,7 +53,7 @@ console:
     #  cpu: "1"
     #  memory: "2G"
   config:
-    issuerURI: ""
+    issuerURI: "" # Cannot have trailing slash
```
acohen4, line 56: what's the reasoning here? Is this discussed in some convo I can reference?

cpepin:
@chainlink @acohen4 
>  Does REACT_APP_RUNTIME_SUBSCRIPTIONS_PREFIX now need to be set in all environments?

Nope, and I'm guessing this will only need to be explicitly set for local deployments. The default value is `/graphql` which works great when the host name isn't localhost `http://localhost/graphql` vs. `https://hub.pachyderm.com/graphql`.

 > Is the current setup for REACT_APP_RUNTIME_ISSUER_URI best practice? (With trailing slash)

I don't know what is necessarily best practice here, I think it's mostly a side-effect of the implementation. We just take the `ISSUER_URI`, append the auth path, and redirect the client browser. It's a naive implementation, and is just using template strings in the client code (i.e. `${issuerUri}${authPath}`). We can presumable make this work with or without a trailing slash, and to me that seems to be the correct approach. I can work on a PR for that.

etc/helm/examples/local-dev-values.yaml
acohen4: based on @cpepin 's  comments, I would think we should remove the trailing slash from `pachd.oauthIssuer`

acohen4 (approved):
LGTM

chainlink:
I'm going to merge this one and then remove when haberdashery#424 lands

PR #6950, opened 2021-09-28
Remove node port exception for ports

chainlink:
This is a breaking change that requires further consideration. This may block helm upgrades for customers that have already installed 2.0RC


chainlink:
Closing in favour of https://github.com/pachyderm/pachyderm/pull/7366

PR #6954, opened 2021-09-29
Fix console ingress annotation

raids:
Based on [`values.yaml`](https://github.com/pachyderm/pachyderm/blob/master/etc/helm/pachyderm/values.yaml#L144-L146), I think the Ingress annotations need be read from `.Values.ingress.annotations`.


chainlink (approved):

chainlink:
Nice catch! @acohen4 This should be on your radar for folks with Ingress issues

PR #6955, opened 2021-09-29
Distributed compaction performance improvements

brycemcanally:


src/internal/miscutil/log.go
```diff
@@ -0,0 +1,22 @@
+package miscutil
+
+import (
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// TODO: refactor into a common logging utility.
+func LogStep(name string, cb func() error) (retErr error) {
```
brendoncarroll, line 10: Not really a fan of this.  I think it would be more idiomatic to return a closure.  Don't need to change it as part of this PR, but I don't think we should port any more logging to this function.

src/server/pfs/server/master.go
```diff
@@ -70,65 +71,78 @@ func (d *driver) finishCommits(ctx context.Context) error {
...
-			validatingDuration := time.Since(start)
-			// Finish the commit.
-			return d.txnEnv.WithWriteContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
-				commitInfo := &pfs.CommitInfo{}
-				if err := d.commits.ReadWrite(txnCtx.SqlTx).Update(pfsdb.CommitKey(commit), commitInfo, func() error {
-					commitInfo.Finished = txnCtx.Timestamp
-					commitInfo.SizeBytesUpperBound = size
-					if commitInfo.Details == nil {
-						commitInfo.Details = &pfs.CommitInfo_Details{}
+		return miscutil.LogStep(fmt.Sprintf("finishing commit %v", commit.ID), func() error {
```
brendoncarroll, line 74: The addition of the callback makes this diff pretty big. I'm under the impression nothing changed here though other than the addition of logging?

src/internal/storage/fileset/compaction.go
```diff
@@ -93,53 +97,68 @@ func NewDistributedCompactor(s *Storage, maxFanIn int, workerFunc CompactionBatc
...
+					tasks = append(tasks, CompactionTask{
+						Inputs:    ids,
+						PathRange: pathRange,
+					})
+					return nil
+				})
+			}); err != nil {
+				return err
+			}
+			taskLens = append(taskLens, len(tasks)-taskLen)
```
brendoncarroll, line 128: I think this line and above can be outside the renewer?

src/internal/storage/fileset/compaction.go
```diff
@@ -93,53 +97,68 @@ func NewDistributedCompactor(s *Storage, maxFanIn int, workerFunc CompactionBatc
...
+			}
+			if len(results) != len(tasks) {
+				return errors.Errorf("results are a different length than tasks")
+			}
+			for _, taskLen := range taskLens {
+				id, err := c.s.Concat(ctx, results[:taskLen], ttl)
+				if err != nil {
+					return err
+				}
+				renewer.Add(*id)
```
brendoncarroll, line 144: Rewewing here makes sense.  Concat takes time, an we were just given an id, before calling Concat again.

src/internal/storage/fileset/compaction.go
```diff
@@ -93,53 +97,68 @@ func NewDistributedCompactor(s *Storage, maxFanIn int, workerFunc CompactionBatc
...
+				return err
+			}
+			taskLens = append(taskLens, len(tasks)-taskLen)
 		}
-		id, err := c.Compact(ctx, ids[start:end], ttl)
-		if err != nil {
-			return nil, err
+		var resultIds []ID
+		if err := miscutil.LogStep(fmt.Sprintf("compacting %v tasks", len(tasks)), func() error {
+			results, err := c.workerFunc(ctx, tasks)
```
brendoncarroll, line 132: In the previous algorithm, we renewed the results from the workerFunc.  This is a spot where we are about to do something that takes significant time (Concat), and we have just been given ids which are not renewed.

src/internal/storage/fileset/compaction.go
brendoncarroll: Yeah all of this just operates on that input, which we can assume will not expire.

src/internal/storage/fileset/compaction.go
```diff
@@ -93,53 +97,68 @@ func NewDistributedCompactor(s *Storage, maxFanIn int, workerFunc CompactionBatc
...
-		tasks = append(tasks, CompactionTask{
-			Inputs:    ids,
-			PathRange: pathRange,
-		})
-		return nil
+		if len(resultIds) == 1 {
+			id = &resultIds[0]
+			return nil
+		}
+		id, err = c.Compact(ctx, resultIds, ttl)
```
brendoncarroll, line 156: ✅   Don't renew here because the next thing we do is return.

brendoncarroll (changes requested):
I think there is a spot we are missing renewal.  Also would appreciate moving the stuff unrelated to renewer to outside the `WithRenew`.

src/server/pfs/server/master.go
brycemcanally: Yeah.

src/internal/storage/fileset/compaction.go
brycemcanally: Yeah, will do.

src/internal/storage/fileset/compaction.go
brycemcanally: Yeah, we actually need to renew at collection time, particularly with this change. We are going to lay out a lot of work, so we need to start renewing immediately when the master receives the results from a task.

brendoncarroll (approved):

PR #6958, opened 2021-09-29
Helm Auth Config with updated console

acohen4:


etc/helm/examples/local-dev-values.yaml
```diff
@@ -14,19 +14,11 @@ pachd:
   metrics:
     enabled: false
   clusterDeploymentID: dev
-  ## if pachd.enterpriseLicenseKey is set, the following values will be interpreted to setup Auth
-  # enterpriseLicenseKey: ""
-  oauthIssuer: "http://pachd:1658"
-  oauthRedirectURI: "http://localhost:30657/authorization-code/callback"
-  userAccessibleOauthIssuerHost: "localhost:30658"
-  mockIDP: true
+  activateEnterprise: false
```
acohen4, line 17: add this to doc....

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -24,26 +24,28 @@ stringData:
...
+    {{- end }}
+    redirect_uri: {{ include "pachyderm.pachdRedirectURI" . }}
     scopes:
     - email
     - profile
     - groups
     - openid 
 
-{{ if eq .Values.deployTarget "LOCAL" }}
+  # TODO: When should this _not_ be specified?
```
acohen4, line 42: I think I can remove this entirely. It's supposed to serve as an example for pushing pachyderm Cluster Role bindings to pachd

chainlink (changes requested):
First pass

chainlink:
Looks like we can totally eliminate the local-values.yaml file with this PR 🎉

chainlink:
Need to change the check in etc/helm/pachyderm/templates/pachd/config-job.yaml to `activateEnterprise` instead of `enterpriseLicenseKey`

etc/helm/pachyderm/templates/pachd/config-secret.yaml
acohen4: yeah sure. How does having this piece in a comment block sound?

etc/helm/pachyderm/values.yaml
```diff
@@ -53,15 +53,15 @@ console:
...
     #  memory: "2G"
   config:
-    issuerURI: "" # Cannot have trailing slash
     reactAppRuntimeIssuerURI: "" # Inferred if running locally or using ingress
     oauthRedirectURI: "" # Infered if running locally or using ingress
-    oauthClientID: "dash"
+    oauthClientID: "console"
     oauthClientSecret: "" # Autogenerated on install if blank
     graphqlPort: 4000
     oauthPachdClientID: pachd
```
chainlink, line 61: `oauthPachdClientID` should be removed

chainlink (changes requested):
One small change, but otherwise looks good

chainlink:
Can you add docs to both the helm docs and to the values.yaml file for the new fields?

PR #6959, opened 2021-09-30
- Add NodePort rules to pachd issuer URI

chainlink:
- Remove trailing slash from reactAppRuntimeIssuerURI in local case
- Add hosts section to TLS section of ingress
- Change config pod enterpriseClusters and enterpriseConfig to use peer port instead of main pachd port for compatibility with TLS deployments where main port is encrypted
- Change console port in port forward to match console (4000)


acohen4 (approved):
Looks good!

PR #6970, opened 2021-10-04
Collect etcd and pg-bouncer logs in debug dumps.

PFedak:
Also add extensions to some files in the dump.


src/server/debug/server/server.go
```diff
@@ -122,11 +123,44 @@ func (s *debugServer) handleRedirect(
 					return err
 				}
 			}
+
```
brycemcanally, line 126: nit, I tend to prefer not having arbitrary spaces.

PFedak: ah sure, I initially had the function inline and I guess I left this when I moved it

PR #6971, opened 2021-10-04
Potential modification to pipelineOp goroutine management

acohen4:
Run pipelineOp goro until the pipeline is deleted, and use a single channel with buf.size = 1 for bump semantics. Delete Ops are assumed by a NotFound Error when querying pipelineInfo


src/server/pps/server/pipeline_controller.go
```diff
@@ -249,6 +201,25 @@ func (op *pipelineOp) step(timestamp time.Time) (retErr error) {
 			retry: false,
 		}
 	}); err != nil {
+		// I don't like the IsNotFoundError works with string comparison.
```
acohen4, line 204: thoughts about this comment @msteffen @PFedak ?

src/server/pps/server/pipeline_controller.go
```diff
@@ -249,6 +201,25 @@ func (op *pipelineOp) step(timestamp time.Time) (retErr error) {
...
+		// What if we add a not found in the message in an instance where the DB records are indeed foud?
+		if errutil.IsNotFoundError(err) {
+			// we interpret no pipeline def as this event being a delete event
+			if err := op.deletePipelineResources(); err != nil {
```
acohen4, line 208: An issue I have with this implementation:
- Does there exist a race condition where the pipelineInfo enters the delete condition, and is bumped before we remove it from the active pipelineOps map? I think this is extremely unlikely, and maybe impossible? But without being sure this makes me uncomfortable. We could expand the critical section to the beginning of the NotFound check, but then we're locking while doing k8s I/O which we really shouldn't be.
cc @msteffen

msteffen: In general, in either this (the "Channel Approach") or #6927 (the "Counter Approach"), what needs to happen inside the critical path of the pipelineOp removing itself from the map is 1) confirming that there are no pending updates, and 2) removing the op from the map. So:
```
func tryFinish() bool {
  op.opsLock.Lock()
  defer op.opsLock.Unlock()
  select {
    case <-op.bumpChan:
      op.bump()  // just cleared the event from the channel >_<. It's ok, just bump ourselves again.
      return false
    default:
      delete(pipeline, opMap)
  }
}
```
In the "channel approach", however, the op _doesn't_ need to hold opsLock in order to check if there are pending updates _outside of the delete path_. So one version of this that is not-recommended and inefficient, but _is_ thread-safe and shows off the difference, is:
```
eventloop:
  for {
    select {
    case <-op.bumpChan:
      op.step()
    default:
      if op.tryFinish() {  // do _another_ select, but this time with the lock held
        return  // op is dead
      }
      continue  // more work to do
    }
  }
}
```

Now, the approach above is not so different from the Counter Approach in overhead, though, because the op attempts to delete itself as soon as `op.bumpChan` is empty: you avoid a lock in `eventloop`, but then immediately acquire a lock you were avoiding in `tryFinish()`. However, where the Channel Approach excels is if you leave the op around until the pipeline is deleted, and then you can mostly avoid the lock:

```
eventloop:
for <-op.bumpChan {  // this runs many times, with no locking
  if step() == opFinished {
    return  // op is dead (only runs ~once)
  }
}
...
func step() error {
  if op.getPipelineInfo() == NotFound {  // only runs ~once in the pipeline's lifetime
    return delete()
  }
  ...
}
...
func delete() error {
  k8sError := op.deletePipelineResources();
  if tryFinish() {  // this locks, but delete() only runs ~once
    // Op not in the map anymore. Die without touching k8s.
    if k8sError != nil { log(k8sError); }
    return opFinished;
  }
  return k8sError; // we are in the map and there's an event. Go back to eventloop
}
```

PR #6975, opened 2021-10-08
Implement group and join input streaming

PFedak:


PFedak:
still need to manage iterator contexts, might be a better way

PR #6976, opened 2021-10-08
Removes postgres pachd init-containers

brendoncarroll:
Removes init-containers used to wait until postgres is ready; pachd now gracefully waits for postgres to come up.

Merge after #6974


tybritten:
Awesome. IMHO the 'db not ready' log should be at least info, not debug. when we were troubleshooting a db connection issue w a customer, the only way we noticed it was the pg bouncer logs. The regularly deployed pachd logs were just silent.

PR #6977, opened 2021-10-09
make pachd cluster scoped resource naming dynamic to avoid conflicts

hobbsh:
This PR fixes cluster scoped resource naming (all RBAC resources) to be dynamic in case there are multiple pachyderm releases in a single cluster. It also updates the non-cluster scoped resources to be dynamic as well.


hobbsh:
> Does that address all of the resource naming problems from your perspective?

@acohen4 Yes it does, thanks! I also totally missed the documentation around the value that controlled the cluster-level RBAC resources, so that would've sufficed too. It is definitely good to have the dependency removed entirely though!

PR #7026, opened 2021-10-22
Breaking out an opStep object to contain step specific state

acohen4:


acohen4:
Closing - these changes were addressed in https://github.com/pachyderm/pachyderm/pull/7064

PR #7032, opened 2021-10-27
pachsql: ParseURL

brendoncarroll:
Adds a ParseURL function to pachsql.  Renamed package `sql` to `pachsql` to avoid conflict with standard library package name.


src/internal/pachsql/url.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, err
+	}
+	port, err := strconv.Atoi(u.Port())
+	if err != nil {
+		return nil, err
+	}
+	params := make(map[string]string)
+	for k, v := range u.Query() {
+		if len(v) > 0 {
+			params[k] = v[len(v)-1]
```
acohen4, line 32: do we already know what param keys may have multiple values?

src/internal/pachsql/url.go
```diff
@@ -0,0 +1,104 @@
...
+	case "mysql":
+		driver = "mysql"
+		dsn = mySQLDSN(u, password)
+	default:
+		return nil, errors.Errorf("database protocol %q not supported", u.Protocol)
+	}
+	return sqlx.Open(driver, dsn)
+}
+
+func postgresDSN(u URL, password string) string {
```
acohen4: does postgres not support additional params? Thinking about the unused `u.Params` field in this func

src/internal/pachsql/url_test.go
```diff
@@ -0,0 +1,46 @@
...
+				Host:     "10.0.0.1",
+				Port:     9000,
+				Database: "mydb",
+				Params: map[string]string{
+					"sslmode": "disable",
+				},
+			},
+		},
+		{
+			In: "mysql://jbond@10.0.0.1:1007/martini?shaken=true&stirred=false",
```
acohen4, line 27: lol <3

src/internal/pachsql/db.go
```diff
@@ -0,0 +1,61 @@
...
+	case "mysql":
+		driver = "mysql"
+		dsn = mySQLDSN(u, password)
+	default:
+		return nil, errors.Errorf("database protocol %q not supported", u.Protocol)
+	}
+	return sqlx.Open(driver, dsn)
+}
+
+func postgresDSN(u URL, password string) string {
```
acohen4, line 35: Should the URL params also be passed to the PG DSN?

src/internal/pachsql/url.go
brendoncarroll: I don't think any of them will have multiple values.  A URL is allowed have multiple values per parameter though.

src/internal/pachsql/db.go
brendoncarroll: Yeah, good catch, will add that.

src/internal/pachsql/db.go
```diff
@@ -0,0 +1,64 @@
...
+		fields["password"] = password
+	}
+	for k, v := range u.Params {
+		fields[k] = v
+	}
+	var dsnParts []string
+	for k, v := range fields {
+		dsnParts = append(dsnParts, k+"="+v)
+	}
+	return strings.Join(dsnParts, " ")
```
acohen4, line 52: Should have asked this before, is the postgres DSN unordered?

acohen4 (approved):
LGTM!

src/internal/pachsql/db.go
brendoncarroll: I'm pretty sure it is.  This section you commented on is taken from `dbutil`

PR #7034, opened 2021-10-28
Replaces `sqlx.DB` with `pachsql.DB`

brendoncarroll:
Blocked on #7032 

This should be a non-functional change.  Just a refactor, split out from the db URL PR.


src/internal/ppsutil/util.go
```diff
@@ -32,6 +32,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/dbutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
```
acohen4, line 35: Should the sqlx import be removed here?

acohen4 (approved):
LGTM. There are a few places where sqlx is still being imported. It mostly looks like we could remove them

PR #7035, opened 2021-10-28
Structured Data Package with SQL Result Materialization

brendoncarroll:
Creates a structured data package `sdata` with a `TupleWriter` interface and writers for JSON and CSV formats.  Adds a matrix test to check that we can successfully materialize data from all combinations of database and materialization format.

- `sdata` MaterializeSQL can be used to copy rows from a `sql.Rows` to a `TupleWriter`.
- `dockertestenv`: Adds `NewMySQL` for creating a fresh MySQL database suitable for tests.

Needs #7032 #7084


brendoncarroll:
The failing tests are from `sql.NullInt16` missing in go versions <1.17.  I have another PR up to bump the go version.

src/internal/sdata/writers.go
```diff
@@ -0,0 +1,216 @@
...
+	headers []string
+
+	headersWritten bool
+	record         []string
+}
+
+// NewCSVWriter returns a CSVWriter writing to w.
+// If len(headers) == 0, then no headers will be written
+func NewCSVWriter(w io.Writer, headers []string) *CSVWriter {
+	return &CSVWriter{cw: csv.NewWriter(w)}
```
acohen4, line 49: missing setting the headers on the CSVWriter?

src/internal/sdata/writers.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func (m *CSVWriter) WriteTuple(row Tuple) error {
```
acohen4, line 52: `len(row) == len(headers)`, always right?

src/internal/sdata/writers.go
```diff
@@ -0,0 +1,216 @@
...
+
+func (m *JSONWriter) WriteTuple(row Tuple) error {
+	if len(row) != len(m.fields) {
+		return ErrTupleFields{Writer: m, Fields: m.fields, Tuple: row}
+	}
+	record := m.record
+	for i := range row {
+		var y interface{}
+		switch x := row[i].(type) {
+		case *sql.NullInt16:
```
acohen4, line 170: should this case also include the type `*int16` as in the CSVWriter?

acohen4: nvm, we that's handled in the default at the bottom 👍

src/internal/sdata/sdata_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/jmoiron/sqlx"
+	"github.com/pachyderm/pachyderm/v2/src/internal/dockertestenv"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
+	"github.com/pachyderm/pachyderm/v2/src/internal/require"
+)
+
+// TestMaterializeSQL checks that rows can be materialized from all the supported databases,
+// with all the supported writers.
+// It does not check that the writers themselves output in the correct format.
+func TestMaterializeSQL(t *testing.T) {
```
acohen4, line 25: sweet test 👍

src/internal/sdata/sdata_test.go
```diff
@@ -0,0 +1,185 @@
...
+		cols = append(cols, col)
+	}
+	return "(" + strings.Join(cols, ", ") + ")"
+}
+
+func formatValues(x interface{}, db *pachsql.DB) string {
+	var placeholder func(i int) string
+	switch db.DriverName() {
+	case "pgx":
+		placeholder = func(i int) string { return "$" + strconv.Itoa(i+1) }
```
acohen4, line 148: didn't realize the format strings differ between drivers. Figured it was the same since it's evaluated in the sql library

src/internal/sdata/writers.go
brendoncarroll: Yeah.  That should never come up, but if it did, it wouldn't panic with out of bounds below, you would just get more columns for the entry.

acohen4: makes sense! My only follow up question would be: would it make sense to include the validation present in JSONWriter below that throws `ErrTupleFields` error

acohen4 (approved):
Looks great! I left one small follow up questions

PR #7037, opened 2021-10-29
Rename pipelineOp -> pipelineController

acohen4:
This change is purely aesthetic


PFedak (approved):
LGTM, agree that op doesn't really make sense any more

src/server/pps/server/pipeline_controller.go
```diff
@@ -69,7 +75,7 @@ type pipelineOp struct {
 	crashingMonitorCancel func()
 
 	bumpChan  chan struct{}
-	opManager *opManager
+	opManager *pcManager
```
msteffen, line 78: would it make sense to rename opManager -> pcMgr as well? (per the change in the ppsMaster?)

src/server/pps/server/pipeline_controller.go
```diff
@@ -93,21 +99,21 @@ func (m *ppsMaster) newPipelineOp(ctx context.Context, cancel context.CancelFunc
...
 //
 // This function is expected to be called while holding the opManager lock.
 // This helps guard the critical section when the goroutine is cleaned up
-// as part of a pipeline deletion event in pipelineOp.tryFinish().
+// as part of a pipeline deletion event in pipelineController.tryFinish().
 //
 // Note: since op.bumpChan is a buffered channel of length 1, Bump() will
 // add to the channel if it's empty, and do nothing otherwise.
-func (op *pipelineOp) Bump() {
+func (op *pipelineController) Bump() {
```
msteffen, line 116: would it make sense to change the receiver names from `op` to `pc` here as well?

msteffen (commented):
Overall LGTM, with two extra naming changes

src/server/pps/server/pipeline_controller.go
acohen4: it would make so much sense! lol . I caught this after merging - I'll make another PR to change these

PR #7040, opened 2021-10-29
Concurrency Limiter in PPS Master to protect k8s API server from becoming overwhelmed

acohen4:


src/server/pps/server/pipeline_controller.go
```diff
@@ -187,6 +198,11 @@ func (op *pipelineController) tryFinish() {
 func (op *pipelineController) step(timestamp time.Time) (retErr error) {
 	log.Debugf("PPS master: processing event for %q", op.pipelineInfo.Pipeline)
 
+	// the limiter intends to guard the k8s API server from being overwhelmed by many concurrent requests
+	// that could arise from many concurrent pipelineController goros.
```
msteffen, line 202: This is probably something that has to be profiled, but I'm a little anxious that the critical path here is bigger than I expected—I think that, to just capture the k8s RPCs, we'd just hold the limiter in `getRC()`, `updateRC()` and `createPipelineResources()`, but this will also cover reading the pipeline info and such.

Do you think it's better to capture the whole update? Would it make sense to just hold the limiter during those calls?

msteffen (commented):
Two questions, but only one about structure

src/server/pps/server/pipeline_controller.go
```diff
@@ -37,6 +38,8 @@ const (
 	rcExpected
 )
 
+const concurrency int = 10
```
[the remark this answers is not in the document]
acohen4: I like it!

src/server/pps/server/pipeline_controller.go
acohen4: very good Q - I agree that adding the limiter in the lower level calls would allow for more throughput. It makes sense to me to do it that way. The decision to make the limiter work at a coarser level was more geared towards simplicity and being easy to reason about, but I don't think it's much simpler than your suggestion

src/server/pps/server/pipeline_controller.go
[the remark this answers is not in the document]
msteffen: actually, second thought: it's more annoying, but would it make sense to tie this to an env var in `env.Config`? I just have visions of some customer running into this and us being stuck

src/server/pps/server/pipeline_controller.go
acohen4: @msteffen so I just made the change - the one thing that I now realize makes me nervous is the possibility that one of these functions that call `limiter.Acquire()` could in theory call another one, meaning that we need to be careful about this code in the future. For that reason, I actually think I would prefer the coarser locking. What do you think?

acohen4: We could also eliminate the risk by adding a little more complexity - perhaps wrapping the limiter calls with a check to see if this pc already hit acquire. Not sure if it's worthwhile though

src/server/pps/server/pipeline_controller.go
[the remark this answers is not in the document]
acohen4: 100% !

src/server/pps/server/pipeline_controller.go
msteffen: My opinion is that that guarantee would be best to enforce at the programmer layer: put scary comments around the functions, or put them in their own file with a stern admonition to never add recursive calls to them (like we had in monitor.go).

I don't really know what the right thing to do here is, and I think to find out we'd have to measure the performance impact of different locking granularities (I don't know if we have any PPS scale tests but that seems like the way to do it). My instinct would be to use the slightly finer locking in `get/updateRC()` and `createPipelineResources()` and avoid introducing reentrant deadlocks with comments, but I have no evidence that the finer locking matters and you're the one in the trenches right now, so if you think courser locking is better I'm happy to approve it.

msteffen (approved):
LGTM

src/server/pps/server/pipeline_controller.go
acohen4: I don't really know what's best here either, but I think I'll go with your suggested approach of handling it at the programmer level makes sense. I create another PR to move all the k8s client methods to the bottom of the file here: https://github.com/pachyderm/pachyderm/pull/7045

PR #7047, opened 2021-11-01
Further improve commit finishing performance

brycemcanally:
This PR makes several changes to commit finishing that should improve the overall performance.

**Distribute shard and concat tasks during compaction:**
The prior implementation of compaction would process the shard and concat tasks in the PFS master, which is heavy lifting since it requires getting chunks and going through all of the file metadata when compacting a list of file sets. Now, these two tasks are represented similarly to the way a compact task is represented and will be distributed to compaction workers in the cluster. This change is useful because it distributes work that can be processed in parallel, but also because it removes the majority of the heavy lifting from the PFS master which is responsible for processing the finishing of all of the commits in the system and garbage collection.

**Parallelize commit finishing:**
Since the majority of the heavy lifting has been removed from the PFS master, we can now parallelize the finishing of commits. The commit finishing parallelization is handled by a goroutine per repo. This model makes more sense than a set of workers that can process any commit because the ordering of commit finishing matters within repos, but not across repos.

**Smaller performance changes:**
- Change the compaction property check to iterate back through the larger levels (starting at the last that satisfies the compaction property) to determine which level to start the compaction from. This prevents recursively compacting the levels which can add a significant amount of duplicate work when compacting a commit.
- Reorder job commit finishing to ensure the meta commit is finished before output commit. This change makes it such that standby / autoscaling won't scale down the workers before the compaction of the meta commit is completed.
- Compact file sets created through the file set API to keep the fan in of a file set below the max fan in (prevents reads that can thrash the cache when reading the file set directly).
- Add a limiter to the chunk writer to throttle the data being written in (this is normally handled indirectly, but the file set compaction change does not throttle the data being written to the chunk writer).

**Bug fixes:**
- Fix a gRPC connection leak that occurs when connecting to the worker API servers.
- Prevent the job registry from processing jobs in the finishing state (jobs that are finished as far as the registry is concerned, but not finished in terms of the output / meta commits). Also, prevent the clearing of commits in the finishing state.
- Fix a deadlock in the postgres listener.


src/internal/ppsutil/util.go
```diff
@@ -333,14 +333,14 @@ func FinishJob(pachClient *client.APIClient, jobInfo *pps.JobInfo, state pps.Job
 	// store commit states.
 	_, err := pachClient.RunBatchInTransaction(func(builder *client.TransactionBuilder) error {
 		if _, err := builder.PfsAPIClient.FinishCommit(pachClient.Ctx(), &pfs.FinishCommitRequest{
-			Commit: jobInfo.OutputCommit,
```
brendoncarroll, line 336: Why does the MetaCommit have to come first?

src/internal/storage/chunk/writer.go
```diff
@@ -89,7 +89,8 @@ func newWriter(ctx context.Context, client Client, memCache kv.GetPut, deduper *
 		},
 		buf:   &bytes.Buffer{},
 		stats: &stats{},
-		chain: NewTaskChain(cancelCtx),
+		// TODO: Make task chain parallelism configurable?
+		chain: NewTaskChain(cancelCtx, 100),
```
brendoncarroll, line 93: Don't we bottleneck on the rolling hash here?  Might make sense to set this to `runtime.GOMAXPROCS(0)` if it's CPU bound.

src/internal/work/work.go
```diff
@@ -117,11 +118,20 @@ type Master struct {
 	*taskEtcd
 	taskID    string
 	taskEntry *taskEntry
+	ctx       context.Context
```
brendoncarroll, line 121: Not a fan of storing contexts.  But you say below this will be replaced by the TaskService, so not a big deal.

src/server/debug/server/server.go
```diff
@@ -242,6 +242,11 @@ func (s *debugServer) handleWorkerRedirect(tw *tar.Writer, pod *v1.Pod, collectW
 	if err != nil {
 		return err
 	}
+	defer func() {
+		if err := c.Close(); retErr == nil {
```
brendoncarroll, line 246: I don't think it makes sense to return this error.  Just logging it would be fine.

src/server/pfs/server/driver.go
```diff
@@ -1625,7 +1625,7 @@ func (d *driver) clearCommit(ctx context.Context, commit *pfs.Commit) error {
 	if err != nil {
 		return err
 	}
-	if commitInfo.Finished != nil {
+	if commitInfo.Finishing != nil {
 		return errors.Errorf("cannot clear finished commit")
```
brendoncarroll, line 1629: Does this error need to update to reflect the changed behavior?

src/server/pfs/server/driver_file.go
```diff
@@ -92,7 +92,7 @@ func (d *driver) withCommitUnorderedWriter(ctx context.Context, renewer *fileset
 	return d.commitStore.AddFileSet(ctx, commit, *id)
 }
 
-func (d *driver) withUnorderedWriter(ctx context.Context, renewer *fileset.Renewer, compact bool, cb func(*fileset.UnorderedWriter) error, opts ...fileset.UnorderedWriterOption) (*fileset.ID, error) {
```
brendoncarroll, line 95: So now we just always compact.

src/server/pfs/server/env.go
```diff
@@ -60,7 +59,7 @@ func EnvFromServiceEnv(env serviceenv.ServiceEnv, txnEnv *txnenv.TransactionEnv)
 		GetPPSServer:  env.PpsServer,
 		GetPachClient: env.GetPachClient,
 
-		BackgroundContext: ctx,
+		BackgroundContext: env.Context(),
```
brendoncarroll, line 62: :+1:

src/server/pfs/server/master.go
```diff
@@ -58,11 +58,42 @@ func (d *driver) master(ctx context.Context) {
 }
 
 func (d *driver) finishCommits(ctx context.Context) error {
-	compactor, err := newCompactor(d.env.BackgroundContext, d.storage, d.etcdClient, d.prefix, d.env.StorageConfig.StorageCompactionMaxFanIn)
+	repos := make(map[string]context.CancelFunc)
```
brendoncarroll, line 61: Is this missing a deferred for loop to cancel all of these?

src/server/pfs/server/pfsserver.proto
```diff
@@ -1,20 +1,40 @@
 syntax = "proto3";
 
-package pfsserver;
+package fileset;
```
brendoncarroll, line 3: The other named seems more fitting.  It's still in the pfs server package.

src/server/pfs/server/pfsserver.proto
```diff
@@ -1,20 +1,40 @@
...
+  repeated CompactTask compact_tasks = 2;
 }
 
 message PathRange {
   string lower = 1;
   string upper = 2;
 }
+
+message CompactTask {
+  int64 index = 1;
```
brendoncarroll, line 22: Why do all of these have an index field?

src/internal/ppsutil/util.go
brycemcanally: The reason for that change is mentioned in the main comment of the PR. I don't think this change is going to address the issue anymore with the goroutine per repo model though. I will probably need to add a wait on the meta commit in the code that handles the running / standby transition.

src/internal/storage/chunk/writer.go
brycemcanally: The rolling hash can be the bottleneck in certain cases, but it wouldn't be in the copy case since we don't go through the rolling hash in most cases (the concat step which copies the shards and uses the rolling hash for the edges of the shards). Even in the case where we are not copying, if the upload to the object storage provider is slower than the rolling hash, we will just keep buffering chunks in memory until the full data has been written (assuming the data is coming into the writer fast enough, which is the case for the compact at the close of an unordered writer).

src/internal/work/work.go
brycemcanally: Yeah, this is going away with the task service.

src/internal/ppsutil/util.go
brendoncarroll: So we start scaling pipelines before the Transaction is complete.  What if it's rolled back?

src/server/debug/server/server.go
brycemcanally: Yeah, that makes sense. It is probably preferable to be able to complete the request with a log about a leak than just fail.

src/server/pfs/server/driver.go
brycemcanally: Not sure, the clients are exposed to this process through finish commit. The finishing without finished state is just something that causes waiting under the hood.

src/server/pfs/server/driver_file.go
brycemcanally: No, we only compact for file sets created through the file set API. Also, the compaction just keeps the file sets under the fan in, it doesn't compact into one file set.

src/server/pfs/server/master.go
brycemcanally: I was leaning on the context always being canceled if we exit this function (context canceled elsewhere or returning an error canceling the error group context). It is probably not good to rely on that though, so I can add a deferred cancel to ensure those contexts are canceled.

src/server/pfs/server/pfsserver.proto
brycemcanally: Yeah, this was a mistake when shuffling around the protos.

src/server/pfs/server/pfsserver.proto
brycemcanally: The order we collect the results in doesn't correspond with the order the tasks were created, so we need this to map back to the ordering of the tasks.

src/server/pfs/server/compactor.go
```diff
@@ -99,49 +50,302 @@ func (c *compactor) Compact(ctx context.Context, ids []fileset.ID, ttl time.Dura
...
-					return err
-				}
-				result, err = serializeCompactionResult(&CompactionTaskResult{
-					Index: task.Index,
-					Id:    id.HexString(),
-				})
-				return err
-			}); err != nil {
-				return nil, err
+			shardTask, err := deserializeShardTask(subtask.Data)
```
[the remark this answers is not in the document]
brycemcanally: It does error when a type mismatch occurs. It wouldn't be getting through CI if it didn't. I think this pattern is fine, although we may want to special case the mismatch error, like you mentioned, so we can capture other errors immediately. As is, a different error can occur and then be overwritten by the subsequent mismatch errors.

src/internal/ppsutil/util.go
brycemcanally: No, the order of operations in a transaction correspond to the order the relevant events are sent to listeners when the transaction completes (operation order corresponds to the ordering of the row processing of the trigger which performs the notify calls on the relevant channels).

src/server/pfs/server/compactor.go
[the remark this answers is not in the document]
brycemcanally: Actually, I'm going to switch to the switch pattern you suggested. That is cleaner.

brendoncarroll (approved):

PR #7051, opened 2021-11-03
Use a real context in the chunk renewer

brycemcanally:
This PR refactors the chunk renewer setup slightly to allow us to use a real context. A real context is needed to cancel the renewal in the error case because a file set writer is only closed when it is successful.


brendoncarroll (approved):
lgtm

PR #7057, opened 2021-11-03
Let users Deploy a pachd cluster linked to an external Enterprise Server

acohen4:
This PR should enable the following sequence of deployments to represent a pachyderm instance registering with an existing enterprise server:

1.) `helm install pachyderm etc/helm/pachyderm --namespace enterprise --set pachd.enterpriseLicenseKey=$(cat ent.txt) -f etc/helm/examples/enterprise-dev.yaml`

2.) `helm install pachyderm etc/helm/pachyderm --set oidc. issuerURI=<...>  --set pachd.activateEnterpriseMember=true --set pachd.enterpriseServerAddress=<...> --set pachd.enterpriseCallbackAddress=<...> --set pachd.enterpriseRootToken=<...> `


etc/helm/pachyderm/templates/_helpers.tpl
```diff
@@ -89,7 +89,7 @@ http://localhost:4000/oauth/callback/?inline=true
 
 {{- define "pachyderm.pachdRedirectURI" -}}
 {{- if .Values.pachd.oauthRedirectURI -}}
-{{ .Values.console.config.oauthRedirectURI -}}
```
chainlink, line 92: Good catch! This looks like an unrelated bug?

acohen4: Yeah I could break it out into its own PR too

etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -45,21 +45,28 @@ stringData:
...
+{{ if .Values.pachd.activateEnterpriseMember }}
+
+  enterpriseServerAddress: {{ required "To register a pachd instance with an enterprise server, the enterprise server's address must be provided." .Values.pachd.enterpriseServerAddress }}
+
+  enterpriseRootToken: {{ required "To register this pachd as an enterprise member, the enterprise root token must be provided." .Values.pachd.enterpriseRootToken }}
+
   # TODO: Add config option for non embedded / multiple pachs?
   # enterpriseClusters is the set of pachds covered by license service 
   enterpriseClusters: |
-    - address: grpc://localhost:1653
```
chainlink, line 51: Are we getting rid of the localhost case because it's assumed if it's not passed into the config pod?

chainlink (approved):
Small question, but otherwise looks good!

etc/helm/pachyderm/templates/pachd/config-secret.yaml
acohen4: Yeah it's actually being implied by the top level "enterpriseSecret" entry in the secret. When it's provided, the pachd cluster will set local as a client to itself - aka embedded enterprise

acohen4: So in this case, it wasn't necessary to include the enterpriseClusters / enterpriseConfig declarations until now, since they were being implied by the "enterpriseSecret" config.

etc/helm/pachyderm/templates/_helpers.tpl
chainlink: I think that'd be good from a release notes perspective

PR #7058, opened 2021-11-04
Move caching out of chunk.Get

brendoncarroll:


src/internal/storage/chunk/reader.go
```diff
@@ -126,14 +129,43 @@ func (dr *DataReader) DataRef() *DataRef {
...
-		if dr.offset > dr.dataRef.SizeBytes {
-			return errors.Errorf("DataReader.offset cannot be greater than the dataRef size. offset size: %v, dataRef size: %v.", dr.offset, dr.dataRef.SizeBytes)
-		}
+	if dr.offset > dr.dataRef.SizeBytes {
+		return errors.Errorf("DataReader.offset cannot be greater than the dataRef size. offset size: %v, dataRef size: %v.", dr.offset, dr.dataRef.SizeBytes)
+	}
+	ref := dr.dataRef.Ref
+	b := backoff.NewExponentialBackOff()
+	b.InitialInterval = 1 * time.Millisecond
+	cb := func(chunk []byte) error {
```
brycemcanally, line 144: Could we just inline this in the getFromCache call?

brycemcanally (approved):
lgtm, just one minor change. This is definitely cleaner.

src/internal/storage/chunk/reader.go
brendoncarroll: yeah sure

PR #7066, opened 2021-11-05
PipelineController Step Full Changes

acohen4:
Context for the PPS Master parallelization project, of which this is part: [Google doc](https://docs.google.com/document/d/1ICe3h45YdMM5x8F1bNBm77jIlnBw4ltuOIDmavSs_54/edit?usp=sharing)

### DIFF
--- a/src/server/pps/server/pipeline_controller.go
+++ b/src/server/pps/server/pipeline_controller.go
@@ -69,7 +69,6 @@ func newPcManager(maxConcurrentK8sRequests int) *pcManager {
 // pcStep captures the state that's used within a single handling of a pipeline event
 type pcStep struct {
 	pc           *pipelineController
-	ctx          context.Context
 	pipelineInfo *pps.PipelineInfo
 	rc           *v1.ReplicationController
 }
@@ -192,15 +191,9 @@ func (pc *pipelineController) step(timestamp time.Time) (isDelete bool, retErr e
 		tracing.FinishAnySpan(span, "err", retErr)
 	}()
 
-	// if we fail to create a new step, there was an error querying the pipeline info, and there's nothing we can do
-	if step, err := pc.newStep(); err != nil {
-		msg := fmt.Sprintf("failed to set up step data to handle event for pipeline '%s': %v", pc.pipeline, errors.Wrapf(err, "failing pipeline %q", pc.pipeline))
-		log.Errorf("PPS master: " + msg)
-		failError := step.setPipelineFailure(msg)
-		if failError != nil {
-			log.Errorf("PPS master: error failing pipeline '%s': %v", pc.pipeline,
-				errors.Wrapf(failError, "error failing pipeline %q, after error (%v)", pc.pipeline, err))
-		}
+	if step, ctx, err := pc.newStep(); err != nil {
+		// if we fail to create a new step, there was an error querying the pipeline info, and there's nothing we can do
+		log.Errorf("PPS master: failed to set up step data to handle event for pipeline '%s': %v", pc.pipeline, errors.Wrapf(err, "failing pipeline %q", pc.pipeline))
 		return false, err
 	} else if step == nil {
 		// interpret the event as a delete operation
@@ -217,12 +210,12 @@ func (pc *pipelineController) step(timestamp time.Time) (isDelete bool, retErr e
 			// set pc.rc
 			// TODO(msteffen) should this fail the pipeline? (currently getRC will restart
 			// the pipeline indefinitely)
-			if err := step.getRC(noExpectation); err != nil && !errors.Is(err, errRCNotFound) {
+			if err := step.getRC(ctx, noExpectation); err != nil && !errors.Is(err, errRCNotFound) {
 				return err
 			}
 
 			// Create/Modify/Delete pipeline resources as needed per new state
-			return step.updateRcAndState()
+			return step.updateRcAndState(ctx)
 		}, backoff.NewExponentialBackOff(), func(err error, d time.Duration) error {
 			errCount++
 			if errors.As(err, &stepErr) {
@@ -235,7 +228,7 @@ func (pc *pipelineController) step(timestamp time.Time) (isDelete bool, retErr e
 			return errors.Wrapf(err, "could not update resource for pipeline %q", pc.pipeline)
 		})
 		if errors.As(err, &stepErr) && stepErr.failPipeline {
-			failError := step.setPipelineFailure(fmt.Sprintf("could not update resources after %d attempts: %v", errCount, err))
+			failError := step.setPipelineFailure(ctx, fmt.Sprintf("could not update resources after %d attempts: %v", errCount, err))
 			if failError != nil {
 				log.Errorf("PPS master: error starting a pipelineController for pipeline '%s': %v", pc.pipeline,
 					errors.Wrapf(failError, "error failing pipeline %q (%v)", pc.pipeline, stepErr))
@@ -246,16 +239,16 @@ func (pc *pipelineController) step(timestamp time.Time) (isDelete bool, retErr e
 }
 
 // returns nil, nil if the step is found to be a delete operation
-func (pc *pipelineController) newStep() (*pcStep, error) {
+func (pc *pipelineController) newStep() (*pcStep, context.Context, error) {
 
 	// query pipelineInfo
 	var pi *pps.PipelineInfo
 	var err error
 	if pi, err = pc.tryLoadLatestPipelineInfo(); err != nil && collection.IsErrNotFound(err) {
 		// if the pipeline info is not found, interpret the operation as a delete
-		return nil, nil
+		return nil, nil, nil
 	} else if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	tracing.TagAnySpan(pc.ctx,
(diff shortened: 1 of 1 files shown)


src/server/pps/server/pipeline_controller.go
```diff
@@ -188,48 +190,34 @@ func (pc *pipelineController) step(timestamp time.Time) (retErr error) {
...
-	// Process the pipeline event
-	var errCount int
 	var stepErr stepError
+	errCount := 0
 	err = backoff.RetryNotify(func() error {
 		// set pc.rc
 		// TODO(msteffen) should this fail the pipeline? (currently getRC will restart
 		// the pipeline indefinitely)
-		if err := step.getRC(stepCtx, noExpectation); err != nil && !errors.Is(err, errRCNotFound) {
+		if err := step.getRC(noExpectation); err != nil && !errors.Is(err, errRCNotFound) {
```
msteffen, line 220: minor style nit, but now that I see the code laid out, with `getRC` and `run()` together in this retry loop, I wonder if it would make sense to rename `run()` to `tryUpdateRC()` or some such—make the retry loop the "RC portion" of the logic, basically.

src/server/pps/server/pipeline_controller.go
```diff
@@ -149,11 +150,9 @@ func (pc *pipelineController) Start(timestamp time.Time) {
 		case <-pc.ctx.Done():
 			return
 		case <-pc.bumpChan:
-			err := pc.step(timestamp)
-			// we've given up on the step, check if the error indicated that the pipeline should fail
-			if err != nil {
-				log.Errorf("PPS master: error starting a pipelineController for pipeline '%s': %v", pc.pipeline,
-					errors.Wrapf(err, "failing pipeline %q", pc.pipeline))
+			// pc.step returns true if the pipeline was deleted, and the controller can try to shutdown
+			if pc.step(timestamp) {
```
msteffen, line 154: I haven't read the surrounding code so this suggestion may make no sense, but: do we use `timestamp` outside of `pc.step()`? If not, would it make sense to instantiate it inside of `pc.step()`? (basically to limit its lifetime to the function it's used in)

src/server/pps/server/pipeline_controller.go
```diff
@@ -188,48 +190,34 @@ func (pc *pipelineController) step(timestamp time.Time) (retErr error) {
...
-	var pi *pps.PipelineInfo
-	var err error
-	if pi, err = pc.tryLoadLatestPipelineInfo(); err != nil && collection.IsErrNotFound(err) {
-		// if the pipeline info is not found, interpret the operation as a delete
+	// if we fail to create a new step, there was an error querying the pipeline info, and there's nothing we can do
+	step, err := pc.newStep()
+	if err != nil {
+		finalErr = err
+		log.Errorf("PPS master: failed to set up step data to handle event for pipeline '%s': %v", pc.pipeline,
+			errors.Wrapf(err, "failing pipeline %q", pc.pipeline))
```
msteffen, line 201: I may be misreading this, but I think there may be a bug here, where the log line says that it's failing the pipeline, but the pipeline doesn't get failed. If it makes sense to move this into the retry loop with `getRC`, that could be one fix, but we could also call `setPipelineFailure` in two places.

src/server/pps/server/pipeline_controller.go
```diff
@@ -188,48 +190,34 @@ func (pc *pipelineController) step(timestamp time.Time) (retErr error) {
 		tracing.TagAnySpan(span, "pollpipelines-event", "true")
 	}
 	defer func() {
-		tracing.FinishAnySpan(span, "err", retErr)
+		tracing.FinishAnySpan(span, "err", finalErr)
```
msteffen, line 193: If you're going to read `finalErr` in a defer statement, I think it might make sense to make `finalErr` a return value and return it, even if the caller doesn't read it. I think that the risk of using a local variable in a defer statement is that local variables are easy to accidentally shadow, but if it's a return variable, then anywhere that calls `return x, err` will work.

src/server/pps/server/pipeline_controller.go
```diff
@@ -188,48 +190,34 @@ func (pc *pipelineController) step(timestamp time.Time) (retErr error) {
...
+	// if we fail to create a new step, there was an error querying the pipeline info, and there's nothing we can do
+	step, err := pc.newStep()
+	if err != nil {
+		finalErr = err
+		log.Errorf("PPS master: failed to set up step data to handle event for pipeline '%s': %v", pc.pipeline,
+			errors.Wrapf(err, "failing pipeline %q", pc.pipeline))
+	}
+
+	// interpret the event as a delete operation
+	if step == nil {
```
msteffen, line 205: Would it make sense to make this an `else if` off of the above block? Like
```
step, err := pc.newStep();
if err != nil {
  log.Errorf("failing...")
  err := pc.failPipeline()
  return true, err
} else if step == nil {
  log.Printf("deleting...")
  err := pc.deletePipelineResources()
  return true, err
}

backoff.RetryNotify(... rc stuff ...)
```

src/server/pps/server/pipeline_controller.go
```diff
@@ -188,48 +190,34 @@ func (pc *pipelineController) step(timestamp time.Time) (retErr error) {
...
-		pc:           pc,
-		ctx:          stepCtx,
-		pipelineInfo: pi,
+		return true
 	}
 
-	// Process the pipeline event
-	var errCount int
 	var stepErr stepError
+	errCount := 0
```
msteffen, line 215: doesn't matter, but just FYI ints are initialized to 0 by default in Go.

msteffen (commented):
bunch of minor suggestions, but this is looking very close!

src/server/pps/server/pipeline_controller.go
acohen4: yeah that's much more descriptive than run. Do you have a preference between `tryUpdateRC()` and `updateRcAndState()`?

src/server/pps/server/pipeline_controller.go
acohen4: You're right. I updated the message. There's actually nothing we can do in this case, since we weren't able to retrieve the pipelineInfo from the DB. We can only log the error

src/server/pps/server/pipeline_controller.go
```diff
@@ -149,11 +149,9 @@ func (pc *pipelineController) Start(timestamp time.Time) {
 		case <-pc.ctx.Done():
 			return
 		case ts := <-pc.bumpChan:
-			err := pc.step(ts)
-			// we've given up on the step, check if the error indicated that the pipeline should fail
-			if err != nil {
-				log.Errorf("PPS master: error starting a pipelineController for pipeline '%s': %v", pc.pipeline,
-					errors.Wrapf(err, "failing pipeline %q", pc.pipeline))
+			// pc.step returns true if the pipeline was deleted, and the controller can try to shutdown
+			if isDelete, _ := pc.step(ts); isDelete {
```
msteffen, line 153: Without having traced the code too carefully, it seems like it might still make sense to log any errors here, since they're still returned? Wdyt?

msteffen (approved):
One small suggestion, but overall LGTM! Nice!

src/server/pps/server/pipeline_controller.go
msteffen: I suggested returning this error and then having the caller not read it in an earlier comment 🤦‍♀️. Feel free to ignore (I guess I do think the error should be logged

src/server/pps/server/pipeline_controller.go
```diff
@@ -191,19 +191,64 @@ func (pc *pipelineController) step(timestamp time.Time) (retErr error) {
...
+					return nil
+				}
+			}
+			return errors.Wrapf(err, "could not update resource for pipeline %q", pc.pipeline)
+		})
+		if errors.As(err, &stepErr) && stepErr.failPipeline {
+			failError := step.setPipelineFailure(ctx, fmt.Sprintf("could not update resources after %d attempts: %v", errCount, err))
+			if failError != nil {
+				log.Errorf("PPS master: error starting a pipelineController for pipeline '%s': %v", pc.pipeline,
+					errors.Wrapf(failError, "error failing pipeline %q (%v)", pc.pipeline, stepErr))
```
msteffen, line 234: I would similarly do here:
```
log.Errorf("PPS master: error failing pipeline '%s' after step error (%v): %v", pc.pipeline, stepErr, failErr)
errors.Wrapf(failError, "error failing pipeline %q  after step error (%v)", pc.pipeline, stepErr)
```

msteffen (commented):
Few more comments, actually :grimacing:. Sill looks really good overall!!

src/server/pps/server/pipeline_controller.go
```diff
@@ -191,19 +191,64 @@ func (pc *pipelineController) step(timestamp time.Time) (retErr error) {
...
+	} else if step == nil {
+		// interpret the event as a delete operation
+		if err := pc.deletePipelineResources(); err != nil {
+			log.Errorf("PPS master: error deleting pipelineController resources for pipeline '%s': %v", pc.pipeline,
+				errors.Wrapf(err, "failing pipeline %q", pc.pipeline))
+			return true, err
+		}
+		return true, nil
+	} else {
+		var stepErr stepError
```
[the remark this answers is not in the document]
PFedak: this is correct, just needed for unwrapping in the `errors.As`. I guess it could be inline in the ifs 
`if stepErr := new(stepError); errors.As(err, &stepErr)`
or just have two declared closer to where it's used, but the interface for unwrapping is just a little annoying

src/server/pps/server/pipeline_controller.go
```diff
@@ -191,19 +191,64 @@ func (pc *pipelineController) step(timestamp time.Time) (retErr error) {
 		tracing.FinishAnySpan(span, "err", retErr)
 	}()
 
+	if step, ctx, err := pc.newStep(); err != nil {
```
PFedak, line 194: since both of the first two branches return, I'd slightly prefer having the last big block indented one level less

PFedak (approved):
LGTM, just nits

src/server/pps/server/pipeline_controller.go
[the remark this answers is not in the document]
acohen4: great nit 👍 also love the idiom "belt-and-suspenders" lol

src/server/pps/server/pipeline_controller.go
```diff
@@ -176,7 +174,9 @@ func (pc *pipelineController) tryFinish() {
 // 1. retrieves its full pipeline spec and RC into the 'Details' field
 // 2. makes whatever changes are needed to bring the RC in line with the (new) spec
 // 3. updates 'pipelineInfo', if needed, to reflect the action it just took
-func (pc *pipelineController) step(timestamp time.Time) (retErr error) {
+//
+// returns true if the pipeline is deleted, and the pipelineController can try to shutdown
+func (pc *pipelineController) step(timestamp time.Time) (isDelete bool, retErr error) {
 	log.Debugf("PPS master: processing event for %q", pc.pipeline)
 
 	// Handle tracing
```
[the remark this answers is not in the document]
msteffen: I'm not sure what was going on when I loaded this (the indented code was inside one of the GitHub code folds, so that might have had something to do with it), but the code seems totally fine now, so I'm marking this resolved.

src/server/pps/server/pipeline_controller.go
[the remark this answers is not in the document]
msteffen: Got it, now I see how this works—thanks!!

PR #7070, opened 2021-11-05
Introduce helm global.disableSecurityContexts

acohen4:


etc/helm/pachyderm/values.yaml
```diff
@@ -14,7 +14,7 @@ global:
     # If blank, a value will be generated by the postgres subchart
```
chainlink, line 14: Looks like an auto formatter got run on this file?

chainlink (changes requested):
Couple small comments. What's the context behind introducing a global here?

etc/helm/pachyderm/values.yaml
acohen4: autoformatter is my arch nemesis 😡

acohen4:
> Couple small comments. What's the context behind introducing a global here?

Manually setting a bunch of the secuirty contexts to false is cumbersome for OpenShift deployments. Right now you would:
`pachd.securityContext.enabled: false

etcd.securityContext.enabled: false

postgresql.securityContext.enabled: false

postgresql.containerSecurityContext.enabled: false`

slack thread: https://pachyderm.slack.com/archives/C01SMKHC71B/p1636035529059000

tybritten (approved):

chainlink:
So, looking a little deeper here, I have some concerns:
- Ideally, we shouldn't modify the embedded postgres chart at all, which makes upgrading to new chart versions easier 
- The bitnami stuff is specifically designed to work with Openshift, and by disabling the security contexts, we're doing an end run around alot of the work they've done to make it work https://github.com/bitnami/charts/blob/master/bitnami/postgresql/templates/statefulset.yaml#L91-L118
- As has been mentioned in prior PRs, the security context stuff was added specifically to work with openshift, so if that's not working, then we should modify it to be compatible.

acohen4:
> So, looking a little deeper here, I have some concerns:
> 
> * Ideally, we shouldn't modify the embedded postgres chart at all, which makes upgrading to new chart versions easier
> * The bitnami stuff is specifically designed to work with Openshift, and by disabling the security contexts, we're doing an end run around alot of the work they've done to make it work https://github.com/bitnami/charts/blob/master/bitnami/postgresql/templates/statefulset.yaml#L91-L118
> * As has been mentioned in prior PRs, the security context stuff was added specifically to work with openshift, so if that's not working, then we should modify it to be compatible.

I didn't realize that chart was worked on by bitnami. You make good points. @tybritten what do you think?

chainlink (approved):

PR #7077, opened 2021-11-08
Improve tracker GC query performance.

PFedak:
In some database engines, NOT IN prevents certain optimizations (in
particular, treating the query as an anti-join) even when all columns
involved are marked NOT NULL. Switching this to an equivalent NOT EXISTS
query resulted in a dramatic speedup when tested on a customer cluster.

Additionally, to avoid needlessly holding a query open, add a limit to
the deletable object query and simply iterate over the results in a
slice. We already run the query multiple times until completion, so no
other code needs to change.


brendoncarroll (approved):
This LGTM.  At some point we may want to change the interface to just return a slice, and pass in the limit as an argument.

PR #7079, opened 2021-11-08
Adding ingress.uriHttpsProtoOverride helm value

acohen4:


chainlink:
What about `upstreamTLS` for a name of the flag?

acohen4:
> upstreamTLS

is the TLS being handled upstream the right description? I probably don't have as good of an understanding here

tybritten:
i think the name makes sense as is. basically, we're setting the various URIs that are derived from the ingress address and most of the time, we'll set the protocol corrected based on `ingress.tls.enabled` but there are some ingress exceptions where tls termination is handled upstream.

tybritten (approved):

chainlink:
Overall, I think this flag is just going to cause confusion. I'm also concerned about adding another flag to support a non standard edge case. I think if we want to support ingress in openshift, we should do it right and add a ingressroute object to the helm chart (presumably you need to create one manually outside the helmchart anyway for pach itself?)

tybritten:
sure we could do that. My thought was it might not only be a openshift use case? Any environment (like ALB in AWS, etc) where SSL can be terminated off-cluster before the actual ingress object would need to be able to do this. maybe there's a cleaner way to do it?

nadegepepin (approved):
LGTM

PR #6817, opened 2021-09-07
add max-txn-ops, nodeselector and toleration

GuiGav:
Following this issue https://github.com/pachyderm/helmchart/issues/211
-    add nodeselector and toleration
-    fix bug in etcd command if dynamicNodes > 1
-    add max-txn-ops variable for etcd

PR #7082, opened 2021-11-09
Adding nodeSelectors to helm contributed in #6817

acohen4:


chainlink (approved):
Looks good, fyi I'm pretty sure scaling etcd replicas past 1 is broken (and we typically don't use with pach because of raft perf concerns) As such, we might want to omit that flag for now

GuiGav:
Hi @chainlink , this is suppose to fix the etcd problem (at least it does on 1.13). But indeed I decided to come back to 1 replica as scaling created more problems than it solved because of the raft perf you mentionned

chainlink (approved):

PR #7083, opened 2021-11-10
script for deploying a production like setup to gcp

seslattery:
Definitely open to suggestions on a better spot for this to live.  But ideally this is for documentation purposes, we can link to this script at the start of the gcp prod docs as another resource.


nadegepepin (approved):
LGTM!

PR #7085, opened 2021-11-10
Add special error message for inspect job/commit.

### LINKED ISSUE #7065: pachctl inspect job workflow is unintuitive
**What happened?**:
I did the tutorial on Pachyderm 2, had a job run then did a `pachctl list job` followed by a `pachctl inspect job <jobid>`.

This returned a `invalid format "<jobid>": expected pipeline@job-id`

The problem is that doing a `pachctl list job` shows no info about which pipelines were involved in the job. For a simple scenario such as the tutorial, it is trivial to guess that the pipeline is edges, but for a complex production scenario with multiple pipelines involved with the job, this leaves the user guessing.

I would expect a `pachctl inspect job <jobid>` to return some aggregated information about all the pipelines were involved in the job. 

Alternatively, doing a `pachctl list job` should specify which pipelines were involved with the job in a new column and maintain the `pipeline@job-id` format.

As described above.

PFedak:
It seems reasonable to expect "inspect job <ID>" to do what "list job
<ID>" does, so if we detect that usage we should nudge users towards
the intended command.

Fix #7065

### DIFF
--- a/src/server/pfs/cmds/cmds.go
+++ b/src/server/pfs/cmds/cmds.go
@@ -367,7 +367,7 @@ $ {{alias}} test@fork -p XXX`,
 		Run: cmdutil.RunFixedArgs(1, func(args []string) error {
 			commit, err := cmdutil.ParseCommit(args[0])
 			if err != nil && uuid.IsUUIDWithoutDashes(args[0]) {
-				return errors.New(`Use "list commit <id>" to see commits with a given ID across different repos `)
+				return errors.New(`Use "list commit <id>" to see commits with a given ID across different repos`)
 			} else {
 				return err
 			}
--- a/src/server/pfs/cmds/cmds.go
+++ b/src/server/pfs/cmds/cmds.go
@@ -366,7 +366,9 @@ $ {{alias}} test@fork -p XXX`,
 		Long:  "Return info about a commit.",
 		Run: cmdutil.RunFixedArgs(1, func(args []string) error {
 			commit, err := cmdutil.ParseCommit(args[0])
-			if err != nil {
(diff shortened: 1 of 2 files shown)


nadegepepin (approved):
LGTM

src/server/pfs/cmds/cmds.go
```diff
@@ -366,7 +366,9 @@ $ {{alias}} test@fork -p XXX`,
 		Long:  "Return info about a commit.",
 		Run: cmdutil.RunFixedArgs(1, func(args []string) error {
 			commit, err := cmdutil.ParseCommit(args[0])
-			if err != nil {
+			if err != nil && uuid.IsUUIDWithoutDashes(args[0]) {
+				return errors.New(`Use "list commit <id>" to see commits with a given ID across different repos `)
```
Tryneus, line 370: nit:
```suggestion
				return errors.New(`Use "list commit <id>" to see commits with a given ID across different repos`)
```

src/server/pps/cmds/cmds.go
```diff
@@ -76,7 +76,9 @@ If the job fails, the output commit will not be populated with data.`,
 		Long:  "Return info about a job.",
 		Run: cmdutil.RunFixedArgs(1, func(args []string) error {
 			job, err := cmdutil.ParseJob(args[0])
-			if err != nil {
+			if err != nil && uuid.IsUUIDWithoutDashes(args[0]) {
+				return errors.New(`Use "list job <id>" to see jobs with a given ID across different pipelines `)
```
Tryneus, line 80:
```suggestion
				return errors.New(`Use "list job <id>" to see jobs with a given ID across different pipelines`)
```

Tryneus (approved):
LGTM

PR #7087, opened 2021-11-10
Updating Etcd and k8s client versions, and simplifying go module

acohen4:
Most signficantly, this change includes module upgrades to the k8s + etcd clients, as well as using the corresponding etcd image version. 

To address concerns around upgrading the etcd statefulset:

- Based on the docs, upgrading from 3.3 -> 3.4 -> 3.5 can be done via rolling upgrade assuming we aren’t using deprecated flags/APIs which we’re not.
https://etcd.io/docs/v3.5/upgrades/upgrade_3_4/ , https://etcd.io/docs/v3.5/upgrades/upgrade_3_5/
- I also simply validated this with a simple pipeline locally - This does bring up that upgrade testing is probably worthwhile to think about more generally


chainlink:
Oh man, sorry you had to wade in here as well. I've got PRs for most of this stuff too :/ 
- https://github.com/pachyderm/pachyderm/pull/6804/files
- https://github.com/pachyderm/pachyderm/pull/6803
- (pach v1) https://github.com/pachyderm/pachyderm/pull/6024/files

My thinking on the prometheus stuff is to vendor it into pach since it's literally one RPC (and then we eliminate the dependency entirely)

I proposed upgrading etcd before we shipped 2.0, but the concern there was around known perf of 3.3 vs the unknown of 3.5. The consensus at the time is that we need to perf test 3.5 before going there. 3.5 is emphatically better in many ways (including full module support) so I think we should definitely do this, but probably in a separate PR. 

The grpc stuff is also a long story. We're stuck on an old version because of gogoproto (and etcd) The linter is technically right, `google.golang.org/protobuf/proto` is the new library (and the old one has been deprecated) I found a couple of libraries that let us replicate the naming conventions from gogo, and a couple of other features we use, so we can indeed get off gogo. That combined with the etcd upgrade should get us off the old protos.

acohen4:
> Oh man, sorry you had to wade in here as well. I've got PRs for most of this stuff too :/
> 
> * https://github.com/pachyderm/pachyderm/pull/6804/files
> * [Initial commit for vendoring loki client into pach2 #6803](https://github.com/pachyderm/pachyderm/pull/6803)
> * (pach v1) https://github.com/pachyderm/pachyderm/pull/6024/files
> 
> My thinking on the prometheus stuff is to vendor it into pach since it's literally one RPC (and then we eliminate the dependency entirely)
> 
> I proposed upgrading etcd before we shipped 2.0, but the concern there was around known perf of 3.3 vs the unknown of 3.5. The consensus at the time is that we need to perf test 3.5 before going there. 3.5 is emphatically better in many ways (including full module support) so I think we should definitely do this, but probably in a separate PR.
> 
> The grpc stuff is also a long story. We're stuck on an old version because of gogoproto (and etcd) The linter is technically right, `google.golang.org/protobuf/proto` is the new library (and the old one has been deprecated) I found a couple of libraries that let us replicate the naming conventions from gogo, and a couple of other features we use, so we can indeed get off gogo. That combined with the etcd upgrade should get us off the old protos.

That's great stuff! Based on these changes, it actually looks like the grpc version is compatible with gogo proto which is good! I think we should re-surface the etcd-client upgrade conversation now that we hit 2.0, and have load testing capabilities to test changes including the etcd upgrade in preparation for 2.1

go.mod
```diff
@@ -5,60 +5,58 @@ go 1.17
 require (
 	cloud.google.com/go v0.84.0
 	cloud.google.com/go/storage v1.10.0
-	github.com/Azure/azure-sdk-for-go v36.1.0+incompatible
+	github.com/Azure/azure-sdk-for-go v54.0.0+incompatible
```
brendoncarroll, line 8: lol, this is what real semver looks like.

go.mod
```diff
@@ -70,88 +68,94 @@ require (
...
-	k8s.io/apimachinery v0.17.4
-	k8s.io/client-go v12.0.0+incompatible
+	gotest.tools/v3 v3.0.3 // indirect
+	k8s.io/api v0.22.3
+	k8s.io/apimachinery v0.22.3
+	k8s.io/client-go v0.22.3
 	modernc.org/mathutil v1.0.0
 )
 
 require (
```
brendoncarroll, line 104: Is there a reason these are separate?  `go mod tidy` should give you 2 stanzas. One for direct, and one for indirect.  If you remove all the `// indirect` comments (just the comments, not the entries) from the top stanza, it should remove or keep those entries for you.

chainlink:
As mentioned in slack, we need to handle the etcd upgrade case here (ie. when you run a 2.0.1 to 2.1 pach upgrade, does etcd just restart and is happy, or is there more to it. We should also look to see if any CLI options have changed or been added/removed in 3.5)

chainlink:
Also, helm changes like the etcd version bump aren't fully captured in CI, and the chart is vendored into Hub

src/internal/config/config.go
```diff
@@ -126,7 +126,10 @@ func validateCachedConfig() (bool, error) {
 	if cachedConfig.UserID == "" {
 		updated = true
 		log.Debugln("No UserID present in config - generating new one.")
-		uuid := uuid.NewV4()
+		uuid, err := uuid.NewV4()
```
brendoncarroll, line 129: We have a uuid package internally, which doesn't return an error.  Try `uuid.NewWithDashes()`

src/internal/dockertestenv/util.go
```diff
@@ -94,7 +94,7 @@ func ensureContainer(ctx context.Context, dclient docker.APIClient, containerNam
 	hostConfig := &container.HostConfig{
 		PortBindings: portBindings,
 	}
-	resp, err := dclient.ContainerCreate(ctx, containerConfig, hostConfig, nil, containerName)
+	resp, err := dclient.ContainerCreate(ctx, containerConfig, hostConfig, nil, nil, containerName)
```
brendoncarroll, line 97: I assume not setting the platform requests the host OS and Architecture.
https://pkg.go.dev/github.com/docker/docker/client#Client.ContainerCreate

If the INTERNAL tests are passing then it's probably fine.

src/internal/pool/pool.go
```diff
@@ -40,7 +40,7 @@ type Pool struct {
 func NewPool(kubeClient *kube.Clientset, namespace string, serviceName string, port int, queueSize int64, opts ...grpc.DialOption) (*Pool, error) {
 	endpointsInterface := kubeClient.CoreV1().Endpoints(namespace)
 
-	watch, err := endpointsInterface.Watch(metav1.ListOptions{
+	watch, err := endpointsInterface.Watch(context.TODO(), metav1.ListOptions{
```
brendoncarroll, line 43: I think this could be a `context.Background` since the watch is cancelled when the connection is closed below.  But TODO is fine for now.

chainlink:
I think our protos are still being compiled using the old library? https://github.com/pachyderm/pachyderm/blob/master/etc/proto/Dockerfile#L15

src/server/pps/server/s3g_sidecar.go
```diff
@@ -259,14 +259,15 @@ func (s *k8sServiceCreatingJobHandler) OnCreate(ctx context.Context, jobInfo *pp
 	}
 }
 
-func (s *k8sServiceCreatingJobHandler) OnTerminate(_ context.Context, job *pps.Job) {
+func (s *k8sServiceCreatingJobHandler) OnTerminate(ctx context.Context, job *pps.Job) {
```
brendoncarroll, line 262: :+1:

src/server/pps/server/pipeline_controller.go
```diff
@@ -787,28 +788,28 @@ func (pc *pipelineController) deletePipelineKubeResources() (retErr error) {
 
 	// Delete any services associated with pc.pipeline
 	selector := fmt.Sprintf("%s=%s", pipelineNameLabel, pc.pipeline)
-	opts := &metav1.DeleteOptions{
+	opts := metav1.DeleteOptions{
 		OrphanDependents: &falseVal,
 	}
-	services, err := kubeClient.CoreV1().Services(namespace).List(metav1.ListOptions{LabelSelector: selector})
+	services, err := kubeClient.CoreV1().Services(namespace).List(pc.ctx, metav1.ListOptions{LabelSelector: selector})
```
brendoncarroll, line 794: We probably don't want to just take the stored context here.  It would be nice to actually have control over how long `deletePipelineKubeResources` can take before it times out.  That might end up being the stored context, at a level up.

chainlink:
A couple more lingering refs to the old protobuf lib
https://github.com/pachyderm/pachyderm/blob/master/src/server/pfs/server/driver_file.go#L9
https://github.com/pachyderm/pachyderm/blob/master/src/server/pps/server/api_server.go#L18

src/internal/uuid/uuid.go
```diff
@@ -14,7 +14,10 @@ import (
 func New() string {
 	var result string
 	backoff.RetryNotify(func() error {
```
brendoncarroll, line 16: I'm not sure why we backoff here.

src/internal/watch/watch.go
```diff
@@ -141,6 +141,14 @@ func NewEtcdWatcher(ctx context.Context, client *etcd.Client, trimPrefix, prefix
 				// use new "nextRevision"
 				internalWatcher = etcd.NewWatcher(client)
 				rch = internalWatcher.Watch(ctx, prefix, watchOptions(nextRevision)...)
+				// We interpret a response with nil Error and 0 events as a server side cancellation
+				// as described in https://github.com/etcd-io/etcd/blob/client/v3.5.1/client/v3/watch.go#L53
+				//
+				// "If the context "ctx" is canceled or timed out, returned "WatchChan" is closed,
+				// and "WatchResponse" from this closed channel has zero events and nil "Err()"."
+				if resp.Err() == nil && len(resp.Events) == 0 {
```
brendoncarroll, line 149: Isn't the `ok` variable above more direct evidence the channel has been closed?

src/server/cmd/pachctl/cmd/cmd_test.go
```diff
@@ -94,8 +94,11 @@ func testConfig(t *testing.T, pachdAddressStr string) *os.File {
 	pachdAddress, err := grpcutil.ParsePachdAddress(pachdAddressStr)
 	require.NoError(t, err)
 
+	uid, err := uuid.NewV4()
```
brendoncarroll, line 97: Use our uuid library here.

src/server/pfs/server/driver_file.go
```diff
@@ -6,7 +6,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/golang/protobuf/proto"
+	"github.com/gogo/protobuf/proto"
```
brendoncarroll, line 9: Was this intentional.  Weren't we trying to move away from gogo because it's unmaintained?

src/server/pps/server/api_server.go
```diff
@@ -3076,7 +3071,7 @@ func RepoNameToEnvString(repoName string) string {
 }
 
 func (a *apiServer) rcPods(rcName string) ([]v1.Pod, error) {
```
brendoncarroll, line 3073: This is another good place to add a context argument.

src/server/pps/server/api_server.go
```diff
@@ -269,7 +269,7 @@ func validateTransform(transform *pps.Transform) error {
 func (a *apiServer) validateKube() {
```
brendoncarroll, line 269: This could take a context argument.

brendoncarroll (commented):
This is a great change.

@chainlink brings up a good point about upgrading the etcd cluster itself.  Is the client backwards compatible?  That would be something we can pull out.

Some questions throughout.  There are a few places where we have a stored context (which is not a great pattern) and I think we should avoid using it except in the main control loop for the component that's storing it.  Contexts should almost always just be passed down the call tree.

chainlink:
Overall super glad to be getting off these ancient libs!

acohen4:
@brendoncarroll @chainlink thank you both for the thorough comments. I'll respond to them all, but first off regarding the etcd image upgrade concerns:

- Based on the docs, upgrading from 3.3 -> 3.4 -> 3.5 can be done via rolling upgrade assuming we aren’t using deprecated flags/APIs which we’re not.
https://etcd.io/docs/v3.5/upgrades/upgrade_3_4/ , https://etcd.io/docs/v3.5/upgrades/upgrade_3_5/
- I also simply validated this with a simple pipeline locally - This does bring up that upgrade testing is probably worthwhile to think about more generally 

@chainlink do you know where else in CI I need to update the etcd version?

chainlink:
@acohen4 the hub tests (and load test if they run in CI?) It's a little tricky, but you need to ship the new helmchart to hub staging and you can test there. https://github.com/pachyderm/hub/blob/master/deployment/operations/install_pachd.go#L26

src/internal/uuid/uuid.go
acohen4: me neither

src/server/pfs/server/driver_file.go
acohen4: this was intentional. Right now, we are 99% using gogoproto. When we migrate, we'll want to do so to the updated google backed proto lib which is incompatible with the previous. I removed this because that module ("github.com/golang/protobuf/proto") is deprecated

src/internal/watch/watch.go
acohen4: it is, but the way the updated etcd client signals that the ctx was cancelled on the server-side is by including `Err()==nil  && len(resp.Events) == 0`

src/internal/dockertestenv/util.go
acohen4: I wasn't really sure here, but yeah things seem to be okay? I can also dig in a bit more

src/server/pps/server/pipeline_controller.go
acohen4: yeah I agree with this. I'm actually making some changes to the pipeline controller separately and would like to handle this in a separate PR if that seems alright

chainlink:
Hrm, from the etcd docs:

```
To upgrade an existing etcd deployment to 3.5, the running cluster must be 3.4 or greater. If it’s before 3.4, please upgrade to 3.4 before upgrading to 3.5.
```

https://etcd.io/docs/v3.5/upgrades/upgrade_3_5/#upgrade-requirements

etc/proto/Dockerfile
```diff
@@ -12,8 +12,6 @@ RUN \
 RUN unzip protoc.zip -d /
 RUN cp -r /include /bin
 
-RUN go get -u -v github.com/golang/protobuf/proto
```
acohen4, line 15: this doesn't seem to be doing anything since we're using gogoproto in our protoc process

brendoncarroll (approved):
LGTM

PR #7088, opened 2021-11-11
First pass at supporting reading from open commits in s3g

lukemarsden:
S3G currently errors when trying to list or read objects in open commits:

```
2021-11-11 05:52:37,366 - MainThread - botocore.parsers - DEBUG - Response body:
b'<?xml version="1.0" encoding="UTF-8"?>\n<Error><Code>InternalError</Code><Message>timestamp: nil Timestamp</Message><Resource>/master.hello</Resource><RequestId>9afa06d0-853e-49d8-b7
```

I hypothesize this is due to files in open commits not yet having a `Committed` time recorded in Pach, so we hit the error case here: https://github.com/pachyderm/pachyderm/blob/master/src/server/pfs/s3/bucket.go#L20-L23

In protobuf module source we can see the same error message we're getting on the client in the function `validateTimestamp` that `TimestampFromProto` calls:
```
func validateTimestamp(ts *Timestamp) error {
	if ts == nil {
		return errors.New("timestamp: nil Timestamp")
	}
```

This PR makes S3G just return 1970-01-01 timestamps for those files, because listing objects with the wrong modified time is better than failing to list them at all.

A less wrong answer might be to list a more recent known time (lower bound) on the modification time, but doing that well requires more insight than I have at present. (e.g. can we return the correct commit time for files that were last modified in earlier commits, and return the timestamp of the start of the open commit for files that have been modified in the open commit?)

### DIFF
--- a/src/server/pfs/s3/bucket.go
+++ b/src/server/pfs/s3/bucket.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/gogo/protobuf/types"
 	glob "github.com/pachyderm/ohmyglob"
 	"github.com/pachyderm/pachyderm/v2/src/internal/ancestry"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errutil"
@@ -17,7 +16,7 @@ import (
 )
 
 func newContents(fileInfo *pfsClient.FileInfo) (s2.Contents, error) {
-	t, err := types.TimestampFromProto(fileInfo.Committed)
+	t, err := permissiveTimestampFromProto(fileInfo)
 	if err != nil {
 		return s2.Contents{}, err
 	}
--- a/src/server/pfs/s3/object.go
+++ b/src/server/pfs/s3/object.go
@@ -5,12 +5,33 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/gogo/protobuf/types"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errutil"
+	pfsClient "github.com/pachyderm/pachyderm/v2/src/pfs"
 	pfsServer "github.com/pachyderm/pachyderm/v2/src/server/pfs"
 	"github.com/pachyderm/s2"
 )
+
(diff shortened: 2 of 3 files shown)


lukemarsden:
This works with S3G + rclone mount 🎉

src/server/pfs/s3/bucket.go
```diff
@@ -17,7 +16,7 @@ import (
 )
 
 func newContents(fileInfo *pfsClient.FileInfo) (s2.Contents, error) {
-	t, err := types.TimestampFromProto(fileInfo.Committed)
+	t, err := permissiveTimestampFromProto(fileInfo)
```
PFedak, line 19: I have a slight style preference for just inlining this function. In fact, looking at the code of `TimestampFromProto`, I wonder if we should just be ignoring these errors. It explicitly treats a nil timestamp as the zero time, and the only other errors are from out-of-range dates which shouldn't be an issue since we are the ones generating the timestamps.

src/server/pfs/s3/object.go
```diff
@@ -5,13 +5,34 @@ import (
...
 	pfsServer "github.com/pachyderm/pachyderm/v2/src/server/pfs"
 	"github.com/pachyderm/s2"
 )
 
+func permissiveTimestampFromProto(fileInfo *pfsClient.FileInfo) (time.Time, error) {
+	// fileInfo.Committed can be nil in the case of reading from an open
+	// commit. In that case, it's better to list the file with no known
+	// timestamp than to return an error.
+
+	// TODO: it would be less wrong to use the commit start time than to return
```
PFedak, line 22: I think this is too late to do anything about this, though, it would need to be a change in the iteration in pfs/server/source.go  
It doesn't look like we currently use that field to indicate whether the source commit is finished, so it would probably be okay to use the start time there

PFedak (changes requested):
Making the modified time more accurate based on commit history is interesting but it seems like it would be an expensive operation with what we currently expose

msteffen:
Closing PR, as it's obsolete. I believe this was superceded by https://github.com/pachyderm/pachyderm/pull/8129

PR #7089, opened 2021-11-12
Add confirm prompt to port-forward.

PFedak:
If the port forwarder doesn't clean up the context when it exits,
subsequent calls will error out, requiring a user to go in and delete
lines from the config file. For a quick fix, just add a confirm that
will start forwarding anyway if asked.


Tryneus (approved):
LGTM

PR #7091, opened 2021-11-12
[DOC-92] [ci-skip] Added new reducer pipeline and tagged new 2.0.0 image

nadegepepin:


nadegepepin:
I decided against including a reducer (3rd pipeline) that would aggregate the content of all files in each directory (in the processor output repo) and create one log file per directory. The reasoning behind this decision is that it adds complexity while it is not directly related to our first goal: provide a simple explanation of how to use a spout pipeline.

PR #7092, opened 2021-11-12
Stop pachctl mount panicking on unmount and 200x performance 😉 

lukemarsden:
Looks like the go-fuse lib changed its behavior from the consumer having to call Serve() to Mount() calling Serve() automatically, this change switches the extra Serve() for a Wait() so that we just wait on the lib's call. This fixes the following panic (and various flavors like it):

```
panic: sync: negative WaitGroup counter

goroutine 835 [running]:
sync.(*WaitGroup).Add(0xc000db80b8, 0xc0016c01b0)
        /snap/go/8627/src/sync/waitgroup.go:74 +0x105
sync.(*WaitGroup).Done(0x0)
        /snap/go/8627/src/sync/waitgroup.go:99 +0x25
github.com/hanwen/go-fuse/v2/fuse.(*Server).loop(0xc000db8000, 0x0)
        /home/luke/gocode/pkg/mod/github.com/hanwen/go-fuse/v2@v2.0.3/fuse/server.go:448 +0x194
github.com/hanwen/go-fuse/v2/fuse.(*Server).Serve(0xc000db8000)
        /home/luke/gocode/pkg/mod/github.com/hanwen/go-fuse/v2@v2.0.3/fuse/server.go:367 +0x35
created by github.com/hanwen/go-fuse/v2/fs.Mount
        /home/luke/gocode/pkg/mod/github.com/hanwen/go-fuse/v2@v2.0.3/fs/mount.go:32 +0x150
FAIL    github.com/pachyderm/pachyderm/v2/src/server/pfs/fuse   12.416s
```

Also, upgrade go-fuse to the latest version (although the previous version we were using had the same problem).

FInally, disable progress bars as they cause throughput to drop from 200MB/sec to 1MB/sec 😮

### DIFF
--- a/src/server/pfs/fuse/fuse.go
+++ b/src/server/pfs/fuse/fuse.go
@@ -1,6 +1,7 @@
 package fuse
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/signal"
@@ -73,6 +74,11 @@ func Mount(c *client.APIClient, target string, opts *Options) (retErr error) {
 			}
 		}
 	}()
+	fmt.Println("Uploading files to Pachyderm...")
+	// Rendering progress bars for thousands of files significantly slows down
+	// throughput. Disabling progress bars takes throughput from 1MB/sec to
+	// 200MB/sec on my system, when uploading 18K small files.
+	progress.Disable()
 	for path, state := range root.files {
 		if state != dirty {
 			continue
@@ -100,5 +106,6 @@ func Mount(c *client.APIClient, target string, opts *Options) (retErr error) {
 			return err
 		}
 	}
+	fmt.Println("Done!")
 	return nil
 }
--- a/src/server/pfs/fuse/fuse.go
+++ b/src/server/pfs/fuse/fuse.go
@@ -53,7 +53,7 @@ func Mount(c *client.APIClient, target string, opts *Options) (retErr error) {
 		}
 		server.Unmount()
 	}()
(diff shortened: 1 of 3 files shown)


lukemarsden:
Note: the FUSE tests don't actually seem to be running on master. They were failing.

src/server/pfs/fuse/fuse.go
```diff
@@ -73,6 +74,11 @@ func Mount(c *client.APIClient, target string, opts *Options) (retErr error) {
 			}
 		}
 	}()
+	fmt.Println("Uploading files to Pachyderm...")
+	// Rendering progress bars for thousands of files significantly slows down
+	// throughput. Disabling progress bars takes throughput from 1MB/sec to
+	// 200MB/sec on my system, when uploading 18K small files.
+	progress.Disable()
```
albscui, line 81: Instead of hard disabling, maybe control it via options?

msteffen (approved):
LGTM—what a find!

src/server/pfs/fuse/fuse.go
lukemarsden: I don't think we should ever give people the option to turn on something which destroys performance so badly 😄 especially not with `pachctl mount` which - as it's an interactive mount - is potentially more likely to deal with many small files.

lukemarsden:
The failing tests are in PPS and so unrelated to this change, merging.

PR #7094, opened 2021-11-12
Add `pachctl delete commit` command

Tryneus:
Between 1.x and 2.x, `PFS.DeleteCommit` was split into two operations:

* `SquashCommitSet` removes a commit-set which has children, but the children will still contain the data from that commit.
* `DropCommitSet` removes a commit-set that has no children

This change was made at the API level, but `pachctl` was not updated accordingly, so only `SquashCommitSet` is available on the command-line.  This adds back the `pachctl delete commit` command that will use the `PFS.DropCommitSet` RPC.  I did not name it `pachctl drop commit` to avoid adding a new verb to the CLI, but could be convinced otherwise.


src/server/pfs/cmds/cmds.go
```diff
@@ -771,6 +771,27 @@ The squash will fail if it includes a commit with no children`,
 	shell.RegisterCompletionFunc(squashCommit, shell.BranchCompletion)
 	commands = append(commands, cmdutil.CreateAlias(squashCommit, "squash commit"))
 
+	deleteCommit := &cobra.Command{
+		Use:   "{{alias}} <commit-id>",
+		Short: "Delete the sub-commits of a commit.",
+		Long: `Delete the sub-commits of a commit.  The data in the sub-commits will be lost.
+This operation is only supported if the commit has no children.`,
```
PFedak, line 778: I guess more precisely

```suggestion
This operation is only supported if none of the sub-commits have children.`,
```

src/server/pfs/cmds/cmds.go
```diff
@@ -771,6 +771,27 @@ The squash will fail if it includes a commit with no children`,
 	shell.RegisterCompletionFunc(squashCommit, shell.BranchCompletion)
 	commands = append(commands, cmdutil.CreateAlias(squashCommit, "squash commit"))
 
+	deleteCommit := &cobra.Command{
+		Use:   "{{alias}} <commit-id>",
+		Short: "Delete the sub-commits of a commit.",
```
PFedak, line 776: feels a little weird that only squash and delete use "sub-commit" now but I don't really have a suggestion

PFedak (approved):
LGTM broadly - did we discuss calling the command `drop commit` at some point? I guess there isn't as much of a need to distance it from the old command anymore. If we end up adding real delete support, are you anticipating we would add it to this command? I could imagine this being separate but maybe that's just too confusing.

PR #7101, opened 2021-11-15
Increase wait time for pipeline crashing recovery test.

PFedak:
The crashing monitor interval is 15 seconds, so the previous limits
might only allow 15 seconds in some cases. I'm not sure what we should
really expect, but increasing it a bit more seems safe.


msteffen (approved):
LGTM

PR #7107, opened 2021-11-17
Disambiguate some commit and job logs.

PFedak:
A few logs and error messages did not show the branch that commits were
made on, which could be ambiguous and isn't consistent with how commits
are displayed elsewhere.


Tryneus (approved):
Good catch, LGTM pending CI

PR #7108, opened 2021-11-17
New Entrypoint `pachtf` and `transforms` package

brendoncarroll:
Adds a package transforms for pipeline transforms that we intend to provide batteries included.
Adds a new entrypoint pachtf which will be the program that calls into the transforms library package.

`transforms.SQLIngest` reads queries from its input directory and writes results to the output directory.

### DIFF
--- a/src/internal/secrets/secrets.go
+++ b/src/internal/secrets/secrets.go
@@ -1,0 +1,11 @@
+package secrets
+
+import "fmt"
+
+// Secret is the type of secret data.
+// It prevents the actual contents of the secret from being logged
+type Secret string
+
+func (s Secret) String() string {
+	return fmt.Sprintf("Secret{len=%d}", len(s))
+}
--- a/src/internal/transforms/sqlingest.go
+++ b/src/internal/transforms/sqlingest.go
@@ -25,7 +25,8 @@ type SQLIngestParams struct {
 	Format   string
 }
 
-// SQLIngest connects to a SQL database at params.URL and runs params.Query.
+// SQLIngest connects to a SQL database at params.URL and runs queries
+// read from files in the input.
 // The resulting rows are written to files in params.OutputDir.
 // The format of the output file is controlled by params.Format.
 // Valid options are "json" and "csv"
--- a/src/internal/transforms/sqlingest.go
+++ b/src/internal/transforms/sqlingest.go
@@ -2,10 +2,7 @@ package transforms
 
 import (
 	"context"
-	"fmt"
 	"io"
-	"os"
-	"path/filepath"
 	"time"
 
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
@@ -15,6 +12,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// SQLIngestParams are the parameters passed to SQLIngest
 type SQLIngestParams struct {
 	// Instrumentation
 	Logger *logrus.Logger
@@ -24,10 +22,7 @@ type SQLIngestParams struct {
 
 	URL      pachsql.URL
 	Password secrets.Secret
-	Query    string
 	Format   string
-	// Shard affects the number appended to the file path.
-	Shard int
 }
 
 // SQLIngest connects to a SQL database at params.URL and runs params.Query.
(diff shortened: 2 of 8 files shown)


src/internal/transforms/transforms.go
```diff
@@ -0,0 +1,61 @@
...
+		inputFile, err := os.OpenFile(inputPath, os.O_RDONLY, 0)
+		if err != nil {
+			return err
+		}
+		defer inputFile.Close()
+		outputFile, err := os.OpenFile(outputPath, os.O_WRONLY|os.O_CREATE, 0755)
+		if err != nil {
+			return err
+		}
+		defer outputFile.Close()
```
acohen4, line 55: I assume there's no harm in having this defer evaluate with an error most of the time when we first reach `outputFile.Close()` at the end of this method?

src/internal/secrets/secrets.go
```diff
@@ -0,0 +1,11 @@
+package secrets
+
+import "fmt"
+
+// Secret is the type of secret data.
+// It prevents the actual contents of the secret from being logged
+type Secret string
+
+func (s Secret) String() string {
```
acohen4, line 9: clever!

src/internal/transforms/sqlingest.go
```diff
@@ -0,0 +1,104 @@
...
+	log.Infof("DONE")
+	return nil
+}
+
+type writerFactory = func(w io.Writer, fieldNames []string) sdata.TupleWriter
+
+func makeWriterFactory(formatName string) (writerFactory, error) {
+	var factory writerFactory
+	switch formatName {
+	case "json", "jsonlines":
```
acohen4, line 92: Comment for future: I feel like it could be nice to make the format values an Enum

WDYT?

src/internal/transforms/sqlingest.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// PFS In/Out
+	InputDir, OutputDir string
+
+	URL      pachsql.URL
+	Password secrets.Secret
+	Format   string
+}
+
+// SQLIngest connects to a SQL database at params.URL and runs params.Query.
```
acohen4, line 28: maybe just fix the params.Query part of the comment?

src/internal/transforms/sqlingest.go
brendoncarroll: I was thinking that eventually we would have something like a `FormatSpec` object which would have more parameters.  Something like this

```go
type FormatSpec struct {
  *JSONSpec
  *CSVSpec
}

type CSVSpec struct {
  IncludeHeaders bool
}
```

Not really sure where that should live maybe `transforms`, but `sdata` also kind of makes sense.  For now I think we will just create these objects based on the simple strings.  I think having the CSV headers would actually just create problems for people, so there's no way to enable it.

src/internal/transforms/transforms.go
brendoncarroll: Yeah the double close is not an issue.  The `return outputFile.Close()` is what matters.

src/server/cmd/pachtf/main.go
```diff
@@ -0,0 +1,65 @@
...
+	urlStr, formatName := args[0], args[1]
+	password, ok := os.LookupEnv(passwordEnvar)
+	if !ok {
+		return errors.Errorf("must set %v", passwordEnvar)
+	}
+	u, err := pachsql.ParseURL(urlStr)
+	if err != nil {
+		return err
+	}
+	inputDir := filepath.FromSlash(pfs + "/in")
```
acohen4, line 53: What's the thought behind the "/in" directory? I'm not immediately sure why it would be better (or worse) than having query files at the top level of the input repo.

brendoncarroll: This is the path under /pfs that the queries are written to.  It can't be empty and has to match the pipeline spec.

acohen4: Do you think it would be worth making it configurable? I think having it hard coded would be problematic if there were multiple SQL queries being ingressed since there would have to be a separate input repo for each

brendoncarroll: The repo name can be mapped to `in` in the pipeline spec, using the `name` field.  I think hardcoding a value here is preferable to threading through the repo name.

acohen4: sounds good to me 👍

acohen4 (approved):
LGTM!

PR #7114, opened 2021-11-18
Make background tasks wait on service env.

PFedak:
Previously, there was no guarantee that e.g. commit finishing would wait
for the api servers to be initialized. Pachyderm operations that use
transactions rely on various circular references being set up between
servers, so they would result in a panic if performed too early.

This change just makes all transactions wait on `txnEnv.Initialize`, which we 
already call at the end of setup to create the circular references.

### DIFF
--- a/src/internal/serviceenv/service_env.go
+++ b/src/internal/serviceenv/service_env.go
@@ -63,7 +63,6 @@ type ServiceEnv interface {
 	ClusterID() string
 	Context() context.Context
 	Logger() *log.Logger
-	ServerInitDone() <-chan struct{}
 	Close() error
 }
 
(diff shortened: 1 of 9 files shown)


acohen4 (approved):
LGTM!!

brycemcanally:
Is there a way to make this such that the clients of the service environment don't have to know to wait on the initialize (i.e. encapsulate the waiting in the service environment implementatation). This doesn't seem like a good permanent solution as is, but if a proper solution is more involved then we can use this with a TODO mentioning that this needs more work.

PFedak:
Agreed on this being ugly, the circular nature of the references makes it a little strange. I'm having trouble thinking through the cleaner implementation, though, basically we want to create a bunch of structs with pointers between them, and separately they want to run their own private background tasks once the pointers are created. Maybe reordering things is easier for now, then

brycemcanally (approved):
lgtm, we should probably change the title of the PR as well.

PR #7115, opened 2021-11-18
Use NewOnUserMachine() instead of NewFromURI in go client example

acohen4:
The reason for this change is that NewOnUserMachine pulls auth information from the user's context


Tryneus (approved):
LGTM

PR #7117, opened 2021-11-19
Migrating Enterprise Service config from etcd -> postgres

acohen4:


src/internal/clusterstate/v2.0.0.go
```diff
@@ -98,5 +98,3 @@ var state_2_0_0 migrations.State = migrations.InitialState().
 	Apply("add rotation_token_expiry to identity.config table", func(ctx context.Context, env migrations.Env) error {
 		return identity.AddRotationTokenExpiryConfig(ctx, env.Tx)
 	})
-	// DO NOT MODIFY THIS STATE
```
acohen4, line 101: will revert this

src/internal/dockertestenv/etcd.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"testing"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/require"
+	etcd "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/client/v3/mock/mockserver"
+)
+
+func NewTestEtcd(t *testing.T) *etcd.Client {
```
brendoncarroll, line 11: We have a package `testetcd`.  Which provides an in memory etcd.  In this package I would expect this function to bring up a docker image.

src/internal/migrations/migrations.go
```diff
@@ -18,14 +20,18 @@ type Env struct {
 	// TODO: etcd
 	ObjectClient obj.Client
 	Tx           *pachsql.Tx
+	EtcdClient   *clientv3.Client
```
brendoncarroll, line 23: 👍

src/internal/migrations/migrations.go
```diff
@@ -18,14 +20,18 @@ type Env struct {
...
+	EtcdPrefix   string
```
brendoncarroll, line 24: I don't think this makes sense in the migrations.Env object.  There is one etcd per cluster, we should be using the whole thing right?

src/internal/migrations/migrations.go
```diff
@@ -18,14 +20,18 @@ type Env struct {
...
 }
 
 // MakeEnv returns a new Env
 // The only advantage to this contructor is you can be sure all the fields are set.
 // You can also create an Env using a struct literal.
-func MakeEnv(objC obj.Client) Env {
+func MakeEnv(objC obj.Client, senv serviceenv.ServiceEnv) Env {
```
brendoncarroll, line 30: Can you just depend on the etcd.Client here?

src/server/enterprise/server/env.go
```diff
@@ -33,3 +46,36 @@ func EnvFromServiceEnv(senv serviceenv.ServiceEnv, etcdPrefix string) Env {
...
+	return col.NewPostgresCollection(
+		"enterpriseConfig",
+		db,
+		listener,
+		&ec.EnterpriseConfig{},
+		nil,
+	)
+}
+
+func TryEnterpriseConfigPostgresMigration(ctx context.Context, migrationEnv migrations.Env) error {
```
brendoncarroll, line 60: Maybe drop the `Try` here.  If it doesn't work then pachd can't come up, so just trying isn't really an option.

brendoncarroll: Also probably worth commenting this function to explain what's going on.  Not everyone has enterprise, some people do, if they do, this migrates it to postgres, etc.

src/server/enterprise/server/api_server.go
```diff
@@ -200,16 +201,20 @@ func (a *apiServer) Activate(ctx context.Context, req *ec.ActivateRequest) (resp
...
+		if err := a.configCol.ReadWrite(txCtx.SqlTx).Put(configKey, &ec.EnterpriseConfig{
 			LicenseServer: req.LicenseServer,
 			Id:            req.Id,
 			Secret:        req.Secret,
 		}); err != nil {
 			return err
 		}
-
-		return a.enterpriseTokenCol.ReadWrite(stm).Put(enterpriseTokenKey, record)
+		if _, err := col.NewSTM(ctx, a.env.EtcdClient, func(stm col.STM) error {
```
brendoncarroll, line 212: What's going on here.  Do we still need etcd?

src/internal/dockertestenv/etcd.go
brendoncarroll: If you need this exact function, I would put it in `testetcd` and explain why it's different from the other functions in there, and when you should use it.  Otherwise take the client out of the EtcdEnv that you get from `testetcd`

src/server/enterprise/server/env.go
```diff
@@ -4,13 +4,22 @@ import (
...
 	logrus "github.com/sirupsen/logrus"
 	clientv3 "go.etcd.io/etcd/client/v3"
 )
 
 type Env struct {
+	DB       *pachsql.DB
+	Listener col.PostgresListener
+	TxnEnv   *txnenv.TransactionEnv
+
 	EtcdClient *clientv3.Client
```
brendoncarroll, line 23: Related to my above comment, why can't we get rid of etcd here?

src/server/enterprise/server/api_server.go
```diff
@@ -67,7 +68,7 @@ func NewEnterpriseServer(env Env, heartbeat bool) (ec.APIServer, error) {
 		env:                  env,
 		enterpriseTokenCache: keycache.NewCache(env.BackgroundContext, enterpriseTokenCol.ReadOnly(env.BackgroundContext), enterpriseTokenKey, defaultEnterpriseRecord),
 		enterpriseTokenCol:   enterpriseTokenCol,
-		configCol:            col.NewEtcdCollection(env.EtcdClient, env.EtcdPrefix, nil, &ec.EnterpriseConfig{}, nil, nil),
+		configCol:            EnterpriseConfigCollection(env.DB, env.Listener),
```
brendoncarroll, line 71: I like that there is a function for this now.

brendoncarroll (changes requested):
I think we need to figure out a way to test the migration codepath.  At the moment it's not exercised because we never stop at the 2.0.0 point and setup the enterprise service. We don't need the whole cluster, we just need an etcd (from `testetcd`) and a postgres (from `dockertestenv`)
1. run the 2.0.0 migrations on the postgres
2. create the legacy record in etcd.  Refactor if you have to, make sure this runs as much of the same code as the real production path.
3. Apply your migration.
4. Check that the record is in postgres.

We also need another migration to clear the state in etcd.  We can't do it in the same migration because we have to commit to postgres first, before we can delete from etcd. 
You can add that migration right after the one you have now and it should work.  Should also probably check that the record is gone at the end of the test.

src/server/enterprise/server/api_server.go
acohen4: Background: 
- LS (license service) stores the org's license.
- ES (enterprise service) heartbeats to LS to verify that the license is still valid to continue enabling enterprise
- Here we are moving ES's copy of LS communication info `enterpriseConfig` over to PG
- ES keeps a secondary copy of the license retrieved in the Heartbeat stored in etcd collection called `enterpriseToken`. It's written to infrequently but pulled via `keycache.Cache`. It's valuable to cache it because it's technically accessed at each auth.WhoAmI call.

So whether we need etcd is a good question. My inclination is to leave it as is since it's supposedly more performant - though I don't know for certain it's worth the extra bit of complexity. The enterpriseToken collection also happens to be ephemeral. 

WDYT?

acohen4:
> I think we need to figure out a way to test the migration codepath. At the moment it's not exercised because we never stop at the 2.0.0 point and setup the enterprise service. We don't need the whole cluster, we just need an etcd (from `testetcd`) and a postgres (from `dockertestenv`)

So this is what I was hoping would be accomplished by the upgrade test. Right now because master doesn't have the migration, the cluster first builds, activates enterprise/auth, does some basic pipeline creation. Then post-upgrade, the migrations gets applied and we check everything still works. I recognize one downside is that this test execution is not codified and rerun. Do you think that's what we should aim for?

Also good catch that etcd isn't being cleaned up right now during the migration. I think adding another migration there makes sense, and I need to figure out the best way to model that in a test since the upgrade test really is working at the PR level

src/internal/migrations/migrations_test.go
```diff
@@ -48,3 +52,54 @@ func TestMigration(t *testing.T) {
 	require.NoError(t, db.GetContext(ctx, &max, `SELECT max(id) FROM migrations`))
 	assert.Equal(t, state.Number(), max)
 }
+
+func TestEnterpriseConfigMigration(t *testing.T) {
```
brendoncarroll, line 56: This probably isn't a good place for this test.  Now the migrations package depends on enterprise, collections, all stuff it doesn't need. 

You should move this into the enterprise server package.  If you run into dependency issues you can rename the package to `server_test` and Go will compile a separate package, avoiding the circular dependency.

src/internal/clusterstate/v2.1.0.go
```diff
@@ -0,0 +1,16 @@
+package clusterstate
+
+import (
+	"context"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/migrations"
+	enterpriseserver "github.com/pachyderm/pachyderm/v2/src/server/enterprise/server"
+)
+
+var state_2_1_0 migrations.State = state_2_0_0.
```
brendoncarroll, line 10: 👍

src/server/enterprise/server/api_server.go
```diff
@@ -200,15 +201,20 @@ func (a *apiServer) Activate(ctx context.Context, req *ec.ActivateRequest) (resp
...
 			Secret:        req.Secret,
 		}); err != nil {
 			return err
 		}
+		return nil
+	}); err != nil {
+		return nil, err
+	}
 
+	if _, err := col.NewSTM(ctx, a.env.EtcdClient, func(stm col.STM) error {
```
brendoncarroll, line 217: So we are writing this to postgres, and etcd.  Why does it also need to be in etcd.  What happens if the etcd state is cleared?

src/internal/migrations/migrations_test.go
brendoncarroll: Otherwise this test looks good though.

src/server/enterprise/server/api_server.go
acohen4: So here we're writing to a separate collection that's backed by etcd. Reading into the code here I think there may be a bug that I'm thinking through. In general, the enterprise service should heartbeat to the license service to get it's token routinely, and this token is retrieved over an etcd cache. 

Regardless, I think if I make a change to this logic it should be a separate PR

src/server/enterprise/server/api_server.go
brendoncarroll: I think it's fine to cache stuff in etcd.  Idk about it being more performant though.  We could also cache it locally.

brendoncarroll (approved):
LGTM

PR #7118, opened 2021-11-23
Create a basic upgrade test in CI

acohen4:


.circleci/config.yml
```diff
@@ -221,4 +236,9 @@ workflows:
                 - LOAD8
                 - LOAD9
                 - LOAD10
-
+  rootless:
```
brendoncarroll, line 239: rootless snuck back in.  Maybe just comment it out?  It's gone on master.

brendoncarroll (approved):
LGTM

.circleci/config.yml
acohen4: very sneaky. Will take it back out

PR #7119, opened 2021-11-23
Remove mathutil dependency

bbonenfant:
This package is not strictly necessary and it's dependency was causing trouble for M1 macs.

I'm not sure why CI isn't passing -- none of the changes I've made seem to be causing them to fail, but I could be misunderstanding the logs.


brendoncarroll (approved):
LGTM

PR #7120, opened 2021-11-23
Leave workers scaled up until meta compaction finishes.

PFedak:
Autoscaling waits until a job's output commit is fully finished before
potentially scaling down workers. The meta commit is similarly expensive
to finish, so it would benefit from keeping the same number of workers
around.


brycemcanally (approved):
lgtm

PR #7124, opened 2021-11-24
Check spec commit to determine RC freshness.

PFedak:
If a pipeline was deleted and recreated quickly, it's possible for the
delete event not to get processed. This is fine as long as the pipeline
controller notices the RC is out of date and recreates the pipeline
resources. However, our freshness checks only use the version number,
which could be the same before and after, resulting in the RC being
incorrectly left up.

Also propagate update timestamps through to the trace and bump the wait
time for the flaky test back down now that the actual bug is fixed.


acohen4 (approved):
LGTM! Thanks :)

PR #7129, opened 2021-11-29
SQL Query Generation from Cron Tick

brendoncarroll:
Adds a transform `sql-gen-queries` to `pachtf` which reads cron timestamp files from the input repo, and generates SQL queries with the timestamp as a comment in the output repo.

This is a component of a SQL ingest DAG, it can be replaced with something more complicated to allow for custom sharding, or snapshotting the database table before ingest.

### DIFF
--- a/Dockerfile.pachtf
+++ b/Dockerfile.pachtf
@@ -1,4 +1,4 @@
-FROM scratch
+FROM alpine:3.15
 
 LABEL name="Pachyderm" \
       vendor="Pachyderm"
--- a/src/internal/transforms/sqlingest.go
+++ b/src/internal/transforms/sqlingest.go
@@ -100,7 +100,7 @@ func makeWriterFactory(formatName string) (writerFactory, error) {
 		}
 	case "csv":
 		factory = func(w io.Writer, fieldNames []string) sdata.TupleWriter {
-			return sdata.NewCSVWriter(w, fieldNames)
+			return sdata.NewCSVWriter(w, nil)
 		}
 	default:
 		return nil, errors.Errorf("unrecognized format %v", formatName)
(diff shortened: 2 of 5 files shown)


Dockerfile.pachtf
```diff
@@ -1,4 +1,4 @@
-FROM scratch
+FROM alpine:3.15
```
acohen4, line 1: why the image swap?

src/internal/transforms/sqlingest.go
```diff
@@ -96,10 +100,48 @@ func makeWriterFactory(formatName string) (writerFactory, error) {
 		}
 	case "csv":
 		factory = func(w io.Writer, fieldNames []string) sdata.TupleWriter {
-			return sdata.NewCSVWriter(w, fieldNames)
+			return sdata.NewCSVWriter(w, nil)
```
acohen4, line 103: is this change intentional?

src/internal/transforms/sqlingest.go
```diff
@@ -96,10 +100,48 @@ func makeWriterFactory(formatName string) (writerFactory, error) {
...
+
+// SQLQueryGeneration generates queries with a timestamp in the comments
+func SQLQueryGeneration(ctx context.Context, params SQLQueryGenerationParams) error {
+	timestamp, err := readCronTimestamp(params.Logger, params.InputDir)
+	if err != nil {
+		return err
+	}
+	timestampComment := fmt.Sprintf("-- %d\n", timestamp)
+	contents := timestampComment + params.Query
+	outputPath := filepath.Join(params.OutputDir, "0000")
```
acohen4, line 126: I'm missing something here. why "pfs/out/0000" ?

src/server/cmd/pachtf/main.go
```diff
@@ -26,14 +26,20 @@ func main() {
...
 		log.Fatalf("unrecognized transform name %q", transformName)
 	}
+	if err := entrypoint(ctx, log, transformArgs); err != nil {
+		log.Fatal(err)
+	}
+}
+
+type Entrypoint = func(ctx context.Context, log *logrus.Logger, args []string) error
+
+var entrypoints = map[string]Entrypoint{
```
acohen4, line 40: 👍

Dockerfile.pachtf
brendoncarroll: We assume a temp directory exists in the worker, and this was causing the pipeline to break.  I wasn't sure what else we needed so I just switched to a real base image instead.

src/internal/transforms/sqlingest.go
brendoncarroll: Yeah, setting the csv writer fields to nil removes the header.  I don't really see the benefit of the header since the user will know what the asked for and it makes splitting and concatenating files more difficult.

src/internal/transforms/sqlingest.go
brendoncarroll: `0000` is just the path that the query generation writes to by default.  I don't want users to get accustomed to something like RESULTS.json.  They should be writing code to consume an unknown number of files from the output, which will be based on how many separate queries they asked for in the input.  I'm open to suggestions here, I was thinking you would just increment this number if we ever start generating more than one query.

src/internal/transforms/sqlingest.go
acohen4: makes sense 👍

src/server/cmd/pachtf/main.go
```diff
@@ -75,8 +90,14 @@ func sqlGenQueries(ctx context.Context, log *logrus.Logger, args []string) error
 		return errors.Errorf("must provide query")
 	}
 	query := args[0]
-	inputDir := filepath.FromSlash(pfs + "/in")
-	outputDir := filepath.FromSlash(pfsOut)
+	inputDir, err := filepath.EvalSymlinks(filepath.FromSlash(pfs + "/in"))
```
acohen4, line 93: what's the case where these paths would need symlinks resolved?

acohen4 (approved):
LGTM

src/server/cmd/pachtf/main.go
brendoncarroll: The input subdirectories (the immediate children of `/pfs/`) are symlinks.  `filepath.WalkDir` doesn't follow symlinks.

PR #7132, opened 2021-11-30
Add helm value: pachd.activateAuth that can be set to false to activate the enterprise license without auth

acohen4:


acohen4:
FYI @dgeorg42 @tybritten

etc/helm/pachyderm/values.yaml
```diff
@@ -217,6 +217,9 @@ pachd:
   ## if pachd.activateEnterpriseMember is set, enterprise will be activated and connected to an existing enterprise server.
   ## if pachd.enterpriseLicenseKey is set, enterprise will be activated.
   activateEnterpriseMember: false
+  ## if pachd.setupEnterpriseOnly is set, enterprise service will be activated on the cluster without additional enterprise features such as auth.
+  setupEnterpriseOnly: false
```
chainlink, line 221: Perhaps this could be `activateAuth` and just always set to true?

chainlink (approved):
One small comment, but otherwise looks good

PR #7141, opened 2021-12-02
Update pachub links to point at landing page

danthebaker:
To be put live once the new hub landing page is live in WordPress


npepin-hub (approved):
LGTM

PR #7143, opened 2021-12-06
Task Service

brycemcanally:
This PR implements a redesign of the work package, which is now being called the task package. The primary goals of these changes are:
- Enable multiple processes / goroutines to manage tasks within a namespace / group.
- Simplify the interface: input is an Any proto and output is an Any proto or error and added a utility function that takes a task and returns a result (as compared to having to set up a collection callback).
- Change the scheduling of tasks to be based on fairness across groups rather than strict priority based on group creation order.
- Change the cleanup to be based on lease renewal in etcd rather than explicit cleanup at the startup of a master component (this enables anything in the system to manage tasks in a namespace / group and have the cleanup be handled automatically).
- Compute stable IDs based on the content of the provided input Any to enable us to add a task service caching layer.

The original design document is here: https://docs.google.com/document/d/1UNWe-Pu8_CcApMxf2TazMQ1XxlQPV8dk1_Kg11tljDQ.
The final interfaces and documentation are in `src/internal/task/task.go`.


src/internal/task/task.proto
```diff
@@ -13,20 +13,18 @@ enum State {
...
+message Group {}
+
 message Task {
   string id = 1 [(gogoproto.customname) = "ID"];
-  google.protobuf.Any data = 2;
-}
-
-message TaskInfo {
-  Task task = 1;
   State state = 2;
```
brendoncarroll, line 20: These fields, including state could potentially be a oneof for RUNNING, SUCCESS, and FAILURE.

src/internal/task/etcd_task_service.go
```diff
@@ -0,0 +1,340 @@
...
+	})
+}
+
+func (es *etcdService) Maker(namespace string) Maker {
+	namespaceEtcd := newNamespaceEtcd(es.etcdClient, es.etcdPrefix, namespace)
+	return newEtcdMaker(namespaceEtcd)
+}
+
+type Maker interface {
+	Make(ctx context.Context, cb ProcessFunc) error
```
brendoncarroll, line 206: `GetOne` was more descriptive.  I'm not sure how many times this will call cb, or if it blocks without a comment.

src/internal/task/etcd_task_service.go
```diff
@@ -0,0 +1,340 @@
...
+
+const (
+	groupPrefix = "/group"
+	taskPrefix  = "/task"
+	claimPrefix = "/claim"
+)
+
+type Service interface {
+	// TODO: Should group ID parameter?
+	Doer(ctx context.Context, namespace, groupID string, cb func(context.Context, Doer) error) error
```
brendoncarroll, line 27: Why does this need a context?, why does it need to return an error, and why does it take a callback instead of returning the doer.

src/internal/collection/transaction.go
```diff
@@ -220,6 +221,13 @@ func (s *stm) Put(key, val string, ttl int64, ptr uintptr) error {
 	return nil
 }
 
+func (s *stm) PutLease(key, val string, lease v3.LeaseID, ptr uintptr) error {
+	s.Lock()
+	defer s.Unlock()
+	s.wset[key] = stmPut{val, ttl, v3.OpPut(key, val, v3.WithLease(lease)), ptr}
```
brendoncarroll, line 227: Does the linter complain about no fields when creating the `stmPut` ?

src/internal/collection/types.go
```diff
@@ -42,6 +42,8 @@ type EtcdCollection interface {
 	// Claim attempts to claim a key and run the passed in callback with
 	// the context for the claim.
 	Claim(ctx context.Context, key string, val proto.Message, f func(context.Context) error) error
+
+	WithRenewer(ctx context.Context, cb func(context.Context, *Renewer) error) error
```
brendoncarroll, line 46: The callback is necessary here so that a new context scoped to the renewer can be passed through?

src/internal/task/etcd_task_service.go
brendoncarroll: The callback in particular is going to make initialization very difficult.

A system (like PFS) should be able to receive a Service, and then spawn Doers or makers and pass them to the master and compaction worker.  As written, the entire lifetime of the compaction worker or pfs master would have to be limited to inside the callback.

src/internal/task/etcd_task_service.go
brendoncarroll: `ProcessOne` with the callback seems pretty descriptive too.

src/internal/task/etcd_task_service.go
```diff
@@ -0,0 +1,340 @@
...
+	*namespaceEtcd
+}
+
+func newEtcdMaker(namespaceEtcd *namespaceEtcd) Maker {
+	return &etcdMaker{
+		namespaceEtcd: namespaceEtcd,
+	}
+}
+
+type ProcessFunc func(context.Context, *types.Any) (*types.Any, error)
```
brendoncarroll, line 219: This should be a type alias, it improves the compiler errors when you pass the wrong type.

src/internal/task/etcd_task_service_test.go
```diff
@@ -0,0 +1,175 @@
...
+	// Set up task service.
+	s := NewEtcdService(env.EtcdClient, "")
+	// Set up workers.
+	workerCtx, workerCancel := context.WithCancel(context.Background())
+	workerEg, errCtx := errgroup.WithContext(workerCtx)
+	for i := 0; i < numWorkers; i++ {
+		workerEg.Go(func() error {
+			maker := s.Maker("")
+			for {
+				ctx, cancel := context.WithCancel(errCtx)
```
brendoncarroll, line 63: This cancel leaks.

src/internal/task/etcd_task_service_test.go
```diff
@@ -0,0 +1,175 @@
...
+			var inputs []*types.Any
+			for j := 0; j < numTasks; j++ {
+				input, err := serializeTestTask(&TestTask{ID: strconv.Itoa(j)})
+				if err != nil {
+					return err
+				}
+				inputs = append(inputs, input)
+				created[i][j] = true
+			}
+			ctx, cancel := context.WithCancel(errCtx)
```
brendoncarroll, line 109: Cancel leaks here too.

src/internal/task/etcd_task_service_test.go
```diff
@@ -0,0 +1,175 @@
...
+func test(t *testing.T, workerFailProb, groupCancelProb, taskFailProb float64) {
+	seed := time.Now().UTC().UnixNano()
+	rand.Seed(seed)
+	msg := seedStr(seed)
+	env := testetcd.NewEnv(t)
+	numGroups := 10
+	numTasks := 10
+	numWorkers := 5
+	// Set up task service.
+	s := NewEtcdService(env.EtcdClient, "")
```
brendoncarroll, line 55: You might consider adding an argument to test: `s Service`, and then creating a fresh etcd from `testetcd` as part of the setup of each test below.

src/internal/task/etcd_task_queue.go
```diff
@@ -0,0 +1,113 @@
...
+				cycleKey = ""
+			default:
+			}
+			tq.requeueGroup(key)
+		}
+	}()
+	return tq
+}
+
+func (tq *taskQueue) group(ctx context.Context, groupID string, cb func(context.Context, chan taskFunc)) error {
```
brendoncarroll, line 74: Is this method how you put tasks in the queue?

src/internal/task/etcd_task_queue.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	ctx, cancel := context.WithCancel(ctx)
+	taskFuncChan := make(chan taskFunc, bufSize)
+	ge := &groupEntry{
+		cancel:       cancel,
+		taskFuncChan: taskFuncChan,
+	}
+	tq.groups.Set(groupID, ge)
+	go func() {
+		cb(ctx, taskFuncChan)
```
brendoncarroll, line 88: Why is this async?

src/internal/task/etcd_task_service.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+		result = task.Output
+		return errutil.ErrBreak
+	}); err != nil {
+		return nil, err
+	}
+	return result, nil
+}
+
+type CollectFunc func(int, *types.Any, error) error
```
brendoncarroll, line 144: Could be a type alias.

src/internal/task/etcd_task_service.go
```diff
@@ -0,0 +1,340 @@
...
+func newEtcdDoer(namespaceEtcd *namespaceEtcd, groupID string, renewer *col.Renewer) Doer {
+	return &etcdDoer{
+		namespaceEtcd: namespaceEtcd,
+		groupID:       groupID,
+		renewer:       renewer,
+	}
+}
+
+func (ed *etcdDoer) Do(ctx context.Context, any *types.Any) (*types.Any, error) {
+	// TODO: include any type in hash?
```
brendoncarroll, line 101: We need to remarshal the whole Any here and then call `pachhash.Sum` on that.

src/internal/task/etcd_task_service.go
```diff
@@ -0,0 +1,340 @@
...
+		etcdClient,
+		etcdPrefix,
+		nil,
+		template,
+		nil,
+		nil,
+	)
+}
+
+type Doer interface {
```
brendoncarroll, line 81: I think the `Doer` and `Maker` interfaces should be with the Service so the whole model is contiguous in one place.  We also need to comments on those interfaces.

src/internal/task/etcd_task_service.go
brendoncarroll: +1 to  the `int`

src/internal/task/etcd_task_service.go
brycemcanally: The `One` part of those names is very misleading. The expected behavior would be to always be making tasks then executing the process function (until the context is canceled). It is more like a for each, but I left it as `Make` for parity with the `Doer` interface (`Doer` has `Do`, `Maker` has `Make`). I don't have too strong of an opinion about this other than `One` shouldn't be a part of the name. Also, agreed on the documentation, that is what I am working on now.

src/internal/task/etcd_task_service.go
brycemcanally: We don't want the doers to be scoped like that because we want the groups to be at a lower level of granularity (i.e. the pfs master would create a doer for each repo or commit (it is commit now) being processed rather than a doer for all of the commits that will be processed). The PFS / PPS code has been migrated over if you want to see what it looks like. We need to return a callback and close the renewer, which is why I went with this pattern, but we could also handle that by exposing a close method and returning a context. The callback pattern just seems safer / cleaner for this.

src/internal/collection/transaction.go
brycemcanally: No.

src/internal/collection/types.go
brycemcanally: Yes.

src/internal/task/etcd_task_queue.go
brycemcanally: This is how you set up a group in the internal task queue. Tasks are sent through the provided channel in the callback. This is used by the makers for managing the queuing of tasks across groups. Each group has a queue that is fed into this task queue which attempts to ensure fair execution of tasks across groups (tasks are the unit of scheduling, so it may not necessarily be fair in practice for groups with many small tasks vs groups with larger tasks, but I don't think that is a scheduling problem we should try to solve).

src/internal/task/etcd_task_queue.go
brycemcanally: We need to kick off goroutines through the maker as groups are added that will manage the queuing for the groups (basically watch the appropriate queues and send the tasks to the task queue). It could be at a level up, but this seems like a pretty good way to manage it. The context will get canceled if the group is deleted, which should be respected by the callback.

src/server/pps/server/server.go
```diff
@@ -45,12 +48,15 @@ type Env struct {
...
 func EnvFromServiceEnv(senv serviceenv.ServiceEnv, txnEnv *txnenv.TransactionEnv, reporter *metrics.Reporter) Env {
+	etcdPrefix := path.Join(senv.Config().EtcdPrefix, senv.Config().PPSEtcdPrefix)
 	return Env{
 		DB:            senv.GetDBClient(),
 		TxnEnv:        txnEnv,
 		Listener:      senv.GetPostgresListener(),
 		KubeClient:    senv.GetKubeClient(),
 		EtcdClient:    senv.GetEtcdClient(),
+		EtcdPrefix:    etcdPrefix,
+		TaskService:   senv.GetTaskService(etcdPrefix),
```
brendoncarroll, line 59: Do we actually need the etcdClient for anything other than the task service?

brycemcanally: Yeah, for the code that connects to the worker API servers. The connection info is recorded in etcd by the workers.

src/internal/task/task.go
```diff
@@ -0,0 +1,69 @@
...
+type Source interface {
+	// Iterate iterates through tasks until the provided context is canceled.
+	// For each task, the process function will be called and the results
+	// will be returned to the Doer that created the task.
+	Iterate(ctx context.Context, cb ProcessFunc) error
+}
+
+// CollectFunc is the type of a function that is used for collecting the output of a stream / batch of tasks.
+// Index is the index of a task with respect to the order in which the task was created in the stream / batch.
+type CollectFunc = func(index int64, output *types.Any, _ error) error
```
brendoncarroll, line 64: This should probably be an `int` since it is going to be used as an index to a slice.

src/internal/task/util.go
```diff
@@ -0,0 +1,46 @@
...
+	})
+	eg.Go(func() error {
+		for _, input := range inputs {
+			select {
+			case inputChan <- input:
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		}
+		close(inputChan)
```
brendoncarroll, line 42: deferring this close is probably safer, just in case `doer.Do` isn't respecting the context.

src/internal/task/task.go
brycemcanally: I tend to use int64 for things that may need to support very high scale at some point. Although it probably won't be needed, the downside (casting to an int) doesn't really outweigh the potential upside (supporting higher scale) imo.

src/internal/task/util.go
brycemcanally: Closing the input channel signals that all of the input has been sent. If there is a bug in the Do implementation that causes it to not respect the context, then we are more likely to know where the issue is if we get stuck there vs returning with incorrect results (not all of the input would have been processed).

src/server/worker/pipeline/transform/registry.go
```diff
@@ -283,16 +277,18 @@ func (reg *registry) processJobRunning(pj *pendingJob) error {
...
+	taskDoer := reg.driver.NewTaskDoer(pj.ji.Job.ID)
+	if err := func() error {
+		if err := pj.withParallelDatums(ctx, func(ctx context.Context, dit datum.Iterator) error {
+			return reg.processDatums(ctx, pj, taskDoer, dit)
 		}); err != nil {
 			return err
 		}
-		return pj.withSerialDatums(master.Ctx(), func(ctx context.Context, dit datum.Iterator) error {
-			return reg.processDatums(ctx, pj, master, dit)
+		return pj.withSerialDatums(ctx, func(ctx context.Context, dit datum.Iterator) error {
```
brendoncarroll, line 288: Was this supposed to change from the `master.Ctx()` to the `pachClient.Ctx()`

src/server/pfs/server/pfsserver.proto
```diff
@@ -4,13 +4,11 @@ package pfsserver;
 option go_package = "github.com/pachyderm/pachyderm/v2/src/server/pfs/server";
 
 message ShardTask {
-  int64 index = 1;
```
brendoncarroll, line 7: 🎉

src/server/pfs/server/pfsserver.proto
```diff
@@ -4,13 +4,11 @@ package pfsserver;
...
-  repeated string inputs = 2;
+  repeated string inputs = 1;
```
brendoncarroll, line 7: It's probably safer to not renumber these, but I guess if we clear etcd, and the upgrade happens on all the workers and pachds before any work is scheduled, then it shouldn't be a problem.

src/internal/task/etcd_queue.go
```diff
@@ -0,0 +1,113 @@
...
+			iter := tq.groups.IterFunc()
+			kv, ok := iter()
+			tq.mu.Unlock()
+			var key string
+			if ok {
+				key = kv.Key.(string)
+			}
+			// Wait if there are no group entries or a cycle has been completed without finding a task.
+			if !ok || key == cycleKey {
+				time.Sleep(waitTime)
```
brendoncarroll, line 53: This is probably fine since the wait time is so short, but generally this should be a select from a ticker and `ctx.Done()`

src/server/worker/pipeline/transform/registry.go
brycemcanally: Yes, there isn't a master anymore and the renewal is handled in the Do call. I'm not sure what you are asking about a master lock, there is no locking involved here (or needed). Each Do call operates in its own prefix in etcd.

brendoncarroll (approved):
LGTM

src/server/pfs/server/pfsserver.proto
brycemcanally: We shouldn't support compatibility with prior etcd state. It would be a bug if we used any of it anyways. The model we should have is that all of the etcd state is deleted during an upgrade and we should be able to make whatever changes we want to the format / what we store in etcd.

PR #7144, opened 2021-12-07
Adding Helm value: pachd.customRootCertsDir to be used for bootstrapping by config-pod -> pachd

acohen4:


chainlink:
Having a value under the pachd key that doesn't affect pachd seems not ideal. Does this affect pach ssl at all?

acohen4:
Did not go with this solution. went with #7155

PR #7146, opened 2021-12-08
upgrade ingress from v1beta to v1

tybritten:
Some ingress controllers don't treat paths without `*` as prefix types in v1beta. Also moved the `/*` to the end as the catchall


chainlink:
Hi @tybritten which ingress controllers has this been tested against?

tybritten:
I had tested it against alb and traefik but now it doesn't work on nginx if you're on 1.19+...might be time to just switch to v1. The current config doesn't work with alb

chainlink:
Switching to v1 makes sense to me, ingress has been out of beta for a while

tybritten:
ok updated

acohen4 (approved):
LGTM

acohen4 (commented):
Still one failing test though

PR #7151, opened 2021-12-08
Context timeout interceptor

acohen4:


src/internal/middleware/context/interceptor.go
```diff
@@ -0,0 +1,99 @@
+package context
```
brendoncarroll, line 1: This is the name of a standard library package.  Almost every package we have references `context` so this package will always be difficult to refer to.  Maybe something like `ctxmiddleware` or `ctxintercept`

src/internal/middleware/context/interceptor.go
```diff
@@ -0,0 +1,99 @@
...
+	"/pps_v2.API/RunLoadTest":        unlimited,
+	"/pps_v2.API/RunLoadTestDefault": unlimited,
+}
+
+func setTimeout(fullMethod string, ctx context.Context) context.Context {
+	if timeout, ok := customTimeoutMethods[fullMethod]; ok {
+		if timeout == unlimited {
+			return ctx
+		}
+		newCtx, _ := context.WithTimeout(ctx, timeout)
```
brendoncarroll, line 83: This _ is a cancel function.  You have to call it to prevent leaking resources.  You can call it as many times as you like, as long as the number of times is a positive integer.

https://pkg.go.dev/context
> Failing to call the CancelFunc leaks the child and its children until the parent is canceled or the timer fires.

You might change `setTimeout` to `getTimeout(fullMethod string) time.Duration`.  You can use 0 or nil to represent unlimited; I'll leave that to you.  Then in the functions down below create the context, and defer call the cancel function, so it's always cancelled when the function exits.
```go
// check for your unlimited value here
ctx, cf := context.WithTimeout(ctx, timeout)
defer cf()
```

jrockway:
Part of https://pachyderm.atlassian.net/browse/CORE-281

PR #7152, opened 2021-12-08
Stream PostgresCol List() from row buffer

acohen4:


src/internal/collection/postgres_collection.go
```diff
@@ -301,20 +326,63 @@ func (c *postgresCollection) list(
...
 		}
 	}
 
 	if opts.Limit > 0 {
 		query += fmt.Sprintf(" limit %d", opts.Limit)
 	}
+	return query, args, nil
+}
 
+func (c *postgresCollection) list(
```
brendoncarroll, line 374: This will not return a consistent view of the database if we have to do multiple queries.

brendoncarroll:
This could be fine as a potential fix, but I think we should weigh some of the options.

1. This implementation. We code around inconsistent reads if and when that ever comes up.
2. Keep the list method the same. Change the call sites to read into a buffer.  The buffer can just grow forever, and we can deal with OOMing when the time comes.
3. Change the list method to buffer everything in memory, deal with OOMs later, don't change any call sites, callers no longer have the ability to stream results, it's always buffered.
4. Read up to a maximum size into a buffer, if the maximum is exceeded, return an error.  If clients start receiving the error, they can query with an offset or more predicates.  There is already no expectation of consistency between API calls, so this would not be surprising behavior.

src/internal/collection/postgres_collection.go
acohen4: right we would have to be okay with that here. @brycemcanally mentioned that's the downside of this approach, but that it should be okay for these operations. What do you think?

acohen4: Still not perfect, but one note is we can guarantee that no two rows will be returned twice within the same stream since the results are always sorted, if a row got added/updated between queries we would not revisit it since it would have already been passed in the sort order. That is only with the exception of `SortOrder==ASC && TargetColumn==UPDATED_AT`

brendoncarroll (approved):
There was some discussion in slack about inconsistent reads, and it seems like it will be fine for list because list is for read only collections.

src/internal/collection/postgres_collection.go
```diff
@@ -301,20 +326,63 @@ func (c *postgresCollection) list(
 		query += " where " + strings.Join(fields, " and ")
 	}
 
+	afterCondition := func(t SortTarget, ord SortOrder) (string, error) {
```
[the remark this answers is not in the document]
acohen4: great catch, and suggestion :)

Tryneus (approved):
LGTM

src/internal/collection/postgres_collection.go
```diff
@@ -598,6 +630,14 @@ func (c *postgresReadOnlyCollection) Watch(opts ...watch.Option) (watch.Watcher,
 			return
 		}
 
+		// if waitEv was set but not sent yet, send it now
+		if waitEv != nil && interested(waitEv, options.SortTarget) {
```
acohen4, line 634: whether the watcher is interested should decide whether we send here

src/internal/collection/postgres_collection.go
```diff
@@ -72,14 +77,21 @@ func WithNotFoundMessage(format func(interface{}) string) Option {
 	}
 }
 
+func WithListBufferCapOverride(cap int) Option {
```
brycemcanally, line 80: This would probably be fine as `WithListBufferCapacity`, the override is implicit when using the functional option.

src/internal/collection/postgres_collection.go
```diff
@@ -301,20 +328,72 @@ func (c *postgresCollection) list(
...
 		query += fmt.Sprintf(" limit %d", opts.Limit)
 	}
+	return query, args, nil
+}
 
+func (c *postgresCollection) list(
+	ctx context.Context,
+	withFields map[string]string,
+	opts *Options,
+	greedy bool,
```
brycemcanally, line 389: Do we need a greedy option anymore? I don't think we do now that we don't hold a transaction open when listing.

src/internal/collection/postgres_collection.go
```diff
@@ -351,30 +430,72 @@ func (c *postgresCollection) list(
 		return nil
 	}
 
-	result := &model{}
-	for rows.Next() {
-		if err := rows.StructScan(result); err != nil {
-			return c.mapSQLError(err, "")
+	// To avoid holding the DB connection open for an unknown duration
```
brycemcanally, line 433: This is a bit pedantic, but we should probably be more precise about the fact that we are opening / closing a transaction during the read rather than opening / closing a database connection. The database connection to pgbouncer will remain open, but by finishing the transaction we will ensure that a real database connection is freed up for a different client.

src/internal/collection/postgres_collection.go
```diff
@@ -440,48 +561,91 @@ func (c *postgresReadOnlyCollection) Count() (int64, error) {
...
+	// event emition to the watcher if we encounter a listed record that is
+	// in the future of a buffered event
+	if err := c.list(withFields, &Options{Target: options.SortTarget, Order: etcd.SortAscend}, func(m *model) error {
+		if err := proto.Unmarshal(m.Proto, val); err != nil {
+			return errors.EnsureStack(err)
+		}
+		lastTs := lastTimestamp(m, options.SortTarget)
+
+		if bufEvent == nil {
+		loop:
```
brycemcanally, line 589: I don't think you need the loop here. You could just use a select with an empty default to proceed if an event isn't available.

src/internal/collection/postgres_collection.go
```diff
@@ -440,48 +561,91 @@ func (c *postgresReadOnlyCollection) Count() (int64, error) {
...
+				}
 			}
+		}
 
-			if lastUpdated.Before(m.UpdatedAt) {
-				lastUpdated = m.UpdatedAt
+		if bufEvent != nil {
+			if bufEvent.time.Unix() <= lastTs.Unix() {
+				if err := watcher.sendInitial(bufEvent.WatchEvent(c.ctx, watcher.db, watcher.template)); err != nil {
+					watcher.sendInitial(&watch.Event{Type: watch.EventError, Err: err})
```
brycemcanally, line 603: I think you could just return the error here. The same logic will be applied by the error handling for the outer function.

src/internal/collection/postgres_listener.go
```diff
@@ -217,6 +217,7 @@ type postgresEvent struct {
 	protoData []byte          // the serialized protobuf of the row data (exclusive with storedID)
 	storedID  string          // the ID of a temporary row in the large_notifications table
 	err       error           // any error that occurred in parsing
+	create    bool
```
brycemcanally, line 220: Is this used anywhere?

src/internal/collection/postgres_collection.go
```diff
@@ -351,30 +430,72 @@ func (c *postgresCollection) list(
...
+	// dictated by the client's callback, we:
+	// (1) buffer the SQL rows
+	// (2) close the connection
+	// (3) apply f(), the client's callback, to results in the buffer
+	// (4) if the buffer was full, re-execute the query, offset by key, and repeat (1)
+	bufferResults := func(rs *sqlx.Rows) ([]*model, int, error) {
+		defer rs.Close()
+		var rowCnt int
+		rowsBuffer := make([]*model, 0, c.listBufferCapacity)
+		result := &model{}
```
brycemcanally, line 443: You could just declare the model in the for loop since it is reset each iteration.

src/internal/collection/postgres_listener.go
acohen4: no reverting this. Was using this in the previous logic iteration

src/internal/collection/postgres_collection.go
```diff
@@ -351,30 +430,72 @@ func (c *postgresCollection) list(
...
-	for rows.Next() {
-		if err := rows.StructScan(result); err != nil {
-			return c.mapSQLError(err, "")
+	// To avoid holding the DB connection open for an unknown duration
+	// dictated by the client's callback, we:
+	// (1) buffer the SQL rows
+	// (2) close the connection
+	// (3) apply f(), the client's callback, to results in the buffer
+	// (4) if the buffer was full, re-execute the query, offset by key, and repeat (1)
+	bufferResults := func(rs *sqlx.Rows) ([]*model, int, error) {
```
brycemcanally, line 439: Could we make this function just do everything for getting a batch (get the query string, perform the query, and buffer the rows)? Having the row close being handled by a different function from the one that created it seems like a bad code smell. We could also just inline this in the for loop with the `if err := func() {...}(); err != nil {...}` pattern if the reason this exists is for the defer.

src/internal/collection/postgres_collection.go
acohen4: You're right 👍

acohen4: Based on the initial slack thread, we decided that the "Dirty read semantics" that are exhibited by Lists that use multiple buffers are acceptable for ReadOnly but, ReadWrite may be more risky.

src/internal/collection/postgres_collection.go
```diff
@@ -301,80 +330,131 @@ func (c *postgresCollection) list(
...
+	f func(*model) error,
+) error {
+	// To avoid holding a transaction open (which holds a DB connection) for an unknown duration
+	// dictated by the client's callback, we:
+	// (1) buffer the SQL rows
+	// (2) close the connection
+	// (3) apply f(), the client's callback, to results in the buffer
+	// (4) if the buffer was full, re-execute the query, offset by key, and repeat (1)
+	bufferResults := func(last *model) ([]*model, bool, error) {
+		query, args, err := c.listQueryStr(ctx, withFields, opts, last)
```
brycemcanally, line 401: We should use the limit option and set it to the list buffer capacity. That can improve the query performance in certain cases and prevent more rows than we intend to read from being read on the server side / sent.

src/internal/collection/postgres_collection.go
```diff
@@ -301,80 +330,131 @@ func (c *postgresCollection) list(
...
+	ctx context.Context,
+	withFields map[string]string,
+	opts *Options,
+	q sqlx.ExtContext,
+	f func(*model) error,
+) error {
+	// To avoid holding a transaction open (which holds a DB connection) for an unknown duration
+	// dictated by the client's callback, we:
+	// (1) buffer the SQL rows
+	// (2) close the connection
```
brycemcanally, line 397: Close the transaction. We don't really explicitly do this as much as read a subset of the query results. Switching to the limit approach may warrant rewording this as performing a query with a limit to get a batch.

src/internal/collection/postgres_collection.go
```diff
@@ -279,14 +295,27 @@ func targetToSQL(target etcd.SortTarget) (string, error) {
...
+		return m.Key, nil
+	}
+	return "", errors.Errorf("unsupported sort target for postgres collections: %d", target)
+}
+
+func sortOrderOperator(order SortOrder) string {
+	if order == SortDescend {
+		return "<"
+	}
+	// SortAcend + SortNone
```
brycemcanally, line 314: `SortAcend` -> `SortAscend`

src/internal/collection/postgres_collection.go
acohen4: great point. I think it's worth considering the "OFFSET" option as well, that could be used to avoid the `WHERE <TARGET> <OP> <VALUE>` logic used to start at an offset by comparing the last seen value. I think the current WHERE clause approach makes more sense since it's slightly less sensitive to concurrency issues where the data is modified between batches

brycemcanally (approved):
lgtm

PR #6930, opened 2021-09-26
Pipeline Templates

msteffen:
This PR implements the MVP of Pipeline Templates, giving pachctl the ability to execute Jsonnet templates in {create,update} pipeline. For context, and the design, see https://docs.google.com/document/d/1VC8SY2hmf5BQShJSmUvQ_q0hXV6ErRl25BmgK1f_KtI/edit

PR #7154, opened 2021-12-08
Pipeline Templates RPCs

brendoncarroll:
Creates a library for pachyderm templates, which are jsonnet based templates, first prototyped in #6930.  The plan is to implement most of the functionality on the server side, so that it can be used by console.

- Adds RenderTemplate and ApplyTemplate rpcs to pps.


sqlingest.jsonnet
```diff
@@ -0,0 +1,56 @@
+local pachyderm = {
```
brendoncarroll, line 1: I will find a better spot for this file.

src/internal/pachtmpl/pachtmpl.go
```diff
@@ -0,0 +1,105 @@
...
+		vm.TLAVar(key, value)
+	}
+	output, err := vm.EvaluateAnonymousSnippet("main", string(tmpl))
+	if err != nil {
+		return "", errors.Wrapf(err, "template err")
+	}
+	return output, nil
+}
+
+// Eval evaluates the jsonnet at entrypointPath using fsContext to resolve imports
```
acohen4, line 39: A little unclear to me what the arguments are meant to be here and whether we're using this function. What's the use-case we're addressing here? Also what's the 'fs' in 'fsContext'?

acohen4 (approved):
Overall LGTM! I left one question around the Eval snippet that I don't really understand

src/internal/pachtmpl/pachtmpl.go
brendoncarroll: It stands for filesystem.  It's similar to a docker context, the idea is that you have a set of jsonnet files in the local filesystem and you want to evaluate the entrypoint, which could reference the other files through import statements.  

We still have to figure out the best way to collect the relevant files and send them over the wire.  That's not done yet, but if and when we figure out they, you would plug in the files there.

brendoncarroll: Explained another way: The map is from file paths to file contents, and the entrypoint path is the path with the contents to evaluate.  When evaluating the contents of the file at entrypoint, import statements may result in additional lookups in the map.

msteffen (approved):
One small question

src/server/pps/cmds/cmds.go
```diff
@@ -679,14 +680,18 @@ All jobs created by a pipeline will create commits in the pipeline's output repo
...
 		Short: "Create a new pipeline.",
 		Long:  "Create a new pipeline from a pipeline specification. For details on the format, see https://docs.pachyderm.com/latest/reference/pipeline_spec/.",
 		Run: cmdutil.RunFixedArgs(0, func(args []string) (retErr error) {
-			return pipelineHelper(false, pushImages, registry, username, pipelinePath, false)
+			return pipelineHelper(false, pushImages, registry, username, pipelinePath, jsonnetPath, jsonnetArgs, false)
 		}),
 	}
-	createPipeline.Flags().StringVarP(&pipelinePath, "file", "f", "-", "The JSON file containing the pipeline, it can be a url or local file. - reads from stdin.")
+	createPipeline.Flags().StringVarP(&pipelinePath, "file", "f", "", "A JSON file (url or filepath) containing one or more pipelines. \"-\" reads from stdin (the default behavior). Exactly one of --file and --jsonnet must be set.")
+	createPipeline.Flags().StringVar(&jsonnetPath, "jsonnet", "", "A Jsonnet template file (url or filepath) for one or more pipelines. \"-\" reads from stdin. Exactly one of --file and --jsonnet must be set. Jsonnet templates must contain a top-level function; strings can be passed to this function with --arg (below)")
```
msteffen, line 693: Nit: would it be possible to mark this feature beta in the flag description, so that we're not constrained by semver in terms of renaming things, etc. in the early iterations on this feature?

msteffen (approved):
One more request

etc/contributing/create_sql_ingest.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
brendoncarroll: I can make a PR to the examples repo.  Should I drop those files from this PR?

msteffen:
INT-467 - merging pipeline templates

PR #7155, opened 2021-12-09
Expose global.customCaCerts helm value for use of custom System Certs

acohen4:
to instruct usage of pachd-tls-cert#tls.crt k8s secret as the root cert for pachd, console, and enterprise-server


tybritten:
Looks good. I'd add a comment explaining the setting so it gets picked up by docs

tybritten (approved):

acohen4:
> Looks good. I'd add a comment explaining the setting so it gets picked up by docs

Great call 👍

chainlink (approved):

tybritten (approved):

chainlink:
nit: probably should be `global.customCACerts`

PR #7157, opened 2021-12-09
Allow existing secrets for PostgreSQL Passwords

tybritten:


etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -65,8 +65,8 @@ spec:
         - name: POSTGRES_PASSWORD
           valueFrom:
             secretKeyRef:
-              name: postgres # Must match secret setup by postgres subchart or postgres-secret.yaml
-              key: postgresql-password
+              name: {{ .Values.global.postgresqlExistingSecretName | default "postgres" }}
```
hobbsh, line 68: Shouldn't this be `.Values.global.postgresql.postgresqlExistingSecretName`?

etc/helm/pachyderm/templates/pachd/postgres-secret.yaml
```diff
@@ -2,7 +2,7 @@
 SPDX-FileCopyrightText: Pachyderm, Inc. <info@pachyderm.com>
 SPDX-License-Identifier: Apache-2.0
 */ -}}
-{{- if and .Values.pachd.enabled  (not .Values.postgresql.enabled) -}}
+{{- if and .Values.pachd.enabled  (not .Values.postgresql.enabled) (not .Values.global.postgresqlExistingSecret)-}}
```
hobbsh, line 5: This should be `not .Values.global.postgresqlExistingSecretName` I think

etc/helm/pachyderm/values.yaml
```diff
@@ -15,6 +15,10 @@ global:
     # When using autogenerated value for the initial install, it must be pulled from the
     # postgres secret and added to values.yaml for future helm upgrades
     postgresqlPassword: "" 
+    # If you want to supply the postgresql password in an existing secret, leave Password blank and
+    # Supply the name of the existing secret in the namespace and the key in that secret with the password
+    postgresqlExistingSecretName: ""
```
hobbsh, line 20: Possibly a bit pedantic, but I think having `postgresql` in the value name is redundant, because the value path already contains it. `existingSecretName`/`existingSecretKey` makes more sense to me, but I get that you're following the established pattern here.

etc/helm/pachyderm/templates/pachd/deployment.yaml
hobbsh: I think you also have an extra space in front of `key` here, throwing things off.

etc/helm/pachyderm/values.yaml
tybritten: agree but you've got it

etc/helm/pachyderm/templates/pachd/postgres-secret.yaml
```diff
@@ -2,7 +2,7 @@
 SPDX-FileCopyrightText: Pachyderm, Inc. <info@pachyderm.com>
 SPDX-License-Identifier: Apache-2.0
 */ -}}
-{{- if and .Values.pachd.enabled  (not .Values.postgresql.enabled) -}}
+{{- if and .Values.pachd.enabled  (not .Values.postgresql.enabled) (not .Values.global.postgresqul.postgresqlExistingSecretName)-}}
```
acohen4, line 5: I think there needs to be a space between the last value and the `-}}`?

etc/helm/pachyderm/templates/pachd/postgres-secret.yaml
```diff
@@ -2,7 +2,7 @@
 SPDX-FileCopyrightText: Pachyderm, Inc. <info@pachyderm.com>
 SPDX-License-Identifier: Apache-2.0
 */ -}}
-{{- if and .Values.pachd.enabled  (not .Values.postgresql.enabled) -}}
+{{- if and .Values.pachd.enabled  (not .Values.postgresql.enabled) (not .Values.global.postgresqul.postgresqlExistingSecretName) -}}
```
acohen4, line 5: ah! I think it's also postgresql is mispelled in .Values.global.**postgresqul**.postgresqlExistingSecretName

etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -65,8 +65,8 @@ spec:
         - name: POSTGRES_PASSWORD
           valueFrom:
             secretKeyRef:
-              name: postgres # Must match secret setup by postgres subchart or postgres-secret.yaml
-              key: postgresql-password
+              name: {{ .Values.global.postgresql.postgresqlExistingSecretName | default "postgres" }} 
+              key: {{ .Values.global.postgresql.postgresqlExistingSecretKey | default "postgres-password" }}
```
acohen4, line 69: the default here was previously `postgresql-password`

tybritten: yes, i just saw that

acohen4 (approved):
LGTM!

PR #7158, opened 2021-12-09
Use request context in s3 gateway.

PFedak:
The s3 gateway was keeping requests open even after the corresponding
http request had been closed, making it very easy to OOM on larger
requests.


acohen4 (approved):
LGTM

PR #7164, opened 2021-12-09
Have `pachctl list job` show the reason for killed jobs

Tryneus:
Noticed this when doing a review of some debug steps, since killed jobs should always have a reason listed.


src/server/pps/pretty/pretty.go
```diff
@@ -65,7 +65,7 @@ func PrintJobInfo(w io.Writer, jobInfo *ppsclient.JobInfo, fullTimestamps bool)
 	fmt.Fprintf(w, "%s\t", Progress(jobInfo))
 	fmt.Fprintf(w, "%s\t", pretty.Size(jobInfo.Stats.DownloadBytes))
 	fmt.Fprintf(w, "%s\t", pretty.Size(jobInfo.Stats.UploadBytes))
-	if jobInfo.State == ppsclient.JobState_JOB_FAILURE {
+	if jobInfo.State == ppsclient.JobState_JOB_FAILURE || jobInfo.State == ppsclient.JobState_JOB_KILLED {
```
PFedak, line 68: we could just always print a reason if it's nonempty, doesn't look like we fill it with anything by default

PFedak (approved):
LGTM

PR #7167, opened 2021-12-10
Collect pipeline and job history in debug dump

brycemcanally:


PFedak (approved):
LGTM

I guess we can have people set the "limit" option if getting all the jobs starts to impact things

PR #7171, opened 2021-12-11
Expose Pach Cluster Role Binding config in helm

acohen4:


chainlink (approved):

chainlink:
I still think we should rename this to avoid confusion with kubernetes RBAC

acohen4:
> I still think we should rename this to avoid confusion with kubernetes RBAC

Yeah I agree the naming is confusing but don't know the value of changing it given that it's not very user facing and we comment it's meaning.
Once we add config-pod to the main pachyderm repo, it will be a little easier to do (in 1 PR / release)

chainlink:
Once it's in the helm config it's alot more difficult to change (maintaining backward compatibility)

PR #7179, opened 2021-12-13
K8S Tolerations should be array's not objects

acohen4:


chainlink (approved):
Perhaps we should add a test here?

PR #7180, opened 2021-12-13
CheckStorage rpc and chunk layer integrity checking

brendoncarroll:
Adds `CheckStorage` rpc to the PFS API, and a method `Check` to the chunk storage layer.  Check iterates through chunk entries and checks if a corresponding chunk exists.

Refactors `chunk.Client.Create` to be more careful about returning values from closures passed to `dbutil.WithTx`

Ideally, this would be run async as a background job, and even more ideally: automatically on a schedule, but until we tackle the the async jobs for import/export we won't have a good solution for that.

TODO:
- [x]  There is a case where garbage collection can delete the object between when we list it and when we check object storage.  This is somewhat mitigated with the predicate `tombstone = false`, but it is still possible.  On not-found errors, we should probably re-check the entries table, and if it's not there anymore, consider it okay.


src/internal/testpachd/mock_pachd.go
```diff
@@ -777,6 +781,12 @@ func (api *pfsServerAPI) RunLoadTestDefault(ctx context.Context, req *types.Empt
 	}
 	return nil, errors.Errorf("unhandled pachd mock pfs.RunLoadTestDefault")
 }
+func (api *pfsServerAPI) CheckStorage(ctx context.Context, req *pfs.CheckStorageRequest) (*pfs.CheckStorageResponse, error) {
+	if api.mock.CheckStorage.handler != nil {
+		return api.mock.CheckStorage.handler(ctx, req)
+	}
+	return nil, errors.Errorf("unhandled pachd mock pfs.RunLoadTestDefault")
```
jrockway, line 788: I think this error message is copied and pasted from the above handler?

jrockway (approved):
Very nice.  We have health checks that run periodically in Hub and could add this there.

src/client/transaction.go
```diff
@@ -419,6 +419,9 @@ func (c *pfsBuilderClient) RunLoadTest(ctx context.Context, req *pfs.RunLoadTest
 func (c *pfsBuilderClient) RunLoadTestDefault(ctx context.Context, req *types.Empty, opts ...grpc.CallOption) (*pfs.RunLoadTestResponse, error) {
 	return nil, unsupportedError("RunLoadTestDefault")
 }
+func (c *pfsBuilderClient) CheckStorage(ctx context.Context, req *pfs.CheckStorageRequest, opts ...grpc.CallOption) (*pfs.CheckStorageResponse, error) {
+	return nil, unsupportedError("RunLoadTestDefault")
```
brycemcanally, line 423: `RunLoadTestDefault` -> `CheckStorage`

src/internal/obj/local_client.go
```diff
@@ -150,7 +150,7 @@ func (c *fsClient) transformError(err error, name string) error {
 		return nil
 	}
 	if os.IsNotExist(err) || strings.HasSuffix(err.Error(), ": no such file or directory") {
-		return pacherr.NewNotExist(c.rootDir, name)
+		return pacherr.NewNotExist(c.BucketURL().String(), name)
```
brycemcanally, line 153: :+1:

src/internal/storage/chunk/client.go
```diff
@@ -50,14 +50,41 @@ func (c *trackedClient) Create(ctx context.Context, md Metadata, chunkData []byt
...
 	chunkTID := chunkID.TrackerID()
-	var needUpload bool
-	var gen uint64
-	if err := dbutil.WithTx(ctx, c.db, func(tx *pachsql.Tx) error {
+	if err := dbutil.WithTx(ctx, c.db, func(tx *pachsql.Tx) (retErr error) {
+		// This function can be called multiple times.  In order to be sure
+		// that needsUpload and gen are set every run, we nil them here.
+		// This also protects this us from dbutil.WithTx dropping an error, if
+		// we only set these values before a `return nil`
+		needUpload, gen = nil, nil
```
brycemcanally, line 87: I would prefer us to just set the default value for the non-pointer types rather than having to dereference elsewhere. The pattern of returning default values or an error seems fine for non-pointer types and pointer types. Also, `protects this us from` -> `protects us from`. I'm not sure if we should really be coding around bugs in our own code, but ensuring the defaults are reset after a retry makes sense to me.

src/internal/storage/chunk/client.go
```diff
@@ -69,44 +96,51 @@ func (c *trackedClient) Create(ctx context.Context, md Metadata, chunkData []byt
...
-		return nil, err
+		return err
 	}
-	return chunkID, nil
+	affected, err := res.RowsAffected()
+	if err != nil {
+		return err
+	}
+	if affected < 1 {
+		return errors.Errorf("no chunk entry for object post upload: chunk=%v gen=%v", chunkID, gen)
```
brycemcanally, line 134: It seems like this error and the one below should have the same outcome. Either both return an error or panic. These would be pretty serious errors so maybe a panic makes sense, but taking down the whole process may be extreme.

src/internal/storage/chunk/client.go
```diff
@@ -132,6 +166,54 @@ func (c *trackedClient) Close() error {
...
+	}
+	if last == nil {
+		return errors.Errorf("no entries for chunk %v", id)
+	}
+	return nil
+}
+
+// CheckEntries runs an integrity check on the objects in object storage.
+// It lists through chunks with IDs >= first, lexicographically.
+func (c *trackedClient) CheckEntries(ctx context.Context, first []byte, limit int, readChunks bool) (last ID, _ error) {
```
brycemcanally, line 183: Wouldn't this produce an error if a chunk was correctly deleted between the query and the check? It seems like it should requery after an existence error to ensure that it wasn't deleted from the chunk table. Also, why is the iteration exposed through multiple layers like this (the first / last stuff). Is the concern about a long running transaction? If so, I'm of the opinion that it is fine to have a few things in the system that have longer running transactions (such as GC) in the interest of simplicity. If the concern is performance and the results becoming stale, this might be another use case for the task service similar to GC.

brycemcanally: Lol, just realized the extra check is the TODO in the comment. My bad. The iteration strategy question still stands.

src/internal/storage/chunk/client.go
brendoncarroll: This first error (line 134) could happen if there was a long pause when uploading a chunk.  It shouldn't take the whole process down.  Affecting more than one, would be a logic error.  It means postgres has an incorrect primary key, or something serious enough to want to prevent any database activity.

src/internal/storage/chunk/client.go
brendoncarroll: I wanted to capture the interactions with the chunk table in one place. Since these entries can be in flux it doesn't really make sense to hold open a connection.  It's better to do a few rows at a time, then get a few more.   We can make the iteration at the next layer up more complicated, or add parallelism.

src/internal/storage/chunk/client.go
brendoncarroll: The default value for needUpload is false, which is a little dangerous.  Panicking on a nil dereference is preferrable to silently skipping a chunk upload.  And I agree generally about not being defensive against our own code, but it would be particularly bad to have an error in this section.

brycemcanally: I don't think we should do this at all, even here where we want to be extra safe. I think we should stick to, and rely on, the pattern of nil error means the results are valid and a non-nil error means the results aren't. The decreased readability is worse than the possibility that we will catch the caller using the results with a non-nil error imo (since I think it is really unlikely to happen).

src/internal/storage/chunk/storage.go
```diff
@@ -81,3 +82,35 @@ func (s *Storage) List(ctx context.Context, cb func(id ID) error) error {
 func (s *Storage) NewDeleter() track.Deleter {
 	return &deleter{}
 }
+
+// Check runs an integrity check on the objects in object storage.
+// It will check objects for chunks with IDs in the range [first, last)
+// As a special case: if len(end) == 0 then it is ignored.
+func (s *Storage) Check(ctx context.Context, begin, end []byte, readChunks bool) (int, error) {
```
brycemcanally, line 89: It seems like this should expose checking one or all chunks rather than a range. What is the use of checking a range of chunks? Also, we can check past the end id with the current implementation.

brendoncarroll: So that this work can be parallelized.   We can use the Task Service at the PFS level to make multiple calls to `Check`.  For now though, it provides a way to resume manually if something goes wrong.

As long as the batch size is reasonably small I don't think the overrun is that big of a deal.  It's not like it mutates anything.

brycemcanally: Okay, that makes sense. Does it still make sense to batch under the hood in that case? It seems like we should have a function that shards the chunks similar to the way we shard file sets and the workers would call this check function with those batches. Also it seems like it would make sense for the exported function to operate on chunk.IDs.

src/internal/storage/chunk/client.go
brendoncarroll: > The decreased readability
This seems like a made up reason.  There are reasons not like this, but I don't see a case for readability being one of them.  The structure of the code and names are totally unchanged, there is a single additional assignment to nil, and an additional `*` when the variable is read.  It wouldn't impact the speed at which the average developer could traverse this area or understand what's going on.  I'd agree that it's inefficient to pass around a pointer to a smaller than pointer object, and perhaps not idiomatic Go.

What about changing this to a `skipUpload` boolean so that the default false is less dangerous, and removing the pointers?

src/internal/storage/chunk/storage.go
brendoncarroll: For the sharding I was thinking we would generate distinct prefixes in a loop.  First `[]byte{0}` then `[]byte{1}` etc. all the way up to `[]byte{255}`.  In this case begin and end would not be valid chunk.IDs, since they are not long enough.  I guess we could also pad them with a suffix of zeros.  At some point we should change ID to `type ID = pachhash.Output`

src/internal/storage/chunk/client.go
brycemcanally: The not idiomatic Go part is why it is less readable imo. Using a pointer to a builtin type usually makes me have to consider how it is set (when the default isn't valid) and where it could be referenced, plus the additional step to actually access the value. I'm not convinced inverting the value is much safer because if the gen is incorrect then the upload will screw things up. I think we shouldn't be relying on setting the results in a particularly way to prevent the client from using them when they shouldn't. Setting the error should be enough.

brycemcanally (approved):
lgtm

PR #7183, opened 2021-12-15
Configure Enterprise secrets using k8s secret refs 

acohen4:
The following secrets can now be configured using secret refs
- pachd.oauthClientSecret 
- pachd.rootToken 
- pachd.enterpriseRootToken 
- pachd.enterpriseSecret
- pachd.enterpriseLicenseKey
- oidc.upstreamIDPs


chainlink (changes requested):
I think perhaps we're overcomplicating things here. Many charts using the `existingSecret` convention. As an example `.Values.global.postgresql.existingSecret`  https://docs.bitnami.com/kubernetes/infrastructure/apache-airflow/administration/use-existing-secrets/ 
Further, I think we're ok to specify a static key in the secret here.

etc/helm/pachyderm/templates/pachd/config-job.yaml
```diff
@@ -22,7 +22,7 @@ spec:
     spec:
       containers:
       - name: config-pod
-        image: pachyderm/config-pod:0.4
+        image: config-pod:test-4
```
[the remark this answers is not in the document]
acohen4: for sure. As soon as the other PR ships I'll update this

etc/helm/pachyderm/templates/pachd/config-job.yaml
```diff
@@ -39,6 +39,48 @@ spec:
...
+        {{- end }}
+        {{- if .Values.pachd.enterpriseLicenseKeySecretName }}
+        - name: ENTERPRISE_LICENSE_KEY
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Values.pachd.enterpriseLicenseKeySecretName }}
+              key: enterpriseLicenseKey
+        {{- end }}   
+        {{- if .Values.oidc.upstreamIDPsSecretName }}
+        - name: UPSTREAM_IDPS
```
chainlink, line 78: Isn't this an array? What would a configuration here look like?

etc/helm/pachyderm/templates/pachd/config-job.yaml
```diff
@@ -39,6 +39,48 @@ spec:
         - name: PACH_ADDR
           value: "grpc://pach-enterprise:1650"
         {{- end }}
+        {{- if .Values.pachd.rootTokenSecretName }}    
+        - name: ROOT_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Values.pachd.rootTokenSecretName }}
+              key: rootToken
```
chainlink, line 47: The existing format for `existingSecret` keys (from postgres anyway) is dashed instead of camel case (ie. `root-token`)  The other option here is what we use for the other secrets (capitalized, so it's easier to pull in as `envFrom` My preference would be dashed. Sorry to be a stickler here, but i think the consistency is important (having all existing secrets follow one format anyway)

chainlink (changes requested):
Couple of comments

etc/helm/pachyderm/templates/pachd/config-job.yaml
acohen4: configuration would look the same as specified in the values.yaml for oidc.upstreamIDPs. It's a YAML array. Since it all gets encoded into bytes for the secret, we just deserialize it in the config-pod into YAML when it's read

chainlink (approved):

etc/helm/pachyderm/templates/pachd/config-job.yaml
chainlink: We may want to test this, (I'm not sure new lines and env vars are going to play nicely here?)

acohen4: Tested this locally 👍

PR #7185, opened 2021-12-15
Remove port_forwarders from pach config if port-forward process is dead

msteffen:


src/client/client.go
```diff
@@ -545,6 +546,30 @@ func NewEnterpriseClientOnUserMachine(prefix string, options ...Option) (*APICli
...
+		}
+		// If old port-forward pid isn't "pachctl port-forward", it's dead.
+		if pfname, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/cmdline", pid)); err == nil {
+			if !bytes.HasPrefix(pfname, []byte{'p', 'a', 'c', 'h', 'c', 't', 'l'}) {
+				return false
+			}
+		} else {
+			log.Errorf("Can't read /proc/%d/cmdline to establish whether old port-forward process is alive: %v", pid, err)
+			return true // assume alive
+		}
```
PFedak, line 568: I looked at what an OS-independent version of this would look like, and it's probably not worth it. `os.FindProcess` only errors on not exist on Windows, and there's not a straightforward way to get the executable name, anyway.

src/client/client.go
```diff
@@ -545,6 +546,30 @@ func NewEnterpriseClientOnUserMachine(prefix string, options ...Option) (*APICli
...
+		// If old port-forward pid no longer exists at all, it's dead.
+		if _, err := os.Stat(fmt.Sprintf("/proc/%d", pid)); errors.Is(err, os.ErrNotExist) {
+			return false
+		} else if err != nil {
+			log.Errorf("Can't stat /proc/%d to establish whether old port-forward process is alive: %v", pid, err)
+			return true // assume alive
+		}
+		// If old port-forward pid isn't "pachctl port-forward", it's dead.
+		if pfname, err := ioutil.ReadFile(fmt.Sprintf("/proc/%d/cmdline", pid)); err == nil {
+			if !bytes.HasPrefix(pfname, []byte{'p', 'a', 'c', 'h', 'c', 't', 'l'}) {
```
PFedak, line 562:
```suggestion
			if !strings.HasPrefix(string(pfname), "pachctl") {
```
looks a little nicer to me, maybe there's some reason to avoid it

src/client/client.go
```diff
@@ -558,13 +583,20 @@ func newOnUserMachine(cfg *config.Config, context *config.Context, contextName,
...
-		if ok {
-			log.Debugf("Connecting to explicitly port forwarded pachd instance on port %d", pachdLocalPort)
-			pachdAddress = &grpcutil.PachdAddress{
-				Secured: false,
-				Host:    "localhost",
-				Port:    uint16(pachdLocalPort),
+		if !CheckPortForwardLiveness(context) {
+			context.PortForwarders = nil
+			if err = cfg.Write(); err != nil {
+				return nil, errors.Wrap(err, "could not write config to remove port-forward ports")
```
PFedak, line 589: part of me is a little nervous about blocking every command if you somehow make your config non-writable after the port-forwarder is killed, but it doesn't seem very likely

PFedak (approved):
LGTM, just the one thing I'm a little unsure of

src/client/client.go
msteffen: No real reason, LGTM

src/client/client.go
msteffen: This reservation makes sense. One alternative that occurred to me is that this code could just warn users that they need to restart `pachctl port-forward`, though an issue is that if `pachctl port-forward` also can't modify the config, then other `pachctl` commands will have no way to use it—they'll keep checking the old pid for liveness, keep re-discoving that the old port-forward process is dead, and keep doing internal port-forwarding even though a new port-forward process has come up as a replacement.

I'm not really sure what a good solution would be. I was planning to merge this PR if/when the tests are green, but I'd be happy to revise this here (if you know what to do) or in a followup PR if I merge before you respond—just let me know.

msteffen:
Just wanted to update this PR to say thatbI haven't forgotten about it, it's stuck on a nil dereference bug in CI. I probably won't be able to bathe it for another week and a half or so due to the 2.1.x code freeze, but my plan is to merge it once that's over

PR #7186, opened 2021-12-15
Annotate which errors are caused by users and don't log stack traces for them

Tryneus:
This PR adds a new type `errors.UserError` that error types can embed, which will be recognized by the RPC `Log` function to suppress logging stack traces.

### DIFF
--- a/src/internal/errors/errors.go
+++ b/src/internal/errors/errors.go
@@ -33,6 +33,20 @@ var (
 	WithStack = errors.WithStack
 )
 
+// UserError is a type used to indicate that an error is not an internal error
+// so that we can log it differently (i.e. do not log at the ERROR level with a
+// stack trace)
+type UserError struct{}
+
+func (err UserError) IsUserError() bool {
+	return true
+}
+
+// Interface for type-checking user errors
+type UserErrorer interface {
+	IsUserError() bool
+}
+
 // StackTrace is stack of Frames from innermost (newest) to outermost (oldest).
 type StackTrace = errors.StackTrace
 
--- a/src/internal/log/log.go
+++ b/src/internal/log/log.go
@@ -21,8 +21,8 @@ var reportMetricsOnce sync.Once
 
 // Logger is a helper for emitting our grpc API logs
 type Logger interface {
-	Log(request interface{}, response interface{}, err error, duration time.Duration)
-	LogAtLevelFromDepth(request interface{}, response interface{}, err error, duration time.Duration, level logrus.Level, depth int)
+	Log(request interface{}, response interface{}, err interface{}, duration time.Duration)
+	LogAtLevelFromDepth(request interface{}, response interface{}, err interface{}, duration time.Duration, level logrus.Level, depth int)
 }
 
 type logger struct {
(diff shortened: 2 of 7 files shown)


src/internal/errors/errors.go
```diff
@@ -33,6 +33,20 @@ var (
...
+// so that we can log it differently (i.e. do not log at the ERROR level with a
+// stack trace)
+type UserError struct{}
+
+func (err UserError) IsUserError() bool {
+	return true
+}
+
+// Interface for type-checking user errors
+type UserErrorer interface {
```
Tryneus, line 46: I really dislike this, but couldn't find another way to do this without making it more cumbersome on user code.

brendoncarroll: If you are just relying on the embedding to do a type switch.  You could probably make the interface method private `isUserError()` and make it return nothing.  `UserError` is in the same package, so it can implement it.  Otherwise, when would someone ever implement IsUserError and have it return false? Wouldn't they just not implement it?

src/internal/errors/errors.go
```diff
@@ -33,6 +33,20 @@ var (
 	WithStack = errors.WithStack
 )
 
+// UserError is a type used to indicate that an error is not an internal error
+// so that we can log it differently (i.e. do not log at the ERROR level with a
+// stack trace)
+type UserError struct{}
```
brendoncarroll, line 39: There is another package `pacherr` which has some of the structured error stuff.  I have been using that for structured error handling, while trying to phase out the string matching error handling in this package.

Separately: What do you think about having a `Hint` field here, or something that allows us to communicate helpful information to the user about how to resolve the error?

src/internal/log/log.go
```diff
@@ -21,8 +21,8 @@ var reportMetricsOnce sync.Once
 
 // Logger is a helper for emitting our grpc API logs
 type Logger interface {
-	Log(request interface{}, response interface{}, err error, duration time.Duration)
-	LogAtLevelFromDepth(request interface{}, response interface{}, err error, duration time.Duration, level logrus.Level, depth int)
+	Log(request interface{}, response interface{}, err interface{}, duration time.Duration)
```
brendoncarroll, line 24: Why was this necessary?

src/internal/log/log.go
```diff
@@ -191,13 +191,17 @@ func (l *logger) LogAtLevelFromDepth(request interface{}, response interface{},
...
 	}
 	if err != nil {
-		// "err" itself might be a code or even an empty struct
-		fields["error"] = err.Error()
-		var frames []string
-		errors.ForEachStackFrame(err, func(frame errors.Frame) {
-			frames = append(frames, fmt.Sprintf("%+v", frame))
-		})
-		fields["stack"] = frames
+		if goerr, ok := err.(error); ok {
```
brendoncarroll, line 194: I don't think we should define our own type for this check, after looking this over.  We should check to see if it is a grpc status, or implements a `GRPCStatus() *status.Status` method.  If it doesn't, log it.  If it does, switch on the Code (which is like an HTTP status code), and don't log errors originating on the client.  By default we should log errors, we should only skip logging for specific codes.

Then when we want to stop an error from making it into the logs, we implement GRPCStatus for that error type.  That has the additional benefit of being able to inspect the errors using the grpc libraries.

src/internal/errors/errors.go
brendoncarroll: I don't think this additional type is the best approach, see my comment below

brendoncarroll:
These are the grpc codes.
https://pkg.go.dev/google.golang.org/grpc/codes

NotExists, and AlreadyExists probably capture most of the types in pfs.  InvalidArgument would probably work for anything else.

Tryneus:
Closing this PR in favor of a different approach.

PR #7206, opened 2021-12-21
Don't fail pipeline for transient database issues.

PFedak:
An error fetching pipeline info to create an "op" should only fail the
pipeline immediately if there's no hope of success on retry, which is
really only true of not found errors.


msteffen (approved):
LGTM

PR #7207, opened 2021-12-21
[FRON-223] feat(console): add helm values for telemetry opt out

jeffrifwald:
Related PR: https://github.com/pachyderm/haberdashery/pull/484


etc/helm/pachyderm/templates/console/deployment.yaml
```diff
@@ -48,6 +48,8 @@ spec:
           value: {{ include "pachyderm.issuerURI" . | quote}}
         - name: REACT_APP_RUNTIME_ISSUER_URI
           value: {{ include "pachyderm.reactAppRuntimeIssuerURI" . | quote}}
+        - name: REACT_APP_RUNTIME_DISABLE_TELEMETRY
+          value: {{ .Values.console.config.disableTelemetry | quote}}
```
jeffrifwald, line 52: Not sure if I need the `| quote` here for a boolean value, but I left it in.

chainlink (approved):

PR #7214, opened 2022-01-04
Pachctl mount server

lukemarsden:
Introduce `pachctl mount-server` subcommand which runs a small REST server which allows mounting and unmounting of repos in the specified mount directory dynamically.


src/server/pfs/fuse/fuse_test.go
```diff
@@ -279,7 +279,7 @@ func TestRepoOpts(t *testing.T) {
 			},
 		},
 		RepoOptions: map[string]*RepoOptions{
-			"repo1": {},
+			"repo1": {Repo: "repo1"},
```
msteffen, line 282: I don't actually know the fuse library's API that well, but if RepoOptions is part of the client's user-facing API, we may be obligated by semver to infer a default value for `Repo` if it's unset (in case an existing user is calling our client with `Repo` unset; they should be able to upgrade without it breaking)

src/server/pfs/fuse/loopback.go
```diff
@@ -44,10 +45,10 @@ type loopbackRoot struct {
...
 	c *client.APIClient
 
-	repoOpts map[string]*RepoOptions
-	branches map[string]string
-	commits  map[string]string
-	files    map[string]fileState
+	repoOpts map[string]*RepoOptions // key is mount name
+	branches map[string]string       // key is mount name
+	commits  map[string]string       // key is mount name
+	files    map[string]fileState    // key is {mount_name}/{path}
```
msteffen, line 51: :pray:

src/server/pfs/fuse/loopback.go
```diff
@@ -519,25 +520,21 @@ func newLoopbackRoot(root, target string, c *client.APIClient, opts *Options) (*
...
-	for _, ri := range ris {
-		if len(ro) > 0 && ro[ri.Repo.Name] == nil {
-			continue
-		}
-		p := n.repoPath(ri)
+	ros := n.root().repoOpts
+	// we only mount explicitly named repos now
+	for name := range ros {
+		p := n.namePath(name)
+		fmt.Printf("> MKDIR %s, %s\n", name, p)
```
msteffen, line 533: I'm not sure if this is leftover println debugging or if we want to keep these in, but if it's the latter, it might make sense to use logrus here, and log these at DEBUG? WDYT?

msteffen (commented):
A few initial comments

src/server/pfs/fuse/fuse_test.go
lukemarsden: I think it's just consumed by the CLI for `pachctl mount` which has been updated.

src/server/pfs/fuse/loopback.go
lukemarsden: removed

src/server/pfs/cmds/mount_unix.go
```diff
@@ -108,6 +113,37 @@ func mountCmds() []*cobra.Command {
...
 	commands = append(commands, cmdutil.CreateAlias(mount, "mount"))
 
+	var daemonize bool
+	var mountDir string
+	var socket string
+	var logFile string
+	mountServer := &cobra.Command{
+		Use:   "{{alias}}",
+		Short: "Start a mount server for controlling FUSE mounts via a local gRPC API.",
+		Long:  "Starts a gRPC mount server on the given socket, logging to the given log-file, daemonizing and exiting once it is active if requested, otherwise running in the foreground and logging to stdout.\n\nSee TODO:<link to docs> for documentation on available methods on the API.",
```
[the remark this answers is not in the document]
lukemarsden: I've removed it. I don't think we need to publish the REST spec for this API in the public pachyderm docs just yet, as this feature is mostly intended for our own use. Maybe in a future release though.

msteffen (commented):
Few more comments, but I'm still working through the meat. Thanks for being so patient!

.circleci/config.yml
```diff
@@ -183,6 +183,7 @@ workflows:
               - AUTH
               - ENTERPRISE
               - PFS
+              - FUSE
```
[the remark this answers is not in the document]
lukemarsden: PFS tests take 7m4s
FUSE tests take 3m20s
Slowest test bucket seems to be PPS6 at 23m49s!

So merging them won't make overall test cycle any slower.

So yeah, we should merge them.

src/server/pfs/fuse/loopback.go
```diff
@@ -649,19 +659,20 @@ func (n *loopbackNode) commit(repo string) (string, error) {
 	// You can access branches that don't exist, which allows you to create
 	// branches through the fuse mount.
 	if errutil.IsNotFoundError(err) {
-		n.root().commits[repo] = ""
+		n.root().commits[name] = ""
 		return "", nil
 	}
-	n.root().commits[repo] = bi.Head.ID
+	n.root().commits[name] = bi.Head.ID
+	fmt.Printf("*** COMMIT END commits: %+v\n", n.root().commits)
```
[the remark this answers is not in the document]
lukemarsden: fixed, thanks!

src/server/pfs/fuse/loopback.go
```diff
@@ -621,25 +623,33 @@ func (n *loopbackNode) trimTargetPath(path string) string {
...
 		defer n.root().mu.Unlock()
-		commit, ok := n.root().commits[repo]
+		commit, ok := n.root().commits[name]
 		return commit, ok
 	}(); ok {
 		return commit, nil
 	}
-	branch := n.root().branch(repo)
-	bi, err := n.root().c.InspectBranch(repo, branch)
+	fmt.Printf("*** COMMIT repoOpts: %+v\n", n.root().repoOpts)
```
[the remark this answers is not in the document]
lukemarsden: fixed, thanks!

src/server/pfs/fuse/loopback.go
```diff
@@ -566,21 +558,31 @@ func (n *loopbackNode) download(path string, state fileState) (retErr error) {
...
+	commit, err := n.commit(name)
 	if err != nil {
 		return err
 	}
 	if commit == "" {
 		return nil
 	}
-	if err := n.c().ListFile(client.NewCommit(parts[0], branch, commit), pathpkg.Join(parts[1:]...), func(fi *pfs.FileInfo) (retErr error) {
+	// XXX translate name -> repo
+	fmt.Printf("*** COMMIT repoOpts: %+v\n", n.root().repoOpts)
```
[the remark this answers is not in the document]
lukemarsden: fixed, thanks!

.circleci/config.yml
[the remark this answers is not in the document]
lukemarsden: done!

msteffen (commented):
Few more comments. Mostly focused on `server.go`

src/server/pfs/fuse/fuse.go
```diff
@@ -13,12 +13,48 @@ import (
...
+			// different named paths.
+			branch := "master" // XXX is the old implementation always defaulting to master?
+			bi, err := c.InspectBranch(ri.Repo.Name, branch)
+			if err != nil && !errutil.IsNotFoundError(err) {
+				return err
+			}
+			isOutputBranch := bi != nil && len(bi.Provenance) > 0
+			write := opts.Write
+			if isOutputBranch {
+				write = false
```
msteffen, line 49: could the above be just `write := opts.Write && (bi == nil || len(bi.Provenance) == 0)` (or `write := opts.Write && !isOutputBranch)` if you prefer the temporary variable)?

src/server/pfs/fuse/options.go
```diff
@@ -82,17 +86,17 @@ func (o *Options) validate(c *client.APIClient) error {
 	if o == nil {
 		return nil
 	}
-	for repo, opts := range o.RepoOptions {
+	for _, opts := range o.RepoOptions {
 		if opts.Write {
 			if uuid.IsUUIDWithoutDashes(opts.Branch) {
-				return errors.Errorf("can't mount commit %s@%s in Write mode (mount a branch instead)", repo, opts.Branch)
+				return errors.Errorf("can't mount commit %s@%s %s in Write mode (mount a branch instead)", opts.Repo, opts.Branch)
```
msteffen, line 92: I think there might be a printf bug here (there are three `%s` tokens in the format string, but only two arguments: `opts.Repo` and `opts.Branch`). I think you meant to add the name as a printf argument but never quite got to it.

src/server/pfs/fuse/options.go
```diff
@@ -82,17 +86,17 @@ func (o *Options) validate(c *client.APIClient) error {
...
 			}
-			bi, err := c.InspectBranch(repo, opts.Branch)
+			bi, err := c.InspectBranch(opts.Repo, opts.Branch)
 			if err != nil && !errutil.IsNotFoundError(err) {
 				return err
 			}
 			if bi != nil && len(bi.Provenance) > 0 {
-				return errors.Errorf("can't mount branch %s@%s in Write mode because it's an output branch", repo, opts.Branch)
+				return errors.Errorf("can't mount branch %s@%s as %s in Write mode because it's an output branch", opts.Repo, opts.Branch)
```
msteffen, line 99: Same comment as above, I there's a printf bug here

src/server/pfs/fuse/loopback.go
```diff
@@ -621,25 +623,32 @@ func (n *loopbackNode) trimTargetPath(path string) string {
...
 		return commit, ok
 	}(); ok {
 		return commit, nil
 	}
-	branch := n.root().branch(repo)
-	bi, err := n.root().c.InspectBranch(repo, branch)
+	ro, ok := n.root().repoOpts[name]
+	if !ok {
+		// not ready or something?
+		// XXX should this be an error?
```
msteffen, line 646: I think it would make sense to return an error (or maybe even just log an error) here.

src/server/pfs/fuse/loopback.go
```diff
@@ -566,21 +560,29 @@ func (n *loopbackNode) download(path string, state fileState) (retErr error) {
...
 		return err
 	}
 	if commit == "" {
 		return nil
 	}
-	if err := n.c().ListFile(client.NewCommit(parts[0], branch, commit), pathpkg.Join(parts[1:]...), func(fi *pfs.FileInfo) (retErr error) {
+	ro, ok := n.root().repoOpts[name]
+	if !ok {
+		// not ready or something?
+		// XXX should this be an error?
```
msteffen, line 577: Might make sense to either return or log an error

src/server/pfs/fuse/loopback.go
```diff
@@ -566,21 +560,29 @@ func (n *loopbackNode) download(path string, state fileState) (retErr error) {
...
+		return nil
+	}
+	repoName := ro.Repo
+	if err := n.c().ListFile(client.NewCommit(repoName, branch, commit), pathpkg.Join(parts[1:]...), func(fi *pfs.FileInfo) (retErr error) {
```
msteffen, line 581: I really did not realize that pachctl mount worked like this. Good to know...

src/server/pfs/fuse/server.go
```diff
@@ -0,0 +1,850 @@
...
+// http://www.timschmelmer.com/2013/08/state-machines-with-state-functions-in.html
+
+// setup
+
+type StateFn func(*MountStateMachine) StateFn
+
+func (m *MountStateMachine) transitionedTo(state, status string) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	fmt.Printf("[%s] %s -> %s\n", m.MountKey, m.State, state)
```
msteffen, line 544: stray?

src/server/pfs/fuse/server.go
```diff
@@ -0,0 +1,850 @@
...
+	func() {
+		m.manager.mu.Lock()
+		defer m.manager.mu.Unlock()
+		m.manager.root.repoOpts[m.MountState.Name] = &RepoOptions{
+			Repo:   m.MountKey.Repo,
+			Branch: m.MountKey.Branch,
+			Write:  m.Mode == "rw",
+		}
+		m.manager.root.branches[m.Name] = m.MountKey.Branch
+		fmt.Printf("*** MOUNTINGSTATE repoOpts: %+v\n", m.manager.root.repoOpts)
```
msteffen, line 611: stray?

src/server/pfs/fuse/server.go
```diff
@@ -0,0 +1,850 @@
...
+	Mode       string   `json:"mode"`       // "ro", "rw", or "" if unknown/unspecified. written by client
+	State      string   `json:"state"`      // "unmounted", "mounting", "mounted", "pushing", "unmounted", "error". written by fsm
+	Status     string   `json:"status"`     // human readable string with additional info wrt State, e.g. an error message for the error state. written by fsm
+	Mountpoint string   `json:"mountpoint"` // where on the filesystem it's mounted. written by fsm. can also be derived from {MountDir}/{Name}
+}
+
+type MountStateMachine struct {
+	MountState
+	manager   *MountManager
+	mu        sync.Mutex
```
msteffen, line 478: I'm still assembling my mental image of how this all works, but I don't totally understand the purpose of this mutex. My understanding is that `MaybeStartFsm` uses `manager.mu` and `manager.States` to ensure that there's one goro (and one instance of `Run()`) running for a given `name` at a time. Does anything besides the state functions write to `MountStateMachine`? Could two state functions ever write to the same `MountStateMachine` concurrently, despite `manager.States` guarding `Run()`? Or is the intent to prevent reading partial writes?

msteffen (commented):
One real question and several nits. I think once all the stray debug lines and such are cleaned up this will be ready to merge

src/server/pfs/fuse/server.go
lukemarsden: There's a mutex for the global MountManager and a mutex for each MountStateMachine. The MountManager mutex is to guard access to the `mm.States` map (since go maps can't be accessed concurrently).

For the per mount MountStateMachine mutex:

Different parts of `MountStateMachine.MountState` are written by different parts of the system, as documented in the comments in the `MountState` struct, e.g.
```
	Mode       string   `json:"mode"`       // "ro", "rw", or "" if unknown/unspecified. written by client
	State      string   `json:"state"`      // "unmounted", "mounting", "mounted", "pushing", "unmounted", "error". written by fsm
```
(where fsm means finite state machine)

I find it's always useful to document which agent in the system is responsible for writing to a given variable, and generally not having two agents both try and write to the same.

So, there might be different agents writing to different parts of the MountState from different goroutines, so holding a mutex on writing to those values seems wise to avoid overlapping executions of code that wants to modify the state. We could probably optimize this with an RWLock but I don't think it's performance critical code: we always hold the locks for a few milliseconds at most probably.

src/server/pfs/fuse/server.go
lukemarsden: This one's intentional, it's useful to be able to see the behavior of the system in terms of which state transitions it's performing.

src/server/pfs/fuse/server.go
lukemarsden: fixed

src/server/pfs/fuse/loopback.go
lukemarsden: fixed

src/server/pfs/fuse/loopback.go
lukemarsden: fixed

src/server/pfs/fuse/options.go
lukemarsden: good catch, I also forgot to thread the mount name through into the RepoOpts struct, which I've done now

src/server/pfs/fuse/options.go
lukemarsden: fixed, thanks

src/server/pfs/fuse/fuse.go
lukemarsden: yeah, i just thought this was easier to read

src/server/pfs/fuse/server.go
```diff
@@ -0,0 +1,850 @@
...
+	fmt.Println("Sending mount request...")
+	mm.States[key].requests <- Request{
+		Mount:  true,
+		Repo:   key.Repo,
+		Branch: key.Branch,
+		Commit: key.Commit,
+		Name:   name,
+		Mode:   mode,
+	}
+	fmt.Println("sent mount request!")
```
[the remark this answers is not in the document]
lukemarsden: fixed

src/server/pfs/fuse/server.go
```diff
@@ -0,0 +1,850 @@
...
+	defer mm.mu.Unlock()
+
+	// fetch list of available repos & branches from pachyderm, and overlay that
+	// with their mount states
+	lr := map[string]RepoResponse{}
+	repos, err := mm.Client.ListRepo()
+	if err != nil {
+		return lr, err
+	}
+	fmt.Printf("repos: %+v\n", repos)
```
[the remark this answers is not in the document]
lukemarsden: fixed

src/server/pfs/fuse/server_test.go
```diff
@@ -0,0 +1,187 @@
...
+			MountDir: dir,
+		}
+	}
+	if sopts.Unmount == nil {
+		sopts.Unmount = make(chan struct{})
+	}
+	unmounted := make(chan struct{})
+	var mountErr error
+	defer func() {
+		fmt.Printf("=== IN DEFER FUNC, CLOSING ===\n")
```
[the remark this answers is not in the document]
lukemarsden: fixed

src/server/pfs/fuse/server.go
```diff
@@ -0,0 +1,840 @@
...
+}
+
+func NewMountManager(c *client.APIClient, target string, opts *Options) (ret *MountManager, retErr error) {
+	if err := opts.validate(c); err != nil {
+		return nil, err
+	}
+	commits := make(map[string]string)
+	for repo, branch := range opts.getBranches() {
+		if uuid.IsUUIDWithoutDashes(branch) {
+			commits[repo] = branch
```
smalyala, line 264: What is `commits` being used for? There doesn't seem to be a reference to it afterwards.

src/server/pfs/fuse/server.go
msteffen: Ah, I didn't realize `Mode` could change after startup, I'll try to understand that part a bit better

lukemarsden: It doesn't right now but it could, e.g a "remount"

src/server/pfs/fuse/server.go
msteffen: How hard would it be to switch to logrus for these?

src/server/pfs/fuse/server.go
msteffen: Yeah I can't quickly tell either?

msteffen (approved):
A few more comments, but LGTM once those are addressed.

I think the next steps are to update the notebook backend to the mount server, and then break the mount server into its own binary for releases?

Here's a more specific list of remaining issues for this code (deposited here for my future self to file, subject to your assent):
- [ ] automating the release of the mount server binary
- [ ] blocking filesystem operations during upload
- [ ] moving to logrus to match other core pach logging
- [ ] test for cp issue (a.k.a. ls issue)

src/server/pfs/fuse/server_test.go
```diff
@@ -0,0 +1,187 @@
...
+	err := env.PachClient.PutFile(commit, "dir/file1", strings.NewReader("foo"))
+	require.NoError(t, err)
+	err = env.PachClient.PutFile(commit, "dir/file2", strings.NewReader("foo"))
+	require.NoError(t, err)
+	withServerMount(t, env.PachClient, nil, func(mountPoint string) {
+
+		_, err := put("repos/repo/master/_mount?name=repo&mode=ro")
+		require.NoError(t, err)
+
+		fmt.Printf("=====> MOUNTPOINT IS %s\n", mountPoint)
```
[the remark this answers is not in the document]
msteffen: I think this got overlooked in the last pass Ave still needs removing?

src/server/pfs/fuse/server.go
lukemarsden: done! the output looks pretty now 😁 

image

lukemarsden:
Thanks! Those items look good to me. I'll move the list into the [scratchpad doc](https://docs.google.com/document/d/17gYGN3dwFRnw97KreCJIFSH8fVx9q2a1tew_mctl3Xg/edit).

I've finished addressing all the review feedback. Ready to merge?

PR #7217, opened 2022-01-04
Setting the Worker RC's init container security context to the same as the user + storage containers'

acohen4:
the init container is a go script that moves the worker and pachctl binaries to be accessible from the user container


chainlink:
I think we want to keep pach security context and user security context separate?

acohen4:
> I think we want to keep pach security context and user security context separate?

I think this is a simplification we can make. It probably doesn't make sense to change this for something that's going out in a patch if not necessary. 

I'll close this PR. Opening this one as an alternative: https://github.com/pachyderm/pachyderm/pull/7218

PR #7218, opened 2022-01-05
Unset Worker SecurityContexts when pachd.securityContext.enabled=false

acohen4:


brendoncarroll (approved):
LGTM

etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -152,6 +152,10 @@ spec:
         - name: NO_PROXY
           value:  {{.Values.global.noProxy}}
         {{ end }}
+        {{- if not .Values.pachd.securityContext.enabled }}
```
chainlink, line 155: This feels a little weird, I think we should be pulling using the .enabled value as the value of the env var itself

chainlink (changes requested):
See note

PR #7219, opened 2022-01-05
OpenShift in CI

acohen4:


msteffen:
Could we potentially use this to test github.com/pachyderm/openshift-operator?

acohen4:
> Could we potentially use this to test github.com/pachyderm/openshift-operator?

absolutely 👍

PR #7225, opened 2022-01-05
Use published helm manifest in pre-upgrade deployment of the Upgrade Test

acohen4:
Remove the git-magic happening in the Upgrade test by first deploying a published pachyderm version

### DIFF
--- a/etc/testing/circle/launch-published.sh
+++ b/etc/testing/circle/launch-published.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -ex
+
+# shellcheck disable=SC1090
+source "$(dirname "$0")/env.sh"
+
+# deploy object storage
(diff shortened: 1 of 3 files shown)


etc/testing/circle/launch-published.sh
```diff
@@ -0,0 +1,22 @@
...
+# shellcheck disable=SC1090
+source "$(dirname "$0")/env.sh"
+
+# deploy object storage
+kubectl apply -f etc/testing/minio.yaml
+
+helm repo add pach https://helm.pachyderm.com
+helm repo update
+
+VERSION="2.0.4"
```
acohen4, line 14: for the meantime this version is hard-coded. The intention is that in the near future we will make this test run in a go script against a series of published versions

PFedak (approved):
LGTM

PR #7227, opened 2022-01-06
Pass postgres secret reference from pachd to workers

Tryneus:
If a user wants to bring their own postgres instance, they may specify `postgresqlExistingSecretName` and `postgresqlExistingSecretKey` when deploying the helm chart.  Unfortunately, PPS did not take this into account when generating the worker RCs for a pipeline, so workers would still be trying to connect using the default postgres key reference.  This PR fixes this by having the worker RC load the secret reference from the running `pachd` pod and passing it to workers.


acohen4 (approved):
LGTM

PR #7228, opened 2022-01-06
Add loki log collection to debug dump

brycemcanally:
This PR adds functionality for collecting logs in the debug dump from loki when it is deployed. The logs are collected / stored in the same file structure (each container has a well defined file path) as the direct Kubernetes logs we normally collect. This gives us a way to collect the logs in the debug dump format when the direct Kubernetes logs are inaccessible (lost due to autoscaling or rotation due to the volume). This PR also includes a few additional logs in the work layout step of job processing.


src/server/debug/server/server.go
```diff
@@ -575,6 +583,20 @@ func (s *debugServer) collectLogs(tw *tar.Writer, pod, container string, prefix
 	}, prefix...)
 }
 
+func (s *debugServer) collectLogsLoki(tw *tar.Writer, pod, container string, prefix ...string) error {
+	return collectDebugFile(tw, "logs-loki", "txt", func(w io.Writer) error {
+		queryStr := `{pod="` + pod
```
PFedak, line 588: I was worried for a bit that we might have to worry about special characters but `=` does seem to just be strings and not regex

src/server/debug/server/server.go
```diff
@@ -601,7 +623,96 @@ func (s *debugServer) collectPipelineDumpFunc(pachClient *client.APIClient, limi
...
+				}
+				if err := cb(stream.Labels, entry.Line); err != nil {
+					return err
+				}
+			}
+		}
+		if numLogs < maxLogs {
+			return nil
+		}
+		start = start.Add(time.Nanosecond)
```
PFedak, line 715: is the only difference between this and `lokiutil.QueryRange` the lack of sorting?

src/server/debug/server/server.go
```diff
@@ -601,7 +623,96 @@ func (s *debugServer) collectPipelineDumpFunc(pachClient *client.APIClient, limi
...
+				name = join(prefix[0], name)
+			}
+			return writeErrorFile(tw, err, name)
+		}
+		return nil
+	}
+}
+
+func (s *debugServer) forEachWorkerLoki(pipelineInfo *pps.PipelineInfo, cb func(string) error) error {
+	pods, err := s.getWorkerPodsLoki(pipelineInfo)
```
PFedak, line 652: am I understanding correctly that this reads through all the pipeline logs to find unique `pod` label values?

it seems like there should be some way to use the `GET /loki/api/v1/label/<name>/values` endpoint and select the ones that match the expected pod names for a pipeline, but maybe that's not worth it for now

PFedak (approved):
LGTM, most of my comments probably aren't worth delaying this PR

src/server/debug/server/server.go
```diff
@@ -575,6 +583,20 @@ func (s *debugServer) collectLogs(tw *tar.Writer, pod, container string, prefix
...
 }
 
+func (s *debugServer) collectLogsLoki(tw *tar.Writer, pod, container string, prefix ...string) error {
+	return collectDebugFile(tw, "logs-loki", "txt", func(w io.Writer) error {
+		queryStr := `{pod="` + pod
+		if container != "" {
+			queryStr += `", container="` + container
+		}
+		queryStr += `"}`
+		return s.queryLoki(queryStr, func(_ loki.LabelSet, line string) error {
```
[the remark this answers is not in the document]
brycemcanally: yeah, I'll add that.

src/server/debug/server/server.go
brycemcanally: Also the label sets in the callback.

src/server/debug/server/server.go
brycemcanally: Yeah, I didn't see a clear way to get the values based on the value of a different label. That endpoint gives values for a particular label.

PR #7231, opened 2022-01-07
Include usernames in gRPC logs.

PFedak:


jrockway (approved):
Good cleanup.  Having the context for these per-request logs is generally nice (when tracing is enabled, you can dump the span ID for example, and then correlate log messages with the Jaeger traces).

PR #7235, opened 2022-01-11
Pass postgres secret reference from pachd to workers [2.0.x]

Tryneus:
Cherry-pick #7227 into 2.0.x

### DIFF
--- a/src/server/pps/server/worker_rc.go
+++ b/src/server/pps/server/worker_rc.go
@@ -629,8 +629,8 @@ func (a *apiServer) getWorkerOptions(ctx context.Context, pipelineInfo *pps.Pipe
 	}
 
 	// Get the reference to the postgres secret used by the current pod
-	podName := pc.env.Config.PachdPodName
-	selfPodInfo, err := pc.env.KubeClient.CoreV1().Pods(pc.namespace).Get(ctx, podName, metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


src/server/pps/server/worker_rc.go
```diff
@@ -479,7 +475,7 @@ func hashAuthToken(token string) string {
 	return base64.RawURLEncoding.EncodeToString(h[:])
 }
 
-func (a *apiServer) getWorkerOptions(pipelineInfo *pps.PipelineInfo) (*workerOptions, error) {
+func (a *apiServer) getWorkerOptions(ctx context.Context, pipelineInfo *pps.PipelineInfo) (*workerOptions, error) {
```
PFedak, line 478: am I missing something or is the context unused? (I guess the k8s client uses some implied context?)

PFedak (approved):

src/server/pps/server/worker_rc.go
Tryneus: no, you're right, it's unused - this is just an artifact of the port from `master` (and part of why the cherry-pick wasn't trivial) - we've changed kubernetes client versions in `master`, and the new version uses contexts on all the calls.  This parameter is extraneous, but I'm inclined to just leave it here to reduce the delta between `master` and `2.0.x`.

PR #7236, opened 2022-01-11
Run upgrade tests in go

acohen4:
The intention here is to add some basic deployment test utilities we can build on, with the first client being system upgrade tests.

One point of difficulty for me on OSX is actually running a test locally that expects to connect with a NodePort on minikube. I tried to address this with port-forwarding in test, but that's still quite flaky


src/internal/testutil/deploy.go
```diff
@@ -0,0 +1,116 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kube "k8s.io/client-go/kubernetes"
+)
+
+const (
+	helmChartLocalPath     = "../../../etc/helm/pachyderm"
+	helmChartPublishedPath = "pach/pachyderm"
+	helmRelease            = "test-release"
+	ns                     = "default"
```
chainlink, line 27: You might want to use different namespaces instead of deleting (and waiting for PVCs to clear) between each run

chainlink (approved):
Overall I like this and it represents a nice "upgrade" to our upgrade tests. 

One nit I had is that it's a bit too tight in terms of passing functions to each other. It might be a bit easier to follow if you had:

```
funcA
funcB
funcC
```

Instead of
```
funcC(funcB(funcA())))
```

Maybe that's just me though

src/internal/testutil/deploy.go
```diff
@@ -0,0 +1,130 @@
...
+		SetStrValues: map[string]string{
+			"pachd.storage.minio.signature": "",
+			"pachd.storage.minio.secure":    "false",
+		},
+	}
+}
+
+// WaitUntilReady attempts to ping the database until the context is cancelled.
+// Progress information is written to log
+func WaitUntilReady(ctx context.Context, log *logrus.Logger, db *sqlx.DB) error {
```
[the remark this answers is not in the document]
acohen4: nope! removed this

src/internal/testutil/pach_client.go
```diff
@@ -32,3 +32,18 @@ func GetPachClient(t testing.TB) *client.APIClient {
 	}
 	return pachClient
 }
+
+// GetPachClient gets a pachyderm client for use in tests. It works for tests
+// running both inside and outside the cluster. Unlike GetPachClient,
+// multiple calls will return new pach client instances
+func GetNewPachClient(t testing.TB) *client.APIClient {
```
brendoncarroll, line 39: I was trying to create something similar to this in this P.R. https://github.com/pachyderm/pachyderm/pull/7126

The idea is to contain logic about how to use minikube (which is just a specific implementation of k8s) into a separate package, and make `testutil` only utility functions which do not allocate resources.  What do you think?

Regardless of the direction there: `GetNew` -> `New`

src/testing/upgrade/upgrade_test.go
```diff
@@ -2,6 +2,7 @@ package main
```
brendoncarroll, line 2: I really like this pattern of e2e tests that are just a package main.

src/internal/testutil/auth.go
```diff
@@ -43,6 +40,34 @@ func ActivateAuth(tb testing.TB) {
 	require.NoError(tb, err)
 }
 
+func ActivateAuthNewClient(tb testing.TB) {
```
brendoncarroll, line 43: If the functions in this file are all new additions (so only stuff in this PR is calling them), I think we should avoid adding any helpers that know about the singleton cluster.

Any new authentication helpers here should take a pach client, something like
```AuthenticatePachClient(t testing.TB, c *client.APIClient, subject string) *client.APIClient```

The client passed by an argument remains unchanged and the newly returned pach client has the updated authentication.  That would be a true utility function, only adding value, and not costing anything in terms of side effects or resource allocation, which may or may not be desired.  It could work for any pachclient even if we didn't deploy on minikube.

acohen4: great point

src/internal/testutil/pach_client.go
brendoncarroll: So right now we get PachClients from `minikubetestenv`, but we could also get PachClients from an in memory `testpachd` package.  `testutil` continues to be useful even after that switch.

brendoncarroll (approved):
See comments, no blockers.

PR #7241, opened 2022-01-13
Move RPC logging into an interceptor

Tryneus:
This moves the RPC logging functionality all into one place, from `src/internal/log` to `src/internal/middleware/logging`, and RPCs no longer need to explicitly log their request and response.  In addition:

* Got rid of the janky stack inspection that RPC logging did to figure out the RPC name, since we have that provided by grpc to the interceptor
* Performed a bit of a reorg and cleanup up the logging code that was moved into the middleware
* Removed the dependency on the logger for several API servers
* Provided a way to customize the logging behavior for certain endpoints, similar to how the auth middleware does it
  * Some auth endpoints customize the log level (for example, `WhoAmI` is always debug level unless an error occurs, and 'not activated' errors are moved to the debug level)
  * Several APIs have sensitive information that we don't want to log, so some endpoints copy the request/response and redact that information
* Streaming RPCs will now always log the first received message as the request and count the number of responses to log 'stream containing %d objects' in the response field
 
I think in the future, it would be better to annotate the protobufs for sensitive fields and not have all this necessary boilerplate, but that's a separate project.

### DIFF
--- a/src/internal/middleware/logging/interceptor.go
+++ b/src/internal/middleware/logging/interceptor.go
@@ -21,12 +21,12 @@ import (
 
 type logConfig struct {
 	// interfaces here are the request and response protobufs, respectively
-	level             func(interface{}, interface{}, error) logrus.Level
+	level             func(error) logrus.Level
 	transformRequest  func(interface{}) interface{}
 	transformResponse func(interface{}) interface{}
 }
 
-func defaultLevel(req interface{}, res interface{}, err error) logrus.Level {
+func defaultLevel(err error) logrus.Level {
 	if err == nil {
 		return logrus.InfoLevel
 	}
@@ -38,7 +38,7 @@ var defaultConfig = logConfig{level: defaultLevel}
 // The config used for auth endpoints suppresses 'not activated' errors to the
 // debug level
 var authConfig = logConfig{
-	level: func(req interface{}, resp interface{}, err error) logrus.Level {
+	level: func(err error) logrus.Level {
 		if err == nil {
 			return logrus.InfoLevel
 		} else if auth.IsErrNotActivated(err) {
@@ -195,11 +195,11 @@ var endpoints = map[string]logConfig{
 	"/auth_v2.API/RevokeAuthTokensForUser":    authConfig,
 
 	"/auth_v2.API/WhoAmI": {
-		level: func(req interface{}, resp interface{}, err error) logrus.Level {
+		level: func(err error) logrus.Level {
 			if err == nil {
 				return logrus.DebugLevel
 			}
-			return authConfig.level(req, resp, err)
+			return authConfig.level(err)
 		},
 	},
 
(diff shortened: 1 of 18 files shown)


src/server/cmd/pachd/main.go
```diff
@@ -258,7 +262,21 @@ func doEnterpriseMode(config interface{}) (retErr error) {
 	}
 
 	// Setup Internal Pachd GRPC Server.
-	internalServer, err := grpcutil.NewServer(context.Background(), false, grpc.ChainUnaryInterceptor(tracing.UnaryServerInterceptor(), authInterceptor.InterceptUnary), grpc.StreamInterceptor(authInterceptor.InterceptStream))
+	internalServer, err := grpcutil.NewServer(
+		context.Background(),
+		false,
+		grpc.ChainUnaryInterceptor(
+			errorsmw.UnaryServerInterceptor,
```
Tryneus, line 269: Also added `errorsmw` here since I had missed these servers in #7238.

src/internal/middleware/logging/interceptor.go
```diff
@@ -0,0 +1,462 @@
...
+
+	"github.com/pachyderm/pachyderm/v2/src/auth"
+	"github.com/pachyderm/pachyderm/v2/src/enterprise"
+	"github.com/pachyderm/pachyderm/v2/src/identity"
+	"github.com/pachyderm/pachyderm/v2/src/license"
+)
+
+type logConfig struct {
+	// interfaces here are the request and response protobufs, respectively
+	level             func(interface{}, interface{}, error) logrus.Level
```
acohen4, line 24: it looks like we aren't using the request/response arguments for logConfig.level(). What're your thoughts for having them in the signature?

src/server/pfs/server/api_server.go
```diff
@@ -452,11 +394,7 @@ func (a *apiServer) modifyFile(ctx context.Context, uw *fileset.UnorderedWriter,
 			}
 		case *pfs.ModifyFileRequest_CopyFile:
 			cf := mod.CopyFile
-			if err := func() (retErr error) {
-				func() { a.Log(ctx, cf, nil, nil, 0) }()
```
acohen4, line 456: do we still want to keep this logging since it's not at the RPC boundary?

src/internal/middleware/logging/interceptor.go
Tryneus: I'd be ok with removing it - I figured for customization it'd be nice to have all the information right there, but if it's ever needed (and it probably won't be), it's not too hard to add back in.

src/server/pfs/server/api_server.go
Tryneus: I debated this for a bit and decided not to keep it because none of the other branches in the switch log here.  I think if we want to log on this kind of thing, it should be done generically for all modify file operations.  Also, I suspect this may have just been left over from some debugging.

PR #7244, opened 2022-01-13
Add protoc plugin for autogenerating some protobuf boilerplate

Tryneus:
This mostly serves as a proof-of-concept and platform for more protobuf autogeneration in the future.  There are several places that should probably be moved to generated code rather than being kept in sync with the `.proto` files by hand.

This PR creates a `protoc` plugin under `etc/proto/pachgen`, which crawls the `.proto` definitions in the repository and puts them through any defined generators, which do terrible terrible things with golang's `text/template` abomination to generate the text of the output generated file (which should end in `.gen.go` and includes a comment header to warn developers away from hand-editing it).

Currently, there is only a generator for `src/client/transaction.gen.go`, which creates implementations of each grpc API client that errors every call with 'unsupported'.  This is used by the transaction client as the default behavior unless a custom handler is defined for that RPC (as transactions are only supported on distinct methods).

In the future, generators could be made for:
 * the mock server used in tests
 * a mock client for new unit-ier tests
 * example data or protobuf fixtures for tests to use
 * redaction functions for protobuf messages with sensitive information (which I wish I had in #7241)


src/client/transaction.gen.go
```diff
@@ -0,0 +1,569 @@
+// Code generated by pachgen (etc/proto/pachgen). DO NOT EDIT.
```
albscui, line 1: `DO NOT EDIT` reinforces at the IDE level.

etc/proto/run.sh
```diff
@@ -41,7 +48,19 @@ Mgoogle/protobuf/any.proto=github.com/gogo/protobuf/types,\
...
+    "-I${GOPATH}/pkg/mod/github.com/gogo/protobuf@${GOGO_PROTO_VERSION}" \
+    --pach_out="../v2/src" \
+    "${proto_files[@]}" > /dev/stderr
 
+popd > /dev/stderr
+
+
+# TODO (brendon): figure out how to configure protoc
+pushd v2 > /dev/stderr
+gofmt -w . > /dev/stderr
```
albscui, line 65: The original form of the generated code had some formatting issues with indentation and import sort order. The indentation issue is easily solved, but I couldn't figure out how to sort a `range` in Go templates.

etc/proto/pachgen/transaction_client.go
```diff
@@ -0,0 +1,108 @@
...
+	CodeGenerators = append(CodeGenerators, &TransactionClientGenerator{})
+}
+
+type TransactionClientGenerator struct {
+	Protos []*descriptor.FileDescriptorProto
+}
+
+var funcs = map[string]interface{}{
+	"importPath": func(proto *descriptor.FileDescriptorProto) string {
+		if *proto.Package == "taskapi" {
```
albscui, line 25: The fact that the proto package is called `taskapi` while the Go package in `src/` is called `task` makes this import a bit awkward.

etc/proto/pachgen/transaction_client.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		return fmt.Sprintf(`"github.com/pachyderm/pachyderm/v2/src/%s"`, path.Dir(*proto.Name))
+	},
+	"hasAPI": func(proto *descriptor.FileDescriptorProto) bool {
+		for _, service := range proto.Service {
+			if *service.Name == "API" || *service.Name == "Debug" {
+				return true
+			}
+		}
+		return *proto.Package == "taskapi"
```
albscui, line 36: Need this so that we can import `taskapi`.

PFedak: seems like this is really `shouldImport`. I wonder if this would be better handled by preprocessing in the go code that collects all direct package dependencies

etc/proto/pachgen/transaction_client.go
PFedak: the way protoc handles this is to name every single import `<protoPackageName> <go/package/path>`, so you don't have to special case. I'd prefer that here

PFedak (commented):
Not quite done reviewing yet, overall I'm on board. Just have to think through some of the detection pieces

etc/proto/run.sh
albscui: Hmm I tried `goimports`, but it also changes other generated protos.

etc/proto/pachgen/transaction_client.go
albscui: For proto packages with `_v2`, should we remove the `_v2` or import as-is?

etc/proto/Dockerfile
```diff
@@ -3,19 +3,26 @@ LABEL maintainer="msteffen@pachyderm.io"
...
+RUN go install -v github.com/gogo/protobuf/protoc-gen-gofast@v1.3.2 github.com/gogo/protobuf/protoc-gen-gogofast@v1.3.2 
 RUN mkdir -p ${GOPATH}/src/github.com/pachyderm/pachyderm
 RUN date +%s >/last_run_time
 
+RUN go install golang.org/x/tools/cmd/goimports@latest
+
+# Build the bespoke Pachyderm codegen plugin
+COPY pachgen ${GOPATH}/src/github.com/pachyderm/pachyderm/etc/proto/pachgen
+WORKDIR ${GOPATH}/src/github.com/pachyderm/pachyderm/etc/proto/pachgen
+RUN go mod init && go mod tidy && go build -o "${GOPATH}/bin/protoc-gen-pach" > /dev/stderr
```
PFedak, line 25: does that `/dev/stderr` redirect send it to that directory in the image?

etc/proto/build.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 # This script build protos using the pachyderm_proto image.
 
-find src -regex ".*\.proto" -print0 \
-| xargs -0 tar cf - \
-| docker run -i --rm pachyderm_proto \
-| tar xf -
+read -ra proto_files < <(find src -name "*.proto" -print0 | xargs -0)
+read -ra pachgen_files < <(find etc/proto/pachgen -print0 | xargs -0)
```
PFedak, line 5: do we still need this?

etc/proto/pachgen/transaction_client.go
```diff
@@ -0,0 +1,113 @@
...
+			}
+		}
+		// special case because taskapi has no services defined
+		return *proto.Package == "taskapi"
+	},
+	"isAPI": func(proto *descriptor.ServiceDescriptorProto) bool {
+		return *proto.Name == "API" || *proto.Name == "Debug"
+	},
+	"isClientStreaming": func(proto *descriptor.MethodDescriptorProto) bool {
+		return proto.ClientStreaming != nil && *proto.ClientStreaming
```
PFedak, line 47: since these are protos, could just use `GetClientStreaming()`, though I guess we don't do that a lot in our codebase

etc/proto/Dockerfile
albscui: oh right, this was taken from a script, but I think we can just drop the `> /dev/stderr` here.

PFedak (approved):
LGTM

PR #7246, opened 2022-01-17
Eliminate no-op scaling updates & inline autoscaling

msteffen:
Also, handle a corner case in the pipeline poller, where `watchPipelines` observes an event after the master is cancelled but before `watchPipelines` is killed. Finally, fix a typo in the `watchPipelines` comment, and clarify another in `pollPipelines`.

### DIFF
--- a/src/server/pps/server/monitor.go
+++ b/src/server/pps/server/monitor.go
@@ -11,7 +11,6 @@ import (
 	"github.com/robfig/cron"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/pachyderm/pachyderm/v2/src/client"
 	"github.com/pachyderm/pachyderm/v2/src/internal/backoff"
@@ -22,7 +21,6 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/tracing/extended"
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
 	"github.com/pachyderm/pachyderm/v2/src/pps"
-	"github.com/pachyderm/pachyderm/v2/src/server/worker/driver"
 	workerserver "github.com/pachyderm/pachyderm/v2/src/server/worker/server"
 )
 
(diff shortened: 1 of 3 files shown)


src/server/pps/server/pipeline_controller.go
```diff
@@ -748,19 +796,20 @@ func (step *pcStep) getRC(ctx context.Context, expectation rcExpectation) (retEr
...
-	rc := step.pc.env.KubeClient.CoreV1().ReplicationControllers(step.pc.namespace)
-
-	newRC := *step.rc
 	// Apply op's update to rc
-	update(&newRC)
-	// write updated RC to k8s
-	if _, err := rc.Update(ctx, &newRC, metav1.UpdateOptions{}); err != nil {
-		return newRetriableError(err, "error updating RC")
+	newRC := *step.rc
+	if update(&newRC) {
```
msteffen, line 807: This is the main change that eliminates no-op scaling updates

src/server/pps/server/poller.go
```diff
@@ -282,20 +284,27 @@ func (m *ppsMaster) watchPipelines(ctx context.Context) {
...
+				e = &pipelineEvent{
 					eventType: deleteEv,
 					pipeline:  pipelineName,
 					timestamp: time.Unix(event.Rev, 0),
 				}
 			}
+			select {
+			case m.eventCh <- e:
+				continue
+			case <-m.masterCtx.Done():
```
msteffen, line 305: The semantic change in this file: this ensures pollPipelines can exit after the PPS master restarts or loses mastership or something.

src/server/pps/server/pipeline_controller.go
```diff
@@ -555,23 +556,69 @@ func (step *pcStep) scaleUpPipeline(ctx context.Context) (retErr error) {
...
+			if !step.pipelineInfo.Details.Autoscaling {
+				return maxScale // don't bother if Autoscaling is off
+			}
+			if curScale == 0 {
+				return 1 // make one pod to be the worker master & calculate tasks
+			}
+
+			// Master is scheduled; see if tasks have been calculated
+			taskService := step.pc.env.TaskService
+			nTasks64, _, err := taskService.TaskCount(ctx, driver.TaskNamespace(step.pipelineInfo))
```
msteffen, line 581: image

src/server/pps/server/pipeline_controller.go
brendoncarroll: Is step a protocol buffer?  Is so, we should use proto.Clone instead of dereference then indirect.

msteffen: It's not a proto, but by investigating I found out that it does have a `DeepCopy()` method, which this now uses.

PFedak (approved):
LGTM

src/server/pps/server/pipeline_controller.go
```diff
@@ -186,7 +187,7 @@ func (pc *pipelineController) step(timestamp time.Time) (isDelete bool, retErr e
 	if !timestamp.IsZero() {
 		tracing.TagAnySpan(span, "update-time", timestamp)
 	} else {
-		tracing.TagAnySpan(span, "pollpipelines-event", "true")
+		tracing.TagAnySpan(span, "pollpipelines-or-autoscaling-event", "true")
```
acohen4, line 190: 👍

src/server/pps/server/pipeline_controller.go
acohen4, line 581: What does nTasks represent when queried? Is it number of Datums to process? When does this value get updated - as tasks are being completed or at the end of a job?

src/server/pps/server/monitor.go
```diff
@@ -249,49 +247,6 @@ func (pc *pipelineController) monitorPipeline(ctx context.Context, pipelineInfo
 			}, backoff.NewInfiniteBackOff(),
 				backoff.NotifyCtx(ctx, "monitorPipeline for "+pipeline))
 		})
-		if pipelineInfo.Details.ParallelismSpec != nil && pipelineInfo.Details.ParallelismSpec.Constant > 1 && pipelineInfo.Details.Autoscaling {
```
acohen4, line 252: Could you remind me what the main issue was that prompted moving this logic into scaleUpPipeline? Was it contention over the k8s API server with queries for the RC?

src/server/pps/server/pipeline_controller.go
msteffen: `nTasks` is the number of datums in the job total. IIUC, it's computed once, by the worker master, after the Job is started

src/server/pps/server/monitor.go
msteffen: At a high level, it's mostly to simplify the code, so that the only places where a pipeline RC is getting updated is by the pipeline's pipeline_controller, running in a synchronous loop. This sort of came out of the ["k8s stuckness" issue](https://github.com/pachyderm/pachyderm/pull/6038) from last year.

To explain "simplify the code" in a little more detail: when debugging that issue, one theory we had was that there was interference between `pollPipelines` and the autoscaling goro: if a pipeline was RUNNING and autoscaling was enabled, and then `step()` would set the number of pipeline replicas to 1, and if `pollPipelines` bumps a pipeline that's in the middle of running, that will drop its RC replicas down to 1 until the autoscaling goro runs and raises it again. That actually doesn't happen because `scaleUpPipeline()` only adjusts an RC's scale if it's currently at 0, but with the code written this way, the theory doesn't make sense because the RC isn't subject to any contention.

FWIW, I actually think the whole `monitorPipeline` goro could be inlined in the pipeline RC, making all pipeline updates synchronous. I actually think we could go a step farther and eliminate `STANDBY` completely, and have the handler for `RUNNING` (as well as `InspectPipeline`) just look at the pipeline's output repo for ready commits. Basically `RUNNING`+no ready commits=`STANDBY`, and then we wouldn't need a goro to explicitly change the pipeline's state. We'd just need a way to bump the pipelineController as soon as a new output commit is ready, but the poller would eventually fix things anyway in this model. All of this seemed out of scope for the change though.

msteffen: (also, if you think this change is a mistake, and we should limit this PR to eliminating no-op scaling updates, I'd be happy to do that)

msteffen:
I know that there's an issue with `TestAutoscalingStandby`. It takes several runs to reproduce the issue locally, but what seems to happen is that everything is going well and then the worker starts logging `lease keepalive response queue is full; dropping response send` from the etcd client every 3 seconds. Still investigating

src/server/pps/server/monitor.go
acohen4: Ok that makes a lot of sense! I totally agree with the approach just wasn't seeing the literal contension issue I remember chatting about at a high level

acohen4 (approved):
This is great!

PR #7250, opened 2022-01-19
Removing the pachyderm-storage-secret from the user container

acohen4:


tybritten (approved):
LGTM

PFedak (approved):
lgtm, i guess in principle someone could have been using this to access object storage directly from their pipeline code but that seems unlikely

PR #7256, opened 2022-01-21
Create API for listing tasks.

PFedak:


src/internal/middleware/auth/interceptor.go
```diff
@@ -224,6 +224,12 @@ var authHandlers = map[string]authHandler{
 
 	// TODO: Only the pachd sidecar instances should be able to use this endpoint.
 	"/proxy.API/Listen": unauthenticated,
+
+	//
+	// Task API
+	//
+
+	"/taskapi.API/ListTask": authDisabledOr(authenticated),
```
PFedak, line 232: unfortunately the test I wrote didn't catch that I missed adding the interceptor

src/taskapi/task.proto
```diff
@@ -0,0 +1,31 @@
...
+
+message TaskInfo {
+  string id = 1 [(gogoproto.customname) = "ID"];
+  string full_key = 3;
+  string input_type = 4;
+  State state = 5;
+  string reason = 6;
+}
+
+message ListTaskRequest {
```
brycemcanally, line 24: If we are going to have the task API be a separate gRPC service, then we need to include a service parameter (pfs, pps, ...). Also, we should be able to make the etcd prefixes fixed such that we can define constants for the various services.

src/taskapi/task.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
+
+package taskapi;
+option go_package = "github.com/pachyderm/pachyderm/v2/src/taskapi";
+
+import "gogoproto/gogo.proto";
+
+enum State {
+  STATE_UNKNOWN = 0;
```
brycemcanally, line 9: This can just be `UNKNOWN`, the state is implicit.

src/taskapi/task.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
+
+package taskapi;
```
brycemcanally, line 3: I don't think the api part is really needed. I think we should just keep the package as task and disambiguate with import renames if we need both `src/task` and `src/internal/task`.

src/client/task.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"io"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/grpcutil"
+	"github.com/pachyderm/pachyderm/v2/src/taskapi"
+)
+
+// ListTask lists tasks in the given namespace and group
+func (c APIClient) ListTask(namespace string) ([]*taskapi.TaskInfo, error) {
```
brycemcanally, line 13: I tend to prefer doing the gRPC scrub in a defer so I can just return err in the actual function.

src/client/task.go
```diff
@@ -0,0 +1,37 @@
...
+			Namespace: namespace,
+		},
+	)
+	var out []*taskapi.TaskInfo
+	if err != nil {
+		return nil, grpcutil.ScrubGRPC(err)
+	}
+	for {
+		ti, err := stream.Recv()
+		if errors.Is(err, io.EOF) {
```
brycemcanally, line 29: Would you change this, to this:
```
if err != nil {
	if errors.Is(err, io.EOF) {
		break
	}
	return err
}
```
It reads a bit better imo, and it is a pattern I have been gradually moving our client wrappers to.

src/client/task.go
```diff
@@ -0,0 +1,37 @@
...
+	ctx, cancel := context.WithCancel(c.Ctx())
+	defer cancel()
+
+	stream, err := c.TaskClient.ListTask(
+		ctx,
+		&taskapi.ListTaskRequest{
+			Namespace: namespace,
+		},
+	)
+	var out []*taskapi.TaskInfo
```
brycemcanally, line 23: It seems like this declaration should be after the error check.

src/server/task/server/api_server.go
```diff
@@ -0,0 +1,69 @@
...
+
+	etcd "go.etcd.io/etcd/client/v3"
+)
+
+type apiServer struct {
+	ppsPrefix, pfsPrefix string
+	etcdClient           *etcd.Client
+}
+
+func NewAPIServer(base, pps, pfs string, etcdClient *etcd.Client) taskapi.APIServer {
```
brycemcanally, line 20: I think we should just take an env here and access the prefix fields when we need them rather than storing them in the api server. Also, etcd can be accessed from the env as well.

src/server/task/server/api_server.go
```diff
@@ -0,0 +1,69 @@
...
+	if req.Namespace == "" {
+		return errors.New("Must include a task namespace")
+	}
+	prefix := a.ppsPrefix
+	if strings.HasPrefix(req.Namespace, "storage") {
+		prefix = a.pfsPrefix
+	}
+	etcdCols := task.NewNamespaceEtcd(a.etcdClient, prefix, req.Namespace)
+	var taskData task.Task
+	etcdCols.TaskCol.ReadOnly(ctx).List(&taskData, col.DefaultOptions(), func(key string) error {
```
brycemcanally, line 51: Looks like we are dropping the list error, we can just return the result of this call.

brycemcanally: Also, this functionality should probably be built into the task service. This is pulling the details out of a particular implementation. I think we could handle all of this (including the claim state) internally and expose a list method through the service interface. We would still want to have a separate type that we convert to though.

src/server/task/cmds/cmds.go
```diff
@@ -0,0 +1,51 @@
...
+
+func Cmds() []*cobra.Command {
+	var commands []*cobra.Command
+
+	var raw bool
+	var output string
+	outputFlags := cmdutil.OutputFlags(&raw, &output)
+
+	listTask := &cobra.Command{
+		Hidden: true, // don't show in the list of commands
```
brycemcanally, line 21: :+1:

src/taskapi/task.proto
PFedak: unfortunately proto seems to need `protoPackage.messageName` to be unique, so this can't just be `task` unless I also change some of the messages. I can, however, change the go package to just `task`

src/server/task/server/api_server.go
PFedak: reworked to put everything in the task service, with simple wrapper implementations

src/taskapi/task.proto
PFedak: whoops

src/taskapi/task.proto
```diff
@@ -0,0 +1,31 @@
...
+
+enum State {
+  STATE_UNKNOWN = 0;
+  RUNNING = 1;
+  SUCCESS = 2;
+  FAILURE = 3;
+  CLAIMED = 4; // not a real state used by task logic
+}
+
+message TaskInfo {
```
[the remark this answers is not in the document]
PFedak: Did what seemed to be the right thing using `types.DynamicAny`

src/client/transaction.go
```diff
@@ -423,6 +424,9 @@ func (c *pfsBuilderClient) RunLoadTestDefault(ctx context.Context, req *types.Em
 func (c *pfsBuilderClient) CheckStorage(ctx context.Context, req *pfs.CheckStorageRequest, opts ...grpc.CallOption) (*pfs.CheckStorageResponse, error) {
 	return nil, unsupportedError("CheckStorage")
 }
+func (c *pfsBuilderClient) ListTask(ctx context.Context, in *task.ListTaskRequest, opts ...grpc.CallOption) (pfs.API_ListTaskClient, error) {
+	return nil, unsupportedError("RunLoadTestDefault")
```
brycemcanally, line 428: `RunLoadTestDefault` -> `ListTask`

src/internal/task/task.go
```diff
@@ -39,6 +39,8 @@ type Service interface {
 	NewSource(namespace string) Source
 	// TaskCount returns how many tasks are in a namespace and how many are claimed.
 	TaskCount(ctx context.Context, namespace string) (tasks int64, claims int64, _ error)
+	// ListTaskF calls a function on every task under a given namespace/etcd prefix
+	ListTaskF(ctx context.Context, prefix string, f func(key string, data *Task, claimed bool) error) error
```
brycemcanally, line 43: We have been moving away from the `F` suffix for callback based functions. We should just have this be `ListTask`. Also, we tend to use `cb` for the variable name of the callback function.

src/internal/task/etcd_service.go
```diff
@@ -63,6 +66,61 @@ func (es *etcdService) TaskCount(ctx context.Context, namespace string) (tasks i
 	return nTasks, nClaims, nil
 }
 
+func (es *etcdService) ListTaskF(ctx context.Context, namespace string, f func(string, *Task, bool) error) error {
+	etcdCols := newNamespaceEtcd(es.etcdClient, es.etcdPrefix, namespace)
+	var taskData Task
+	var claim Claim
+	etcdCols.taskCol.ReadOnly(ctx).List(&taskData, col.DefaultOptions(), func(key string) error {
```
brycemcanally, line 73: We are missing a potential list error here, we can just return here to collect it.

src/internal/task/etcd_service.go
```diff
@@ -63,6 +66,61 @@ func (es *etcdService) TaskCount(ctx context.Context, namespace string) (tasks i
...
+		return taskapi.State_FAILURE
+	}
+	return taskapi.State_UNKNOWN
+}
+
+func HandleList(ctx context.Context, svc Service, req *taskapi.ListTaskRequest, send func(info *taskapi.TaskInfo) error) error {
+	var marshaler jsonpb.Marshaler
+	return errors.EnsureStack(svc.ListTaskF(ctx, req.Namespace, func(key string, data *Task, claimed bool) error {
+		var state taskapi.State
+		if claimed {
```
brycemcanally, line 99: The way I prefer to structure stuff like this is to just do the default unconditionally, then if claimed overwrite with the claimed state. So like this:
```
state := translateTaskState(data.State)
if claimed {
  state = taskapi.State_CLAIMED
}
```

src/pfs/pfs.proto
```diff
@@ -524,4 +526,7 @@ service API {
   rpc RunLoadTest(RunLoadTestRequest) returns (RunLoadTestResponse) {}
   // RunLoadTestDefault runs the default load tests.
   rpc RunLoadTestDefault(google.protobuf.Empty) returns (RunLoadTestResponse) {}
+
+  // ListTask lists storage tasks under the given etcd prefix
```
brycemcanally, line 530: Lets go ahead and make this more general. So, just `lists PFS tasks`.

src/pps/pps.proto
```diff
@@ -772,5 +774,8 @@ service API {
 
   // RenderTemplate renders the provided template and arguments into a list of Pipeline specicifications
   rpc RenderTemplate(RenderTemplateRequest) returns (RenderTemplateResponse) {}
+
+  // ListTask lists pipeline tasks under the given etcd prefix
```
brycemcanally, line 778: `lists PPS tasks`

src/internal/task/task.go
brycemcanally: It might actually make sense to just call this `List` since task is implicit in the package.

brycemcanally: Also, we should be able to remove the `TaskCount` method from the interface. We could move that to a utility function based on the `List` method. See the `util.go` file for an example using the `Do` method.

src/internal/task/etcd_service.go
```diff
@@ -63,6 +66,61 @@ func (es *etcdService) TaskCount(ctx context.Context, namespace string) (tasks i
...
+		return taskapi.State_RUNNING
+	case State_SUCCESS:
+		return taskapi.State_SUCCESS
+	case State_FAILURE:
+		return taskapi.State_FAILURE
+	}
+	return taskapi.State_UNKNOWN
+}
+
+func HandleList(ctx context.Context, svc Service, req *taskapi.ListTaskRequest, send func(info *taskapi.TaskInfo) error) error {
```
brycemcanally, line 95: We should move this to the `util.go` file.

src/client/transaction.go
PFedak: whoops

src/internal/task/etcd_service.go
PFedak: sorry, don't know how I missed that a second time

src/internal/task/etcd_service.go
PFedak: makes sense, not sure why I wrote it the way I did

src/task/task.proto
```diff
@@ -0,0 +1,28 @@
...
+  UNKNOWN = 0;
+  RUNNING = 1;
+  SUCCESS = 2;
+  FAILURE = 3;
+  CLAIMED = 4; // not a real state used by task logic
+}
+
+message TaskInfo {
+  string id = 1 [(gogoproto.customname) = "ID"];
+  string full_key = 3;
```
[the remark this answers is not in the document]
PFedak: The concern I had with namespace and group is how to extract it properly from the key. If you can assume the group has no slashes, we can just split the string. Otherwise it seems like we in general need to look for the group collection by trying different prefixes. Is there some simple thing I'm missing here?

src/taskapi/task.proto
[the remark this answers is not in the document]
brycemcanally: Awesome, that is exactly what we want.

src/task/task.proto
```diff
@@ -0,0 +1,28 @@
...
+
+enum State {
+  UNKNOWN = 0;
+  RUNNING = 1;
+  SUCCESS = 2;
+  FAILURE = 3;
+  CLAIMED = 4; // not a real state used by task logic
+}
+
+message TaskInfo {
```
brycemcanally, line 21: A couple of nitpicky things:
- Be sure not to skip any field numbers (i.e. 2). Any gaps should be set to reserved, but that should only occur when a field is being removed while maintaining compatibility.
- Collocate the input fields (type, data) in the message. So the last two could be for the input fields.

src/internal/task/util.go
```diff
@@ -45,3 +48,62 @@ func DoBatch(ctx context.Context, doer Doer, inputs []*types.Any, cb CollectFunc
...
+	case State_SUCCESS:
+		return taskapi.State_SUCCESS
+	case State_FAILURE:
+		return taskapi.State_FAILURE
+	}
+	return taskapi.State_UNKNOWN
+}
+
+// HandleList implements the functionality for an arbitrary service's ListTask gRPC
+func HandleList(ctx context.Context, svc Service, req *taskapi.ListTaskRequest, send func(info *taskapi.TaskInfo) error) error {
```
brycemcanally, line 65: We could probably just rename this to `List`.

src/internal/task/task.proto
```diff
@@ -30,3 +30,9 @@ message Claim {}
 message TestTask {
   string id = 1 [(gogoproto.customname) = "ID"];
 }
+
+message TaskKey {
```
brycemcanally, line 34: I'm starting to think the key should just be an implementation detail, and we might be better served by just reporting the namespace / group. If that seems reasonable, then we can have a group structure similar to the taskapi one or just return the namespace and group in the callback. Either would be fine for the time being.

src/internal/task/util.go
```diff
@@ -45,3 +48,62 @@ func DoBatch(ctx context.Context, doer Doer, inputs []*types.Any, cb CollectFunc
...
+		if err := types.UnmarshalAny(data.Input, &input); err != nil {
+			return errors.EnsureStack(err)
+		}
+		inputJSON, err := marshaler.MarshalToString(input.Message)
+		if err != nil {
+			return errors.EnsureStack(err)
+		}
+		info := &taskapi.TaskInfo{
+			ID: data.ID,
+			Group: &taskapi.Group{
```
brycemcanally, line 82: Very minor nit, but I think we should in general initialize in order with respect to the order that the fields are defined. So, namespace, then group.

src/client/task.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/internal/grpcutil"
+	"github.com/pachyderm/pachyderm/v2/src/task"
+)
+
+// ListTask lists tasks in the given namespace and group
+func (c APIClient) ListTask(service string, namespace, group string) (infos []*task.TaskInfo, retErr error) {
+	ctx, cancel := context.WithCancel(c.Ctx())
+	defer cancel()
+	defer func() {
+		if retErr != nil {
```
brycemcanally, line 17: This condition is unnecessary.

src/client/task.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+	"io"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/grpcutil"
+	"github.com/pachyderm/pachyderm/v2/src/task"
+)
+
+// ListTask lists tasks in the given namespace and group
+func (c APIClient) ListTask(service string, namespace, group string) (infos []*task.TaskInfo, retErr error) {
```
brycemcanally, line 13: Lets change this to a callback based interface so it can scale.

src/internal/task/cmds/cmds.go
```diff
@@ -0,0 +1,61 @@
...
+	var output string
+	outputFlags := cmdutil.OutputFlags(&raw, &output)
+
+	var group string
+	listTask := &cobra.Command{
+		Hidden: true, // don't show in the list of commands
+		Use:    "{{alias}} <service> [<namespace>]",
+		Short:  "Return info about tasks in a namespace.",
+		Long:   "Return info about tasks in a namespace.",
+		Run: cmdutil.RunBoundedArgs(1, 2, func(args []string) error {
```
brycemcanally, line 26: We should probably require the service and have the namespace & group be optional flags.

src/internal/task/cmds/cmds.go
```diff
@@ -0,0 +1,61 @@
...
+
+	var raw bool
+	var output string
+	outputFlags := cmdutil.OutputFlags(&raw, &output)
+
+	var group string
+	listTask := &cobra.Command{
+		Hidden: true, // don't show in the list of commands
+		Use:    "{{alias}} <service> [<namespace>]",
+		Short:  "Return info about tasks in a namespace.",
```
brycemcanally, line 24: Could we just remove the namespace part or change it to service. The service seems to be the primary command, and the namespace & group are just filters.

src/task/task.proto
PFedak: ah, must have missed when I was switching things around

src/internal/task/task.proto
PFedak: I was thinking it would be good to have some way of getting to the underlying task just in case, but I can't really think of a case where we'd want that.

src/internal/task/cmds/cmds.go
```diff
@@ -0,0 +1,62 @@
...
+				return errors.Errorf("must set a task namespace to list a group")
+			}
+
+			client, err := pachdclient.NewOnUserMachine("user")
+			if err != nil {
+				return err
+			}
+			defer client.Close()
+
+			var taskInfos []*task.TaskInfo
```
brycemcanally, line 38: We should also stream the results here.

PFedak: ah, of course, sorry

src/internal/task/util.go
```diff
@@ -45,3 +48,61 @@ func DoBatch(ctx context.Context, doer Doer, inputs []*types.Any, cb CollectFunc
...
+			Reason:    data.Reason,
+			InputType: data.Input.TypeUrl,
+			InputData: inputJSON,
+		}
+		return errors.EnsureStack(send(info))
+	}))
+}
+
+// Count returns the number of tasks and claims in the given namespace and group (if nonempty)
+func Count(ctx context.Context, namespace, group string, service Service) (tasks int64, claims int64, retErr error) {
```
brycemcanally, line 96: nit, can this be ordered as ctx, service, namespace, group.

brycemcanally (approved):
lgtm, just one more minor nit and CI being green after merging in the autoscaling fix.

src/server/worker/driver/driver.go
```diff
@@ -41,7 +41,7 @@ import (
 // TaskNamespace returns the namespace used by the task package for this
 // pipeline.
 func TaskNamespace(pipelineInfo *pps.PipelineInfo) string {
-	return fmt.Sprintf("/pipeline-%s/v%d", pipelineInfo.Pipeline.Name, pipelineInfo.Version)
+	return fmt.Sprintf("/pipeline-%s@v%d", pipelineInfo.Pipeline.Name, pipelineInfo.Version)
```
brycemcanally, line 44: Can we add the same padding we have in the pps version key (8).

PFedak: I've just reused the key function, but can copy if it's preferable to keep them separate.

PR #7261, opened 2022-01-24
Block on establishing pg proxy Listen()

acohen4:
There are two fixes here involving the Watcher implementation.
1.) For workers that listen for events via a proxy that streams events from a pachd instance, we change that proxy's grpc so that it signals when it is completely set up and ready to stream events. We make sure to wait for the initially empty event on the client side.

2.) Fix a bug in the Watcher logic where we incorrectly manage the last timestamp variable that is used once the watcher hands off streaming from the list() call to database events

### DIFF
--- a/src/client/proxy.go
+++ b/src/client/proxy.go
@@ -48,6 +48,10 @@ func (ppl *proxyPostgresListener) setup() error {
 	return err
 }
 
+// TODO: for now, calls to Register() will block on the Listener being established
+// on the main pachd instance. This could become a bottleneck if many Watchers are
+// instantiated at once. Consider finer grained locking at the channel level using
+// a mechanism such as the "WorkDeduper" (src/internal/miscutil/work_deduper.go).
 func (ppl *proxyPostgresListener) Register(notifier col.Notifier) error {
 	ppl.mu.Lock()
 	defer ppl.mu.Unlock()
@@ -83,6 +87,7 @@ func (ppl *proxyPostgresListener) listen(notifier col.Notifier) {
 		notifier.Error(err)
 		cancel()
 		delete(ppl.channelInfos, channel)
+		return
 	}
 
 	go func() {
--- a/src/client/proxy.go
+++ b/src/client/proxy.go
@@ -66,10 +66,20 @@ func (ppl *proxyPostgresListener) listen(notifier col.Notifier) {
 	channel := notifier.Channel()
 	ctx, cancel := context.WithCancel(context.Background())
 	ppl.channelInfos[channel] = newChannelInfo(cancel, notifier)
-	listenClient, err := ppl.client.Listen(ctx, &proxy.ListenRequest{
-		Channel: channel,
-	})
-	if err != nil {
+
+	var listenClient proxy.API_ListenClient
+	if err := func() error {
+		var err error
+		listenClient, err = ppl.client.Listen(ctx, &proxy.ListenRequest{
+			Channel: channel,
+		})
+		if err != nil {
+			return err
+		}
+		// wait for the first init event to be returned by the server
+		_, err = listenClient.Recv()
+		return err
+	}(); err != nil {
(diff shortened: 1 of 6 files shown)


src/server/proxy/server/api_server.go
```diff
@@ -29,6 +29,17 @@ func (a *APIServer) Listen(request *proxy.ListenRequest, server proxy.API_Listen
 	if err := listener.Register(notifier); err != nil {
 		return errors.EnsureStack(err)
 	}
+
+	// send initial empty event to indicate to client that the listener has been registered
+	if err := server.Send(&proxy.ListenResponse{
+		Extra: "",
+	}); err != nil {
+		notifier.sendError(err)
+		return
```
PFedak, line 38:
```suggestion
		return err
```
pretty sure this is all you need since we haven't started the notifier routine yet

PFedak (approved):
LGTM, should probably add a TODO comment on `Register` about the locking scope issues we mentioned, as well as something describing the contract we are establishing w/r/t being ready to listen after `Register` returns.

src/server/proxy/server/api_server.go
PFedak: oh actually as is you wouldn't return the error, that's a weird pitfall of named return values

src/client/proxy.go
```diff
@@ -66,14 +66,27 @@ func (ppl *proxyPostgresListener) listen(notifier col.Notifier) {
...
+		if err != nil {
+			return err
+		}
+		// wait for the first init event to be returned by the server
+		_, err = listenClient.Recv()
+		return err
+	}(); err != nil {
+		notifier.Error(err)
+		cancel()
+		delete(ppl.channelInfos, channel)
```
PFedak, line 85: presumably return after this

src/server/proxy/server/api_server.go
acohen4: eek! good catch

src/client/proxy.go
```diff
@@ -48,6 +48,10 @@ func (ppl *proxyPostgresListener) setup() error {
 	return err
 }
 
+// TODO: for now, calls to Register() will block on the Listener being established
```
brycemcanally, line 51: :+1:

src/internal/collection/postgres_collection.go
```diff
@@ -556,6 +557,7 @@ func (c *postgresReadOnlyCollection) watchRoutine(watcher *postgresWatcher, opti
 		}
 
 		if bufEvent != nil && bufEvent.time.Unix() <= lastTs.Unix() {
+			logrus.Errorf("Handing off watcher!! sending buf event: %v ; and not sending list element: %v\n", bufEvent.key, m.Key)
```
brycemcanally, line 560: This should be removed.

brycemcanally (commented):
We should document the new Listen gRPC behavior in both the service definition in the proto file and the actual implementation.

brycemcanally (approved):
lgtm, just some minor changes needed.

src/internal/collection/postgres_listener.go
```diff
@@ -108,33 +108,25 @@ func (pw *postgresWatcher) Close() {
...
+func (pw *postgresWatcher) forwardNotifications(ctx context.Context) {
 	for {
-		// This allows for double notification on a matching timestamp which is
-		// better than missing a notification. This will _always_ happen on the last
-		// put event if a change is made while doing the list.
-		// We could try to use the postgres xmin value instead of the modified
-		// timestamp, but that has to detect wraparounds, which is non-trivial.
 		select {
 		case eventData := <-pw.buf:
-			if !eventData.time.Before(startTime) || eventData.err != nil {
```
acohen4, line 122: basically I only removed this conditional check. startTime comparison no longer applied, and the second condition looked erroneous to me

PR #7262, opened 2022-01-24
Task service caching layer and job checkpointing

brycemcanally:
This PR implements the task service caching layer and integrates it with the worker master to enable job checkpointing.

**Task service caching layer:**

Clients of the task service can now provide an implementation of a cache interface (defined by the task service) to use for caching tasks. The reason for having the client provide the implementation is because the details of caching different types of tasks can vary (i.e. caching file sets requires ensuring they are not garbage collected). The task service will attempt to get from the cache before creating a task and will only cache a task if it was processed without an error.

The PFS cache is the first implementation of the cache interface and provides functionality for caching an Any proto and associating a list of file sets with the entry (the file sets will not be garbage collected while the entry is in the cache). The cache is an LRU cache built on top of Postgres and transactionally manages entries / tracking of file sets. The PFS cache is exposed through gRPC for use by the worker shim.

**Job checkpointing:**

As of 2.0, jobs that hit transient errors in the worker master have to restart from scratch. This is less than ideal for jobs that have a lot of datums / longer running datums. It would be better to be able to checkpoint the tasks as they are being completed. The task service caching layer was added to enable this type of checkpointing since it is how we represent tasks being done in a job.

After adding the caching layer and using it in the job processing for caching tasks, we still needed to make some changes to the job preprocessing steps to ensure we get the correct caching behavior. The reason we wouldn't get the correct caching behavior out of the box is because the caching is based on task ids, which is based on the fields in the proto that define the task. One of the fields that is commonly used is a file set id that holds files relevant for the job processing step. File set ids are not currently based on the content and are instead uuids. This means that the steps that produce these file sets need to be cached as well to ensure that we eventually get back to a cache hit that was based on stable fields. This required refactoring the job preprocessing steps to use the task service which is a net win for the following reasons:
- It enables the caching behavior we want for job checkpointing
- It makes the preprocessing steps more unit testable (TODO)
- It will give us more visibility into the preprocessing steps since work is being done on a task service API for viewing the tasks in the system.
- It lays the foundation for distributing the preprocessing work more effectively, particularly taking load off of the worker master which is responsible for managing all of the jobs for a pipeline.


src/server/worker/pipeline/transform/worker.go
```diff
@@ -16,18 +18,251 @@ import (
...
+					FileSetId:    resp.FileSetId,
+				})
+				if err != nil {
+					return err
+				}
+				data, err := proto.Marshal(input)
+				if err != nil {
+					return errors.EnsureStack(err)
+				}
+				return errors.EnsureStack(mf.PutFile(name, bytes.NewReader(data)))
```
PFedak, line 209: so, if i'm understanding correctly, the change here is that the computed datum sets get written to/read from a fileset before the actual datum set tasks are created, which slows things down a little in order to add another checkpoint step

PFedak (approved):
My comments are minor, this looks so much better and clearer - I'm excited!

src/internal/storage/fileset/postgres_cache.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/internal/dbutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/track"
+)
+
+// CreatePostgresCacheV1 creates the table for a cache.
+// DO NOT MODIFY THIS FUNCTION
+// IT HAS BEEN USED IN A RELEASED MIGRATION
+func CreatePostgresCacheV1(ctx context.Context, tx *pachsql.Tx) error {
```
brendoncarroll, line 18: I hadn't originally intended the V0 names to correspond with 2.0.x, but it makes sense.

src/internal/storage/fileset/postgres_cache.go
```diff
@@ -0,0 +1,122 @@
...
+		ON CONFLICT (key) DO NOTHING
+	`, key, value, ids)
+	if err != nil {
+		return errors.EnsureStack(err)
+	}
+	var pointsTo []string
+	for _, id := range ids {
+		pointsTo = append(pointsTo, id.TrackerID())
+	}
+	return errors.EnsureStack(c.tracker.CreateTx(tx, cacheTrackerKey(key), pointsTo, track.NoTTL))
```
brendoncarroll, line 77: Shouldn't the `EnsureStack` be in the tracker, since we also control that code?

src/internal/storage/fileset/postgres_cache.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/track"
+)
+
+// CreatePostgresCacheV1 creates the table for a cache.
+// DO NOT MODIFY THIS FUNCTION
+// IT HAS BEEN USED IN A RELEASED MIGRATION
+func CreatePostgresCacheV1(ctx context.Context, tx *pachsql.Tx) error {
+	const schema = `
+	CREATE TABLE storage.cache (
+		key text NOT NULL PRIMARY KEY,
```
brendoncarroll, line 21: This could also be a `BYTEA`.  Are you hex-encoding the IDs?

src/internal/storage/fileset/postgres_cache.go
```diff
@@ -0,0 +1,122 @@
...
+		value_pb BYTEA NOT NULL,
+		ids UUID[] NOT NULL,
```
brendoncarroll, line 23: These ids refer to filesets?

src/internal/storage/fileset/storage.go
```diff
@@ -320,14 +320,8 @@ func (s *Storage) newGC() *track.GarbageCollector {
...
 func (s *Storage) exists(ctx context.Context, id ID) (bool, error) {
-	_, err := s.store.Get(ctx, id)
-	if err != nil {
-		if errors.Is(err, ErrFileSetNotExists) {
-			return false, nil
-		}
-		return false, errors.EnsureStack(err)
-	}
-	return true, nil
+	exists, err := s.store.Exists(ctx, id)
```
brendoncarroll, line 323: :+1:

src/internal/storage/fileset/postgres_cache_test.go
```diff
@@ -0,0 +1,82 @@
...
+func newTestCache(t *testing.T, db *pachsql.DB, tr track.Tracker, maxSize int) *Cache {
+	ctx := context.Background()
+	tx := db.MustBegin()
+	tx.MustExec(`CREATE SCHEMA IF NOT EXISTS storage`)
+	require.NoError(t, CreatePostgresCacheV1(ctx, tx))
+	require.NoError(t, tx.Commit())
+	return NewCache(db, tr, maxSize)
+}
+
+func TestPostgresCache(t *testing.T) {
```
brendoncarroll, line 25: 👍

src/internal/task/task.go
```diff
@@ -67,3 +67,8 @@ type CollectFunc = func(index int64, output *types.Any, _ error) error
 // If an error occurs, then it should be returned.
 // This error will be propagated back to the Doer that created the task.
 type ProcessFunc = func(ctx context.Context, input *types.Any) (output *types.Any, _ error)
+
+type Cache interface {
```
brendoncarroll, line 71: I think this and the logic above would be better encapsulated as a `CachedDoer`.  So the base service is unmodified and the CachedDoer wraps an inner doer + a cache.  The etcd service is pretty complicated, and I think it's useful valuable to keep it as simple as possible.

src/internal/testutil/kubeclient.go
```diff
@@ -44,15 +44,13 @@ func GetKubeClient(t testing.TB) *kube.Clientset {
 	return k
 }
 
-// DeletePachdPod deletes the pachd pod in a test cluster (restarting it, e.g.
-// to retart the PPS master)
-func DeletePachdPod(t testing.TB) {
+func DeletePod(t testing.TB, app string) {
```
brendoncarroll, line 47: Is this related to the Task Service stuff?

src/server/pfs/server/api_server.go
```diff
@@ -719,6 +719,29 @@ func (a *apiServer) CheckStorage(ctx context.Context, req *pfs.CheckStorageReque
 	}, nil
 }
 
+func (a *apiServer) PutCache(ctx context.Context, req *pfs.PutCacheRequest) (resp *types.Empty, retErr error) {
```
brendoncarroll, line 722: This ended up being pretty clean.  I remember we talked about how complex this would be a while ago.

src/internal/storage/renew/string_set.go
```diff
@@ -71,6 +71,17 @@ func (ss *StringSet) Add(ctx context.Context, x string) error {
 	return nil
 }
 
+// Compose returns the result of a compose call on all of the strings being renewed.
+func (ss *StringSet) Compose(ctx context.Context) (string, error) {
```
brendoncarroll, line 75: This doesn't affect the internal state of the StringSet like the background composing loop does.  What is this for?

brendoncarroll (approved):
A few questions, and I think we should not include the caching logic in the Etcd Task Doer.  Otherwise LGTM.

src/internal/testutil/kubeclient.go
brendoncarroll: Ahh there were worker changes too, resolving.

src/internal/storage/fileset/postgres_cache.go
PFedak: Unfortunately the wrapcheck linter assumes interfaces could have implementations outside your code, which is part of the reason we have so many. I've been looking into forking it to ignore interfaces we define, but it's a little complicated as it's bundled with `golangci`

src/server/pachyderm_test.go
```diff
@@ -9801,6 +9801,61 @@ func TestStandbyTransitions(t *testing.T) {
...
+	defer cancel()
+	go func() {
+		ticker := time.NewTicker(time.Minute)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				tu.DeletePod(t, "etcd")
```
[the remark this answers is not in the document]
brycemcanally: The issue I was hitting was due to the master lock retry being pretty long ~20s in the test case, which made me think the job logic was stuck. When the master lock is lost, we hit a "fatal" error in the job logic, which will get retried at a higher level when the master lock is reacquired. Bumping up the test to be a bit longer and tweaking the delete pod frequency addresses the issue by still introducing failures fast enough to prevent full job completion, but slow enough to allow some room between the retry backoff and some processing.

src/server/worker/pipeline/transform/pending_job.go
```diff
@@ -329,10 +363,64 @@ func withDatumFileSet(pachClient *client.APIClient, cb func(*datum.Set) error) (
...
-func (pj *pendingJob) skippableDatum(meta1, meta2 *datum.Meta) bool {
-	if pj.noSkip {
-		return false
-	}
+func skippableDatum(meta1, meta2 *datum.Meta) bool {
 	// If the hashes are equal and the second datum was processed, then skip it.
 	return meta1.Hash == meta2.Hash && meta2.State == datum.State_PROCESSED
 }
+
+func serializeUploadDatumsTask(task *UploadDatumsTask) (*types.Any, error) {
```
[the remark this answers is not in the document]
brycemcanally: I'll add a TODO for this, the boilerplate is a bit annoying, but it is a common pattern we have in a few other places that I would like to keep for now in the interesting of keeping the URLs consistent.

src/server/worker/pipeline/transform/cache.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (c *cache) Put(ctx context.Context, key string, output *types.Any) error {
+	var fileSetIds []string
+	switch {
+	case types.Is(output, &UploadDatumsTaskResult{}):
+		udt, err := deserializeUploadDatumsTaskResult(output)
+		if err != nil {
+			return err
+		}
+		fileSetIds = append(fileSetIds, udt.FileSetId)
```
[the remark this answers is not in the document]
brycemcanally: Would that make things simpler if we still need the type switch here?

src/server/worker/pipeline/transform/pending_job.go
```diff
@@ -162,154 +149,201 @@ func (pj *pendingJob) withDeleter(pachClient *client.APIClient, cb func(datum.De
...
+		}
+		baseFileSetID = result.FileSetId
+		return nil
+	}); err != nil {
+		return "", errors.EnsureStack(err)
+	}
+	return baseFileSetID, nil
+}
+
+func (pj *pendingJob) createJobDatumFileSetParallel(ctx context.Context, taskDoer task.Doer, renewer *renew.StringSet, fileSetID, baseFileSetID string) (string, error) {
```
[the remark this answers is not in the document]
brycemcanally: I do think I can refactor this code for more reuse. That was in the back of my mind while coding this, but I put it on the back burner. I'll see what I can do now. I think the cache should stay behind the doer.

src/server/worker/pipeline/transform/worker.go
brycemcanally: That is true, but the performance hit is minimal considering it is one file per datum set. This model will enable more significant performance improvements down the line (i.e. parallel and distributed preprocessing).

src/internal/storage/fileset/postgres_cache.go
brycemcanally: I can go with v0 if that is more consistent with what we have up to this point. I have some concerns about the migration process, particularly naming migrations without full semver and the packages managing the different versions rather than snapshotting them elsewhere, but I don't think we need to address that now.

src/internal/storage/fileset/postgres_cache.go
brycemcanally: The linter does a pretty good job ensuring we have a stack at the boundaries of our code and third party libraries, so I am in favor of only using `EnsureStack` when the linter says we need it. Requiring it at our interface boundaries is less than ideal, but it isn't as bad as us just throwing them at arbitrary error propagation sites.

src/internal/storage/fileset/postgres_cache.go
brycemcanally: The task IDs are hex encoded. What would be the benefit of `BYTEA` if the key field is expected to be small and printable characters? I'm just wondering if the performance / space improvements is going to outweigh the additional complexity that comes with managing binary data.

src/internal/storage/fileset/postgres_cache.go
brycemcanally: yes

src/internal/task/task.go
brycemcanally: Yeah, that is what I started with, but that doesn't work well with the index functionality. Each task received through the channel increments an internal index that maps back to the order the tasks were created. Since most of the caching functionality is handled by the cache implementation, I don't think it adds that much complexity to the Doer, but I am open to alternatives that handle the index issue.

src/internal/storage/renew/string_set.go
brycemcanally: Right, this is just to get a single file set that references all of the file sets being renewed. This is useful for keeping the file sets being renewed alive across processes. The current use case is that we have a task that creates all of the file sets for the datum sets and we want the worker master to be able to keep those alive while the job is being processed (see `CreateDatumSetsTask`)

src/internal/task/task.go
brendoncarroll: The CacheDoer.Do method reads from the channel, checks the cache, if there's a hit, call cb, otherwise send it along to the inner Do implementation.  In the callback, put it in the cache, and call the outer cb.

I guess you do need to track the indexes on the outer level too. You have to track the number of cache hits when you send each item in so that you can map the inner indexes to the outer indexes.  The inner ones will be lower by the number of cache hits.  That probably looks like a `map[int]int` and a `var index, cacheHit int`.  You add to the map when you send, and lookup/remove in the callback.  Increment count every time something comes in and cacheHit every time you skip something.

That does seem more complicated just to isolate the caching.  I'm generally in favor of internal complexity over surface area complexity i.e. requiring consumers to understand an additional type when calling `NewDoer`.  But if you already tried it and didn't like how it came out then I'll defer to you.

src/internal/storage/fileset/postgres_cache.go
brendoncarroll: I'd say keep it as `V1`.  I think most people will also interpret it that way at first glance.

src/internal/storage/fileset/postgres_cache.go
brendoncarroll: I thought the task IDs were just `pachhash.Output`s?

RE performance: Just 2x the storage and 2x over the wire.  I was a little concerned about that for the chunks and filesets, but this table will be smaller. 

I'm familiar with the human readability argument.  It makes the most sense when you are debugging in a SQL shell.  Encoding can sometimes lead to issues double encoding/unencoding.  Although that's more common in languages other than Go where `[]byte` is clearly used for binary and all the hex strings are `string`s.

src/internal/storage/fileset/postgres_cache.go
brendoncarroll: Ah okay got it.

src/internal/task/task.go
brycemcanally: The in-memory map is a scalability concern and I think the interface becomes more complicated if you need to pass index info to the internal Doer. My thinking is that we will eventually make the interface based on functional options, so you wouldn't really need to understand a cache unless you want to use it (you don't really need to understand it now to not use it). I used to start with an optional parameter for this type of stuff, but I am leaning more towards the explicit setting of the parameter to nil when unused and moving to functional options when more parameters are added over time.

src/internal/storage/fileset/postgres_cache.go
brycemcanally: No, they are hex encoded. We use pachhash, then hex encode immediately and use that string everywhere else. It seems like the benefit of using strings everywhere outweighs the cost of 2x data for keys this small.

src/internal/storage/fileset/postgres_cache.go
```diff
@@ -0,0 +1,122 @@
...
+		accessed_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
```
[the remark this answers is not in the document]
brycemcanally: I am looking into an approach where we have a namespace, a background process that cleans up namespaces when they are no longer needed (i.e. commit / job finishes), and increasing the cache size significantly. The cache size would be large enough and configurable to hold several jobs worth of tasks, but would still be bounded if the background process is cleaning up slowly for some reason.

src/server/worker/pipeline/transform/pending_job.go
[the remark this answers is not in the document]
brycemcanally: I think I am going to punt on this and add a TODO. It does seem like there is a decent bit of shared functionality (similar to the serialize stuff), but I don't see a path forward yet that keeps the added complexity to a minimum.

src/internal/storage/fileset/postgres_cache.go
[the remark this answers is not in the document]
PFedak: sounds good

src/server/worker/pipeline/transform/cache.go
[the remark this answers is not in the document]
PFedak: I was imagining this function could turn into something like
```go
var msg types.DynamicAny
types.UnmarshalAny(output, &msg)
if lister, ok := msg.Message.(FileSetLister); ok {
    fileSetIds = append(fileSetIds, lister.List()...)
}
```
but regardless it doesn't need to be part of this PR

src/server/worker/pipeline/transform/worker.go
PFedak: Agreed, I wasn't questioning the tradeoff, just making sure I wasn't missing anything

src/internal/task/task.go
brendoncarroll: I disagree about having functional options in an interface.  The larger the interface the less useful it is, and functional options greatly increase the parameter space of a given method and therefore the size of the interface.

Going to resolve this.  Keeping the cache stuff mixed in is fine.

src/internal/storage/fileset/postgres_cache.go
brendoncarroll: I think using strings everywhere is actually a drawback in terms of lost typing information.  The `String` method allows the pachhash.Output to print nicely.  And you can always convert to a string at the last minute before you do anything.

With regards to the column type, TEXT is fine.

src/server/worker/pipeline/transform/cache.go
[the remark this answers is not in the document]
brendoncarroll: IIUC this would be adding methods to the protocol buffers result types?  I think that's a good idea, we do something similar with the primitive and composite filesets.

It doesn't actually simplify the Put method does it, because of the Any?  Would we unwrap the Any and then check that it converts to the `FilesetTask` interface, and then error if it doesn't.  So in order to use the fileset cache the inner protocol buffer has to implement the interface.

src/internal/task/task.go
brycemcanally: I might actually agree with you there, although I would want to think through the pros / cons a bit more.

src/server/worker/pipeline/transform/cache.go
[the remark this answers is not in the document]
brycemcanally: Yeah, I think that would just add complexity without any benefit at this point, but it would be a better design if we needed that information elsewhere. I don't think that is likely, but we could easily move to that model if that is ever the case.

src/internal/task/etcd_service.go
```diff
@@ -33,16 +34,16 @@ type etcdService struct {
 func NewEtcdService(etcdClient *etcd.Client, etcdPrefix string) Service {
 	return &etcdService{
 		etcdClient: etcdClient,
-		etcdPrefix: etcdPrefix,
+		etcdPrefix: path.Join(etcdPrefix, version.PrettyVersion()),
```
brendoncarroll, line 37: This is the line. All our versioning problems are solved.

src/internal/task/etcd_service.go
```diff
@@ -206,6 +224,7 @@ func computeTaskID(input *types.Any) (string, error) {
 	if err != nil {
 		return "", errors.EnsureStack(err)
 	}
+	val = append(val, []byte(version.PrettyVersion())...)
```
brendoncarroll, line 227: 👍

PR #7265, opened 2022-01-25
CORE-75 Switch log levels based on gRPC status codes in logging middleware

albscui:
Currently, all errors returned by our system are logged at ERROR level. However, we want to reduce the noise in our logs because some errors are caused by user error, and are not relevant for debugging real problems.

This PR:

- adds the ability to control the log level based on the gRPC status codes returned by the handlers 
- implements `GRPCStatus()` interface for a subset of the pfs server errors
- updates the widely used `EnsureStack(error)` function to return a wrapped error that still implements the GRPCStatus interface so that the middleware can get the status from these errors.
 
Note that it is still up the gRPC handlers to leverage these structured errors, so if a gRPC handler does not return these structured errors, then the logs would still be ERROR level. This PR updated `inspectRepo` and `inspectBranch` to return the appropriate errors, but it's not exhaustive. This PR mostly only provides a foundation for future logging level changes.

### DIFF
--- a/src/server/pfs/server/driver.go
+++ b/src/server/pfs/server/driver.go
@@ -228,7 +228,7 @@ func (d *driver) inspectRepo(txnCtx *txncontext.TransactionContext, repo *pfs.Re
 	}
 	repoInfo := &pfs.RepoInfo{}
 	if err := d.repos.ReadWrite(txnCtx.SqlTx).Get(repo, repoInfo); err != nil {
-		return nil, errors.EnsureStack(err)
+		return nil, errors.EnsureStack(pfsserver.ErrRepoNotFound{Repo: repo})
 	}
 	if includeAuth {
 		resp, err := d.env.AuthServer.GetPermissionsInTransaction(txnCtx, &auth.GetPermissionsRequest{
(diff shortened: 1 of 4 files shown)


jrockway (approved):
This looks very nice.  Check to see if anyone else on the Slack thread wanted to take a look before submitting.

src/server/pfs/server/driver.go
```diff
@@ -228,7 +228,7 @@ func (d *driver) inspectRepo(txnCtx *txncontext.TransactionContext, repo *pfs.Re
 	}
 	repoInfo := &pfs.RepoInfo{}
 	if err := d.repos.ReadWrite(txnCtx.SqlTx).Get(repo, repoInfo); err != nil {
-		return nil, errors.EnsureStack(err)
+		return nil, errors.EnsureStack(pfsserver.ErrRepoNotFound{Repo: repo})
```
albscui, line 231: This makes me slightly uncomfortable, since we are losing the `err` from  `.Get(repo)`, but I'm making an assumption that `.Get(repo)` errors can be translated to `RepoNotFound`. An alternative to this approach could be to add an `error` field to `ErrRepoNotFound`, so that it wraps whatever error is returned by `.Get(repo)`.

brendoncarroll (approved):
LGTM

PR #7271, opened 2022-01-27
Deploy Loki by default

acohen4:


jrockway (approved):
Good idea!  I think our code expects Loki 1.x and not 2.x, but we could fix that.  (Would break Hub though.  Not sure if we care.)

chainlink:
I think we're loki 2.0 compliant actually (we'd need to double check)

We should vendor the loki chart in (similar to what we do with PG) so folks don't need to run the `helm dep` commands

chainlink:
Also a couple of ops things:
- I believe loki deploys as a daemon set by default, which we may want to modify
- Loki will collect all pod logs by default, which we probably want to modify to capture just pachd logs
- We should look into the PV that loki is creating to make sure it's large enough and of sufficient IOPS to be useful.

src/internal/serviceenv/config.go
```diff
@@ -21,8 +21,8 @@ type GlobalConfiguration struct {
 	Namespace                      string `env:"PACH_NAMESPACE,default=default"`
 	StorageRoot                    string `env:"PACH_ROOT,default=/pach"`
 	GCPercent                      int    `env:"GC_PERCENT,default=50"`
-	LokiHost                       string `env:"LOKI_SERVICE_HOST"`
-	LokiPort                       string `env:"LOKI_SERVICE_PORT"`
+	LokiHost                       string `env:"PACHYDERM_LOKI_SERVICE_HOST"`
```
acohen4, line 24: One thing I don't like about this right now, is that it forces the client to use our subchart, and restricts them from separately deploying loki.

acohen4:
> I think we're loki 2.0 compliant actually (we'd need to double check)
> 
> We should vendor the loki chart in (similar to what we do with PG) so folks don't need to run the `helm dep` commands

My understanding is that it's not necessary since we're including the tar.gz file in our `/charts` dir

acohen4:
> * I believe loki deploys as a daemon set by default, which we may want to modify

it deploys promtail as a daemonset so that it can push logs from each node in the cluster. I think we probably won't be able to get away from this

> * Loki will collect all pod logs by default, which we probably want to modify to capture just pachd logs

good point 👍 need to investigate this

> * We should look into the PV that loki is creating to make sure it's large enough and of sufficient IOPS to be useful.

also good point - I need to make a few more changes for this

brycemcanally:
We should look into deploying it with the storage being backed by the object storage bucket that is created for Pachyderm (that is an option in Loki 2.0 with the BoltDB shipper storage). Also, be sure that if we have a filter for log collection, it doesn't filter out any of the pods / containers we care about (i.e. user container in worker pods and etcd / pgbouncer pods). I think we have a common label that is called `suite=pachyderm` that can be used for filtering.

src/internal/serviceenv/config.go
chainlink: we could pass a name override to the loki chart perhaps

acohen4:
@chainlink so the nice news is that we can configure the loki-stack chart as before, by prefixing all of it's values with the `loki-stack.` helm object. I included persistence + size with defaults. For example, modifying the pvc size looks like `loki-stack.loki.persistence.size="5Gi"`

src/internal/serviceenv/config.go
acohen4: how does that work?

etc/helm/pachyderm/values.yaml
```diff
@@ -180,8 +180,22 @@ ingress:
       create: false
       crt: ""
       key: ""
+
 loki:
   enabled: true
+
+# locki-stack contains values that will be passed to the loki-stack subchart
```
chainlink, line 184: Should we add a note on how to set the storageclass here?

acohen4:
> Good idea! I think our code expects Loki 1.x and not 2.x, but we could fix that. (Would break Hub though. Not sure if we care.)

@jrockway I think getting this to work on hub will definitely require some work, at the very least because the k8s exposed env-var for service discovery changes. Before merging, should we discuss the ops changes required for having this work on hub? Alternatively, we could make the values of those env-vars configurable so hub has to do little to nothing to upgrade

etc/helm/pachyderm/values.yaml
acohen4: added 👍

src/internal/serviceenv/config.go
acohen4: so we can repurpose these config values to LokiHost, LokiPort to store the env-var name instead - maybe change their names to LokiHostVar and LokiPortVar

acohen4:
> > Good idea! I think our code expects Loki 1.x and not 2.x, but we could fix that. (Would break Hub though. Not sure if we care.)
> 
> @jrockway I think getting this to work on hub will definitely require some work, at the very least because the k8s exposed env-var for service discovery changes. Before merging, should we discuss the ops changes required for having this work on hub? Alternatively, we could make the values of those env-vars configurable so hub has to do little to nothing to upgrade

I made it so that pachd uses the env-var to figure out which unix variable to reference. This note should no longer be of concern

chainlink (approved):

PR #7272, opened 2022-01-28
[ENT-39] Implement pause and unpause

robert-uhl:
This PR adds pause and unpause functionality as part of the enterprise feature set. Pause quiesces a Pachyderm cluster, terminating all worker pods and restarting all `pachd` pods in a special paused state; unpause reverses this.

It adds two subcommands to `pachctl`: `pachctl enterprise pause` and `pachctl enterprise unpause`.


src/auth/auth.proto
```diff
@@ -171,6 +171,7 @@ enum Permission {
   CLUSTER_ENTERPRISE_HEARTBEAT           = 115;
   CLUSTER_ENTERPRISE_GET_CODE            = 116;
   CLUSTER_ENTERPRISE_DEACTIVATE          = 117;
+  CLUSTER_ENTERPRISE_PAUSE               = 149;
```
acohen4, line 174: Missing `CLUSTER_ENTERPRISE_UNPAUSE`?

src/server/enterprise/server/api_server.go
```diff
@@ -330,3 +345,95 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
 }
+
+func (a *apiServer) Pause(ctx context.Context, req *ec.PauseRequest) (resp *ec.PauseResponse, retErr error) {
+	if req.Namespace == "" {
+		req.Namespace = "default"
+	}
+	if err := a.setPauseState(ctx, req.Namespace, true); err != nil {
+		return nil, err
+	}
+	if err := scaleDownWorkers(ctx, a.env.GetKubeClient(), req.Namespace); err != nil {
```
acohen4, line 356: I think there may also be a race condition here, where we scale down the workers before pachds. I can imagine a situation where the pachds change the values of the worker RCs back to the spec before they are rolled over. 

I'm not sure what the best solution is here. We could scale down on startup, but that makes for a confusing control flow

src/server/enterprise/server/api_server.go
```diff
@@ -330,3 +345,95 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+		return nil, errors.EnsureStack(err)
+	}
+	if err := rollPachd(ctx, a.env.GetKubeClient(), req.Namespace); err != nil {
+		return nil, errors.EnsureStack(err)
+	}
+
+	return &ec.PauseResponse{}, nil
```
acohen4, line 363: maybe a naive question - but is it possible for the pod to restart before we return the response?

robert-uhl:
Re. changing `Deployment.Spec.Template.Annotations`, that does indeed cause the deployment to execute a rolling restart of each `pachd` pod. By setting the annotation to the current timestamp, one knows that it will be a change.

robert-uhl:
I implemented graceful stop of the gRPC server in order to let `Pause` return.

brendoncarroll:
IMO this is fundamentally the wrong approach.  Normally high level feedback would be provided in a design doc, but the first I saw the doc or this PR was this morning.  I saw a doc for the Velero backup stuff, but nothing for pause/unpause.  Since there is already an implementation, I'm going to provide feedback here.

Having pachd scale down itself is difficult to reason about for the same reasons self-modifying code is difficult to reason about.  It's usually possible to express what you want with a system that doesn't self-modify.  Both the enterprise server and pachctl could do this easily and sanely.

It looks like we decide whether the cluster is paused based on reading a value out of the database at startup? That seems like it could eventually give the wrong value if you called Pause then Unpause in between container restarts.  Setting a parameter or environment variable is a better approach.  In order to keep track of whether the pachds are done switching, we would need to write back to the database once they are done switching modes, for each pachd, not just the one serving the RPC.  The source of truth for the pachd mode should always be its k8s configuration.

An RPC called Pause should block until the cluster is paused, and should be idempotent and retryable.  This will make it easy for the caller to move onto whatever is next, knowing that pause is done.  On a non-pachd system this could be implemented as polling the k8s API or waiting for events.

Pachd does a lot of things and has a lot of people changing it's source at once; I don't think an approach where you have to go around trying to stop everything is a good idea.  The shutdown path for all of these services is totally untested (because it hasn't been important), and it's too easy to leak worker or master background processes and totally break this feature.

A better approach is to use the enterprise server, which IIRC receives heartbeats from all the clusters it manages, and could connect back to them if it wanted to.  Pause scales pachd to 0 and makes sure that's done, then returns.  Pachctl could also work for single cluster deployment for people who don't have an enterprise server.  I don't thing k8s access is really that much to ask considering this operation will probably done by administrators prior to upgrading (which also needs k8s access).  Data scientists will probably not be backing up pachyderm themselves.  If they aren't backing up postgres and object storage, why do they need to use Pause/Unpause?

As far as this implementation goes:
Sprinkling branches throughout the initialization does not inspire confidence that everything which needs to be disabled has been disabled.  We should create another pachd mode called pause, spin up just the enterprise server, database and external API to serve the unpause RPC.  This would be easy to do if pause was just another pachd mode.  Alternatively you could initialize a database connection sooner.

robert-uhl:
Per our meeting this morning, I’ll rework this PR:

 - Rather than use a database value, pachd will take a command line argument, which will be set by the redeployment machinery.
 - This will start it in paused mode, which will be an independent mode of operation.
 - The Pause & Unpause RPCs will be idempotent.
 - I will add an additional RPC to indicate whether the workspace is paused, partially paused (i.e., some pachds running in paused mode and some not) or unpaused.
 - I will add license & identity APIs to the paused mode API set.

robert-uhl:
> 1. Summarize the approach in a "Design Doc" and share it in eng for feedback.
> 2. Add some integration tests

Both done.

> Design Question: Right now it looks to me like we aren't scaling down pachds, mainly pausing them from doing background work and scaling down their workers. Is this intentional?

Correct. It _would_ be possible to scale `pachd` down to one pod, and then scale back up, but I figured that the additional complexity is probably not worth it, especially given our longer-term direction to do this remotely, from the enterprise server.

> What are the major costs that customers have in a cluster that's driving this feature - is it the "work" done by pachd, or the pods themeselves? Is this something we'd do in the future?

In this case, we’re not worried about cost but mutable state. If cost does end up being a driver for pause, then I think it would be appropriate to consider scaling `pachd` down to a single pod.

src/server/enterprise/server/api_server.go
```diff
@@ -330,3 +341,114 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+	}
+	// Updating the spec rolls the deployment, killing each pod and causing
+	// a new one to start.
+	for i, c := range d.Spec.Template.Spec.Containers {
+		if len(c.Command) == 0 {
+			continue
+		}
+		if c.Command[0] == "/pachd" {
+			if paused {
+				c.Command = []string{"/pachd", "--mode", "paused"}
```
acohen4, line 370: would this interact in a weird way with helm? Wouldn't an upgrade overwrite these values? I think recording the state via a change on the container spec may not be the way to go. Would a configmap make more sense if we're using kubernetes/env as a place to store this info?

etc/helm/pachyderm/templates/pachd/rbac/role.yaml
```diff
@@ -47,4 +47,15 @@ rules:
...
+- apiGroups:
+  - "apps"
+  - "extensions"
+  resources:
+  - deployments
+  - deployments/scale
+  verbs:
+  - get
+  - list
+  - patch
```
acohen4, line 59: just a reminder that we should be able to remove these perms in case we decide not to modify the deployments

src/server/cmd/pachd/main.go
```diff
@@ -924,7 +930,331 @@ func doFullMode(config interface{}) (retErr error) {
...
+				true,
+			)
+			if err != nil {
+				return err
+			}
+			eprsclient.RegisterAPIServer(externalServer.Server, enterpriseAPIServer)
+			env.SetEnterpriseServer(enterpriseAPIServer)
+			// Stop workers because unpaused pachds in the process
+			// of rolling may have started them back up.
+			if err := e.StopWorkers(ctx); err != nil {
```
acohen4, line 1075: should we have this call be done out of the enterprise service? Since this feature relies on restarts and startup-time logic I'm contemplating whether the logic would be easier to follow if we handle this at the enterprise service startup. WDYT?

acohen4 (changes requested):

src/server/cmd/pachd/main.go
robert-uhl: I’m not opposed to that, but would it live in `eprsserver.NewEnterpriseServer`, or in a new method on the enterprise server? If so, wouldn’t that new method basically be a renamed `StopWorkers`? Maybe that’s a good thing, because it abstracts the exact knowledge out a bit?

etc/helm/pachyderm/templates/pachd/rbac/role.yaml
robert-uhl: Still need them for the deployment rolling portion.

src/server/enterprise/server/api_server.go
robert-uhl: I’ve rewritten the code to use a ConfigMap unmanaged by Helm (which thus will not be overwritten by Helm operations).

src/server/enterprise/server/api_server.go
```diff
@@ -330,3 +343,192 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
 
 	return &ec.DeactivateResponse{}, nil
 }
+
+func (a *apiServer) Pause(ctx context.Context, req *ec.PauseRequest) (resp *ec.PauseResponse, retErr error) {
```
acohen4, line 347: We should account for this API call being called on a cluster running in `enterprise` mode. I think right now we would transition such a cluster through `enterprise -> paused -> full`. If it's easiest, I think preventing pausing of an enterprise mode cluster would be fine for now

src/server/enterprise/server/api_server.go
```diff
@@ -330,3 +343,192 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+	if err := scaleDownWorkers(ctx, a.env.GetKubeClient(), a.env.Namespace); err != nil {
```
acohen4, line 348: Do we still need this call here since it's scaling down workers on startup of a cluster in `paused` mode?

src/server/enterprise/server/api_server.go
```diff
@@ -330,3 +343,192 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+	if err := rollPachd(ctx, a.env.GetKubeClient(), a.env.Namespace, true); err != nil {
+		return nil, errors.EnsureStack(err)
+	}
+
+	return &ec.PauseResponse{}, nil
+}
+
+var updatedAtFieldName = "pachyderm.com/updatedAt"
+
+func rollPachd(ctx context.Context, kc *kubernetes.Clientset, namespace string, paused bool) error {
```
acohen4, line 360: I think adding a comment here highlighting the approach would be helpful

src/server/cmd/pachd/main.go
```diff
@@ -924,7 +936,333 @@ func doFullMode(config interface{}) (retErr error) {
...
+		server.TLSConfig = &gotls.Config{GetCertificate: cLoader.GetCertificate}
+		return errors.EnsureStack(server.ListenAndServeTLS(certPath, keyPath))
+	})
+	go waitForError("Prometheus Server", errChan, requireNoncriticalServers, func() error {
+		http.Handle("/metrics", promhttp.Handler())
+		return errors.EnsureStack(http.ListenAndServe(fmt.Sprintf(":%v", env.Config().PrometheusPort), nil))
+	})
+	go func(c chan os.Signal) {
+		<-c
+		log.Println("terminating; waiting for enterprise server to gracefully stop")
```
acohen4, line 962: this message should read for "pachd server" right?

src/server/cmd/pachd/main.go
```diff
@@ -924,7 +936,333 @@ func doFullMode(config interface{}) (retErr error) {
...
+		var g, _ = errgroup.WithContext(ctx)
+		g.Go(func() error { externalServer.Server.GracefulStop(); return nil })
+		g.Go(func() error { internalServer.Server.GracefulStop(); return nil })
+		g.Wait()
+		log.Println("gRPC server gracefully stopped")
+	}(interruptChan)
+	return <-errChan
+}
+
+func doPausedMode(config interface{}) (retErr error) {
```
acohen4, line 972: What you have here makes sense to me, but it would be easier to catch issues if we didn't need to duplicate so much code between modes. I think it could be nice to refactor this file a bit in a separate PR. That being said it's up to you, if you don't want to handle that right now.

src/server/enterprise/server/enterprise_test.go
```diff
@@ -302,3 +302,98 @@ func TestEnterpriseConfigMigration(t *testing.T) {
...
+			return errors.Errorf("could not get pause status %v", err)
+		}
+		if resp.Status == enterprise.PauseStatusResponse_UNPAUSED {
+			return nil
+		}
+		return errors.Errorf("status: %v", resp.Status)
+	}, bo)
+}
+
+func TestDoublePauseUnpause(t *testing.T) {
```
acohen4, line 354: Maybe one more test case would be to set Paused + Unpaused in succession, so that the second command would complete before the first command causes a restart

acohen4 (changes requested):

go.mod
```diff
@@ -173,6 +173,7 @@ require (
 	github.com/klauspost/cpuid v1.3.1 // indirect
 	github.com/kr/pretty v0.2.1 // indirect
 	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de // indirect
+	github.com/lithammer/dedent v1.1.0 // indirect
```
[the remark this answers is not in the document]
robert-uhl: Because I neglected to run `go mod tidy` … fixing.

src/server/enterprise/server/api_server.go
robert-uhl: I think with the ConfigMap approach the enterprise server is safe, as it gets started with a verbatim `--mode=enterprise`, with no reference to the ConfigMap value.

It does make sense to generate an error when trying to pause an enterprise server, though. Adding that …

src/server/enterprise/server/api_server.go
robert-uhl: I thought of it as belt-and-suspenders, on the theory that it doesn’t hurt to do it multiple times (and in fact with a multiple-`pachd` cluster the scale-down will happen once for each `pachd`). Thoughts?

src/server/cmd/pachd/main.go
```diff
@@ -924,7 +936,333 @@ func doFullMode(config interface{}) (retErr error) {
...
+	}
+
+	// must run InstallJaegerTracer before InitWithKube/pach client initialization
+	if endpoint := tracing.InstallJaegerTracerFromEnv(); endpoint != "" {
+		log.Printf("connecting to Jaeger at %q", endpoint)
+	} else {
+		log.Printf("no Jaeger collector found (JAEGER_COLLECTOR_SERVICE_HOST not set)")
+	}
+	env := serviceenv.InitWithKube(serviceenv.NewConfiguration(config))
+	profileutil.StartCloudProfiler("pachyderm-pachd-full", env.Config())
```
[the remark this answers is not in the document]
robert-uhl: Probably not — fixing.

src/server/cmd/pachd/main.go
robert-uhl: I agree, but I think that should be a separate PR. Also, the tricky thing is that they are almost but not quite the same, and I worry that a `doAllModes(mode, config)` might end up being its own kind of spaghetti. I think that there is probably an elegant solution, but it requires some thought & planning.

src/client/transaction.go
```diff
@@ -335,3 +335,311 @@ func (c *ppsBuilderClient) CreatePipeline(ctx context.Context, req *pps.CreatePi
 	c.tb.requests = append(c.tb.requests, &transaction.TransactionRequest{CreatePipeline: req})
 	return nil, nil
 }
+
+func (c *pfsBuilderClient) ActivateAuth(ctx context.Context, req *pfs.ActivateAuthRequest, opts ...grpc.CallOption) (*pfs.ActivateAuthResponse, error) {
```
acohen4, line 339: @robert-uhl is this generated? Could you elaborate on what's going on here and how it relates to the other changes?

src/server/enterprise/server/api_server.go
acohen4: With the spirit of making the logic easy to follow, I think that if it's not necessary for this feature to behave correctly, we shouldn't include it. 

To the second point, it isn't necessarily a "good thing" that all pachd's call this - perhaps if there were a lot of pachds, all the calls could overwhelm the Kube API Server. I don't know (and don't think) if that's a realistic concern.

src/server/enterprise/server/api_server.go
acohen4: I think it may still be worth parameterizing it such that when you "unpause", the previous mode is reset as opposed to the hard-coded "full" mode. 

The case I have in mind, is that pretty soon we'll want to run pachd in a mode where the identity gRPC service is turned off (in the case where it's relying on the identity service of the enterprise server). Perhaps there are other ways of configuring it, but this could naturally be implemented with a corresponding mode which would be different than full.

src/server/enterprise/server/enterprise_test.go
```diff
@@ -302,3 +302,136 @@ func TestEnterpriseConfigMigration(t *testing.T) {
...
+		}
+		if resp.Status == enterprise.PauseStatusResponse_PAUSED {
+			return nil
+		}
+		return errors.Errorf("status: %v", resp.Status)
+	}, bo)
+
+	// ListRepo should return an error since the cluster is paused now
+	_, err = client.ListRepo()
+	require.NotNil(t, err)
```
acohen4, line 341: nit: for consistency, we should use require.NoError() here

src/server/enterprise/server/api_server.go
robert-uhl: The previous mode could be stored as another value in the ConfigMap, then used to replace `MODE` on an unpause. What do you think of that as a solution?

src/client/transaction.go
robert-uhl: I believe that I picked the wrong leg of a diff when merging.

I’ve redone the merge.

src/server/enterprise/server/enterprise_test.go
robert-uhl: In this case, an error is expected (thus the `NotNil`), so I think it would be `require.YesError`. Updating.

src/server/enterprise/server/api_server.go
robert-uhl: This is implemented now — let me know what you think.

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/fsouza/go-dockerclient v1.4.1
 	github.com/go-sql-driver/mysql v1.6.0
 	github.com/gogo/protobuf v1.3.2
+	github.com/golang/protobuf v1.5.2
```
acohen4, line 28: where is this getting used? We probably want to be using gogo proto referenced above

src/client/transaction.gen.go
```diff
@@ -471,6 +417,20 @@ func (c *unsupportedPpsBuilderClient) ListTask(_ context.Context, _ *taskapi.Lis
 	return nil, unsupportedError("ListTask")
 }
 
+type unsupportedDebugBuilderClient struct{}
```
acohen4, line 420: this file should also be reverted right?

src/internal/middleware/auth/interceptor.go
```diff
@@ -77,6 +77,9 @@ var authHandlers = map[string]authHandler{
 	"/enterprise_v2.API/GetActivationCode": authDisabledOr(clusterPermissions(auth.Permission_CLUSTER_ENTERPRISE_GET_CODE)),
 	"/enterprise_v2.API/Deactivate":        authDisabledOr(clusterPermissions(auth.Permission_CLUSTER_ENTERPRISE_DEACTIVATE)),
 	"/enterprise_v2.API/Heartbeat":         authDisabledOr(clusterPermissions(auth.Permission_CLUSTER_ENTERPRISE_HEARTBEAT)),
+	"/enterprise_v2.API/Pause":             authDisabledOr(clusterPermissions(auth.Permission_CLUSTER_ENTERPRISE_PAUSE)),
+	"/enterprise_v2.API/Unpause":           authDisabledOr(clusterPermissions(auth.Permission_CLUSTER_ENTERPRISE_PAUSE)),
+	"/enterprise_v2.API/PauseStatus":       authDisabledOr(clusterPermissions(auth.Permission_CLUSTER_ENTERPRISE_PAUSE)),
```
acohen4, line 82: The UX for a non-Admin learning that their cluster is paused seems like it may be lacking. It might make sense to have the PauseStatus API be available for users other than ClusterAdmin. I think it would be worth clarifying this with product

src/server/enterprise/server/api_server.go
```diff
@@ -332,3 +346,237 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+		return nil, errors.Errorf("cannot pause a sidecar or enterprise server")
+	}
+	if err := a.rollPachd(ctx, true); err != nil {
+		return nil, errors.EnsureStack(err)
+	}
+
+	return &ec.PauseResponse{}, nil
+}
+
+const updatedAtFieldName = "pachyderm.com/updatedAt"
```
acohen4, line 363: this should belong in the const block at the top of the file

src/server/enterprise/server/api_server.go
```diff
@@ -332,3 +346,237 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
 
 	return &ec.DeactivateResponse{}, nil
 }
+
+// Pause sets the cluster to paused mode, restarting all pachds in a paused
+// status.  If they are already paused, it is a no-op.
+func (a *apiServer) Pause(ctx context.Context, req *ec.PauseRequest) (resp *ec.PauseResponse, retErr error) {
+	if a.env.Mode == SidecarMode || a.env.Mode == EnterpriseMode {
```
acohen4, line 353: I think it may make more sense to make this an allow list as opposed to a disallow list. i.e. `if a.env.Mode != "full" then error`. This way if new modes are added in the future without knowledge of this feature, we protect them from being paused unless desired

src/server/enterprise/server/api_server.go
```diff
@@ -388,11 +396,8 @@ func rollPachd(ctx context.Context, kc *kubernetes.Clientset, namespace string,
 			},
 			Data: make(map[string]string),
 		}
-		if paused {
-			c.Data["MODE"] = "paused"
-		} else {
-			c.Data["MODE"] = "full"
-		}
+		c.Data["MODE"] = "paused"
+		c.Data["previous-mode"] = "full"
```
acohen4, line 400: I don't think we know that we're in "full" mode at this point

src/server/enterprise/server/api_server.go
```diff
@@ -332,3 +346,237 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+		// Since pachd-config is not managed by Helm, it may not exist.
+		c = &v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "pachd-config",
+				Namespace: namespace,
+			},
+			Data: make(map[string]string),
+		}
+		c.Data["MODE"] = "paused"
+		c.Data["previous-mode"] = "full"
```
acohen4, line 400: I don't think we know that we're in "full" mode at this point

src/server/enterprise/server/api_server.go
```diff
@@ -332,3 +346,237 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+	} else {
+		// Curiously, Get can return no error and an empty configmap!
+		// If this happens, need to set the name and namespace.
+		if c.ObjectMeta.Name == "" {
+			c.ObjectMeta = metav1.ObjectMeta{
+				Name:      "pachd-config",
+				Namespace: namespace,
+			}
+		}
+		if c.Data == nil {
```
acohen4, line 415: I'm not sure I understand this logic. Wouldn't `c.Data["previous-mode"] != "paused"` here if the configmap is empty?

src/server/enterprise/server/api_server.go
```diff
@@ -332,3 +346,237 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+			// Short-circuit and do not update if configmap
+			// is already in the correct state.  This keeps
+			// the updatedAt semantics correct when checking
+			// the pause status.
+			return nil
+		}
+		if c.Data["previous-mode"] == "" {
+			return errors.Errorf("empty previous-mode in pachd-config ConfigMap")
+		}
+		mode := c.Data["MODE"]
```
acohen4, line 436: I think I understand my confusion here better. I think what was hard for me to follow is the use of "previous-mode" key as having an initial value of "paused" which isn't semantically true. I think it would be clearer if the logic here was explicit, i.e.:
`if paused then c.Data["previous-mode] = c.Data["MODE"]; c.Data["MODE"] = "paused"; else c.Data["MODE"] = originalMode`

Maybe if we were able to capture the original mode the semantics here would be simpler. What do you think?

src/server/enterprise/server/api_server.go
```diff
@@ -332,3 +346,237 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+		if _, err := cc.Update(ctx, c, metav1.UpdateOptions{}); err != nil {
+			return errors.Errorf("could not update configmap: %w", err)
+		}
+	}
+
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		dd := kc.AppsV1().Deployments(namespace)
+		d, err := dd.Get(ctx, "pachd", metav1.GetOptions{})
+		if err != nil {
+			return err //nolint:wrapcheck
```
acohen4, line 452: what's `//nolint:wrapcheck`?

src/server/enterprise/server/api_server.go
```diff
@@ -332,3 +346,237 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+	var sawAfter, sawBefore bool
+	for _, p := range pp.Items {
+		if p.CreationTimestamp.Time.Before(updatedAt) {
+			sawBefore = true
+		} else {
+			sawAfter = true
+		}
+	}
+	switch {
+	case sawBefore && sawAfter:
```
acohen4, line 555: nit: I think setting the correct value of `PauseStatus` in the switch statement and handling return after would save a bunch of lines of code. Up to you

src/server/cmd/pachd/main.go
```diff
@@ -926,7 +948,333 @@ func doFullMode(config interface{}) (retErr error) {
...
+			}
+			authclient.RegisterAPIServer(externalServer.Server, authAPIServer)
+			env.SetAuthServer(authAPIServer)
+			return nil
+		}); err != nil {
+			return err
+		}
+		if err := logGRPCServerSetup("Enterprise API", func() error {
+			e := eprsserver.EnvFromServiceEnv(env, path.Join(env.Config().EtcdPrefix, env.Config().EnterpriseEtcdPrefix), txnEnv)
+			e.Mode = eprsserver.PausedMode
```
acohen4, line 1083: why do we need to set the paused mode here?

src/server/cmd/pachd/main.go
```diff
@@ -864,8 +884,10 @@ func doFullMode(config interface{}) (retErr error) {
 			return err
 		}
 		if err := logGRPCServerSetup("Enterprise API", func() error {
+			e := eprsserver.EnvFromServiceEnv(env, path.Join(env.Config().EtcdPrefix, env.Config().EnterpriseEtcdPrefix), txnEnv)
+			e.Mode = eprsserver.UnpausedMode
```
acohen4, line 888: can we make the default value for `e.Mode = eprsserver.UnpausedMode` so that we don't have to explicitly set it besides in paused mode?

go.mod
robert-uhl: Looks like it’s in `pachgen`:

```
etc/proto/pachgen/main.go
9:      descriptor "github.com/golang/protobuf/protoc-gen-go/descriptor"
10:     plugin "github.com/golang/protobuf/protoc-gen-go/plugin"

etc/proto/pachgen/transaction_client.go
11:     descriptor "github.com/golang/protobuf/protoc-gen-go/descriptor"
12:     plugin "github.com/golang/protobuf/
```

src/client/transaction.gen.go
robert-uhl: Regenerated, I think, rather than reverted. I ran `make proto` again, and am adding a commit with its updates.

src/server/enterprise/server/api_server.go
robert-uhl: I believe so. This is within the `if k8serrors.IsNotFound(err)` block, which means that there is no ConfigMap. `rollPachd` is only called for full or paused mode, and if there is no ConfigMap then the cluster has to be in full mode.

src/server/enterprise/server/api_server.go
robert-uhl: In this case, `rollPachd` is creating the configmap as it should currently exist, with a non-nil Data member and `MODE`/`previous-mode` (incidentally, that naming is to emphasise that `previous-mode` is not intended to be used as an environment variable). The current, pre-roll state of an empty configmap is equivalent to full mode, paused previous mode.

This is needed for the `paused && c.Data["MODE"] == "paused"`/`if !paused && c.Data["MODE"] != "paused"` conditionals and `c.Data["MODE"] = c.Data["previous-mode"]` later on.

src/server/enterprise/server/api_server.go
robert-uhl: Would it help if rather than constructing a new ConfigMap twice in similar but inverted ways, there were a function to create a new base-state ConfigMap, and a function to invert it? I think that might make the logic clearer.

src/server/cmd/pachd/main.go
robert-uhl: The zero value is unpaused, so this works, but I think that explicit is probably better than implicit. I can kinda see both sides though.

acohen4: I think we should make paused opt-in for future modes, so if we could avoid forcing all modes to implement the "base logic" for the feature, I think that would be preferable. WDYT?

src/internal/middleware/auth/interceptor.go
pappasilenus: If this is a status API that merely allows the user to find out if the cluster is paused, that's a great point. Any user with cluster access should be able to know the cluster state.  I believe we may have a little feature debt in pachctl that preceeds any API call with a "Hey, is the cluster paused?" call so that we can give them an informative error message rather than just hanging.  How can we best indicate that debt and get the work in the queue?

robert-uhl: I think that can be handled with a dedicated PPS & PFS implementation which returns an appropriate error code for every API call. Looking at https://pkg.go.dev/google.golang.org/grpc/codes, I think a decent argument can be made for either `FailedPrecondition` or `Unavailable`. I would like to do this as a follow-on story though.

src/server/cmd/pachd/main.go
robert-uhl: Hmmm, I could change `server.PauseMode` to enumerate pausable modes (+paused mode) instead of all possible modes, and have the zero value be `UnpausableMode`. Then `Pause` and `Unpause` could key off of that.

The end result would be that one need not implement anything for a non-pausable mode.

src/internal/middleware/auth/interceptor.go
brendoncarroll: If the cluster is paused the other API servers (PFS & PPS) won't be active.  So GRPC will probably say the method doesn't exist.  We can change what is printed to the user in pachctl, but I don't think we should change the fact that those services aren't even running.

robert-uhl:
Offline discussion delved into the need to support pausable modes
other than full and the interplay with Helm, which may not manage the
pause state because we do not wish to reset the pause state after a
Helm operation.  The issue here is determining which state to unpause
into.  This is what we decided on:

 - A new environment variable `UNPAUSED_STATE` which indicates
   the original start-up state of the cluster.
 - The `Pause` RPC will no longer track the previous mode of the
   cluster; it will just set the current mode to paused if it is not
   already set.
 - The `Unpause` RPC will set the current mode of the cluster to the value of
   `UNPAUSED_STATE`, or if it is unset to full mode.
   - The value of `UNPAUSED_STATE` will be plumbed through to
     `Unpause` from the `main` package via `server.EnvFromServerEnv`.
 - Helm will be used to set `UNPAUSED_STATE` appropriately.

src/server/enterprise/server/api_server.go
acohen4: looks good!

src/server/enterprise/server/api_server.go
```diff
@@ -332,3 +348,228 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+	return nil
+}
+
+func newPachdConfigMap(namespace string) *v1.ConfigMap {
+	return &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "pachd-config",
+			Namespace: namespace,
+		},
+		Data: map[string]string{"MODE": "full"},
```
acohen4, line 461: this should be set to env.UnpausedMode right?

src/server/enterprise/server/api_server.go
```diff
@@ -332,3 +348,228 @@ func (a *apiServer) Deactivate(ctx context.Context, req *ec.DeactivateRequest) (
...
+		}); err != nil {
+			return errors.Errorf("could not scale down %s: %v", w.GetName(), err)
+		}
+	}
+	return nil
+}
+
+func (a *apiServer) Unpause(ctx context.Context, req *ec.UnpauseRequest) (resp *ec.UnpauseResponse, retErr error) {
+	if a.env.mode == UnpausableMode {
+		return nil, errors.Errorf("this pachd instance is not in a pausable mode")
```
acohen4, line 489: change error message to "pachd is already unpaused."

src/server/enterprise/server/env.go
```diff
@@ -24,24 +26,65 @@ type Env struct {
...
+
+const (
+	UnpausableMode PauseMode = iota
+	FullMode
+	PausedMode
+)
+
+type Option func(Env) Env
+
+func WithEtcdPrefix(etcdPrefix string) Option {
```
acohen4, line 48: hey sorry if I was sending mixed messages here, but I think etcd prefix is not optional. If so then I think it's better to have been passed as an argument as before

src/internal/middleware/auth/interceptor.go
acohen4: We could add logic to pachctl that would run the PauseStatus API under the condition that it gets "service doesn't exist" error. Agreed those changes should be handled in a separate PR

acohen4, line 82: Instead of creating another permission here, you can just set the value in the map to `authenticated` which will allow any cluster user to run the PauseStatus API

src/server/auth/server/roles.go
```diff
@@ -50,6 +50,7 @@ func init() {
 			auth.Permission_REPO_ADD_PIPELINE_READER,
 			auth.Permission_REPO_REMOVE_PIPELINE_READER,
 			auth.Permission_PIPELINE_LIST_JOB,
+			auth.Permission_CLUSTER_ENTERPRISE_PAUSE_STATUS,
```
acohen4, line 53: if we go with the `authenticated` approach mentioned, we can remove the `auth.Permission_CLUSTER_ENTERPRISE_PAUSE_STATUS`

src/internal/testutil/local/pachd.go
```diff
@@ -201,7 +201,7 @@ func RunLocal() (retErr error) {
 		}
 		if err := logGRPCServerSetup("Enterprise API", func() error {
 			enterpriseAPIServer, err := eprsserver.NewEnterpriseServer(
-				eprsserver.EnvFromServiceEnv(env, path.Join(env.Config().EtcdPrefix, env.Config().EnterpriseEtcdPrefix), txnEnv), true)
+				eprsserver.EnvFromServiceEnv(env, txnEnv, eprsserver.WithEtcdPrefix(path.Join(env.Config().EtcdPrefix, env.Config().EnterpriseEtcdPrefix))), true)
```
acohen4, line 204: I think we can revert these changes if we always require an etcd prefix

src/server/enterprise/server/env.go
```diff
@@ -24,24 +26,65 @@ type Env struct {
...
-func EnvFromServiceEnv(senv serviceenv.ServiceEnv, etcdPrefix string, txEnv *txnenv.TransactionEnv) Env {
-	return Env{
+func WithUnpausedMode(mode string) Option {
+	return func(e Env) Env {
+		e.UnpausedMode = mode
+		return e
+	}
+}
+
+func WithMode(mode PauseMode) Option {
```
acohen4, line 62: why do we need mode & unpaused mode? it doesn't look like e.mode gets used right?

src/server/enterprise/server/api_server.go
robert-uhl: No, this is the argument passed to `--mode`, so it needs to be ∈{`full`, `paused`, `enterprise`, `sidecar`}.

robert-uhl: Weeell, it will get overwritten before being saved in every current code path, but … you are right, and I’ve fixed it so future code will be correct.

src/server/enterprise/server/api_server.go
robert-uhl: That is true but it is not an error to unpause a full-mode cluster; in this case the error really is that the cluster can neither be paused nor unpaused. Maybe it is clearer to state ‘this pachd instance is not in an unpausable mode’?

src/server/enterprise/server/env.go
robert-uhl: No, it’s used at ll. 189 & 312, in `Activate` and `Deactivate`, to disable activation & deactivation when in a paused status.

src/server/enterprise/server/api_server.go
acohen4: works for me!

src/server/enterprise/server/env.go
acohen4: ah got it! My b 👍

acohen4 (approved):
LGTM! Thanks for working through all the back and forth!

PR #7276, opened 2022-02-01
Make Cron Auth work

acohen4:
This PR makes it so that Cron pipelines work with Auth. 

The fundamental issue is that the pipeline auth subject, needs Write permissions to an input repo. I've added an internal transaction API to do so. 

I also added a test case that creates a Cron pipeline, and turns on auth

### DIFF
--- a/src/server/auth/server/api_server.go
+++ b/src/server/auth/server/api_server.go
@@ -713,6 +713,19 @@ func (a *apiServer) AddPipelineReaderToRepoInTransaction(txnCtx *txncontext.Tran
 	return a.setUserRoleBindingInTransaction(txnCtx, &auth.Resource{Type: auth.ResourceType_REPO, Name: sourceRepo}, auth.PipelinePrefix+pipeline, []string{auth.RepoReaderRole})
 }
 
+// AddPipelineReaderToRepoInTransaction gives a pipeline access to write data to the specified source repo.
+// The only time a pipeline needs write permission for a source repo is in the case of Cron inputs.
+// This is distinct from ModifyRoleBinding because AddPipelineWriter is a less expansive permission
+// that is included in the repoWriter role, versus being able to modify all role bindings which is
+// part of repoOwner. This method is for internal use and is not exposed as an RPC.
+func (a *apiServer) AddPipelineWriterToSourceRepoInTransaction(txnCtx *txncontext.TransactionContext, sourceRepo, pipeline string) error {
+	// Check that the user is allowed to add a pipeline to write to the output repo.
+	if err := a.CheckRepoIsAuthorizedInTransaction(txnCtx, &pfs.Repo{Type: pfs.UserRepoType, Name: sourceRepo}, auth.Permission_REPO_ADD_PIPELINE_WRITER); err != nil {
+		return err
+	}
+	return a.setUserRoleBindingInTransaction(txnCtx, &auth.Resource{Type: auth.ResourceType_REPO, Name: sourceRepo}, auth.PipelinePrefix+pipeline, []string{auth.RepoWriterRole})
+}
+
 // AddPipelineWriterToRepoInTransaction gives a pipeline access to write to it's own output repo.
 // This is distinct from ModifyRoleBinding because AddPipelineWriter is a less expansive permission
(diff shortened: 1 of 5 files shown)


src/server/auth/server/api_server.go
```diff
@@ -713,6 +713,19 @@ func (a *apiServer) AddPipelineReaderToRepoInTransaction(txnCtx *txncontext.Tran
...
 }
 
+// AddPipelineReaderToRepoInTransaction gives a pipeline access to write data to the specified source repo.
+// The only time a pipeline needs write permission for a source repo is in the case of Cron inputs.
+// This is distinct from ModifyRoleBinding because AddPipelineWriter is a less expansive permission
+// that is included in the repoWriter role, versus being able to modify all role bindings which is
+// part of repoOwner. This method is for internal use and is not exposed as an RPC.
+func (a *apiServer) AddPipelineWriterToSourceRepoInTransaction(txnCtx *txncontext.TransactionContext, sourceRepo, pipeline string) error {
+	// Check that the user is allowed to add a pipeline to write to the output repo.
+	if err := a.CheckRepoIsAuthorizedInTransaction(txnCtx, &pfs.Repo{Type: pfs.UserRepoType, Name: sourceRepo}, auth.Permission_REPO_ADD_PIPELINE_WRITER); err != nil {
```
acohen4, line 723: Still thinking through the semantics here. Right now it's saying that a user needs write access to the source repo for a pipeline they created to write to it. I think this makes sense, but generally we think in terms of the pipelines output repo.

msteffen: Yeah I'm not sure what the ideal model is either, but this seems at least reasonable

msteffen (approved):
Overall LGTM, though maybe Peter or Armaan or someone should check too? Also, does the pipeline's auth token need to be passed to the Cron goro (in think in monitor.go)? Is that in this PR and I missed it?

acohen4:
> Overall LGTM, though maybe Peter or Armaan or someone should check too? Also, does the pipeline's auth token need to be passed to the Cron goro (in think in monitor.go)? Is that in this PR and I missed it?

@msteffen Good q! So what happens there is that the pipeline_controller receives an event from the DB update to the pipelineInfo getting an auth token. It then recognizes the RC is stale because of the mismatching Auth token, and restarts the pipelines which restart the monitoring goros that do Autoscaling / Crons

src/server/pps/server/api_server.go
```diff
@@ -1617,7 +1618,10 @@ func (a *apiServer) fixPipelineInputRepoACLsInTransaction(txnCtx *txncontext.Tra
 			if _, ok := remove[repo]; ok {
 				delete(remove, repo)
 			} else {
-				add[repo] = struct{}{}
+				addRead[repo] = struct{}{}
+				if input.Cron != nil {
+					addWrite[repo] = struct{}{}
+				}
```
PFedak, line 1624: strictly speaking I think there's a bug here, though a user probably deserves it if they get themselves into the situation. If the previous pipeline version was using a (possibly-non-master) branch from `cron_repo` as a normal PFS input, and the new pipeline starts using `cron_repo` as a cron input, the write permission won't be added.

src/server/auth/server/testing/admin_test.go
```diff
@@ -516,6 +517,95 @@ func TestPreActivationPipelinesKeepRunningAfterActivation(t *testing.T) {
...
+
+	// alice creates a pipeline
+	pipeline1 := tu.UniqueString("cron1-")
+	require.NoError(t, aliceClient.CreatePipeline(
+		pipeline1,
+		"",
+		[]string{"/bin/bash"},
+		[]string{"cp /pfs/time/* /pfs/out/"},
+		nil,
+		client.NewCronInput("time", "@every 10s"),
```
PFedak, line 550: could probably make this shorter to speed up the test, unless there's some effect of the delay I'm missing

PFedak (approved):
lgtm, surprised we were missing this for so long

src/server/pps/server/api_server.go
acohen4: good catch. I'm not sure whether it makes sense to prevent that - WDYT?

PR #7278, opened 2022-02-01
Checking for null identity/auth/license responses during log redactions

acohen4:


src/server/auth/server/api_server.go
```diff
@@ -953,8 +953,15 @@ func (a *apiServer) GetRoleBinding(ctx context.Context, req *auth.GetRoleBinding
 
 // GetRobotToken implements the protobuf auth.GetRobotToken RPC
 func (a *apiServer) GetRobotToken(ctx context.Context, req *auth.GetRobotTokenRequest) (resp *auth.GetRobotTokenResponse, retErr error) {
+	removeSecret := func(r *auth.GetRobotTokenResponse) *auth.GetRobotTokenResponse {
+		if r == nil {
+			return nil
+		}
+		copyResp := proto.Clone(r).(*auth.GetRobotTokenResponse)
```
armaanv, line 960: why do we need to make a copy here?

armaanv (approved):

src/server/auth/server/api_server.go
acohen4: good eye, I missed there was supposed to be a line below to set the token value to "" so that it wouldn't show up in logs. Instead I just removed logging the response in this case entirely since the Token is the only field

PR #7280, opened 2022-02-02
Expose JOIN_ON and GROUP_BY env variables

albscui:
The scheme for the new env variables are:
- `PACH_DATUM_<input.name>_JOIN_ON`
- `PACH_DATUM_<input.name>_GROUP_BY`

### DIFF
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -9843,3 +9843,119 @@ func TestLoad(t *testing.T) {
 	require.NoError(t, cmdutil.Encoder("", buf).EncodeProto(resp))
 	require.Equal(t, "", resp.Error, buf.String())
 }
+
+func TestPipelineEnvJoinOn(t *testing.T) {
+	c := tu.GetPachClient(t)
+	require.NoError(t, c.DeleteAll())
+
+	// create repos
+	repo1 := tu.UniqueString("TestPipelineEnvJoinOn_repo1")
+	require.NoError(t, c.CreateRepo(repo1))
+	repo2 := tu.UniqueString("TestPipelineEnvJoinOn_repo2")
+	require.NoError(t, c.CreateRepo(repo2))
+
+	input := client.NewJoinInput(
+		client.NewPFSInput(repo1, "/(*)"),
+		client.NewPFSInput(repo2, "/(*)"),
+	)
+	input.Join[0].Pfs.Name = "repo1"
+	input.Join[1].Pfs.Name = "repo2"
+	input.Join[0].Pfs.JoinOn = "$1"
+	input.Join[1].Pfs.JoinOn = "$1"
+
+	require.NoError(t, c.PutFile(client.NewCommit(repo1, "master", ""), "a", strings.NewReader("foo")))
+	require.NoError(t, c.PutFile(client.NewCommit(repo2, "master", ""), "a", strings.NewReader("bar")))
+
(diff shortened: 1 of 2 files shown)


src/server/pachyderm_test.go
```diff
@@ -9843,3 +9843,119 @@ func TestLoad(t *testing.T) {
...
+	pipeline := tu.UniqueString("pipeline")
+	_, err = c.PpsAPIClient.CreatePipeline(
+		context.Background(),
+		&pps.CreatePipelineRequest{
+			Pipeline: client.NewPipeline(pipeline),
+			Transform: &pps.Transform{
+				Cmd: []string{"bash"},
+				Stdin: []string{
+					"mkdir -p /pfs/out/$PACH_DATUM_repo1_GROUP_BY",
+					"echo $PACH_DATUM_repo1_GROUP_BY> /pfs/out/$PACH_DATUM_repo1_GROUP_BY/PACH_DATUM_repo1_GROUP_BY",
```
PFedak, line 9937: seems like you could just use `touch` since the value is in the path. You could also make the file name shorter, would make this line a little easier to read

src/server/worker/driver/driver.go
```diff
@@ -496,6 +496,8 @@ func (d *driver) UserCodeEnv(
 	for _, input := range inputs {
 		result = append(result, fmt.Sprintf("%s=%s", input.Name, filepath.Join(d.InputDir(), input.Name, input.FileInfo.File.Path)))
 		result = append(result, fmt.Sprintf("%s_COMMIT=%s", input.Name, input.FileInfo.File.Commit.ID))
+		result = append(result, fmt.Sprintf("PACH_DATUM_%s_JOIN_ON=%s", input.Name, input.JoinOn))
+		result = append(result, fmt.Sprintf("PACH_DATUM_%s_GROUP_BY=%s", input.Name, input.GroupBy))
```
PFedak, line 500: my initial concerns with this, though I think ultimately none are necessary to address:
- we'll have these values even when they aren't used. We could skip if empty, but I suppose someone could actually rely on joining/grouping on the empty string
- there can be duplicates if there was a `Group` (multiple `common.Input` entries from the same input), but it looks like `cmd.Start()` just uses the last one
- maybe more of a concern: I don't think we do any validation on input names, and the join/group key can be basically anything. The latter is probably fine, but if the input name has `=` it would lead to the wrong variable being created. This seems unlikely, but maybe we should think about validating input names at some point.

PFedak (approved):
looks good, the things I was worried about are probably nonissues. The group example should be updated to use this, not sure if that should be in this PR or something @nadegepepin should take on

nadegepepin:
@albscui, I wouldn't mind changing the example `group` example and possibly the `join`.  This will help me document those new variables.

src/server/worker/driver/driver.go
albscui: > we'll have these values even when they aren't used. We could skip if empty, but I suppose someone could actually rely on joining/grouping on the empty string
True, then I think we should skip setting these env variables if empty. Also I'm not sure why a user would ever want to join_on/group_by empty values.
> there can be duplicates if there was a Group (multiple common.Input entries from the same input), but it looks like cmd.Start() just uses the last one
I see... but that is fine since the values for these would always be equal right?
> Definitely. We should create a ticket to track this.

PR #7281, opened 2022-02-02
Changed help text of list file - path no longer interpreted as a glob pattern in 2.0

nadegepepin:


BOsterbuhr (approved):
Perfect thank you!

PR #7282, opened 2022-02-02
Changed help text of list file - path no longer interpreted as a glob pattern in 2.0

nadegepepin:
Backport from #7281


BOsterbuhr (approved):
Perfect thanks again

PR #7284, opened 2022-02-02
Configure Chunk GC via env var

albscui:
This PR gives the user the option to either turn off the Chunk GC, or tune its cadence via environment variable. Turning off the GC is useful when the user is testing out a new version of Pachyderm, and wants to potentially rollback to a backup of a previously known good state. The reason we want to turn off the Chunk GC is so that we don't lose any real chunks in object storage if there are any deletes after the upgrade.

The env variable is called `STORAGE_CHUNK_GC_PERIOD`, which is an integer in unit seconds.

The high level user flow for turning off GC:
- spin down the cluster, take a backup of Postgres
- upgrade their Pachyderm cluster to the new release, and set `STORAGE_CHUNK_GC_PERIOD` to any integer <= 0. 
- if something breaks, and the user wish to rollback to a previous version of Pachyderm with the backup of Postgres, then
- they can simply restore Postgres using the backup, then downgrade Pachyderm

### DIFF
--- a/src/internal/serviceenv/config.go
+++ b/src/internal/serviceenv/config.go
@@ -102,18 +102,17 @@ type PachdSpecificConfiguration struct {
 
 // StorageConfiguration contains the storage configuration.
 type StorageConfiguration struct {
-	StorageMemoryThreshold         int64  `env:"STORAGE_MEMORY_THRESHOLD"`
-	StorageShardThreshold          int64  `env:"STORAGE_SHARD_THRESHOLD"`
-	StorageLevelFactor             int64  `env:"STORAGE_LEVEL_FACTOR"`
-	StorageUploadConcurrencyLimit  int    `env:"STORAGE_UPLOAD_CONCURRENCY_LIMIT,default=100"`
-	StoragePutFileConcurrencyLimit int    `env:"STORAGE_PUT_FILE_CONCURRENCY_LIMIT,default=100"`
-	StorageGCTimeout               string `env:"STORAGE_GC_TIMEOUT"`
-	StorageGCPeriod                int64  `env:"STORAGE_GC_PERIOD,default=60"`
-	StorageChunkGCPeriod           int64  `env:"STORAGE_CHUNK_GC_PERIOD,default=60"`
-	StorageCompactionMaxFanIn      int    `env:"STORAGE_COMPACTION_MAX_FANIN,default=10"`
-	StorageFileSetsMaxOpen         int    `env:"STORAGE_FILESETS_MAX_OPEN,default=50"`
-	StorageDiskCacheSize           int    `env:"STORAGE_DISK_CACHE_SIZE,default=100"`
-	StorageMemoryCacheSize         int    `env:"STORAGE_MEMORY_CACHE_SIZE,default=100"`
+	StorageMemoryThreshold         int64 `env:"STORAGE_MEMORY_THRESHOLD"`
+	StorageShardThreshold          int64 `env:"STORAGE_SHARD_THRESHOLD"`
+	StorageLevelFactor             int64 `env:"STORAGE_LEVEL_FACTOR"`
+	StorageUploadConcurrencyLimit  int   `env:"STORAGE_UPLOAD_CONCURRENCY_LIMIT,default=100"`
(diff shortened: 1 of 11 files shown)


src/server/pfs/server/master.go
```diff
@@ -44,10 +43,19 @@ func (d *driver) master(ctx context.Context) {
...
 		})
-		eg.Go(func() error {
-			gc := chunk.NewGC(d.storage.ChunkStorage())
-			return gc.RunForever(ctx)
-		})
+		if d.env.StorageConfig.StorageGCPolling == "-1" {
+			// skipping GC is useful for potentially rolling back to a previous state
+			log.Info("Skipping initialization of Chunk GC because STORAGE_GC_POLLING set to -1")
+		} else {
+			gcPeriod, err := time.ParseDuration(d.env.StorageConfig.StorageGCPolling)
```
brendoncarroll, line 50: This is a little late in the initialization to parse this.   This errors the master, but then it will restart, only to error again.

I think this should be integer seconds so we can rely on the env parsing stuff to fail if it can't be parsed into an integer.  Above we should check for < 0.  So all integer values are allowed.

src/server/pfs/server/master.go
```diff
@@ -44,10 +43,19 @@ func (d *driver) master(ctx context.Context) {
...
+			if err != nil {
+				return errors.EnsureStack(err)
+			}
+			eg.Go(func() error {
+				gc := chunk.NewGC(d.storage.ChunkStorage())
```
brendoncarroll, line 55: Can you also add `STORAGE_CHUNK_GC_POLLING` to control this loop separately?

src/server/pfs/server/master.go
```diff
@@ -44,10 +43,19 @@ func (d *driver) master(ctx context.Context) {
 		eg.Go(func() error {
 			return d.storage.GC(ctx)
 		})
-		eg.Go(func() error {
-			gc := chunk.NewGC(d.storage.ChunkStorage())
-			return gc.RunForever(ctx)
-		})
+		if d.env.StorageConfig.StorageGCPolling == "-1" {
+			// skipping GC is useful for potentially rolling back to a previous state
+			log.Info("Skipping initialization of Chunk GC because STORAGE_GC_POLLING set to -1")
```
brendoncarroll, line 48: The log line should be different for chunk vs general GC.

src/internal/storage/chunk/gc.go
```diff
@@ -10,18 +10,23 @@ import (
...
 // NewGC returns a new garbage collector operating on s
 func NewGC(s *Storage) *GarbageCollector {
-	return &GarbageCollector{s: s, log: logrus.StandardLogger()}
+	return &GarbageCollector{s: s, log: logrus.StandardLogger(), period: s.gcPeriod}
 }
 
 // RunForever calls RunOnce until the context is cancelled, logging any errors.
 func (gc *GarbageCollector) RunForever(ctx context.Context) error {
-	ticker := time.NewTicker(time.Minute)
+	if gc.period <= 0 {
```
albscui, line 25: `<=0` because ticker only accepts positive values.

src/internal/serviceenv/config.go
```diff
@@ -107,8 +107,9 @@ type StorageConfiguration struct {
 	StorageLevelFactor             int64  `env:"STORAGE_LEVEL_FACTOR"`
 	StorageUploadConcurrencyLimit  int    `env:"STORAGE_UPLOAD_CONCURRENCY_LIMIT,default=100"`
 	StoragePutFileConcurrencyLimit int    `env:"STORAGE_PUT_FILE_CONCURRENCY_LIMIT,default=100"`
-	StorageGCPolling               string `env:"STORAGE_GC_POLLING"`
```
brendoncarroll, line 110: This wasn't used anywhere, so we removed it.

src/internal/testpachd/real_env.go
```diff
@@ -131,7 +131,6 @@ func DefaultConfigOptions(config *serviceenv.Configuration) {
 	config.StorageMemoryThreshold = units.GB
 	config.StorageShardThreshold = units.GB
 	config.StorageLevelFactor = 10
-	config.StorageGCPolling = "30s"
```
brendoncarroll, line 134: not used

brendoncarroll (approved):
LGTM

PR #7285, opened 2022-02-03
Add Helm Value for GC Interval

acohen4:


etc/helm/pachyderm/values.yaml
```diff
@@ -417,6 +417,7 @@ pachd:
     # object storage uploads per Pachd instance.  It is analogous to
     # the --upload-concurrency-limit argument to pachctl deploy.
     uploadConcurrencyLimit: 100
+  pfsGCInterval: 1m
```
brycemcanally, line 420: We should rename this to `storageGCPolling`. We should be consistent with the environment variable names in the Helm chart.

brycemcanally: We need to gradually make the environment variables be more globally descriptive through namespacing with prefixes (that is what we do with the storage layer configuration by prefixing them with `STORAGE`). The configuration at higher levels should be consistent with the final configuration (the environment variable) because it is going to lead to confusion if we have to track mappings from higher level configuration to lower level configuration.

etc/helm/pachyderm/values.yaml
```diff
@@ -417,6 +417,7 @@ pachd:
     # object storage uploads per Pachd instance.  It is analogous to
     # the --upload-concurrency-limit argument to pachctl deploy.
     uploadConcurrencyLimit: 100
+  storageGCPolling: 1m
```
brycemcanally, line 420: Can we not have a default set? I would prefer to have defaults handled at the lowest level possible.

albscui (approved):
LGTM!

brycemcanally (approved):
lgtm

PR #7288, opened 2022-02-03
Try (auto)scaling up until at max.

PFedak:
Tasks do not in general get created all at once, and for pipeline in
particular we expect two waves of tasks from parallel and serial datums.


PFedak:
I missed this in #7246, it's not generally correct that the task count won't change after being observed to be non-zero. I think this is causing flakes in `TestPipelineAutoscaling`

msteffen (approved):
LGTM. @acohen4 might be interested in this too

acohen4 (approved):
LGTM!

PR #7293, opened 2022-02-07
Implement a clear cache operation for the PFS cache

brycemcanally:
This PR implements a clear cache operation for the PFS cache that operates on a tag prefix. Entries that are added to the PFS cache can now include an optional tag that can be used with the clear cache operation to control the lifetime of a cache entry. The LRU eviction policy still applies to the PFS cache, but providing a way to eagerly clear entries allows us to keep the storage under control while also allowing a high number of entries before evicting (cache entries are only kept around if they are needed). The current use case for this (and the PFS cache in general) is to clear job tasks. When a job reaches a terminal state or a pipeline is stopped / deleted, we want to clear the relevant entries from the cache. In this PR, we handle this through explicit calls to the clear cache gRPC at points where this occurs (worker master terminates job, job is stopped by a user, pipeline is deleted). This isn't the ideal long term solution because the clearing operation isn't captured in one place and we cannot guarantee completion without it being transactional (no-go because it can be large) or asynchronous with retries. We will eventually move to a model where this can be handled asynchronously and will probably be based on events for the state transitions of jobs. The solution we have in this PR is essentially a best effort to clear the cache, which will fallback on the LRU eviction policy in the worst case.


brendoncarroll (approved):
LGTM

PFedak (approved):
Should we clear the cache entirely on upgrades?

Also, it would be nice to be able to cache `UploadDatumTask` with a `pipeline` tag rather than `pipeline@job`, so that they can be reused by subsequent jobs. Seems like this requires passing in a separate `Doer` to those functions.

brycemcanally:
Clearing the cache during an upgrade makes sense, but it isn't necessary right now. Losing the caching of the upload task is unfortunate, but clearing the storage seems more valuable in the short term.

PR #7295, opened 2022-02-08
S3 Gateway: Return 404 instead of 400 on trailing slash

msteffen:
Jira issues: [CORE-336]

[CORE-336]: https://pachyderm.atlassian.net/browse/CORE-336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/pfs/s3/object.go
+++ b/src/server/pfs/s3/object.go
@@ -24,7 +24,7 @@ func (c *controller) GetObject(r *http.Request, bucketName, file, version string
 	}
 
 	if strings.HasSuffix(file, "/") {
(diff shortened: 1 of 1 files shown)


PFedak (approved):
LGTM, could add a test for this I guess

msteffen:
I know I need to add a test for this but I'm really worried that if I don't hit merge right now I'll forget until 30 seconds after the release goes out. I have a to-do to add a test in a followup.

PR #7299, opened 2022-02-09
Testing enterprise deployments

acohen4:
Adds a basic test deploying pachd w/ auth locally with MinIO backend


src/internal/minikubetestenv/deploy.go
```diff
@@ -78,6 +131,14 @@ func UpgradeRelease(t *testing.T, ctx context.Context, kubeClient *kube.Clientse
...
 
+func InstallReleaseEnterprise(t *testing.T, ctx context.Context, kubeClient *kube.Clientset, user string) *client.APIClient {
+	createSecretEnterpriseKeySecret(t, ctx, kubeClient, ns)
+	opts := union(localDeploymentWithMinioOptions(ns, localImage), withEnterprise(t, ns))
+	require.NoError(t, helm.InstallE(t, opts, helmChartLocalPath, helmRelease))
+	waitForPachd(t, ctx, kubeClient, ns, localImage)
+	return testutil.AuthenticatedPachClientPostActivate(t, testutil.NewPachClient(t), user)
+}
+
 func DeleteRelease(t *testing.T, ctx context.Context, kubeClient *kube.Clientset) {
```
brendoncarroll, line 142: Let's not expose this, and instead register cleanup in InstallReleaseEnterprise, with `t.Cleanup`, so it is cleaned up at the end of the test.

src/testing/deploy/deploy_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/auth"
+	"github.com/pachyderm/pachyderm/v2/src/client"
+	"github.com/pachyderm/pachyderm/v2/src/internal/minikubetestenv"
+	"github.com/pachyderm/pachyderm/v2/src/internal/require"
+	"github.com/pachyderm/pachyderm/v2/src/internal/testutil"
+)
+
+func TestDeployEnterprise(t *testing.T) {
+	k := testutil.GetKubeClient(t)
+	minikubetestenv.DeleteRelease(t, context.Background(), k) // cleanup cluster
```
brendoncarroll, line 18: I know we've had to be defensive with cleanup in the past, but I would prefer that these new APIs don't force that on the tests.  Can we do this DeleteRelease in `InstallReleaseEnterprise` if there is already one there?

src/internal/minikubetestenv/deploy.go
```diff
@@ -78,6 +131,14 @@ func UpgradeRelease(t *testing.T, ctx context.Context, kubeClient *kube.Clientse
 	return testutil.AuthenticatedPachClient(t, testutil.NewPachClient(t), user)
 }
 
+func InstallReleaseEnterprise(t *testing.T, ctx context.Context, kubeClient *kube.Clientset, user string) *client.APIClient {
```
brendoncarroll, line 134: `*testing.T` -> `testing.TB`

TB is the common interface between tests and benchmarks, it makes this function slightly more useful although we don't have a ton of benchmarks.  Both work with the `require` package, and both support `Cleanup`.

brendoncarroll (approved):
A few tweaks to setup and teardown, otherwise LGTM.

PR #7302, opened 2022-02-10
Add unmounting all repos handler

### LINKED ISSUE #7300: [Integrations][Notebooks] Unmount all repos at once in Go mount server
An unmount all function would be nice to have in the Go mount server for the following:
- Jupyter Lab Python backend supports unmounting all repos. Currently, we're iterating through all mounted repos and calling unmount on each of them.
- For switching cluster endpoints, we'd want to unmount all repos before we switch to the new endpoint.

smalyala:
- Adds a handler for unmounting all repos in mount server
- Unmounts all repos before switching to new cluster endpoint

Closes #7300 

https://pachyderm.atlassian.net/browse/INT-518


msteffen:
The current diff is showing 900+ files changed, but that seems like a diff problem—later `master` has been merged into the feature branch but not the base branch? I wonder if it would make sense to merge the base into master (after rebasing of needed) and then base this PR on master.

If you have a minute, could you make a PR merging mount-server-master into master? I can be the approver/merger of needed. Then we can just go back to merging into master and delete mount-server-master

smalyala:
Yep, sounds good. There was a flaky test which is why I merged in master.

msteffen (commented):
LGTM

PR #7307, opened 2022-02-11
obj: Don't hold locks while doing Gets in the cacheClient

brendoncarroll:
Recently @jrockway observed something like 10k goroutines waiting on this lock.  It prevents any concurrent object storage access. 

The extremely coarse locking was chosen to deal with the semantics of the FS backed object client at the time.  That obj.Client implementation has since been improved to have typical object storage semantics, and so we no longer need to hold locks during Puts to the FS client because Puts are atomic.

### DIFF
--- a/src/internal/obj/cache_client.go
+++ b/src/internal/obj/cache_client.go
@@ -12,7 +12,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/pacherr"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


src/internal/obj/cache_client.go
```diff
@@ -65,19 +66,51 @@ func NewCacheClient(slow, fast Client, size int) Client {
...
+// getSlow retrieves an object from c.slow and concurrently writes it to w and to c.fast.
+// c.mu is not acquired until after a succesful put to prevent holding the lock while doing IO,
+// and to prevent a slow w from holding up cache access for other callers.
+func (c *cacheClient) getSlow(ctx context.Context, p string, w io.Writer) error {
+	return miscutil.WithPipe(func(w2 io.Writer) error {
+		mw := io.MultiWriter(w, w2)
+		return c.slow.Get(ctx, p, mw)
+	}, func(r io.Reader) error {
+		if err := c.fast.Put(ctx, p, r); err != nil {
+			return errors.EnsureStack(err)
```
brycemcanally, line 108: Maybe we should just log here rather than fail the get?

brycemcanally (approved):
lgtm

PR #7308, opened 2022-02-11
Fix flaky transactions

brycemcanally:
This PR is intended to fix the recent issue with flaky transactions.

The issue was that a recent change to inspect repo & inspect branch made it such that all errors during the get operation would be converted to not found errors. This was likely hiding a serialization error that we were not respecting in a larger transaction which was leading to transaction abortion errors in later queries. This was predominately affecting the deletion of cron pipelines because they are probably the most subject to serialization errors since they have a component performing operations in parallel with the deletion that is occurring.


acohen4 (approved):
LGTM

PR #7309, opened 2022-02-12
[2.0.x] obj: Don't hold locks while doing Gets in the cacheClient (#7307)

brendoncarroll:
This applies PR #7307


brycemcanally (approved):
lgtm

PR #7310, opened 2022-02-14
Start Enterprise Heartbeat routine immediately

acohen4:
the enterprise service caches the license token it gets when it heartbeats to the license service at some cadence. Here we make sure to heartbeat to the license server as soon as the server wakes up. This helps cover an edge case where pachd is restarted with a cleared etcd instance. It would be expected to fail filling the cache until it hearbeated


PFedak (approved):
LGTM, I guess an alternate fix would be to have `GetState` perform a heartbeat instead of returning `State_NONE`, but this is good to have regardless.

PR #7311, opened 2022-02-14
Add clear cache entry in auth interceptor

brycemcanally:
This PR adds the entry for clear cache in the auth interceptor. Clear cache is best effort under the hood, so this issue went unnoticed since it doesn't cause failures.


PFedak (approved):
LGTM

PR #7315, opened 2022-02-15
Fix a couple task service issues

brycemcanally:
This PR fixes a couple of task service issues:
- When workers would update the task entries in etcd, they would remove the lease which ensures cleanup when the client crashes. A put operation in etcd does this by default, but there is an option to ignore the lease which prevents this from happening. The update operation on etcd collections now does this by default.
- Tracker entries were not cleared by the postgres cache clear operation. The unit testing was updated to test this properly.

### DIFF
--- a/src/internal/storage/fileset/postgres_cache_test.go
+++ b/src/internal/storage/fileset/postgres_cache_test.go
@@ -61,7 +61,7 @@ func TestPostgresCache(t *testing.T) {
 	checkNotExists := func(i int) {
 		_, err := cache.Get(ctx, strconv.Itoa(i))
 		require.YesError(t, err)
-		exists, err := storage.exists(ctx, ids[0])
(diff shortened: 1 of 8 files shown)


src/internal/storage/fileset/postgres_cache_test.go
```diff
@@ -61,7 +61,7 @@ func TestPostgresCache(t *testing.T) {
 	checkNotExists := func(i int) {
 		_, err := cache.Get(ctx, strconv.Itoa(i))
 		require.YesError(t, err)
-		exists, err := storage.exists(ctx, ids[0])
+		exists, err := storage.exists(ctx, ids[i])
```
PFedak, line 64: ooh, sorry I missed this

PFedak (approved):
LGTM, I spent some time going over how etcd leases work to make sure I was following.

PR #7316, opened 2022-02-16
Update rabbitmq image for pach 2.1.0-rc.1

msteffen:


dgeorg42:
@msteffen - In the Makefile, should we change `CONTAINER_VERSION := 2.1.0-rc.1` to just `CONTAINER_VERSION := 2.1`?  I'm just thinking it will look unfinished when we go GA if the image is still using "rc.1" in the tag.

Functionally, though, the change works as I was able to successfully run the example on 2.1 rc.1 using the new image.

PR #7332, opened 2022-02-18
Allow incorrect use of update pipeline.

PFedak:
With auth enabled, an update pipeline call results in a missing role
binding error as it tries to authorize a write on a repo that doesn't
exist. We detect spurious updates in the rest of the logic, and will
error in the case the repo already exists, so this auth check seems
unintended and doesn't add security.


acohen4 (approved):
LGTM - I agree that the change in logic more accurately conforms to the auth model. When "update" is used as "create" it should be interpreted as such with regards to checking auth perms

PR #7333, opened 2022-02-18
Open commits performance changes

brycemcanally:
This PR includes a set of changes to improve performance when dealing with open commit chains. 

**The user facing changes are:**
- Open commits with an unfinished base commit (most recent ancestor commit that isn't errored) are no longer readable. This only impacts output repos with parallel jobs since we already prevent the creation of a commit without a finished parent in source repos.
- Errored commits (commits associated with a killed / failed job) are no longer compacted. They are still readable, but the performance will be worse since they are not compacted. This seems like the right tradeoff in most cases and it simplifies the above change since a job can be killed before its base commit is finished.
- Repo sizes are based on the most recent commit in master that is finished (rather than open).
- ~~Commits that are not readable due to this change will have a sentinel value set for the `SizeBytesUpperBound` field (TODO)~~ (We are going to leave the size as 0 until it is in a readable state).

The problem we are hitting in a few different ways is that we are creating n file sets for each commit operation where n is the number of open ancestor commits. With list operations this can end up being n^2, which creates a lot of garbage in the system that further degrades performance. Another issue that is somewhat related is that commit finishing order is based on an updated at timestamp, which can be the same for multiple commits when a pipeline is stopped. This scenario causes the commits to be finished in a pathological order which results with the first issue and more seriously a lot of duplicate compaction work. The primary change to address these issues is a new invariant that prevents us from getting a file set for a commit with an unfinished base commit. Having this invariant ensures that there is no way to hit these problems anywhere in the system. The changes in this PR are the new invariant and the corresponding changes required to satisfy the invariant.

### DIFF
--- a/src/server/pfs/pfs.go
+++ b/src/server/pfs/pfs.go
@@ -228,7 +228,7 @@ func (e ErrCommitNotFinished) Error() string {
 }
 
 func (e ErrBaseCommitNotFinished) Error() string {
-	return fmt.Sprintf("base commit %v for commit %v not finished", e.BaseCommit, e.Commit)
+	return fmt.Sprintf("base commit %v for commit %v unfinished", e.BaseCommit, e.Commit)
 }
 
 func (e ErrAmbiguousCommit) Error() string {
@@ -263,7 +263,7 @@ var (
 	fileNotFoundRe            = regexp.MustCompile(`file .+ not found`)
 	outputCommitNotFinishedRe = regexp.MustCompile("output commit .+ not finished")
 	commitNotFinishedRe       = regexp.MustCompile("commit .+ not finished")
-	baseCommitNotFinishedRe   = regexp.MustCompile("base commit .+ not finished")
+	baseCommitNotFinishedRe   = regexp.MustCompile("base commit .+ unfinished")
 	ambiguousCommitRe         = regexp.MustCompile("commit .+ is ambiguous")
 	inconsistentCommitRe      = regexp.MustCompile("branch already has a commit in this transaction")
 	commitOnOutputBranchRe    = regexp.MustCompile("cannot start a commit on an output branch")
--- a/src/server/pfs/pfs.go
+++ b/src/server/pfs/pfs.go
@@ -84,6 +84,12 @@ type ErrCommitNotFinished struct {
 	Commit *pfs.Commit
 }
 
+// ErrBaseCommitNotFinished represents an error where a base commit has not been finished.
+type ErrBaseCommitNotFinished struct {
+	BaseCommit *pfs.Commit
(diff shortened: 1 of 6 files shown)


src/server/pfs/pfs.go
```diff
@@ -253,6 +263,7 @@ var (
 	fileNotFoundRe            = regexp.MustCompile(`file .+ not found`)
 	outputCommitNotFinishedRe = regexp.MustCompile("output commit .+ not finished")
 	commitNotFinishedRe       = regexp.MustCompile("commit .+ not finished")
```
PFedak, line 265: This other one isn't actually used, which is good as it matches both of the others. It'd be nice if there's a way to get rid of these regex entirely after that grpc status change

src/server/pfs/server/master.go
```diff
@@ -122,7 +123,7 @@ func (d *driver) finishRepoCommits(ctx context.Context, compactor *compactor, re
 			return nil
 		}
 		commit := commitInfo.Commit
-		return miscutil.LogStep(fmt.Sprintf("finishing commit %v", commit.ID), func() error {
+		return miscutil.LogStep(fmt.Sprintf("finishing commit %v", commit), func() error {
```
PFedak, line 126: thanks, I keep meaning to fix this

src/server/pfs/server/driver.go
```diff
@@ -1173,8 +1176,9 @@ func (d *driver) listCommit(
 					ci = cis[len(cis)-1-i]
 				}
 				var err error
+				// TODO: Set SizeBytesUpperBound to a sentinel value when the base is not finished.
 				ci.SizeBytesUpperBound, err = d.commitSizeUpperBound(ctx, ci.Commit)
-				if err != nil {
+				if err != nil && !pfsserver.IsBaseCommitNotFinishedErr(err) {
```
PFedak, line 1181: could use `errors.Is` as it's not crossing a gRPC boundary

PFedak (approved):
LGTM, glad this was mostly localized. I think we have other tests that cover cases with multiple failed commits so we probably don't need an explicit one

src/server/pfs/server/driver.go
brycemcanally: Yeah, I just went for consistency here. Also, something that I didn't know until recently is that `errors.Is` is a type and value check and `errors.As` is a type check. So, we should use `errors.As` when we only care about the type.

src/server/pfs/pfs.go
brycemcanally: Yeah, I'm going to go ahead and make the base commit not finished regex not matchable by the not finished regex, but we should eventually come up with a better way to handle the types than regex matching.

PR #7335, opened 2022-02-21
Remove pachd's init container

acohen4:


etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -236,8 +236,4 @@ spec:
         secret:
           secretName: {{ required "If pachd.tls.enabled, you must set pachd.tls.secretName" .Values.pachd.tls.secretName | quote }}
       {{- end }}
-      initContainers:
-      - name: init-etcd
-        image: busybox:1.28
-        command: ['sh', '-c', "until wget -O- http://etcd:2379/version > /dev/null; do echo waiting for etcd; sleep 2; done"]
```
PFedak, line 242: look like we already have a 3 minute wait in pachd to connect to etcd that most operations block on, any idea if anything else relied on this?

PFedak (approved):
I don't think I know enough to dispute this, but it makes sense as far as I can tell

etc/helm/pachyderm/templates/pachd/deployment.yaml
acohen4: where do we set the 3 minute wait?

acohen4:
> I don't think I know enough to dispute this, but it makes sense as far as I can tell

looking through it, the etcd client creation should actually block on successful connection. This also happens after we connect to postgres which usually takes longer to setup anyway

etc/helm/pachyderm/templates/pachd/deployment.yaml
PFedak: https://github.com/pachyderm/pachyderm/blob/425ef6c1f55f030fa26d923ef788dcce4f82e80b/src/internal/serviceenv/service_env.go#L246

PR #7336, opened 2022-02-21
pps/cmds: change jsonnet args to StringArrayVar from StringSliceVar

brendoncarroll:
Prior to this change the `,` character would cause the query to be split, resulting in two arguments starting with`query=`.  Only the 2nd argument took effect, making the query invalid.  This change avoids any additional parsing.


src/server/pps/cmds/cmds.go
```diff
@@ -707,7 +707,7 @@ All jobs created by a pipeline will create commits in the pipeline's output repo
...
 	updatePipeline.Flags().StringVar(&jsonnetPath, "jsonnet", "", "BETA: A Jsonnet template file (url or filepath) for one or more pipelines. \"-\" reads from stdin. Exactly one of --file and --jsonnet must be set. Jsonnet templates must contain a top-level function; strings can be passed to this function with --arg (below)")
-	updatePipeline.Flags().StringSliceVar(&jsonnetArgs, "arg", nil, "Top-level argument passed to the Jsonnet template in --jsonnet (which must be set if any --arg arugments are passed). Value must be of the form 'param=value'. For multiple args, --arg may be set more than once, or it may be passed a comma-separated list of 'param=value' pairs.")
+	updatePipeline.Flags().StringArrayVar(&jsonnetArgs, "arg", nil, "Top-level argument passed to the Jsonnet template in --jsonnet (which must be set if any --arg arugments are passed). Value must be of the form 'param=value'. For multiple args, --arg may be set more than once, or it may be passed a comma-separated list of 'param=value' pairs.")
```
albscui, line 710: Do we want to remove the `For multiple args, --arg may be set more than once, or it may be passed a comma-separated list of 'param=value' pairs` comment from here as well?

albscui (approved):
LGTM

PR #7177, opened 2021-12-13
[2.0.x Backport] Adding nodeSelectors to helm contributed in #6817 (#7082)

acohen4:
* Adding nodeSelectors to helm contributed in #6817

* Adding comma between etcd initial nodes

PR #7339, opened 2022-02-22
Add node selectors to many components.

PFedak:


etc/helm/pachyderm/values.yaml
```diff
@@ -503,6 +507,8 @@ postgresql:
     size: 10Gi
     labels:
       suite: pachyderm
+  primary:
+    nodeSelector: {}
```
PFedak, line 511: I'm not clear if this needs to be here, or should be here by convention, as this is just passing the value through to the subchart

acohen4:
could you also add values for tolerations (same way as nodeSelector). I have an example PR here: https://github.com/pachyderm/pachyderm/pull/7177/files

Also, could you run `make schema` from ./etc/helm ? It adds some type-checking to our helm templates

etc/helm/pachyderm/values.yaml
acohen4: where'd "primary" come from lol?

PFedak: ah, maybe I'm doing it wrong. It seemed like the pod-related stuff in the postegresql chart was under `primary:`, maybe I don't understand how that works

acohen4: so my bad - you're totally right here. It's how the postgresql subchart distinguishes between config for the primary write PG replica vs the read replicas that can be deployed https://github.com/pachyderm/pachyderm/blob/master/etc/helm/pachyderm/charts/postgresql/templates/statefulset-readreplicas.yaml. Could you also add a similar readReplicas section?

acohen4 (approved):
LGTM!

tybritten (approved):
same, looks good.

etc/helm/pachyderm/values.yaml
PFedak: ah, nice catch, that naming makes more sense now

PR #7344, opened 2022-02-23
Add approximate compaction progress estimates.

PFedak:
This implementation uses a new label field in the task proto to track
info not actually necessary for task processing.


PFedak:
Not sure this solution really makes sense, just kind of brainstorming

PR #7346, opened 2022-02-24
CORE-288 Add Snowflake driver

albscui:
Added the official Snowflake Go driver  to `pachsql`. Tests against Snowflake are against an existing Warehouse, but we create and teardown databases for each test. We leverage environment variables configured in CI for authenticating against Snowflake.


src/internal/sdata/sdata_test.go
```diff
@@ -93,7 +116,6 @@ func setupTable(t testing.TB, db *pachsql.DB) {
 		TimeNull     sql.NullTime
 	}
 	_, err := db.Exec(`CREATE TABLE test_data (
-		id SERIAL PRIMARY KEY NOT NULL,
```
albscui, line 96: Snowflake doesn't support `SERIAL`

src/internal/sdata/sdata_test.go
```diff
@@ -74,6 +80,21 @@ func TestMaterializeSQL(t *testing.T) {
 	}
 }
 
+// newSnowSQL creates an emphermeral database in a real Snowflake instance.
+func newSnowSQL(t testing.TB) *sqlx.DB {
+	user := os.Getenv("SNOW_USER")
```
brendoncarroll, line 85: I'd prefer to write out "snowflake" so no one has to guess about what the environment variable is for if they see it outside this context.

src/internal/sdata/sdata_test.go
```diff
@@ -74,6 +80,21 @@ func TestMaterializeSQL(t *testing.T) {
...
+	password := os.Getenv("SNOW_PASSWORD")
+	account_identifier := os.Getenv("SNOW_ACCOUNT")
+	dsn := fmt.Sprintf("%s:%s@%s", user, password, account_identifier)
+
+	db, err := sqlx.Open("snowflake", dsn)
```
brendoncarroll, line 90: We should use our own `pachsql.OpenURL` so that code path also gets exercised during this test.

src/internal/pachsql/url.go
```diff
@@ -26,7 +26,16 @@ func ParseURL(x string) (*URL, error) {
 	}
 	port, err := strconv.Atoi(u.Port())
 	if err != nil {
-		return nil, errors.EnsureStack(errors.Wrapf(err, "parsing url port: %w"))
```
brendoncarroll, line 29: If we do this, then we can't tell whether the port was not set.  I think it makes more sense to do this in the `*DSN` methods and handle the case where port is 0 separately.  That way we move some of the protocol specific logic out of here.  What do you think?

src/internal/pachsql/db.go
```diff
@@ -77,6 +82,35 @@ func mySQLDSN(u URL, password string) string {
...
 
+func snowflakeDSN(u URL, password string) string {
+	// A Snowflake account_identifier uniquely identifies a Snowflake account.
+	// account_identifer is embedded in the hostname, eg <account_identifier>.snowflakecomputing.com
+	// however, the "snowflakecomputing.com" can be left out
+	// example: jsmith@my_organization-my_account/mydb/testschema?warehouse=mywh
+	// in this case, the account_identifier is my_organization-my_account
+	var account string
+	if strings.HasSuffix(u.Host, ".snowflakecomputing.com") {
+		account = strings.Split(u.Host, ".snowflakecomputing.com")[0]
```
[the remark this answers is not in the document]
brendoncarroll: Regardless, should probably define this as a constant at the top of the function since it's kind of long and you type it out twice.

src/internal/pachsql/url.go
albscui: As discussed, the desired logic is to default to `port = 0` if the user doesn't specify a port. We then check for this and default to specific port numbers in the DB specific DSN functions.

src/internal/pachsql/db.go
```diff
@@ -26,27 +32,41 @@ type Tx = sqlx.Tx
...
+		return nil, errors.Wrapf(err, "failed to generate DSN: %v", u)
+	}
 	res, err := sqlx.Open(driver, dsn)
 	return res, errors.EnsureStack(err)
 }
 
-func postgresDSN(u URL, password string) string {
+func postgresDSN(u URL, password string) (string, error) {
+	if u.Schema != "" {
+		return "", errors.New("postgres DSN should not contain schema name")
```
brendoncarroll, line 60: 👍

src/internal/pachsql/db.go
```diff
@@ -59,10 +79,17 @@ func postgresDSN(u URL, password string) string {
...
 		dsnParts = append(dsnParts, k+"="+v)
 	}
-	return strings.Join(dsnParts, " ")
+	return strings.Join(dsnParts, " "), nil
 }
 
-func mySQLDSN(u URL, password string) string {
+func mySQLDSN(u URL, password string) (string, error) {
+	if u.Schema != "" {
+		return "", errors.New("mysql DSN should not contain schema name")
```
brendoncarroll, line 87: 👍

src/internal/pachsql/url.go
brendoncarroll: I don't think this will give the behavior we want. I would expect port to be 0 whenever there is an error here.   We need to check `if u.Port() != ""`.  If no port was parsed from the url then don't call `strconv.Atoi` to try to parse it.  If there was a port parsed then try to parse it and always error if it fails.

src/internal/sdata/sdata_test.go
```diff
@@ -122,40 +125,39 @@ func setupTable(t testing.TB, db *pachsql.DB) {
 		if i > 0 {
 			fz.Fuzz(&x)
 		}
-		_, err = db.Exec(`INSERT INTO test_data `+formatColumns(x)+` VALUES `+formatValues(x, db), makeArgs(x)...)
+		insertStatement := `INSERT INTO test_data ` + formatColumns("id", x) + ` VALUES ` + formatValues(i, x, db)
+		fmt.Println(insertStatement)
```
brendoncarroll, line 129: Don't print with `fmt`.  Call `t.Log`.

src/internal/sdata/writers.go
```diff
@@ -137,6 +137,8 @@ func (m *CSVWriter) format(x interface{}) (string, error) {
 		} else {
 			y = null
 		}
+	case *time.Time:
```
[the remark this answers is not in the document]
brendoncarroll: Okay maybe it's just GitHub.

brendoncarroll (approved):
LGTM

PR #7349, opened 2022-02-24
Mount server master

smalyala:
Code freeze for 2.1.0 is over so merge mount server branch back into `master`


.circleci/config.yml
```diff
@@ -168,10 +168,39 @@ jobs:
       - run: etc/testing/circle/start-minikube.sh
       - run: etc/testing/circle/build.sh
       - run: etc/testing/circle/deploy_test.sh
+  build-pachctl:
```
msteffen, line 171:
```suggestion
  # build pachctl and push to GCS bucket, so that it can be bundled into the
  # Jupyter-Pachyderm extension
  build-pachctl:
```

src/server/pfs/fuse/server.go
```diff
@@ -447,6 +562,90 @@ func Server(c *client.APIClient, sopts *ServerOptions) error {
...
+				}
+				*c = *(newClientPtr)
+			}
+		}
+	}
+
+	return clusterStatus, nil
+}
+
+func jsonMarshal(t interface{}) ([]byte, error) {
```
msteffen, line 641: Just noting here, we should use pachyderm's serde package (which has extra handling for protobufs and a few other tweaks) here, for consistency with the rest of pachyderm/pachyderm

msteffen (approved):
One suggested change, otherwise LGTM

src/server/pfs/fuse/server.go
smalyala: When returning the auth URL login, the default JSON encoding was replacing characters that invalidated the URL. Hence I created this function that disabled escaping certain HTML characters. 

I took a look at the `EncodeTransform` function in the serde package. Super useful that it allows transformations on the data to be returned. Correct me if I'm wrong but that wouldn't help in this case right since my issue isn't with the data itself but rather a setting of the encoder, which `EncodeTransform` doesn't allow me to modify.

PR #7350, opened 2022-02-24
[2.1.x Backport] Finer grained locking of Index Writer data (#7290)

acohen4:
* Finer grained locking of Index Writer data

* Setting level.firstIdx, lastIdx while writing indexes instead of during the chunk creation callback

* Add levelWriter.lastCbIdx and levelWriter.lastIdk

* Account for the entire Index data including the datum when considering whether to skip the first index in a chunk during the index writer callback


brycemcanally (approved):
lgtm

PR #7354, opened 2022-02-24
Add nodeSelector and annotations to some helm templates.  (#7339) (#7345)

PFedak:


etc/helm/pachyderm/templates/console/service.yaml
```diff
@@ -6,6 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 apiVersion: v1
 kind: Service
 metadata:
+  annotations: {{- toYaml .Values.console.service.annotations | nindent 4 }}
```
chainlink, line 9: We probably should surround this in an if statement

chainlink (approved):
Looks good with one small edit

etc/helm/pachyderm/templates/console/service.yaml
PFedak: ah, fair, I had seen some places where we let the bare `{}` through but that seems fine

chainlink: It might work, generally safer to go with the if though

PR #7356, opened 2022-02-25
pachctl: only print "reading from stdin" if stdin is a terminal

jrockway:
Users have reported being annoyed by "Reading from stdin." messages.  This suppresses them when stdin is not a terminal.  We also print the message to stderr now.

### DIFF
--- a/src/internal/cmdutil/util.go
+++ b/src/internal/cmdutil/util.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/mattn/go-isatty"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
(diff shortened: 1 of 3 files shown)


PFedak (approved):
LGTM, looks like we print a similar message in `readPipelineBytes` in `server/pps/cmds` where we could use this, though I imagine fewer people see that one

jrockway:
Oh, I got the readPipelineBytes one, I think.  Is there another one? The existence of two places where we print this is what talked me into adding the helper function.

PFedak:
Oh, whoops, no, not sure how I missed that.

jrockway:
No worries.  That's why there's a comments section!

PR #7357, opened 2022-02-25
Pool pachyderm clusters for parallel integration tests

acohen4:
This change introduces a cluster factory utility to the `minikubetestenv` package that can be used to programmatically create pachyderm clusters and associated clients across namespaces on a kubernetes cluster.

Open questions:
- One thing that could be annoying is that tests that have new clusters spinned up will appear to take longer than the actual execution does. This could screw up stats overtime if we care about that

### DIFF
--- a/etc/helm/pachyderm/templates/pachd/config-job.yaml
+++ b/etc/helm/pachyderm/templates/pachd/config-job.yaml
@@ -38,11 +38,7 @@ spec:
         {{- end}}
         {{- if .Values.enterpriseServer.enabled }}
         - name: PACH_ADDR
-<<<<<<< HEAD
           value: "grpc://pach-enterprise:"{{ .Values.enterpriseServer.service.apiGRPCPort }}    
-=======
-          value: "grpc://pach-enterprise:31650"        
->>>>>>> 504264a37f (Making ports flexible and such)
         {{- end }}
         {{- if .Values.pachd.rootTokenSecretName }}    
         - name: ROOT_TOKEN
--- a/etc/helm/pachyderm/templates/pachd/config-job.yaml
+++ b/etc/helm/pachyderm/templates/pachd/config-job.yaml
@@ -38,7 +38,11 @@ spec:
         {{- end}}
         {{- if .Values.enterpriseServer.enabled }}
         - name: PACH_ADDR
+<<<<<<< HEAD
           value: "grpc://pach-enterprise:"{{ .Values.enterpriseServer.service.apiGRPCPort }}    
+=======
+          value: "grpc://pach-enterprise:31650"        
+>>>>>>> 504264a37f (Making ports flexible and such)
         {{- end }}
         {{- if .Values.pachd.rootTokenSecretName }}    
         - name: ROOT_TOKEN
--- a/etc/testing/circle/start-minikube.sh
+++ b/etc/testing/circle/start-minikube.sh
@@ -9,8 +9,8 @@ VERSION=v1.19.0
 minikube_args=(
   "--vm-driver=docker"
   "--kubernetes-version=${VERSION}"
-  "--cpus=8"
-  "--memory=2Gi"
+  "--cpus=7"
+  "--memory=12Gi"
 )
 while getopts ":v" opt; do
   case "${opt}" in
--- a/etc/testing/circle/start-minikube.sh
+++ b/etc/testing/circle/start-minikube.sh
@@ -9,7 +9,7 @@ VERSION=v1.19.0
 minikube_args=(
   "--vm-driver=docker"
   "--kubernetes-version=${VERSION}"
-  "--cpus=4"
+  "--cpus=8"
   "--memory=2Gi"
 )
 while getopts ":v" opt; do
--- a/etc/testing/circle/start-minikube.sh
+++ b/etc/testing/circle/start-minikube.sh
@@ -9,6 +9,8 @@ VERSION=v1.19.0
 minikube_args=(
   "--vm-driver=docker"
   "--kubernetes-version=${VERSION}"
+  "--cpus=4"
+  "--memory=2Gi"
 )
 while getopts ":v" opt; do
   case "${opt}" in
--- a/src/internal/minikubetestenv/client_factory.go
+++ b/src/internal/minikubetestenv/client_factory.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 
 	"github.com/pachyderm/pachyderm/v2/src/client"
-	"github.com/pachyderm/pachyderm/v2/src/internal/require"
 	"github.com/pachyderm/pachyderm/v2/src/internal/testutil"
 	"golang.org/x/sync/semaphore"
 )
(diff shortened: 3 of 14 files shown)


etc/helm/pachyderm/templates/pachd/config-job.yaml
```diff
@@ -37,13 +37,8 @@ spec:
...
         {{- end}}
         {{- if .Values.enterpriseServer.enabled }}
-        {{- if eq .Values.enterpriseServer.service.type "NodePort" }}
         - name: PACH_ADDR
-          value: "grpc://pach-enterprise:1650"
-        {{- else }}
-        - name: PACH_ADDR
-          value: "grpc://pach-enterprise:30650"        
-        {{- end }}
+          value: "grpc://pach-enterprise:31650"
```
acohen4, line 41: make this port configurable

acohen4: also do we need to warn customers about ports changing? this change probably shouldn't go out till 2.2

src/internal/minikubetestenv/deploy.go
```diff
@@ -22,36 +25,69 @@ import (
...
+	}
+	return address
 }
 
 func localDeploymentWithMinioOptions(namespace, image string) *helm.Options {
+	os := runtime.GOOS
+	serviceType := ""
+	switch os {
+	case "darwin":
+		serviceType = "LoadBalancer"
```
acohen4, line 78: while working on this I learned that the preferred way to access kube services over localhost on mac is by declaring them as type: `LoadBalancer`, and running `minikube tunnel` in a separate shell session. This works much better than port-forward

src/internal/minikubetestenv/client_factory.go
```diff
@@ -0,0 +1,102 @@
...
+	poolSize        = 3
+	namespacePrefix = "test-cluster-"
+)
+
+var (
+	clusterFactory *ClusterFactory
+	setup          sync.Once
+)
+
+type ClusterFactory struct {
```
brendoncarroll, line 26: 🚀 This is incredible.  I thought we were legit years away from this.

src/internal/minikubetestenv/deploy.go
chainlink: Death to port-forward! 🔪

src/internal/minikubetestenv/deploy.go
```diff
@@ -22,36 +25,69 @@ import (
...
-func getPachAddress(t testing.TB) *grpcutil.PachdAddress {
-	cfg, err := config.Read(false, false)
+type DeployOpts struct {
+	Version      string
+	Enterprise   bool
+	AuthUser     string
+	CleanupAfter bool
+	// Because NodePorts are cluster-wide, we use a PortOffset to
+	// assign separate ports per deployment.
+	// NOTE: it might make more sense to declare port instead of offset
```
chainlink, line 40: We might want to have a data structure which maps namespace name to ports. I'm encountering the same issue on the pulumi side setting up multiple pachs in a given cluster.

chainlink:
Agree with collapsing nodeport (I even have a PR here :D https://github.com/pachyderm/pachyderm/pull/6950/files) I think it probably should be in a separate PR though as it's a breaking change.

A few general thoughts:
- Ensuring pachd is fully up and ready to accept data before pointing the tests at them?
- Why not just run a suite/namespace per test run instead of pooling them?
- Making sure we're tearing down the resources (PVCs, etc) after each test run / at the end of the runs
- Getting away from port forwarding is definitely _long_ overdue, so nice work there!

src/internal/minikubetestenv/client_factory.go
```diff
@@ -0,0 +1,102 @@
...
+	managedClusters   map[string]*client.APIClient
+	availableClusters map[string]bool
```
brendoncarroll, line 28: I think this could be a` map[string]struct{}`.  I don't see this assigned to false anywhere only deleted from.

src/internal/minikubetestenv/client_factory.go
```diff
@@ -0,0 +1,102 @@
...
+	cf.managedClusters[assigned] = nil // hold my place in line
+	managedCount := len(cf.managedClusters)
+	cf.mu.Unlock()
+
+	c := InstallRelease(t,
+		context.Background(),
+		assigned,
+		kube,
+		&DeployOpts{
+			PortOffset: uint16(managedCount * 11),
```
brendoncarroll, line 70: So `mangagedClusters` only ever grows? Otherwise I think there would be problems here.  Maybe comment that invariant somewhere.

src/internal/minikubetestenv/client_factory.go
```diff
@@ -0,0 +1,102 @@
...
+func AcquireCluster(t testing.TB, ctx context.Context) (*client.APIClient, string) {
+	setup.Do(func() {
+		clusterFactory = &ClusterFactory{
+			managedClusters:   map[string]*client.APIClient{},
+			availableClusters: map[string]bool{},
+			sem:               *semaphore.NewWeighted(poolSize),
+		}
+	})
+
+	clusterFactory.sem.Acquire(ctx, 1)
```
brendoncarroll, line 89: let's `require.NoError` the result of this.  If the context expires then `Acquire` should return an error.

src/internal/minikubetestenv/client_factory.go
```diff
@@ -0,0 +1,102 @@
...
+		})
+	cf.mu.Lock()
+	cf.managedClusters[assigned] = c
+	cf.mu.Unlock()
+	return assigned, c
+}
+
+// AcquireCluster returns a pachyderm APIClient from one of a pool of managed pachyderm
+// clusters deployed in separate namespace, along with the associated namespace
+func AcquireCluster(t testing.TB, ctx context.Context) (*client.APIClient, string) {
```
brendoncarroll, line 80: Not sure how much value the context adds here.  We assume context.Backrgound() elsewhere.  Here, we use it to set the deadline for sem.Acquire.  I think this will be run at the beginning of every test, and in practice no one will want fine grained control over how long to wait for a cluster.  If it takes longer than 5 minutes, maybe something is wrong.  It's not something developers will be thinking about when they are writing tests.  I think we should make that decision (about how long to wait) here.

src/internal/minikubetestenv/client_factory.go
```diff
@@ -0,0 +1,102 @@
...
+	assigned, c := clusterFactory.acquireFreeCluster()
+	if assigned == "" {
+		assigned, c = clusterFactory.acquireNewCluster(t)
+	}
+	t.Cleanup(func() {
+		clusterFactory.sem.Release(1)
+		clusterFactory.mu.Lock()
+		clusterFactory.availableClusters[assigned] = true
+		clusterFactory.mu.Unlock()
+	})
```
brendoncarroll, line 100: What do you think about doing a delete all before we return it to the pool?

brendoncarroll (approved):
These changes are great.  `AcquireCluster` fixes everything wrong with GetPachClient.  And `InstallRelease` taking any k8s client is pretty cool too.

I left a couple small comments on the ClientFactory section, only one is really a bug/possibly unintended behavior. 

It looks like `AcquireCluster` is going to be the most called function in this package, I think we should make it more foolproof by calling DeleteAll either during cleanup, after acquisition, or both.

We had discussed cleanup in InstallRelease before, but with the client factory, it  looks like that is less necessary than cleaning up in AcquireCluster, sorry if I added noise/confusion there.

src/internal/minikubetestenv/client_factory.go
acohen4: I'm all for it! I think it's simpler to migrate over the tests, I think maybe it's best to add a `TODO` here for a follow up PR

src/internal/minikubetestenv/client_factory.go
acohen4: sounds good to me - i agree 👍 I didn't put much thought behind use-cases here and thought it'd be nice to have.

src/internal/minikubetestenv/client_factory.go
acohen4: good call out - yes the list is ever growing

src/internal/minikubetestenv/client_factory.go
```diff
@@ -0,0 +1,102 @@
...
+
+	clusterFactory.sem.Acquire(ctx, 1)
+	var assigned string
+	assigned, c := clusterFactory.acquireFreeCluster()
+	if assigned == "" {
+		assigned, c = clusterFactory.acquireNewCluster(t)
+	}
+	t.Cleanup(func() {
+		clusterFactory.sem.Release(1)
```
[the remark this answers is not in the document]
acohen4: You're right. I think the right way to do this is to have the semaphore release last. Basically all access to `mu` are nested within the access to `sem`

acohen4:
> A few general thoughts:
> 
> * Ensuring pachd is fully up and ready to accept data before pointing the tests at them?

This is handled with `deploy.go#waitForPachd()`

> * Why not just run a suite/namespace per test run instead of pooling them?

Good q. Right now mostly for flexibility. It's not clear that more is merrier right now. For example, after a certain number of clusters, k8s scheduling fails with CPU limits. This seems to be a slightly smaller amount of clusters than available CPUs. I think we should probably increase our CI instances to xlarge, but at the moment we have 4 CPUs at our disposal. It's faster to reuse clusters than teardown and rebuild

> * Making sure we're tearing down the resources (PVCs, etc) after each test run / at the end of the runs

Related to the above point - right now I don't think this is the way to go for the client_factory. That being said, the utilities in deploy.go do cleanup at the end, and those are the utilities used in the deployment specific tests

src/internal/minikubetestenv/client_factory.go
chainlink: Currently, does it just release the used cluster back into the pool in it's current state?

etc/testing/circle/start-minikube.sh
```diff
@@ -9,6 +9,8 @@ VERSION=v1.19.0
 minikube_args=(
   "--vm-driver=docker"
   "--kubernetes-version=${VERSION}"
+  "--cpus=7"
+  "--memory=2Gi"
```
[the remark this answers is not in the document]
acohen4: what value makes sense/

src/internal/minikubetestenv/client_factory.go
acohen4: yes - right now the existing tests manage cleanup anyway. As part of completing the migration, I will try to clean the state here

PR #7366, opened 2022-03-02
Set Nodeports to the same values as the service port

acohen4:
I think this shouldn't be a breaking change unless customers use extra annotations in their setup that are outside our ingress. Since this is a significant change, I think we should prep for it being released in 2.2

### DIFF
--- a/etc/helm/pachyderm/templates/enterprise-server/service.yaml
+++ b/etc/helm/pachyderm/templates/enterprise-server/service.yaml
@@ -20,35 +20,30 @@ metadata:
 spec:
   ports:
   - name: api-grpc-port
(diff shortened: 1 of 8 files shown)


chainlink:
There's a coming change with https://pachyderm.slack.com/archives/C02C8TUF493/p1646256970496589 that we'll want to coordinate with.

chainlink (approved):
Overall I think this is a good change, but I have two considerations:
- Other services on cluster that use `peer-port` will break with this, so we'll need to give folks a heads up
- Clients _should_ be using `externalService` for ingress to pach, but if they're not this will break as well

tybritten (approved):
LGTM

PR #7367, opened 2022-03-02
Receive keep alive responses for worker ip renewal

brycemcanally:
This PR makes it such that we receive the keep alive responses when renewing a worker's ip. This will remove the warning messages that can be seen when the internal buffer in the keep alive implementation is filled.

### DIFF
--- a/src/server/cmd/worker/main.go
+++ b/src/server/cmd/worker/main.go
@@ -86,9 +86,19 @@ func do(config interface{}) error {
 	}
 
 	// keepalive forever
-	if _, err := env.GetEtcdClient().KeepAlive(context.Background(), resp.ID); err != nil {
+	keepAliveChan, err := env.GetEtcdClient().KeepAlive(context.Background(), resp.ID)
+	if err != nil {
 		return errors.Wrapf(err, "error with KeepAlive")
 	}
+	go func() {
+		for {
+			_, more := <-keepAliveChan
+			if !more {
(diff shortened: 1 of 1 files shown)


src/server/cmd/worker/main.go
```diff
@@ -86,9 +86,19 @@ func do(config interface{}) error {
...
-	if _, err := env.GetEtcdClient().KeepAlive(context.Background(), resp.ID); err != nil {
+	keepAliveChan, err := env.GetEtcdClient().KeepAlive(context.Background(), resp.ID)
+	if err != nil {
 		return errors.Wrapf(err, "error with KeepAlive")
 	}
+	go func() {
+		for {
+			_, more := <-keepAliveChan
+			if !more {
+				log.Errorf("failed to renew worker ip etcd lease")
```
PFedak, line 97: From my understanding, if there's any actual problem renewing the lease, some actual operation will fail later, right? Obviously this doesn't introduce any new bug here.

PFedak (approved):
LGTM, I guess we could put this in a util with an `onFail` handler since I can't imagine we'd ever want different logic, but it also seems unlikely that we'll add more indefinite keepAlives like this

src/server/cmd/worker/main.go
brycemcanally: I think it is mostly used for collecting info from workers, although the restart datum functionality would be affected. We should eventually move to using kubernetes for this rather than having the workers register their ip in etcd.

PR #7370, opened 2022-03-02
Structured Data: CSV & JSON Parsers

brendoncarroll:
Adds parsers for JSON lines and CSV formats.


src/internal/sdata/json.go
```diff
@@ -0,0 +1,137 @@
...
+		record[m.fields[i]] = y
+	}
+	return errors.EnsureStack(m.enc.Encode(record))
+}
+
+func (m *JSONWriter) Flush() error {
+	return errors.EnsureStack(m.bufw.Flush())
+}
+
+type jsonParser struct {
```
albscui, line 88: Why is `jsonParser` private, while `JSONWriter` public?

src/internal/sdata/csv.go
```diff
@@ -150,78 +126,29 @@ func (m *CSVWriter) Flush() error {
 	return errors.EnsureStack(m.cw.Error())
 }
 
-// JSONWriter writes tuples as newline separated json objects.
-type JSONWriter struct {
-	bufw   *bufio.Writer
-	enc    *json.Encoder
-	fields []string
-	record map[string]interface{}
+type csvParser struct {
```
albscui, line 129: Why is `csvParser` private while `CSVWriter` public?

src/internal/sdata/sdata_test.go
```diff
@@ -20,6 +20,86 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/testsnowflake"
 )
 
+// TestFormatParse is a round trip from a Tuple through formatting and parsing
+// back to a Tuple again.
+func TestFormatParse(t *testing.T) {
```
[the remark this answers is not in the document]
albscui: nit: this seems like a good place for a Go fuzz test, https://go.dev/doc/tutorial/fuzz

src/internal/sdata/json.go
```diff
@@ -0,0 +1,137 @@
...
+		record: make(map[string]interface{}, len(fieldNames)),
+	}
+}
+
+func (m *JSONWriter) WriteTuple(row Tuple) error {
+	if len(row) != len(m.fields) {
+		return ErrTupleFields{Writer: m, Fields: m.fields, Tuple: row}
+	}
+	record := m.record
+	for i := range row {
```
albscui, line 37: Shouldn't we be leveraging `convert()` here like in the CSVWriter?

albscui (changes requested):

src/internal/sdata/sdata_test.go
```diff
@@ -20,6 +20,86 @@ import (
...
+		c := ""
+		d := sql.NullInt64{}
+		return Tuple{&a, &b, &c, &d}
+	}
+	fieldNames := []string{"a", "b", "c", "d"}
+	for _, tc := range testCases {
+		t.Run(tc.Name, func(t *testing.T) {
+			const N = 10
+			buf := &bytes.Buffer{}
+			fz := fuzz.New()
```
albscui, line 62: Do we need to create new fuzzer for each test case? Can we just create a single fuzzer outside the loop?

brendoncarroll: I was doing this so it would be deterministic and produce the same values for each test.  We also need to set the rand.Source, which I added.

src/internal/sdata/convert.go
```diff
@@ -0,0 +1,434 @@
...
+	for _, layout := range []string{
+		time.RFC3339Nano,
+		time.RFC1123Z,
+		time.RFC822Z,
+		time.Kitchen,
+		time.ANSIC,
+	} {
+		t, err := time.Parse(layout, x)
+		if err == nil {
+			return t, errors.EnsureStack(err)
```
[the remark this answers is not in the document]
brendoncarroll: I guess we don't. Removed.

src/internal/sdata/convert.go
```diff
@@ -0,0 +1,434 @@
...
+		return asInt8(dst, x)
+	case *int16:
+		return asInt16(dst, x)
+	case *int32:
+		return asInt32(dst, x)
+	case *int64:
+		return asInt64(dst, x)
+	case *float64:
+		return asFloat64(dst, x)
+	case *string:
```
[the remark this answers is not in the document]
brendoncarroll: I added this, but it doesn't work for CSV because there isn't a way to encode a null string in CSV.  So I also held off of adding it to the tests.

albscui (approved):
LGTM!

src/internal/sdata/sdata_test.go
albscui: Just for my own understanding, but I'd assume that you only need to set the random seed once for all fuzzing operations to be deterministic?

brendoncarroll: If we want to get the same data for each run of the test we have to seed it with the same source every run.

PR #7373, opened 2022-03-03
Fix debug dump segfault on failed jobs

albscui:
Currently, when a user try to get a debug dump for a pipeline/job that failed, they get a cryptic `EOF` error. This is actually caused by a segmentation fault in our code, which crashes `pachd`.

### DIFF
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -9228,15 +9228,21 @@ func TestDebug(t *testing.T) {
 
 	expectedFiles, pipelines := tu.DebugFiles(t, dataRepo)
 
-	for _, p := range pipelines {
+	for i, p := range pipelines {
+		cmdStdin := []string{
+			fmt.Sprintf("cp /pfs/%s/* /pfs/out/", dataRepo),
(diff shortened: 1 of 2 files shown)


src/server/pachyderm_test.go
```diff
@@ -9228,15 +9228,20 @@ func TestDebug(t *testing.T) {
...
 	expectedFiles, pipelines := tu.DebugFiles(t, dataRepo)
 
-	for _, p := range pipelines {
+	for i, p := range pipelines {
+		cmdStdin := []string{
+			fmt.Sprintf("cp /pfs/%s/* /pfs/out/", dataRepo),
+			"sleep 3",
+		}
+		if i == 0 {
+			// fail on purpose
```
acohen4, line 9237: I think we should add more context to the comment so that looking back we know why this is tested and what we 
should expect

acohen4 (approved):
Left a nit. LGTM!

brendoncarroll (approved):

brycemcanally (approved):
lgtm

PR #7375, opened 2022-03-04
Add post upgrade hook for configpod

tybritten:
This is why configjobs are only running on install not upgrade


acohen4 (approved):
Ah! Thank you!

PR #7379, opened 2022-03-04
[DOC-332] Minor changes on local deploy

nadegepepin:


doc/docs/2.0.x/getting_started/local_installation.md
```diff
@@ -174,7 +172,7 @@ When done with the [Prerequisites](#prerequisites), deploy Pachyderm on your loc
 
 !!! Note  "When deploying locally with Console"
```
nitinjainsj, line 173: Should we change the wording for "Install Pachyderm with Console" -- I'd expect people coming to this step for the first time will not know what is "console". Instead something like "full featured" or "all features" might be more intuitive.

nitinjainsj (approved):
LGTM

doc/docs/2.0.x/getting_started/local_installation.md
nadegepepin: I actually think Console makes more sense. "Full feature" does not really give a clear idea of what we imply. And when deploying locally, the only Enterprise feature you care about is Console.

PR #7383, opened 2022-03-04
Don't propagate TotalFileSet if parent commit errored in alias commits

albscui:
Updating `2.1.x` branch first because there are changes in the `master` branch that no longer reproduces the `sql: no rows in result set` error. We can cherry-pick this change into master later.

### DIFF
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -9972,3 +9972,47 @@ func TestLoad(t *testing.T) {
 	require.NoError(t, cmdutil.Encoder("", buf).EncodeProto(resp))
 	require.Equal(t, "", resp.Error, buf.String())
 }
+
+func TestPutFileNoErrorOnErroredParentCommit(t *testing.T) {
+	// Test whether PutFile still works on a repo where a parent commit errored
+	if testing.Short() {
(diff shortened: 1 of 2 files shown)


src/server/pachyderm_test.go
```diff
@@ -9972,3 +9972,56 @@ func TestLoad(t *testing.T) {
...
+			client.NewPFSInput("A", "/*"),
+		),
+		"",
+		false,
+	))
+	require.NoError(t, c.PutFile(client.NewCommit("inA", "master", ""), "file", strings.NewReader("foo")))
+	commitInfo, err := c.InspectCommit("A", "master", "")
+	require.NoError(t, err)
+	_, err = c.WaitCommitSetAll(commitInfo.Commit.ID)
+	require.NoError(t, err)
```
PFedak, line 10025: could add a check that the output commit did actually fail to make sure we're really testing it

PFedak (approved):
LGTM, only note is that this adds a single extra hop in lookup for aliases when the parent had a validation error. We could alternatively do a best effort fetch by handling `ErrNoTotalFileSet` and ignoring it if the commit is errored, but nothing should be reading those commits, anyway, so I think this can be ignored

PR #7387, opened 2022-03-09
[ENT-102] pachgen is non-deterministic

robert-uhl:


albscui (changes requested):
There doesn't seem to be any new code generated, but rather just rearranging old code. I wonder what caused this? When I try running `make proto` locally on master, this file doesn't get changed.

robert-uhl:
Is there anything non-deterministic about `pachgen`? Maybe it behaves differently on different platforms? I ran `make proto` on x86 Linux.

robert-uhl:
Albert & I ran this down. After some testing, we were able to confirm that it was due to `find . -name "*.proto"` returning differently-ordered results on our two machines. There are two approaches we could take: one is to sort the results of `find`, and one is to sort the generated code, so no matter what the order of the input is the output is the same. After some discussion, we decided that sorting in the code generator is the better choice, because that protects against any semantically-null changes like file ordering, and it means that we can continue to use `-print0/-0` in `run.sh` without changes.

albscui (changes requested):
Can we move the sort implementation to `pachgen/main.go` ? Since we plan on adding more generators in the future.

Maybe just sort the proto files before this line https://github.com/pachyderm/pachyderm/blob/master/etc/proto/pachgen/main.go#L41

PR #7388, opened 2022-03-09
Distribute load testing

brycemcanally:
This PR distributes the load generation part of the load testing tooling. The distributed work is managed by the new task service, so the benefits that come with that will apply here (i.e. task scalability and visibility). This change will allow us to more quickly generate large scale workloads by taking advantage of more machines in a cluster. 

A few changes were made to the load testing tooling to enable this change:
- A load specification is now a protobuf type internally rather than a normal Go type, which makes it easier to interface with the task service. The YAML interface remains though, we just convert to JSON then protobuf now.
- Each put file modification is now a task and is performed by creating a file set and returning it to the load test doer which will add the file set to the commit.
- Validation was redesigned to be more scalable and distributable. The model we now have is probabilistic and consists of computing a hash for each file path + size (eventually will be content when we distribute validation) and XORing them to produce a final hash for the state of a commit.


src/internal/pfsload/worker.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/internal/backoff"
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/miscutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/task"
+	log "github.com/sirupsen/logrus"
+)
+
+const namespace = "pfsload"
+
+func Worker(pachClient *client.APIClient, taskService task.Service) error {
```
brendoncarroll, line 20: I think you could get by with just the pfs.APIClient interface instead of the whole client.APIClient.

src/internal/pfsload/validator.go
```diff
@@ -5,201 +5,150 @@ import (
...
+	return &Validator{
 		spec:   spec,
-		buffer: fileset.NewBuffer(),
 		random: random,
-	}
-	return &validatorClient{
-		Client:    client,
-		validator: v,
-	}, v, nil
+		hash:   pfs.NewHash().Sum(nil),
```
brendoncarroll, line 33: nit: these could be `pachhash.Output`s

brendoncarroll (approved):
A couple suggestions, otherwise LGTM.

src/internal/pfsload/worker.go
brycemcanally: I need the full client for the auth functionality, but with some improvements to auth (primarily refactoring the functionality into utilities) we should be able to do that eventually.

PR #7390, opened 2022-03-09
Fix Secure (HTTPS) setting for Minio

tybritten:
currently only `"1"` enables this. Using `ParseBool` to accept `"true"` and update values file documentation.

### DIFF
--- a/etc/helm/pachyderm/values.yaml
+++ b/etc/helm/pachyderm/values.yaml
@@ -420,11 +420,17 @@ pachd:
       id: ""
       secret: ""
     minio:
+      # minio bucket name
       bucket: ""
+      # the minio endpoint. Should only be the hostname:port, no http/https.
       endpoint: ""
+      # the username/id with readwrite access to the bucket.
       id: ""
+      # the secret/password of the user with readwrite access to the bucket.
       secret: ""
+      # enable https for minio with "true" defaults to "false"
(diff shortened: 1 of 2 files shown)


etc/helm/pachyderm/values.yaml
```diff
@@ -420,11 +420,17 @@ pachd:
...
+      # minio bucket name
       bucket: ""
+      # the minio endpoint. Should only be the hostname:port, no http/https.
       endpoint: ""
+      # the username/id with readwrite access to the bucket.
       id: ""
+      # the secret/password of the user with readwrite access to the bucket.
       secret: ""
+      # enable https for minio with "true" defaults to "false"
       secure: ""
```
brendoncarroll, line 432: I'm not really an authority on the helm chart. But IMO this should be a boolean and not a string.

@chainlink @acohen4 any thoughts?

acohen4: I agree it should be a boolean. Only concern is helm backwards compatibility - we usually don't want to modify existing field types, especially not in point releases

tybritten: Yeah thats why I left it as a string, especially because `ParseBool` can handle both options cleanly

acohen4 (approved):

PR #7393, opened 2022-03-10
Start simplifying pipeline_controller.go 

acohen4:
The changes in this PR are purely for the purposes of simplifying/refactoring code

### DIFF
--- a/src/server/pps/server/pipeline_controller.go
+++ b/src/server/pps/server/pipeline_controller.go
@@ -32,14 +32,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-type rcExpectation byte
-
-const (
-	noExpectation rcExpectation = iota
-	noRCExpected
-	rcExpected
-)
-
 func max(is ...int) int {
 	if len(is) == 0 {
 		return 0
@@ -211,7 +203,7 @@ func (pc *pipelineController) step(timestamp time.Time) (isDelete bool, retErr e
 		// set pc.rc
 		// TODO(msteffen) should this fail the pipeline? (currently getRC will restart
 		// the pipeline indefinitely)
-		if err := step.getRC(ctx, noExpectation); err != nil && !errors.Is(err, errRCNotFound) {
+		if err := step.getRC(ctx); err != nil && !errors.Is(err, errRCNotFound) {
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


src/server/pps/server/pipeline_controller.go
```diff
@@ -418,11 +394,11 @@ func (step *pcStep) rcIsFresh() bool {
 		log.Errorf("PPS master: RC for %q is nil", step.pipelineInfo.Pipeline.Name)
 		return false
 	}
-	expectedName := ""
-	if step.pipelineInfo != nil {
-		expectedName = ppsutil.PipelineRcName(step.pipelineInfo.Pipeline.Name, step.pipelineInfo.Version)
+	if step.pipelineInfo == nil {
```
PFedak, line 397: weird that this check was present previously - there's no situation where we should hit this, right?

src/server/pps/server/pipeline_controller.go
```diff
@@ -271,20 +263,58 @@ func (pc *pipelineController) newStep() (*pcStep, context.Context, error) {
...
+	}
+
+	if step.rc != nil && !step.rcIsFresh() {
+		// old RC is not down yet
+		return step.restartPipeline(ctx, "stale RC") // step() will be called again after collection write
+	}
+
+	if step.pipelineInfo.State == pps.PipelineState_PIPELINE_PAUSED {
+		if !step.pipelineInfo.Stopped {
+			// StartPipeline has been called (so spec commit is updated), but new spec
```
PFedak, line 290: `StartPipeline` no longer updates the spec commit, though I guess this is using it to refer to the pipeline spec information in general, hmm

PFedak, line 290: `StartPipeline` no longer updates the spec commit

src/server/pps/server/pipeline_controller.go
```diff
@@ -271,20 +263,58 @@ func (pc *pipelineController) newStep() (*pcStep, context.Context, error) {
...
+		}
+		return nil
+	}
+
+	if step.rc != nil && !step.rcIsFresh() {
+		// old RC is not down yet
+		return step.restartPipeline(ctx, "stale RC") // step() will be called again after collection write
+	}
+
+	if step.pipelineInfo.State == pps.PipelineState_PIPELINE_PAUSED {
```
PFedak, line 288: since there's not really a repetition/priority issue, this block could be left in the switch, just adding `&& state != PAUSED` to the condition below this

PFedak, line 288: since there's not really a repetition/priority issue, this block could be left in the switch, just adding `&& state != PAUSED` to the condition below this

PFedak (approved):
LGTM, looking forward to more :)

src/server/pps/server/pipeline_controller.go
msteffen: I know this is a drive-by comment, but I saw this and got curious: [here's the commit where this check was originally added](https://github.com/pachyderm/pachyderm/blob/8bacf0e1a62c555d190c4b53f4e7305a9bb62089/src/server/pps/server/pipeline_controller.go#L588), and as you can see, this code was totally different back then.

I think what I thought could happen was that the PPS master could read the `etcdPipelineInfo` from etcd, but then fail to read the full pipelineInfo from PPS for whatever reason, and then call `op.restartPipeline()`, which called `op.rcIsFresh()`, which would then crash without this check (though, reading through the code again, it seems like the pipeline should just fail in that case, so maybe even then it was unnecessary? I'm not totally sure, though :joy:)

PR #7394, opened 2022-03-10
Migrate more tests to AcquireClient()

acohen4:


src/internal/testutil/cmds.go
```diff
@@ -116,11 +118,64 @@ which match >/dev/null || {
...
+		require.NoError(t, err)
+		// remove the empty file so that a config can be generated
+		require.NoError(t, os.Remove(config))
+		t.Cleanup(func() {
+			// since this call gets run multiple times, ignore error
+			_ = os.Remove(config)
+		})
+		return BashCmd(`
+		export PACH_CONFIG={{.config}}
+		pachctl config set context  --overwrite testing <<EOF
```
albscui, line 155: Just wondering: this implies that we can't use `PachctlBashCmd` to test `pachctl config` since it depends on it. Do you think we can move this out somewhere?

src/internal/testutil/auth.go
```diff
@@ -47,6 +47,12 @@ func ActivateAuth(tb testing.TB) {
 	activateAuthHelper(tb, client)
 }
 
+// ActivateAuthClient activates the auth service in the test cluster, if it isn't already enabled
+func ActivateAuthClient(tb testing.TB, c *client.APIClient) {
```
albscui, line 51: Can you clarify why we need to maintain the old `ActivateAuth`?

albscui (commented):
Figuring out which tests to migrate vs not to seems to be quite an undertaking. Can you clarify your approach on picking which tests to migrate?

This actually looks good to me, but I'll wait for the clarifications before approving.

src/internal/minikubetestenv/client_factory.go
```diff
@@ -114,9 +139,22 @@ func AcquireCluster(t testing.TB) (*client.APIClient, string) {
...
+	as := &acquireSettings{}
+	for _, o := range opts {
+		o(as)
+	}
 	assigned, c := clusterFactory.acquireFreeCluster()
 	if assigned == "" {
-		assigned, c = clusterFactory.acquireNewCluster(t)
+		assigned, c = clusterFactory.acquireNewCluster(t, as)
+	}
+	// in the case loki is requested, upgrade the cluster to include it
```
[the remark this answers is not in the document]
acohen4: I was contemplating this, but the reason this needs to be here is that the cluster may have been deployed by a test not requiring loki. We're then faced with the challenge of satisfying a future test scheduled to the cluster that does require it. By upgrading here, we make sure Loki is always available for a test requiring it. Basically `ClusterFactory.acquireNewCluster` doesn't always get called. 
`ClusterFactory.acquireNewCluster` will actually deploy with Loki if called by a test needing it.

src/internal/testutil/cmds.go
acohen4: that's true, I actually decided to go with this approach after looking at the `pachctl config` tests that do something similar. `pachctl config` actually doesn't need an API client to test, so I think it's alright to leave it as is. Those tests are located here: https://github.com/pachyderm/pachyderm/blob/master/src/server/config/cmds_test.go#L13

src/internal/testutil/auth.go
acohen4: `ActivateAuth` can go away once the migration towards this model is complete. There are still several cmd tests using the current approach that would need to be migrated first

acohen4:
> Figuring out which tests to migrate vs not to seems to be quite an undertaking. Can you clarify your approach on picking which tests to migrate?

The plan is to migrate the rest of the tests using the cluster in a future PR. There are a bunch of cmd tests that need some more infrastructural logic to be added, as well as `TestDebug` which is tricky

PR #7414, opened 2022-03-17
Propagating Custom certs referenced in pachd's SSL_CERT_DIR to side-cars

acohen4:


etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -150,6 +150,10 @@ spec:
         - name: SSL_CERT_DIR
           value:  /pachd-tls-cert
         {{- end }}
+        {{- if .Values.pachd.tls.enabled }}
```
tybritten, line 153: shouldn't this be if global.customcacerts?

tybritten (commented):
just one question. wasn't sure if you want to set that value unless customcacert is set

etc/helm/pachyderm/templates/pachd/deployment.yaml
acohen4: I think either this or `{{- if and .Values.pachd.tls.enabled .Values.global.customCaCerts }}` as we have elsewhere should work. Right now the only use-case to pass the secret name to pachd is for the `customCaCerts` case, but this value is more general. I can change it to be consistent

tybritten: Lgtm. Testing locally now. I'll package up my setup so we can make a permanent test

tybritten (approved):
Tested this locally and works as intended.

src/internal/serviceenv/config.go
```diff
@@ -98,6 +98,7 @@ type PachdSpecificConfiguration struct {
 	// TODO: Merge this with the worker specific pod name (PPS_POD_NAME) into a global configuration pod name.
 	PachdPodName                 string `env:"PACHD_POD_NAME,required"`
 	EnableWorkerSecurityContexts bool   `env:"ENABLE_WORKER_SECURITY_CONTEXTS,default=true"`
+	TLSCertsSecretName           string `env:"TLS_CERT_SECRET_NAME,default="`
```
albscui, line 101:
```suggestion
	TLSCertSecretName           string `env:"TLS_CERT_SECRET_NAME,default="`
```

albscui (approved):
LGTM! Just a nit on the `TLSCertsSecretName`.

src/server/pps/server/worker_rc.go
```diff
@@ -88,6 +88,22 @@ func getPachctlSecretVolumeAndMount(secret string) (v1.Volume, v1.VolumeMount) {
 		}
 }
 
+// getTLSCertSecretVolumeAndMount returns a Volume and VolumeMount object
+// configured for the pach-tls secret to be stored in pipeline side-cars.
+func getTLSCertSecretVolumeAndMount(secret, mountPath string) (v1.Volume, v1.VolumeMount) {
+	return v1.Volume{
+			Name: secret,
```
albscui, line 95: Is name of the volume something we control? Maybe call it something like `<secret>-volume` to distinguish it from the actual secret name?

acohen4: it is something that we control. We could do that, but to be consistent with other volumes, mounts we're creating here, I think just using the secret name makes sense

PR #7415, opened 2022-03-17
storage/fileset/index: Fix issue where `levelWriter` callback points to wrong level.

brendoncarroll:
Previously we checked if a level needed to be created using the read lock with `numLevels`, then if necessary, we created a level using `createLevel(lw *levelWriter)`.

The callback for a levelWriter is configured using the index which that levelWriter will be assigned to.  It can't go anywhere on the level stack.  It has to go at it's index or the callback will be messed up.

In `createLevel` we append without checking the length, potentially inserting a levelWriter with a callback for the wrong level at the end of the stack of levels.

These changes check whether a level needs to be added while holding the lock, and then add it, so that no other levelWriters will be created in between.  It also does a check with the read lock first, so that in the common case we only need the read lock to get a level writer.

### DIFF
--- a/src/internal/storage/fileset/index/writer.go
+++ b/src/internal/storage/fileset/index/writer.go
@@ -75,14 +75,25 @@ func (w *Writer) writeIndex(idx *Index, level int) error {
 	return errors.EnsureStack(err)
 }
 
+// setupLevel ensures that the level exists, creating one if necessary.
+// setupLevels takes levelsMu in read and write mode as needed, and releases it before returning
 func (w *Writer) setupLevel(idx *Index, level int) {
-	if level == w.numLevels() {
+	// check first with the read lock
+	if level < w.numLevels() {
+		// this should be the common case
+		return
+	}
+	// then get the write lock, check again and maybe create another level.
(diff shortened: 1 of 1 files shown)


src/internal/storage/fileset/index/writer.go
```diff
@@ -75,14 +75,25 @@ func (w *Writer) writeIndex(idx *Index, level int) error {
...
+	// check first with the read lock
+	if level < w.numLevels() {
+		// this should be the common case
+		return
+	}
+	// then get the write lock, check again and maybe create another level.
+	w.levelsMu.Lock()
+	defer w.levelsMu.RLock()
+	if level < len(w.levels) {
 		cw := w.chunks.NewWriter(w.ctx, w.tmpID, w.callback(level), chunk.WithRollingHashConfig(averageBits, int64(level)))
```
brendoncarroll, line 90: This is where the level is set for the new potential levelWriter.

brendoncarroll:
Closing, after talking with @acohen4.  This would require two callbacks to be triggered for the same level, which is not supposed to happen.

PR #7420, opened 2022-03-21
Initial script for pushing redhat marketplace images

msteffen:
Tested in CircleCI by https://github.com/pachyderm/pachyderm/pull/7421

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -261,6 +261,19 @@ jobs:
       - checkout
       - run: sudo npm install -g prettier
       - run: prettier -c etc/helm/pachyderm/values.yaml .circleci/config.yml
+  push_redhat:
+    docker:
+      - image: cimg/go:1.17.3
+    steps:
+      - checkout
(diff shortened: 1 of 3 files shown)


molinamelendezj (approved):
LGTM. 

I may split out the publish step from `etc/redhat/push_images.sh` later on into its on stage.

@msteffen just be on same page this will also fire on custom releases , is that okay?

msteffen:
Splitting `push_images.sh` into its own stage seems totally fine to me—as long as it runs and pushes an image to Red Hat every time we cut a release, they should be happy.

> @msteffen just be on same page this will also fire on custom releases , is that okay?

That's totally fine (it's also fine if it doesn't, but the extra images don't hurt anything. I was just trying to keep things simple, and to ensure it ran at least on every normal release)

molinamelendezj:
PPS2 is flaky test, it was very stubborn but it passed now. Will have to look into it some more...merging

PR #7427, opened 2022-03-22
add the --wait all for minikube start

molinamelendezj:
There are constant failures on circle ci workflow due to `Exiting due to PROVIDER_DOCKER_NOT_RUNNING: deadline exceeded running "docker version --format -": signal: killed`

this is because we aren't waiting for the kubernetes daemon to fully be up and operational before running minikube start, so adding this check prior to minikube start. and added the wait all flag to minikube to make sure it is also ready to avoid this sporadic failures.

### DIFF
--- a/etc/testing/circle/start-minikube.sh
+++ b/etc/testing/circle/start-minikube.sh
@@ -11,6 +11,7 @@ minikube_args=(
   "--kubernetes-version=${VERSION}"
   "--cpus=7"
   "--memory=12Gi"
+  "--wait=all"
 )
 while getopts ":v" opt; do
   case "${opt}" in
--- a/etc/testing/circle/start-minikube.sh
+++ b/etc/testing/circle/start-minikube.sh
@@ -25,6 +25,17 @@ while getopts ":v" opt; do
   esac
 done
 
+# wait for docker or timeout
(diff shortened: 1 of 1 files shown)


etc/testing/circle/start-minikube.sh
```diff
@@ -11,6 +11,7 @@ minikube_args=(
   "--kubernetes-version=${VERSION}"
   "--cpus=7"
   "--memory=12Gi"
+  "--wait=all"
```
acohen4, line 14: why is `wait` necessary here if we're already waiting for docker below?

molinamelendezj: just an extra layer of protection. this tells minikube start to [wait for Kubernetes components to verify and wait for after starting a cluster. ](https://minikube.sigs.k8s.io/docs/commands/start/)

molinamelendezj: just an extra layer of protection. this tells minikube start to [wait for Kubernetes components to verify and wait for after starting a cluster. ](https://minikube.sigs.k8s.io/docs/commands/start/)

acohen4 (approved):
LGTM!!

PR #7428, opened 2022-03-22
Fix missing files (flaky correctness issue) & switch keys to mount name

smalyala:
- Fix correctness issue https://pachyderm.atlassian.net/browse/INT-549
- Switches indexing into the state machine from `MountKey` to the mount name
- Adds a server handler for listing mounts (requested for FE ease of use)

An example response for `GET /mounts` looks like:
```
{
  "mounted": {
    "foomaster2": {
      "name": "foomaster2",
      "mount_key": {
        "repo": "foo",
        "branch": "master",
        "commit": ""
      },
      "mode": "ro",
      "state": "mounted",
      "status": "",
      "mountpoint": ""
    }
  },
  "unmounted": [
    {
      "name": "",
      "mount_key": {
        "repo": "bar",
        "branch": "master",
        "commit": ""
      },
      "mode": "",
      "state": "unmounted",
      "status": "",
      "mountpoint": ""
    },
    {
      "name": "",
      "mount_key": {
        "repo": "foo",
        "branch": "dev",
        "commit": ""
      },
      "mode": "",
      "state": "unmounted",
      "status": "",
      "mountpoint": ""
    }
  ]
}
```

Jira: [INT-560] [INT-549]

[INT-560]: https://pachyderm.atlassian.net/browse/INT-560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lukemarsden:
If this is a non-backward compatible API change then we need to put the new endpoints under a v2 URL

lukemarsden:
oh sorry, looks like `/mounts` is a new endpoint (so not backward incompatible), and you've also done the work to keep the existing endpoint working the same, is that right?

smalyala:
Talking with Peter more on this, I think it makes sense to hold off on this `/mounts` endpoint until we have a use case for it/decide to implement multiple mounts in the extension. That way we can design the response based on what's needed. Peter's okay with the FE continuing the use of `/repos` result for populating the UI.

I'll eject the `/mounts` commits into a separate branch so this one is just the mount key indexing switch. However, this will still require a change in the response from `/repos` since indexing by mounts allows for the same repo/branch to be mounted more than once. Hence, we'll probably use a similar repos endpoint but with `v2` attached to it somewhere. I'll add that here.

smalyala:
Moved the commits related to adding an endpoint for listing mounts to branch https://github.com/pachyderm/pachyderm/tree/mount-server-mounts-handler. Still need to add `v2` to this one.

smalyala:
This PR (https://github.com/pachyderm/jupyterlab-pachyderm/pull/104) bumps API to `v2` on the Python shim.

lukemarsden (approved):

PR #7433, opened 2022-03-24
Exit finishing early for errored commits.

PFedak:
Also add a previous job in the StopJob test to exercise an edge case
where we finish commits out of order.


brycemcanally (approved):
lgtm

PR #7436, opened 2022-03-25
Batch copy operations in unordered writer.

PFedak:
Previously, copy operations during a modify file stream forced a new
underlying fileset to be created, which made operations with many
separate copies (e.g. shuffle pipelines) unnecessarily slow. The
unordered writer only really needs to serialize its pending writes when
it hits the memory limit, and copies have the advantage that they can be
recorded without eating into that budget.

This change records copy operations in the unordered writer, deferring
the actual traversal of the copy source and writes to serialization
time. To avoid issues with path ordering, serialization must occur if a
pending copy might overlap existing puts/deletes or vice versa, so the
unordered writer must check this before making any changes.

A future implementation could interleave files from copies and
puts/deletes, eliminating the need for this new check.

### DIFF
--- a/src/internal/storage/fileset/buffer.go
+++ b/src/internal/storage/fileset/buffer.go
@@ -13,10 +13,16 @@ type Buffer struct {
 }
 
 type file struct {
-	path  string
-	datum string
-	buf   *bytes.Buffer
-	copy  File
+	path     string
+	datum    string
+	contents []fileContent
+}
+
+// contents are either raw bytes to be appended or an existing file to be copies
+// Exactly one will be non-nil
+type fileContent struct {
+	buf  *bytes.Buffer
+	copy File
 }
 
 func NewBuffer() *Buffer {
@@ -36,7 +42,6 @@ func (b *Buffer) addInternal(path, datum string) *file {
 		datumFiles[datum] = &file{
 			path:  path,
 			datum: datum,
-			buf:   &bytes.Buffer{},
 		}
 	}
 	return datumFiles[datum]
@@ -44,10 +49,12 @@ func (b *Buffer) addInternal(path, datum string) *file {
 
 func (b *Buffer) Add(path, datum string) io.Writer {
 	f := b.addInternal(path, datum)
-	if f.copy != nil {
-		panic("unsafe add")
+	if len(f.contents) > 0 && f.contents[len(f.contents)-1].copy == nil {
+		return f.contents[len(f.contents)-1].buf
 	}
-	return f.buf
+	buf := &bytes.Buffer{}
+	f.contents = append(f.contents, fileContent{buf: buf})
+	return buf
 }
 
 func (b *Buffer) Delete(path, datum string) {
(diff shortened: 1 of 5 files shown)


src/internal/storage/fileset/buffer.go
```diff
@@ -8,14 +8,22 @@ import (
...
 type file struct {
-	path  string
-	datum string
-	buf   *bytes.Buffer
+	path     string
+	datum    string
+	contents []fileContent
+}
+
+// contents are either raw bytes to be appended or an existing file to be copies
```
brycemcanally, line 22: copies -> copied

src/internal/storage/fileset/buffer.go
```diff
@@ -25,52 +33,75 @@ func NewBuffer() *Buffer {
 	}
 }
 
-func (b *Buffer) Add(path, datum string) io.Writer {
+func (b *Buffer) addInternal(path, datum string) *file {
```
brycemcanally, line 36: I think just `add` is fine. It being unexported implicitly means it is internal.

src/internal/storage/fileset/buffer.go
```diff
@@ -25,52 +33,75 @@ func NewBuffer() *Buffer {
...
 	path = Clean(path, false)
 	if _, ok := b.additive[path]; !ok {
 		b.additive[path] = make(map[string]*file)
 	}
 	datumFiles := b.additive[path]
 	if _, ok := datumFiles[datum]; !ok {
+		b.fileCount++
```
brycemcanally, line 43: This is super nitpicky, but I generally prefer the primary functionality (creating the datum file) to come before side effects (updating the file count). It just reads better imo.

src/internal/storage/fileset/storage.go
```diff
@@ -94,7 +94,7 @@ func (s *Storage) ChunkStorage() *chunk.Storage {
 
 // NewUnorderedWriter creates a new unordered file set writer.
 func (s *Storage) NewUnorderedWriter(ctx context.Context, opts ...UnorderedWriterOption) (*UnorderedWriter, error) {
-	return newUnorderedWriter(ctx, s, s.memThreshold, opts...)
+	return newUnorderedWriter(ctx, s, s.memThreshold, s.shardSizeThreshold/2, opts...)
```
brycemcanally, line 97: This needs to be the count, not the size.

src/internal/storage/fileset/unordered_writer.go
```diff
@@ -74,6 +76,11 @@ func (uw *UnorderedWriter) Put(p, datum string, appendFile bool, r io.Reader) (r
 			w = uw.buffer.Add(p, datum)
 		}
 	}
+
```
brycemcanally, line 79: nit, arbitrary space.

src/internal/storage/fileset/unordered_writer.go
```diff
@@ -157,27 +166,27 @@ func (uw *UnorderedWriter) Delete(p, datum string) error {
...
-			if !appendFile {
-				if err := w.Delete(f.Index().Path, datum); err != nil {
-					return err
-				}
-			}
-			return w.Copy(f, datum)
-		})
-		return errors.EnsureStack(err)
-	})
+
```
brycemcanally, line 179: arbitrary space

src/internal/storage/fileset/buffer.go
PFedak: makes sense, I was also debating just running through everything to re-count every once in a while rather than the incremental bookkeeping

src/internal/storage/fileset/buffer.go
```diff
@@ -25,52 +33,75 @@ func NewBuffer() *Buffer {
...
+		for file, datumFiles := range b.additive {
 			if strings.HasPrefix(file, path) {
+				b.fileCount -= len(datumFiles)
 				delete(b.additive, file)
 			}
 		}
 		return
 	}
 	if datumFiles, ok := b.additive[path]; ok {
+		b.fileCount -= len(datumFiles)
```
brycemcanally, line 76: I don't think this is the correct update. We should decrement by 1 if the path & datum pair existed in the additive set.

PFedak: ah of course, you're right

brycemcanally (approved):
lgtm

PR #7438, opened 2022-03-25
Refactor to reduce complexity in Pipeline Controller

acohen4:


src/server/pps/server/pipeline_controller.go
```diff
@@ -397,50 +384,49 @@ func (pc *pipelineController) loadLatestPipelineInfo(message *pps.PipelineInfo)
...
+			pi.Pipeline.Name, rcAuthTokenHash, hashAuthToken(pi.AuthToken))
 		return false
-	case rcPipelineVersion != strconv.FormatUint(step.pipelineInfo.Version, 10):
+	case rcPipelineVersion != strconv.FormatUint(pi.Version, 10):
 		log.Errorf("PPS master: pipeline version in %q looks stale %s != %d",
-			step.pipelineInfo.Pipeline.Name, rcPipelineVersion, step.pipelineInfo.Version)
+			pi.Pipeline.Name, rcPipelineVersion, pi.Version)
 		return false
-	case rcSpecCommit != step.pipelineInfo.SpecCommit.ID:
+	case rcSpecCommit != pi.SpecCommit.ID:
```
PFedak, line 411: just for fun, can we move this check first. we had one customer concerned about logging auth tokens, and even though it's a hash, we might as well do it less

src/server/pps/server/pipeline_controller.go
```diff
@@ -397,50 +384,49 @@ func (pc *pipelineController) loadLatestPipelineInfo(message *pps.PipelineInfo)
...
-	rcSpecCommit := step.rc.ObjectMeta.Annotations[pipelineSpecCommitAnnotation]
+	rcName := rc.ObjectMeta.Name
+	rcPachVersion := rc.ObjectMeta.Annotations[pachVersionAnnotation]
+	rcAuthTokenHash := rc.ObjectMeta.Annotations[hashedAuthTokenAnnotation]
+	rcPipelineVersion := rc.ObjectMeta.Annotations[pipelineVersionAnnotation]
+	rcSpecCommit := rc.ObjectMeta.Annotations[pipelineSpecCommitAnnotation]
 	switch {
-	case rcAuthTokenHash != hashAuthToken(step.pipelineInfo.AuthToken):
+	case rcAuthTokenHash != hashAuthToken(pi.AuthToken):
 		log.Errorf("PPS master: auth token in %q is stale %s != %s",
```
PFedak, line 404: I know it's been like this forever, but these should be warning at most, maybe even just info, as they are totally expected

src/server/pps/server/pipeline_controller.go
```diff
@@ -761,13 +728,14 @@ func (step *pcStep) getRC(ctx context.Context) (retErr error) {
 		if errCount >= maxErrCount {
 			invalidRCState := errors.Is(err, errTooManyRCs) || errors.Is(err, errStaleRC)
 			if invalidRCState {
-				return step.restartPipeline(ctx, fmt.Sprintf("could not get RC after %d attempts: %v", errCount, err))
+				return pc.restartPipeline(ctx, pi, rc, fmt.Sprintf("could not get RC after %d attempts: %v", errCount, err))
 			}
 			return err //return whatever the most recent error was
 		}
-		log.Errorf("PPS master: error retrieving RC for %q: %v; retrying in %v", step.pc.pipeline, err, d)
+		log.Errorf("PPS master: error retrieving RC for %q: %v; retrying in %v", pc.pipeline, err, d)
```
PFedak, line 735: can we make this a warning, too, while we're at it

PFedak (commented):
LGTM

src/server/pps/server/pipeline_controller.go
```diff
@@ -455,24 +441,23 @@ func (step *pcStep) setPipelineState(ctx context.Context, state pps.PipelineStat
...
+func (pc *pipelineController) startPipelineMonitor(pi *pps.PipelineInfo) {
 	// since *pc.pipelineInfo may be modified on a Bump, and startMonitor() passes its
 	// input pipelineInfo to a monitor goroutine, send a pointer to a copy of pc.pipelineInfo to
 	// avoid a race condition
-	pi := *step.pipelineInfo
-	if step.pc.monitorCancel == nil {
-		step.pc.monitorCancel = step.pc.startMonitor(step.pc.ctx, &pi)
+	if pc.monitorCancel == nil {
+		// TODO(acohen4): assess whether we should remove &(*pi)
+		pc.monitorCancel = pc.startMonitor(pc.ctx, &(*pi))
```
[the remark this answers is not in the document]
acohen4: at some point in time we were concerned about this because  `pi` was attached to the `pipelineController`, so we wanted to make sure that monitor goros were working on their own copies. 

I actually just removed this entirely. `pi` is now generated on each `step()` and the monitor goros don't even modify `pi`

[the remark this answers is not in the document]
acohen4: also good catch on the shallow copy

src/server/pps/server/pipeline_controller.go
acohen4: I'll just move auth check last?

PFedak (approved):

PR #7442, opened 2022-03-25
Decouple content defined chunking from file batching

brycemcanally:
This PR decouples the content defined chunking from file batching by splitting up the chunk writer into three simpler components. This PR also includes a split in the write path for files based on their size. Files larger than or equal to the file batching threshold get written to their own chunks, while files smaller than the file batching threshold get batched with other files in chunks.

**Splitting up the chunk writer:**
The chunk writer contained all of the functionality for content defined chunking, batching files into chunks, and creating chunks. A new abstraction was created for each of these:
- Chunker: performs content defined chunking on a byte stream (this is technically a utility function).
- Batcher: batches entries into chunks based on a threshold.
- Uploader: manages upload tasks.

These new components should be easier to reason about while maintaining the same benefits of the chunk writer, primarily the parallel processing. Both the batcher and uploader have asynchronous interfaces to support the processing of chunks asynchronously. These interfaces are simpler and more well defined than the asynchronous interface provided by the chunk writer.

**Splitting the file write path based on size:**
The file set writer will now create two additive indexes: one for files larger than or equal to the file batching threshold and one for files smaller than the file batching threshold. These two indexes are merged to produce one additive index when the file set writer is closed. Files larger than or equal to the file batching threshold will go through an uploader and files smaller than the file batching threshold will go through a batcher.

Something noteworthy about this change is that content defined chunking is no longer applied to the files smaller than the file batching threshold. This is functionally equivalent to the current system because the chunk minimum size configuration would prevent split points being found in the middle of files below the file batching threshold and the artificial split point configuration is equivalent to the file batching threshold.


acohen4:
These changes are awesome - I find the current state of things is much easier to reason about. I'm still groking merge.go(s) and fileset/storage.go, but otherwise the changes generally look good. Particularly for the index writer, I like the direction

src/internal/miscutil/miscutil.go
```diff
@@ -25,3 +27,64 @@ func WithPipe(wcb func(w io.Writer) error, rcb func(r io.Reader) error) error {
...
+type Iterator struct {
+	peek     interface{}
+	dataChan chan interface{}
+	errChan  chan error
+}
+
+// NewIterator creates a new iterator.
+func NewIterator(ctx context.Context, iterate func(func(interface{}) error) error) *Iterator {
+	dataChan := make(chan interface{})
+	errChan := make(chan error, 1)
```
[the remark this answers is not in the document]
brycemcanally: So the goroutine can exit as soon as an error occurs. It doesn't make a huge difference, but it is a common pattern I follow for error channels.

src/internal/storage/fileset/writer.go
```diff
@@ -20,32 +23,42 @@ import (
...
 }
 
 func newWriter(ctx context.Context, storage *Storage, tracker track.Tracker, chunks *chunk.Storage, opts ...WriterOption) *Writer {
 	w := &Writer{
-		ctx:     ctx,
-		storage: storage,
-		tracker: tracker,
+		ctx:            ctx,
+		storage:        storage,
+		batchThreshold: DefaultBatchThreshold,
```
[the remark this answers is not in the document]
brycemcanally: yeah, the batching threshold in this PR is compatible with the current system, in terms of file hashes. Different batching thresholds can affect the hashes of certain files.

src/internal/storage/fileset/index/writer.go
```diff
@@ -54,77 +55,56 @@ func (w *Writer) writeIndex(idx *Index, level int) error {
...
 	} else {
 		return errors.New("either Index.Range or Index.File must be set, but not both.")
 	}
-	// Create an annotation for each index.
-	if err := l.cw.Annotate(&chunk.Annotation{
-		RefDataRefs: refDataRefs,
-		Data:        idx,
-	}); err != nil {
-		return err
+	l.buf.Reset()
```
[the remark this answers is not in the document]
brycemcanally: I stored it in the level writer so we can reuse the allocated memory. Resetting a bytes buffer keeps the allocated memory for future writes.

src/internal/storage/fileset/writer.go
```diff
@@ -102,80 +100,80 @@ func (w *Writer) checkPath(prevIdx, idx *index.Index) error {
...
-				if err := w.indexFunc(w.lastIdx); err != nil {
-					return err
-				}
-			}
-			w.lastIdx = idx
+// Copy copies a file to the file set writer.
+func (w *Writer) Copy(file File, datum string) error {
+	idx := file.Index()
+	size := index.SizeBytes(idx)
+	if size >= int64(w.batchThreshold) && chunk.FullDataRefs(idx.File.DataRefs) {
```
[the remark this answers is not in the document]
brycemcanally: Not for files produced before this change. It is correct that a file produced after this change will satisfy the second condition if it satisfies the first.

src/internal/storage/chunk/gc_test.go
```diff
@@ -1,81 +1,65 @@
...
-	rng := rand.New(rand.NewSource(seed))
-
-	cb := func(_ []*Annotation) error { return nil }
-	w := s.NewWriter(ctx, "test-writer", cb)
-	require.NoError(t, w.Annotate(&Annotation{}))
-	_, err := io.Copy(w, io.LimitReader(rng, size))
-	require.NoError(t, err)
-	require.NoError(t, w.Close())
-}
+// TODO: Reenable.
```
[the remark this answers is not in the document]
brycemcanally: `writeRandom` uses the chunk writer.

src/internal/storage/fileset/merge.go
```diff
@@ -112,24 +114,38 @@ func (mfr *MergeFileReader) Content(ctx context.Context, w io.Writer, opts ...ch
...
-			resolvedDataRefs = append(resolvedDataRefs, annotations[0].NextDataRef)
+	var hashes [][]byte
+	size := index.SizeBytes(mfr.idx)
+	if size >= DefaultBatchThreshold {
+		// TODO: Optimize to handle large files that can mostly be copy by reference?
+		if err := miscutil.WithPipe(func(w io.Writer) error {
+			r := mfr.chunks.NewReader(ctx, mfr.idx.File.DataRefs)
+			return r.Get(w)
+		}, func(r io.Reader) error {
+			uploader := mfr.chunks.NewUploader(ctx, "chunk-uploader-resolver", true, func(_ interface{}, dataRefs []*chunk.DataRef) error {
```
brycemcanally, line 127: @acohen4 We compute the stable hash for a file spread across multiple file sets here and use the `noUpload` option since we are just reading the file.

src/internal/storage/chunk/uploader.go
```diff
@@ -0,0 +1,108 @@
...
+func upload(ctx context.Context, client Client, chunkBytes []byte, pointsTo []ID, noUpload bool) (*DataRef, error) {
+	md := Metadata{
+		Size:     len(chunkBytes),
+		PointsTo: pointsTo,
+	}
+	createFunc := func(ctx context.Context, data []byte) (ID, error) {
+		res, err := client.Create(ctx, md, data)
+		return res, errors.EnsureStack(err)
+	}
+	if noUpload {
```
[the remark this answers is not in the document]
brycemcanally: Refer to comment below.

src/internal/miscutil/miscutil.go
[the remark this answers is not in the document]
acohen4: makes sense, hadn't put too much thought into this previously 👍

src/internal/storage/fileset/storage.go
```diff
@@ -225,17 +226,44 @@ func (s *Storage) getPrimitives(ctx context.Context, ids []ID) ([]*Primitive, er
...
+	return s.newPrimitive(ctx, &Primitive{
+		Additive:  additive,
+		Deletive:  deletive,
+		SizeBytes: size,
+	}, ttl)
+}
+
+func (s *Storage) concat(ctx context.Context, indexes []*index.Index) (*index.Index, error) {
+	mergeIndex := index.NewWriter(ctx, s.ChunkStorage(), "index-concat")
+	if err := index.Merge(ctx, s.ChunkStorage(), indexes, func(idx *index.Index) error {
```
PFedak, line 261: Just to make sure I'm following, the main functional change here is that the concat step of compaction will now just always concatenate indices, whereas before it might actually do an expensive rewrite of the underlying data?

src/internal/storage/fileset/index/merge.go
```diff
@@ -0,0 +1,57 @@
...
+	pq := stream.NewPriorityQueue(ss, compare)
+	return pq.Iterate(func(ss []stream.Stream) error {
+		for _, s := range ss {
+			if err := cb(s.(*indexStream).index); err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+}
```
PFedak, line 34: will also be nice to redo this with generics

src/internal/storage/fileset/writer.go
```diff
@@ -102,80 +101,82 @@ func (w *Writer) checkPath(prevIdx, idx *index.Index) error {
...
-				return nil, err
-			}
-		}
-	}
-	// Close the index writers.
 	additiveIdx, err := w.additive.Close()
 	if err != nil {
 		return nil, err
 	}
+	// The batcher writes to the additive small index, so close it first.
```
PFedak, line 155: could call it `additiveBatched` or something, not sure which is more important to convey

src/internal/storage/chunk/batcher.go
```diff
@@ -0,0 +1,133 @@
...
+	"golang.org/x/sync/semaphore"
+)
+
+// ChunkFunc is a function that provides the metadata for the entries in a chunk and a data reference to the chunk.
+type ChunkFunc = func([]interface{}, *DataRef) error
+
+// EntryFunc is a function that provides the metadata for an entry and a data reference to the entry in a chunk.
+// Size zero entries will have a nil data reference.
+type EntryFunc = func(interface{}, *DataRef) error
+
```
PFedak, line 15: did I miss somewhere where the metadata isn't `*index.Index/nil` or is this just for flexibility?

src/internal/storage/chunk/batcher.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+			return func() error {
+				for _, entry := range entries {
+					if err := b.entryFunc(entry.meta, entry.dataRef); err != nil {
+						return err
+					}
+				}
+				return nil
+			}, nil
+		}
```
PFedak, line 106: Having the two options feels a little messy to me vs just returning `dataRef, metas, []computedDataRefs` and having the caller use what they want, but I don't feel too strongly about it

src/internal/storage/fileset/writer.go
```diff
@@ -102,80 +101,82 @@ func (w *Writer) checkPath(prevIdx, idx *index.Index) error {
...
-				if err := w.indexFunc(w.lastIdx); err != nil {
-					return err
-				}
-			}
-			w.lastIdx = idx
+// Copy copies a file to the file set writer.
+func (w *Writer) Copy(file File, datum string) error {
+	idx := file.Index()
+	size := index.SizeBytes(idx)
+	if size >= int64(w.batchThreshold) && chunk.StableDataRefs(idx.File.DataRefs) {
```
PFedak, line 123: the stable condition is mainly for backwards compatibility, right? (And I guess if the batch threshold changes) Thought I saw a comment about that somewhere but maybe I was imagining it

src/internal/storage/fileset/writer.go
```diff
@@ -102,80 +101,82 @@ func (w *Writer) checkPath(prevIdx, idx *index.Index) error {
...
+			Path: idx.Path,
+			File: &index.File{
+				Datum: datum,
+			},
 		}
-	}
-	return nil
+		return w.uploader.Copy(copyIdx, idx.File.DataRefs)
+	}
+	// TODO: Optimize to handle large files that can mostly be copy by reference?
```
PFedak, line 136: for the below-threshold case or supporting older large files? I don't see how you can do a copy by reference in those cases with the current model

src/internal/storage/fileset/writer.go
```diff
@@ -102,80 +101,82 @@ func (w *Writer) checkPath(prevIdx, idx *index.Index) error {
...
 	}
+	// The batcher writes to the additive small index, so close it first.
+	if err := w.batcher.Close(); err != nil {
+		return nil, err
+	}
+	additiveSmallIdx, err := w.additiveSmall.Close()
+	if err != nil {
+		return nil, err
+	}
+	// TODO: Rethink / rework merging the two additive indexes?
```
PFedak, line 163: would the alternative be preserving a second additive field in the primitive filesets or you mean something like merging as we go/in the background?

PFedak (approved):
Apologies for taking so long, honestly this was a pleasure to read, I'm so happy you were able to get this done. I had a couple of questions just to make sure I'm not missing some changes, but the new logic itself is so much easier to reason about.

src/internal/storage/chunk/chunker.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+		data := buf[:n]
+		for _, b := range data {
+			chunkBuf = append(chunkBuf, b)
+			if len(chunkBuf) >= DefaultMaxChunkSize {
+				if err := cb(chunkBuf); err != nil {
+					return err
+				}
+				resetHash(hash)
+				chunkBuf = nil
```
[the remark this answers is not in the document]
brycemcanally: A buffer couldn't be freed until the client was done with it, which would add complexity that would outweigh the benefit of reuse imo.

src/internal/storage/fileset/fileset_test.go
```diff
@@ -81,8 +71,12 @@ func TestWriteThenRead(t *testing.T) {
 	fileNames := index.Generate("abc")
 	files := []*testFile{}
 	for _, fileName := range fileNames {
+		var datums []string
 		for _, datumInt := range random.Perm(maxDatums) {
-			datum := fmt.Sprintf("%08x", datumInt)
+			datums = append(datums, fmt.Sprintf("%08x", datumInt))
+		}
+		sort.Strings(datums)
```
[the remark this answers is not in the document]
brycemcanally: yeah, I'm not sure why it was implemented like this. We can probably leave it as is for now and revisit later.

src/internal/storage/fileset/storage.go
brycemcanally: Right. The existing system would, in general, run the chunks at the edges of the shards through the content defined chunking to get the stable split points. This isn't really necessary anymore and the tradeoff for removing it is slight potential for less deduplication but no longer having to move file content through the PFS master.

src/internal/storage/fileset/writer.go
brycemcanally: sure

src/internal/storage/chunk/batcher.go
brycemcanally: It is primarily for flexibility, but it would also be a circular dependency if it was typed to `*index.Index`.

src/internal/storage/chunk/batcher.go
brycemcanally: I prefer this approach because it makes things simpler for the client and the additional complexity in this abstraction is minimal. Also, it is slightly more performant since the hashes aren't computed for the chunk case (index entries).

src/internal/storage/fileset/writer.go
brycemcanally: right, existing files were not written with the batch threshold in mind so they might not be eligible for copy by reference.

src/internal/storage/chunk/chunker.go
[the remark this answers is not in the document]
PFedak: yeah, probably true

src/internal/storage/fileset/writer.go
brycemcanally: Large files are spread across multiple file sets during upload, so a compaction would rewrite them. I am thinking that is fine for now because it is only once, but it is less than ideal.

src/internal/storage/fileset/writer.go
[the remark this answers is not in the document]
acohen4: just to clarify, the batching threshold can later be updated to pretty much any value without affecting existing files right?

acohen4 (approved):

src/internal/storage/fileset/writer.go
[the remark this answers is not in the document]
brycemcanally: yeah

src/internal/storage/chunk/gc_test.go
```diff
@@ -2,7 +2,7 @@ package chunk
 
 import (
 	"context"
-	"io"
+	io "io"
```
PFedak, line 5: is this a weird editor/goimports thing or is there some collision this is avoiding?

PFedak (approved):

src/internal/storage/chunk/gc_test.go
brycemcanally: editor / goimport, I don't know why that happens.

PR #7443, opened 2022-03-28
Fix pachtcl port-forward so that it directs by default at the pod ports

acohen4:
copy of #7426


robert-uhl (approved):
I see that this copies PR #7426, which @msteffen approved. LGTM

PR #7446, opened 2022-03-28
PPS master event nits

acohen4:
This is some logic cleanup that was missed at the end of the "parallelize pps master" work. We now decide on whether an event was a delete by querying the database at the beginning of a pipeline controller's step.

### DIFF
--- a/src/server/pps/server/poller.go
+++ b/src/server/pps/server/poller.go
@@ -124,7 +124,7 @@ func (m *ppsMaster) pollPipelines(ctx context.Context) {
 						return errors.New("'pipelineName' label missing from rc " + rc.Name)
 					}
 					if !dbPipelines[pipeline] {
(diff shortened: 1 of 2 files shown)


src/server/pps/server/poller.go
```diff
@@ -284,26 +284,19 @@ func (m *ppsMaster) watchPipelines(ctx context.Context) {
...
+				case <-m.masterCtx.Done():
+					return errors.Wrap(err, "pipeline event arrived while master is restarting")
 				}
-			}
-			select {
-			case m.eventCh <- e:
-				continue
-			case <-m.masterCtx.Done():
-				return errors.Wrap(err, "pipeline event arrived while master is restarting")
+			case watch.EventError:
```
acohen4, line 298: we weren't handling this case before. On errors we would silently send a nil event to the master

PFedak (approved):
LGTM

PR #7449, opened 2022-03-29
Start to break out composite uses of k8s in pipeline_controller into kubeDriver

acohen4:


src/server/pps/server/poller.go
```diff
@@ -176,25 +172,17 @@ func (m *ppsMaster) pollPipelines(ctx context.Context) {
...
 		}
-		defer kubePipelineWatch.Stop()
+		defer cancel()
 	WatchLoop:
 		for {
 			select {
 			case <-ctx.Done():
 				return nil
-			case event := <-kubePipelineWatch.ResultChan():
+			case event := <-watch:
```
PFedak, line 185: oh one thing I was curious about here - I think really we should be doing `event, ok := <- watch` and restarting the watch if `!ok` - I think that explains the "stuck in a loop returning empty events" behavior

src/server/pps/server/kube_driver.go
```diff
@@ -0,0 +1,175 @@
...
+			LabelSelector: metav1.FormatLabelSelector(metav1.SetAsLabelSelector(
+				map[string]string{
+					"component": "worker",
+				})),
+			Watch: true,
+		})
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "failed to watch kubernetes pods")
+	}
+	return kubePipelineWatch.ResultChan(), func() { kubePipelineWatch.Stop() }, nil
```
PFedak, line 174: Does this not work?

```suggestion
	return kubePipelineWatch.ResultChan(), kubePipelineWatch.Stop, nil
```

src/server/pps/server/pipeline_controller.go
```diff
@@ -634,25 +625,9 @@ func (pc *pipelineController) deletePipelineResources() (retErr error) {
...
+	return pc.kd.DeletePipelineResources(pc.ctx, pc.pipeline)
 }
 
-//============================================================================================
-//
-// The below functions all acquire the counting semaphore lock: the pc.pcMgr.limiter to
-// prevent too many concurrent requests against kubernetes on behalf of different pipelines.
-// The functions below this block should not call the functions above it to avoid deadlocks.
-//
-//============================================================================================
```
PFedak, line 646: 👏

src/server/pps/server/pipeline_controller.go
```diff
@@ -79,6 +73,7 @@ type pipelineController struct {
 
 	bumpChan chan time.Time
 	pcMgr    *pcManager
+	kd       *kubeDriver
```
PFedak, line 76: seems like we can lose the k8s-relevant fields `namespace`, `etcdPrefix`) from the pipelineController

PFedak (approved):
LGTM

src/server/pps/server/pipeline_controller.go
acohen4: right!

src/server/pps/server/kube_driver.go
acohen4: no, I was being silly

src/server/pps/server/poller.go
acohen4: I'm not sure i follow the empty events bit. But stuck in a loop definitely makes sense

src/server/pps/server/poller.go
```diff
@@ -176,25 +172,27 @@ func (m *ppsMaster) pollPipelines(ctx context.Context) {
...
-						"component": "worker",
-					})),
-				Watch: true,
-			})
+		watch, cancel, err := m.kd.WatchPipelinePods(ctx)
 		if err != nil {
 			return errors.Wrap(err, "failed to watch kubernetes pods")
 		}
-		defer kubePipelineWatch.Stop()
+		defer cancel()
```
PFedak, line 179: weird technical go question here I'm going to have to look up: when is the function in a defer bound, exactly? It seems like the closure must be created at the `defer` line, so we will keep putting `defer`s on the stack which refer to old `watch`es. Probably not an issue, but if this wasn't how go worked and instead overwriting `cancel` changed it, this would be an issue as `cancel` could end up nil.

src/server/pps/server/poller.go
```diff
@@ -176,25 +172,27 @@ func (m *ppsMaster) pollPipelines(ctx context.Context) {
...
+			case event, ok := <-watch:
+				if !ok {
+					log.Warn("kubernetes pod watch unexpectdly ended - restarting watch")
+					cancel()
+					watch, cancel, err = m.kd.WatchPipelinePods(ctx)
+					if err != nil {
+						return errors.Wrap(err, "failed to watch kubernetes pods")
+					}
+					defer cancel()
+					continue
```
PFedak, line 194: I think I prefer just logging and returning `ErrContinue` here unless there's an obstacle I'm not seeing

PR #7451, opened 2022-03-29
Speed up pause/unpause test

robert-uhl:
There is an issue in the current branch where the timestamps used had only second resolution, which caused incorrect results when the `pachd` Pod restarted in the same second as the `pachd-config` ConfigMap was created. I believe that this slowed down recognition of the cluster paused state by the PauseStatus RPC.

End result is that the ENTERPRISE tests run about ten minutes faster.


acohen4 (approved):
LGTM! Nice catch

PR #7453, opened 2022-03-30
Address CVE-2021-3121

robert-uhl:
This updates package `github.com/gogo/protobuf` to v1.3.2 in the Helm chart and the Go RabbitMQ spout example. As part of the latter, it updates its `Dockerfile` to build with the latest Go 1.17 Alpine image rather than a no-longer available one.


acohen4 (approved):
LGTM

PR #7466, opened 2022-04-04
Support Dex Secret files for IDPs

tybritten:
Some dex IDPs (like google https://dexidp.io/docs/connectors/google/) require credentials in a file mounted to the pod so dex can use it. This enables it in the chart.


acohen4 (approved):
LGTM!

PR #7467, opened 2022-04-04
re-enable rootless ci tests

molinamelendezj:
### Re-enable rootless ci tests

- fixes the transient DNS issue inside `docker build`, by passing the `docker build --network=host` flag, which rest assured, the network specified in build is different from network specified in run. The network specified in build is only for downloading packages that are necessary for building image. The network specified in run is for the containers itself. There seems to be an issue with this `docker`/`minikube` version combination. Will punt the upgrade work on a later PR since that may break a lot more.
- also fixes an odd issue where the storage classes were not getting created, adding more memory to `minikube start` and wait for its rootless docker to be completely up seems to fix the issue.


acohen4 (approved):
LGTM!

PR #7470, opened 2022-04-05
Refresh commit into before compaction retries.

PFedak:
Pachd could get stuck finishing a commit from a killed job.


brycemcanally:
I kind of wonder if we would be better off just not allowing the stopping of a job that is in the finishing state? I think the stop job functionality is primarily intended for users to stop their code for some reason, not ours. We would actually need something that watches the job state or tracks the current commits to terminate the compaction early due to a stop job. I think considering a commit in the finishing state as not errorable by anything other than the master would be easier to reason about. We can add a bound to the finishing retry and error afterwards if being stuck in finishing is problematic, although that doesn't solve the problem as much as make it more apparent.

PR #7472, opened 2022-04-05
Enable TLS support for Postgres

tybritten:
`.Values.global.postgresql.postgresqlSSL` is currently not wired up because the bitnami image version we're using doesn't work with it. This replaces the bitnami image with our own and configures it.


BOsterbuhr:
I am not sure if this is an issue or not but shouldn't this:
https://github.com/pachyderm/pachyderm/blob/61a0bdf1e5e09a476a2c33ec360622ca4eb07438/etc/helm/pachyderm/templates/pachd/deployment.yaml#L67-L70
Be this instead?
```
{{- if .Values.global.postgresql.postgresqlSSL }}
- name: POSTGRES_SSL
  value: {{ .Values.global.postgresql.postgresqlSSL | quote }}
{{- end }}
```

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
```diff
@@ -55,14 +55,26 @@ spec:
           value: "300"
         - name: PGBOUNCER_IGNORE_STARTUP_PARAMETERS
           value: "extra_float_digits"
+        - name: SERVER_TLS_SSLMODE
```
claytonlemons, line 58: Should this be `PGBOUNCER_SERVER_TLS_SSLMODE`? It's currently what I have in my pgbouncer deployment

tybritten: Good catch. It works with the newer pg image we were using but I fixed it in both places.

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
```diff
@@ -72,9 +84,29 @@ spec:
           requests:
             cpu: 250m
             memory: 256M
+        volumeMounts:
+        {{- if .Values.global.postgresql.postgresqlSSLSecret }}
+        - mountPath: /pg-tls-cert
+          name: pg-tls-cert
+        {{- end }}
+        {{- if .Values.global.postgresql.postgresqlSSLSecret }}
```
acohen4, line 92: should be `...postgresqlSSLCACert`

acohen4 (approved):
needs one small fix, otherwise LGTM!

chainlink:
Given that were completely swapping out the PG Bouncer image this should be included in the review https://github.com/pachyderm/pg-bouncer

chainlink:
Doesn't look like the `PGBOUNCER_DATABASE` env var is included in the underlying image

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
```diff
@@ -41,8 +41,8 @@ spec:
          {{- else }}
           value: {{ .Values.global.postgresql.postgresqlHost }}
          {{- end }}
-        - name: PGBOUNCER_PORT
-          value: "5432"
+        - name: POSTGRESQL_PORT
```
chainlink, line 44: Is this the port PG is running on, or the port PG Bouncer is running on?

chainlink:
`PGBOUNCER_UNIX_SOCKET_DIR` is also missing from the underlying image (not sure if we need it anymore?

chainlink:
Looks like `PGBOUNCER_DEFAULT_POOL_SIZE` is missing from the underlying image

chainlink (changes requested):
Recommend adding the missing env vars (or removing them from the config)

chainlink:
It looks like the bitnami image supports SSL? https://github.com/bitnami/bitnami-docker-pgbouncer#securing-pgbouncer-traffic

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
```diff
@@ -55,14 +55,26 @@ spec:
...
+         {{- end }}
         - name: PGBOUNCER_UNIX_SOCKET_DIR
           value: " "
         - name: POSTGRESQL_PASSWORD
           valueFrom:
             secretKeyRef:
               name: {{ .Values.global.postgresql.postgresqlExistingSecretName | default "postgres" }} 
               key: {{ .Values.global.postgresql.postgresqlExistingSecretKey | default "postgresql-password" }}
-        image: bitnami/pgbouncer:1.16.0
+        image: pachyderm/pgbouncer:1.17.0-1
```
nitinjainsj, line 77: For this specific change, is there a reason to use our Pachydeerm version of pgbouncer? Bitmai version should already support TLS.

chainlink:
Have all of the options in https://github.com/pachyderm/pg-bouncer/blob/main/entrypoint.sh been exercised and verified they're being passed through correctly?

etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -64,10 +64,6 @@ spec:
           value: {{ required "postgresql username required" .Values.global.postgresql.postgresqlUsername | quote }}
         - name: POSTGRES_DATABASE
           value: {{ required "postgresql database name required" .Values.global.postgresql.postgresqlDatabase | quote }}
-        {{- if .Values.global.postgresql.ssl }}
```
chainlink, line 67: Do we use postgres SSL in pachd? I had a look in the code, we read in the env var, but it's not clear we do anything with it.

chainlink (approved):
Looks good, but need to regenerate the values schema file.

PR #7474, opened 2022-04-05
Ensure shard path ranges are disjoint.

PFedak:


src/internal/storage/fileset/index/reader.go
```diff
@@ -90,26 +96,36 @@ func (r *Reader) topLevel() pbutil.Reader {
...
+// For a range filter, this means the name is >= to the lower bound and the datum (if provided)
+// is >= the lower bound datum at the lower bound path itself
 // For a prefix filter, this means the name is >= to the prefix.
-func (r *Reader) atStart(name string) bool {
+func (r *Reader) atStart(name, datum string) bool {
 	if r.filter == nil {
 		return true
 	}
 	if r.filter.pathRange != nil && r.filter.pathRange.Lower != "" {
+		if name == r.filter.pathRange.Lower && datum != "" && r.filter.pathRange.LowerDatum != "" {
```
brycemcanally, line 107: Could we refactor the path range checks into helper functions like this:
```
func atStartRange(pathRange *pathRange, name, datum string) bool {
	if pathRange.Lower == "" {
		return true
	}
	if name >= pathRange.Lower {
		if name != pathRange.Lower || pathRange.LowerDatum == "" {
			return true
		}
		return datum >= pathRange.LowerDatum
	}
	return false
}
```
This is a bit easier to read for the range check. Also, I don't think the preexisting functionality when a range is set with an empty lower or upper is what we want, although it would be correct. It would go into the prefix check, which is fine because the prefix would be an empty string, but that is not really what we would want conceptually. It would be better to just return the results for a check based on the path range when the path range is set.

PFedak: Ah you're totally right, that's a much better idea. Falling back to the prefix there is pretty awkward

brycemcanally (approved):
lgtm

PR #7475, opened 2022-04-05
Remove the restart call from inside getRC

acohen4:


src/server/pps/server/pipeline_controller.go
```diff
@@ -717,14 +717,15 @@ func (pc *pipelineController) getRC(ctx context.Context, pi *pps.PipelineInfo) (
 		if errCount >= maxErrCount {
 			invalidRCState := errors.Is(err, errTooManyRCs) || errors.Is(err, errStaleRC)
 			if invalidRCState {
-				return pc.restartPipeline(ctx, pi, rc, fmt.Sprintf("could not get RC after %d attempts: %v", errCount, err))
+				restart = fmt.Sprintf("could not get RC after %d attempts: %v", errCount, err)
```
PFedak, line 720: i feel like this error message was never accurate, maybe `restart` should just be a bool since it will always be the same string anyway

src/server/pps/server/pipeline_controller.go
```diff
@@ -190,14 +190,17 @@ func (pc *pipelineController) step(timestamp time.Time) (isDelete bool, retErr e
 		return true, nil
 	}
 	var stepErr stepError
+	// TODO(msteffen) should this fail the pipeline? (currently getRC will restart
+	// the pipeline indefinitely)
+	rc, restart, err := pc.getRC(ctx, pi)
+	if restart != "" {
+		return false, pc.restartPipeline(ctx, pi, rc, restart)
+	}
+	if err != nil && !errors.Is(err, errRCNotFound) {
```
PFedak, line 199: maybe `errRCNotFound` just shouldn't be returned. Also seems like `errUnexpectedRC` can be deleted so it's just k8s errors that cause this

src/server/pps/server/pipeline_controller.go
```diff
@@ -190,14 +190,17 @@ func (pc *pipelineController) step(timestamp time.Time) (isDelete bool, retErr e
...
+		return false, err
+	}
 	errCount := 0
 	err = backoff.RetryNotify(func() error {
-		// TODO(msteffen) should this fail the pipeline? (currently getRC will restart
-		// the pipeline indefinitely)
-		rc, err := pc.getRC(ctx, pi)
-		if err != nil && !errors.Is(err, errRCNotFound) {
-			return err
-		}
```
PFedak, line 200: Hmm, I think I missed before, but I'm worried that when we moved refreshing pipeline info out of this retry loop we caused potential regressions on some of the reasons we adopted the top-level retry. Maybe worth thinking through this. Moving `getRC` out seems safer but I still worry that in the case of a retriable error we are going to run the logic on a version of the RC before any partial changes

PFedak (commented):
the changes with `getRC` itself look great, but I'm making myself a little nervous about the overall retry logic - maybe `step()` should just be calling `Bump()` in those cases instead of the inner retry or something along those lines.

src/server/pps/server/pipeline_controller.go
acohen4: Totally ok if we decide not to commit to this part of the change. 

The reason I think it's worthwhile is that in the current state of things, functions like getRC have their own retry logic that becomes nested within the "top level" retry logic which I think makes things more difficult to reason about.

That being said, I didn't notice that potential error and agree that it's cause for concern, and maybe for moving all 3 operations into the top-level retry. It seems like the main class of retriable error is one where `setPipelineState()` fails, which I suppose we interpret as potentially being due to some DB race issue that prompt us to re-`step` the pipeline controller? If that's the case I think `Bump()` may be a little difficult to use for this - we wouldn't have a clear way to say "we retried enough times"

PFedak: that's a fair point. Yeah, a lot of what drove the earlier changes in this structure was avoiding nested retries, and I think that's a good goal. It just violates some assumptions in subtle ways - it probably doesn't matter as much here, as we don't have any transaction-type guarantees, but we'd be explicitly exposing ourselves to those issues rather than them mostly being limited to weird cases where users modify the cluster directly. I think on the balance this is good, and in the long term these decisions probably won't matter because more "pure" logic will naturally avoid these questions.

PFedak (approved):
I don't think my vague concerns about moving `getRC` out of the retry loop are enough of a reason to rethink this. The guarantees of the rest of the logic w/r/t idempotence and error recovery should be enough.

PR #7476, opened 2022-04-05
Use a singleton pach address object in deploy.go

acohen4:


src/internal/minikubetestenv/deploy.go
```diff
@@ -32,8 +32,28 @@ const (
 	licenseKeySecretName   = "enterprise-license-key-secret"
 )
 
-// defensively lock around helm calls
-var mu sync.Mutex
+var (
+	mu sync.Mutex // defensively lock around helm calls
+
+	computedPachAddress *grpcutil.PachdAddress
```
acohen4, line 38: since we depend on the address to assign node ports, we compute it once to avoid potential race conditions

molinamelendezj:
[CORE-415](https://pachyderm.atlassian.net/browse/CORE-415)

molinamelendezj (approved):
👀

PR #7485, opened 2022-04-08
temporary circleci release definition

molinamelendezj:
This is a temporary circle ci release workflow definition to avoid dependency on local releases. This workflow will be improved/modernized over time and by no means the final configuration.


jrockway (approved):
Very nice!

PR #7489, opened 2022-04-08
Make pipeline logic and side-effect handling explicit

acohen4:


src/server/pps/server/pipeline_controller.go
```diff
@@ -54,6 +54,116 @@ func newPcManager() *pcManager {
 	}
 }
 
+type sideEffectName int32
+
+const (
+	sideEffectName_RESOURCES        sideEffectName = 0
```
PFedak, line 60: maybe KUBERNETES_RESOURCES to be super clear

src/server/pps/server/pipeline_controller.go
```diff
@@ -242,95 +356,98 @@ func (pc *pipelineController) fetchState() (*pps.PipelineInfo, context.Context,
...
-			target = pps.PipelineState_PIPELINE_STANDBY
+			return pps.PipelineState_PIPELINE_STANDBY, sideEffects, nil
 		}
-		return pc.setPipelineState(ctx, pi.SpecCommit, target, "")
+		return pps.PipelineState_PIPELINE_RUNNING, sideEffects, nil
 	}
 	if rc == nil {
 		// may happen if an external system deletes the RC
-		return pc.restartPipeline(ctx, pi, rc, "missing RC")
+		return pi.State, []sideEffect{RestartSideEffect()}, nil
```
PFedak, line 383: probably doesn't make sense to change in this PR, but the restart code also changes the pipeline's state. To better adhere to this model, I think it could make sense to return RESTARTING here and move that out of the actual restart code. Right now, if the restart side effect code modified the actual `pipelineInfo` passed in rather than just the database, this would be a bug

src/server/pps/server/pipeline_controller.go
```diff
@@ -242,95 +356,98 @@ func (pc *pipelineController) fetchState() (*pps.PipelineInfo, context.Context,
...
+}
+
+func (pc *pipelineController) apply(ctx context.Context, pi *pps.PipelineInfo, rc *v1.ReplicationController, target pps.PipelineState, sideEffects []sideEffect) error {
+	for _, s := range sideEffects {
+		if err := s.apply(ctx, pc, pi, rc); err != nil {
+			return err
+		}
+	}
+	if target != pi.State {
+		return pc.setPipelineState(ctx, pi.SpecCommit, target, "")
```
PFedak, line 450: I'm torn between returning the "desired" pipeline state, as you're doing here, and making this explicitly about a change, where you check against some explicit "don't care" or "same" value which gets returned instead of `pi.State` in all of those places. I think the desired state is better, but it would be good to eliminate the weirdness about the restarts in that case.

Part of the issue is that `setPipelineState` itself has the side effect of causing a bump. Avoiding the same start/end case will be less important if/when we remove the cases where we rely on multiple "steps" to completely resolve an update.

src/server/pps/server/pipeline_controller.go
```diff
@@ -54,6 +54,116 @@ func newPcManager() *pcManager {
...
+
+type sideEffectToggle int32
+
+const (
+	sideEffectToggle_NONE sideEffectToggle = 0
+	sideEffectToggle_UP   sideEffectToggle = 1
+	sideEffectToggle_DOWN sideEffectToggle = 2
+)
+
+// sideEffect returns a tuple first with the error representing the result of the side effect,
```
acohen4, line 76: need to fix this comment

PFedak (approved):
LGTM, I like where we landed on this

src/server/pps/server/pipeline_controller.go
acohen4: right. And more generally the side effects shouldn't be updating the pipelineInfo in the DB

PR #7493, opened 2022-04-11
Add buffering to debug dump stream

brycemcanally:
This PR adds buffering to the debug dump streaming. This should improve the reliability of the debug dump functionality in the case where workers crash or get deleted (e.g. standby / autoscaling) midstream. Also, the goroutine dump was switched to the proto format to enable the use of `go tool pprof` for exploring / visualizing the goroutine dump.


brycemcanally:
Yeah, your description of the issue is correct. Having the subset of the expected files from before the error was an intentional goal of this change, but it also makes the implementation simpler / better because we don't need to buffer the full worker tar stream in the local filesystem.

PR #7494, opened 2022-04-11
Decouple content defined chunking from file batching (#7442)

brycemcanally:
2.1.x PR for #7442


src/internal/serviceenv/config.go
```diff
@@ -33,9 +33,9 @@ type GlobalConfiguration struct {
 	PostgresUser                   string `env:"POSTGRES_USER,required"`
 	PostgresPassword               string `env:"POSTGRES_PASSWORD"`
 	PostgresMaxOpenConns           int    `env:"POSTGRES_MAX_OPEN_CONNS,default=10"`
-	PostgresMaxIdleConns           int    `env:"POSTGRES_MAX_IDLE_CONNS,default=2"`
+	PostgresMaxIdleConns           int    `env:"POSTGRES_MAX_IDLE_CONNS,default=10"`
 	PGBouncerMaxOpenConns          int    `env:"PG_BOUNCER_MAX_OPEN_CONNS,default=10"`
-	PGBouncerMaxIdleConns          int    `env:"PG_BOUNCER_MAX_IDLE_CONNS,default=2"`
+	PGBouncerMaxIdleConns          int    `env:"PG_BOUNCER_MAX_IDLE_CONNS,default=10"`
```
PFedak, line 38: were these also in the master PR? seems good regardless

PFedak (approved):
LGTM

src/internal/serviceenv/config.go
brycemcanally: yes, I noticed that we were frequently reconnecting to pgbouncer during my tests. The issue was that go sql's connection pooling is pretty aggressive about when it considers a connection is idle.

PR #7495, opened 2022-04-11
Significantly improve performance in FUSE mounts

lukemarsden:
(credit to @smalyala for this code, just cherry-picking it here).

In testing, this makes `ls` on a 100,000 file pachyderm repo take just 6 seconds, down from 25 minutes.


albscui (approved):
LGTM!

PR #7500, opened 2022-04-11
pachctl port-forward: cleanup on SIGHUP (window closed)

jrockway:
This will clean up port forwards as though C-c was pressed when closing the terminal window that's running pachctl port-forward.  Tested on Linux by closing the terminal that was running this; we clean up successfully and the next invocation works perfectly (no warning about not having cleaned it up).

This uses syscall which is generally [non-portable](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=670s) and should be gated by build tags, but SIGHUP is in syscall for OS X, Linux, and Windows, so I think we're OK not doing anything special here.  (In fact, I tried all GOOSes and didn't find any without SIGHUP.)

### DIFF
--- a/src/server/cmd/pachctl/cmd/cmd.go
+++ b/src/server/cmd/pachctl/cmd/cmd.go
@@ -655,8 +655,9 @@ This resets the cluster to its initial state.`,
 
 			fmt.Println("CTRL-C to exit")
 			ch := make(chan os.Signal, 1)
-			signal.Notify(ch, os.Interrupt)
-			signal.Notify(ch, syscall.SIGHUP)
+			// Handle Control-C, closing the terminal window, and pkill (and friends)
+			// cleanly.
+			signal.Notify(ch, os.Interrupt, syscall.SIGHUP, syscall.SIGTERM)
 			<-ch
 
 			return nil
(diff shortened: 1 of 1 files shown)


src/server/cmd/pachctl/cmd/cmd.go
```diff
@@ -655,6 +656,7 @@ This resets the cluster to its initial state.`,
 			fmt.Println("CTRL-C to exit")
 			ch := make(chan os.Signal, 1)
 			signal.Notify(ch, os.Interrupt)
+			signal.Notify(ch, syscall.SIGHUP)
```
PFedak, line 659: looks like notify actually takes `... os.Signal` so you could combine these. I don't know a lot about signals, does it make sense to include the other termination signals here?

PFedak (approved):
LGTM

src/server/cmd/pachctl/cmd/cmd.go
jrockway: good call on making it all one line.  will do that.

Yeah, I think we could add SIGTERM here.  Will do that.  Everything else (like SIGQUIT / Control-\) I think we should just let be unhandled; if someone is manually sending signals here then maybe they have some reason to exit uncleanly.  SIGTERM is sent if you are shutting down your computer, I think, so that's a reasonable time to clean up.

PR #7503, opened 2022-04-12
Expose configuration for compaction sharding

brycemcanally:
This PR exposes configuration for the compaction sharding. The two environment variables are `STORAGE_COMPACTION_SHARD_SIZE_THRESHOLD` and `STORAGE_COMPACTION_SHARD_COUNT_THRESHOLD`. The corresponding helm configurations are `compactionShardSizeThreshold` and `compactionShardCountThreshold`. The shard size corresponds to the total size of the files in a shard and the shard count corresponds to the total number of files in a shard. If either criteria is met, a shard will be created.

### DIFF
--- a/src/internal/serviceenv/config.go
+++ b/src/internal/serviceenv/config.go
@@ -103,17 +103,18 @@ type PachdSpecificConfiguration struct {
 
 // StorageConfiguration contains the storage configuration.
 type StorageConfiguration struct {
-	StorageMemoryThreshold         int64 `env:"STORAGE_MEMORY_THRESHOLD"`
-	StorageShardThreshold          int64 `env:"STORAGE_SHARD_THRESHOLD"`
-	StorageLevelFactor             int64 `env:"STORAGE_LEVEL_FACTOR"`
-	StorageUploadConcurrencyLimit  int   `env:"STORAGE_UPLOAD_CONCURRENCY_LIMIT,default=100"`
-	StoragePutFileConcurrencyLimit int   `env:"STORAGE_PUT_FILE_CONCURRENCY_LIMIT,default=100"`
-	StorageGCPeriod                int64 `env:"STORAGE_GC_PERIOD,default=60"`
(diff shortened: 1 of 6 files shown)


src/internal/serviceenv/config.go
```diff
@@ -103,17 +103,18 @@ type PachdSpecificConfiguration struct {
...
-	StorageGCPeriod                int64 `env:"STORAGE_GC_PERIOD,default=60"`
-	StorageChunkGCPeriod           int64 `env:"STORAGE_CHUNK_GC_PERIOD,default=60"`
-	StorageCompactionMaxFanIn      int   `env:"STORAGE_COMPACTION_MAX_FANIN,default=10"`
-	StorageFileSetsMaxOpen         int   `env:"STORAGE_FILESETS_MAX_OPEN,default=50"`
-	StorageDiskCacheSize           int   `env:"STORAGE_DISK_CACHE_SIZE,default=100"`
-	StorageMemoryCacheSize         int   `env:"STORAGE_MEMORY_CACHE_SIZE,default=100"`
+	StorageMemoryThreshold               int64 `env:"STORAGE_MEMORY_THRESHOLD"`
+	StorageCompactionShardSizeThreshold  int64 `env:"STORAGE_COMPACTION_SHARD_SIZE_THRESHOLD"`
+	StorageCompactionShardCountThreshold int64 `env:"STORAGE_COMPACTION_SHARD_COUNT_THRESHOLD"`
+	StorageLevelFactor                   int64 `env:"STORAGE_LEVEL_FACTOR"`
```
acohen4, line 109: will this be interpreted as default value of 0?

brycemcanally: Yeah, a zero value means either the environment variable is zero or it is unset. We create a functional option if it is greater than zero, otherwise the internal default value will be used.

acohen4 (approved):
LGTM

PR #7507, opened 2022-04-13
NewSQLTupleWriter and tests

albscui:
Added `NewSQLTupleWriter` which takes a `Tuple` and writes it to a target database.

### DIFF
--- a/src/internal/sdata/sdata.go
+++ b/src/internal/sdata/sdata.go
@@ -1,7 +1,6 @@
 package sdata
 
 import (
-	"context"
 	"database/sql"
 	"io"
 	"time"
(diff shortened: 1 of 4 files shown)


src/internal/sdata/sdata.go
```diff
@@ -108,11 +108,31 @@ func Copy(w TupleWriter, r TupleReader, row Tuple) (n int, _ error) {
...
 		return nil, err
 	}
+	return NewTupleFromTableInfo(info)
+}
+
+func NewTupleFromTableTx(ctx context.Context, tx *pachsql.Tx, tableName string) (Tuple, error) {
+	info, err := pachsql.GetTableInfoTx(tx, tableName)
+	if err != nil {
+		return nil, err
+	}
```
PFedak, line 135: this should use `NewTupleFromTableInfo`, too

src/internal/sdata/sdata.go
```diff
@@ -108,11 +108,31 @@ func Copy(w TupleWriter, r TupleReader, row Tuple) (n int, _ error) {
...
+
 func NewTupleFromTable(ctx context.Context, db *pachsql.DB, tableName string) (Tuple, error) {
 	info, err := pachsql.GetTableInfo(ctx, db, tableName)
 	if err != nil {
 		return nil, err
 	}
+	return NewTupleFromTableInfo(info)
+}
+
+func NewTupleFromTableTx(ctx context.Context, tx *pachsql.Tx, tableName string) (Tuple, error) {
```
PFedak, line 131: this and `NewTupleFromTable` are still unused, right? To be honest, now that they are just two function calls maybe we don't need them

PFedak (approved):
LGTM

PR #7513, opened 2022-04-14
Added comment on shard size/number thresholds params + dex database name override param in values.yaml

nadegepepin:


nadegepepin:
Mirrors https://github.com/pachyderm/pachyderm/pull/7509 on branch 2.1.x

PR #7517, opened 2022-04-16
WIP: Add oidc.additionalClients field

claytonlemons:
### Purpose
This PR enables users to specify additional clients for the dex OIDC provider.

### Motivation
This feature satisfies a specific requirement for JupyterHub users, which is to allow users to reuse the ID token returned by JupyterHub's authenticator to retrieve a session token for `pachctl auth login -t`. However, this feature may be useful for other services that would like to authenticate against the dex provider and reuse the ID token.

### Details
JupyterHub allows one to specify an `auth_state_hook` that grabs authentication information (including an ID token) returned by an OIDC provider (dex, in this case):
```python
 async def auth_state_hook(self, auth_state):
     if auth_state is not None:
         self.env["OIDC_ID_TOKEN"] = auth_state["access_token"] # dex stores the ID token as an "access_token"
 c.KubeSpawner.auth_state_hook = auth_state_hook
```
When a Jupyter notebook server starts, the OIDC_ID_TOKEN will be available in the user's environment, allowing one to run the following command automatically as part of a startup script:
```bash
echo ${OIDC_ID_TOKEN} | pachctl auth login --id-token
```
Now, when a user logs into JupyterHub and starts a notebook server, the user will already have a `pachctl` session token created for them. 

Note: it is not possible for JupyterHub to use the same upstream provider as dex. The reason is because `pachctl` expects the ID token to be issued by dex. Using the upstream provider directly will result in an ID token with a different issuer.

### Example
Here's what a user might add to their values.yaml for Pachyderm's helm chart:
```yaml
oidc:
  additionalClients:
    - id: jupyter-hub
      name: jupyter-hub
      secret: {user's-secret} # this secret is then used in the JupyterHub authenticator
      redirect_uris:
        - https://{jupyter-hub-domain}/hub/oauth_callback
```

### TODO

- [ ] Determine if user should be allowed to specify whether an additional client is also a trusted peer of the pachd client
- [ ] Update schema
- [ ] Finish unit tests

### DIFF
--- a/etc/helm/pachyderm/templates/pachd/config-secret.yaml
+++ b/etc/helm/pachyderm/templates/pachd/config-secret.yaml
@@ -3,9 +3,9 @@ SPDX-FileCopyrightText: Pachyderm, Inc. <info@pachyderm.com>
 SPDX-License-Identifier: Apache-2.0
 */ -}}
 
-{{- /* 
-if pachd.activateEnterprise is set, always run the bootstrap config. 
-On Installs, setting the pachd.enterpriseLicenseKey or pachd.activateEnterpriseMember 
+{{- /*
+if pachd.activateEnterprise is set, always run the bootstrap config.
+On Installs, setting the pachd.enterpriseLicenseKey or pachd.activateEnterpriseMember
 will also start bootstrapping.
 */ -}}
 {{ if or .Values.pachd.activateEnterprise (and .Release.IsInstall (or .Values.pachd.enterpriseLicenseKeySecretName .Values.pachd.enterpriseLicenseKey .Values.pachd.activateEnterpriseMember) ) }}
@@ -57,7 +57,7 @@ stringData:
   {{- end }}
 
   # TODO: Add config option for non embedded / multiple pachs?
-  # enterpriseClusters is the set of pachds covered by license service 
+  # enterpriseClusters is the set of pachds covered by license service
   enterpriseClusters: |
     - address: {{ .Values.pachd.enterpriseCallbackAddress }}
       id: {{ required "To register a pachd instance with an enterprise server, a unique ID must be set for this instance" .Values.pachd.oauthClientID }}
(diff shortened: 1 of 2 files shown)


etc/helm/pachyderm/templates/pachd/config-secret.yaml
```diff
@@ -132,6 +132,9 @@ stringData:
       {{- if .Values.console.enabled}}
       trusted_peers:
       - {{ .Values.console.config.oauthClientID | quote }}
+      {{- range .Values.oidc.additionalClients }}
+      - {{ .id | quote }}
```
claytonlemons, line 136: > Determine if user should be allowed to specify whether an additional client is also a trusted peer of the pachd client


As this is a WIP, would love some feedback about this^ particular issue. Should each additional client be automatically added as a trusted peer to the pachd client? If not, should we require the user to opt-in or opt-out of making an additional client a trusted peer?
For example, we could add the following field to the `additionalClients` schema:
```yaml
- id: my-additional-client
  name: my-additional-client
  secret: ...
  redirect_uris:
    - ...
  trusted_by_pachd: true # should the default be true or false?

**Comment by chainlink** [comment]
@acohen4 What do you think?

File: etc/helm/pachyderm/templates/pachd/config-secret.yaml (hunk 1 above)
Comment (acohen4) [reply etc/helm/pachyderm/templates/pachd/config-secret.yaml:136]: Good question. Since pachd is the only system minting access tokens, I think that it's fair to assume at this point that all systems that are configured to authenticate against it will want to permission to request pachyderm access tokens. In which case what you have proposed so far is great. 

What do you think @chainlink ?

File: etc/helm/pachyderm/templates/pachd/config-secret.yaml [hunk 2]
```diff
@@ -140,6 +143,15 @@ stringData:
...
     {{- end }}
+    {{- range .Values.oidc.additionalClients }}
+    - id: {{ .id }}
+      name: {{ .name }}
+      secret: {{ .secret }}
+      redirect_uris:
+      {{- range .redirect_uris }}
+      - {{ . }}
+      {{- end }}
+    {{- end }}
```
zjgoodman, line 154: would recommend just spreading what's in the values.yaml instead of manually duplicating it yourself
```yaml
{{- toYaml .Values.oidc.additionalClients | nindent 4 }}
```

### Proof that this works
Template
```yaml
  oidcClients: |
    - id: sample
      name: sample
      secret: sample
      redirect_uris:
      - sample
    {{- toYaml .Values.additionalClients | nindent 4 }}
```
Values
```yaml
additionalClients:
  - id: hi
    name: sup
    secret: secret
    redirect_uris:
      - url1
      - url2
```
Rendered output
```yaml
  oidcClients: |
    - id: sample
      name: sample
      secret: sample
      redirect_uris:
      - sample
    - id: hi
      name: sup
      redirect_uris:
      - url1
      - url2
      secret: secret
```

chainlink:
👋 Hey guys, we've decided to go with https://github.com/pachyderm/pachyderm/pull/8002 which fits with our new gitops tooling (reading things in directly to pach vs the config pod)

msteffen:

PR #7518, opened 2022-04-18
Working script

robert-uhl:
The major update is making sure that this creates a working cluster in
the specified region, and that it is durable when the default output
option is something other than JSON.

Notably, this does not yet fix the tags.


molinamelendezj (approved):
👍🏼

PR #7520, opened 2022-04-19
PPS Master Unit Testing Groundwork [CORE-360]

acohen4:


src/server/pps/server/server.go
```diff
@@ -45,14 +45,40 @@ type Env struct {
 	Logger            *logrus.Logger
 	Config            serviceenv.Configuration
 }
+type envOpts struct {
```
acohen4, line 48: will revert this file

src/server/pps/server/pipeline_state_driver.go
```diff
@@ -0,0 +1,272 @@
...
+	msd := &mockStateDriver{}
+	msd.reset()
+	msd.eChan = make(chan *watch.Event, 1)
+	msd.doneEChan = make(chan struct{}, 5)
+	return msd
+}
+
+func (msd *mockStateDriver) SetState(ctx context.Context, specCommit *pfs.Commit, state pps.PipelineState, reason string) error {
+	if pipeline, ok := msd.specCommits[specCommit.ID]; ok {
+		if pi, ok := msd.pipelines[pipeline]; ok {
```
PFedak, line 185: can't decide if it's important to capture, but strictly speaking `SetPipelineState` will only modify the version of the pipeline at that spec commit, which may no longer be current. This is important for concurrent updates, so I expect something like 

- send pipeline event for a failed pipeline while blocking `getRC` from completing
- update pipeline externally
- unblock `getRC` to let the old step continue with a stale pipeline info

to behave incorrectly in this model, with the pipeline staying in failure instead of going into starting etc

even if we don't want to capture that, this model feels a little weird - you could just be using the repo name from the spec commit, right? Or is the idea to test that the spec commit ID made it into the map properly?

src/server/pps/server/pipeline_state_driver.go
```diff
@@ -0,0 +1,272 @@
...
+	})
+	return pi, err
+}
+
+func (sd *stateDriver) loadLatestPipelineInfo(ctx context.Context, pipeline string, message *pps.PipelineInfo) error {
+	specCommit, err := ppsutil.FindPipelineSpecCommit(ctx, sd.pfsApi, *sd.txEnv, pipeline)
+	if err != nil {
+		return errors.Wrapf(err, "could not find spec commit for pipeline %q", pipeline)
+	}
+	// TODO: which context should be passed here? master's ctx?
```
PFedak, line 160: what situations are you thinking of where the request/pipeline controller context wouldn't be appropriate? We're already a few function calls deep at this point, I guess

src/server/pps/server/pipeline_controller_test.go
```diff
@@ -0,0 +1,686 @@
...
+			},
+			expectedStates: []pps.PipelineState{
+				pps.PipelineState_PIPELINE_STARTING,
+				// STANDBY is set consecutively because ppsutil.SetPipelineState
+				// will still update the PipelineInfo even if the desired state
+				// is already set
+				pps.PipelineState_PIPELINE_STANDBY,
+				pps.PipelineState_PIPELINE_STANDBY,
+				pps.PipelineState_PIPELINE_RUNNING,
+				pps.PipelineState_PIPELINE_STANDBY,
```
PFedak, line 232: ok, I'm definitely missing something - isn't the `assertWhen` going to succeed when the last state is RUNNING? Is it just going back quickly enough that STANDBY shows up anyway?

src/server/pps/server/pipeline_controller_test.go
```diff
@@ -0,0 +1,686 @@
...
+	require.Equal(t, 200, inspectCount)
+}
+
+func TestAutoscalingManyTasks(t *testing.T) {
+	stateDriver, infraDriver, mockPachd := ppsMasterHandles(t)
+	pipeline := tu.UniqueString(t.Name())
+	mockAutoscaling(mockPachd, 100, 1)
+	mockPachd.PFS.InspectCommit.Use(func(context.Context, *pfs.InspectCommitRequest) (*pfs.CommitInfo, error) {
+		// add some delay so that Scaling Up will kick in before the commit is otherwise considered done
+		time.Sleep(time.Second)
```
PFedak, line 285: I'd generally prefer to do things like this using channels (or directly checking the condition you're waiting for) if it's not too messy

src/server/pps/server/pipeline_controller.go
```diff
@@ -717,18 +652,18 @@ func (pc *pipelineController) scaleDownPipeline(ctx context.Context, pi *pps.Pip
 // if errorState {
 //   return pc.restartPipeline("entered error state")
 // }
```
PFedak, line 654: update comment

src/server/pps/server/pipeline_controller.go
```diff
@@ -295,22 +296,22 @@ func (pc *pipelineController) step(timestamp time.Time) (isDelete bool, retErr e
...
-	targetState, sideEffects, err := evaluate(pi, rc)
+	targetState, sideEffects, reason, err := evaluate(pi, rc)
 	if err != nil {
 		return false, err
 	}
 	errCount := 0
 	err = backoff.RetryNotify(func() error {
 		// Create/Modify/Delete pipeline resources as needed per new state
-		return pc.apply(ctx, pi, rc, targetState, sideEffects)
+		return pc.apply(ctx, pi, rc, targetState, sideEffects, reason)
```
PFedak, line 314: hmm, feels like we narrowly avoided a bug before in the case of an external RC deletion while RUNNING? `getRC` would return a nil `rc` but `restart` would be `false`, `evaluate` would say we need to restart, `apply` would error on the restart because of how that function worked. It was okay because we errored out of the backoff, too, and would just print the failure at the top level. Still makes me a little nervous

PFedak (commented):
I still want to look through the tests you've added to see if there's anything that I'd like to see. I think we can probably delete the `TestStandbyTransitions` test and some of the other `Autoscaling` ones now, which is nice

src/server/pps/server/infra_driver.go
```diff
@@ -0,0 +1,149 @@
...
+const (
+	mockInfraOp_CREATE = 0
+	mockInfraOp_DELETE = 1
+	mockInfraOp_READ   = 2
+)
+
+type mockInfraDriver struct {
+	rcs           map[string]v1.ReplicationController // indexed by pipeline name
+	calls         map[string]map[mockInfraOp]int      // indexed by pipeline name
+	elapsedScales map[string][]int32                  // indexed by pipeline name
```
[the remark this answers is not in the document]
acohen4: totally agree 👍

src/server/pps/server/pipeline_controller.go
```diff
@@ -336,63 +337,40 @@ func (pc *pipelineController) step(timestamp time.Time) (isDelete bool, retErr e
...
-			return pps.PipelineState_PIPELINE_STANDBY, sideEffects, nil
+			return pps.PipelineState_PIPELINE_STANDBY, sideEffects, "", nil
 		}
-		return pps.PipelineState_PIPELINE_RUNNING, sideEffects, nil
+		return pps.PipelineState_PIPELINE_RUNNING, sideEffects, "", nil
 	}
 	if rc == nil {
 		// may happen if an external system deletes the RC
-		return pps.PipelineState_PIPELINE_RESTARTING, []sideEffect{RestartSideEffect()}, nil
+		return pps.PipelineState_PIPELINE_RESTARTING, []sideEffect{RestartSideEffect()}, "missing RC", nil
```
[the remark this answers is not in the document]
acohen4: yeah it does seem silly

src/server/pps/server/pipeline_state_driver.go
acohen4: I actually think this TODO is outdated. The function should be passed the PC's ctx

src/server/pps/server/pipeline_controller_test.go
```diff
@@ -0,0 +1,686 @@
...
+		return &pfs.CommitInfo{}, nil
+	})
+	mockPachd.PFS.SubscribeCommit.Use(func(req *pfs.SubscribeCommitRequest, server pfs.API_SubscribeCommitServer) error {
+		for i := 0; i < commitCount; i++ {
+			server.Send(&pfs.CommitInfo{
+				Commit: client.NewCommit(req.Repo.Name, "", uuid.NewWithoutDashes()),
+			})
+		}
+		// keep the subscribe stream open longer
+		time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
acohen4: yes that's way better :)

src/server/pps/server/pipeline_controller_test.go
```diff
@@ -0,0 +1,686 @@
...
+					if k >= len(actualStates) || actualStates[k] != expected {
+						break
+					}
+					k++
+					if j == len(states)-1 {
+						return nil
+					}
+				}
+			}
+			return errors.New("change hasn't reflected")
```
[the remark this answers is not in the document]
acohen4: No you're right. I'm going to separate out the two cases

src/server/pps/server/pipeline_controller_test.go
acohen4: yeah you're right. I removed `assertWhen` due to discussion above

src/server/pps/server/pipeline_state_driver.go
```diff
@@ -0,0 +1,272 @@
...
+	states      map[string][]pps.PipelineState // tracks all of the
+	eChan       chan *watch.Event
+	doneEChan   chan struct{} // supports closing eChan
+}
+
+func newMockStateDriver() *mockStateDriver {
+	msd := &mockStateDriver{}
+	msd.reset()
+	msd.eChan = make(chan *watch.Event, 1)
+	msd.doneEChan = make(chan struct{}, 5)
```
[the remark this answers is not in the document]
acohen4: I changed the name of it here. I was using it to force closing the channel that is otherwise closed by ctx.Done()

src/server/pps/server/pipeline_state_driver.go
acohen4: yeah the model here should be inverted so that `pipelines` -> `specCommit` -> `pipelineInfo`

src/server/pps/server/monitor.go
```diff
@@ -109,6 +107,11 @@ func (pc *pipelineController) monitorPipeline(ctx context.Context, pipelineInfo
 		eg.Go(func() error {
 			defer close(ciChan)
 			return backoff.RetryNotify(func() error {
+				select {
+				case <-ctx.Done():
+					return errutil.ErrBreak
+				default:
+				}
```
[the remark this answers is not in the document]
acohen4: I changed this the RetryWithCancel

src/server/pps/server/pipeline_controller.go
acohen4: right (yikes)

src/server/pps/server/pipeline_state_driver.go
```diff
@@ -157,116 +158,143 @@ func (sd *stateDriver) loadLatestPipelineInfo(ctx context.Context, pipeline stri
...
 }
 
-func (msd *mockStateDriver) Watch(ctx context.Context) (<-chan *watch.Event, func(), error) {
+func (d *mockStateDriver) Watch(ctx context.Context) (<-chan *watch.Event, func(), error) {
 	go func() {
 		select {
 		case <-ctx.Done():
-			close(msd.eChan)
+			d.eChan <- &watch.Event{Type: watch.EventError, Err: ctx.Err()}
+			close(d.eChan)
```
PFedak, line 232: could close in a defer to handle both, but not super important

src/server/pps/server/pipeline_state_driver.go
```diff
@@ -157,116 +158,143 @@ func (sd *stateDriver) loadLatestPipelineInfo(ctx context.Context, pipeline stri
...
+	d.eChan <- &watch.Event{
 		Key:  []byte(fmt.Sprintf("%s@%s", pi.Pipeline.Name, pi.SpecCommit.ID)),
 		Type: et,
 	}
 }
 
-func (msd *mockStateDriver) cancelWatch() {
-	msd.doneEChan <- struct{}{}
+func (d *mockStateDriver) cancelWatch() {
+	d.closeEChan <- struct{}{}
```
PFedak, line 289: can also be a close

PFedak (approved):
LGTM, just depends if you want to remove the (very slow) integration test versions of the autoscaling/standby tests you're reproducing here

PR #7522, opened 2022-04-19
CORE-387 New Egress API

albscui:
Added a new `Egress` interface for copying data from Pachyderm to an external storage system. Currently, we support 2 types of outputs: Object Storage, and SQL databases.

The new RPC can be called via the PFS API client, or a pipeline's `"egress"` field.

Example:

First create a repo `input_repo` with `/test_table/0000`. Add some data in the form of CSV or JSON. Then create a table in your target databaes called `test_table`.

```json
{
    "pipeline": {
        "name": "egress"
    },
    "input": {
        "pfs": {
            "repo": "input_repo",
            "glob": "/",
            "name": "in"
        }
    },
    "transform": {
        "image": "ubuntu:22.04",
        "cmd": [
            "bash"
        ],
        "stdin": [
            "cp -r /pfs/in/* /pfs/out"
        ]
    },
    "egress": {
        "sql_database": {
            "url": "snowflake://albertpachyderm@CDKSVXM-PN23804/ALBERT_TEST/PUBLIC?warehouse=COMPUTE_WH",
            "file_format": {
                "type": "CSV"
            },
            "secret": {
                "k8s_secret": "snowflakesecret",
                "key": "PACHYDERM_SQL_PASSWORD"
            }
        }
    }
}
```

To do: expose this via `pachctl`, something like:
```
pachctl egress <src_repo> <target_url>
```


src/internal/dockertestenv/postgres.go
```diff
@@ -48,39 +48,39 @@ func NewTestDBConfig(t testing.TB) serviceenv.ConfigOption {
...
+		c.PGBouncerPort = PGBouncerPort
 
 		c.PostgresUser = testutil.DefaultPostgresUser
 	}
 }
 
 func NewTestDB(t testing.TB) *pachsql.DB {
 	opts := NewTestDBOptions(t)
-	return testutil.NewTestDB(t, opts)
+	return testutil.OpenDB(t, opts...)
```
albscui, line 63: This was a bug because both `NewTestDBOptions` and `NewTestDB` call `CreateEphemeralDB`, thereby unintentionally creating two databases instead of one.

src/internal/dockertestenv/postgres.go
```diff
@@ -48,39 +48,36 @@ func NewTestDBConfig(t testing.TB) serviceenv.ConfigOption {
...
+		c.PGBouncerPort = PGBouncerPort
 
 		c.PostgresUser = testutil.DefaultPostgresUser
 	}
 }
 
 func NewTestDB(t testing.TB) *pachsql.DB {
-	opts := NewTestDBOptions(t)
-	return testutil.NewTestDB(t, opts)
+	return testutil.OpenDB(t, NewTestDBOptions(t)...)
```
albscui, line 62: Fixing a bug here because both `NewTestDBOptions()` and `NewTestDB()` call `CreateEphemeralDB()` which means this is creating two databases but we only use one.

src/internal/dockertestenv/postgres.go
```diff
@@ -20,17 +20,17 @@ import (
...
 
-func postgresHost() string {
+func PostgresHost() string {
 	return getDockerHost()
 }
 
-func pgBouncerHost() string {
-	return postgresHost()
+func PGBouncerHost() string {
+	return PostgresHost()
```
PFedak, line 33: I'm assuming this is for a test that needs to build a URL? It looks like we already have `dbutil.GetDSN()`, does that allow you to get around the refactoring you did here or is there more you need?

src/internal/obj/factory.go
```diff
@@ -471,43 +472,52 @@ func (s ObjectStoreURL) String() string {
 	return fmt.Sprintf("%s://%s/%s", s.Scheme, s.Bucket, s.Object)
 }
 
+type URLFormatError struct {
+	Msg string
+	URL string
+}
+
+func (e *URLFormatError) Error() string { return fmt.Sprintf("%s: %s", e.Msg, e.URL) }
+
```
PFedak, line 481: I think we can drop the changes to this file now

src/internal/sdata/sdata.go
```diff
@@ -90,8 +91,8 @@ func NewTupleFromColumnTypes(cTypes []*sql.ColumnType) (Tuple, error) {
 	return row, nil
 }
 
-// Copy copies a tuple from w to r.  Row is used to indicate the correct shape of read data.
-func Copy(w TupleWriter, r TupleReader, row Tuple) (n int, _ error) {
+// Copy copies a tuple from r to w. Row is used to indicate the correct shape of read data.
+func Copy(r TupleReader, w TupleWriter, row Tuple) (n int, _ error) {
```
PFedak, line 95: for what it's worth, go standard library functions seem to put writers first

src/pfs/pfs.proto
```diff
@@ -462,6 +462,51 @@ message RunLoadTestResponse {
   google.protobuf.Duration duration = 5;
 }
 
+message ObjectStorageEgressOptions {
```
PFedak, line 465: maybe we can drop `Options` from the type name

src/pfs/pfs.proto
```diff
@@ -462,6 +462,51 @@ message RunLoadTestResponse {
...
+  message FileFormat {
+    enum Type {
+        UNKNOWN = 0;
+        CSV = 1;
+        JSON = 2;
+        PARQUET = 3;
+    }
+    message Options {
+      repeated string json_field_names = 1;
+    }
```
PFedak, line 479: reiterating some comments I already made, but I don't think this needs to be nested another level in.

src/pfs/pfs.proto
```diff
@@ -462,6 +462,51 @@ message RunLoadTestResponse {
...
+    Type type = 1;
+    Options options = 2;
+  }
+  message Secret {
+    string k8s_secret = 1;
+    string key = 2;
+    string mount_path = 3;
+    string env_var = 4;
+  }
```
PFedak, line 488: `mount_path` is ignored, right? and I'd rather we not give people the option to set `env_var` if there's only one working value for it, unless the plan is to unify this with the `pps` type when we get a chance

src/pfs/pfs.proto
```diff
@@ -462,6 +462,51 @@ message RunLoadTestResponse {
...
+    ObjectStorageEgressOptions obj = 2;
+    SQLEgressOptions sql = 3;
+  }
+}
+message EgressResponse {
+  message SQLEgressResult {
+    map<string, int64> sql_rows_written = 1;
+  }
+  oneof ret {
+    int64 bytes_written = 1;
```
PFedak, line 505: if we're doing a one of, both should be wrapped in messages just in case

src/server/pfs/server/egress.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/miscutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
+	"github.com/pachyderm/pachyderm/v2/src/internal/sdata"
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset"
+	"github.com/pachyderm/pachyderm/v2/src/pfs"
+	"golang.org/x/net/context"
+)
+
+type egressResult map[string]int64
```
PFedak, line 17: you could just use the actual `SQLEgressResult` type

src/internal/dockertestenv/postgres.go
PFedak: wow, the combination of name and signature in `NewTestDBOptions(t testing.TB) []dbutil.Option` seems almost deliberately misleading, not surprised that happened

src/server/pfs/server/egress.go
```diff
@@ -0,0 +1,99 @@
...
+
+func getEgressPassword() (string, error) {
+	const passwordEnvar = "PACHYDERM_SQL_PASSWORD" // TODO move this to a package for sharing
+	password, ok := os.LookupEnv(passwordEnvar)
+	if !ok {
+		return "", errors.Errorf("must set %v", passwordEnvar)
+	}
+	return password, nil
```
PFedak, line 25: should we enforce this in `ValidateInput` during pipeline creation? Either the user needs to put the secret field in the egress, or we hope they provide it some way, I feel like it makes some things simpler if we assume we're the ones creating it

src/server/pfs/server/egress.go
```diff
@@ -0,0 +1,99 @@
...
+	// cache tableInfos because multiple files can belong to the same table
+	tableInfos := make(map[string]*pachsql.TableInfo)
+	rowsWritten := make(egressResult)
+	err = src.Iterate(ctx, func(fi *pfs.FileInfo, file fileset.File) error {
+		if fi.FileType != pfs.FileType_FILE {
+			return nil
+		}
+
+		tableName := strings.Split(fi.File.Path, "/")[1]
+		tableInfo, prs := tableInfos[tableName]
```
PFedak, line 61: usually people use `ok` for this, I guess you mean `pr`e`s`ent?

src/server/worker/pipeline/transform/registry.go
```diff
@@ -461,10 +461,21 @@ func deserializeDatumSet(any *types.Any) (*DatumSet, error) {
...
+		request.Target = &pfs.EgressRequest_Obj{Obj: egress.GetObj()}
+	} else {
+		switch egress.Target.(type) {
+		case *pps.Egress_Obj:
+			request.Target = &pfs.EgressRequest_Obj{Obj: egress.GetObj()}
+		case *pps.Egress_Sql:
+			request.Target = &pfs.EgressRequest_Sql{Sql: egress.GetSql()}
+		}
+	}
+	if _, err := client.Egress(client.Ctx(), &request); err != nil && !pfsserver.IsFileNotFoundErr(err) {
```
PFedak, line 478: maybe put a TODO here around detecting unrecoverable errors in the SQL case (bad password, datatype mismatch)

PFedak (commented):
looking a lot better, might want to float the proto changes in front of someone else, too

src/internal/sdata/sql.go
```diff
@@ -2,36 +2,82 @@ package sdata
...
 
+// limit the number of rows
+const rowLimit = 1000
+
 // SQLTupleWriter writes tuples to a SQL database.
 type SQLTupleWriter struct {
 	tx              *pachsql.Tx
+	tableInfo       *pachsql.TableInfo
 	insertStatement string
+	buf             []interface{}
```
PFedak, line 20: Making this `[]Tuple` so you don't have to keep checking for the column length seems good, unless then executing the actual query is annoying. Speaking of which, do we validate anywhere that the tuple lengths are compatible? seems like it would be good to error early

src/internal/sdata/sql.go
```diff
@@ -2,36 +2,82 @@ package sdata
...
-	_, err := m.tx.Exec(m.insertStatement, t...)
-	return errors.EnsureStack(err)
+	if len(m.buf)/len(m.tableInfo.Columns) >= rowLimit {
+		m.Flush()
+	}
+	for i := range t {
+		v := reflect.New(reflect.TypeOf(t[i]).Elem())
+		v.Elem().Set(reflect.ValueOf(t[i]).Elem())
+		m.buf = append(m.buf, v.Interface())
+	}
```
PFedak, line 31: this feels general enough to make it a method on `Tuple`

src/internal/sdata/sql.go
```diff
@@ -2,36 +2,82 @@ package sdata
...
+func (m *SQLTupleWriter) GeneratePreparedStatement() (*pachsql.Stmt, error) {
+	if len(m.buf) == 0 {
+		return nil, nil
+	}
+	driverName := m.tx.DriverName()
+	placeholders := []string{} // a list of (?, ?, ...)
+
+	// construct list of placeholders by accumulating elements into a row first
+	row := []string{}
+	for i := 0; i < len(m.buf); i++ {
```
PFedak, line 62:
```suggestion
	for i := range m.buf {
```

src/server/pfs/server/egress.go
```diff
@@ -0,0 +1,119 @@
...
+			}
+			tableInfos[tableName] = tableInfo
+			result, err := tx.Exec(fmt.Sprintf("DELETE FROM %s.%s", tableInfo.Schema, tableInfo.Name))
+			if err != nil {
+				return errors.EnsureStack(err)
+			}
+			_, err = result.RowsAffected()
+			if err != nil {
+				return errors.EnsureStack(err)
+			}
```
PFedak, line 87: not really sure what this error would mean, but seems like the right way to handle it anyway

src/server/pps/server/api_server.go
```diff
@@ -1493,6 +1493,24 @@ func (a *apiServer) validateEnterpriseChecks(ctx context.Context, req *pps.Creat
...
+		return nil
+	}
+	if target, ok := egress.Target.(*pps.Egress_SqlDatabase); ok {
+		secret := target.SqlDatabase.GetSecret()
+		if secret == nil {
+			return errors.New("egress.sql_database.secret is required")
+		}
+		if secret.K8SSecret == "" || secret.Key == "" {
+			return errors.New("egress.sql_database.secret.k8s_secret and egress.sql_database.secret.key are required")
+		}
```
PFedak, line 1509: I guess also check `Columns` isn't set for `CSV`

src/internal/sdata/sql.go
```diff
@@ -2,36 +2,82 @@ package sdata
...
 import (
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
 )
 
+// limit the number of rows
```
PFedak, line 12: maybe mention this is specifically about batching here, though it's not a big file I guess

PFedak (approved):
I think this looks good enough to merge given our discussion in the meeting today. Had a couple of small suggestions but given how awkward passing arguments into the prepared statement is you can probably ignore some of that.

src/server/pfs/server/egress.go
albscui: I'm gonna just remove this since the some db drivers don't even implement `RowsAffected()`.

src/server/pps/server/api_server.go
albscui: I actually think `Columns` can be useful for both CSV and JSON. For example, a user could potentially have data that represents a subset of the columns in the target table, and want all the other columns to be filled with null values.

PR #7525, opened 2022-04-19
Add a proxy to expose all external services on one port

jrockway:
This PR adds the envoy proxy, allowing all Pachyderm traffic to be served on a single port.  What we add is:

* The option to create one external port (NodePort, LoadBalancer, or ClusterIP) that serves all Pachyderm services on that port; console, auth, pachd, s3, etc.  Via the helm values at install time.  (You can also "upgrade" to add the proxy, of course.)
* There is the option to also serve legacy ports through the proxy.  Envoy always opens these, but the k8s service is gated by Helm (so you can still use a LoadBalancer or NodePort for legacy services; that's like 30650 for gRPC, etc.)
* Both of these options give you nice access logs showing every Pachyderm request (how long it took, what the outcome was -- did the server disconnect, did the client disconnect, were there no healthy pachd instances at the time, etc.) The tests output these logs, so you can get a concise summary of what the tests were doing to Pachyderm.
* This feature is totally optional right now and disabled by default, but the proxy is used for all the tests, and all test access to Pachyderm goes through the proxy (includes minikubetestenv tests, and tests to auth and S3; not just grpc requests).
* Envoy is not run as root, so it's compatible with clusters that restrict privileged operations.  We don't try to bind privileged ports either.
* We create a new service for Envoy to use to talk to pachd and console; this way the user can renumber the ports in `pachd.service` and we don't break.  It also lets Envoy discover all replicas and avoid routing through kube-proxy (`clusterIP: None`).
* I tested long file uploads with a small program that uploads a file a 1 byte per second for an hour.  Maybe we should check this in somewhere.

There is some excitement going on.  Envoy needs a config file, and each port we bind needs a configuration.  Because we open the proxy port (8080) and one port for each service (30650 and friends), there is a lot of duplication.  I used jsonnet to abstract over that; see etc/generate-envoy-json/envoy.jsonnet.  This is strictly a developer-only compile-time thing; there are tests to ensure that the jsonnet file is run and deposited in etc/helm/pachyderm/envoy.json (also added tests to make sure you ran jsonnetfmt!).   Users can't affect the config file, everyone gets the same one and I plan to keep it that way.  That's why helm isn't used; it's a pain to write Envoy config files, and templating them is even harder, and putting the result in a configmap thru Helm is hardest.  So I just used jsonnet.

There are also a few issues preventing merge right now:
* ~Make sure this is off by default (I think I did that) but on for the tests.~  Done
* ~Ingress should point at this when it's enabled.~ Done
* ~There should be TLS support (for Envoy to terminate TLS, not for mTLS to backend services).~ Tracked in CORE-557.
* ~Updating the deployment tests.  (I want all tests in `master` to go through the proxy, but that makes the config incompatible with older versions of pachyderm.)~
* ~The health checks need some tuning.  Envoy is pretty aggressive about removing dead pachd replicas and serving "no healthy upstream" when you restart pachd; takes about 10 seconds to catch up which is annoying.  This needs some more testing.~ (The problem is not actually health checks, but DNS returning 0 entries for the service until pachd is healthy.  We have a lot of tests that expect pachd to be ready as soon as the Pod is ready, but that's not strictly correct, and going through the proxy exposes this.  As for restarting pachd, that is now handled cleanly; we retry requests that fail to make a TCP connection internally, so users should never see downtime while restarting pachd.  Annoyingly, Envoy's retry system won't let you retry the case where there are 0 backends; this is a bug we should fix upstream.)
* ~There are some "DO NOT MERGE" commits that need to be interactively rebased out; they just change the tests to debug flakiness and weren't that useful.~

A reviewer should make sure they like the new `values.yaml` and the descriptions there, because that will be the documentation for this feature until 2.3 when we enable it by default.

I'm testing this by running Kind such that port 30080 becomes localhost:80, or rather 192.168.1.70:80, and using that for everything.  (This PR does not solve the issue of console and pachd talking to themselves over the external network, but it happens to work totally fine with kind.)

Put this in `kind.yaml`: 
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
    - role: control-plane
      kubeadmConfigPatches:
          - |
              kind: InitConfiguration
              nodeRegistration:
                  kubeletExtraArgs:
                      node-labels: "ingress-ready=true"
      extraPortMappings:
          - containerPort: 30080
            hostPort: 80
            protocol: TCP
          - containerPort: 30443
            hostPort: 443
            protocol: TCP
```

Then make a values file (i just hard code the secrets locally so they don't change with each install):
```yaml

deployTarget: LOCAL
pachd:
    metrics:
        enabled: false
    clusterDeploymentID: dev
    activateEnterprise: true
    enterpriseLicenseKey: "XXX"
    enterpriseSecret: "50PO0HynZjt7oWqWbKnQlLN6j2dl97rq"
    rootToken: "V4Ptmxchcj04TY5vmngJAD0RiuJ3JYc6"
    oauthClientSecret: "p2BDmthvG91dKS22n1YnbvmZW0ZRYW91"
    storage:
        backend: LOCAL
    localhostIssuer: "true"
postgresql:
    postgresqlPassword: "abc123321cba"
    postgresqlPostgresPassword: "npLvsvhGcF"
console:
    enabled: true
    config:
        oauthClientSecret: "Hz1iySrX5Gm8owpRrYMM8vY7NXfgxhUs"
ingress:
    enabled: false
    host: 192.168.1.70
proxy:
    enabled: true
    service:
        type: NodePort
```

Then you can install, like:
```sh
kind delete cluster
kind create cluster --config kind.yaml
kubectx kind-kind
kubectl apply -f pachd-config.yaml
helm install pachyderm ./pachyderm/etc/helm/pachyderm --values values.yaml
```

(pachd-config.yaml is:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
    name: pachd-config
data:
    MODE: full
```
Might not be needed anymore)

Now localhost:80 (or rather ingress.host:80) will serve everything.  You can hit http://192.168.1.70 (etc.) in your browser to see console.   You can configure pachctl like this:
```json
{
    "user_id": "6ccb1f70-c7fa-49c6-906f-002a8c60c514",
    "v2": {
        "active_context": "local",
        "contexts": {
            "local": {
                "pachd_address": "grpc://192.168.1.70:80",
                "session_token": "V4Ptmxchcj04TY5vmngJAD0RiuJ3JYc6",
                "cluster_deployment_id": "dev"
            }
        }
    }
}
```
And use pachctl normally.

Finally, you can configure the Minio client to make this work as an S3 source like:
```json
{
    "version": "10",
    "aliases": {
        "local": {
            "url": "http://localhost",
            "accessKey": "V4Ptmxchcj04TY5vmngJAD0RiuJ3JYc6",
            "secretKey": "V4Ptmxchcj04TY5vmngJAD0RiuJ3JYc6",
            "api": "S3v4",
            "path": "auto"
        }
    }
}
```

Some interesting things to do are look at the pachyderm-proxy access logs `kubectl logs pachyderm-proxy-....` and metrics `kubectl port-forward pachyderm-proxy-... 9901` `curl localhost:9901/stats` etc.


etc/helm/pachyderm/envoy.yaml
```diff
@@ -0,0 +1,225 @@
...
+                use_remote_address: true
+                stat_prefix: pachyderm_envoy_http
+                route_config:
+                  virtual_hosts:
+                    - name: any
+                      domains: ["*"]
+                      routes:
+                        - match:
+                            prefix: /
+                            grpc: {}
```
chainlink, line 63: If only routing grpc was always this easy 😅

chainlink, line 63: If only routing grpc was always this easy 😅

jrockway: Yeah they thought of this use case in advance 😂

src/internal/minikubetestenv/deploy.go
```diff
@@ -223,9 +249,13 @@ func pachClient(t testing.TB, pachAddress *grpcutil.PachdAddress, authUser, name
 		if err != nil {
 			return errors.Wrapf(err, "failed to connect to pachd on port %v", pachAddress.Port)
 		}
+		// Ensure that pachd is really ready to receive requests.
```
acohen4, line 252: why's this check necessary? I thought that if we're able to create (& connect) a new client, we should be good to go

jrockway: Basically the proxy is being too smart.  The timeline right after deployment looks something like:

* pachd pod is ready
* k8s does a health check
* endpoints become ready
* endpoints are added to DNS
* envoy polls DNS to discover the endpoints
* envoy health checks the service
* traffic is ready to be routed

If you send a request between <pachd pod is ready> and <traffic is ready to be routed>, Envoy returns "no healthy upstream" instead of blocking until ready.  This is different than a raw TCP connection through kube-proxy.

Luckily I think disabling the health check changes the behavior to blocking until ready.  That's something I'm trying to avoid, because the message from pachctl when pachd is actually unhealthy is quite good ("no healthy upstream"), but there is some latency on transitioning from unhealthy to healthy (worst case: 2 * (connect timeout + health check interval)) and I need to get rid of that.

There a couple of options that I am testing:
1. Just don't health check.
2. Change the polling interval while unhealthy to something higher (`unhealthy_interval` https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/health_check.proto#config-core-v3-healthcheck)
3. Infer health from responses ("outlier detection": https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier#arch-overview-outlier-detection)

We will see which yields the best results.

etc/generate-envoy-config/envoy.libsonnet
```diff
@@ -0,0 +1,201 @@
...
+          name: 'envoy.access_loggers.stderr',
+          typed_config: {
+            '@type': 'type.googleapis.com/envoy.extensions.access_loggers.stream.v3.StderrAccessLog',
+          },
+        },
+      ],
+      address: {
+        socket_address: {
+          address: '0.0.0.0',
+          port_value: 9901,
```
acohen4, line 15: which port is this?

etc/generate-envoy-config/envoy.jsonnet
```diff
@@ -0,0 +1,196 @@
...
+  'pachd-grpc': {
+    internal_port: 1650,
+    external_port: 30650,
+    grpc: true,
+    service: 'pachd-proxy-backend',
+    routes: [
+      {
+        match: {
+          grpc: {},
+          prefix: '/',
```
acohen4, line 12: so grpc to '/' will go to pachd and '/' without grpc will go to console?

etc/generate-envoy-config/envoy.jsonnet
```diff
@@ -0,0 +1,196 @@
...
+        },
+        route: {
+          cluster: 'pachd-grpc',
+          timeout: '604800s',
```
acohen4, line 16: why this timeout?

etc/generate-envoy-config/envoy.jsonnet
```diff
@@ -0,0 +1,196 @@
...
+        },
+      },
+    ],
+    health_check: {
```
acohen4, line 20: where's the reference to these health check values?

etc/generate-envoy-config/envoy.jsonnet
```diff
@@ -0,0 +1,196 @@
...
+      healthy_threshold: 1,
+      interval: '10s',
+      timeout: '10s',
+      unhealthy_threshold: 2,
+      no_traffic_interval: '10s',
+      no_traffic_healthy_interval: '10s',
+    },
+  },
+  'pachd-s3': {
+    internal_port: 1600,
```
acohen4, line 31: why are both internal ports and external ports required?

etc/generate-envoy-config/envoy.jsonnet
```diff
@@ -0,0 +1,196 @@
...
+      },
+    }; [
+      base {  // S3v4
+        match: {
+          prefix: '/',
+          headers: [
+            {
+              name: 'authorization',
+              string_match: {
+                prefix: 'AWS4-HMAC-SHA256',
```
acohen4, line 48: where is this prefix passed for S3?

etc/generate-envoy-config/envoy.jsonnet
```diff
@@ -0,0 +1,196 @@
...
+    routes: [
+      {
+        match: {
+          prefix: '/',
+        },
+        route: {
+          cluster: 'console',
+          idle_timeout: '600s',
+          timeout: '3600s',
+          upgrade_configs: [
```
acohen4, line 128: how are upgrade_configs used?

etc/generate-envoy-config/envoy.libsonnet
```diff
@@ -0,0 +1,201 @@
...
+        {
+          name: 'envoy.resource_monitors.fixed_heap',
+          typed_config: {
+            '@type': 'type.googleapis.com/envoy.extensions.resource_monitors.fixed_heap.v3.FixedHeapConfig',
+            max_heap_size_bytes: 500000000,
+          },
+        },
+      ],
+    },
+    layered_runtime: {
```
acohen4, line 59: what does this do?

etc/helm/pachyderm/envoy.json
```diff
@@ -0,0 +1,1027 @@
+{
```
acohen4, line 1: what's the process a dev should go through to updating this file when changing the jsonnet?

etc/generate-envoy-config/envoy.jsonnet
```diff
@@ -0,0 +1,196 @@
...
+      routes=std.flatMap(function(name) services[name].routes, ['pachd-grpc', 'pachd-s3', 'pachd-identity', 'pachd-oidc', 'console'])
+    ),
+  ] + [
+    local svc = services[name];
+    Envoy.httpListener(
+      port=svc.internal_port,
+      name='direct-' + name,
+      routes=svc.routes,
+    )
+    for name in ['pachd-grpc', 'pachd-s3', 'pachd-identity', 'pachd-oidc', 'console', 'pachd-metrics']
```
acohen4, line 185: maybe it would be helpful to clarify which services aren't included here, either by comment or `services - [...]`?
WDYT?

etc/helm/pachyderm/templates/proxy/service.yaml
```diff
@@ -0,0 +1,83 @@
...
+  {{- if .Values.proxy.service.annotations }}
+  annotations:
+    {{- toYaml .Values.proxy.service.annotations | nindent 4 }}
+  {{- end }}
+  name: pachyderm-proxy
+  namespace: {{ .Release.Namespace }}
+spec:
+  type: {{ .Values.proxy.service.type }}
+  ports:
+    - name: http-port
```
acohen4, line 20: left another comment, but are we planning on exposing https port at some point?

acohen4 (commented):
Overall LGTM. I did leave a bunch of questions. I'm also reading up on the envoy docs in the meantime so I grok the config better

src/internal/minikubetestenv/deploy.go
jrockway: I talked a little about it in the description, but basically we can't avoid the "unhealthy upstream".  In our case it's not a symptom of not having a passing health check early on, but rather that DNS hasn't been updated (and polled), and thus there are 0 upstream IP addresses to route traffic to.  (The flow is: a pachd pod starts, the pod becomes ready, the k8s endpoint controller updates the endpoint, CoreDNS receives a notification from the k8s apiserver that this happened and makes the pod available in the service [because "headless service", no kube-proxy hop, rather the ready pods are returned as unique A records], and finally Envoy polls DNS and gets the address.)

I added retries inside Envoy for certain types of connection failure (when we're sure that pachd couldn't have already processed the RPCs), but unfortunately this is not a case where Envoy thinks it can retry.  (It doesn't understand that addresses arrive asynchronously and that waiting can solve the problem; someone reported a bug for this, but they worked around it and closed it.  We might want to reopen that.) 

In the end, I think this is largely a problem for our tests and not actual users.  I fixed the tests to wait for a successful RPC instead of pod readiness.  With the retries, things work very well during version upgrades / rolling restarts.  If a pachd replica is dead and Envoy hasn't noticed yet, it will have trouble opening a connection, and that is retryable.  So requests will transparently go to a new replica even when Envoy's view of the system is outdated.

(So how does this work without Envoy?  We use normal kube-proxy services and not headless services, so there is an IP address even when pachd isn't running.  It seems that the way kube-proxy sets up the network, connections will actually be delayed until there are endpoints.  So you don't really even have to wait for pod readiness to send a request, you just need to know the TCP timeout and when pachd will be ready, and any TCP connection in that interval will succeed.  Envoy, on the other hand, just says "nope" and rejects the request.  Unfortunate.

As for "why not just use a normal service and not a headless service", Envoy already introduces an extra network hop, so adding a second hop through kube-proxy seems non-ideal.  We could do that, though.)

etc/generate-envoy-config/envoy.jsonnet
```diff
@@ -0,0 +1,196 @@
...
+          cluster: 'pachd-metrics',
+          idle_timeout: '60s',
+          timeout: '60s',
+        },
+      },
+    ],
+  },
+};
+
+// This is the generation code.
```
acohen4, line 169: how many threads is envoy running?

etc/helm/pachyderm/templates/_helpers.tpl
```diff
@@ -46,9 +46,14 @@ http
 
 {{- define "pachyderm.issuerURI" -}}
 {{- if .Values.oidc.issuerURI -}}
+{{- if and .Values.proxy.enabled (not (hasSuffix "/dex" .Values.oidc.issuerURI)) -}}
+{{ fail (printf "With the proxy enabled, oidc.issuerURI must end with /dex, not %s" .Values.oidc.issuerURI) }}
+{{- end -}}
 {{ .Values.oidc.issuerURI }}
```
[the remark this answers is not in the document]
jrockway: Yeah, this one caught me by surprise.  I thought we mandated "/dex", but it's just the default configuration when Ingress is enabled.  Would like to move in the direction of not making these things configurable; less choices, less confusion.

etc/helm/pachyderm/templates/ingress/ingress-via-proxy.yaml
```diff
@@ -0,0 +1,33 @@
...
+metadata:
+  name: "pachyderm-proxy"
+  annotations: {{ toYaml .Values.ingress.annotations | nindent 4 }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: "pachyderm-proxy"
+    suite: "pachyderm"
+spec:
+  {{- if .Values.ingress.tls.enabled }}
+  tls:
```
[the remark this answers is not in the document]
jrockway: This particular TLS section is just the Ingress definition where the Ingress controller terminates TLS and sends cleartext to Envoy, which sends cleartext to pachd and friends.  I'll add a TLS test when we make the proxy terminate TLS (issue CORE-577).

etc/helm/pachyderm/templates/proxy/deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+        suite: pachyderm
+        {{- if .Values.proxy.labels }}
+        {{- toYaml .Values.proxy.labels | nindent 8 }}
+        {{- end }}
+    spec:
+      affinity:
+        podAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - weight: 100
+              podAffinityTerm:
```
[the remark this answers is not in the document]
jrockway: "Prefer a node that is already running pachd."

In the case where there is 1 pachd and 1 pachyderm-proxy, it's nice if they're on the same node.  If not, then oh well.

etc/helm/pachyderm/templates/proxy/deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+      {{- include "pachyderm.imagePullSecrets" . | indent 6 }}
+      containers:
+        - name: envoy
+          command:
+            - envoy
+          args:
+            - -c
+            - /etc/envoy/envoy.json
+            - --service-node
+            - $(MY_POD_NAME)
```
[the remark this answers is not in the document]
jrockway: We don't really need to know those yet.  If we use other features, the information can help: https://www.envoyproxy.io/docs/envoy/latest/operations/cli.html?highlight=bootstrap%20node#cmdoption-service-cluster

etc/helm/pachyderm/templates/proxy/deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+          image: {{ .Values.proxy.envoyImage }}
+          ports:
+            - name: admin-port
+              containerPort: 9901
+            - name: http-port
+              containerPort: 8080
+            # - name: https
```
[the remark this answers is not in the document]
jrockway: Not really.  I think I will literally be uncommenting these in the next PR, so it's annoying to delete them now ;)

etc/generate-envoy-config/envoy.jsonnet
jrockway: Yup.

etc/generate-envoy-config/envoy.libsonnet
jrockway: Envoy admin interface.  Used for the k8s health check (/server_info = live, /ready = ready).  Also serves /stats/prometheus for prometheus enthusiasts.

etc/generate-envoy-config/envoy.jsonnet
jrockway: We don't use the external port.  I kept them as documentation.

etc/generate-envoy-config/envoy.jsonnet
jrockway: S3 V4 sends authorization headers like: 

```
Authorization: AWS4-HMAC-SHA256 Credential=V4Ptmxchcj04TY5vmngJAD0RiuJ3JYc6/20220517/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
```

S3 V2 sends authorization headers like:

```
Authorization: AWS **REDACTED**:**REDACTED**
```

Those are the only protocols that "S2" (our in-house S3 clone) supports.

Obviously this is somewhat flawed in that if you wanted to talk to console and stuck "Authorization: AWS is not in use" in the HTTP request, it wouldn't be routed correctly.

etc/generate-envoy-config/envoy.jsonnet
jrockway: This is for websockets, which is what console's frontend uses to talk to it.  (Streaming updates as the server state changes.)

HTTP upgrades are insanely complicated: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/upgrades

The docs imply that upgrades being enabled is the default, but not upgrades that support websockets (only HTTP CONNECT).

etc/generate-envoy-config/envoy.libsonnet
jrockway: Basically you get a warning in the logs if you don't have this.  "Runtime" is their buzzword for "configuration".  The runtime has layers that can be read from this file, other files on disk (like a configmap, and it will update without having to restart envoy and drain traffic), or a service discovery server.  We are just using it to set a limit on connections to something less than infinity.

etc/helm/pachyderm/envoy.json
jrockway: "make" in this directory.  If you forget, CI will remind you.

(So... why not put a comment in this file saying that it's autogenerated.  Why not use YAML instead of JSON?  jsonnet didn't have a way to pass a comment at the very top level through 😢.)

etc/generate-envoy-config/envoy.jsonnet
jrockway: I pushed a revision that explains why the multiplexed route table has services listed in a particular order (the priority order in the route table is important).  I made this statement use all the services, added a comment, and made sure that the order doesn't depend on hashtable order.  (It doesn't matter what order the direct listeners show up in, but it will make CI claim that you forgot to regenerate the file if the map key iteration order changes.)

etc/helm/pachyderm/templates/proxy/service.yaml
jrockway: Yup, CORE-577.

etc/generate-envoy-config/envoy.jsonnet
jrockway: I don't make any effort to set it.  Envoy is largely event-based: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/intro/threading_model

etc/generate-envoy-config/envoy.jsonnet
acohen4: gotcha makes sense. Could you leave a comment about that somewhere? I feel like that would confuse me if I stumbled across it otherwise

src/internal/minikubetestenv/deploy.go
```diff
@@ -322,6 +364,9 @@ func putRelease(t testing.TB, ctx context.Context, namespace string, kubeClient
 	if opts.Loki {
 		helmOpts = union(helmOpts, withLokiOptions(namespace, int(pachAddress.Port)))
 	}
+	if !(opts.Version == "" || strings.HasPrefix(opts.Version, "2.3")) {
```
acohen4, line 367: should we add a TODO to remove this once 2.4 hits?

src/internal/minikubetestenv/deploy.go
acohen4: Ok thanks for the thorough description. I think the key detail I was missing was the "why does it work now" comparison

src/internal/minikubetestenv/deploy.go
jrockway: For the time being I opened CORE-654 to remind us.

etc/helm/pachyderm/values.yaml
```diff
@@ -637,3 +638,60 @@ oidc:
...
+  # to pachd or Console.
+  enabled: false
+  # The number of proxy replicas to run.  1 should be fine, but if you want more for higher
+  # availability, that's perfectly reasonable.  Each replica can handle 50,000 concurrent
+  # connections.  There is an affinity rule to prefer scheduling the proxy pods on the same node as
+  # pachd, so a number here that matches the number of pachd replicas is a fine configuration.
+  # (Note that we don't guarantee to keep the proxy<->pachd traffic on-node or even in-region.)
+  replicas: 1
+  # The envoy image to pull.
+  envoyImage: envoyproxy/envoy:v1.22.0
```
acohen4, line 656: I think this should be image.repository, image.tag, image.pullPolicy like we do in other places

acohen4 (changes requested):

etc/helm/pachyderm/values.yaml
jrockway: Will update.

acohen4: also remove the `envoyImage`

jrockway: doh!

jrockway: that is fixed, and now all the tests pass!

acohen4 (approved):
LGTM!!

etc/generate-envoy-config/envoy.jsonnet
jrockway: Basically if there's more than one pachd, and one of them isn't working, we can avoid sending traffic to that one.  Otherwise, if we look at the error logs we can see things like "no healthy upstream" and know that, at the time, pachd wasn't being responsive.  (Will be nice in debug dumps, for example.)

PR #7529, opened 2022-04-21
move ListPipelineInfo to ppsutil to reduce PPS master dependency on APIServer

acohen4:


albscui (approved):
LGTM! My nit is that the name `ListPipelineInfo` doesn't tell the caller about all the other stuff it's doing, including `checkPipelineVersion` and applying `f`. But this is a strict refactor, and we can try to improve the naming later.

PR #7531, opened 2022-04-21
Batch copy operations in unordered writer. (#7436)

PFedak:
Previously, copy operations during a modify file stream forced a new
underlying file set to be created, which made operations with many
separate copies (e.g. shuffle pipelines) unnecessarily slow. The
unordered writer only really needs to serialize its pending writes when
it hits the memory limit, and copies have the advantage that they can be
recorded without eating into that budget.

This change records per-file copy operations in the unordered writer, 
deferring the actual data reference writes to serialization time. It also 
sets a file count threshold for serialization, enforcing a lower bound on 
the size of file sets produced from copies.

### DIFF
--- a/src/internal/storage/fileset/buffer.go
+++ b/src/internal/storage/fileset/buffer.go
@@ -8,14 +8,22 @@ import (
 )
 
 type Buffer struct {
-	additive map[string]map[string]*file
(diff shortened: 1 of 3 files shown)


brycemcanally (approved):
lgtm

PR #7533, opened 2022-04-21
Check for validation errors when choosing a job's base commit. (#7519)

PFedak:
Previously the base meta commit for a job was the most recent non-errored ancestor. As validation errors only show up in the output commit, we could end up picking a base commit that the job code would later reject, leading to an infinite loop. We might later want to enforce that output and meta commits have the same error status once finished, but this change will avoid the loop regardless.


brycemcanally (approved):
lgtm

PR #7538, opened 2022-04-25
Implement compaction cache

brycemcanally:
This PR implements caching for compaction tasks (similar to the caching model we use for job processing tasks). Also, the validation step when finishing a commit has been converted to a task that runs through the task service. This will take more load off of the PFS master and make the validation step eligible for caching.


PFedak (approved):
LGTM

PR #7539, opened 2022-04-25
Eagerly delete task service group entries

brycemcanally:
This PR makes it such that we eagerly delete group entries created through the task service. Currently, we rely on lease expiration to delete the group entries, which is correct, but can result with an unnecessary delay when canceling work in the case that the Doer hits an error. This can be impactful to how quickly work resumes in the retry case (i.e. compaction).


PFedak (approved):
LGTM

PR #7544, opened 2022-04-26
Make postgres watches respect contexts.

PFedak:
Postgres watches could block indefinitely during the initial list backfill, as nothing connected the context to the watcher's `done` channel.


acohen4 (approved):
LGTM. How'd you notice this bug?

acohen4:
> LGTM. How'd you notice this bug?

NVM, I see the support issue now and just found this in unit tests too

PR #7545, opened 2022-04-26
Migrate Admin Tests

acohen4:


robert-uhl:
LGTM! I think the usage of `t.Parallel()` is good, given that the tests use different clusters & namespaces.

robert-uhl (approved):

PR #7546, opened 2022-04-26
Migrate Auth tests to Client Factory

acohen4:


robert-uhl (approved):
Looks good — appears to be a pretty straightforward conversion.

PR #7549, opened 2022-04-26
[CORE-295] Check repo existence before put file

robert-uhl:
Previously, `pachctl put file` would upload the file in question only to find that the indicated repo didn’t exist; this change checks for the existence of the repo first.

### DIFF
--- a/src/server/pfs/cmds/cmds.go
+++ b/src/server/pfs/cmds/cmds.go
@@ -1035,6 +1035,7 @@ $ {{alias}} repo@branch -i http://host/path`,
 			if err != nil {
 				return err
 			}
+
 			opts := []client.Option{client.WithMaxConcurrentStreams(parallelism)}
 			if compress {
 				opts = append(opts, client.WithGZIPCompression())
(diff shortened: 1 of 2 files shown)


src/server/pfs/cmds/cmds.go
```diff
@@ -1045,6 +1046,13 @@ $ {{alias}} repo@branch -i http://host/path`,
 			}
 			defer c.Close()
 			defer progress.Wait()
+			// check whether or not the repo exists before attempting to upload
+			if _, err = c.InspectRepo(file.Commit.Branch.Repo.Name); err != nil {
+				if errutil.IsNotFoundError(err) {
+					return errors.Errorf("repo %s not found", file.Commit.Branch.Repo.Name)
```
PFedak, line 1052: you can probably just return `err` as is here

src/server/pfs/cmds/cmds.go
```diff
@@ -1045,6 +1046,13 @@ $ {{alias}} repo@branch -i http://host/path`,
...
+				}
+				return errors.Errorf("could not inspect repo %s: %w", file.Commit.Branch.Repo.Name, err)
```
PFedak, line 1054: I think we're preferring using `errors.Wrapf` now

src/server/pfs/cmds/cmds_test.go
```diff
@@ -79,6 +80,24 @@ func TestPutFileSplit(t *testing.T) {
...
+                (dd if=/dev/urandom bs=1024 count=$((1024 * 1024)) 2>/dev/null | pachctl put file {{.repo}}@master:random 2>&1 || true)\
+                  | match "repo {{.repo}} not found"
+`,
+		"repo", repoName).Run())
+	// experimentally, the repo check should take less than three quarters
+	// of a second, while uploading 1GB of data takes well over two seconds
+	if time.Since(start) > 2*time.Second {
+		t.Error("pachctl put file appeared not to check if repo existed before uploading")
+	}
+}
```
PFedak, line 99: Maybe a more straightforward way to test this would be to use `pachctl put file -f does-not-exist, then make sure we get an error about the repo rather than the missing file?

PFedak (approved):
Thanks for the fix! I'd rather not rely on timing in the test, so hopefully the alternative I suggested or something along those lines suffices

src/server/pfs/cmds/cmds_test.go
robert-uhl: That _is_ dependent on the ordering of the two checks (repo versus file), but I like that it doesn’t rely on timing. I’m changing the test, and adding a comment so if the ordering changes the developer working on it has a hint about why the test failed.

PR #7550, opened 2022-04-27
Collect all minikubetestenv logs, avoid minikubetestenv port conflicts

jrockway:
This is two unrelated changes ;)

kube_debug.sh: Dump all pachyderm logs, not just pachd.  Also do this for the default namespace and all 6 possible minikubetestenv namespaces.  Very helpful for debugging stuff :)

kubernetes-prometheus/all-manifests.yaml: This allocates a bunch of NodePorts with no `nodePort` specified.  Sometimes it picks a NodePort in the range 30660-31020, which are used by Pachyderm when started with minikubetestenv.  This results in the tests failing (at max timeout!) when this happens.  The change just hard-codes port numbers that don't conflict with anything.

### DIFF
--- a/etc/kubernetes-prometheus/manifests-all.yaml
+++ b/etc/kubernetes-prometheus/manifests-all.yaml
@@ -392,6 +392,7 @@ spec:
   ports:
   - name: alertmanager
     protocol: TCP
+    nodePort: 30200
     port: 9093
     targetPort: 9093
 ---
@@ -2420,6 +2421,7 @@ spec:
   type: NodePort
   ports:
     - port: 3000
+      nodePort: 32001
   selector:
     app: grafana
     component: core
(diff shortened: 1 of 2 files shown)


acohen4 (approved):
LGTM!! Where does the prom manifest get applied?

jrockway:
> LGTM!! Where does the prom manifest get applied?

In the Makefile for certain tests:

https://github.com/pachyderm/pachyderm/blob/master/Makefile#L333 is the rule, and it gets used sneakily by tests like this: https://github.com/pachyderm/pachyderm/blob/master/Makefile#L315

I don't think we run the tests that need this, though: 

https://github.com/pachyderm/pachyderm/blob/master/src/server/worker/stats/stats_test.go

I see we test worker/ but not worker/..., so they probably don't run.  (And I have a feeling I broke them several months ago when restructuring prometheus metrics 😱)

PR #7558, opened 2022-04-29
[CORE-463] Make inputs failed message clearer

robert-uhl:
This change differentiates between jobs which fail while running & jobs which never run at all; it does this by adding a new job state “not run.” End users will need to update to a newer version of `pachctl` in order to work with jobs in a not-run state.

The not-run state cascades from one not-run job to its child jobs. There is a descriptive reason for each not-run job indicating why it was not run (this is an expansion of the old “inputs failed” message).

The not-run state was previously termed “cancelled,” hence some of the current commit titles.


src/pps/pps.proto
```diff
@@ -69,6 +69,7 @@ enum JobState {
   JOB_KILLED = 6;
   JOB_EGRESSING = 7;
   JOB_FINISHING = 8;
+  JOB_NOT_RUN = 9;
```
brycemcanally, line 72: It seems kind of weird to have a state with a negative in it. @jdoliner what are your thoughts on this? Canceled seems fine to me since it was canceled internally.

src/server/worker/pipeline/transform/registry.go
```diff
@@ -256,8 +266,8 @@ func (reg *registry) processJobStarting(pj *pendingJob) error {
 		return err
 	}
 	if len(failed) > 0 {
-		reason := fmt.Sprintf("inputs failed: %s", strings.Join(failed, ", "))
-		return reg.failJob(pj, reason)
+		reason := fmt.Sprintf("Cancelled because the following upstream pipelines failed: %s", strings.Join(failed, ", "))
```
brycemcanally, line 269: This just gives us the immediate upstream pipelines rather than the full path of not run / canceled jobs. This may be fine considering the job set will have the not run / canceled state for all of the jobs in the downstream path, but I just want to make sure this meets the requirements.

src/server/pps/cmds/cmds_test.go
```diff
@@ -123,6 +123,62 @@ func TestRawFullPipelineInfo(t *testing.T) {
 		`).Run())
 }
 
+func TestFailedPipelineInfo(t *testing.T) {
```
brycemcanally, line 126: It seems like this test should be geared more towards jobs than pipelines. We are testing a failed job not a failed pipeline. Also, it seems like this is testing the not run / canceled state specifically since that is the only thing being checked, so it probably makes sense to name the test according to that.

src/server/pachyderm_test.go
```diff
@@ -1191,6 +1191,101 @@ func TestPipelineFailure(t *testing.T) {
...
 }
 
+func TestInputFailure(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration tests in short mode")
+	}
+	t.Parallel()
+	c, _ := minikubetestenv.AcquireCluster(t)
+
+	dataRepo := tu.UniqueString("TestPipelineFailure_data")
```
brycemcanally, line 1201: Probably should update the name.

src/pps/pps.proto
robert-uhl: Back to ‘cancelled,’ then? Or ‘aborted’? ‘Not run’ was the best we came up with at the design review. Some other suggestions:

 - never run
 - cannot run
 - unrunnable (too opaque?)
 - inputs failed (that was the old error message; maybe it makes sense as the state)
 - input failure
 - precondition failed (yikes that is verbose!)

brycemcanally: Not run doesn't tell you whether the system or user caused it not to run and it isn't as obviously terminal as canceled is. Also, I'm not sure why we need to emphasize that the job didn't go into the running state. That information alone doesn't tell you what happened, they would need to look at the details regardless. If we want it to be really clear from the state, then inputs failed seems like the right choice, but my same concerns in terms of the longevity of this model still apply (having a state for every terminal scenario seems like a bad design).

src/server/worker/pipeline/transform/registry.go
robert-uhl: Confirmed with @dgeorg42 that this suffices for the time being.

src/pps/pps.proto
jdoliner: I generally agree that having a failure state for every terminal scenario is a bad design. However I think it's more reasonable to have different terminal states that reflect the different states the job's filesystems can be in due to when the job terminated. We already have `FAILURE` and `KILLED` those reflect a difference in why the job is in a terminal state. But they also reflect a difference in the possible states of the fs for the job. You have some guarantees about what the metacommit will look like if it's in `FAILURE` (I believe) whereas in `KILLED` there aren't any since the user can kill the job at any point during the execution. The new state is a guarantee that nothing will be present in the meta commit, that to me is the significance of the fact that it never went into the running state. `NOT_RUN` reflects that to me, although I can see how it might lead to the question of why didn't pachyderm run my job. Of course inspecting the job should answer that question, but that's been true so we're trying to make it more clear from the state. I kind of like `UNRUNNABLE` it suggests to me that it's a property of the job that prevents it from ever running, not a decision the user or system made. But it has the downside that someone could conceivably think the state is transitory and it might become runnable. So I'm not sure if that's better. Similarly `UNSTARTABLE` seems like it gets the idea across. I actually think that the most descriptive option might be `DEAD_ON_ARRIVAL` I think it's a little too whimsical unfortunately but it has exactly the right connotation. Does this reasoning make sense to you @brycemcanally and do you think `UNRUNNABLE` expresses the right things? I agree it is a little weird to have the word "not" in a state.

brycemcanally: Inputs failed would reflect the same thing about the meta commit and would be obviously terminal. Tbh, I don't feel too strongly about what we go with for now because I don't think any of the short term options are good. We can go with not run if that makes sense to you guys, just make sure it is consistent in the code (I see some places that still reference canceled).

robert-uhl: I kinda like that `UNRUNNABLE` — while it is a negative — does not have ‘not’ in it. I think that the sense of it might be even better than `NOT_RUN` or `CANCELLED`, in that the job is not runnable (because its inputs are not available); that’s more apt than just saying that the job wasn’t run, or that it was cancelled.

jdoliner: K, let's commit to `UNRUNNABLE`

src/server/pps/cmds/cmds_test.go
```diff
@@ -128,6 +128,66 @@ func TestRawFullPipelineInfo(t *testing.T) {
 		`).Run())
 }
 
+func TestNotRunJobInfo(t *testing.T) {
```
brycemcanally, line 131: nit, `NotRun` -> `Unrunnable`

brycemcanally (approved):
lgtm, just one minor change.

PR #7559, opened 2022-04-29
Avoid using 31800 for all nodeports; give each test a unique port

jrockway:
I think it's possible for 3 tests that each request port 31800 as a NodePort to run concurrently, so this gives each of them a unique port number.  Then they can run in the same cluster concurrently.

It's not clear to me how much flakiness this has failed; I have seen two of the tests run concurrently and one failed, but it might have actually failed to some brokenness with minikube and not due to this issue.  But, one less thing to worry about.


acohen4 (approved):
LGTM! Good catch. I think we could also add a PortAllocating endpoint in minikubetestenv, to service these sort of tests

PR #7572, opened 2022-05-02
CORE-558 rename "k8s_secret" to "name" in Egress secret

albscui:


src/server/pfs/pfs.go
```diff
@@ -429,7 +429,7 @@ func IsDropWithChildrenErr(err error) bool {
 func ValidateSQLDatabaseEgress(sql *pfs.SQLDatabaseEgress) error {
 	if sql != nil {
 		if secret := sql.GetSecret(); secret != nil {
-			if secret.K8SSecret != "" && secret.Key != "" {
+			if secret.Name != "" {
 				return nil
 			}
 			return errors.Errorf("egress.sql_database.secret.k8s_secret and egress.sql_database.secret.key are required")
```
jrockway, line 435: The field name in the errors.Errorf needs to change.

src/server/pfs/pfs.go
```diff
@@ -429,7 +429,7 @@ func IsDropWithChildrenErr(err error) bool {
 func ValidateSQLDatabaseEgress(sql *pfs.SQLDatabaseEgress) error {
 	if sql != nil {
 		if secret := sql.GetSecret(); secret != nil {
-			if secret.K8SSecret != "" && secret.Key != "" {
+			if secret.Name != "" {
```
jrockway, line 432: If `Key` is empty, what are we going to use as the name of the key?  I guess the empty string is a string, but if you try to create a secret with a key that is the empty string, you get:

```
# secrets "foo" was not valid:
# * data[]: Invalid value: "": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')
```

src/server/pfs/pfs.go
```diff
@@ -429,7 +429,7 @@ func IsDropWithChildrenErr(err error) bool {
 func ValidateSQLDatabaseEgress(sql *pfs.SQLDatabaseEgress) error {
 	if sql != nil {
```
jrockway, line 430: Outside the scope this PR, but I'd prefer returning to indenting, i.e.

```go
if sql == nil {
    return nil
}
secret := sql.GetSecret()
if secret == nil {
   return errors.Errorf("egress.sql_database.secret is required")
}
// ...
```

This is not quite the rule: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow but "Try to keep the normal code path at a minimal indentation" is good advice.

jrockway (approved):
LGTM.  One error message still refers to `k8s_secret` and needs to be changed.

I did check out the branch and did `rg -S 'k8s.?secret' -g '*.go'` to see if there are any other cases like this, but didn't find any.

src/server/pfs/pfs.go
albscui: Oh oops didn't mean to remove the check on the key.

src/server/pfs/pfs.go
albscui: Agreed, this way is cleaner. I normally structure my code like this when dealing with errors, but in this case it's kind of the opposite where a nil value results in error.

PR #7577, opened 2022-05-03
Migrate Cmd Tests to Client Factory

acohen4:


src/internal/config/config.go
```diff
@@ -270,20 +269,24 @@ func (c *Config) write() error {
...
-		if err = ioutil.WriteFile(p, rawConfig, 0644); err != nil {
-			return errors.Wrapf(err, "failed to copy updated config file from %s to %s", tmpfile.Name(), p)
+		if err = ioutil.WriteFile(path, rawConfig, 0644); err != nil {
+			return errors.Wrapf(err, "failed to copy updated config file from %s to %s", tmpfile.Name(), path)
 		}
 	}
 
 	// essentially short-cuts reading the new config back from disk
-	cachedConfig = proto.Clone(c).(*Config)
+	// Note: this will only fail in the case that the caller passes
```
albscui, line 285: Just to clarify, the failure is due to when the caller passes in some random path that is not consistent with `configPath()`? Should we error in this case?

albscui (approved):
LGTM!

src/internal/config/config.go
acohen4: maybe fail is the wrong word in the comment. The intention behind these changes is to expand the PachConfig API so that we can modify configs at arbitrary paths from go code. The benefit here is for testing. Basically we don't care to cache the config globally in the case that we are passing an arbitrary pach_config path from the API

acohen4: maybe fail is the wrong word in the comment. The intention behind these changes is to expand the PachConfig API so that we can modify configs at arbitrary paths from go code. The benefit here is for testing. Basically we don't care to cache the config globally in the case that we are passing an arbitrary pach_config path from the API

PR #7581, opened 2022-05-03
[CORE-476] Show total input datums rather than running total

robert-uhl:
No test added because it would be tricky to test, due to the fact that it would have to rely on timing.


src/server/worker/pipeline/transform/pending_job.go
```diff
@@ -160,12 +159,30 @@ func (pj *pendingJob) withDeleter(pachClient *client.APIClient, cb func(datum.De
 	})
 }
 
+// writeDatumCount gets the total count of input datums for the job and updates
+// its DataTotal field.
+func (pj *pendingJob) writeDatumCount(ctx context.Context, taskDoer task.Doer) error {
+	pachClient := pj.driver.PachClient().WithCtx(ctx)
+	return pachClient.WithRenewer(func(ctx context.Context, renewer *renew.StringSet) error {
```
brycemcanally, line 166: nit, the scope of the renewer could be reduced to just the create full job datum file set.

brycemcanally (approved):
lgtm, just a minor nit. Also, you could add a test to specifically check that the total is available before any datums are processed by having a longer running datum and checking a few seconds after kicking off the job, but I'm not sure if we want more timing based tests like that.

PR #7586, opened 2022-05-04
Today's random CI fixes

jrockway:
* Update dependencies to newer version; latest k8s and machine image
* Updating the machine image broke the gcloud installer script, but fortunately the machine already has a newer version of gcloud installed anyway, so I deleted the orb-based install step.
* Fix a few persistently flaky tests
* I made the container build and starting of minikube happen in parallel; saves about 45-60 seconds.


acohen4 (approved):
LGTM

PR #7587, opened 2022-05-04
[CORE-520] Capture docker image SHA for datums

armaanv:


src/server/pps/server/kube_driver.go
```diff
@@ -173,3 +173,15 @@ func (kd *kubeDriver) WatchPipelinePods(ctx context.Context) (<-chan watch.Event
...
+func (kd *kubeDriver) GetWorkerImageID(ctx context.Context, podName string) (string, error) {
+	pod, err := kd.kubeClient.CoreV1().Pods(kd.namespace).Get(
+		ctx,
+		podName,
+		metav1.GetOptions{})
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get kubernetes pod")
+	}
+
+	return pod.Status.ContainerStatuses[1].ImageID, nil // second container should be the pipeline worker, first should be pachd
```
brycemcanally, line 186: We should probably iterate through the containers and check the names to be sure we got the right one. We could probably also make this function more general by parameterizing the container name.

src/server/worker/pipeline/transform/status.go
```diff
@@ -19,6 +19,7 @@ type Status struct {
 	jobID       string
 	datumStatus *pps.DatumStatus
 	cancel      func()
+	ImageID     string
```
brycemcanally, line 22: I don't think we should use the status to pipe this information through, we should probably just pipe it through as another parameter in the functions for now. The image id isn't needed by the status functionality.

src/server/worker/pipeline/transform/worker.go
```diff
@@ -377,6 +377,8 @@ func handleDatumSet(driver driver.Driver, logger logs.TaggedLogger, datumSet *Da
 								return errors.EnsureStack(driver.RunUserErrorHandlingCode(runCtx, logger, env))
 							}))
 						}
+						// maybe proto/clone meta here
```
brycemcanally, line 380: We should probably go ahead and do the clone. We could add `meta = proto.Clone(meta).(*datum.Meta)` as the first line in the callback.

src/server/pps/server/kube_driver.go
```diff
@@ -173,3 +173,15 @@ func (kd *kubeDriver) WatchPipelinePods(ctx context.Context) (<-chan watch.Event
 	}
 	return kubePipelineWatch.ResultChan(), kubePipelineWatch.Stop, nil
 }
+
+func (kd *kubeDriver) GetWorkerImageID(ctx context.Context, podName string) (string, error) {
```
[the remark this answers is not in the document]
acohen4: I agree - the kubeDriver is intended as a layer of abstraction for the PPS Master process (in pachd) that manages the allocating pipeline resources. I think just passing a KubeClient to the Worker & calling a utility function on it as Bryce suggested makes more sense

etc/helm/pachyderm/templates/pachd/rbac/worker-role.yaml
```diff
@@ -16,6 +16,7 @@ rules:
   - ""
   resources:
   - services
+  - pods
```
acohen4, line 19: this is something I'm just noticing because we're adding a resource here, but I don't think the worker should be able to modify any k8s resources. I can open another PR removing the modify verbs here

src/server/worker/worker.go
```diff
@@ -73,21 +74,25 @@ func NewWorker(
...
 
-func (w *Worker) worker() {
+func (w *Worker) worker(env serviceenv.ServiceEnv) {
 	ctx := w.driver.PachClient().Ctx()
 	logger := logs.NewStatlessLogger(w.driver.PipelineInfo())
 
 	backoff.RetryUntilCancel(ctx, func() error {
 		eg, ctx := errgroup.WithContext(ctx)
 		driver := w.driver.WithContext(ctx)
+		imageID, err := getUserContainerImageID(ctx, env)
```
acohen4, line 88: since the Image ID is static for the lifetime of the Worker process, I think it might make sense to load and reference it through the driver

armaanv: This makes sense. I've moved the function into the driver and made it more generic for container image IDs

etc/helm/pachyderm/templates/pachd/rbac/worker-role.yaml
acohen4: yep makes sense!

src/server/worker/pipeline/transform/transform_test.go
```diff
@@ -110,7 +110,7 @@ func newWorkerSpawnerPair(t *testing.T, dbConfig serviceenv.ConfigOption, pipeli
 
 	eg.Go(func() error {
 		err := backoff.RetryUntilCancel(env.driver.PachClient().Ctx(), func() error {
-			return Worker(env.driver.PachClient().Ctx(), env.driver, env.logger, &Status{})
+			return Worker(env.driver.PachClient().Ctx(), env.driver, env.logger, &Status{}, "")
```
acohen4, line 113: this call should use the original Worker() signature

armaanv: good catch, fixed it in the latest commit

acohen4 (approved):
LGTM

PR #7588, opened 2022-05-04
Adds a churn example.

jdoliner:


examples/churn/churn.json
```diff
@@ -0,0 +1,36 @@
...
+        "cpu": 1
+    },
+    "egress": {
+        "sql_database": {
+            "url": "snowflake://albertpachyderm@CDKSVXM-PN23804/CHURN_DEMO/PUBLIC?warehouse=COMPUTE_WH",
+            "file_format": {
+                "type": "CSV"
+            },
+            "secret": {
+                "k8s_secret": "snowflakesecret",
```
albscui, line 31:
```suggestion
                "name": "snowflakesecret",
```

albscui (approved):
LGTM, except we renamed `k8s_secret` to just `name`.

albscui:
Do we have to upload `examples/churn/data.sql`? Seems like quite a large file. Is it possible to make it smaller?

jdoliner:
> Do we have to upload `examples/churn/data.sql`? Seems like quite a large file. Is it possible to make it smaller?

I could slim it down a bit and the example would still work.

PR #7589, opened 2022-05-04
Fix a flaky etcd collection test, improve etcd logging

jrockway:
This PR attempts to fix the VERY flaky `InterruptionDuringBackfill` test.  I'm not 100% sure this is the correct fix.  What I think is happening here is that we're writing to a channel and canceling the context, and expecting to get the cancellation and not any of the writes.  But the writes happen before the cancellation (depending on how fast the computer is today), so I have no idea how that could work.  As a result, I've adjusted the test to drain the channel looking for the cancellation, and confirming that the cancellation is the last event.

We also had some code to try to integrate etcd's logger with the tests, and filter out noisy startup messages.  At some point etcd dropped support for capnslog, which is the logger we're expecting to adjust to make that happen.  They use Zap now, so I've changed the tests to use Zap.  This has two benefits; the startup messages are now filtered, and the tests log to the per-test logger.  This lets you see exactly which test the etcd messages are from.  (The client also logs stuff, and I've made those logs go to the same place.  They are tagged with "etcd-client" so you can distinguish between them.)

### DIFF
--- a/src/internal/collection/watch_test.go
+++ b/src/internal/collection/watch_test.go
@@ -252,7 +252,8 @@ func watchTests(
 			watcher := makeWatcher(ctx, t, reader)
 			tester := NewWatchTester(t, writer, watcher)
 			cancel()
-			for i := 0; i < 10; i++ {
+			var canceled bool
+			for i := 0; i < 11; i++ {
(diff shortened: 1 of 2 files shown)


acohen4 (approved):
LGTM

PR #7591, opened 2022-05-04
Deploy Loki  in test clusters

acohen4:


src/server/pachyderm_test.go
```diff
@@ -9325,41 +9324,48 @@ func TestDebug(t *testing.T) {
...
-			if g.Match(hdr.Name) {
-				delete(expectedFiles, pattern)
-				break
+		if len(expectedFiles) > 0 {
+			t.Logf("got files: %v", gotFiles)
+			var names []string
+			for n := range expectedFiles {
+				names = append(names, n)
 			}
+			t.Logf("no files match: %v", names)
```
jrockway, line 9362: We can get rid of this `t.Logf` now that the error condition is "hasn't produced: %v" (couple lines down) instead of "6 != 0" ;)

jrockway (approved):
Looks good!

PR #7592, opened 2022-05-04
CORE-542 Handle SQL numeric types with precision and scale gracefully

albscui:
SQL `numeric` types can have a `precision` and `scale`. The `precision` represents the count of significant digits in the number; the `scale` represents the count of digits to the right of the decimal point. Currently, our system doesn't handle numeric types correctly. This PR addresses this gap, by consuming numeric data as strings.

The end to end flow of data would look something like:

```
numeric in SQL -> string -> CSV/JSON file -> string -> numeric SQL
```

What does this mean for CSV? Nothing, since everything in CSV are strings.
What does this mean for JSON? Instead of storing the numeric as a number in JSON, we are just going to store it as string.

Note that there seems to be a bug in Snowflake's numeric to string conversion. Created an issue  https://github.com/snowflakedb/gosnowflake/issues/578


src/internal/pachsql/table_info.go
```diff
@@ -82,11 +86,20 @@ func GetTableInfoTx(tx *Tx, tablePath string) (*TableInfo, error) {
...
+
+	var cinfos []ColumnInfo
+	var precision, scale sql.NullInt64
 	for rows.Next() {
 		var ci ColumnInfo
-		if err := rows.Scan(&ci.Name, &ci.DataType, &ci.IsNullable); err != nil {
+		if err := rows.Scan(&ci.Name, &ci.DataType, &ci.IsNullable, &precision, &scale); err != nil {
 			return nil, errors.EnsureStack(err)
 		}
+		if precision.Valid {
```
jrockway, line 97: Do you even need the `Valid` check here?  If it's false, then the `Int64` is 0, which is what `ci.Precision ` would be if you hadn't assigned to it.

src/internal/sdata/sdata.go
```diff
@@ -113,61 +114,64 @@ func NewTupleFromTableInfo(info *pachsql.TableInfo) (Tuple, error) {
...
+	// FIXME: for now, use either int64 or float64 for arbitrary precision numbers,
+	// but this is not enough for numbers outside their range.
+	case "NUMERIC", "NUMBER", "DECIMAL", "FIXED":
+		if nullable {
+			if scale == 0 && precision > 0 {
+				return new(sql.NullInt64), nil
+			}
+			return new(sql.NullFloat64), nil
+		}
+		if scale == 0 && precision > 0 {
```
jrockway, line 156: I'm not totally convinced that NUMERIC without a decimal point is representable as an integer.  Consider 2^65:

```go
func main() {
        x := math.Pow(2, 65)
        fmt.Printf("%v %v\n", x, int64(x))
}
```
```
$ go run main.go
3.6893488147419103e+19 -9223372036854775808
```

I think both cases have to be `big.Rat` objects.  (I would also not try to round according to precision, let the database handle that.)

src/internal/sdata/sdata.go
```diff
@@ -113,61 +114,64 @@ func NewTupleFromTableInfo(info *pachsql.TableInfo) (Tuple, error) {
...
-	// TODO account for precision and scale as well
-	case "FLOAT", "FLOAT8", "REAL", "DOUBLE PRECISION", "FIXED":
+		return new(float64), nil
+	case "FLOAT", "FLOAT8", "REAL", "DOUBLE PRECISION":
 		if nullable {
 			return new(sql.NullFloat64), nil
-		} else {
-			return new(float64), nil
 		}
+		return new(float64), nil
```
jrockway, line 164: This one I'm less sure about.  Does every float32 map to exactly one float64, and are the string representation (for JSON/CSV) equal?  I don't know the answer to that, but it would be a hard to detect way to lose/corrupt data.

src/internal/sdata/sdata.go
```diff
@@ -113,61 +114,64 @@ func NewTupleFromTableInfo(info *pachsql.TableInfo) (Tuple, error) {
...
 		}
-	case "DATE", "TIMESTAMP", "TIMESTAMP_NTZ", "TIMESTAMP WITHOUT TIME ZONE":
+		return new(string), nil
+	case "DATE", "TIME", "TIMESTAMP", "TIMESTAMP_LTZ", "TIMESTAMP_NTZ", "TIMESTAMP_TZ", "TIMESTAMP WITH TIME ZONE", "TIMESTAMP WITHOUT TIME ZONE":
 		if nullable {
 			return new(sql.NullTime), nil
-		} else {
-			return new(time.Time), nil
 		}
+		return new(time.Time), nil
```
jrockway, line 174: This is another case I don't know how to handle.  Is 2022-05-04T7:33Z::"timestamp without time zone" == to 2022-04-04T6:33Z+01::"timestamp without time zone"?  Do users even care?  (I just worry about serializing a time zone into the CSV or whatever when there wasn't one in the database.)

jrockway (changes requested):
I'm really paranoid about throwing away information from the database by casting rationals into floats and I think we should try and avoid that.  I can be talked out of it, but I think we should give big.Rat a try before falling back to float conversions.  (Floating point math is a nice optimization to make scientific computation fast, but if it's stuff like taxes then someone goes to jail over the precision loss.  Pachyderm should support both ;)

src/internal/sdata/sdata.go
```diff
@@ -113,61 +114,64 @@ func NewTupleFromTableInfo(info *pachsql.TableInfo) (Tuple, error) {
...
+		return new(int32), nil
+	case "BIGINT", "INT8":
 		if nullable {
 			return new(sql.NullInt64), nil
-		} else {
+		}
+		return new(int64), nil
+	// FIXME: for now, use either int64 or float64 for arbitrary precision numbers,
+	// but this is not enough for numbers outside their range.
+	case "NUMERIC", "NUMBER", "DECIMAL", "FIXED":
```
albscui, line 149: This is the core logic change.

jrockway (approved):
This looks good.  We should talk about what to do regarding timestamps without time zones.  Maybe doing nothing is OK; the user's code should know to discard the time zone information even if there is a time zone in the CSV/JSON.  But we should do a little research and see what the right thing to do is.

For outputting numeric data in JSON, I am not sure whether or not we should quote it.  Javascript doesn't have integers, but JSON doesn't care, the spec says it's just base 10 digits with - (-1), . (1.23), and e (1.23e45).  If the number from the database meets their formatting requirements, maybe we should just dump it into the JSON without quotes.  (Do not know how to do that in Go though.)  Always quoting it is certainly a fair approach, though.  If you were writing Go code, you'd just make your unmarshaling destination something like `struct { ID int, NumericNumber big.Rat, ...}` and it would work.  If you say "float64" it probably doesn't work though, and if you do have float64s in their database, they might be annoyed.  Dunno.  Always being a string is certainly consistent, but not always the most convenient.  

We can't really change the format with pachtf releases; we would need a way for a user to specify which "generation" of output they want.  If we ever feel like making it more convenient in the future.

src/internal/sdata/convert.go
```diff
@@ -437,3 +438,14 @@ func parseTime(x string) (t time.Time, err error) {
 func isNullString(x string) bool {
 	return x == "null" || x == "nil" || len(x) == 0
 }
+
+// formatTimestampNTZ trims the "Z" at the end of a timestamp if the "Z" exists
+// this is assuming that the timestamp has no time zone
+// example: 2022-05-06T20:18:10Z -> 2022-05-06T20:18:10
```
jrockway, line 444: I'd add a note explaining why we have to mutate a string instead of just formatting it correctly the first time.  (Because we don't know the type, and database/sql converts it to a time.Time before we can even get it as a string.)

jrockway (approved):
Looks good.  

We chatted on Slack about the literal string "null" floating around in the CSV code.

PR #7599, opened 2022-05-06
Enable copying root directory of source repo

acohen4:


src/server/pfs/server/paths.go
```diff
@@ -59,6 +60,14 @@ func cleanPath(p string) string {
 	return "/" + strings.Trim(p, "/")
 }
 
+// adds a trailing slash to the path if not already present
```
BOsterbuhr, line 63: Does this make `pachctl copy file repo1@master repo2@master` the same as `pachctl copy file repo1@master:/ repo2@master:/`?

src/server/pfs/server/driver_file.go
```diff
@@ -167,7 +167,7 @@ func (d *driver) copyFile(ctx context.Context, uw *fileset.UnorderedWriter, dst
 		return err
 	}
 	fs = fileset.NewIndexFilter(fs, func(idx *index.Index) bool {
-		return idx.Path == srcPath || strings.HasPrefix(idx.Path, srcPath+"/")
+		return idx.Path == srcPath || strings.HasPrefix(idx.Path, trailSlash(srcPath))
```
brycemcanally, line 170: You could use `fileset.Clean(srcPath, true)` for this. That is what we use in list file.

PR #7610, opened 2022-05-10
[CORE-551] add retry to egress

armaanv:
This adds a retry to the egress processing step of the pipeline


src/server/worker/pipeline/transform/registry.go
```diff
@@ -488,11 +488,20 @@ func (reg *registry) processJobEgressing(pj *pendingJob) error {
...
+			return err
+		}
+		return nil
+	}, backoff.NewInfiniteBackOff(), func(err error, d time.Duration) error {
+		pj.logger.Logf("error processing egress: %v, retrying in %v", err, d)
+		return nil
+	}); err != nil {
 		return err
 	}
+
```
brycemcanally, line 504: nit, arbitrary space.

src/server/pachyderm_test.go
```diff
@@ -1191,6 +1191,63 @@ func TestPipelineFailure(t *testing.T) {
 	require.True(t, strings.Contains(jobInfo.Reason, "datum"))
 }
 
+func TestEgressFailure(t *testing.T) {
+	// setup
```
brycemcanally, line 1195: nit, I don't think comments like this are useful and only take up space. Comments that describe something that isn't obvious from the code is helpful, but I don't think the comments in this test tell me anything that couldn't be easily gleaned from the code.

src/server/pachyderm_test.go
```diff
@@ -1191,6 +1191,63 @@ func TestPipelineFailure(t *testing.T) {
 	require.True(t, strings.Contains(jobInfo.Reason, "datum"))
 }
 
+func TestEgressFailure(t *testing.T) {
```
albscui, line 1194: Maybe add a check at the end to verify that `JobInfo.Restart` didn't change?

brycemcanally, line 1194: I don't think the job would terminate in this case. Does this test pass?

albcui (approved):
LGTM. Maybe add that check on the JobInfo state.

src/server/pachyderm_test.go
```diff
@@ -1191,6 +1191,54 @@ func TestPipelineFailure(t *testing.T) {
...
+	commit, err := c.StartCommit(repo, "master")
+	require.NoError(t, err)
+	require.NoError(t, c.PutFile(commit, "/test_table/0000", strings.NewReader("1,Foo\n2,Bar")))
+	require.NoError(t, c.FinishCommit(repo, "master", commit.ID))
+	var jobInfos []*pps.JobInfo
+	require.NoErrorWithinTRetry(t, time.Minute, func() error {
+		jobInfos, err = c.ListJob(pipeline, nil, -1, true)
+		require.NoError(t, err)
+		if len(jobInfos) != 1 {
+			return errors.Errorf("expected 1 jobs, got %d", len(jobInfos))
```
PFedak, line 1233: pretty sure this will never be true with the current ordering, you'll get a job when you create the pipeline, then a second one when you create the commit. You can either create the commit with data first or expect two jobs

I'm going to make a PR to fix `NoErrorWithinTRetry` to actually show the error you care about

src/server/pachyderm_test.go
```diff
@@ -1191,6 +1191,54 @@ func TestPipelineFailure(t *testing.T) {
...
+		}
+		return nil
+	})
+	jobInfo, err := c.InspectJob(pipeline, jobInfos[0].Job.ID, false)
+	require.NoError(t, err)
+	require.Equal(t, pps.JobState_JOB_EGRESSING, jobInfo.State)
```
PFedak, line 1239: I don't have a great solution but I'd be slightly happier if this wasn't just a point-in-time check. In particular this could potentially have succeeded before your change if it caught it at the right time in the retry cycle (although I guess it's vanishingly quick).

PFedak (commented):
sorry for the drive-by, heard your comment in standup and was curious

src/server/pachyderm_test.go
armaanv: I agree that this point in time check is not the best solution. I added in a sleep and a check to make sure that the job output exists to ensure that the retry is being exercised

src/server/pachyderm_test.go
armaanv: Thanks, I moved the commit before the pipeline

src/server/pachyderm_test.go
PFedak: I guess this is fine. You could do something like wait for it to be in egressing and then check in a loop several times that it's still egressing, but maybe it's not worth worrying about

PR #7615, opened 2022-05-11
Capture the relevant error in test retry loop.

PFedak:
Previously we would always return the context cancellation.


PFedak:
actually it looks like we used to do the right thing, but a refactor clobbered the error

PR #7618, opened 2022-05-11
CORE-586 Update how we handle null values in CSV

albscui:
TLDR 

```
        Go                    CSV
{"", `""`, nil, "null"} -> "","""""",,null
```

For CSV files, we special-case the string `null` to represent null values. This is problematic because it could collide with the literal string `null` in the database. This PR resolves this issue by using empty-value to represent null values, and a special-case string `""` to represent empty string. However, Go's `encoding/csv`'s Writer doesn't allow us to write `nil` values, nor write the literal `""`. To achieve this, we have to implement our own custom csv Reader and Writer, but instead of writing this from scratch, we just forked `encoding/csv` and modified its guts.


jrockway (approved):
This looks pretty good to me.  I would like probably a few more CSV-related tests, but maybe fuzzing is enough.

I think we should get rid of any special signifiers of null, see comment in regards to "isNullString".  I think that function should always return false; there should be no way of representing a null string inside our code other than (*string)(nil).

Kind of debated not hitting approve based on that, but let's chat about it.

jrockway (approved):
I like the removal of isNullString.  Internal and external consistency.  Hopefully users will not hate this :)

PR #7619, opened 2022-05-11
Wait for initial commit to finish in TestPipelineAutoscaling.

PFedak:
There was a race condition where, due to parallel jobs, the first
intended job could start while the initial (empty) job was still being
processed, leading to an unexpected scale of two. While totally correct
behavior, the test expects serial jobs.


acohen4 (approved):
LGTM

PR #7628, opened 2022-05-12
[CORE-377] Surface missing secret

robert-uhl:


src/internal/testpachd/mock_transaction.go
```diff
@@ -145,7 +147,7 @@ func (api *ppsTransactionAPI) UpdateJobStateInTransaction(txnCtx *txncontext.Tra
 	return errors.Errorf("unhandled pachd mock: pps.UpdateJobStateInTransaction")
 }
 
-func (api *ppsTransactionAPI) CreatePipelineInTransaction(txnCtx *txncontext.TransactionContext, req *pps.CreatePipelineRequest) error {
+func (api *ppsTransactionAPI) CreatePipelineInTransaction(ctx context.Context, txnCtx *txncontext.TransactionContext, req *pps.CreatePipelineRequest) error {
```
robert-uhl, line 150: Note that despite the similar names, `context.Context` and `txncontext.TransactionContext` are very different things; `txncontext.TransactionContext` does _not_ implement `context.Context`.

robert-uhl:
In offline discussion @PFedak revealed that the lack of a context in `CreatePipelineInTransaction` was deliberate, out of concern for long-running code inside a transaction. I’ll move the secret validation outside of the transaction.

src/server/pps/server/api_server.go
```diff
@@ -1494,6 +1495,19 @@ func (a *apiServer) validateEnterpriseChecks(ctx context.Context, req *pps.Creat
 	return nil
 }
 
+func (a *apiServer) validateSecret(ctx context.Context, req *pps.CreatePipelineRequest) error {
+	for _, s := range req.GetTransform().GetSecrets() {
+		_, err := a.env.KubeClient.CoreV1().Secrets(a.namespace).Get(ctx, s.Name, metav1.GetOptions{})
```
PFedak, line 1500: should we check `s.Key` exists, too? Not sure we're guaranteed to have access to that, I guess

src/server/pps/server/api_server.go
```diff
@@ -1760,10 +1778,6 @@ func (a *apiServer) CreatePipeline(ctx context.Context, request *pps.CreatePipel
 }
 
 func (a *apiServer) initializePipelineInfo(request *pps.CreatePipelineRequest, oldPipelineInfo *pps.PipelineInfo) (*pps.PipelineInfo, error) {
-	if err := a.validatePipelineRequest(request); err != nil {
```
PFedak, line 1763: we still want this, right?

PFedak (approved):
lgtm once you add back `validatePipelineRequest`

src/server/pps/server/api_server.go
robert-uhl: Yup, they did☺

src/server/pps/cmds/cmds_test.go
```diff
@@ -929,3 +929,33 @@ pachctl list job -x | match ' / 2'
 `).Run()
 	}, "expected to see two datums")
 }
+
+func TestPipelineWithoutSecrets(t *testing.T) {
```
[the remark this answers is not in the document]
robert-uhl: I thought it was appropriate since the CORE-377 is about UX.

src/server/pps/server/api_server.go
PFedak: oh, I actually meant in the secret object returned by k8s as well, but I don't really know enough to say if that makes sense

PFedak (approved):
lgtm

src/server/pps/server/api_server.go
robert-uhl: Ah, I think that _does_ make sense. Implemented that as well.

PR #7631, opened 2022-05-13
Fix errgroup contexts in taskDoer.Do()

acohen4:
because we previously created this `eg` without a context, it wouldn't break the inputChan consuming forloop when the eg.Go() would fail


albscui (approved):
LGTM!

src/internal/task/etcd_service.go
```diff
@@ -118,17 +118,17 @@ func newEtcdDoer(namespaceEtcd *namespaceEtcd, group string, cache Cache) Doer {
 
 func (ed *etcdDoer) Do(ctx context.Context, inputChan chan *types.Any, cb CollectFunc) error {
 	return ed.withGroup(ctx, func(ctx context.Context, renewer *col.Renewer) error {
-		var eg errgroup.Group
 		prefix := path.Join(ed.group, uuid.NewWithoutDashes())
 		done := make(chan struct{})
 		var count int64
 		ctx, cancel := context.WithCancel(ctx)
+		eg, egCtx := errgroup.WithContext(ctx)
```
brycemcanally, line 125: I think we should just overwrite `ctx`. It is fine to skip the deletion when an error occurs because it will get cleaned up by the lease expiration. Also, for a bit of clarity, this isn't addressing a correctness issue, just ensuring fail-fast behavior when the watch independently errors?

src/internal/task/etcd_service.go
```diff
@@ -118,15 +118,20 @@ func newEtcdDoer(namespaceEtcd *namespaceEtcd, group string, cache Cache) Doer {
...
 	return ed.withGroup(ctx, func(ctx context.Context, renewer *col.Renewer) error {
-		var eg errgroup.Group
 		prefix := path.Join(ed.group, uuid.NewWithoutDashes())
 		done := make(chan struct{})
 		var count int64
-		ctx, cancel := context.WithCancel(ctx)
 		defer func() {
-			cancel()
-			eg.Wait()
+			if _, err := col.NewSTM(ctx, ed.etcdClient, func(stm col.STM) error {
```
PFedak, line 125: I guess it doesn't matter given bryce's comment, but you could use the original `ctx` value here, rather than the error group context by doing `defer func(ctx context.Context){ ... }(ctx)`

PFedak (approved):

src/internal/task/etcd_service.go
```diff
@@ -162,60 +168,57 @@ func (ed *etcdDoer) Do(ctx context.Context, inputChan chan *types.Any, cb Collec
...
-		})
-		defer func() {
-			if _, err := col.NewSTM(ctx, ed.etcdClient, func(stm col.STM) error {
-				if err := ed.taskCol.ReadWrite(stm).DeleteAllPrefix(prefix); err != nil {
-					return errors.EnsureStack(err)
-				}
-				return errors.EnsureStack(ed.claimCol.ReadWrite(stm).DeleteAllPrefix(prefix))
-			}); err != nil {
-				fmt.Printf("errored deleting tasks with the prefix %v: %v\n", prefix, err)
+			if err != nil && !errors.As(err, context.Canceled) {
```
brycemcanally, line 171: I don't think this is correct. If we finish laying out the work and the context gets canceled by the caller while waiting, we would return nil rather than context canceled.

PR #7633, opened 2022-05-13
[INT-614] Cleanup s3-gateway APIClient usage

bbonenfant:
Refactor s3-gateway code to use `serviceenv.GetPachClient` as opposed to creating a new client for each request

Initial testing shows that this does in fact significantly speed up performance of the S3 gateway. Uploading 30 1kb files (one at a time) using these changes takes roughly 10s; when testing on current release (pachd v2.2.0), uploading 30 1kb files takes upwards of 2 minutes.


src/server/pfs/s3/auth.go
```diff
@@ -43,12 +40,7 @@ func (c *controller) SecretKey(r *http.Request, accessKey string, region *string
 
 func (c *controller) CustomAuth(r *http.Request) (bool, error) {
 	c.logger.Debug("CustomAuth")
-
-	pc, err := c.clientFactory()
-	if err != nil {
-		return false, errors.Wrapf(err, "could not create a pach client for auth")
-	}
-
+	pc := c.clientFactory(r.Context())
```
msteffen, line 43: Minor nit: it seems like it might be more conventional to store `env` in the `controller` and call `env.GetPachClient()` here, rather than specifically storing `env.GetPachClient` in `c.clientFactory`. But it's functionally the same (and technically slightly more flexible to store only the function if that's all you need).

@PFedak you're the core pach engineer here, so I don't know if you have an opinion on this.

msteffen (approved):
LGTM, nice work! This was fast.

@bbonenfant I don't know if there's any percentage in running a small benchmark on your laptop (N reads through grpc vs. through S3G) but if it's not too hard, it could be cool to confirm to ourselves that it's better now (plus we could check it in).

bbonenfant:
@msteffen I edited the summary with this info, but I'll also post it here:
> Initial testing shows that this does in fact significantly speed up performance of the S3 gateway. Uploading 30 1kb files (one at a time) using these changes takes roughly 10s; when testing on current release (pachd v2.2.0), uploading 30 1kb files takes upwards of 2 minutes.

PFedak (approved):
This code looks fine, but I'm confused about where performance gains are coming from. The only functional change is to the local testing deployment, right? Is this something customers actually use? Do we even use this in CI?

msteffen:
> This code looks fine, but I'm confused about where performance gains are coming from. The only functional change is to the local testing deployment, right? Is this something customers actually use? Do we even use this in CI?

I'm not actually sure, now that you point it out. I had it in my head that we weren't caching clients, and when I saw a piece of code that called `client.NewFromURI` get updated I assumed that was it, and am only now noticing that that code is in `testutil`.

The only other thing I noted in this PR is that pachd's main() function was, I believe, using the wrong context here, which may have been part of the issue: https://github.com/pachyderm/pachyderm/pull/7633/files#diff-1aa4ab905e56432d62cd27077bfb5fbd35430854b2cbaab1de9cf2603e294c10L1289

Other than that, I don't know? @bbonenfant when you tested it, were you using the pachd created by `testutil` by any chance (which would explain the improvement)? Or did you have a minikube cluster running? Part of me is like "we should benchmark to understand the fix" but part of me is like "why benchmark something that's already fixed?" I think if you have the time, Ben, figuring out what changed could actually be really useful (on the general principle that all bugs are really a bug in your understanding, rather than a bug in the code. The root cause of the slowness could be something that affects other parts of the S3G, or something that could be improved in other parts of Pachyderm) but my sense is that if you can't figure it out in a day or so, it may not be worth spending more time on than that.

WDYT?

bbonenfant:
I'm not sure what `testutil` means in this context, but I was testing using the pachd container created by the `docker-build` make recipe. Also I'm running everything in minikube. I'll do a little more investigating though, since it won't hurt to get more familiar with the inner workings of this module.

PFedak:
> I'm not sure what `testutil` means in this context, but I was testing using the pachd container created by the `docker-build` make recipe. Also I'm running everything in minikube. I'll do a little more investigating though, since it won't hurt to get more familiar with the inner workings of this module.

Basically, the place you changed from creating a new connection to `localhost` to a pachclient is part of a special setup for testing pachd locally (not to be confused with a local deployment). If you're running pachyderm normally, that change shouldn't be relevant. The rest of the changes don't seem like they would cause meaningful performances changes (creating a new `APIClient` struct is just copying a small amount of memory, no new grpc connections etc were being created there). Maybe it would be useful to try to get a breakdown of where the extra overhead was before this change?

msteffen:
We need to merge this into the 2.x branch

PR #7634, opened 2022-05-16
[CORE-559] add a http handler for oidc health check

armaanv:


jrockway (approved):
Couple things:

1) You should fix https://pachyderm.atlassian.net/browse/CORE-467 as part of this.  (You'll notice that you added a / handler to the OIDC server, but now the metrics server also handles / 😂)

2) Ideally this health check would actually check the health of the OIDC service.

I think this is pretty OK for unblocking people that are trying to set up GKE ingress though, so both can be done in a followup.

PR #7637, opened 2022-05-16
[CORE-610] Support Snowflake variant type

robert-uhl:
This PR adds support for the Snowflake VARIANT type. In Go, it is represented as an `interface{}`. This requires a special bulk insert form, particular to Snowflake.

It also refactors the tests to have a single master list of supported databases (so far: PostgreSQL, MySQL & Snowflake).

Finally, it corrects a typo in a function name.


src/internal/sdata/sql.go
```diff
@@ -78,9 +78,24 @@ func (m *SQLTupleWriter) GeneratePreparedStatement() (*pachsql.Stmt, error) {
...
-		tableInfo.Name,
-		strings.Join(tableInfo.ColumnNames(), ", "))
-	return &SQLTupleWriter{tx, tableInfo, insertStatement, []Tuple{}}
+	var s string
+	if tableInfo.Driver == "snowflake" {
+		var asClauses []string
+		for i, col := range tableInfo.Columns {
+			var asClause string
+			if col.DataType == "VARIANT" {
+				asClause = fmt.Sprintf(`to_variant(COLUMN%d) as %s`, i+1, col.Name)
```
albscui, line 87: This can be simplified to just `to_variant(COLUMN%d)`, because I don't think the `as` here is doing anything.

We should explicitly add the column names as part of the `INSERT INTO`. For example, let  `co1` be a `VARIANT` type:
```
INSERT INTO <table_name> (col1, col2, ...) SELECT to_variant(COLUMN1), COLUMN2, ... FROM VALUES 
```

src/internal/sdata/sdata_test.go
```diff
@@ -224,8 +184,24 @@ func TestSQLTupleWriter(suite *testing.T) {
 				// for mysql compatibility
 				*ti = time.Now()
 			})
+			fz.Funcs(func(x *interface{}, co fuzz.Continue) {
```
albscui, line 187: Hmm, this seems like a duplicate of https://github.com/pachyderm/pachyderm/pull/7637/files#diff-16ac2fdb87cc8640c10a72c7be681809939ca82348c89981bef0b2de48477645R115

I feel like we can refactor `GenerateTestData` to just generate data and not execute the SQL. Then we can re-use that logic where is needed.

albscui (commented):
Thank you so much for improving the test db utility functions! When I was working with this stuff, it felt like I was digging myself into a hole that I couldn't get out of.

Overall, this looks good, just needs some cleanup. I think the `INSERT INTO` in `NewSQLTupleWriter` can be refined a bit.

src/internal/pachsql/util.go
```diff
@@ -72,66 +71,159 @@ type TestRow struct {
...
-	var cols []string
-	rty := reflect.TypeOf(x)
-	for i := 0; i < rty.NumField(); i++ {
-		field := rty.Field(i)
-		col := field.Tag.Get("column")
-		cols = append(cols, col)
-	}
-	return "(" + strings.Join(cols, ", ") + ")"
-}
+// func GenerateTestData(db *DB, tableName string, n int, row SetIDer) error {
```
albscui, line 107: cleanup

src/internal/pachsql/util.go
```diff
@@ -72,66 +71,159 @@ type TestRow struct {
...
-	for i := 0; i < n; i++ {
-		fz.Fuzz(&row)
-		row.Id = int16(i)
-		if _, err := db.Exec(insertStatement, makeArgs(row)...); err != nil {
-			return errors.EnsureStack(err)
-		}
-	}
-	return nil
-}
+// type SetIDer interface {
```
albscui, line 103: cleanup

albscui (approved):
LGTM, just need to clean up commented out code

PR #7639, opened 2022-05-17
[CORE-653] Fix deployment test

robert-uhl:
Helm chart version was not merged into the options structure used when deploying, which resulted in deploying an incompatible combination of chart and image.


acohen4 (approved):
LGTM

PR #7635, opened 2022-05-16
Worker Master deletes meta files for unskippable datums

acohen4:
Execute point deletes from the client on the PPS Worker Master because directory deletes are slow when operating on a commit with many intermediate filesets

PR #7642, opened 2022-05-17
[2.2.x Backport] Worker Master deletes meta files for unskippable datums (#7635)

acohen4:
* Worker Master deletes meta files for unskippable datums

* Fix path for deleting unskippable base meta commit datum job input files


brycemcanally (approved):
lgtm

PR #7653, opened 2022-05-19
[Mount Server] Add logic for repo access

smalyala:
- Adds a new `authorization` field in the response for `GET /repos` indicating what level of access a user has on a repo
- Prevents user from mounting a repo they don't have access to

Jira: https://pachyderm.atlassian.net/browse/INT-623


src/server/pfs/fuse/server.go
```diff
@@ -188,27 +188,41 @@ func (mm *MountManager) ListByRepos() (ListRepoResponse, error) {
...
+				}
+				if br.Mount == nil {
+					br.Mount = append(br.Mount, MountState{State: "unmounted"})
+				}
+				rr.Branches[branch.Branch.Name] = br
 			}
-			if br.Mount == nil {
-				br.Mount = append(br.Mount, MountState{State: "unmounted"})
+			if repo.AuthInfo == nil {
+				rr.Authorization = "off"
```
pleeko, line 220: This is the case for workspaces where auth is not enabled correct?

src/server/pfs/fuse/server.go
```diff
@@ -858,6 +872,24 @@ func jsonMarshal(t interface{}) ([]byte, error) {
...
+func hasRepoRead(permissions []auth.Permission) bool {
+	for _, p := range permissions {
+		if p == auth.Permission_REPO_READ {
+			return true
+		}
+	}
+	return false
+}
+
+func hasRepoWrite(permissions []auth.Permission) bool {
```
pleeko, line 884: We don't need to check for the role `repoOwner` because we are checking for the specific write permission inside of the role? Would we have to add a similar function in the future if we wanted to check for other privileges like:
```
auth.Permission_REPO_DELETE,
auth.Permission_REPO_INSPECT_COMMIT,
auth.Permission_REPO_LIST_COMMIT,
auth.Permission_REPO_DELETE_COMMIT,
auth.Permission_REPO_CREATE_BRANCH,
auth.Permission_REPO_LIST_BRANCH,
auth.Permission_REPO_DELETE_BRANCH,
auth.Permission_REPO_LIST_FILE
```

src/server/pfs/fuse/server_test.go
```diff
@@ -172,6 +172,39 @@ func TestBasicServerDifferingNames(t *testing.T) {
 	})
 }
 
+func TestRepoAccess(t *testing.T) {
+	c, _ := minikubetestenv.AcquireCluster(t)
+	tu.ActivateAuthClient(t, c)
+	alice, bob := "robot"+tu.UniqueString("alice"), "robot"+tu.UniqueString("bob")
+	aliceClient, bobClient := tu.AuthenticateClient(t, c, alice), tu.AuthenticateClient(t, c, bob)
```
pleeko, line 179: How are the different permissions being applied to the two clients here?

pleeko (approved):
Looks good! I have a few questions

src/server/pfs/fuse/server.go
smalyala: off- auth disabled
none- auth enabled and user has no permissions

src/server/pfs/fuse/server.go
smalyala: Good question. A user (i.e. root, robot, pipeline, etc.) can't be created by mixing and matching permissions. They can only be given roles (i.e. repoReader, repoWriter, etc.). https://pachyderm.slack.com/archives/C02C8TUF493/p1653398328191089

As a result, the roles that contain the `repo_write` permission are finite and known: `repoWriter`, `repoOwner`, and `clusterAdmin`. We're okay with these roles writing to a repo which means checking to see if a user has the `repo_write` permission is sufficient (which relates to your first question). 

To your second question, all of those (except for `repo_delete`) are associated with the `repoWriter` role which passes our `hasRepoWrite()` check. If we later allow a user to delete a mounted repo, that would be a slight tweak, which we can address if we decide to add that to notebooks.

PR #7654, opened 2022-05-19
ActivateAuth with a root token injected via pachd's environment

acohen4:


jrockway (approved):
This ended up being a lot cleaner than I imagined.  I like the part of the background job that does a "rotate" in the background.  I think this is going to be really pleasant for users.

src/server/auth/server/api_server.go
```diff
@@ -153,7 +155,7 @@ func (a *apiServer) envBootstrap(ctx context.Context) {
 			return nil
```
jrockway, line 155: I think we might prefer the "Successfully configured auth server via environment" message to be explicitly logged here.  I think with it deferred we also log when panicking with a failure message?

src/testing/deploy/deploy_test.go
```diff
@@ -34,6 +34,28 @@ func TestDeployEnterprise(t *testing.T) {
...
+	require.Equal(t, auth.RootUser, whoami.Username)
+	// set new root token via env
+	opts.AuthUser = ""
+	token := "new-root-token"
+	opts.ValueOverrides = map[string]string{"pachd.rootToken": token}
+	c = minikubetestenv.UpgradeRelease(t, context.Background(), "default", k, opts)
+	c.SetAuthToken(token)
+	whoami, err = c.AuthAPIClient.WhoAmI(c.Ctx(), &auth.WhoAmIRequest{})
+	require.NoError(t, err)
+	require.Equal(t, auth.RootUser, whoami.Username)
```
jrockway, line 56: Can we also check that the old token doesn't somehow still work?

(Some random complexity that you don't have to deal with here, but we might want to think about in the future.  I like non-aborting assertions for these test cases where we want to test two things, so that when it fails, it will say "old key didn't work\n new key didn't work".  That gives you more information about the state of the system when the test fails. 

 I don't actually know if we have non-aborting assertions in our custom assertion library, though.)

jrockway (approved):
I continue to enjoy this feature!  Some small logging and testing nits below.  Very minor.

src/testing/deploy/deploy_test.go
acohen4: we don't have the non-aborting assertions, but I can add the check :)

PR #7655, opened 2022-05-19
Ensure Snowflake connection string logic works

albscui:
Originally, we wanted to add support for allowing users to specify their role in the Snowflake connection string. It's now apparent that this is already supported, but not verified. This PR adds tests to verify that this feature, along with others are supported, as well as a small refactor in how we parse the Snowflake URL.

### DIFF
--- a/src/internal/pachsql/db.go
+++ b/src/internal/pachsql/db.go
@@ -118,29 +118,29 @@ func snowflakeDSN(u URL, password string) (string, error) {
 	// however, the "snowflakecomputing.com" can be left out
 	// example: jsmith@my_organization-my_account/mydb/testschema?warehouse=mywh
 	// in this case, the account_identifier is my_organization-my_account
-	var account, host string
-	if u.Port == 0 {
-		account = strings.TrimSuffix(u.Host, SnowflakeComputingDomain)
-	} else {
-		account = u.Params["account"]
-		host = u.Host
-	}
-
 	params := make(map[string]*string, len(u.Params))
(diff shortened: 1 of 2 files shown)


src/internal/pachsql/db.go
```diff
@@ -118,29 +118,29 @@ func snowflakeDSN(u URL, password string) (string, error) {
...
 
 	cfg := &sf.Config{
-		Account:   account,
-		User:      u.User,
-		Password:  password,
-		Database:  u.Database,
-		Schema:    u.Schema,
-		Host:      host,
-		Port:      int(u.Port),
-		Warehouse: u.Params["warehouse"],
```
robert-uhl, line 142: Is dropping `Warehouse` deliberate?

Maybe it would be good to just pass the URL minus the scheme to `snowflake.ParseDSN`, or use a package like https://github.com/xo/dburl?

albscui: Yea, I realized the `warehouse` was being duplicated in the final dsn generated, and that the more general way is to just feed everything via the `params`.

I agree with your suggestion about wanting to leverage `snowflake.ParseDSN`.  The way we separate out the password from the user's URL makes this a bit tricky.  `ParseDSN` errors if you don't provide a password, so we have to inject the password into the user's URL. Also, we would have to do a major refactor of `pachsql` to generalize this across all databases. For that reason, I think we should keep things the way they are for now.

robert-uhl (approved):
LGTM.

PR #7663, opened 2022-05-20
Include gRPC status codes for more errors.

PFedak:


jrockway (approved):
I tried this out locally with a version of pachctl that prints the gRPC code that's received from the wire, and everything is coming through ok.  This looks good to me.

We need to follow up and make some other related errors have codes.  Opening a ticket for that.

src/internal/middleware/errors/interceptor.go
```diff
@@ -5,28 +5,55 @@ import (
...
+		if res, ok := status.FromError(cursor); ok {
+			code = res.Code()
+			break
 		}
+		cursor = errors.Unwrap(cursor)
 	}
-	return err
+	return &gRPCStatusError{
+		code: code,
+		err:  err,
```
jrockway, line 40: The sticky bit here is that we end up with a somewhat verbose error message.  For example, if we end up with an error chain like:

```go
fmt.Errorf("wrap: %w", status.Errorf(codes.FailedPrecondition, "i personally do not like the montage pipeline"))
```
Then this now prints as: `wrap: rpc error: code = FailedPrecondition desc = i personally do not like the montage pipeline`, whereas before it was `i personally do not like the montage pipeline`.  What we really want, but is hard to get, is `wrap: i personally do not like the montage pipeline`.

I think what you have here is an improvement, since information is never lost, but there is some confusion cost for people looking at the messages.  (The hard-coded string "rpc error" that status.Error() adds might technically be inaccurate, for example.)  I don't think we can really do better, easily, because we can't get at the constructor (or private fields) of every error that implements `interface { GRPCStatus() *Status }`, so we can't rewrite them.  We also can't guarantee that we control the `Error` method on the chain of errors; it can format however it wants.  

Long term, we probably need our own API for this, so methods can annotate their replies with GRPC status codes, without introducing the "rpc error: code = %s desc = %s" in the middle of an error chain.  Sigh!

jrockway: But you know what, we rarely use `status.Error` or `status.Errorf` directly, so this all might be perfectly OK.  Just some stuff in identity that might look messy if we don't make a dedicated error type for them:

```
src/identity/identity.go
12:     ErrInvalidID = status.Error(codes.Internal, "ID does not exist")
15:     ErrAlreadyExists = status.Error(codes.Internal, "ID already exists")

src/auth/auth.go
102:    ErrNotActivated = status.Error(codes.Unimplemented, "the auth service is not activated")
106:    ErrAlreadyActivated = status.Error(codes.Unimplemented, "the auth service is already activated")
112:    ErrNotSignedIn = status.Error(codes.Unauthenticated, "no authentication token (try logging in)")
116:    ErrNoMetadata = status.Error(codes.Internal, "no authentication metadata (try logging in)")
120:    ErrBadToken = status.Error(codes.Unauthenticated, "provided auth token is corrupted or has expired (try logging in again)")
124:    ErrExpiredToken = status.Error(codes.Internal, "token expiration is in the past")

src/license/license.go
12:     ErrDuplicateClusterID = status.Error(codes.Unimplemented, "cluster ID is not unique")
15:     ErrInvalidIDOrSecret = status.Error(codes.Unimplemented, "cluster ID or secret is not valid")
18:     ErrNotActivated = status.Error(codes.Unimplemented, "cluster does not have enterprise features enabled")
```

src/internal/middleware/errors/interceptor.go
```diff
@@ -22,12 +23,25 @@ func StreamServerInterceptor(srv interface{}, stream grpc.ServerStream, info *gr
...
+	}
+	if cursor == nil && err != nil {
+		// this should never happen, we unwrapped an error consisting only of stack traces
+		// reset to the original error, we have no idea what this is
+		cursor = err
+	}
+	if _, ok := status.FromError(cursor); ok {
 		// handles nil as well
-		return err //nolint:wrapcheck
+		return cursor //nolint:wrapcheck
```
jrockway, line 42: Any reason not to return `&gRPCStatusError{code: code, err: err}` from this branch?  That way we'd be keeping the stack trace, I think.  (Maybe gRPCStatusError should implement StackTracer and delegate to the underlying?  That would definitely complicate this PR, though, so maybe not worth it.)

jrockway (approved):
New changes look good.  I have a question, but it's not blocking?

src/internal/middleware/errors/interceptor.go
PFedak: Yes, sorry, I should have been clearer. I explicitly want to strip the stack (the grpc handler code flattens the error immediately after, anyway) to avoid the double-wrapping issue that was happening with auth errors. The only issue was they were being wrapped with stack traces - I think it's fine for us to use `status.Error` in cases like that where the request hasn't even really started yet.

jrockway: Makes sense!

src/internal/middleware/errors/interceptor.go
PFedak: As discussed elsewhere, I think this is handled now - we pass through the auth cases properly, and the pattern in other parts of the code of implementing the status code interface will work properly

jrockway: Agree.

PR #7664, opened 2022-05-21
Let Envoy terminate TLS

jrockway:
This adds a configuration to values.yaml to let Envoy terminate TLS.  If you have a cert (or can get one with, say, cert-manager), then you can put this directly on the Internet and get HTTPS support.

The way this works is that we have different envoy config files for the TLS and non-TLS case, generated with `etc/generate-envoy-config` as normal.

I've adjusted the routes for the TLS and non-TLS case to be as friendly as possible.  If you install the non-TLS version and port-forward port 8443 on the pachyderm-proxy pod, you'll get a message saying to reconfigure helm to enable TLS.  (Note that the LoadBalancer/NodePort won't have this port forwarded, however.  You have to dig pretty deep to start debugging.)

For the TLS case, we still serve traffic on port 80, but it redirects to 443.  (Actually, it redirects by changing the scheme from http to https, so if you're using non-standard port numbers, it won't work.)  Port 443 will respond to cleartext requests (`GET http://example.com:443`) with the same redirect.

To test this locally, you'll need to start console with `NODE_TLS_REJECT_UNAUTHORIZED=0` (because it goes through the proxy to get the openid config) and hack up pachctl to trust whatever self-signed certs you're using for testing.  If you do that, it does work though, including websockets.

This PR adds test for the proxy, including console.  Console is deployed as an nginx container that serves static HTML that we make sure we can retrieve through the proxy.  We also test file uploads that are larger than the GRPC max message size, and S3 downloads.  Some refactoring of minikubetestenv was necessary to let us cleanly deploy with TLS; minikubetestenv records "extra options" that were used and when a cluster is reassigned, it is upgraded to match the desired options if it didn't already have those options.  (This does not work if you use `-cluster.reuse` but I don't think that ever worked correctly if the last test did an upgrade.)  Maybe it would be best to always upgrade after a cluster is assigned?  I think most of the time, that will make `reuse` work?

Need to test:
[x] ~~Against a real/trusted TLS certificate.~~ (Works.)
[x] ~~That cert rotation works (things have changed since I wrote https://jrock.us/posts/rotating-envoy-certs/)~~ (works.)
[x] ~~How settings here interact with `ingress.tls` (namely making the issuer uri https://).~~  (Done.  Will lead to problems if someone tries to do http -> ingress controller -> https -> proxy, but that is not really possible with the current state of the helm chart.)

Need to add:
[x] ~~HSTS headers~~
[x] ~~Tests that deploy this during CI, and try to make grpc, auth, and s3 requests over HTTPS.~~ (Added in proxy_test.go.)

### DIFF
--- a/etc/generate-envoy-config/envoy.libsonnet
+++ b/etc/generate-envoy-config/envoy.libsonnet
@@ -90,7 +90,7 @@
     ],
   },
 
-  httpConnectionManager(name, routes):: {
+  httpConnectionManager(name, routes=[], response_headers_to_add={}):: {
     name: 'envoy.http_connection_manager',
     typed_config: {
       '@type': 'type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager',
@@ -138,6 +138,15 @@
       request_timeout: '604800s',  // Necessary to allow long file uploads.
       stream_idle_timeout: '600s',  // Only completely idle streams are dropped after this timeout.
       route_config: {
+        [if std.length(response_headers_to_add) > 0 then 'response_headers_to_add' else null]: [
+          {
+            header: {
+              key: key,
+              value: response_headers_to_add[key],
+            },
+          }
+          for key in std.objectFields(response_headers_to_add)
+        ],
         virtual_hosts: [
           {
             domains: ['*'],
@@ -249,7 +258,7 @@
             },
           },
         },
-        filters: [$.httpConnectionManager(name, routes)],
+        filters: [$.httpConnectionManager(name, routes, response_headers_to_add={ 'strict-transport-security': 'max-age=604800' })],
       },
 
       // Redirect to https if this request wasn't sent over TLS.
--- a/etc/generate-envoy-config/envoy.libsonnet
+++ b/etc/generate-envoy-config/envoy.libsonnet
@@ -70,7 +70,7 @@
     },
   },
 
-  loadAssignment(name, address, port): {
+  loadAssignment(name, address, port):: {
     cluster_name: name,
     endpoints: [
       {
@@ -90,6 +90,93 @@
     ],
   },
 
+  httpConnectionManager(name, routes):: {
+    name: 'envoy.http_connection_manager',
+    typed_config: {
+      '@type': 'type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager',
+      access_log: [
+        {
+          name: 'envoy.access_loggers.stdout',
+          typed_config: {
+            '@type': 'type.googleapis.com/envoy.extensions.access_loggers.stream.v3.StdoutAccessLog',
+          },
+        },
+      ],
+      codec_type: 'auto',
+      common_http_protocol_options: {
+        headers_with_underscores_action: 'REJECT_REQUEST',
+        idle_timeout: '60s',
+      },
+      http2_protocol_options: {
+        initial_connection_window_size: 1048576,
+        initial_stream_window_size: 65536,
+        max_concurrent_streams: 100,
+      },
+      http_filters: std.prune([
+        // If there's a GRPC router, enable grpc_stats.
+        if std.foldl(function(last, route) last || 'grpc' in route.match, routes, false) then
+          {
+            name: 'envoy.filters.http.grpc_stats',
+            typed_config: {
+              '@type': 'type.googleapis.com/envoy.extensions.filters.http.grpc_stats.v3.FilterConfig',
+              enable_upstream_stats: true,
+              // Stats for all methods = true would let anyone on the Internet fill up the
+              // statistics buffer with bogus RPC method names, so we have to leave it off.
+              stats_for_all_methods: false,
+            },
+          } else {},
+        {
+          name: 'envoy.filters.http.router',
+          typed_config: {
+            '@type': 'type.googleapis.com/envoy.extensions.filters.http.router.v3.Router',
+          },
+        },
+      ]),
+      http_protocol_options: {
+        accept_http_10: false,
+      },
+      request_timeout: '604800s',  // Necessary to allow long file uploads.
+      stream_idle_timeout: '600s',  // Only completely idle streams are dropped after this timeout.
+      route_config: {
+        virtual_hosts: [
+          {
+            domains: ['*'],
+            name: 'any',
+            routes: routes,
+            retry_policy: {
+              retry_on: 'connect-failure',
+              num_retries: 4,
+              host_selection_retry_max_attempts: 4,
+            },
+          },
+        ],
+      },
+      stat_prefix: name,
+      use_remote_address: true,
+    },
+  },
+
+  redirectToHttpsRoute: {
+    match: {
+      prefix: '/',
+    },
+    redirect: {
+      https_redirect: true,
+    },
+  },
+
+  messageRoute(status, message): {
+    match: {
+      prefix: '/',
+    },
+    direct_response: {
+      status: status,
+      body: {
+        inline_string: message,
+      },
+    },
+  },
+
   httpListener(port, name, routes): {
     name: name,
     per_connection_buffer_limit_bytes: 32768,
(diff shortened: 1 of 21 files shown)


src/internal/testutil/enterprise.go
```diff
@@ -14,10 +14,9 @@ import (
 func GetTestEnterpriseCode(t testing.TB) string {
 	acode, exists := os.LookupEnv("ENT_ACT_CODE")
 	if !exists {
-		t.Error("Enterprise Activation code not found in Env Vars")
+		t.Error("Enterprise activation code not found in environment variable (ENT_ACT_CODE)")
```
acohen4, line 17: <3

src/server/proxy_test.go
```diff
@@ -0,0 +1,301 @@
...
+	}, "console-proxy-backend should have endpoints")
+}
+
+func TestTrafficThroughProxy(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration test in short mode")
+	}
+	t.Parallel()
+	c, ns := minikubetestenv.AcquireCluster(t)
+	deployFakeConsole(t, ns)
```
acohen4, line 220: I keep reading deepFakeConsole

src/internal/testutil/enterprise.go
jrockway: Haha, guess who didn't have that environment variable set when he first tried to run a minikubetestenv test 😂

src/internal/minikubetestenv/client_factory.go
```diff
@@ -146,19 +184,22 @@ func AcquireCluster(t testing.TB, opts ...Option) (*client.APIClient, string) {
...
+	assigned, mc := clusterFactory.acquireFreeCluster()
 	if assigned == "" {
-		assigned, c = clusterFactory.acquireNewCluster(t, as)
+		assigned, mc = clusterFactory.acquireNewCluster(t, as)
 	}
-	// in the case loki is requested, upgrade the cluster to include it
-	if as.WaitForLoki {
-		c = UpgradeRelease(t,
+
+	// If the cluster settings have changed, upgrade the cluster to make them take effect.
```
acohen4, line 192: 👍

etc/helm/pachyderm/templates/proxy/service.yaml
```diff
@@ -17,12 +17,22 @@ metadata:
 spec:
   type: {{ .Values.proxy.service.type }}
   ports:
+    {{- if .Values.proxy.service.httpPort }}
```
acohen4, line 20: if a user wants TLS and not HTTP, do they have to set proxy.service.httpPort=0 ?

etc/helm/pachyderm/values.yaml
```diff
@@ -752,4 +752,23 @@ proxy:
...
+    # ingress controller talk to the proxy over TLS, in which case, it's fine to enable TLS here in
+    # addition to in the ingress section above.)
+    enabled: false
+    # The secret containing "tls.key" and "tls.crt" keys that contain PEM-encoded private key and
+    # certificate material.  Generate one with "kubectl create secret tls <name> --key=tls.key
+    # --cert=tls.cert".  This format is compatible with the secrets produced by cert-manager, and
+    # the proxy will pick up new data when cert-manager rotates the certificate.
+    secretName: ""
+    # If set, generate the secret from values here.  This is intended only for unit tests.
+    secret: {}
```
acohen4, line 774: do we want to add key/cert fields here?

etc/helm/pachyderm/templates/proxy/service.yaml
jrockway: Yeah.  Maybe we should document that better.

etc/helm/pachyderm/templates/proxy/deployment.yaml
```diff
@@ -102,8 +102,17 @@ spec:
...
+              mountPath: /tls
+            {{- end }}
       volumes:
         - name: envoy-config
           configMap:
             name: pachyderm-proxy-config
+        {{- if .Values.proxy.tls.enabled }}
+        - name: tls
+          secret:
+            secretName: {{ .Values.proxy.tls.secretName }}
```
acohen4, line 116: should this hook up to the helm generated `pachyderm-proxy-tls`?

etc/helm/pachyderm/values.yaml
jrockway: This is really a hack to make testing easier; I don't think users will want to configure TLS this way.  (You have to set secretName if you set it here, and the secretName is hard-coded to some name I've already forgotten.)

etc/generate-envoy-config/envoy.libsonnet
```diff
@@ -70,7 +70,7 @@
     },
   },
 
-  loadAssignment(name, address, port): {
+  loadAssignment(name, address, port):: {
```
acohen4, line 73: ?

jrockway: This makes it non-exported.  See "visibilities" in https://jsonnet.org/ref/language.html

etc/generate-envoy-config/envoy.libsonnet
```diff
@@ -102,79 +198,80 @@
...
-                      '*',
-                    ],
-                    name: 'any',
-                    routes: routes,
-                    retry_policy: {
-                      retry_on: 'connect-failure',
-                      num_retries: 4,
-                      host_selection_retry_max_attempts: 4,
+                  name: 'tls',
+                  sds_config: {
```
acohen4, line 247: so does SDS get deployed in a separate pod?

etc/generate-envoy-config/envoy.libsonnet
```diff
@@ -188,7 +285,8 @@
       tcp_keepalive: {},
     },
   },
-  GRPCCluster: self.defaultCluster {
+
+  defaultGRPCCluster:: self.defaultCluster {
```
acohen4, line 289: why is this config necessary for grpc?

etc/generate-envoy-config/envoy.libsonnet
jrockway: xDS can happen over the network or by files containing the messages that would be sent over the network.  In the case of discovering the TLS secret, we just use files, hence `path_config_source`.

When you edit a configmap or secret in k8s, the mounted volumes actually change to reflect the update.  The `watched_directory` stuff picks up on that.  That way when something like cert-manager rotates the TLS secret, we immediately start using it without having to restart.

etc/helm/pachyderm/templates/proxy/deployment.yaml
acohen4: so outcome here is to keep things as they are and ask customers to create their own secret? That makes sense to me

etc/generate-envoy-config/envoy.libsonnet
jrockway: Needs an empty http2_protocol_options to tell envoy that the upstream protocol is http2.  This called "h2c" (http2 over cleartext).  Normally HTTP/2 is negotiated during ALPN, which is part of TLS.

gRPC mandates HTTP/2, but Envoy doesn't know that this cluster will be used for gRPC.  If pachd were doing TLS, then we'd autonegotiate it, but since it's not, we have to force http2 here.

Fun story... one time I randomly ran into someone who was working on Envoy and the empty `http2_protocol_options` was the first thing they mentioned 😂

etc/helm/pachyderm/templates/proxy/service.yaml
acohen4: could we just not run this section of template if tls.enabled?

etc/helm/pachyderm/templates/proxy/deployment.yaml
jrockway: I think you kind of talked me into doing it like the rest of the chart.  You will have to set the name, but I think that's OK.

acohen4 (approved):
LGTM!

etc/helm/pachyderm/templates/proxy/service.yaml
jrockway: It is nice to have HTTP even if you are using TLS because the browser will try port 80 first.  If nothing is listening there, the browser will error before trying HTTPS on port 443.  (I don't know why this is the case.)

What we do when TLS is enabled is make port 80 send a redirect from http://whatever to https://whatever, which upgrades to TLS and starts using port 443 instead.

We do set HSTS ("hypertext strict transport security") headers so that once a user successfully visits pachyderm over https, their browser will prevent them from loading the site over http.  This prevents a man in the middle from downgrading the user to HTTP if they happen to have http://whatever bookmarked.  (I set the caching for this to 1 week, so if a company wants to disable TLS, eventually users will be able to use it without any action.  It's a tradeoff between convenience and security, weighted towards security.  Over on the public Internet, bots are roaming around looking for HSTS headers and they actually ship a list of what they find with the browser.  So if you have never visited, say, jrock.us, and you type http://jrock.us into the URL bar it will never connect to port 80!  This leaves no window for an attacker to MITM the http connection, but of course means I could never turn off HTTPS.)

etc/helm/pachyderm/templates/proxy/deployment.yaml
jrockway: I made the generated secret match the secretName, erroring when it's not set.  So there is now no magic string when generating a secret from key/cert.  Also added an error for the case where you set one of key/cert without the other (easy to spell cert as `cret` or something).

PR #7667, opened 2022-05-24
Refactor main.go 1

acohen4:


src/server/cmd/pachd/main.go
```diff
@@ -181,6 +192,28 @@ func doEnterpriseMode(config interface{}) (retErr error) {
...
+
+func doReadinessCheck(config interface{}) error {
+	env := serviceenv.InitPachOnlyEnv(serviceenv.NewConfiguration(config))
+	return env.GetPachClient(context.Background()).Health()
+}
+
+func doEnterpriseMode(config interface{}) (retErr error) {
+	defer func() {
+		if retErr != nil {
+			log.WithError(retErr).Print("failed to start server")
```
PFedak, line 205: seems like errors from `<-errChan` will end up here, too, not necessarily from starting. goroutine dump is presumably m ore interesting in the unexpected-error-while-running case

src/server/cmd/pachd/main.go
```diff
@@ -102,18 +102,7 @@ func main() {
...
-}
-
-func doEnterpriseMode(config interface{}) (retErr error) {
-	defer func() {
-		if retErr != nil {
-			log.WithError(retErr).Print("failed to start server")
-			pprof.Lookup("goroutine").WriteTo(os.Stderr, 2)
-		}
-	}()
+func commonSetup(config interface{}, service string, initDB, initDexDB bool) (env *serviceenv.NonblockingServiceEnv, err error) {
```
PFedak, line 105: definitely a nit but I strongly dislike multiple boolean arguments. Here I guess it's simple as they are always the same value. From a readability perspective I'd prefer to pass them in a way that includes names, either through functional options or an options struct. Alternatively, maybe just make a separate "setUpDB" function that does both, since it's not strictly "common"

src/server/cmd/pachd/main.go
```diff
@@ -556,48 +539,10 @@ func doFullMode(config interface{}) (retErr error) {
 	var ctx = context.Background()
 	interruptChan := make(chan os.Signal, 1)
 	signal.Notify(interruptChan, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
-	defer func() {
-		if retErr != nil {
-			pprof.Lookup("goroutine").WriteTo(os.Stderr, 2)
-		}
-	}()
```
PFedak, line 563: did you intend to delete this here?

PFedak (approved):
lgtm, though I don't think you meant to delete one of the error handling checks

acohen4:
opened https://github.com/pachyderm/pachyderm/pull/7845 instead

PR #7669, opened 2022-05-25
[2.2.x backport] Add the proxy

jrockway:
This is a somewhat messy backport because the testing code changed out from under us.  I think I cherry-picked the right PR to make everything work though.


jrockway:
Going to try this again.

PR #7673, opened 2022-05-25
Fixes/adds commit mounting.

jdoliner:


smalyala:
Just one question, otw LGTM!

jdoliner:
@smalyala I expanded this to include mounting of files and commits. I also beefed up the tests. Please take another look when you have a chance. Also I'm not sure if I'm just failing at gh but I don't see your question.

smalyala:
Lmk if you can see the question now, should be right above this comment.

PR #7679, opened 2022-05-25
Bootstrap License Server via environment

acohen4:


src/server/license/server/api_server.go
```diff
@@ -35,11 +38,69 @@ func New(env Env) (lc.APIServer, error) {
 		env:     env,
 		license: licenseCollection(env.DB, env.Listener),
 	}
+	go s.envBootstrap(context.Background())
```
jrockway, line 41: Not 100% sure about this but I started a Slack discussion.

I think it would be best if we run this inline, not in the background, and return the error from envBootstrap here.  We do want to prevent the gRPC server from ever starting if envBootstrap fails; it would be weird to pass gRPC health checks for a while, and then when envBoostrap finally panics, shut down the server.  (We also don't want to handle any RPCs before auth is either activated, or we fail to activate it.  Seems like it would be hard to debug that that is happening.)

jrockway (commented):
I like this a lot.  I want to better understand what's going on with envBootstrap's timing before approving this, however.

src/server/license/server/api_server.go
```diff
@@ -35,11 +38,68 @@ func New(env Env) (lc.APIServer, error) {
 		env:     env,
 		license: licenseCollection(env.DB, env.Listener),
 	}
+	s.envBootstrap(context.Background())
```
jrockway, line 41: collect this error and return it

jrockway (approved):
I feel better with the activation being done synchronously.  Do get the error from envBootstrap and return it, but this looks good otherwise.

src/server/license/server/api_server.go
```diff
@@ -30,16 +33,79 @@ type apiServer struct {
 }
 
 // New returns an implementation of license.APIServer.
-func New(env Env) (lc.APIServer, error) {
+func New(env Env, public bool) (lc.APIServer, error) {
 	s := &apiServer{
 		env:     env,
 		license: licenseCollection(env.DB, env.Listener),
 	}
+	// envBootstrap runs concurrently because it reaches out to the enterprise service which heartbeats back to the license service. Therefore it must return to allow its service endpoints to be accessible
```
acohen4, line 41: cc @jrockway

jrockway: I'm still not convinced that `go` is an explicit enough scheduling directive.  I think you have to make bootstrap public and call it from main, after the server is ready to serve.

src/server/license/server/api_server.go
```diff
@@ -30,16 +33,79 @@ type apiServer struct {
...
+		_, err = a.env.EnterpriseServer.Activate(ctx, &ec.ActivateRequest{
+			Id:            localhostEnterpriseClusterId,
+			LicenseServer: localhostPeerAddr,
+			Secret:        a.env.Config.EnterpriseSecret})
+		if err != nil {
+			return errors.EnsureStack(err)
+		}
+		return nil
+	}(); err != nil {
+		panic(fmt.Sprintf("failed to configure the License Server from the environment: %v", err))
```
jrockway, line 94: Do we still want this to panic?  This can probably be de-indented and the individual things above can return errors, and where you write `go envBoostrap()` can instead be `go func() { err := envBootstrap(); if err != nil { panic(err) }`.

jrockway (commented):
I don't know how much of this we should fix in this PR, but I think it's pretty unusual that we can't fully construct our own API server unless another server on another computer can send us requests.  I can see why pachd has a dependency on the enterprise server, but I don't see why the enterprise server has a dependency on pachd (which is what prevents us from running envBootstrap at `New` time).

I think most cases, we aren't really waiting for a remote server to call us, but rather we use RPCs internally so that the various components of pachd aren't coupled at the go code level.  We should try and avoid that because it's interfering with our ability to implement features we want to implement.  (I kind of know why we do it that way; because each API server depends on a ServiceEnv, and if you wanted to call something without going through the server, you'd have to have the ServiceEnv build these objects and retain a reference for the other components, and you can't do that because it would be a circular package dependency.  That is also something we have to fix at some point; the ServiceEnv and API servers should all be in the same package.)

src/server/license/server/api_server.go
```diff
@@ -29,17 +32,76 @@ type apiServer struct {
...
+		if err != nil {
+			if errors.As(err, lc.ErrDuplicateClusterID) {
+				_, err = a.UpdateCluster(ctx, &lc.UpdateClusterRequest{
+					Id:          localhostEnterpriseClusterId,
+					Address:     localhostPeerAddr,
+					UserAddress: localhostPeerAddr,
+					Secret:      a.env.Config.EnterpriseSecret,
+				})
+				if err != nil {
+					return errors.Wrapf(err, "failed to update localhost cluster in the license service")
```
jrockway, line 75: One last nitpicky comment.  I think it's fine to omit the "failed" part from error constructors.  If it succeeded, it wouldn't be an Error object.  So this could be errors.Wrapf(err, "update localhost cluster").  (Ideally whatever handles the returned error would add "bootstrap license service".)

src/server/cmd/pachd/main.go
```diff
@@ -980,6 +988,11 @@ func doFullMode(config interface{}) (retErr error) {
 		g.Wait()
 		log.Println("gRPC server gracefully stopped")
 	}(interruptChan)
+	for _, b := range bootstrappers {
+		if err := b(context.Background()); err != nil {
```
jrockway, line 992: It would be nice if we knew and printed which bootstrapper failed here.  Maybe the []bootstrapFunc could be map[string]bootstrapFunc instead?  (Do we care about order?)

etc/helm/pachyderm/templates/pachd/license-secret.yaml
```diff
@@ -0,0 +1,13 @@
...
+*/ -}}
+{{- if and (not .Values.pachd.enterpriseLicenseKeySecretName) .Values.pachd.enterpriseLicenseKey }} 
+apiVersion: v1
+kind: Secret
+metadata:
+  name: pachyderm-license
+  namespace: {{ .Release.Namespace }}
+stringData:
+  enterprise-license-key: {{ .Values.pachd.enterpriseLicenseKey }}
+{{- end }}
```
jrockway, line 13: I think this is a nice bootstrapping experience.  You can deploy manually, move this secret to your secrets manager, and then remove the license key from the values, and check that in safely.

jrockway (approved):
I like this design.  Very explicit about the need for bootstrapping.

PR #7682, opened 2022-05-26
[FREQQ-105] Emit JSON logs by default, for GCP

msteffen:
This was requested as part of an oncall issue. See https://pachyderm.slack.com/archives/CSELV66P5/p1653413270247859 for context, as well as the referenced issue in the title.

#7683 makes a similar change in the 2.2.x branch, but leaves `text` as the default logging format. By making `json` the default only in master, I hope to give everyone lots of time to adjust to the new format and switch the default back to `text` if appropriate.

### DIFF
--- a/src/internal/log/log.go
+++ b/src/internal/log/log.go
@@ -55,7 +55,18 @@ func Pretty(entry *logrus.Entry) ([]byte, error) {
 	return serialized, nil
 }
 
-var jsonFormatter = &logrus.JSONFormatter{}
+var jsonFormatter = &logrus.JSONFormatter{
+	// Use GCP's field name conventions (absent any better alternative)
+	// https://cloud.google.com/logging/docs/agent/logging/configuration
+	FieldMap: logrus.FieldMap{
+		logrus.FieldKeyTime:  "time",
+		logrus.FieldKeyLevel: "severity",
+		logrus.FieldKeyMsg:   "message",
+	},
+
+	// https://github.com/sirupsen/logrus/pull/162/files
+	TimestampFormat: time.RFC3339Nano,
+}
 
 // JSONPretty is similar to Pretty() above, but it formats logrus log entries as
 // valid JSON objects. This is required by GCP (or else it misunderstands
--- a/src/internal/log/log.go
+++ b/src/internal/log/log.go
@@ -20,6 +20,20 @@ func (f FormatterFunc) Format(entry *logrus.Entry) ([]byte, error) {
 	return f(entry)
 }
 
+// formatServiceAndDuration factors out simple formatting code from Pretty() and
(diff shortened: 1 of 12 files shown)


src/internal/log/log.go
```diff
@@ -55,6 +55,22 @@ func Pretty(entry *logrus.Entry) ([]byte, error) {
 	return serialized, nil
 }
 
+var jsonFormatter = &logrus.JSONFormatter{}
+
+// JSONPretty is similar to Pretty() above, but it formats logrus log entries as
+// valid JSON objects. This is required by GCP (or else it misunderstands
+// severity and treats all log lines as ERROR logs).
+func JSONPretty(entry *logrus.Entry) ([]byte, error) {
```
jrockway, line 63: Not sure what field names JSONFormatter uses out of the box, but Google wants "severity", "message", "time" (in RFC3339).  To solve this particular customer's issue, the log level being "severity" is probably the most important.

If we want to go crazy, here's everything they support: https://cloud.google.com/logging/docs/agent/logging/configuration

chainlink:
I think @seslattery did some work on this at one point.

src/internal/log/log.go
seslattery:
```go
	log.SetFormatter(&log.JSONFormatter{
		FieldMap: log.FieldMap{
			log.FieldKeyTime:  "time",
			log.FieldKeyLevel: "severity",
			log.FieldKeyMsg:   "message",
		},
		// https://github.com/sirupsen/logrus/pull/162/files
		TimestampFormat: time.RFC3339Nano,
	})
```
is how hub and helium do it

msteffen: Done—thanks for the tip!

src/server/cmd/pachd/main.go
```diff
@@ -79,8 +78,7 @@ func init() {
 }
 
 func main() {
-	log.SetFormatter(logutil.FormatterFunc(logutil.Pretty))
-	maxprocs.Set(maxprocs.Logger(log.Printf))
```
jrockway, line 83: We shouldn't delete the call to maxprocs.Set.  This copies k8s limits on the container over to GOMAXPROCS, which is important for reasonable latency.

jrockway (approved):
This looks good, but add back the call to `maxprocs.Set`.

src/server/cmd/pachd/main.go
msteffen: oops, thanks!

msteffen:
It looks like the Netlify checks failed at the "preparing repo" stage. As this PR makes no docs changes, and the failure seems to have occurred prior to even inspecting the contents of this PR, and prior Netlify checks have passed, I'm going to attribute the errors to flakiness in Neflify and merge this PR.

PR #7685, opened 2022-05-27
Many Loki improvements

jrockway:
This PR is Quite The Thing.  I can't vouch for the sanity of the changes because I may have consumed some alcohol and listened to the same 5 seconds of a song on repeat while writing this.  Note to self: delete this tomorrow morning.

With that in mind...

1. I noticed while reviewing loki logs in a debug dump that they weren't separated by newlines.  This adds a newline between lines.  
2. I saw a TODO in the code that we should really pass a context to the Loki client.  Good idea.  Refactored to do that.  (I went a little overboard here; relying on explicit contexts instead of the one contained by server.env.pachClient.  We have to use the context from pachClient.WithContext(ctx).Ctx() in order to pass auth credentials internally, so that still gets a context, but we pass an explicit context around as a general best practice; I didn't want to do different things in different methods, and I didn't want to use pachClient as a budget context.Context.)  I tested Control-C-ing pachctl debug dump, and it cancels everything now.  No goroutine leaks while a user is trying to debug stuff!
4. Did you see the parenthetical remark in that last one about auth?  Yeah, I know that because I didn't do that first.  I noticed some access denied errors in the logs and wanted to see where they were coming from, so I made all access denied errors also log the client IP.  When it's 127.0.0.1 you know you messed up.  (At some point in the future, people will probably want x-forwarded-for from Envoy or their ingress controller, instead of the IP of the ingress controller.  We can do that when someone asks.  Already VERY out of scope for this PR.)
5. I saw a TODO in the code that we should retrieve more than 5000 log lines from Loki.  Extremely good idea.  Now we get 30,000.  The code tried to do this before, but the loop conditions were wrong, and I fixed those, so now it works!  I extensively tested this; there is and was a bug where a container has stdout and stderr streams, logs may appear out of order if one chunk of 5000 logs has both streams in the same chunk.  -We should probably load all 30,000 lines into a slice and sort by time.  That felt horrible to me this evening so I didn't do it, but it's probably a max of 6MB of RAM per log file, so not a big deal.- We do the sorting now, and output in a new format that preserves the Loki stream information.
6. The original thing that prompted me to fix item number 1 (not to be confused with bug #1) was a confusing error message in an `error.txt` from Loki in the debug dump.  I made the Loki client return correct errors, instead of the error from trying to parse the plain-text error as JSON!  Instead of a JSON parse error, you'll get the plain text that the Loki server responded with now.
7. I noticed a while ago on hub that "pachctl logs" didn't work.  This is because we silently discard JSON parse errors from worker logs.   If they aren't valid JSON, we throw them away without even warning about it.  The JSON parse errors are actually a change between Docker and containerd/cri-o container runtimes; Loki expects the raw log file on the host to be in Docker format (configurable!), but if you're using containerd/cri-o, then they're not, and you don't actually get JSON from loki.  Because our code assumes that every record returned from Loki is valid JSON, if you use containerd "pachctl logs" basically stops working.  I estimate this affects 100% of users, because Docker mysteriously disappeared from Kubernetes and was replaced with containerd or cri-o, but nobody edited their promtail configuration to account for this.  Users can of course fix this log parsing (see Slack thread), but I know they won't, so I adjusted the Loki log parsing to handle all 3 possible formats (native/correct promtail config, incorrectly configured Docker logs, incorrectly configured containerd/cri-o logs).  I also now warn when we delete a malformed line, so at least you can look at pachd logs and see that we're throwing away your worker logs.
8. Speaking of malformed lines, we initialized prometheus in the workers before the log format has been set to JSON, but logged a line saying that we're starting up prometheus.  The log format thing turns out to be ... not fixable (circular dependencies) ... so I just deleted that log line to avoid a spurious log parsing error.
9. I added a test to ensure that we get the 30,000 most recent logs, instead of the 30,000 least recent logs.
10. While working on the profiling stuff, I noticed that profile/binary dumps also contain suite=pachyderm logs.  I fixed that issue here.

The new log format looks something like this; the logs are sorted by time, and whenever the stream info changes, we emit a header that captures all the information Loki returned:
```
&map[app:pg-bouncer container:pg-bouncer filename:/var/log/pods/default_pg-bouncer-66d8b86c4-zwhhk_d954f97d-d92e-4fbb-a626-7b42a428cc8a/pg-bouncer/0.log job:default/pg-bouncer namespace:default pod:pg-bouncer-66d8b86c4-zwhhk pod_template_hash:66d8b86c4 suite:pachyderm]
2022-05-26T20:01:05.667064575Z stderr F pgbouncer 20:01:05.66
2022-05-26T20:01:05.668233301Z stderr F pgbouncer 20:01:05.66 Welcome to the Bitnami pgbouncer container
...
2022-05-26T21:12:22.988920326Z stderr F 2022-05-26 21:12:22.988 UTC [1] LOG stats: 25 xacts/s, 45 queries/s, in 5684 B/s, out 11994 B/s, xact 520 us, query 215 us, wait 0 us
&map[app:pg-bouncer container:pg-bouncer filename:/var/log/pods/default_pg-bouncer-66d8b86c4-zwhhk_d954f97d-d92e-4fbb-a626-7b42a428cc8a/pg-bouncer/0.log job:default/pg-bouncer namespace:default pod:pg-bouncer-66d8b86c4-zwhhk pod_template_hash:66d8b86c4 stream:stderr suite:pachyderm]
2022-05-26 21:13:22.987 UTC [1] LOG stats: 25 xacts/s, 45 queries/s, in 5691 B/s, out 11997 B/s, xact 515 us, query 213 us, wait 0 us
2022-05-26 21:13:49.380 UTC [1] LOG C-0x55e6ebf012d0: pachyderm/pachyderm@10.244.0.20:39982 login attempt: db=pachyderm user=pachyderm tls=no
...
```
(In this example, it's because Loki was reconfigured to properly retain the stream name; but this basically happens most often when an app logs to both stdout and stderr.  pachyderm-proxy does that, for example.)

The stream headers always start with '&map[', so you can grep those out.

-We should open a ticket to write some tests for the debug dump that ensure 100% of the logs are in the loki-logs.txt file.  I think we should just run a pipeline, print some stuff (30,000 lines of stuff, or 5001 lines at a minimum), and then see if it's in the debug dump.  That will allow people to edit this code without breaking it.  I might remember to do this.- I sort of did this.  It's not the real Loki, but at least we can add tests for the log-getting code now.

Good evening and good night.  This was fun.  I don't remember what I was supposed to do today, but this is what I did.

### DIFF
--- a/src/internal/lokiutil/client/client.go
+++ b/src/internal/lokiutil/client/client.go
@@ -3,7 +3,6 @@ package client
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"io"
 	"net/http"
 	"net/url"
@@ -78,9 +77,9 @@ func (c *Client) doRequest(ctx context.Context, path, query string, quiet bool, 
 	if resp.StatusCode >= http.StatusBadRequest {
 		body, err := io.ReadAll(resp.Body)
 		if err != nil {
-			body = append(body, []byte(fmt.Sprintf("\nreading body: %v", err))...)
+			return errors.EnsureStack(errors.Errorf("error response from loki: %v (body: %q); additionally, reading body: %v", resp.Status, body, err))
 		}
-		return errors.EnsureStack(errors.Errorf("error response from loki: %v (body: %s)", resp.Status, body))
+		return errors.EnsureStack(errors.Errorf("error response from loki: %v (body: %q)", resp.Status, body))
 	}
 	return errors.EnsureStack(json.NewDecoder(resp.Body).Decode(out))
 }
--- a/src/internal/lokiutil/client/client.go
+++ b/src/internal/lokiutil/client/client.go
@@ -3,6 +3,8 @@ package client
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"path"
@@ -71,8 +73,15 @@ func (c *Client) doRequest(ctx context.Context, path, query string, quiet bool, 
 	if err != nil {
 		return errors.EnsureStack(err)
 	}
-
 	defer resp.Body.Close()
+
+	if resp.StatusCode >= http.StatusBadRequest {
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			body = append(body, []byte(fmt.Sprintf("\nreading body: %v", err))...)
+		}
+		return errors.EnsureStack(errors.Errorf("error response from loki: %v (body: %s)", resp.Status, body))
+	}
 	return errors.EnsureStack(json.NewDecoder(resp.Body).Decode(out))
 }
 
--- a/src/internal/middleware/auth/interceptor.go
+++ b/src/internal/middleware/auth/interceptor.go
@@ -10,6 +10,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/peer"
 )
 
 // authHandlers is a mapping of RPCs to authorization levels required to access them.
@@ -279,6 +280,13 @@ type Interceptor struct {
 	getAuthServer func() authserver.APIServer
 }
 
+func peerName(ctx context.Context) string {
+	if p, ok := peer.FromContext(ctx); ok {
+		return p.Addr.String()
+	}
+	return "<unknown ip>"
+}
+
 // InterceptUnary applies authentication rules to unary RPCs
 func (i *Interceptor) InterceptUnary(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 	a, ok := authHandlers[info.FullMethod]
@@ -290,7 +298,7 @@ func (i *Interceptor) InterceptUnary(ctx context.Context, req interface{}, info 
 	username, err := a(ctx, i.getAuthServer(), info.FullMethod)
 
 	if err != nil {
-		logrus.WithError(err).Errorf("denied unary call %q to user %v\n", info.FullMethod, nameOrUnauthenticated(username))
+		logrus.WithError(err).Errorf("denied unary call %q to user %v@%v", info.FullMethod, nameOrUnauthenticated(username), peerName(ctx))
 		return nil, err
 	}
 
@@ -313,7 +321,7 @@ func (i *Interceptor) InterceptStream(srv interface{}, stream grpc.ServerStream,
 	username, err := a(ctx, i.getAuthServer(), info.FullMethod)
 
 	if err != nil {
-		logrus.WithError(err).Errorf("denied streaming call %q to user %v\n", info.FullMethod, nameOrUnauthenticated(username))
+		logrus.WithError(err).Errorf("denied streaming call %q to user %v@%v", info.FullMethod, nameOrUnauthenticated(username), peerName(ctx))
 		return err
 	}
 
--- a/src/server/debug/server/server.go
+++ b/src/server/debug/server/server.go
@@ -626,8 +626,13 @@ func (s *debugServer) collectLogsLoki(ctx context.Context, tw *tar.Writer, pod, 
 		}
 		queryStr += `"}`
 		return s.queryLoki(ctx, queryStr, func(_ loki.LabelSet, line string) error {
-			_, err := w.Write([]byte(line))
-			return errors.EnsureStack(err)
+			if _, err := w.Write([]byte(line)); err != nil {
+				return errors.EnsureStack(err)
+			}
+			if _, err := w.Write([]byte("\n")); err != nil {
+				return errors.EnsureStack(err)
+			}
+			return nil
 		})
 	}, prefix...)
 }
@@ -747,7 +752,7 @@ func (s *debugServer) queryLoki(ctx context.Context, queryStr string, cb func(lo
 					advancedStart = true
 					start = entry.Timestamp
 				}
-				if err := cb(stream.Labels, entry.Line+"\n"); err != nil {
+				if err := cb(stream.Labels, entry.Line); err != nil {
 					return err
 				}
 			}
--- a/src/server/debug/server/server.go
+++ b/src/server/debug/server/server.go
@@ -810,7 +810,6 @@ func (s *debugServer) queryLoki(ctx context.Context, queryStr string) ([]lokiLog
 		}
 
 		var readThisIteration int
-		var advancedTime bool
 		for _, s := range streams {
 			stream := s // Alias for pointer later.
 			for _, e := range stream.Entries {
@@ -818,7 +817,11 @@ func (s *debugServer) queryLoki(ctx context.Context, queryStr string) ([]lokiLog
 				numLogs++
 				readThisIteration++
 				if entry.Timestamp.Before(end) {
-					advancedTime = true
+					// Because end is an "exclusive" range, if we read any logs
+					// at all we are guaranteed to not read them in the next
+					// iteration.  (If all 5000 logs are at the same time, we
+					// still advance past them on the next iteration, which is
+					// the best we can do under those circumstances.)
 					end = entry.Timestamp
 				}
 				result = append(result, lokiLog{
@@ -826,14 +829,6 @@ func (s *debugServer) queryLoki(ctx context.Context, queryStr string) ([]lokiLog
 					Entry:  &entry,
 				})
 			}
-		}
-		if !advancedTime {
-			// This means we read a chunk of 5000 logs without the timestamp changing,
-			// probably because more than 5000 logs were logged in the same moment of
-			// time.  We can't do anything about this except step past it.  If we didn't
-			// move the end time, we'd just read the same chunk over and over until the
-			// loop ends because of numLogs.
-			end = end.Add(-time.Second)
 		}
 		if readThisIteration == 0 {
 			break
--- a/src/server/debug/server/server.go
+++ b/src/server/debug/server/server.go
@@ -56,10 +56,10 @@ func NewDebugServer(env serviceenv.ServiceEnv, name string, sidecarClient *clien
 	}
 }
 
-type collectPipelineFunc func(*tar.Writer, *pps.PipelineInfo, ...string) error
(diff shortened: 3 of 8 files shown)


src/internal/lokiutil/client/client.go
```diff
@@ -71,8 +73,15 @@ func (c *Client) doRequest(ctx context.Context, path, query string, quiet bool,
...
 	defer resp.Body.Close()
+
+	if resp.StatusCode >= http.StatusBadRequest {
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			body = append(body, []byte(fmt.Sprintf("\nreading body: %v", err))...)
+		}
+		return errors.EnsureStack(errors.Errorf("error response from loki: %v (body: %s)", resp.Status, body))
+	}
 	return errors.EnsureStack(json.NewDecoder(resp.Body).Decode(out))
```
robert-uhl, line 85: Rather than printing that the body contained `"\nreading body: …"` when it did not, why not this instead:

```go
	if resp.StatusCode >= http.StatusBadRequest {
		body, err := io.ReadAll(resp.Body)
		if err != nil {
			return errors.EnsureStack(errors.Errorf("error response from loki: %v (body: %q; err: %v)", resp.Status, body, err))
		}
		return errors.EnsureStack(errors.Errorf("error response from loki: %v (body: %q)", resp.Status, body))
	}
```

FWIW, I also quoted the body string, which might help future parsing.

src/internal/middleware/auth/interceptor.go
```diff
@@ -279,6 +280,13 @@ type Interceptor struct {
 	getAuthServer func() authserver.APIServer
 }
 
+func peerName(ctx context.Context) string {
```
robert-uhl, line 283: Seems a bit surprising that it is called `peerName` but may not actually return a peer name. Looking at how it’s used, it doesn’t make sense to complicate its signature; maybe rename to `peerNameOrUnknown`, `peerString` or something like that?

Or perhaps I am being too nitpicky …

src/server/debug/server/server.go
```diff
@@ -71,66 +71,67 @@ func (s *debugServer) handleRedirect(
 	redirect redirectFunc,
 	collect collectFunc,
 ) error {
+	ctx := pachClient.Ctx() // this context has authorization credentials we need
```
robert-uhl, line 74: It would be clearer to just add the context to `handleRedirect`’s signature, but perhaps that is better in a different PR?

src/server/debug/server/server.go
```diff
@@ -283,13 +286,13 @@ func (s *debugServer) Profile(request *debug.ProfileRequest, server debug.Debug_
...
-		redirectProfileFunc(pachClient.Ctx(), request.Profile),
+		redirectProfileFunc(request.Profile),
 		collectProfileFunc(request.Profile),
 	)
 }
 
 func collectProfileFunc(profile *debug.Profile) collectFunc {
-	return func(tw *tar.Writer, prefix ...string) error {
+	return func(_ context.Context, tw *tar.Writer, _ *client.APIClient, prefix ...string) error {
 		return collectProfile(tw, profile, prefix...)
```
robert-uhl, line 296: `collectProfile` calls `debugProfile` with a callback which calls `writeProfile`, which calls `time.Sleep`; it makes sense to propagate the context down the chain so that `writeProfile` can terminate the sleep early if the context is cancelled.

src/server/debug/server/server.go
```diff
@@ -367,14 +370,12 @@ func collectBinary(tw *tar.Writer, prefix ...string) error {
...
 }
 
-func redirectBinaryFunc(ctx context.Context) redirectFunc {
-	return func(c debug.DebugClient, filter *debug.Filter) (io.Reader, error) {
-		binaryC, err := c.Binary(ctx, &debug.BinaryRequest{Filter: filter})
-		if err != nil {
-			return nil, errors.EnsureStack(err)
-		}
-		return grpcutil.NewStreamingBytesReader(binaryC, nil), nil
+func redirectBinaryFunc(ctx context.Context, c debug.DebugClient, filter *debug.Filter) (io.Reader, error) {
```
robert-uhl, line 373: Since it no longer returns a function, it should probably be renamed to `redirectBinary`.

src/server/debug/server/server.go
```diff
@@ -430,15 +431,15 @@ func (s *debugServer) collectInputRepos(tw *tar.Writer, pachClient *client.APICl
...
+				return s.collectCommits(ctx, tw, pachClient, repoInfo.Repo, limit, repoPrefix)
 			}
 			return err
 		}
 	}
 	return nil
 }
 
-func (s *debugServer) collectCommits(tw *tar.Writer, pachClient *client.APIClient, repo *pfs.Repo, limit int64, prefix ...string) error {
+func (s *debugServer) collectCommits(rctx context.Context, tw *tar.Writer, pachClient *client.APIClient, repo *pfs.Repo, limit int64, prefix ...string) error {
```
robert-uhl, line 442: `rctx`?

src/server/debug/server/server.go
```diff
@@ -461,7 +462,7 @@ func (s *debugServer) collectCommits(tw *tar.Writer, pachClient *client.APIClien
 		},
 	}
 	if err := collectDebugFile(tw, "commits", "json", func(w io.Writer) error {
-		ctx, cancel := context.WithCancel(pachClient.Ctx())
+		ctx, cancel := context.WithCancel(rctx)
```
robert-uhl, line 465: It’s a bit surprising to make the context cancellable this deep in the call stack, especially when it doesn’t appear to be spawning any background tasks which might need cancellation.

I think that this could just be the enclosing context. Could either close over it, or add to the signature of `collectDebugFile` and its callback.

src/server/debug/server/server.go
```diff
@@ -713,43 +714,59 @@ func (s *debugServer) getWorkerPodsLoki(pipelineInfo *pps.PipelineInfo) (map[str
...
 		for _, stream := range streams {
 			for _, entry := range stream.Entries {
 				numLogs++
+				readThisIteration++
 				if entry.Timestamp.After(start) {
+					advancedStart = true
 					start = entry.Timestamp
 				}
-				if err := cb(stream.Labels, entry.Line); err != nil {
+				if err := cb(stream.Labels, entry.Line+"\n"); err != nil {
```
robert-uhl, line 750: Rather than `queryLoki` adding a newline to every line, I think that the responsibility should be on the callback to handle the line appropriately, e.g. quoting it in JSON or writing terminated with a newline in a log file, or whatever.

Looking at where `queryLoki` is called, it looks like `collectLogsLoki` is the one place that actually uses the line. Rather than `w.Write([]byte(line))`, it could just call `fmt.Fprintln(w, line)`.

src/server/pps/server/api_server.go
```diff
@@ -1395,6 +1397,89 @@ func (a *apiServer) getLogsLoki(request *pps.GetLogsRequest, apiGetLogsServer pp
...
 }
 
+func parseLokiLine(inputLine string, msg *pps.LogMessage) error {
+	// There are three possible log formats in Loki, depending on the underlying formatting done
+	// by the container runtime.  Under ideal circumstances, the user of Loki has configured
+	// promtail to remove the runtime-specific logs, but it's quite easy to miss this
+	// configuration aspect, so we correct for it here.  (See:
+	// https://grafana.com/docs/loki/latest/clients/promtail/stages/cri/ and
+	// https://grafana.com/docs/loki/latest/clients/promtail/stages/docker/)
+
```
robert-uhl, line 1407: Rather than closures, why not make all three top-level private functions with a signature `func (string) (*pps.LogMessage, error)`?

robert-uhl (commented):
I _really_ like that you plumbed through the context. It’s a pity that they lengthen function signatures, but they provide some really nice benefits.

src/server/debug/server/server.go
jrockway: I had a variant of that, but it ends up being tricky... there are some calls made to pachd with the thick client, in which case it uses pachClient.Ctx(), and some calls made directly through the autogenerated grpc stubs, which need that exactly context (not server.Context(), which is missing grpc creds needed to make internal calls in the context metadata!)

What this means is that in the 3 callers of this method, they'd have to create a pachClient and put the context in, and then take the context out and pass it to this method.  Seems clearer to only pass the pachClient, so that it looks to the caller like we're doing something weird, which we are.

src/server/debug/server/server.go
jrockway: root context, so ctx doesn't shadow itself when we create the cancelable context.  I think if you call the context you don't want to ever use "ctx", it's easy to accidentally use it and be confused as to why cancelation doesn't work; that's why I didn't call this one "ctx" and the other one "cctx".

src/internal/lokiutil/client/client.go
jrockway: Done.

src/server/debug/server/server.go
jrockway: Done.

src/server/debug/server/server.go
jrockway: It depends on what happens when ForEachCommit returns early with an error; I don't see it shutting down the stream and I think that's what this is trying to achieve.

src/server/debug/server/server.go
jrockway: Done.

jrockway:
PTAL (next week!)

src/server/debug/server/server.go
robert-uhl: > I didn't write this code, and it works, so I'm not too interested in changing it here.

That’s fair!

src/server/debug/server/server.go
```diff
@@ -313,7 +325,14 @@ func writeProfile(w io.Writer, profile *debug.Profile) error {
 				return errors.EnsureStack(err)
 			}
 		}
-		time.Sleep(duration)
+
+		// Wait for either the defined duration, or until the context is done.  We'll eat
+		// the context cancelled / deadline exceeded error here, but I think that's fine
+		// since we don't do any other IO in this function.  (The writer is going to be
```
robert-uhl, line 331: Should that be ‘isn’t’?

robert-uhl (approved):
I am fine addressing the four outstanding items above in a follow-on PR; this is worth getting in as-is.

PR #7686, opened 2022-05-27
[CORE-467] add new mux for oidc and metrics

armaanv:
This prevents the prometheus handler from being served on the OIDC port and vice versa. 

Running `curl -s localhost:30657/metrics` while port forwarding now returns a 404


jrockway (approved):
Looks good.  A nice improvement would be to serve 200 on / instead of 404.  Maybe a message like "pachd oidc ok", and "pachd metrics ok, metrics at /metrics".  We might have a separate issue for that (related to health checks).

armaanv:
> Looks good. A nice improvement would be to serve 200 on / instead of 404. Maybe a message like "pachd oidc ok", and "pachd metrics ok, metrics at /metrics". We might have a separate issue for that (related to health checks).

Yup, I addressed that in https://github.com/pachyderm/pachyderm/pull/7634. I'll rebase that on top of this commit after I merge to master

PR #7687, opened 2022-05-30
Try running CI without awscli

chainlink:
Shaves 1m off the install.sh portion of the test (which gets run on every bucket)


molinamelendezj (approved):
Good find!

PR #7689, opened 2022-05-30
Don't build kafka container before every PPS test run

chainlink:


molinamelendezj (approved):
You definitely finding some good ones. Why was Kafka there to being with!

PR #7694, opened 2022-05-31
[CORE-606] Improve Loki defaults

robert-uhl:
This PR improves the Loki setup:

- It is explicit about retention periods for Pachyderm (1 week) and non-Pachyderm (1 day) pods.
- It increases the default persistence PV.
- It adds support for using Loki and object storage when deploying to GCP and Azure.


etc/deploy/azure/azure-doco-script.sh
```diff
@@ -149,7 +154,9 @@ pachd:
     annotations:
       service.beta.kubernetes.io/azure-load-balancer-resource-group: ${RESOURCE_GROUP}
   image:
-    tag: "2.1.1"
```
PFedak, line 152: hmm, are these supposed to be regularly updated? Should this actually be blank, to inherit from whatever the actual current chart is? This makes me wonder how customers actually construct their values.yaml

etc/deploy/azure/azure-doco-script.sh
```diff
@@ -168,7 +175,54 @@ global:
...
+      ksLi4xaNmjICq44Y3ekQEe5+NauQrz4wlHrQMz2nZQ/1/I6eYs9HRCwBXbsdtTLS
+      R9I4LtD+gdwyah617jzV/OeBHRnDJELqYzmp
+      -----END CERTIFICATE-----
+
+loki-stack:
+  loki:
+    config:
+      schema_config:
+        configs:
+        - from: 1989-11-09
```
PFedak, line 207: 🇩🇪

etc/helm/pachyderm/values.yaml
```diff
@@ -208,11 +208,18 @@ loki-stack:
...
       annotations: {}
       nodeSelector: {}
       tolerations: []
+    config:
+      limits_config:
+        retention_period: 24h
+        retention_stream:
+          - selector: '{suite="pachd"}'
+            priority: 1
+            period: 168h # = 1 day
```
PFedak, line 222: I assume you mean 1 week. It looks like YAML supports `1d` and `1w`, which might be preferable

etc/deploy/gcp/gcp-doco-script.sh
```diff
@@ -23,14 +23,17 @@ SQL_MEM="7680MB"
...
 CLOUDSQL_INSTANCE_NAME="${NAME}-sql"
 GSA_NAME="${NAME}-gsa"
+LOKI_GSA_NAME="${NAME}-loki-gsa"
 STATIC_IP_NAME="${NAME}-ip"
 
 ROLE1="roles/cloudsql.client"
 ROLE2="roles/storage.admin"
 
 SERVICE_ACCOUNT="${GSA_NAME}@${PROJECT_ID}.iam.gserviceaccount.com"
+LOKI_SERVICE_ACCOUNT="${GSA_NAME}@${PROJECT_ID}.iam.gserviceaccount.com"
```
PFedak, line 36: should this be using `LOKI_GSA_NAME`, or are they sharing a service account?

etc/deploy/gcp/gcp-doco-script.sh
```diff
@@ -106,7 +110,17 @@ gcloud iam service-accounts add-iam-policy-binding ${SERVICE_ACCOUNT} \
...
-gcloud sql databases create pachyderm -i ${CLOUDSQL_INSTANCE_NAME}
+gcloud iam service-accounts create ${LOKI_GSA_NAME}
+
+gcloud projects add-iam-policy-binding ${PROJECT_ID} \
+    --member="serviceAccount:${LOKI_SERVICE_ACCOUNT}" \
+    --role="${ROLE2}"
+
+gcloud iam service-accounts keys create "${LOKI_GSA_NAME}-key.json" --iam-account="$LOKI_SERVICE_ACCOUNT"
+
+kubectl create secret generic loki-service-account --from-file="${LOKI_GSA_NAME}-key.json"
```
PFedak, line 121: What exactly is this secret protecting? Is this how we give pachd object storage permissions, too, or do we somehow derive it elsewhere?

PFedak (commented):
I think this makes sense, but what I'm not entirely clear on is the flow of how users actually incorporate this into their own charts. I was kind of hoping there would be a way for loki to use the deployment type to set itself up properly, but this way anyone off the beaten path (no clue how many customers that is) has to at best copy this config over, and at worst repeat some version of the work you did here for a feature that they don't care about.

I didn't hit approve mainly because I think you should have someone else with a better feel for our deploy scripts/process double check things here.

etc/deploy/azure/azure-doco-script.sh
```diff
@@ -168,7 +175,54 @@ global:
...
+      dc93Uk3zyZAsuT3lySNTPx8kmCFcB5kpvcY67Oduhjprl3RjM71oGDHweI12v/ye
+      jl0qhqdNkNwnGjkCAwEAAaNFMEMwHQYDVR0OBBYEFOWdWTCCR1jMrPoIVDaGezq1
+      BE3wMBIGA1UdEwEB/wQIMAYBAf8CAQMwDgYDVR0PAQH/BAQDAgEGMA0GCSqGSIb3
+      DQEBBQUAA4IBAQCFDF2O5G9RaEIFoN27TyclhAO992T9Ldcw46QQF+vaKSm2eT92
+      9hkTI7gQCvlYpNRhcL0EYWoSihfVCr3FvDB81ukMJY2GQE/szKN+OMY3EU/t3Wgx
+      jkzSswF07r51XgdIGn9w/xZchMB5hbgF/X++ZRGjD8ACtPhSNzkE1akxehi/oCr0
+      Epn3o0WC4zxe9Z2etciefC7IpJ5OCBRLbf1wbWsaY71k5h+3zvDyny67G7fyUIhz
+      ksLi4xaNmjICq44Y3ekQEe5+NauQrz4wlHrQMz2nZQ/1/I6eYs9HRCwBXbsdtTLS
+      R9I4LtD+gdwyah617jzV/OeBHRnDJELqYzmp
+      -----END CERTIFICATE-----
```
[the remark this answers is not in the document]
robert-uhl: Yup, this is the certificate for the CA which issues the Azure PostgreSQL servers their certificates — it provides the root of trust pgbouncer needs. Without it, it cannot verify the server and TLS fails.

etc/helm/pachyderm/values.yaml
robert-uhl: > It looks like YAML supports 1d and 1w, which might be preferable

The [Loki docs](https://grafana.com/docs/loki/latest/configuration/#generic-placeholders) indicate that durations must match this regex: `[0-9]+(ns|us|µs|ms|[smh])`; units can only be nanoseconds, microseconds, milliseconds, seconds minutes and hours.

You are completely correct about 1 week — my fault. Fixed in latest.

etc/deploy/gcp/gcp-doco-script.sh
robert-uhl: It is used to give the Loki pod permission to write to object storage:

```yaml
loki-stack:
  loki:
    env:
    - name: GOOGLE_APPLICATION_CREDENTIALS
      value: /etc/secrets/${LOKI_GSA_NAME}-key.json
    extraVolumes:
      - name: loki-service-account
        secret:
          secretName: loki-service-account
    extraVolumeMounts:
      - name: loki-service-account
        mountPath: /etc/secrets
```

etc/deploy/azure/azure-doco-script.sh
robert-uhl: Needed a date in the past, figured that was as good as any☺

etc/deploy/azure/azure-doco-script.sh
robert-uhl: So, the short answer is that it has always been explicit, and no-one has ever changed that.

I _think_ it makes sense to keep it explicit, because it means that the script should always create a proper cluster. The alternative would be to test the script with each new version. That requires some work: I’ll create a story for it.

seslattery (approved):
I think this looks good.  I'm a little hesitant about requiring a second cloud storage bucket for loki if we're making it the default, but that's outside the scope of this PR.

PR #7697, opened 2022-05-31
[CORE-606] Upgrade loki-stack

robert-uhl:
This is a precondition for CORE-606, not the only PR for the story.


src/server/pps/server/api_server.go
```diff
@@ -1315,7 +1315,7 @@ func (a *apiServer) getLogsLoki(request *pps.GetLogsRequest, apiGetLogsServer pp
 		if err := a.env.AuthServer.CheckClusterIsAuthorized(apiGetLogsServer.Context(), auth.Permission_CLUSTER_GET_PACHD_LOGS); err != nil {
 			return errors.EnsureStack(err)
 		}
-		return lokiutil.QueryRange(apiGetLogsServer.Context(), loki, `{app="pachd"}`, time.Now().Add(-since), time.Now(), request.Follow, func(t time.Time, line string) error {
+		return lokiutil.QueryRange(apiGetLogsServer.Context(), loki, `{app="pachd"}`, time.Now().Add(-since), time.Time{}, request.Follow, func(t time.Time, line string) error {
```
acohen4, line 1318: what's up with this?

src/internal/minikubetestenv/deploy.go
```diff
@@ -251,7 +253,7 @@ func union(a, b *helm.Options) *helm.Options {
 }
 
 func waitForPachd(t testing.TB, ctx context.Context, kubeClient *kube.Clientset, namespace, version string) {
-	require.NoError(t, backoff.Retry(func() error {
+	require.NoErrorWithinTRetry(t, 5*time.Minute, func() error {
```
acohen4, line 256: only difference here is that the backoff is exponential. This is probably fine

src/internal/lokiutil/lokiutil.go
```diff
@@ -57,7 +58,16 @@ func QueryRange(ctx context.Context, c *loki.Client, queryStr string, from, thro
 		if err := forEachLine(*resp, func(t time.Time, line string) error {
 			from = t
 			nMsgs++
-			return f(t, line)
+			var entry struct {
```
acohen4, line 61: what's the backstory here?

etc/helm/pachyderm/Chart.yaml
```diff
@@ -21,7 +21,7 @@ type: application
 # each time you make changes to the chart and its templates, including
 # the app version.  Versions are expected to follow Semantic
 # Versioning (https://semver.org/)
-version: 2.1.0
+version: 2.1.1
```
acohen4, line 24: is this necessary? I think we also want it to match appVersion

etc/helm/pachyderm/values.yaml
```diff
@@ -215,6 +215,55 @@ loki-stack:
       tolerations: []
   grafana:
     enabled: false
+  promtail:
+    config:
+      lokiAddress: http://{{ .Release.Name }}-loki:3100/loki/api/v1/push
```
acohen4, line 220: I didn't realize we can resolve substitutions in values.yaml

src/server/pps/server/api_server.go
robert-uhl: What’s really being requested here is the last `since` duration of logs. And when `request.Follow` was true, it wasn’t actually following: it requested a start time of t - since and an end time of t, then a start time of t - since + 1 ns and an end time of t, then a start time of t - since + 2 ns and an end time of t — the end time never increased. And really, for follow mode we don’t _want_ it anyway.

I briefly deleted the argument entirely, since it was never actually used for anything but `time.Now()`, but then decided it was probably better to leave it in but DTRT with a zero value, much like the Loki API itself.

src/internal/lokiutil/lokiutil.go
robert-uhl: So, it looks like in the older version Loki just returned the log line; with the newer version of Loki we get back a structured line which contains the log line as a string. This code parses it out.

etc/helm/pachyderm/Chart.yaml
```diff
@@ -52,15 +52,15 @@ sources:
...
 
-# Note this chart has one diverged change: 
+# Note this chart has one diverged change:
 # Adding a label variable to the volumeClaimTemplate in the Postgres Stateful set
 # TODO: Upstream this change
 dependencies:
 - name: postgresql
   version: "10.8.0"
-  repository: "file://../charts/postgresql"
+  repository: "file://./dependencies/postgresql"
```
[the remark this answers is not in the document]
robert-uhl: `helm dependency` wants to manage `charts/` itself, so I moved the bits we manage to a different directory. It fails to run correctly otherwise.

etc/helm/pachyderm/Chart.yaml
robert-uhl: Yup! ‘This version number should be incremented each time you make changes to the chart and its templates, including the app version.’ This change updates the chart, so the chart version needs to be incremented.

The chart version and the app version are two different things, both monotonically increasing. Increasing the app version necessitates increasing the chart version, but not vice-versa. E.g. correcting a buggy template does not necessarily mean a new release of Pachyderm itself (does it?).

etc/helm/pachyderm/values.yaml
robert-uhl: I stole it from https://github.com/grafana/helm-charts/blob/main/charts/loki-stack/values.yaml#L8

etc/helm/pachyderm/Chart.yaml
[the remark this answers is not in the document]
acohen4: shouldn't helm manage loki-stack though?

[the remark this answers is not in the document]
robert-uhl: Ah, I see what you mean … need to delete the 2.5.1 tarball. Done!

etc/helm/pachyderm/Chart.yaml
acohen4: it's something we do at release time. You shouldn't need to increment it in the PR. Basically we keep version=appVersion. So every release increments this. Ex: https://github.com/pachyderm/pachyderm/blob/v2.2.1/etc/helm/pachyderm/Chart.yaml

robert-uhl: Righto! Reverted.

src/server/debug/server/server.go
```diff
@@ -794,12 +794,11 @@ func (s *debugServer) queryLoki(ctx context.Context, queryStr string) ([]lokiLog
...
 	var result []lokiLog
 
-	// FIXME: run once without an end, then work backwards
-	end := time.Now().Add(time.Minute)            // Account for some clock skew between our node and the Loki node.
+	var end time.Time
 	start := time.Now().Add(-30 * 24 * time.Hour) // 30 days.  (Loki maximum range is 30 days + 1 hour.)
 
-	for numLogs := 0; start.Before(end) && numLogs < maxLogs; {
-		resp, err := c.QueryRange(ctx, queryStr, serverMaxLogs, start, time.Time{}, "BACKWARD", 0, 0, true)
+	for numLogs := 0; end.IsZero() || start.Before(end) && numLogs < maxLogs; {
```
acohen4, line 800: should this be `(end.IsZero() || start.Before(end)) && numLogs < maxLogs`?

robert-uhl: Yes, yes it should be. Fixing.

src/server/debug/server/server.go
```diff
@@ -794,12 +794,11 @@ func (s *debugServer) queryLoki(ctx context.Context, queryStr string) ([]lokiLog
 	// of providing lines in chronological order even when the app uses both stdout and stderr.
 	var result []lokiLog
 
-	// FIXME: run once without an end, then work backwards
```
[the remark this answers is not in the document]
robert-uhl: No, I added that comment in a previous commit in this PR, then removed it after I implemented the right behaviour.

PR #7704, opened 2022-06-01
CORE-695 Add logging response transformer to omit activation code from GetState

FahadBSyed:
## Description
This pull request removes the `ActivationCode` when `pachd` logs `GetState` responses. Previously, `pachd` would log a 'sanitized' version of the `ActivationCode` which did not include any sensitive data. However, the team decide its better to remove the field from the logs altogether to avoid confusion and create some peace of mind.


acohen4 (approved):
LGTM! I think ideally the .gitignore change can be its own PR, but not a hard rule

PR #7717, opened 2022-06-02
Consult RC when un-crashing pipeline.

PFedak:
The missing secrets test was flaking due to the crashing monitor taking
it out of the crashing state prematurely. The logic was wrong in both
directions: it required the maximum number of pods to be present, even
if the RC requested fewer, and could include workers in its count which
were going to be terminated.

While it seems like checking the RC's status should be sufficient, I've
left the worker status check in place to capture e.g. a failure in
networking in the worker service which the worker pod itself doesn't
care about.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,7 +247,7 @@ test-pfs-server:
 test-pps: launch-stats docker-build-spout-test
 	@# Use the count flag to disable test caching for this test suite.
 	PROM_PORT=$$(kubectl --namespace=monitoring get svc/prometheus -o json | jq -r .spec.ports[0].nodePort) \
-	  go test -v -count=1 ./src/server -parallel $(PARALLELISM) -timeout $(TIMEOUT) $(RUN) $(TESTFLAGS)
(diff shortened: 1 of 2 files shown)


jrockway (approved):
This seems reasonable but I'm not 100% sure.  There are at least 3 things running concurrently that can mutate the state.  pollPipelinePods can transition from anything -> CRASHING.  monitorCrashingPipeline can transition from CRASHING -> RUNNING.  monitorPipeline can transition from CRASHING -> STANDBY (among other states).  In the test logs you linked on Slack, each of these happens at least once.  I'm not sure if the problem is the check in monitorCrashingPipeline , or if there's a deeper problem.

For example, there is at least one log line indicative of the line of monitorPipeline around 225 being responsible for the faulty transition to STANDBY. I am not sure how that code ran; there is an infinite loop that checks that the input and output commits are finished, and we couldn't finish that commit if we didn't start the pipeline successfully, which we couldn't do if we couldn't start the pipeline because of a missing secret.  Not sure, but the loop might be running for the previous commit still?

My main conclusion is that we need the Jaeger traces in tests.  And debug-level logs.

PR #7725, opened 2022-06-06
[CORE-606] Loki & logging improvements

robert-uhl:


src/server/debug/server/server.go
```diff
@@ -325,16 +326,18 @@ func writeProfile(ctx context.Context, w io.Writer, profile *debug.Profile) erro
...
+		t := time.NewTimer(duration)
+		select {
+		case <-ctx.Done():
+			if !t.Stop() {
+				<-t.C
+			}
+			return errors.EnsureStack(ctx.Err())
+		case <-t.C:
+			return nil
+		}
```
jrockway, line 340: Why do it this way?  There are now 3 branches to test; ctx.Done + !t.Stop, ctx.Done + t.Stop, and recv on t.C.

src/server/pps/server/api_server.go
```diff
@@ -1405,57 +1446,21 @@ func parseLokiLine(inputLine string, msg *pps.LogMessage) error {
...
+	parsers := []struct {
+		driver string
+		parser func(string, *pps.LogMessage) error
+	}{
 		// The order here is important, because native and docker messages are ambiguous.
 		// It is unfortunate that we can't happy-path native.
-		"cri":    parseCRI,
-		"docker": parseDocker,
-		"native": parseNative,
+		{"cri", parseCRILine},
```
jrockway, line 1457: Yeah not sure what happened here.  I guess the order doesn't actually matter.

jrockway (approved):

src/server/debug/server/server.go
robert-uhl: In order to distinguish between a cancelled context and an exceeded deadline (returning an error in the former situation but no error in the latter).

I think maybe t.Stop() can just be called and its return value be ignored, to be honest, because we’re not going to reuse the channel.

PR #7728, opened 2022-06-06
Check status code when importing jsonnet.

PFedak:
This won't deflake the test, but will avoid cloaking request errors as
parse problems.


acohen4 (approved):
LGTM

PR #7732, opened 2022-06-06
Add Support for Plurals in Pachctl

FahadBSyed:
This Pull Request adds support for the plural version of resources like `pipelines` or `files`. Previously, users had to input commands in a singular form like so:
```
pachctl list repo
```

Now, a user can input:
```
pachctl list repos
```

There's no difference in the execution of these commands under the hood, the plurals are just the aliases of the original commands.

### DIFF
--- a/src/internal/cmdutil/cobra.go
+++ b/src/internal/cmdutil/cobra.go
@@ -319,7 +319,17 @@ func (r *RepeatedStringArg) Type() string {
 // '{{alias}}' will be replaced with the full command path.  These commands can
 // later be merged into the final Command tree using 'MergeCommands' below.
 func CreateAlias(cmd *cobra.Command, invocation string) *cobra.Command {
+	return createAlias(cmd, invocation)
+}
 
+// CreateAliases is like CreateAlias, except it allows us to specify one or more plurals for the
+// last argument in the command with the assumption that the last argument in the command is a resource
+// such as 'repo' or 'pipeline'.
+func CreateAliases(cmd *cobra.Command, invocation string, plurals ...string) *cobra.Command {
+	return createAlias(cmd, invocation, plurals...)
+}
+
+func createAlias(cmd *cobra.Command, invocation string, plurals ...string) *cobra.Command {
 	// Create logical commands for each substring in each invocation
 	var root, prev *cobra.Command
 	args := strings.Split(invocation, " ")
@@ -337,6 +347,11 @@ func CreateAlias(cmd *cobra.Command, invocation string) *cobra.Command {
 				cur.Use = strings.ReplaceAll(cmd.Use, "{{alias}}", arg)
 			}
 			cur.Example = strings.ReplaceAll(cmd.Example, "{{alias}}", fmt.Sprintf("%s %s", os.Args[0], invocation))
+
+			if len(plurals) != 0 {
+				cur.Aliases = append([]string{cur.Use}, plurals...)
+			}
+
 		} else {
 			cur.Use = arg
 		}
@@ -401,8 +416,20 @@ func MergeCommands(root *cobra.Command, children []*cobra.Command) {
 // commands have no functionality except to output some docs and related
 // commands, and should not specify a 'Run' attribute.
 func CreateDocsAlias(command *cobra.Command, invocation string, pattern string) *cobra.Command {
+	return createDocsAlias(command, invocation, pattern)
+}
+
+// CreateDocsAliases is like CreateDocsAlias, except it allows us to specify one or more plurals for the
+// last argument in the command with the assumption that the last argument in the command is a resource
+// such as 'repo' or 'pipeline'.
+func CreateDocsAliases(command *cobra.Command, invocation string, pattern string,
+	plurals ...string) *cobra.Command {
+	return createDocsAlias(command, invocation, pattern, plurals...)
+}
+
+func createDocsAlias(command *cobra.Command, invocation string, pattern string, plurals ...string) *cobra.Command {
 	// This should create a linked-list-shaped tree, follow it to the one leaf
-	root := CreateAlias(command, invocation)
+	root := CreateAliases(command, invocation, plurals...)
 	command = root
 	for len(command.Commands()) != 0 {
 		command = command.Commands()[0]
--- a/src/internal/cmdutil/cobra.go
+++ b/src/internal/cmdutil/cobra.go
@@ -322,14 +322,14 @@ func CreateAlias(cmd *cobra.Command, invocation string) *cobra.Command {
 	return createAlias(cmd, invocation)
 }
 
-// CreateAliases is like CreateAlias, except it allows us to specify one or more plurals for the
+// CreateAliases is like CreateAlias, except it allows us to specify one or more synonyms for the
 // last argument in the command with the assumption that the last argument in the command is a resource
 // such as 'repo' or 'pipeline'.
-func CreateAliases(cmd *cobra.Command, invocation string, plurals ...string) *cobra.Command {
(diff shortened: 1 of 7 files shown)


src/internal/cmdutil/cobra.go
```diff
@@ -319,7 +319,17 @@ func (r *RepeatedStringArg) Type() string {
 // '{{alias}}' will be replaced with the full command path.  These commands can
 // later be merged into the final Command tree using 'MergeCommands' below.
 func CreateAlias(cmd *cobra.Command, invocation string) *cobra.Command {
+	return createAlias(cmd, invocation)
+}
+
+// CreateAliases is like CreateAlias, except it allows us to specify one or more plurals for the
+// last argument in the command with the assumption that the last argument in the command is a resource
+// such as 'repo' or 'pipeline'.
+func CreateAliases(cmd *cobra.Command, invocation string, plurals ...string) *cobra.Command {
```
acohen4, line 328: maybe synonyms > plurals is a more accurate parameter name? WDYT?

src/internal/cmdutil/cobra.go
```diff
@@ -337,6 +347,11 @@ func CreateAlias(cmd *cobra.Command, invocation string) *cobra.Command {
 				cur.Use = strings.ReplaceAll(cmd.Use, "{{alias}}", arg)
 			}
 			cur.Example = strings.ReplaceAll(cmd.Example, "{{alias}}", fmt.Sprintf("%s %s", os.Args[0], invocation))
+
+			if len(plurals) != 0 {
+				cur.Aliases = append([]string{cur.Use}, plurals...)
```
acohen4, line 352: does cur.Use need to be re-included as an alias here?

src/server/pps/cmds/cmds.go
```diff
@@ -44,6 +44,32 @@ import (
...
+	secret   = "secret"
+
+	// Plural variables are used below for user convenience.
+	datums    = "datums"
+	jobs      = "jobs"
+	pipelines = "pipelines"
+	secrets   = "secrets"
+
+	// Command 'verbs' are also used in testing.
+	create  = "create"
```
acohen4, line 61: if we aren't using these variables here I think it would make more sense to define them with the tests

src/server/pfs/cmds/cmds.go
```diff
@@ -41,6 +41,35 @@ import (
...
+	repo   = "repo"
+
+	// Plural variables are used below for user convenience.
+	branches = "branches"
+	commits  = "commits"
+	files    = "files"
+	repos    = "repos"
+
+	// Command 'verbs' are also used in testing.
+	copy      = "copy"
```
acohen4, line 58: like in pps, I think we prefer to define this with the tests

src/server/pfs/cmds/cmds_test.go
```diff
@@ -149,3 +168,62 @@ func TestDiffFile(t *testing.T) {
...
+	pluralCheckTemplate := `
+		pachctl {{RESOURCE_PLURAL}} -h > plural.txt
+		pachctl {{RESOURCE}} -h > singular.txt
+		diff plural.txt singular.txt
+		rm plural.txt singular.txt
+	`
+
+	plurals := pluralsMap()
+
+	if testing.Short() {
```
acohen4, line 216: just a style consistency thing: we should probably continue to run the if testing.Short() block first

src/server/pfs/cmds/cmds_test.go
```diff
@@ -2,14 +2,33 @@ package cmds
...
+	"strings"
 	"testing"
 
 	"github.com/pachyderm/pachyderm/v2/src/internal/minikubetestenv"
 	"github.com/pachyderm/pachyderm/v2/src/internal/require"
 	tu "github.com/pachyderm/pachyderm/v2/src/internal/testutil"
-	"github.com/pachyderm/pachyderm/v2/src/server/pfs/fuse"
 )
 
+func resourcesMap() map[string][]string {
```
acohen4, line 14: nit: I think we can just define resourcesMap & pluralsMap as const variables called `resources` and `plurals` if they aren't being modified

src/server/pfs/cmds/cmds_test.go
```diff
@@ -149,3 +168,62 @@ func TestDiffFile(t *testing.T) {
...
+
+// TestPlurals walks through the command tree for each resource and verb combination defined in PPS.
+// A template is filled in that calls the help flag and the output is compared. It seems like 'match'
+// is unable to compare the outputs correctly, but we can use diff here which returns an exit code of 0
+// if there is no difference.
+func TestPlurals(t *testing.T) {
+	pluralCheckTemplate := `
+		pachctl {{VERB}} {{RESOURCE_PLURAL}} -h > plural.txt
+		pachctl {{VERB}} {{RESOURCE}} -h > singular.txt
+		diff plural.txt singular.txt
```
acohen4, line 180: What's the assertion here? Are we expecting the diffs to be equal?

src/server/auth/cmds/cmds_test.go
```diff
@@ -417,3 +417,41 @@ func TestRotateRootToken(t *testing.T) {
...
+
+	pluralMap := map[string]string{
+		"repo": "repos",
+	}
+
+	if testing.Short() {
+		t.Skip("Skipping integration tests in short mode")
+	}
+
+	c, _ := minikubetestenv.AcquireCluster(t)
```
acohen4, line 445: If we aren't making calls to a pachd instance, acquiring a cluster client shouldn't be necessary. If so, you can use `BashCmd` instead of `PachctlBashCmd` instead

src/internal/cmdutil/cobra.go
FahadBSyed: Yeah I think that makes more sense. This way you could do shorthand versions to like `fi` for `files` or something

src/internal/cmdutil/cobra.go
FahadBSyed: It doesn't actually, and it probably shouldn't. I think doing so actually makes it show up twice in the aliases section:

```
Aliases:
  repo, repo, repos
```

So I can remove it and that should hopefully clean up the aliases section too

src/server/auth/cmds/cmds_test.go
FahadBSyed: Ahh cool, will do!

src/server/pfs/cmds/cmds_test.go
FahadBSyed: I think with Go, we can't define const maps at a package scope (although that would be nice)

src/server/pfs/cmds/cmds_test.go
FahadBSyed: Yes exactly, we know the command is working if the help test is displaying for it. Otherwise, we'll get an error back from pachctl. So we compare the output of the help commands to make sure that they are the same.

src/server/pps/cmds/cmds.go
FahadBSyed: Sure, can do

acohen4 (approved):
LGTM

src/server/pfs/cmds/cmds_test.go
```diff
@@ -2,12 +2,36 @@ package cmds
...
+	get       = "get"
+	glob      = "glob"
+	inspect   = "inspect"
+	list      = "list"
+	put       = "put"
+	squash    = "squash"
+	start     = "start"
+	subscribe = "subscribe"
+	update    = "update"
+	wait      = "job"
```
PFedak, line 34: A little concerning this isn't causing any issue

PR #7741, opened 2022-06-07
[CORE-701] CSV egress

robert-uhl:


src/pfs/pfs.proto
```diff
@@ -472,6 +472,7 @@ message SQLDatabaseEgress {
         CSV = 1;
         JSON = 2;
         PARQUET = 3;
+        HEADER_CSV = 4;
```
albscui, line 475: My reservation is that this could be a bit confusing to the user, there being two CSV formats: `csv` vs `header_csv`. What do you think about adding a `hasHeader` toggle instead?

src/internal/sdata/csv.go
```diff
@@ -174,3 +174,58 @@ func (p *CSVParser) Next(row Tuple) error {
...
 }
+
+type HeaderCSVParser struct {
+	CSVParser
+	fields       []string
+	headerRead   bool
+	fieldIndices map[int]int // maps read column to tuple column
+}
+
+func NewHeaderCSVParser(r io.Reader, fields []string) TupleReader {
```
albscui, line 185: I think it's probably better to push this logic into the `CSVParser`, because we have some existing tests that test only `CSVParser` right now.

src/pfs/pfs.proto
robert-uhl: Updated to imply header row from presence of columns.

PR #7742, opened 2022-06-07
CORE-656 Expose whether to write header via pachtf

albscui:
Now the user would have to supply an additional argument via  `pachctl update pipeline --jsonnet ... --arg  hasHeader=true`


robert-uhl (approved):
LGTM!

robert-uhl (approved):
LGTM!

PR #7744, opened 2022-06-07
Add Enterprise Server/Member deploy test

acohen4:


src/testing/deploy/deploy_test.go
```diff
@@ -86,30 +111,3 @@ func mockIDPLogin(t testing.TB, c *client.APIClient) {
 	require.NoError(t, err)
 	require.Equal(t, "user:"+testutil.DexMockConnectorEmail, whoami.Username)
 }
-
-func TestParallelDeployments(t *testing.T) {
```
acohen4, line 90: This test was really just testing the minikubetestenv functionality. I don't think its worth the amount of time it needs to run

jrockway: I was looking at that yesterday and I'm like wait, why do we care if we can deploy in parallel.  For minikubetestenv is why ;)

jrockway (approved):

PR #7745, opened 2022-06-08
Configurable Security Contexts

tybritten:
- Hardcodes some sane security defaults
 ```
runAsNonRoot: true
allowPrivilegeEscalation: false
readOnlyRootFilesystem: true
capabilities:
    drop:
    - all
```
~- Allows users to configure runAs User/Group globally for the deployment~
- with container fs set as readonly, add emptyDir volumes for containers that need to write locally

FREQQ-110

### CODE REVIEW

File: etc/testing/circle/helm-load-env-values.yaml [hunk 1]
```diff
@@ -1,3 +1,5 @@
+deployTarget: LOCAL
```
tybritten, line 1: had to add this to tell etcd to use a root securitycontext

etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -203,7 +193,7 @@ spec:
           value:  {{.Values.global.noProxy}}
         {{ end }}
         - name: ENABLE_WORKER_SECURITY_CONTEXTS
-          value: {{ .Values.pachd.securityContext.enabled | quote }}
+          value: {{ .Values.global.securityContexts.enabled | quote }}
```
jrockway, line 196: I think some of the changes you're making here should propagate to the worker RCs; src/server/pps/server/worker_rc.go line 332 (sidecar) and line 348 (user code) is where those are defined.

jrockway (approved):
Answering my own questions.  This has to be optional or it breaks Openshift.  CI is already testing this.

Only thing we need to figure out is what to do about workers.  Could probably be done in a follow-up.

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
```diff
@@ -95,13 +103,34 @@ spec:
...
         {{- end }}
+      {{- if .Values.global.securityContexts.enabled }}
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          capabilities:
+              drop:
+              - all
+      securityContext:
+        runAsUser: {{ .Values.global.securityContexts.runAsUser }}
```
hesantosh, line 114: pg-bouncer user is `1001` so global `runAsUser` wont work here. I am assuming its `1000` globally.  https://github.com/bitnami/bitnami-docker-pgbouncer/blob/1.17.0-debian-11-r3/1/debian-11/Dockerfile#L38

etc/helm/pachyderm/templates/cloudsqlAuthProxy/deployment.yaml
```diff
@@ -39,11 +39,20 @@ spec:
           - "-instances={{ .Values.cloudsqlAuthProxy.connectionName }}=tcp:0.0.0.0:{{ .Values.cloudsqlAuthProxy.port }}"
         ports:
         - containerPort: {{ .Values.cloudsqlAuthProxy.port }}
-        securityContext:
-          # The default Cloud SQL proxy image runs as the
-          # "nonroot" user and group (uid: 65532) by default.
-          runAsNonRoot: true
         resources: {{ toYaml .Values.cloudsqlAuthProxy.resources | nindent 10 }}
+      {{- if .Values.global.securityContexts.enabled }}
+        securityContext:
```
hesantosh, line 44: I would suggest something like this for both pod and container securitycontext.
https://gitea.com/gitea/helm-chart/src/commit/dd702a7ae62511c99e4aa3db29421b68f5ea44e3/templates/gitea/statefulset.yaml#L68
I expect we will need to add/remove few values along the way.

hesantosh: If not, `runAsUser`, `runAsGroup` and `runAsNonRoot` needs to be on container and Pod level. I know its not required from K8s POV but thats a requirement in our OPA.

etc/helm/pachyderm/templates/cloudsqlAuthProxy/deployment.yaml
```diff
@@ -39,11 +39,20 @@ spec:
...
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          capabilities:
+              drop:
+              - all
+      securityContext:
```
hesantosh, line 50: Sorry, I might have missed this. `seccompProfile` needs to be on pod securitycontext.
```
        runAsUser: {{ .Values.global.securityContexts.runAsUser }}
        runAsGroup: {{ .Values.global.securityContexts.runAsGroup }}
        fsGroup: {{ .Values.global.securityContexts.runAsGroup }}
        runAsNonRoot: true
        seccompProfile:
          type: RuntimeDefault
```

hesantosh (changes requested):

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
tybritten: This does work. The container permissions have read and execute access for other, so any uid can be used.

tybritten:
We're hard-coding the UID/GIDs for the pachyderm components but a user can still configure the user ones for pipelines via the pipeline spec. For more advanced changes, it requires using a `pod_patch`

src/server/pachyderm_test.go
```diff
@@ -10650,3 +10649,99 @@ func TestDatabaseStats(t *testing.T) {
...
+	require.NoError(t, c.CreateRepo(dataRepo))
+
+	commit1, err := c.StartCommit(dataRepo, "master")
+	require.NoError(t, err)
+	require.NoError(t, c.PutFile(commit1, "file", strings.NewReader("foo"), client.WithAppendPutFile()))
+	require.NoError(t, c.FinishCommit(dataRepo, commit1.Branch.Name, commit1.ID))
+	pipeline := tu.UniqueString("TestSimplePipeline")
+	req := basicPipelineReq(pipeline, dataRepo)
+
+	req.PodPatch = "[{\"op\":\"add\",\"path\":\"/securityContext\",\"value\":{}},{\"op\":\"add\",\"path\":\"/securityContext\",\"value\":{\"runAsGroup\":1000,\"runAsUser\":1000,\"fsGroup\":1000,\"runAsNonRoot\":true,\"seccompProfile\":{\"type\":\"RuntimeDefault\"}}},{\"op\":\"add\",\"path\":\"/containers/0/securityContext\",\"value\":{}},{\"op\":\"add\",\"path\":\"/containers/0/securityContext\",\"value\":{\"runAsGroup\":1000,\"runAsUser\":1000,\"allowPrivilegeEscalation\":false,\"capabilities\":{\"drop\":[\"all\"]},\"readOnlyRootFilesystem\":true}}]"
```
jrockway, line 10719: You can use `` ` `` as the quoting operator and then you don't have to escape `"` in the JSON.

src/server/pps/server/worker_rc.go
```diff
@@ -278,17 +279,36 @@ func (kd *kubeDriver) workerPodSpec(options *workerOptions, pipelineInfo *pps.Pi
...
+	}
+	// add emptydir for /tmp to allow for read only rootfs
+	options.volumes = append(options.volumes, v1.Volume{
+		Name: "tmp",
+		VolumeSource: v1.VolumeSource{
+			EmptyDir: &v1.EmptyDirVolumeSource{},
+		}})
+	tmpDirVolumeMount := v1.VolumeMount{
+		Name:      "tmp",
+		MountPath: "/tmp",
```
jrockway, line 307: What about `/pfs`?  I'm surprised there are no complaints here; maybe our tests aren't writing to `/pfs/out`?

jrockway (approved):
Looks good.  Couple questions below.

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
```diff
@@ -87,6 +87,14 @@ spec:
             cpu: 250m
             memory: 256M
         volumeMounts:
+        - mountPath: /opt/bitnami/pgbouncer/conf/
```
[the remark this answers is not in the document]
tybritten: i tried, there's other stuff in that folder that getn wiped then

src/server/pachyderm_test.go
tybritten: i never remember which languages support that and which don't haha

src/server/pps/server/worker_rc.go
tybritten: somewhere else in the code the emptyDir is created for /pfs. It's in the pod at runtime

etc/helm/pachyderm/templates/etcd/statefulset.yaml
```diff
@@ -73,18 +73,32 @@ spec:
         volumeMounts:
         - mountPath: /var/data/etcd
           name: etcd-storage
+      {{- if .Values.global.securityContexts.enabled }}
+      {{- if and (eq (include "pachyderm.storageBackend" . ) "LOCAL") .Values.pachd.storage.local.requireRoot }}
```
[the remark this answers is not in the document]
tybritten: it shouldn't, i pulled that out already

etc/helm/pachyderm/templates/proxy/deployment.yaml
```diff
@@ -96,16 +96,28 @@ spec:
...
             {{- end }}
+     {{- if .Values.global.securityContexts.enabled }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+                drop:
+                - all
+      securityContext:
+        runAsUser: 101
```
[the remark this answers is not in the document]
tybritten: no we can do that, that's what i do for the workers

tybritten:
One breaking change in this PR:
disabling security contexts is now done globally with a single value at `global. securityContexts.enabled=false`, instead of `securityContexts.enabled=false` under each service. This only really applies to OpenShift users

PR #7747, opened 2022-06-08
Mount server separate binary

smalyala:
- Separate mount server into its own binary
- Build and publish docker image via CI


src/internal/minikubetestenv/deploy.go
```diff
@@ -478,6 +478,9 @@ func putRelease(t testing.TB, ctx context.Context, namespace string, kubeClient
 	if opts.EnterpriseMember {
 		helmOpts = union(helmOpts, withEnterpriseMember(pachAddress.Host, int(pachAddress.Port)))
 	}
+	if opts.EnterpriseMember {
```
chainlink, line 481: This looks unrelated?

chainlink:
Heads up @molinamelendezj This adds a docker build to goreleaser.

src/internal/minikubetestenv/deploy.go
smalyala: Good catch. Removed

chainlink (approved):

PR #7749, opened 2022-06-08
Add zombie data detection option to fsck.

PFedak:


brycemcanally:
I'm concerned about how heavyweight this is. It certainly shouldn't be the default, but even then it could take a really long time for clusters with a lot of commits / numbers of files. Distributing this through the task service could help, but I wonder how much value this type of check provides by going across all of the commits vs a subset. It might be more beneficial in terms of cost to just check commits that are active, probably just the HEAD commits, or have the client specify a commit set / commit. That would work better longer term as well because I can see us eventually getting to a state where older commits are uncompacted to save space.

src/pfs/pfs.proto
```diff
@@ -386,6 +386,9 @@ message DiffFileResponse {
 
 message FsckRequest {
   bool fix = 1;
+  Commit zombie_target = 2;
```
brycemcanally, line 389: I feel like we should have a oneof which sets either a commit or the boolean.

src/server/pfs/cmds/cmds.go
```diff
@@ -1604,13 +1623,15 @@ Objects are a low-level resource and should not be accessed directly by most use
...
 			}
-			if !errors {
+			if !foundErrors {
 				fmt.Println("No errors found.")
 			}
 			return nil
 		}),
 	}
 	fsck.Flags().BoolVarP(&fix, "fix", "f", false, "Attempt to fix as many issues as possible.")
+	fsck.Flags().BoolVar(&fullZombie, "zombie-all", false, "Check all pipelines for zombie files: files corresponding to old inputs that were not properly deleted")
```
brycemcanally, line 1633: Can we have the terminology be consistently `all` or `full`. It seems like `all` would be more consistent with other functionality (i.e. delete pipeline).

src/server/pfs/server/api_server.go
```diff
@@ -629,11 +629,38 @@ func (a *apiServer) DeleteAll(ctx context.Context, request *types.Empty) (respon
...
 	}); err != nil {
 		return err
 	}
+
+	if request.ZombieTarget != nil {
+		return a.driver.detectZombie(ctx, request.ZombieTarget, fsckServer.Send)
+	}
+	if request.FullZombieCheck {
+		// list meta repos as a proxy for finding pipelines
+		a.driver.listRepo(ctx, false, pfs.MetaRepoType, func(info *pfs.RepoInfo) error {
```
brycemcanally, line 644: No error check here?

src/server/pfs/server/driver_fsck.go
```diff
@@ -302,3 +311,99 @@ func (d *driver) fsck(ctx context.Context, fix bool, cb func(*pfs.FsckResponse)
...
+			return errors.EnsureStack(w.Put(common.MetaFilePath(id), id, true,
+				strings.NewReader(f.Index().Path+"\n")))
+		}))
+	})
+	if err != nil {
+		return err
+	}
+	// now merge with the meta commit to look for extra datums in the output commit
+	return d.storage.WithRenewer(ctx, defaultTTL, func(ctx context.Context, r *fileset.Renewer) error {
+		r.Add(ctx, *id)
```
brycemcanally, line 363: Error check?

brycemcanally, line 363: Error check?

src/server/pfs/server/api_server.go
PFedak: Sorry for the sloppiness on these, there must be an off-the-shelf linting option for this

brycemcanally (approved):
lgtm, just one minor nitpick left.

PR #7754, opened 2022-06-09
[CORE-730] Clean glob in multiple places.

PFedak:
Group/join pattern extraction and file globbing would use the same glob
inconsistently, so that a file returned from globbing would fail to
extract and lead to degenerate groups/joins.

It turns out this is a regression from v1, recorded as #5365, so also
re-enable the test that was added.


msteffen (approved):
LGTM. Blast from the past!

brycemcanally (approved):
lgtm

PR #7755, opened 2022-06-09
Bootstrap Identity Service Via environment

acohen4:


etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -185,7 +197,7 @@ spec:
           {{- else }}
           value: {{ .Values.pachd.rootToken | quote }}
           {{- end }}
-        {{- end }}
+        {{- end }}
```
jrockway: extra spaces added at end of line.

etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -167,6 +167,18 @@ spec:
             secretKeyRef:
               name: {{ (include "pachyderm.enterpriseLicenseKeySecretName" . ) | trim | quote }}
               key: "enterprise-license-key"
+          {{- if (include "pachyderm.upstreamIDPsSecretName" .) }}
+        - name: IDP_CONNECTORS
+          valueFrom:
+            secretKeyRef:
```
jrockway, line 175: Maybe not relevant yet, but there is a limit on the total size of the environment.  I think it's 8MB, so we're probably not going to hit it, but something to be aware of if we want an unbounded amount of IDP connectors.

I think if I were implementing this right now, I would have made the environment variable point to a config file, and then mount the secret as a volume, and have pachd read the config file.  No need to change things though, just a matter of personal preference.

src/server/identity/server/api_server.go
```diff
@@ -79,7 +138,14 @@ func (a *apiServer) GetIDPConnector(ctx context.Context, req *identity.GetIDPCon
 	}, nil
 }
 
-func (a *apiServer) UpdateIDPConnector(ctx context.Context, req *identity.UpdateIDPConnectorRequest) (resp *identity.UpdateIDPConnectorResponse, retErr error) {
+func (a *apiServer) UpdateIDPConnector(ctx context.Context, req *identity.UpdateIDPConnectorRequest) (*identity.UpdateIDPConnectorResponse, error) {
+	if a.env.Config.IdentityConnectors != "" {
+		return nil, errors.New("identity.UpdateIDPConnector() is disable when the Identity Connectors are set via environment.")
```
jrockway, line 143: "is disabled" (this typo appears in a few other places; was going to mention each but realized there are at least 4 and it already sounds kind of mean once :))

src/server/identity/server/api_server.go
```diff
@@ -22,7 +23,7 @@ type apiServer struct {
 }
 
 // NewIdentityServer returns an implementation of identity.APIServer.
-func NewIdentityServer(env Env, public bool) identity.APIServer {
+func NewIdentityServer(env Env, public bool) (identity.APIServer, func(context.Context) error) {
```
jrockway, line 26: I guess I didn't pick up on this when you did the license server stuff... but I think it would be better to just have an EnvBootstrap method (and in main.go, collect a slice of `interface { EnvBootstrap(ctx context.Context) error }` objects that you call EnvBootstrap on).

src/server/identity/server/api_server.go
```diff
@@ -37,10 +38,61 @@ func NewIdentityServer(env Env, public bool) identity.APIServer {
...
-	return server
+	return server, server.envBootstrap
+}
+
+func (a *apiServer) envBootstrap(ctx context.Context) error {
+	if a.env.Config.IdentityConfig != "" {
+		a.env.Logger.Info("Started to configure identity server config via environment")
+		var config identity.IdentityServerConfig
+		if err := yaml.Unmarshal([]byte(a.env.Config.IdentityConfig), &config); err != nil {
+			return errors.Wrapf(err, "failed to unmarshal IdentityConfig with value: %s.", a.env.Config.IdentityConfig)
```
jrockway, line 49: minor nit; I'd avoid "failed" in error objects.  would probably say `errors.Wrapf(err, "unmarshal IdentityServerConfig %q", a.env.Config.IdentityConfig)`.

src/server/identity/server/api_server.go
```diff
@@ -37,10 +38,61 @@ func NewIdentityServer(env Env, public bool) identity.APIServer {
...
+				}
+				delete(exIds, c.Id)
+			} else {
+				if _, err := a.createIDPConnector(ctx, &identity.CreateIDPConnectorRequest{Connector: &c}); err != nil {
+					return errors.Wrapf(err, "failed to create connector with ID: %s.", c.Id)
+				}
+			}
+		}
+		for e := range exIds {
+			a.deleteIDPConnector(ctx, &identity.DeleteIDPConnectorRequest{Id: e})
```
jrockway, line 81: If this errors, we should at least print the message as a warning.  (Avoids confusion as to why something didn't get deleted.)

etc/helm/pachyderm/templates/pachd/idp-secret.yaml
```diff
@@ -0,0 +1,14 @@
...
+*/ -}}
+{{- if and ( not .Values.oidc.upstreamIDPsSecretName) (or .Values.oidc.upstreamIDPs .Values.oidc.mockIDP) }} 
+apiVersion: v1
+kind: Secret
+metadata:
+  name: pachyderm-identity
+  namespace: {{ .Release.Namespace }}
+stringData:
+  idps: |
+    {{ include "pachyderm.idps" . }}
```
jrockway, line 13: It's a little action-at-a-distance-y for `pachyderm.idps` to have to be indented by 4 spaces.  It does do that, but it makes more sense to me to do that here.

jrockway (approved):
Looks good.  Some minor complaints about error messages and a suggestion to make EnvBootstrap a method instead of something that gets returned.

src/server/identity/server/api_server.go
acohen4: yikes good catch

src/server/identity/server/api_server.go
acohen4: so I think I didn't do this last time because it would introduce a circular dependency between the main & the IdentityServer

PR #7756, opened 2022-06-09
Add SSL config to listener config

tybritten:
The Postgres direct listener was being instantiated without the SSL config option, this fixes it.


jrockway (approved):
This is exciting.  I think that this means that pachyderm.pachops.com can skip the cloud sql proxy and connect directly to the internal IP (which mandates TLS, unconfigurably on their end).

PR #7759, opened 2022-06-10
[CORE-608] Enable Loki by default

robert-uhl:
This PR deploys & enables Loki by default, and also makes the feature available to all users, not just those with an enterprise license.


jrockway (approved):
This looks great.  Nice work!

PR #7761, opened 2022-06-10
CORE-444 Add basic postgres stats to pachctl debug dump

FahadBSyed:
## Description
This PR expands `pachctl debug dump` to include basic postgres stats such as
- top 100 active activities sorted by time
- row counts for tables in the `pachyderm` database
- table sizes for tables in the `pachyderm` database

The next step is to include a full dump of the `pachyderm` database in the debug dump response


src/debug/debug.proto
```diff
@@ -23,6 +23,7 @@ message Filter {
     bool pachd = 1;	
     pps_v2.Pipeline pipeline = 2;	
     Worker worker = 3;
+    bool postgres = 4;
```
acohen4, line 26: minor suggestion: should we name this `database`? Even though we are tightly coupled to postgres, the API doesn't necessarily need to acknowledge that

src/server/debug/server/server.go
```diff
@@ -927,3 +955,150 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+		if err != nil {
+			return errors.EnsureStack(err)
+		}
+		return errors.EnsureStack(err)
+	}, prefix...)
+}
+
+func queryDatabaseRowCounts(conn *sql.Conn, w io.Writer) error {
+	query := `
+		SELECT schemaname,relname,n_live_tup
```
acohen4, line 1003: I think we might as well pull some more columns here:
`seq_scan`, `idx_scan` both look helpful

src/server/debug/server/server.go
```diff
@@ -927,3 +955,150 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+
+	return writeRows(rows, w)
+}
+
+func writeRows(rows *sql.Rows, w io.Writer) error {
+	cols, err := rows.Columns()
+	if err != nil {
+		return errors.EnsureStack(err)
+	}
+
```
acohen4, line 1053: nit: I think we generally want to avoid adding extra newlines in functions, so I'd remove these. That being said we haven't been explicit about it.

src/server/debug/server/server.go
```diff
@@ -927,3 +955,150 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+	header := strings.Join(cols, "\t") + "\n"
+	tabWriter := tabwriter.NewWriter(w, header)
+
+	pointers := make([]interface{}, len(cols))
```
acohen4, line 1057: I think this sort of operation can be done more simply using the dependency: "github.com/jmoiron/sqlx" that we use for SQL in most of the codebase. MapScan or SliceScan should do the job without needing to load to the list of pointers

acohen4 (changes requested):

src/debug/debug.proto
FahadBSyed: Sure, that makes sense!

src/server/debug/server/server.go
```diff
@@ -927,3 +955,150 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+		}
+		return errors.EnsureStack(err)
+	}, prefix...)
+}
+
+func queryDatabaseRowCounts(conn *sql.Conn, w io.Writer) error {
+	query := `
+		SELECT schemaname,relname,n_live_tup 
+		  FROM pg_stat_user_tables 
+		ORDER BY n_live_tup DESC;
```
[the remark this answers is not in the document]
FahadBSyed: That's a good point, can do!

src/server/debug/server/server.go
FahadBSyed: Ahh gotcha, thanks for the style stuff. I will have to definitely deprogram some old habits 😆

src/server/debug/server/server.go
```diff
@@ -927,3 +955,150 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+
+	return queryDatabase(conn, query, w)
+}
+
+func queryDatabaseTableSizes(conn *sql.Conn, w io.Writer) error {
+	query := `
+		SELECT nspname || '.' || relname AS "relation",
+		pg_size_pretty(pg_total_relation_size(C.oid)) AS "total_size"
+		FROM pg_class C
+		LEFT JOIN pg_namespace N ON (N.oid = C.relnamespace)
```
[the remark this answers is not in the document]
FahadBSyed: So the `LEFT JOIN` is needed for nspname to resolve and combine the namespaces.
The `LIMIT` is just bad copy/paste 🤣 
I'm definitely an SQL noob so I used a combination of docs and stack overflow to determine these queries

src/server/debug/server/server.go
FahadBSyed: Oh cool, it would also be nice to not have to write / maintain this logic ourselves. Thanks for pointing that out!

src/server/debug/server/server_test.go
```diff
@@ -299,3 +302,50 @@ slice samples:
...
+		t.Skip("Skipping integration tests in short mode")
+	}
+
+	c, _ := minikubetestenv.AcquireCluster(t)
+
+	numCommits := 5
+
+	// Load some very basic dummy data into the database.
+	require.NoError(t, tu.PachctlBashCmd(t, c, `
+		pachctl create repo {{.repo}}
```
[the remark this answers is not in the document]
FahadBSyed: Yeah that's definitely true, that would be closer to a true unit test instead of doing an integration test here

src/server/debug/server/server.go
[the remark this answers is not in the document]
acohen4: Ah I see. Should this be an inner join instead? What would it mean for a row in pg_class to have a `relnamespace`  without an entry in pg_namespace?

One more note: it might also make sense here to order by relation (namespace, table)

src/server/debug/server/server.go
```diff
@@ -927,3 +952,123 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+	return nil
+}
+
+func writeRowToTabWriter(rows *sql.Rows, tabWriter *tabwriter.Writer) error {
+	results := make(map[string]interface{})
+	if err := sqlx.MapScan(rows, results); err != nil {
+		return errors.EnsureStack(err)
+	}
+	row := ""
+	for _, value := range results {
```
acohen4, line 1063: I don't think this will print rows in the right order. Maybe SliceScan will?

src/server/debug/server/server.go
```diff
@@ -105,39 +109,60 @@ func (s *debugServer) handleRedirect(
...
-			}
-			if wantAppLogs {
-				// All other pachyderm apps (console, pg-bouncer, etcd, etc.).
-				if err := s.appLogs(ctx, tw); err != nil {
-					return err
-				}
-			}
-			return nil
+
+			return s.collectEverything(pachClient, collectPachd, collectPipeline, collectWorker, redirect, wantAppLogs,
```
jrockway, line 125: I don't see a lot of value in delegating to another function here.  The list of arguments is almost longer than the actual code.

Do keep in mind that this function runs on pachd and every worker, so we can't do things like the database dump here without verifying that we're the first pachd and not a replica or a worker.

src/server/debug/server/server.go
```diff
@@ -105,39 +109,60 @@ func (s *debugServer) handleRedirect(
...
+			return err
+		}
+	}
+	if wantAppLogs {
+		// All other pachyderm apps (console, pg-bouncer, etcd, etc.).
+		if err := s.appLogs(ctx, tw); err != nil {
+			return err
+		}
+	}
+	if err := s.collectDatabaseStats(ctx, tw, postgresPrefix); err != nil {
```
jrockway, line 160: This probably needs a bool to control, so that it only runs on the pachd dump and not the workers.  This code is exceedingly confusing, but "collectPachd" and "collectPipeline" is basically a no-op sometimes.

src/server/debug/server/server.go
```diff
@@ -927,3 +952,123 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+	return queryDatabase(conn, ctx, query, w)
+}
+
+func queryDatabaseActivities(conn *sql.Conn, ctx context.Context, w io.Writer) error {
+	query := `
+		SELECT current_timestamp - query_start as runtime, datname, usename, client_addr, query
+		FROM pg_stat_activity
+		WHERE state != 'idle'
+		ORDER by runtime desc
+		LIMIT 100;
```
jrockway, line 1011: I think we should avoid the limit.

src/server/debug/server/server.go
```diff
@@ -927,3 +952,123 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+		}
+		defer conn.Close()
+		if err = queryFunc(conn, ctx, w); err != nil {
+			return errors.EnsureStack(err)
+		}
+		return errors.EnsureStack(err)
+	}, prefix...)
+}
+
+func queryDatabaseRowCounts(conn *sql.Conn, ctx context.Context, w io.Writer) error {
```
jrockway, line 996: ctx.Context should always be the first argument to a function that takes on.  (And below.)

src/server/debug/server/server.go
```diff
@@ -927,3 +952,123 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+}
+
+func (s *debugServer) collectDatabaseTableSizes(ctx context.Context, tw *tar.Writer, prefix ...string) error {
+	return s.collectDatabaseQuery(ctx, tw, "table-sizes", queryDatabaseTableSizes, prefix...)
+}
+
+func (s *debugServer) collectDatabaseQuery(ctx context.Context, tw *tar.Writer, fileName string,
+	queryFunc func(conn *sql.Conn, ctx context.Context, w io.Writer) error, prefix ...string) error {
+	return collectDebugFile(tw, fileName, "txt", func(w io.Writer) error {
+		conn, err := s.env.GetDBClient().Conn(ctx)
```
jrockway, line 984: Conn() does not limit the query's runtime to this context; it basically controls how long we'll block to establish or reuse a new connection.  I have never called this before and I don't think it's the right thing to do here.  with the DB object, you can `QueryContext(ctx, ...` or `QueryxContext(ctx, ...` to run the query with a timeout.

jrockway: Also, I think the database should be passed as a parameter to the server's constructor.  Makes it explicit that this will make database calls.  (s.env.GetDBClient() will never return a different object.)

src/server/debug/server/server.go
jrockway: columns in the right order, I think you mean.  You can SliceScan or pass in rows.Columns() and iterate over that instead of the map keys.

I think the rows should just be JSON objects, though.  JSONlines format files are easy to analyze with JQ.  The first thing I do with a CSV file is convert it to JSON anyway ;)

src/server/debug/server/server.go
```diff
@@ -927,3 +952,123 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+
+func queryDatabase(conn *sql.Conn, ctx context.Context, query string, w io.Writer) error {
+	rows, err := conn.QueryContext(ctx, query)
+	if err != nil {
+		return errors.EnsureStack(err)
+	}
+	return writeRows(rows, w)
+}
+
+func writeRows(rows *sql.Rows, w io.Writer) error {
```
jrockway, line 1038: I'd consider just making these files JSON.  One JSON object per line representing the row.  the gzip compression of the output stream will negate most of the file size impact that duplicate keys on every line tend to cause.

I'd prefer JSON (technically jsonlines https://jsonlines.org/) so I can query the files with jq later on.

src/server/debug/server/server_test.go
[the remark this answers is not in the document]
jrockway: Will also say that we typically put these integration tests in `src/server/pachyderm_test.go`.  With Armaan's changes to how CI works, tests will be tagged as "kubernetes" or "not kubernetes", and putting this test here makes the unit test un-runnable without kubernetes, which is something we'd like to do less of in the future.

Other places that use k8s in the _server will just have to be cleaned up later, basically, so no need to add new code that has to be cleaned up next week.

And yeah, no shell scripts in the go code unless you are testing `pachctl` itself.  (That is a weird anti-pattern that isn't going to change any time soon; we could have easily done those tests in Go, but for whatever reason didn't.  So we don't have any code coverage information for `pachctl`, which makes the code less reliable and harder to work on.  Big problem.)

jrockway (changes requested):
Very good start!  Some comments.

src/server/debug/server/server_test.go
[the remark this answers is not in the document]
FahadBSyed: Makes sense, I can move this to `pachyderm_test.go` and remove the shell script stuff

src/server/debug/server/server.go
FahadBSyed: Ahh thanks for the explanation, that makes sense

src/server/debug/server/server.go
FahadBSyed: Originally I wanted to pass the conn down to make testing easier, but you make a good point. I'll update this as well

src/server/debug/server/server.go
FahadBSyed: Thats a much better idea, the tab writer is a little annoying anyways 😆

src/server/pachyderm_test.go
```diff
@@ -6,7 +6,9 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
+	"github.com/pachyderm/pachyderm/v2/src/debug"
```
jrockway, line 11: This isn't a big problem, but your IDE is not doing what gofmt/goimport/gopls would do here (which would be to put this import at line 33).  If it can be adjusted to be consistent with the rest of the codebase, that would be ideal.

src/server/pachyderm_test.go
```diff
@@ -10534,3 +10536,97 @@ func TestMissingSecretFailure(t *testing.T) {
...
+		Filter: &debug.Filter_Database{Database: true},
+	}
+
+	require.NoError(t, c.Dump(filter, 100, buff))
+
+	files := untar(t, buff)
+
+	for fileName := range files {
+		if strings.Contains(fileName, ".error") {
+			t.Fatalf("error returned from %s", fileName)
```
jrockway, line 10571: I suppose for consistency this should try to use `require`, but that only seems to have Matches, NoneEquals, and OneOfMatches... not NoneMatches.

General thoughts on assertion libraries though: https://github.com/golang/go/wiki/TestComments#assert-libraries

This is basically why they're not recommended... here you are in a test file that only uses assert libraries, but now you have to go off and write a general assertion function instead of a simple loop and == statement.  For that reason, I don't think you should go write that, but be aware of `require` when working in this codebase.

src/server/debug/server/server.go
```diff
@@ -927,3 +946,101 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+func (s *debugServer) collectDatabaseRowCount(ctx context.Context, tw *tar.Writer, prefix ...string) error {
+	return s.collectDatabaseQuery(ctx, tw, "row-counts", s.queryDatabaseRowCounts, prefix...)
+}
+
+func (s *debugServer) collectDatabaseTableSizes(ctx context.Context, tw *tar.Writer, prefix ...string) error {
+	return s.collectDatabaseQuery(ctx, tw, "table-sizes", s.queryDatabaseTableSizes, prefix...)
+}
+
+func (s *debugServer) collectDatabaseQuery(ctx context.Context, tw *tar.Writer, fileName string,
+	queryFunc func(ctx context.Context, db *pachsql.DB, w io.Writer) error, prefix ...string) error {
```
jrockway, line 976: general comment... sometimes it's helpful to split these function literals into their own type:

```go
type queryFunc func(context.Context, *pachsql.DB, io.Writer) error
func (s *debugServer) collectDatabaseQuery(ctx context.Context, tw *tar.Writer, fileName string,
	fn queryFunc, prefix ...string) error {
```
I don't think it adds any value in this case, though.

jrockway (changes requested):
Looking very good.  Going to "request changes" for this early PR but for comments like this in the future I will just approve in this situation ;)

src/server/debug/server/server.go
```diff
@@ -927,3 +946,101 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+		if err := sqlx.MapScan(rows, row); err != nil {
+			return errors.EnsureStack(err)
+		}
+		processedRows = append(processedRows, row)
+	}
+	jsonQueryResponse, err := json.MarshalIndent(processedRows, "", "\t")
+	if err != nil {
+		return errors.EnsureStack(err)
+	}
+	_, err = w.Write(jsonQueryResponse)
```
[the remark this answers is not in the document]
FahadBSyed: I enjoy learning about this stuff! Thanks

src/server/debug/server/server.go
```diff
@@ -927,3 +946,101 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+	return s.queryDatabase(ctx, query, w)
+}
+
+func (s *debugServer) queryDatabase(ctx context.Context, query string, w io.Writer) error {
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+	rows, err := s.database.QueryContext(ctxWithTimeout, query)
+	if err != nil {
+		return errors.EnsureStack(err)
+	}
```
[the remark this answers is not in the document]
FahadBSyed: Oh good point, that could have been had 😆

src/server/pachyderm_test.go
```diff
@@ -10534,3 +10536,95 @@ func TestMissingSecretFailure(t *testing.T) {
...
+
+func validateRowCounts(t *testing.T, buff bytes.Buffer, numCommits int) {
+	rows := make([]rowCountResults, 0)
+	require.NoError(t, json.Unmarshal(buff.Bytes(), &rows))
+
+	for _, row := range rows {
+		if row.RelName == "commits" {
+			require.Equal(t, row.NLiveTup, numCommits)
+		}
+	}
```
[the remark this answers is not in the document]
FahadBSyed: Good catch, I added some logic to fail the test if we never enter that for-loop

src/server/debug/server/server.go
```diff
@@ -927,3 +946,104 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+		WHERE state != 'idle'
+		ORDER by runtime DESC;
+	`
+	return s.queryDatabase(ctx, query, w)
+}
+
+func (s *debugServer) queryDatabaseTableSizes(ctx context.Context, db *pachsql.DB, w io.Writer) error {
+	query := `
+		SELECT nspname || '.' || relname AS "relation",
+		  pg_size_pretty(pg_total_relation_size(C.oid)) AS "total_size"
```
jrockway, line 1007: I'd rather we didn't make the sizes pretty here.  I think an integer number of bytes is easier to work with.

src/server/debug/server/server.go
```diff
@@ -927,3 +946,104 @@ func redirectDump(ctx context.Context, c debug.DebugClient, filter *debug.Filter
...
+		FROM pg_stat_activity
+		WHERE state != 'idle'
+		ORDER by runtime DESC;
+	`
+	return s.queryDatabase(ctx, query, w)
+}
+
+func (s *debugServer) queryDatabaseTableSizes(ctx context.Context, db *pachsql.DB, w io.Writer) error {
+	query := `
+		SELECT nspname || '.' || relname AS "relation",
```
jrockway, line 1006: This is inconsistent with the other queries.  They return {"schemaname": ..., "relname": ...}, but this one returns them concatenated together.  Maybe not a big deal, but consistency can be good.

src/debug/debug.proto
```diff
@@ -23,6 +23,7 @@ message Filter {
     bool pachd = 1;	
     pps_v2.Pipeline pipeline = 2;	
     Worker worker = 3;
+    bool database = 4;
```
jrockway, line 26: I think this maybe shouldn't be a filter yet.  Filter implies "collect one of these things", and we are intending it to be part of the pachd dump.  Maybe when we do the full database dump, we'll want to make that an option for the pachd dump?  (Annoyingly, that's going to be hard to do.  This proto should probably be `oneof filter { PachdDumpConfig pachd = 1; ... }` and not just a raw bool.  Sigh!)

jrockway (approved):
Looks good!

jrockway (approved):

src/server/debug/server/server.go
```diff
@@ -39,22 +43,25 @@ const (
...
 
 // NewDebugServer creates a new server that serves the debug api over GRPC
-func NewDebugServer(env serviceenv.ServiceEnv, name string, sidecarClient *client.APIClient) debug.DebugServer {
+func NewDebugServer(env serviceenv.ServiceEnv, name string, sidecarClient *client.APIClient, db *pachsql.DB) debug.DebugServer {
 	return &debugServer{
 		env:           env,
 		name:          name,
 		sidecarClient: sidecarClient,
 		marshaller:    &jsonpb.Marshaler{Indent: "  "},
+		database:      db,
```
acohen4, line 64: maybe I missed something but why can't we extract db from env.GetDBClient() here?

src/server/debug/server/server.go
```diff
@@ -82,7 +90,14 @@ func (s *debugServer) handleRedirect(
...
 				case *debug.Filter_Pachd:
-					return collectPachd(ctx, tw, pachClient, pachdContainerPrefix)
+					if err := collectPachd(ctx, tw, pachClient, pachdContainerPrefix); err != nil {
+						return err
+					}
+					if wantDatabase {
+						if err := s.collectDatabaseStats(ctx, tw, databasePrefix); err != nil {
+							return err
+						}
+					}
```
acohen4, line 100: I think we want to return s. collectDatabaseStats here even if it's not an error right?

src/server/debug/server/server.go
FahadBSyed: Can do

src/server/pachyderm_test.go
```diff
@@ -10534,3 +10537,86 @@ func TestMissingSecretFailure(t *testing.T) {
...
+
+		err = c.PutFile(commit, "/file", bytes.NewBufferString("file contents"))
+		require.NoError(t, err, "should be able to add a file to a commit")
+
+		err = c.FinishCommit(repoName, branchName, commit.ID)
+		require.NoError(t, err, "finishing a commit should succeed")
+	}
+}
+
+func validateRowCounts(t *testing.T, buf bytes.Buffer) {
```
acohen4, line 10607: could you move this function definition into TestDatabaseStats since it's only being used in one place. The only reason it's advisable here is because this file has SO many tests it's nice when most of the test definition is in one place.

src/server/debug/server/server.go
FahadBSyed: consistency is definitely important, will update this

src/server/debug/server/server.go
acohen4: I think we would mock env's GetDBClient in that case. I don't think we should expand the signature here

acohen4 (approved):
Overall LGTM! There are a few small comments that should still be addressed. 

Nice job here :)

src/server/debug/server/server.go
jrockway: Yeah we have a serviceenv that works for testing, so in theory not necessary, but I think less coupling is better in general.  We already have the problem where the serviceenv can't actual return built servers (because of a dependency cycle).  If the servers were built by the serviceenv instead of building themselves from a serviceenv, we wouldn't have that problem.

src/debug/debug.proto
FahadBSyed: You make a good point, but having a filter just for database files does help when testing

src/server/pachyderm_test.go
```diff
@@ -10534,3 +10537,86 @@ func TestMissingSecretFailure(t *testing.T) {
...
+
+		if hdr.Name == "database/row-counts.json" {
+			validateRowCounts(t, *fileContents)
+		}
+
+		return nil
+	}))
+}
+
+func createTestCommits(t *testing.T, repoName, branchName string, numCommits int, c *client.APIClient) {
```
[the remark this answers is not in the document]
FahadBSyed: Sure, can do

src/server/pachyderm_test.go
FahadBSyed: Yup, for this file that makes sense

PR #7763, opened 2022-06-13
[CORE-662] Add codecov for unit tests. Separate out unit tests using buildtags

armaanv:
This creates a new 'test-go' circleci job that runs the unit tests. Currently, most of the failed tests in that job depend on dockertestenv. I am considering switching the `k8s` tag to an `integration` test tag and including the dockertestenv tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,8 +28,8 @@ parameters:
 
 jobs:
   test-go:
-    docker:
-      - image: cimg/golang:1.17
+    machine:
+      image: << pipeline.parameters.machine_image >>
     resource_class: xlarge
     environment:
       TEST_RESULTS: /tmp/test-results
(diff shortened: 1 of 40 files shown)


.circleci/config.yml
```diff
@@ -26,6 +27,41 @@ parameters:
     default: ""
 
 jobs:
+  test-go:
+    docker:
+      - image: circleci/golang:1.17
```
jrockway, line 32: should upgrade to the cimg/golang:1.17 image to avoid warnings about this in the UI.  doing this will also change the location of go packages to cache (think they are now in /home/circleci/go instead of /go).

jrockway (approved):
Mechanically this is all looking extremely good.  There are still some failures. 

1. We use Docker to start postgres, mysql, minio, and maybe other things.  I think we need `setup_remote_docker` for that to work.  (This seems to be the bulk of the failures.)  Personally, do not like the approach of setting these up in the tests; would prefer instructions that tell people "setup postgres, mysql, and minio, then run the tests" so that we don't couple to details of docker.  (Sadly there are two implementations of "run something in Docker" too.)  Then in CI, we'd just `-run` those things, and have separate log streams, etc.  But, not for this PR to clean up, so try enabling the remote docker and cross your fingers.
2. I started a numbered list of things to fix because I thought there would be a lot, but it's just that first thing that's causing all the failures I think 😂

.circleci/config.yml
chainlink: I've noticed that running docker in docker machine on circle is much faster than setup_remote_docker since remote docker uses a tiny separate instance (that's size is not configurable) Generally agree with Jon that these deps should be explicit and passed in. Here's an example of running postgres along side in circle https://circleci.com/docs/2.0/databases/

.circleci/config.yml
```diff
@@ -26,6 +27,41 @@ parameters:
...
+    resource_class: xlarge
+    environment:
+      TEST_RESULTS: /tmp/test-results
+    steps:
+      - checkout
+      - run: mkdir ${TEST_RESULTS}
+      - run: go install gotest.tools/gotestsum@latest
+      - run: CGO_ENABLED=0 go install ./src/server/cmd/pachctl
```
chainlink, line 40: Do we need to install pachctl?

chainlink:
Perhaps it makes sense to tag the tests requiring a DB separately as well?

chainlink:
Perhaps it makes sense to tag the tests requiring a DB separately as well?

PR #7765, opened 2022-06-13
envoy: use the default idle_timeout

jrockway:
For some reason, I set a very aggressive `idle_timeout`.  This doesn't cause any harm, but it's annoying.  `debug-dump-server` polls Pachyderm every 60 seconds and it always has to create a new channel.   That doesn't cause any problems (the RPC proceeds as normal transparently), but it's kind of a waste of resources.  I feel like I'm probably not the only person who will want to keep an idle connection around for a while, so I've reverted this to the default of 1 hour:

https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts

Do note that this timeout is only for channels with no RPCs whatsoever; things like WaitCommit, PutFile, etc. are not considered idle.  So there is no harm in this being 60s, it just bugs me.

### DIFF
--- a/etc/generate-envoy-config/envoy.libsonnet
+++ b/etc/generate-envoy-config/envoy.libsonnet
@@ -105,7 +105,7 @@
       codec_type: 'auto',
       common_http_protocol_options: {
         headers_with_underscores_action: 'REJECT_REQUEST',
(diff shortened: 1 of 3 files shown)


acohen4 (approved):
LGTM

PR #7771, opened 2022-06-14
NewAuthServer should include EnvBootstrap

acohen4:


src/server/auth/server/api_server.go
```diff
@@ -83,8 +83,13 @@ type apiServer struct {
 	watchesEnabled bool
 }
 
+type InternalAuthServer struct {
+	APIServer    authiface.APIServer
+	EnvBootstrap func(context.Context) error
+}
```
jrockway, line 89: This is really a lot of work to go to to make a method in `apiServer` public.  Would it be unreasonable to rename `apiServer` to `APIServer` and  `s.envBootstrap` to `s.EnvBootstrap` and skip this indirection?  (Now that I think about it, not strictly necessary to make apiServer public... but I hate constructors that return private types, because then you can only create variables that contain them with `:=`.)

src/server/auth/server/api_server.go
```diff
@@ -83,8 +83,13 @@ type apiServer struct {
 	watchesEnabled bool
 }
 
+type InternalAuthServer struct {
+	APIServer    authiface.APIServer
```
jrockway, line 87: I think you might want to do `type InternalAuthServer struct { authiface.APIServer; EnvBootstrap ... }` so that's it's explicit we're trying to wrap an authiface.APIServer here.  You can then use `InternalAuthServer` like it's any `authiface.APIServer`, and avoid most of the refactoring below.

jrockway (approved):
Some suggestions below, but not blocking by any means.

PR #7773, opened 2022-06-14
Decouple minikubetestenv address from pachctl config

acohen4:


src/internal/minikubetestenv/deploy.go
```diff
@@ -85,20 +87,25 @@ func helmChartLocalPath(t testing.TB) string {
...
-			copy := grpcutil.DefaultPachdAddress
-			computedPachAddress = &copy
+	addr := grpcutil.DefaultPachdAddress
+	if *hostOverride != "" {
+		addr.Host = *hostOverride
+	} else if exposedServiceType() == "NodePort" {
+		in := new(bytes.Buffer)
+		in.WriteString("minikube ip")
+		out := new(bytes.Buffer)
+		cmd := exec.Command("/bin/bash")
```
jrockway, line 97: Any reason we need the indirection through the shell here?  If we do need it, you can also `strings.NewReader("minikube ip")` to save one line of code vs bytes.buffer ;)

jrockway (approved):
Excellent idea.  I actually have a pach context "tests" that I switch to before running the tests to configure this... sometimes I forget and it being automated will be nice.

PR #7779, opened 2022-06-15
fix ci deploy test reporting

molinamelendezj:
fixes issue with test failures on ci. test results upload was recently fixed, a pipe was added to the command `go test`, but missed adding the `pipefail` in the deploy test script so failed tests can fail the workflow.


acohen4 (approved):
LGTM

PR #7782, opened 2022-06-16
Avoid scrubbing get file error message.

PFedak:
Mostly revert #6668, which threw away the original error message in
favor of some not-entirely-clear advice. Also point users directly to -r
in multi-file cases, and clarify the help text.

### DIFF
--- a/src/server/pfs/server/source.go
+++ b/src/server/pfs/server/source.go
@@ -5,11 +5,13 @@ import (
 	"path"
 	"strings"
 
+	"golang.org/x/net/context"
+
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset"
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset/index"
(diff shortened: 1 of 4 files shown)


src/server/pfs/server/source.go
```diff
@@ -200,13 +200,13 @@ func checkSingleFile(ctx context.Context, src Source) error {
...
 		if finfo.FileType != pfs.FileType_FILE {
-			return errors.Errorf("cannot get non-regular file. Try GetFileTAR for directories")
+			return errors.Errorf("cannot get directory or non-regular file. Use GetFileTAR or the -r flag")
 		}
 		if count == 0 {
 			count++
 			return nil
 		}
-		return errors.Errorf("matched multiple files. Try GetFileTAR")
+		return errors.Errorf("matched multiple files. Use GetFileTAR or the -r flag")
```
acohen4, line 209: I feel like we shouldn't reference the CLI flag in the server error. WDYT?

PFedak: Yeah, that's probably true. Do you think I should try to detect and replace it in the CLI instead? I think mentioning `GetFileTAR` in a CLI error is similarly inappropriate

PR #7786, opened 2022-06-16
[CORE-444] Add code to dump a database using pagination

FahadBSyed:


src/internal/pachsql/db.go
```diff
@@ -58,6 +70,102 @@ func OpenURL(u URL, password string) (*DB, error) {
...
+			return nil, errors.EnsureStack(err)
+		}
+		processedRows = append(processedRows, row)
+	}
+	return processedRows, nil
+}
+
+// GetMainColumn returns the 'key' column if it exists for SchemaTable table. Otherwise, it queries the db and returns
+// the column_name where ordinal_position is 1. This column can be used to order future queries against this table.
+func GetMainColumn(ctx context.Context, db *DB, table *SchemaTable) (string, error) {
```
acohen4, line 125: how about KeyColumn()?

src/internal/pachsql/db.go
```diff
@@ -58,6 +70,102 @@ func OpenURL(u URL, password string) (*DB, error) {
...
+		return "", err
+	}
+	if keyExists {
+		return "key", nil
+	}
+	query := fmt.Sprintf(`
+		SELECT column_name FROM INFORMATION_SCHEMA.COLUMNS
+		WHERE table_schema = '%s'
+		  AND table_name = '%s'
+		  AND ordinal_position = 1;
```
acohen4, line 137: is Ordinal position guaranteed to give us the Primary key here?

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,251 @@
...
+func (s *debugServer) collectTable(ctx context.Context, w io.Writer, table *pachsql.SchemaTable) error {
+	column, err := pachsql.GetMainColumn(ctx, s.database, table)
+	if err != nil {
+		return errors.EnsureStack(err)
+	}
+	rows, err := s.getFirstPage(ctx, column, pageSize, table)
+	if err != nil {
+		return errors.EnsureStack(err)
+	}
+	if len(rows) == 0 {
```
acohen4, line 134: should this be `if len(rows < pageSize)`?

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,251 @@
...
+		return writeAllRowsToJSON(rows, w)
+	} else if err := writeRowPageToJSON(rows, w, firstPage); err != nil {
+		return errors.EnsureStack(err)
+	}
+	queryTemplate := fmt.Sprintf(`
+		SELECT * FROM %s.%s WHERE %s > {{lastitem}}
```
acohen4, line 140: this assumes that the WHERE column is unique. Is that a fair assumption?

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,251 @@
...
+	`, table.SchemaName, table.TableName, column, pageSize)
+	pageRows, err := pachsql.GetScannedRows(ctx, s.database, query)
+	if err != nil {
+		return nil, errors.EnsureStack(err)
+	}
+	return pageRows, nil
+}
+
+func setToStringBasedOnType(item interface{}) func(item interface{}) string {
+	switch fmt.Sprint(reflect.TypeOf(item)) {
```
acohen4, line 181: it's standard to type switch like this: https://go.dev/tour/methods/16

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,251 @@
...
+	case "time.Time":
```
acohen4, line 182: it's not clear to me whether the "main" column should have type of time

src/internal/pachsql/db.go
FahadBSyed: Sure can do

src/internal/pachsql/db.go
```diff
@@ -58,6 +70,102 @@ func OpenURL(u URL, password string) (*DB, error) {
...
+	defer rows.Close()
+	var tables []SchemaTable
+	if err := sqlx.StructScan(rows, &tables); err != nil {
+		return nil, errors.EnsureStack(err)
+	}
+	return tables, nil
+}
+
+// GetScannedRows queries db and calls MapScanRows on the response rows.
+func GetScannedRows(ctx context.Context, db *DB, query string) ([]RowMap, error) {
```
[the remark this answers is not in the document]
FahadBSyed: Sure, I can change the name here

src/internal/pachsql/db.go
FahadBSyed: You're right that it doesn't work in general. My solution was to search for a 'key' column and if it didn't exist, then to default to whatever is in ordinal_postion = 1. That seems to work fine for the set of tables we have currently

src/server/debug/server/database_dump.go
FahadBSyed: Yes it does assume that. The assumption should hold up if our KeyColumn() is unique. I _believe_ it is given the tables we have today

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+		return nil
+	default:
+		return errors.New("page should be either firstPage, middlePage, or lastPage")
+	}
+	return writeRowsToJSON(rowMap, w, replacerOpts...)
+}
+
+func writeRowsToJSON(rowMap []pachsql.RowMap, w io.Writer, opts ...string) error {
+	marshalledRows, err := json.Marshal(rowMap)
```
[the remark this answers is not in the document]
FahadBSyed: Sure, I can try that out and see if that simplifies things. We'll have slightly more overhead since we are calling marshal() more, but that probably won't be the bottleneck here

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+func writeAllRowsToJSON(rowMap []pachsql.RowMap, w io.Writer) error {
+	return writeRowsToJSON(rowMap, w, "[{", "[\n\t{", "}]", "}\n]\n", "},", "},\n\t")
+}
+
+func writeRowPageToJSON(rowMap []pachsql.RowMap, w io.Writer, p page) error {
+	var replacerOpts []string
+	switch p {
+	case firstPage:
```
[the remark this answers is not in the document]
FahadBSyed: They are slightly different (first page has a '[' in one of the cases, but I think I am going to implement your simplification suggestion below which means I'll scrap this part of the code altogether.

src/server/pachyderm_test.go
```diff
@@ -10620,11 +10621,11 @@ func TestDatabaseStats(t *testing.T) {
 		hdr, err := f.Header()
 		require.NoError(t, err, "getting database tar file header should succeed")
 		require.NotMatch(t, "^[a-zA-Z0-9_\\-\\/]+\\.error$", hdr.Name)
-
 		if hdr.Name == "database/row-counts.json" {
```
jrockway, line 10624: This wants to be a switch hdr.Name { case ... } statement now.

src/server/debug/server/database_dump.go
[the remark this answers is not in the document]
FahadBSyed: I did this refactor and while it is 'about' as complicated, it is definitely more readable.

src/server/pachyderm_test.go
```diff
@@ -10634,7 +10635,14 @@ func TestDatabaseStats(t *testing.T) {
...
-			require.Equal(t, true, commitsFound, "we should have an entry in row-counts.json for commits")
+			require.Equal(t, true, commitsFound,
+				"we should have an entry in row-counts.json for commits")
+		}
+
+		if hdr.Name == "database/tables/collections/commits.json" {
+			require.NoError(t, json.Unmarshal(fileContents.Bytes(), &rows),
+				"unmarshalling commits.json should succeed")
+			require.Equal(t, numCommits, len(rows), "number of commits should match number of rows")
 		}
```
jrockway, line 10646: This test will still pass if given an empty tar file.  You need to keep some state like `var foundRowCounts, foundCommitsJSON`, set them to true in each branch, and check at the end of the test that they're true.  You don't want to test "if there's a file called database/row-counts.json it's valid", you want to test "there must be a file called database/row-counts.json and it must be valid".

jrockway (changes requested):
I think this can all be a little simpler.  Also what happened to the approach of bundling pg_dump and just running that?  That feels like the most reliable way to dump a database.

src/server/pachyderm_test.go
FahadBSyed: True!

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,251 @@
...
+
+const (
+	firstPage  page = "firstPage"
+	middlePage page = "middlePage"
+	lastPage   page = "lastPage"
+	pageSize        = 500
+)
+
+func (s *debugServer) collectDatabaseDump(ctx context.Context, tw *tar.Writer, prefix ...string) error {
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, 1*time.Minute)
```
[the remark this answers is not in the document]
FahadBSyed: I ran some load tests on pachd to try to simulate a real workload. In 1 minute we could collect most of the tables, but not all of them. In that time, we managed to get about 270000 rows in the chunk_objects.json table before our deadline exceeded,

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,251 @@
...
+		SELECT nspname AS "schemaname", relname, pg_total_relation_size(C.oid) AS "total_size"
+		FROM pg_class C
+		  INNER JOIN pg_namespace N ON (N.oid = C.relnamespace)
+		WHERE nspname NOT IN ('pg_catalog', 'information_schema')
+		  AND C.relkind <> 'i'
+		  AND nspname !~ '^pg_toast'
+		ORDER BY nspname, relname;
+	`
+	return s.writeQueryResponseToJSON(ctx, query, w)
+}
```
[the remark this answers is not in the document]
FahadBSyed: Yeah, I think that is better for extension if we wanted more stats. In the current approach we'd have to define two more functions and then add them to the stats() function. Here instead, we can just add a new entry to the map.

src/server/debug/server/database_dump.go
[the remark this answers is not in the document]
jrockway: ks That's pretty good.  In the future, we probably want to do a lot of our checks concurrently, so can just set a deadline in the request for how long those should run.  (I noticed that when we collect CPU profiles, we do them sequentially; first pachd for 30s, then worker 1 for 30s, then worker 2 for 30s, etc.  that should all be done in parallel.)

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,251 @@
...
+			}
+			return nil
+		}, fullPrefix); err != nil {
+			return errors.EnsureStack(err)
+		}
+	}
+	return nil
+}
+
+func (s *debugServer) collectTable(ctx context.Context, w io.Writer, table *pachsql.SchemaTable) error {
```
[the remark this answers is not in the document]
FahadBSyed: This ended up being way simpler and the buffered writer also made the performance way better!

src/internal/pachsql/db.go
```diff
@@ -58,6 +92,85 @@ func OpenURL(u URL, password string) (*DB, error) {
...
 
+// ListTables returns an array of SchemaTable structs that represent the tables.
+func ListTables(ctx context.Context, db *DB) ([]SchemaTable, error) {
+	rows, err := db.QueryContext(ctx, listTablesQuery)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not list tables")
+	}
+	defer rows.Close()
+	var tables []SchemaTable
+	if err := sqlx.StructScan(rows, &tables); err != nil {
```
jrockway, line 103: I think sqlx.SelectContext will handle iterating over all of the rows and struct-scanning into each individual slice element.

src/internal/pachsql/db.go
```diff
@@ -58,6 +92,85 @@ func OpenURL(u URL, password string) (*DB, error) {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to scan rows")
+	}
+	return scannedRows, nil
+}
+
+// MapScanRows iterates through rows and returns an array of sqlx.MapScan results. This is useful
+// for when you want to capture the results, but don't know what columns the query can return
+// or when you don't want to create a struct to represent the columns.
+func MapScanRows(rows *sql.Rows) ([]RowMap, error) {
```
jrockway, line 126: Yeah, I think this is just sqlx.Select.

src/internal/pachsql/db.go
```diff
@@ -58,6 +92,85 @@ func OpenURL(u URL, password string) (*DB, error) {
...
+		if err := sqlx.MapScan(rows, row); err != nil {
+			return nil, errors.EnsureStack(err)
+		}
+		processedRows = append(processedRows, row)
+	}
+	return processedRows, nil
+}
+
+// KeyColumn returns the 'key' column if it exists for SchemaTable table. Otherwise, it queries the db and returns
+// the column_name where ordinal_position is 1. This column can be used to order future queries against this table.
```
jrockway, line 139: Thinking about this a little more, if you guess this wrong, then there won't be an index that lets the database engine retrieve the rows in this order, so it will have to sort them all before returning them to you.  To be honest, I'm not sure why the debug output needs to be in a particular order.

src/internal/pachsql/db_test.go
```diff
@@ -0,0 +1,167 @@
...
+	require.NoError(t, err)
+	defer mockDB.Close()
+	mockRows := sqlmock.NewRows([]string{schemaName, tableName}).
+		AddRow(collections, commits).
+		AddRow(pfs, commitDiffs)
+	mock.ExpectQuery("").WillReturnRows(mockRows)
+	sqlxDB := sqlx.NewDb(mockDB, mockDBName)
+	rows, err := ScannedRows(context.Background(), sqlxDB, "")
+	require.NoError(t, err)
+	require.Equal(t, 2, len(rows))
```
jrockway, line 95: I am not sure length + spot check is a particularly useful test.  Just check that the rows are what you expect them to be: https://github.com/golang/go/wiki/TestComments#compare-full-structures

src/internal/pachsql/db_test.go
```diff
@@ -0,0 +1,167 @@
...
+	require.Equal(t, rows[0][schemaName], "collections")
+}
+
+func scannedRowsInvalidResponse(t *testing.T) {
+	mockDB, _, err := sqlmock.New()
+	require.NoError(t, err)
+	defer mockDB.Close()
+	sqlxDB := sqlx.NewDb(mockDB, mockDBName)
+	_, err = ScannedRows(context.Background(), sqlxDB, "")
+	require.Equal(t, true, err != nil && strings.HasPrefix(err.Error(), "failed to query database:"),
```
jrockway, line 105: There is no way to debug this test if it fails.  It will print "false != true" when you really want it to print the actual error message that was returned.

I tend to do this: https://github.com/jrockway/json-logs/blob/master/pkg/parse/parse_test.go#L969.  Basically, `cmp.Diff` on the received error vs a special type that matches a regex on the error (or an exact error if you want).

For this codebase, I'd require.YesError, and then require.Matches, though.

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,139 @@
...
+	var errs []error
+	for filename, query := range queries {
+		if err := collectDebugFile(tw, filename, "json", func(w io.Writer) error {
+			rows, err := s.database.QueryContext(ctx, query)
+			if err != nil {
+				return errors.EnsureStack(err)
+			}
+			return s.writeRowsToJSON(rows, w)
+		}, prefix...); err != nil {
+			errs = append(errs, errors.Wrapf(err, "failed to query %q", filename))
```
jrockway, line 58: Errors are implicitly failures, so there is no need to add that to the message.  Just "execute query %v" is fine. Being really nitpicky, I also don't see a lot of value in quoting the filename.  They contain no special characters.

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}
+	if len(errs) > 0 {
+		return fmt.Errorf("%v", errs)
```
jrockway, line 62: I'd throw in an errors.EnsureStack here for consistency.

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (s *debugServer) collectDatabaseTables(ctx context.Context, tw *tar.Writer, prefix ...string) error {
+	tables, err := pachsql.ListTables(ctx, s.database)
+	if err != nil {
+		return errors.EnsureStack(err)
+	}
+	for _, table := range tables {
+		fullPrefix := strings.Join(append(append([]string{}, prefix...), "tables", table.SchemaName), "/")
+		if err := collectDebugFile(tw, table.TableName, "json", func(w io.Writer) error {
+			bufferedWriter := bufio.NewWriter(w)
```
jrockway, line 75: You don't need to change anything here, but when I immediately wrap a function argument, I give the argument a weird name and the wrapped version a normal name.  So I'd do `func(rawWriter io.Writer) { w := bufio.NewWriter(rawWriter) }`.  You use things like `w` for an io.Writer and `ctx` for a context.Context so frequently that it's easy to accidentally use the unwrapped one if it has an unusual name, and it's one of those things that you won't notice and code reviewers won't notice because of how the brain works.

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,139 @@
...
+			if err := s.collectTable(ctx, bufferedWriter, &table); err != nil {
+				return errors.Wrapf(err, "error collecting table %s.%s", table.SchemaName, table.TableName)
```
jrockway, line 77: There are a bunch of these so I'll mention it one more time, but this can be `collect table %s.%s`; we know it's an error because it's of type `error`.

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	return nil
+}
+
+func (s *debugServer) collectTable(ctx context.Context, w io.Writer, table *pachsql.SchemaTable) error {
+	column, err := pachsql.KeyColumn(ctx, s.database, table)
+	if err != nil {
+		return errors.Wrap(err, "error getting key column")
+	}
+	query := fmt.Sprintf(`SELECT * FROM %s.%s ORDER BY %s ASC`, table.SchemaName, table.TableName, column)
```
jrockway, line 95: It's pretty dangerous to interpolate strings that originate from outside of the program directly into SQL queries.  Normally you'd use placeholders, but you can't for table and column names, so you'll have to figure out how to hook in something like https://pkg.go.dev/gopkg.in/jackc/pgx.v1#Connection.QuoteString.

It looks like you can do this:

```
In Go 1.13 and above (*sql.Conn) Raw() can be used to get a *pgx.Conn from
the standard database/sql.DB connection pool. This allows operations that
use pgx specific functionality.

    // Given db is a *sql.DB
    conn, err := db.Conn(context.Background())
    if err != nil {
        // handle error from acquiring connection from DB pool
    }

    err = conn.Raw(func(driverConn interface{}) error {
        conn := driverConn.(*stdlib.Conn).Conn() // conn is a *pgx.Conn
        // Do pgx specific stuff with conn
        conn.CopyFrom(...)
        return nil
    })
    if err != nil {
        // handle error that occurred while using *pgx.Conn
    }
```

I am not sure why our code tries to hide the fact that we use pgx.  It's very annoying in this case.

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,139 @@
...
+			return errors.Wrap(err, "could not marshal row")
+		}
+		if _, err := w.Write(jsonRow); err != nil {
+			return errors.Wrap(err, "could not write table")
+		}
+		wroteFirstRow = true
+	}
+	if err := rows.Err(); err != nil {
+		if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+			w.Write([]byte(",\n\t{\"message\": \"result set was truncated.\"}\n]"))
```
jrockway, line 130: You should put the error message in here.  %q is probably sufficient to be valid JSON.

I'm not sure that DeadlineExceeded and Canceled are the only errors where you'd want to return what you already have.  In theory there could be errors like "server is shutting down", and it would be nice to return what we have so far in that case.   I think if you've written any rows, you might as well write the error message to the file and include what you have so far.  (Check the error on w.Write here; if it's a write error then that error needs to be handled a level up, where it will write error.txt complaining about the write error.)

src/server/debug/server/database_dump.go
```diff
@@ -0,0 +1,139 @@
...
+			return nil
+		}
+		return errors.Wrap(err, "error iterating through rows")
```
jrockway, line 133: This is an error *after* iterating through rows.  This would be a read error that makes rows.Next falsely determine that there are no more rows, I guess.

jrockway (approved):
Very nice!

src/internal/pachsql/db.go
FahadBSyed: Yeah that's true, we probably don't need to sort the response at all and can just iterate via rows.Next()

src/internal/pachsql/db.go
FahadBSyed: Nice, I should have read this library more before I wrote these functions. That means I can get rid of some of these

src/server/debug/server/database_dump.go
FahadBSyed: That convention makes a lot of sense. I like it!

src/internal/pachsql/db_test.go
FahadBSyed: Deep equals would be better

src/server/debug/server/database_dump.go
FahadBSyed: So, I decided to jump to see QuoteString()'s definition and include that since it was pretty simple and call that on the inputs before passing them into the template

src/internal/pachsql/db.go
FahadBSyed: We ended up remove these

PR #7788, opened 2022-06-17
Mount server - API to support refresh button in Jupyter extension

lukemarsden:
In the Jupyter mount extension, we want to display how many commits behind HEAD on a given Pachyderm branch the currently-mounted commit is.

This is the backend API side change to enable this.

After this PR is merged, we should bump the version of the mount-server used in https://github.com/pachyderm/jupyterlab-pachyderm/pull/143 and then we can merge that.


smalyala (approved):
Nice and clean! LGTM

PR #7789, opened 2022-06-17
[CORE-109] Index cache

brycemcanally:
This PR updates #6254 and adds a benchmark that shows the performance improvement for lexicographical order point queries (~250s to ~3s on my machine).


src/internal/storage/fileset/index/cache.go
```diff
@@ -0,0 +1,103 @@
...
+	pbr := pbutil.NewReader(br)
+	var pathOffsets []*pathOffset
+	for {
+		pathOffset := &pathOffset{
+			offset: len(data) - br.Len(),
+		}
+		idx := &Index{}
+		if err := pbr.Read(idx); err != nil {
+			break
+		}
```
PFedak, line 76: I guess you're right that it's not worth checking for EOF here

src/internal/storage/fileset/index/cache.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	return &Cache{
+		storage: storage,
+		cache:   lruCache,
+	}
+}
+
+func (c *Cache) Get(ctx context.Context, chunkRef *chunk.DataRef, filter *pathFilter, w io.Writer) error {
+	c.mu.Lock()
+	v, ok := c.cache.Get(string(chunkRef.Ref.Id))
```
PFedak, line 35: this would break if the same chunk is ever looked up with two offsets, right?

PFedak (approved):
LGTM, I'm a little uneasy about some of the tacit assumptions about fields being unused but I don't think there's really much risk there

src/internal/storage/fileset/index/reader.go
```diff
@@ -167,13 +169,16 @@ func (lr *levelReader) Read(data []byte) (int, error) {
...
-		r := lr.chunks.NewReader(lr.ctx, []*chunk.DataRef{lr.idx.Range.ChunkRef})
 		lr.buf = &bytes.Buffer{}
-		if err := r.Get(lr.buf); err != nil {
-			return err
-		}
+		// TODO: Clone?
+		chunkRef := lr.idx.Range.ChunkRef
 		// Skip offset bytes to get to first index entry in chunk.
-		lr.buf = bytes.NewBuffer(lr.buf.Bytes()[lr.idx.Range.Offset:])
+		chunkRef.OffsetBytes = lr.idx.Range.Offset
```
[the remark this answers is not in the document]
brycemcanally: Yes, index entries no longer span multiple chunks. Prior to https://github.com/pachyderm/pachyderm/pull/7442 they could, so we need to support that. I'm not sure what you mean by relying on `OffsetBytes` being zero for the chunk ref. The chunk ref is really a data ref that references the full index chunk, but we may modify the `OffsetBytes` to skip a set of bytes that correspond to an index entry that spans multiple chunks.

src/internal/storage/fileset/index/cache.go
brycemcanally: We always use the same offset for a cached index chunk. It is the offset of the first index entry that starts in the chunk.

PR #7796, opened 2022-06-20
Few changes to reduce per file overhead

brycemcanally:
This PR contains a few changes to reduce the per file overhead in the storage layer. The changes are:

- Removed the `proto.Clone` when setting up a file reader.
- Added a fast path when copying small files to avoid copying the data through a pipe. This involved refactoring the chunk data reader to be an `io.Reader`. I would like to eventually get rid of the chunk reader abstraction and instead have it be just the `Get` call on the chunk storage abstraction.
- Added skipping for the index merge when closing a file set writer when the files don't go to both the uploader and batcher (all of the files were under the batch threshold or all of the files were above the batch threshold).

These changes improved the performance of compaction tasks for a small file workload by ~33% (~30s to ~20s).

### DIFF
--- a/src/internal/storage/chunk/reader.go
+++ b/src/internal/storage/chunk/reader.go
@@ -1,13 +1,13 @@
 package chunk
 
 import (
+	"bytes"
 	"context"
 	"io"
 	"time"
 
 	"github.com/pachyderm/pachyderm/v2/src/internal/backoff"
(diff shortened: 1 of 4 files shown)


PFedak (approved):
LGTM, nice to see a noticeable change from small, local improvements like these

PR #7804, opened 2022-06-22
Bootstrap Embedded Pach Cluster's Auth via environment [CORE-583] [CORE-584]

acohen4:


etc/helm/pachyderm/templates/console/secret.yaml
```diff
@@ -12,9 +12,6 @@ metadata:
   name: pachyderm-console-secret
   namespace: {{ .Release.Namespace }}
 data:
-{{ if .Release.IsUpgrade }}
-  OAUTH_CLIENT_SECRET: {{ required "For an upgrade release, a value is required for console.config.oauthClientSecret" .Values.console.config.oauthClientSecret | toString | b64enc | quote }}
-{{ else }}
+  OAUTH_CLIENT_ID: {{ .Values.console.config.oauthClientID | b64enc | quote }}
```
acohen4, line 15: this may be a backwards compatability issue where we actually want to have them add this to their secret pre-upgrade

src/server/auth/server/api_server.go
```diff
@@ -262,7 +355,7 @@ func (a *AuthServer) expiredEnterpriseCheck(ctx context.Context, username string
 
 func (a *AuthServer) userHasExpiredClusterAccessCheck(ctx context.Context, username string) error {
 	// Root User, PPS Master, and any Pipeline keep cluster access
-	if username == auth.RootUser || strings.HasPrefix(username, auth.PipelinePrefix) {
+	if username == auth.RootUser || strings.HasPrefix(username, auth.PipelinePrefix) || strings.HasPrefix(username, auth.InternalPrefix) {
```
jrockway, line 358: How good do you feel about a prefix check?  Can my pachyderm username ever be `pipeline:foobar` (not with oauth, but maybe ldap if that's a connector we support)?

etc/helm/pachyderm/templates/pachd/identity-config.yaml
```diff
@@ -12,3 +12,18 @@ data:
...
     id_token_expiry: {{ .Values.oidc.IDTokenExpiry }}
     rotation_token_expiry: {{ .Values.oidc.RotationTokenExpiry }}
+  oidc-clients: |
+    - id: {{ .Values.pachd.oauthClientID }}
+      name: {{ .Values.pachd.oauthClientID }}
+      redirect_uris:
+      - {{ include "pachyderm.pachdRedirectURI" . }}
+      {{- if .Values.console.enabled}}
+      trusted_peers:
+      - {{ .Values.console.config.oauthClientID | quote }}
```
jrockway, line 22: Offtopic for this PR, but I'd love to be able to tack on my own trusted clients here.  For Hub, we had Jupyter on this list.  For our internal pachyderm instance, debug-dump-service is in here.

etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -224,12 +255,27 @@ spec:
...
+          {{- else }}
+          value: {{ required "To register this pachd as an enterprise member, the enterprise root token must be provided." .Values.pachd.enterpriseRootToken | quote }}
+          {{- end }}
+        {{- end }}
         {{ if .Values.global.postgresql.identityDatabaseFullNameOverride }}
         - name: IDENTITY_SERVER_DATABASE
           value: {{ .Values.global.postgresql.identityDatabaseFullNameOverride }}
         {{ end }}
+        - name: UPGRADE_NO_OP
+          value: {{ randAlphaNum 32 }}
```
jrockway, line 278: Customers definitely want this to happen (automatically restart pachd when the secrets change), but this will break argocd, because it just does "helm template" in a loop, and this will make every iteration look to argo like a new change.

I think we'll have to hash each of the secrets we generate and put them here instead.  That way only a legitimate edit of a relevant field in values.yaml triggers this automatic restart.  (Another example; if we set the resources on console to a new value, pachd shouldn't need to restart.)  

Great idea for a feature though.  Let's do it in another PR soon?

etc/helm/pachyderm/templates/enterprise-server/deployment.yaml
```diff
@@ -106,17 +106,50 @@ spec:
...
+        {{- if or (include  "pachyderm.mockIDPRoleBindings" .) .Values.pachd.pachAuthClusterRoleBindings }}
+        - name: AUTH_CLUSTER_RBAC
+          valueFrom:
+            secretKeyRef:
+              name: "pachyderm-auth"
+              key: "cluster-role-bindings"
+        {{- end }}
+        {{- end }}
+        {{- if .Values.console.enabled }}
+        - name: CONSOLE_OAUTH_ID
```
jrockway, line 140: In the enterprise case, can users have multiple consoles?  How do we distinguish them now?  Do they all just use the same OAuth client?

src/server/auth/server/api_server.go
```diff
@@ -124,35 +130,122 @@ func NewAuthServer(env Env, public, requireNoncriticalServers, watchesEnabled bo
...
-		}); err != nil {
-			if !errors.As(err, auth.ErrAlreadyActivated) {
+		ctx = internalauth.AsInternalUser(ctx, internalUser)
+		// handle auth activation
+		if a.env.Config.AuthRootToken != "" {
+			if _, err := a.Activate(ctx, &auth.ActivateRequest{
+				RootToken: a.env.Config.AuthRootToken,
+			}); err != nil {
+				if !errors.As(err, auth.ErrAlreadyActivated) {
+					return err
```
jrockway, line 145: Let's wrap this, and the (existing) error on line 151, and the PPS/PFS ActivateAuth calls.

src/server/auth/server/api_server.go
```diff
@@ -124,35 +130,122 @@ func NewAuthServer(env Env, public, requireNoncriticalServers, watchesEnabled bo
...
+				_, err := a.rotateRootToken(ctx,
+					&auth.RotateRootTokenRequest{
+						RootToken: a.env.Config.AuthRootToken,
+					})
 				return err
+			} else {
+				if a.env.Config.PachdSpecificConfiguration != nil {
+					if _, err := a.env.GetPfsServer().ActivateAuth(ctx, &pfs.ActivateAuthRequest{}); err != nil {
+						return errors.EnsureStack(err)
+					}
```
jrockway, line 156: What happens if the pod dies right here, i.e. PFS gets ActivateAuth called, but not PPS?

src/server/auth/server/api_server.go
```diff
@@ -124,35 +130,122 @@ func NewAuthServer(env Env, public, requireNoncriticalServers, watchesEnabled bo
...
+			existing, err := a.GetRoleBinding(ctx, &auth.GetRoleBindingRequest{
+				Resource: &auth.Resource{Type: auth.ResourceType_CLUSTER},
+			})
+			if err != nil {
+				return errors.Wrapf(err, "get cluster role bindings")
+			}
+			for p := range existing.Binding.Entries {
+				// `pach:` user role bindings cannot be modified
+				if strings.HasPrefix(p, auth.PachPrefix) || strings.HasPrefix(p, auth.InternalPrefix) {
+					continue
```
jrockway, line 225: Let's log something here, so users don't have to guess as to why this was skipped.  I could also be talked into crashing (config is broken, try again), but I'm fine with just a log.

jrockway (approved):
Sorry for the delay on the review.  This is great!

I have some questions; if the answer is "trust me it's fine" you don't have to answer :)

src/server/auth/server/api_server.go
acohen4: yeah this is a bug. need to think about how to best handle this

etc/helm/pachyderm/templates/enterprise-server/deployment.yaml
acohen4: so this is kind of annoying but they're expected to be unique client IDs. We don't have a good way right now of preventing collisions besides customers setting the IDs explictly in the case of enterprise server with many clusters

etc/helm/pachyderm/templates/pachd/deployment.yaml
acohen4: What if we wrap this env-var in an `isUpgrade` conditional? I think we should chat about the ArgoCD stuff more - as a part of the current approach helm will spit out new random values in the secrets each run if left unspecified- will that break an ArgoCD workflow?

src/server/auth/server/api_server.go
acohen4: have a follow up PR to fix this: 
https://github.com/pachyderm/pachyderm/pull/7817/files

src/server/auth/server/api_server.go
acohen4: the only ways customer can mint tokens is either via authenticating in which case we tack on a "user:" prefix to the provided username, or via robots in which case we tack on a "robot:" prefix. We definitely rely on this prefix prepending

PR #7805, opened 2022-06-22
helm: Add a setting to prevent console/pachd from using the external network for OAuth

jrockway:
This is the helm side of https://github.com/pachyderm/haberdashery/pull/658

Basically with this enabled `pachd` and `console` will always do oauth things over the internal network.  You can still have the actual Issuer be external, of course, but only a human user with their browser will ever use that URL.  A new setting that defaults to false enables this (overriding localhostIssuer if explicitly set).

There are probably some missing special cases (enterprise server only on the internal network, proxies that map company.com/pachyderm/dex to pachd:1658/dex), but I want to hold off on adding a config key to set these.   I'd like to gradually remove most of this explicit configuration and have one that just works.  But it will be a while.

I also regenerated the values schema to pick up a couple of recent changes.

### DIFF
--- a/etc/helm/pachyderm/templates/console/deployment.yaml
+++ b/etc/helm/pachyderm/templates/console/deployment.yaml
@@ -48,6 +48,10 @@ spec:
           value: {{ include "pachyderm.issuerURI" . | quote}}
         - name: REACT_APP_RUNTIME_ISSUER_URI
           value: {{ include "pachyderm.reactAppRuntimeIssuerURI" . | quote}}
+        {{- if .Values.oidc.avoidExternalNetwork }}
+        - name: DEX_BASE_URI
+          value: {{ include "pachyderm.dexBaseURI" . | quote}}
+        {{- end }}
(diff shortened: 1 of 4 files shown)


etc/helm/pachyderm/templates/console/deployment.yaml
```diff
@@ -48,6 +48,10 @@ spec:
           value: {{ include "pachyderm.issuerURI" . | quote}}
         - name: REACT_APP_RUNTIME_ISSUER_URI
           value: {{ include "pachyderm.reactAppRuntimeIssuerURI" . | quote}}
+        {{- if .Values.oidc.avoidExternalNetwork }}
+        - name: DEX_BASE_URI
```
acohen4, line 52: How is DEX_BASE_URI different than ISSUER_URI?

tybritten: With this approach issuer uri is always the fqdn the user accesses (ingress.host). Dex_base will be basically straight to pachd

jrockway:
I think it turns out that we don't need this.

PR #7807, opened 2022-06-22
[CORE-455] Implement PPS load test DAG functionality

brycemcanally:
This PR implements functionality for specifying repo / pipeline DAGs to the PPS load testing. The PPS load testing will create the repos / pipelines for the DAG and validate the full DAG as the load test runs. The DAG is specified by a file where each line contains a pipeline and its input pipelines (`pipeline: input1, input2,...`). Source repos are specified by lines with no inputs (`source:`). The order of the lines must correspond with a topological sort of the repos / pipelines in the DAG since the repos / pipelines are created in line order.


src/internal/ppsload/pipeline.go
```diff
@@ -0,0 +1,92 @@
...
+func createPipeline(pachClient *client.APIClient, repo string, inputRepos []string, parallelism int64, podPatch string) error {
+	var inputs []*pps.Input
+	for i, inputRepo := range inputRepos {
+		inputs = append(inputs, &pps.Input{
+			Pfs: &pps.PFSInput{
+				Name:      fmt.Sprint("input-", i),
+				Repo:      inputRepo,
+				Branch:    "master",
+				Glob:      "/*",
+				JoinOn:    "/*",
```
PFedak, line 73: I think this ends up joining everything on the same key, which ends up being a cross.

src/internal/pfsload/validator.go
```diff
@@ -63,7 +63,7 @@ func (v *Validator) Validate(client Client, commit *pfs.Commit) (retErr error) {
 		}
 	}
 	err := client.WaitCommitSet(commit.ID, func(ci *pfs.CommitInfo) error {
-		if ci.Commit.Branch.Repo.Type != pfs.UserRepoType {
+		if ci.Commit.Branch.Repo.Type != pfs.UserRepoType || ci.Origin.Kind == pfs.OriginKind_ALIAS {
```
PFedak, line 66: I assume this is just to avoid repeated work?

PFedak (approved):
Looks good other than tweaking the join to actually join

src/internal/ppsload/pipeline.go
brycemcanally: whoops, good catch.

src/internal/pfsload/validator.go
brycemcanally: No, the alias commits won't have the content we are checking for in the validator.

PR #7447, opened 2022-03-28
Remove error wrap checking from some internal interfaces.

PFedak:
Also remove a handful of extant EnsureStack calls.

PR #7817, opened 2022-06-24
Transactionify Auth activation in EnvBootstrap

acohen4:


src/server/pps/server/api_server.go
```diff
@@ -2944,42 +2938,41 @@ func (a *apiServer) DeleteAll(ctx context.Context, request *types.Empty) (respon
...
+func (a *apiServer) ActivateAuthInTransaction(ctx context.Context, txnCtx *txncontext.TransactionContext, req *pps.ActivateAuthRequest) (resp *pps.ActivateAuthResponse, retErr error) {
 	// Unauthenticated users can't create new pipelines or repos, and users can't
 	// log in while auth is in an intermediate state, so 'pipelines' is exhaustive
-	var eg errgroup.Group
-	pipelineInfo := &pps.PipelineInfo{}
-	if err := a.pipelines.ReadOnly(ctx).List(pipelineInfo, col.DefaultOptions(), func(string) error {
-		pipeline := proto.Clone(pipelineInfo).(*pps.PipelineInfo)
-		pipelineName := pipeline.Pipeline.Name
+	pi := &pps.PipelineInfo{}
+	if err := a.pipelines.ReadOnly(ctx).List(pi, col.DefaultOptions(), func(string) error {
```
jrockway, line 2957: What does ReadOnly here do?  Should the list operation somehow have the txnCtx transaction passed in?  (Looks like Collection.ReadWrite takes that, and we are writing to pipelines, though not to the collection of pipelines, so I'm not 100% sure.)

Real question is what happens if a pipeline is added while this code is running.  I don't know what Postgres does under the hood; if a collection is `select * from collections.pipeline`, does inserting a row into that table invalidate the transaction?

If there's a conflict, where do we retry this transaction?  (I was guessing that things like `updatePipeline` would have a retry loop since they take a function to do the update, but I didn't see one.)

src/server/pps/server/api_server.go
```diff
@@ -2944,42 +2938,41 @@ func (a *apiServer) DeleteAll(ctx context.Context, request *types.Empty) (respon
...
+			pi.AuthToken = token
+			return nil
+		}); err != nil {
+			return errors.Wrapf(err, "could not update \"%s\" with new auth token", pi.Pipeline.Name)
+		}
 		// put 'pipeline' on relevant ACLs
-		if err := a.fixPipelineInputRepoACLs(ctx, pipeline, nil); err != nil {
-			return err
+		if err := a.fixPipelineInputRepoACLsInTransaction(txnCtx, pi, nil); err != nil {
+			return errors.Wrapf(err, "fix repo ACLs for pipeline %q", pi.Pipeline.Name)
```
jrockway, line 2972: It would be nice to log that we've done all of this, but to some extent the logs would have to be done after the transaction commits to be strictly correct.  Maybe DEBUG-level log each pipeline that was updated, and the overall outcome of the transaction a couple levels up?

jrockway (approved):
This looks good to me but I'd also let Peter take a look.

It would be nice to have an API where we can add log messages to a transaction and print them during commit.  All of this stuff could benefit from some logging; otherwise we have no chance of helping a customer that sends us a debug dump when this isn't working.

src/server/pps/server/api_server.go
acohen4: this API is a little unusual - we don't expose List() on the ReadWrite collection because it's implemented across transactions due to buffering. This means that we are listing pipelines from outside a transaction and updating each of them inside one. This does open the possibility that a pipeline is created after the transaction starts and we fail to update it

PFedak (changes requested):
If the transaction interface we're providing doesn't allow doing things we want to take, or forces a particular tradeoff in what guarantees we can make, we should at least consider changing it. Here's my recollection of why we have these various constraints - some might be inaccurate, or no longer apply:

no `ctx` in `txnCtx`: this is to prevent blocking on external services during a transaction (to avoid extending the length of a transaction as well as accidentally repeating non-idempotent operations), as well as calls to other API endpoints (which would get an inconsistent view of pachyderm state, and iirc created issues around exhausting our transaction/connection pools, where everything could be stuck in a top-level call while no lower-level call could proceed). I think this restriction still applies, and if there are non-pachyderm operations we want to perform in a transaction then we could add a specific carveout to the interface.

no `List` in `ReadWriteCollection`: this was initially to avoid someone accidentally e.g. listing all commits and doing a nontrivial operation that would lead to a long/large transaction. I'm not sure this part matters. Subsequently we made `List` batched, I think this was specifically defending against connections being left open, but I think there's also a concern around the same connection exhaustion issue, where the row streaming takes up a slot that might end up preventing the per-row operations from being able to proceed.

My grasp on the exact intricacies around how open `List`-type queries interact with transaction and connection limits is not where I'd like it to be. We have multiple places where we want to transactionally perform a nontrivial operation for several items in a collection. In other settings we either use `GetByIndex` to build the set (which uses the same internal buffering as `List` but is actually executed in the transaction), or build it through other means (e.g. branch propagation) and assume it all fits in memory. 

My naive conclusion is that the only real reason against allowing transactional, batched `List` is concerns about contention due to transaction length, and not any correctness or deadlock issues. I think if that's a real issue, we can split collections into large (commits, jobs) and small (pipelines, repos), and only allow `List` on small ones. I'd like to have someone else validate that, but given that we use `GetByIndex` successfully right now, it seems relatively low risk.

src/server/auth/server/api_server.go
```diff
@@ -387,6 +392,29 @@ func (a *apiServer) hasClusterRole(ctx context.Context, username string, role st
 
 // Activate implements the protobuf auth.Activate RPC
 func (a *apiServer) Activate(ctx context.Context, req *auth.ActivateRequest) (*auth.ActivateResponse, error) {
+	var resp *auth.ActivateResponse
+	if err := a.env.TxnEnv.WithWriteContext(ctx, func(txCtx *txncontext.TransactionContext) error {
+		var err error
+		resp, err = a.activateInTransaction(ctx, txCtx, req)
+		return err
```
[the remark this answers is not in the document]
PFedak: you only need to wrap it if the linter complains 😜 It's supposed to require all error that come from outside of our codebase to be wrapped (excluding explicit creation of errors that already come with stack traces). The main limitation is that it is conservative around interfaces and assumes that a function satisfying an interface can come from anywhere. The plugin was recently updated with an explicit interface whitelist. I started adding some to #7447

src/server/auth/server/api_server.go
```diff
@@ -134,30 +134,35 @@ func (a *apiServer) EnvBootstrap(ctx context.Context) error {
...
-					if _, err := a.env.GetPpsServer().ActivateAuth(ctx, &pps.ActivateAuthRequest{}); err != nil {
-						return errors.EnsureStack(err)
+					_, err := a.rotateRootTokenInTransaction(txCtx,
+						&auth.RotateRootTokenRequest{
+							RootToken: a.env.Config.AuthRootToken,
+						})
+					return errors.Wrapf(err, "rotate root token")
+				} else {
+					if a.env.Config.PachdSpecificConfiguration != nil {
+						if _, err := a.env.GetPfsServer().ActivateAuthInTransaction(ctx, txCtx, &pfs.ActivateAuthRequest{}); err != nil {
```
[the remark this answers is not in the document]
PFedak: There are some reasons we don't do this which I outlined in my larger comment. I'm not saying we should definitely keep to them, but it's not an accident.

src/server/pps/server/api_server.go
PFedak: I kind of like the idea of a per-transaction log buffer, but just DEBUG logging everything seems clearer. We lost a lot of internal info when we moved over to everything internal being a transaction, it would be nice to recover some of it.

src/server/auth/server/api_server.go
```diff
@@ -387,6 +392,29 @@ func (a *apiServer) hasClusterRole(ctx context.Context, username string, role st
...
+		var err error
+		resp, err = a.activateInTransaction(ctx, txCtx, req)
+		return err
+	}); err != nil {
+		return nil, err
+	}
+	// wait until the clusterRoleBinding watcher has updated the local cache
+	// (changing the activation state), so that Activate() is less likely to
+	// race with subsequent calls that expect auth to be activated.
+	if err := backoff.Retry(func() error {
```
PFedak, line 406: should this be a `RetryUntilCancel`? oh, I see that this is just moved

PFedak (approved):
LGTM, glad we could remove that wrinkle. I'm wondering if pipeline auth token generation used to have some expensive component leading to the parallelization. I can imagine someone needing to stop all of their pipelines before activating auth if there are a lot in a busy cluster, but I think that's worth the benefits we get from transactionality

PR #7823, opened 2022-06-27
[CORE-744] Add a simple log wrapper to Loki stream entries

FahadBSyed:
## Description
This PR fixes a bug that breaks structured JSON logging when using `pachctl logs` and Loki. Some log messages written by components using `Klog` use their own format and are being forwarded as-is by Loki. We address this with a check to see if the entry is a JSON formatted string. If it is not, then we wrap it in a struct and serialize it.

### DIFF
--- a/src/internal/lokiutil/lokiutil.go
+++ b/src/internal/lokiutil/lokiutil.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"sort"
-	"strings"
 	"time"
 
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
@@ -48,8 +47,9 @@ func forEachLine(resp loki.QueryResponse, f func(t time.Time, line string) error
 	return nil
 }
 
-func wrapEntryLog(entry streamEntry) (streamEntry, error) {
-	if strings.HasPrefix(entry.Log, "{") {
+func wrapEntryIfNotJSON(entry streamEntry) (streamEntry, error) {
+	err := json.Unmarshal([]byte(entry.Log), &map[string]interface{}{})
+	if err == nil {
 		return entry, nil
 	}
 	// Entries from Klog may not be in JSON and so should be wrapped to maintain structured logging.
@@ -79,7 +79,7 @@ func QueryRange(ctx context.Context, c *loki.Client, queryStr string, from, thro
 			if err != nil {
 				return errors.EnsureStack(err)
 			}
-			entry, err = wrapEntryLog(entry)
+			entry, err = wrapEntryIfNotJSON(entry)
 			if err != nil {
 				return errors.EnsureStack(err)
 			}
--- a/src/internal/lokiutil/lokiutil.go
+++ b/src/internal/lokiutil/lokiutil.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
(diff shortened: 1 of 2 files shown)


src/internal/lokiutil/lokiutil.go
```diff
@@ -47,6 +48,21 @@ func forEachLine(resp loki.QueryResponse, f func(t time.Time, line string) error
 	return nil
 }
 
+func wrapEntryLog(entry streamEntry) (streamEntry, error) {
+	if strings.HasPrefix(entry.Log, "{") {
```
robert-uhl, line 52: I believe that this should be a full unmarshal, not just a check for starting with a curly bracket.

src/internal/lokiutil/lokiutil.go
```diff
@@ -58,15 +74,15 @@ func QueryRange(ctx context.Context, c *loki.Client, queryStr string, from, thro
...
-				Log    string    `json:"log"`
-				Stream string    `json:"stream"`
-				Time   time.Time `json:"time"`
-			}
+			entry := streamEntry{}
 			err := json.Unmarshal([]byte(line), &entry)
 			if err != nil {
 				return errors.EnsureStack(err)
 			}
+			entry, err = wrapEntryLog(entry)
```
robert-uhl, line 82: Rather than a misleadingly-named function call, i _think_ that this could just be legibly inlined. Something like `if err := json.Unmarshal(…); err != nil { if entry.log, err = json.Marshal(…); err != nil {…}}`?

Or maybe just renamed to something like `wrapIfNotJSON`. That name is kinda terrible though.

src/internal/lokiutil/lokiutil.go
FahadBSyed: Sure, yeah that's more robust

src/internal/lokiutil/lokiutil.go
FahadBSyed: As of now I renamed the function to `wrapLogIfNotJSON`, but like I said, I am happy to inline the function if we think the test is not needed or could be done some other way.

robert-uhl (approved):

jrockway:
I am unconvinced that this is the right fix.  There are two layers of unwrapping that we might be dealing with:

1. @robert-uhl recently added code that is trying to interpret the response that the Loki API returned.  It used to return whatever bytes were in the log file, but now it's wrapped in JSON.
2. It's possible for logs to not be parsed correctly at ingest time; instead of getting the bytes `foo` that were logged, we'll get the log-writing machinery's annotations as well `2020-010-1T12:34:56 stderr F foo`.  Do note that sometimes this is valid JSON; specifically when containers are being run by Docker and not CRI-O. 

When I filed this issue, my understanding was that the code here was trying to handle 1; i.e they changed their HTTP response to return an object as JSON instead of the raw log lines.  If the program logged "foo", then Loki would return `{time:"something", message:"foo"}` 100% of the time.  I haven't been able to reproduce this; my Loki definitely doesn't do that.

Looking at the fix, I worry that the underlying motivation for the original code was to solve an issue of class 1, but was actually an issue of class 2.  Do we have any idea if Loki started returning logs in a different format, even if the underlying logs are correct?  (Test with `logcli` or similar.)

FahadBSyed:
Ahh. I think I misunderstood. I definitely tried to solve it from the perspective of solving class 1. I didn't look at why the response from Loki was coming in with some entries not in JSON format in the first place (which is my understanding of class 2). I can look at the diff in Loki versions between Loki 2.3.0 and 2.5.0 (which is the diff in Loki-stack 2.5.0 and 2.6.4 respectively) to see if any logging changes were made. I do think that if we want to maintain backwards compatibility without breaking structured logging, we probably need to correct for loki's output from pachd's perspective rather than needing to configure loki differently depending on whether its loki version 2.3.0 or 2.5.0. But in the case that the changes made to Loki are gated by a flag, we might be able to leverage that.

PR #7826, opened 2022-06-28
[CORE-786] Disable ServiceAccount Automounting of Token on pods without Pachd

tybritten:


tybritten:
It looks like loki does need the same treatment, but there doesn't appear to be a way to handle it via the values file, which means customizing those two charts too...

jrockway:
That's fine.  Bob is working on upgrading Loki and making it installed by default, so I'll update those tickets to make sure we come back and audit its k8s access.

PR #7827, opened 2022-06-28
Remove hardcoded images and move to values file

tybritten:
Pg bouncer and the test image were hard coded in their templates. This moves those to the values file for customers that need to override or pull locally.


etc/helm/pachyderm/values.yaml
```diff
@@ -713,6 +716,9 @@ oidc:
   mockIDP: true
   #TODO scopes:
 
+testConnection:
+  image: alpine
```
jrockway, line 720: can we make this `image: {repository: ..., tag: ...}` to be consistent with the other image specifications?

tybritten: yup

PR #7828, opened 2022-06-28
[CORE-787] Add missing probes and resources

tybritten:
* Adds missing readiness/liveness probes
* Adds Resource configuration to the test connection pod

### DIFF
--- a/etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
+++ b/etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
@@ -80,13 +80,13 @@ spec:
         imagePullPolicy: IfNotPresent
         name: pg-bouncer
         livenessProbe:
-          failureThreshold: 3
+          failureThreshold: 5
           exec:
             command:
               - "/bin/sh"
               - "-c"
               - psql postgresql://$POSTGRESQL_USERNAME:$POSTGRESQL_PASSWORD@127.0.0.1:$PGBOUNCER_PORT/ --tuples-only --command="SELECT 1;" | grep -q "1"
-          periodSeconds: 10
+          periodSeconds: 30
           successThreshold: 1
           timeoutSeconds: 1
         readinessProbe:
--- a/etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
+++ b/etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
@@ -79,6 +79,23 @@ spec:
         image: bitnami/pgbouncer:1.16.1-debian-10-r82
         imagePullPolicy: IfNotPresent
         name: pg-bouncer
+        livenessProbe:
+          failureThreshold: 3
+          exec:
+            command:
+              - "/bin/sh"
+              - "-c"
+              - psql postgresql://$POSTGRESQL_USERNAME:$POSTGRESQL_PASSWORD@127.0.0.1:$PGBOUNCER_PORT/ --tuples-only --command="SELECT 1;" | grep -q "1"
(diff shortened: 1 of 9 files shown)


etc/helm/pachyderm/templates/tests/test-connection.yaml
```diff
@@ -20,6 +20,13 @@ spec:
       image: "alpine"
       command: ['/bin/sh', '-c', '--']
       args: ['while ! nc -vz pachd-peer 30653 ; do sleep 1 ; done']
+      resources:
+          limits:
+            cpu: 250m
+            memory: 500Mi
```
jrockway, line 26: I bet this could be 50Mi like the request.  Hopefully we never use that much memory.

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
```diff
@@ -79,6 +79,23 @@ spec:
...
+          failureThreshold: 3
+          exec:
+            command:
+              - "/bin/sh"
+              - "-c"
+              - psql postgresql://$POSTGRESQL_USERNAME:$POSTGRESQL_PASSWORD@127.0.0.1:$PGBOUNCER_PORT/ --tuples-only --command="SELECT 1;" | grep -q "1"
+          periodSeconds: 10
+          successThreshold: 1
+          timeoutSeconds: 1
+        readinessProbe:
```
jrockway, line 92: Not sure how applicable to pgbouncer this is, but postgres ships with `pg_isready` to see if postgres is ready: https://www.postgresql.org/docs/current/app-pg-isready.html

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
```diff
@@ -79,6 +79,23 @@ spec:
         image: bitnami/pgbouncer:1.16.1-debian-10-r82
         imagePullPolicy: IfNotPresent
         name: pg-bouncer
+        livenessProbe:
+          failureThreshold: 3
+          exec:
+            command:
+              - "/bin/sh"
+              - "-c"
+              - psql postgresql://$POSTGRESQL_USERNAME:$POSTGRESQL_PASSWORD@127.0.0.1:$PGBOUNCER_PORT/ --tuples-only --command="SELECT 1;" | grep -q "1"
```
jrockway, line 88: I like that we're doing a thorough check.  I guess when this consistently returns "too many open connections to database", we'll restart, and then nothing will work for a while.  But maybe some of those dropped database connections were leaked and that's actually a good thing?

jrockway (approved):
This is really nice.  I like the specialist checks for things like pgbouncer; should surface subtle issues more readily.

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
tybritten: yeah it isn't the the pg_bouncer image

jrockway: fair enough!

etc/helm/pachyderm/templates/tests/test-connection.yaml
tybritten: oops. typo

etc/helm/pachyderm/templates/pgbouncer/deployment.yaml
tybritten: yeah, i can test it on one of @BOsterbuhr or @dgeorg42 load testing setups to see how it behaves first

PR #7830, opened 2022-06-28
Remove pachtf Docker image

albscui:
- package pachtf binary in worker
- remove pachtf image
- update Jsonnet to use the embedded binary instead


PFedak (approved):
LGTM, glad this wasn't so bad once you figured out the cert issue

PR #7831, opened 2022-06-28
worker: driver: improve error handling around /etc/passwd and /etc/group inspection

jrockway:
This code should probably change to use the ID directly without checking anything if the ID is a number.  Docker handles this fine, `docker run -u 1234:4321 ubuntu:20.04 cat /etc/passwd` works even though there is no user 1234 or group 4321.

### DIFF
--- a/src/server/worker/driver/driver.go
+++ b/src/server/worker/driver/driver.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"os"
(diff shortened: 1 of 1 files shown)


tybritten:
This isn't changing the behavior, just the error handling, correct?

jrockway:
No behavior change, just a clearer error message.  I did make invalid /etc/passwd and /etc/groups return an error instead of panicking though.

tybritten (approved):

PR #7834, opened 2022-06-28
[CORE-703] add oidc health check to envoy

armaanv:
This adds the envoy health check definition for the pachd-oidc service. pachd-identity already has a health check configured which checks /dex/.well-known/openid-configuration so no changes required there.


jrockway (approved):
Looks good!

PR #7836, opened 2022-06-29
[Mount Server] Read from config file on startup

smalyala:
Connect to the context specified in the config file if valid on mount server startup. This prevents users from having to input cluster connection info every time they start up notebooks. Tested manually.

Jira: https://pachyderm.atlassian.net/browse/INT-650

### DIFF
--- a/src/server/cmd/mount-server/cmd/cmd.go
+++ b/src/server/cmd/mount-server/cmd/cmd.go
@@ -2,7 +2,9 @@ package cmd
 
 import (
 	"os"
+	"time"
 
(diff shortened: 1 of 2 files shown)


lukemarsden (approved):
Looks good. Is it worth writing to this file as well, or do we already do that?

smalyala:
> Looks good. Please merge.
> 
> Non blocking question: Is it worth writing to this file as well, or do we already do that?

Yep, we write to the file when we switch to a new cluster- https://github.com/pachyderm/pachyderm/blob/master/src/server/pfs/fuse/server.go#L955

PR #7838, opened 2022-06-29
Env Bootstrap before starting internal server [CORE-585]

acohen4:


jrockway (approved):
Looks good.  I feel like at some point individual bootstrappers made RPCs, but I guess that must not be happening?

acohen4:
> Looks good. I feel like at some point individual bootstrappers made RPCs, but I guess that must not be happening?

Any RPCs to other services in pachd should be made via the serviceenv which skips gRPC

jrockway:
That makes sense!

PR #7842, opened 2022-06-30
Distinguish transaction metrics with ErrTxDone.

PFedak:
Our read-only operation model is to run the transaction, then roll it
back. Our implementation happens to also attempt to commit the
transaction after rollback, which always produces ErrTxDone. While the
read-only/dry-run code knows to ignore it, this error is reported as a
normal error to metrics, which is misleading.

This change simply labels these errors differently. The
transaction runner doesn't currently know the context it's being run in
(and in fact, this could lead to swallowing legitimate ErrTxDone in a
read-only operation, if possible), so correctly ignoring these errors
would require a larger change.


jrockway (approved):
Very nice!

PR #7844, opened 2022-06-30
[CORE-744] rollback changes and also remove log parsing in lokiutil.

FahadBSyed:
This PR rolls back changes made that attempt to unmarshal the entries from Loki before passing them down to `parseLokiLine()` in `lokiutil.QueryRange()`'s callback. Log correction should only be done by `parseLokiLine`. This will also prevent logs from being swallowed before we attempt to output them.


jrockway (approved):
Woo!

PR #7845, opened 2022-06-30
Continue deduplicating main.go

acohen4:


src/server/cmd/pachd/main.go
```diff
@@ -574,49 +562,17 @@ func doFullMode(config interface{}) (retErr error) {
...
-	if err := dbutil.WaitUntilReady(ctx, log.StandardLogger(), env.GetDBClient()); err != nil {
-		return err
-	}
-	if err := migrations.ApplyMigrations(ctx, env.GetDBClient(), migrations.MakeEnv(nil, env.GetEtcdClient()), clusterstate.DesiredClusterState); err != nil {
+	env, err := setup(config, "pachyderm-pachd-full")
+	if err != nil {
 		return err
 	}
-	if err := migrations.BlockUntil(ctx, env.GetDBClient(), clusterstate.DesiredClusterState); err != nil {
+	if err = setupDB(context.Background(), env); err != nil {
```
jrockway, line 569: Do we need this `err` later in the function?  If not, I prefer `:=` to `=` here.

src/server/cmd/pachd/main.go
```diff
@@ -184,14 +173,19 @@ func doEnterpriseMode(config interface{}) (retErr error) {
 	} else {
 		log.Printf("no Jaeger collector found (JAEGER_COLLECTOR_SERVICE_HOST not set)")
 	}
-	env := serviceenv.InitWithKube(serviceenv.NewConfiguration(config))
-	profileutil.StartCloudProfiler("pachyderm-pachd-enterprise", env.Config())
+	env = serviceenv.InitWithKube(serviceenv.NewConfiguration(config))
+	profileutil.StartCloudProfiler(service, env.Config())
```
jrockway, line 177: I wonder if we should move this below the `log.SetFormatter` call, since it logs.  (I also have a feeling that `automaxprocs` logs pretty early as well... maybe we need to do the same there.)

jrockway (approved):
Nice!  This is a good chunk of code that people will only have to wade through once.

PR #7846, opened 2022-06-30
Register one service instance with both the Internal + External GRPC Servers [CORE-585]

acohen4:


jrockway (approved):
Looks good.  One thing I'd say about the current design is that we use a lot of lines to basically list which RPC servers each variant of pachd will run.  Would be nice if initialization  and deciding what to run were done in separate places, so we have something like:

```go
func NewFullMode(...){
    servers := [PFS, PPS, License, Enterprise, Transaction, Version, Admin, Proxy, Health]
    if !env.Config().EnterpriseMember {
       servers = append(servers, Identity, Auth)
    }
    ...
}

func NewEnterpriseMode(...) {
    servers :=  [License, Auth, Enterprise, Health, Admin, Version, Identity]
    ....
```

PR #7847, opened 2022-07-01
[CORE-607] Add helm templates and config for kube-event-tail

FahadBSyed:
## Description
This PR adds `kube-event-tail` templates to the set of components we can deploy alongside `pachd`. Deploying `kube-event-tail` is a simple way of collecting `kubernetes events` during `pachctl debug dump` calls.  `Kube-event-tail` watches events and logs them to stdout. 

Configuration of `kube-event-tail` is controlled with a new helm object `kubeEventTail` which has three boolean fields:

- enabled: deploys kube-event-tail if true
- clusterScope: determines whether kube-event-tail logs only events in its namespace or collects and logs all events
- podMonitor: controls whether a prometheus pod monitor is deployed for kube-event-tail

### DIFF
--- a/etc/helm/pachyderm/templates/kube-event-tail/deployment.yaml
+++ b/etc/helm/pachyderm/templates/kube-event-tail/deployment.yaml
@@ -1,0 +1,87 @@
+{{- if .Values.kubeEventTail.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: pachyderm-kube-event-tail
+    suite: pachyderm
+  name: pachyderm-kube-event-tail
+  namespace: {{ .Release.Namespace }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: pachyderm-kube-event-tail
+      suite: pachyderm
+  template:
+    metadata:
+      labels:
+        app: pachyderm-kube-event-tail
+        suite: pachyderm
+    spec:
+    {{- include "pachyderm.imagePullSecrets" . | indent 6 }}
+      containers:
+        - name: kube-event-tail
+          env:
+            - name: DEBUG_ADDRESS
+              value: 0.0.0.0:8081
+            - name: NAMESPACE
+          {{- if eq .Values.kubeEventTail.clusterScope true }}
+              value: ""
+          {{ else }}
+              value: {{ .Release.Namespace }}
+          {{- end }}
+          image: "{{ .Values.kubeEventTail.image.repository }}:{{ .Values.kubeEventTail.image.tag }}"
+          imagePullPolicy: {{ .Values.kubeEventTail.image.pullPolicy }}
+          livenessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /healthz
+              port: debug
+              scheme: HTTP
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
+          ports:
+            - containerPort: 8081
+              name: debug
+              protocol: TCP
+          readinessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /healthz
+              port: debug
+              scheme: HTTP
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
+        {{- if .Values.kubeEventTail.resources }}
+          resources: {{ toYaml .Values.kubeEventTail.resources | nindent 12 }}
+        {{- else }}
+          resources:
+            limits:
+              cpu: 1
+              memory: 45Mi
+            requests:
+              cpu: 100m
+              memory: 45Mi
+        {{- end }}
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: File
+    {{- if .Values.global.securityContexts.enabled }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop:
+                - all
+      securityContext:
+        runAsUser: 1000
+        runAsGroup: 1000
+        fsGroup: 1000
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+    {{- end }}
+      serviceAccountName: pachyderm-kube-event-tail
+{{- end }}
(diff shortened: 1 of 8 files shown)


etc/helm/pachyderm/templates/kube-event-tail/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+          env:
+            - name: DEBUG_ADDRESS
+              value: 0.0.0.0:8081
+            - name: NAMESPACE
+{{- if eq .Values.kubeEventTail.clusterScope true }}
+              value: ""
+{{ else }}
+              value: {{ .Release.Namespace }}
+{{- end }}
+          image: docker.io/jrockway/kube-event-tail:v0.0.6
```
jrockway, line 33: Now that I think about it, we will have to push this to the pachyderm namespace at some point.  (We should open up a ticket to do this work; Jose is working on doing this for etcd and pgbouncer as part of our release, so once he defines a way to do this, then we can do the same thing here.)  

The reason for this is because docker has a pretty low pull rate limit, but we pay them to not rate limit anything in the pachyderm namespace.

jrockway, line 33: Also, this unfortunately has to be configurable through values.  We have a customer that isn't connected to the Internet, so they go manually load a container into their cluster, and refer to it as their-company-name/kube-event-tail or similar.  (We ran into this problem with our connectivity check tool which just uses "alpine".   Here is the fix: https://github.com/pachyderm/pachyderm/pull/7827)

etc/helm/pachyderm/templates/kube-event-tail/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+    metadata:
+      labels:
+        app: kube-event-tail
+        suite: pachyderm
+    spec:
+      containers:
+        - name: kube-event-tail
+          env:
+            - name: DEBUG_ADDRESS
+              value: 0.0.0.0:8081
```
jrockway, line 26: BTW, annoying "fun fact" here... the kubelet is actually doing those http readiness probes, not the pod itself, so debug has to be exposed to the network in order for it to get at healthz.  Very annoying, if you ask me :D

etc/helm/pachyderm/templates/kube-event-tail/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+            httpGet:
+              path: /healthz
+              port: debug
+              scheme: HTTP
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 1
+          resources:
+            limits:
+              memory: 45Mi
```
jrockway, line 59: you can set a cpu limit of 1 here.  (non-integer cpu limits are problematic, random k8s.af article: https://erickhun.com/posts/kubernetes-faster-services-no-cpu-limits/; this is why pachd and opinionated-server include the `maxprocs` library, so we never run into that latency problem they describe)

etc/helm/pachyderm/templates/kube-event-tail/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+            requests:
+              cpu: 100m
+              memory: 45Mi
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: File
+      securityContext:
```
jrockway, line 65: tyler has recently merged new default security contexts which you should include here.  these unfortunately have to be configurable (to be turned off).  some customers have a setup where they won't let the pod be submitted if it doesn't match their specs, and some other customers have a setup where they won't let the pod be submitted if it has any security context (they add their own).  it's annoying.

etc/helm/pachyderm/templates/kube-event-tail/deployment.yaml
```diff
@@ -0,0 +1,69 @@
...
+            - name: DEBUG_ADDRESS
+              value: 0.0.0.0:8081
+            - name: NAMESPACE
+{{- if eq .Values.kubeEventTail.clusterScope true }}
+              value: ""
+{{ else }}
+              value: {{ .Release.Namespace }}
+{{- end }}
+          image: docker.io/jrockway/kube-event-tail:v0.0.6
+          imagePullPolicy: IfNotPresent
```
jrockway, line 34: We also have to let customers pass in an image pull secret.  For reasons.  Check out the other manifests for an example.

etc/helm/pachyderm/templates/kube-event-tail/podmonitor.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if and .Values.kubeEventTail.enabled .Values.kubeEventTail.podMonitor }}
```
jrockway, line 1: I don't think we're typically providing PodMonitors; it's just for one particular distribution of prometheus which I'm not sure is widely used (prometheus-operator).  I think we should just skip this file entirely, and not make it an option.  If someone asks us for a pod monitor for kube-event-tail... easy to provide then ;)

etc/helm/pachyderm/values.yaml
```diff
@@ -554,6 +554,14 @@ pachd:
     # deploy.
     create: true
 
+kubeEventTail:
+  # Deploys a lightweight app that watches kubernetes events and echos them to logs.
+  enabled: false
```
jrockway, line 559: Let's default to true.  That will make it deploy during CI, and will tell you if you added the magic security context settings :)

jrockway (commented):
I'm so excited about this.

Sorry for all the comments.  Our "new application in helm" process is completely undocumented, and there are an infinite number of traps for the unwary.  I am actually sure I'm missing something; going to try to get a "helm reviewer's guide" going next week with some feedback from the customer team!

etc/helm/pachyderm/templates/kube-event-tail/deployment.yaml
FahadBSyed: Done + I moved the resources into the chart in case they need to be configurable

etc/helm/pachyderm/templates/kube-event-tail/deployment.yaml
FahadBSyed: I opened https://pachyderm.atlassian.net/browse/CORE-810 to capture this

PR #7850, opened 2022-07-05
[CORE-400] Log processing of tasks

robert-uhl:
Add logs for when tasks are created, submitted, received, executed and
results processed.


src/internal/task/etcd_service.go
```diff
@@ -149,9 +150,18 @@ func (ed *etcdDoer) Do(ctx context.Context, inputChan chan *types.Any, cb Collec
 						fmt.Printf("errored putting task %v in cache: %v\n", key, err)
 					}
 				}
+				log.WithFields(log.Fields{
+					"task-id": task.ID,
+					"error":   err,
+				}).Print("task response received")
```
jrockway, line 156: What level does "print" end up being?

jrockway (approved):
This should be informative for log readers!

src/internal/task/etcd_service.go
robert-uhl: Info.

PR #7854, opened 2022-07-05
[DOC-420] Configuration of Pachyderm through env variables - Removed all mention of bootstrap config 

nadegepepin:


doc/docs/master/deploy-manage/manage/upgrades.md
```diff
@@ -126,54 +116,62 @@ shows all pods as `READY`:
...
 
-- **C - Neither A nor B**
 
-      You have not created secrets ahead of your installation, nor did you provide values directly in the values.yaml; we created default values for you in the default secret `pachyderm-bootstrap-config` (See column C below for their key). 
-      
+### Pachyderm Internal Secrets
 
-!!! Important
-       It is important to note that if no secret name is provided for the fields mentioned in **A**, Pachyderm will retrieve the dedicated plain-text secret values in the helm values (**B**) and populate a generic, default, auto-generated secret (pachyderm-bootstrap-config) at the time of the installation (see keys in **C**). If no value is found in either one of those two cases, default values are used in `pachyderm-bootstrap-config`. This default secret is reset at each upgrade, and new default values are created causing the upgrade to fail unless they are retrieved and set back into their appropriate fields.
+In both cases (A and B above), Pachyderm reads the values provided, then injects them into a set of internal secrets.
```
acohen4, line 156: This is maybe more of a revision comment - I think there are 3 options now: populate secrets directly, populate secrets via k8s Secrets, or DO NOTHING (sorta). The DO NOTHING option would still require today's behavior around Postgres password, but otherwise would only require the user to provide the ENTERPRISE LICENSE and optionally the IDPs if they so choose. ALL the other values will be autogenerated and upgrades should work fine. The main case we can envision for a customer to populate secrets via k8s Secrets is if they use a tool like ArgoCD for gitops because our helm templates will still include random values each time they get generated.

Does that make sense? 
cc @jrockway

doc/docs/master/deploy-manage/manage/upgrades.md
```diff
@@ -126,54 +116,62 @@ shows all pods as `READY`:
...
+|------------|------------|-----|--------|---------|
+|root-token| Root clusterAdmin| pachd.rootTokenSecretName |pachd.rootToken|(pachyderm-auth, rootToken)|
+|root-token|Users give this token in their values.yaml when they install a cluster that is under an [Enterprise Server's](../../enterprise/auth/enterprise-server/setup) umbrella.|pachd.enterpriseRootTokenSecretName|pachd.enterpriseRootToken|????|
+|postgresql-password|Password to your database|global.postgresql.postgresqlExistingSecretName <br> global.postgresql.postgresqlExistingSecretKey |global.postgresql.postgresqlPassword|(postgres, postgresql-password) , (postgres, postgresql-postgres-password)|
+|OAUTH_CLIENT_SECRET|Oauth client secret for Console <br> Required if you set Console|console.config.oauthClientSecretSecretName |console.config.oauthClientSecret|(pachyderm-console-secret, OAUTH_CLIENT_SECRET)|
```
jrockway, line 147: We should mention that this can be any random value.  I have been using:

```
$ openssl rand -base64 42
tJkHm0+8niOtP1F8lAPryO9dGwMV7SL/u/uCZQi24kFuRj+7VYvtj01q
```
I pick -base64 to get more entropy per byte-of-length than -hex (`openssl rand` doesn't have an alphanumeric option like helm does), but I'm not using base64 as base64; i.e. I wouldn't run "base64 -d" on this and make that the real value (which is what would happen if you were editing a secret and put it in `data` instead of `stringData`).  This point might confuse users that know k8s secrets use base64, that's totally unrelated here, I'm just trying to get the best thing out of openssl I can.  Might be too confusing.

doc/docs/master/deploy-manage/manage/upgrades.md
```diff
@@ -126,54 +116,62 @@ shows all pods as `READY`:
...
-In general, these values can be provided in three different ways:
+In general, these values can be provided to Pachyderm in two different ways:
 
-- **A - (RECOMMENDED in production) Create your secret(s) ahead of your cluster creation** 
+- **A - (RECOMMENDED in production) Create your secret(s) ahead of time** 
       
       **Create your secret(s) ahead of your cluster creation** and provide each secret name in your values.yaml at the time of the first installation. 
 
-      Find the list of Secret Keys for each setting and their corresponding secret name field in your values.yaml (Column A) in the table below.
+      For each set, find the secret name field that references your secret in your values.yaml (Column A) and its corresponding Secret Key (First column) in the table below.
```
jrockway, line 131: I think realistically, we'll have to walk users through the full lifecycle or Pachyderm in each of these sections.  So for this, I think we have to do something like:

```markdown
You can do this step by step, or do it all at once.  Or change it later.  Whatever.

#### Activate Enterprise
Create the following secret in your secret manager:
kind: Secret
metadata:
   name: pachyderm-license
stringData:
    enterprise-license-key: <the text your account rep emailed you, not base64>
Edit values.yaml so that `pachd.license.exsitingSecret` is set to <whatever, not sure>.  Then run "helm upgrade".

#### Enable auth
First, generate a root password with openssl.  Then, add it to this secret.  Then, edit the values.  Next, figure out what IDP you're using.  Then write a dex config for it.  Put the config in this secret.  Edit the values.  Run helm upgrade.


#### Set cluster roles
etc.
```

PR #7857, opened 2022-07-05
Continue to use pachd.oauthClientSecretSecretName

acohen4:


etc/helm/pachyderm/templates/pachd/auth-secret.yaml
```diff
@@ -27,6 +26,9 @@ stringData:
     - profile
     - groups
     - openid
+{{ if not .Values.pachd.oauthClientSecretSecretName }}
+  pachd-oauth-client-secret: {{ default (randAlphaNum 32) .Values.pachd.oauthClientSecret }}
```
jrockway, line 30: In the past I've had bad luck putting newlines in places like this.  One of the helm tests ends up failing; if it passes you're good to go.

src/internal/serviceenv/config.go
```diff
@@ -115,6 +115,7 @@ type EnterpriseSpecificConfiguration struct {
 	ActivateAuth            bool   `env:"ACTIVATE_AUTH,default=false"`
 	AuthRootToken           string `env:"AUTH_ROOT_TOKEN,default="`
 	AuthConfig              string `env:"AUTH_CONFIG,default="`
+	AuthClientSecret        string `env:"AUTH_CLIENT_SECRET="`
```
jrockway, line 118: might need to be `,default=` instead of ending in `=`?

src/server/auth/server/api_server.go
```diff
@@ -168,10 +168,15 @@ func (a *apiServer) EnvBootstrap(ctx context.Context) error {
 		// handle oidc clients & this cluster's auth config
 		if a.env.Config.AuthConfig != "" && a.env.Config.IdentityClients != "" {
 			var config auth.OIDCConfig
+			var pachdClientSecret string
 			var clients []identity.OIDCClient
 			if err := yaml.Unmarshal([]byte(a.env.Config.AuthConfig), &config); err != nil {
 				return errors.Wrapf(err, "unmarshal auth config: %q", a.env.Config.AuthConfig)
 			}
+			if err := yaml.Unmarshal([]byte(a.env.Config.AuthClientSecret), &pachdClientSecret); err != nil {
```
jrockway, line 176: Is this YAML or just a raw value?  If YAML, we might need a test for the case where it's unset (dunno how well the empty string unmarshals as YAML).

jrockway (approved):

src/internal/serviceenv/config.go
acohen4: yikes, thx

PR #7858, opened 2022-07-05
[CORE-748] remove parallel job functionality

armaanv:


RaananHadar:
just a question. why are you removing this functionality? it was a big upgrade to Pachyderm

BOsterbuhr:
> just a question. why are you removing this functionality? it was a big upgrade to Pachyderm

Hey @RaananHadar, we are looking at the possibility of removing the parallel job functionality because it will simplify how provenance works currently and could have performance impacts. As you can see this is a draft PR and a change that if implemented would not happen until roughly 2.4
If you could summarize how you currently use this functionality and how it performs for your use case and send that over in Slack it would help inform us moving forward.

msteffen:

PR #7860, opened 2022-07-05
helm: add a startup probe for pachd

jrockway:
This gives pachd 300 seconds to connect to the database (and apply migrations, etc.) before k8s will consider killing it due to liveness check failure.  We will fail liveness checks until the RPC server starts up, which happens after DB initialization.

One might be aware of a "test connection" helm hook; that just connects to pachd after deployment and is the condition for unblocking "helm install".  We don't do any sequencing around postgres -> pgbouncer -> pachd; they are all applied at once and we hope for the best, which is fine.


acohen4 (approved):
LGTM

PR #7861, opened 2022-07-06
core pach release pipeline refactor

molinamelendezj:


src/version/client.go
```diff
@@ -8,16 +8,12 @@ import (
...
-	MajorVersion = 2
-	// MinorVersion is the current minor version for pachyderm.
-	MinorVersion = 2
-	// MicroVersion is the patch number for pachyderm.
-	MicroVersion = 0
-)
-
 var (
+	// Overwritten at build time by linker
+	AppVersion = "2.2.0"
```
chainlink, line 13: Since this gets overwritten, the default value should probably be something noticeable (ie. `TEMPLATED` or something)

goreleaser/docker.yml
```diff
@@ -66,7 +66,7 @@ builds:
     main: main.go
     binary: mount-server
     ldflags:
-      - "{{ .Env.LD_FLAGS }}"
```
chainlink, line 69: Does LD_FLAGS contain other things besides version? (Legacy support)

.circleci/config.yml
```diff
@@ -123,26 +130,93 @@ jobs:
...
     docker:
       - image: gcr.io/public-builds/chart-releaser:v1.2.1
     steps:
-      - run: wget -q https://get.helm.sh/helm-v3.5.2-linux-amd64.tar.gz -O - | tar -xzO linux-amd64/helm > /usr/local/bin/helm && chmod +x /usr/local/bin/helm
+      - attach_workspace:
+          at: ./
+      - run:
+          name: download and install helm
+          command: |
+             wget -q https://get.helm.sh/helm-v3.5.2-linux-amd64.tar.gz -O - | tar -xzO linux-amd64/helm > /usr/local/bin/helm && chmod +x /usr/local/bin/helm
```
chainlink, line 185: We end up downloading and installing helm alot, we should probably cache it

.circleci/config.yml
```diff
@@ -343,100 +470,270 @@ jobs:
...
+  #             only: /^v\d+\.\d+\.\d+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/
+  test-release:
     jobs:
+      - helm-tests:
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              #only: /^v\d+\.\d+\.\d+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/
+              only: /^\d+\.\d+\.\d+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/
```
chainlink, line 629: We should use this `aliases` config to capture the tag regex https://github.com/pachyderm/haberdashery/blob/master/.circleci/config.yml#L3

.circleci/config.yml
```diff
@@ -343,100 +470,270 @@ jobs:
...
-                - PPS5
-                - PPS6
-                - PPS_AUTH
-                - EXAMPLES
-                - INTERNAL
-  helm:
-    when:
-      and:
-        - not: << pipeline.parameters.RELEASE_PACH_CORE >>
+  # circleci:
```
chainlink, line 534: Note to uncomment needed jobs before merge

.circleci/config.yml
```diff
@@ -113,8 +114,14 @@ jobs:
       - restore_cache:
           keys:
             - go-mod-helm-v4-{{ checksum "go.sum" }}
-      - run: wget -q https://get.helm.sh/helm-v3.5.2-linux-amd64.tar.gz -O - | tar -xzO linux-amd64/helm > /home/circleci/bin/helm && chmod +x /home/circleci/bin/helm
-      - run: wget -q https://github.com/instrumenta/kubeval/releases/latest/download/kubeval-linux-amd64.tar.gz -O - | tar -xzO kubeval > /home/circleci/bin/kubeval && chmod +x /home/circleci/bin/kubeval
+      - run:
```
chainlink, line 117: This could get broken out into a smaller change

.circleci/config.yml
```diff
@@ -123,26 +130,93 @@ jobs:
...
+          command: |
+            wget -q https://get.helm.sh/helm-v3.5.2-linux-amd64.tar.gz -O - | tar -xzO linux-amd64/helm > /usr/local/bin/helm && chmod +x /usr/local/bin/helm
+      - run: 
+          name: make cr release dirs
+          command: |
+            mkdir -p cr-release-packages/official
+            mkdir -p cr-release-packages/preview
+      # - run: helm package -d .cr-release-packages/ --version ${CIRCLE_TAG:1} --app-version ${CIRCLE_TAG:1} etc/helm/pachyderm # cr package did not work with embedded chart
+      # TODO: remove and replace only for testing.
+      - run:
```
chainlink, line 155: This double build is a bit awkward. I think it would be cleaner to push to a different staging helm repo which is not indexed by artifacthub. We'll likely need to use the `yq` code in other places though (ie. when doing a custom version / alpha)

src/version/client.go
molinamelendezj: good point. but won't folks building locally need a value? I will test this doesn't get overwritten when local builds are done. something like `TEMPLATED` will then result in a go int parse error.

.circleci/config.yml
```diff
@@ -343,100 +470,270 @@ jobs:
...
+          name: setup pachydermbuildbot git
+          command: |
+            git config --global user.email "buildbot@pachyderm.io"
+            git config --global user.name "pachydermbuildbot"
+      # TODO make this depend on tag regex
+      - run:
+          name: update homebrew formula
+          command: |
+            ./etc/build/update_homebrew.sh "alpha"
```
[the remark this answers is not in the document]
molinamelendezj: no this needs a bit more thought. should be empty if its a regular patch release so that crazy homebrew script in the homebrew repo can index the correct tap. refactoring how we do homebrew releases will be one the first ones to address. its pretty simple to bend it to our will for the time being.

.circleci/config.yml
molinamelendezj: yea i agree its a bit awkward. but introducing a new repo for the helm chart just comes with a lot more to change atm. the cost of repacking files to archive is less than a minute so thought it could be a temporary evil. ideal id like to replace all this with something like chart museum.

src/version/client.go
chainlink: how about 0.0.0 or something obviously wrong?

.circleci/config.yml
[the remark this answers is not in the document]
chainlink: So, we'll need to remove `alpha` before merging this then?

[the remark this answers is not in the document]
molinamelendezj: so the existing update homebrew is hacky, but I don't want to touch it right now. aslong as this gets a null/blank value for only patch releases it does the expected result. so we'll just use the additional value on the tag which will have a value for every time we do a (custom, alpha, prerelease), and have a blank value on patch releases. and since we only trigger this on the regex tag alais should be protection enough for now.

.circleci/config.yml
```diff
@@ -601,139 +612,62 @@ workflows:
...
-              only: /^\d+\.\d+\.\d+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/
+          filters: *only-release-tags
       - helm-build:
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
-              #only: /^v\d+\.\d+\.\d+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/
-              only: /^\d+\.\d+\.\d+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/
+          filters: *only-release-tags
```
chainlink, line 620: I think we want to build helm everytime as well?

.circleci/config.yml
```diff
@@ -456,42 +591,74 @@ workflows:
                 - LOAD5
                 - LOAD6
                 - LOAD7
-                - LOAD8
```
chainlink, line 459: It looks like the rest of these loads tests were deleted and not re-added?

chainlink (approved):

PR #7870, opened 2022-07-07
bump pach versions

molinamelendezj:
<s>bumping pach versions for first `2.3.0-nightly.1` release, temporarily done manually. </s>

bumping pach versions for first `2.3.0-alpha.1` release, temporarily done manually.

### DIFF
--- a/etc/helm/pachyderm/Chart.yaml
+++ b/etc/helm/pachyderm/Chart.yaml
@@ -21,14 +21,14 @@ type: application
 # each time you make changes to the chart and its templates, including
 # the app version.  Versions are expected to follow Semantic
 # Versioning (https://semver.org/)
-version: 2.3.0-nightly.1
+version: 2.3.0-alpha.1
 
 # This is the version number of the application being deployed. This
 # version number should be incremented each time you make changes to
 # the application. Versions are not expected to follow Semantic
 # Versioning. They should reflect the version the application is
 # using.
-appVersion: 2.3.0-nightly.1
+appVersion: 2.3.0-alpha.1
 
 kubeVersion: ">= 1.16.0-0"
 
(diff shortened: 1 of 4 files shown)


molinamelendezj:
@pleeko should we lock console to `2.3.0-rc.1` for these core nightly instead ?

etc/helm/pachyderm/Chart.yaml
```diff
@@ -21,21 +21,21 @@ type: application
 # each time you make changes to the chart and its templates, including
 # the app version.  Versions are expected to follow Semantic
 # Versioning (https://semver.org/)
-version: 2.1.0
+version: 2.3.0-nightly.1
```
jrockway, line 24: I don't think we want the nightly builds to be sequential integers starting from 1.  This should probably be 20210707 or similar.  (The trap here is that if you're incrementing a number, you need an atomic load-increment-store operation; consider the case where two nightly releases are running concurrently.  They will probably both do the same increment operation and overwrite each others' work.)

Another option may be: `2.3.0-nightly.20210707+(commit sha)` if we ever have two nightly builds on the same day.

See
https://pkg.go.dev/golang.org/x/mod/semver for the format available in the version number.

jrockway (approved):
We should check that goreleaser plumbs the "nightly" tag we decide upon into the AdditionalVersion field correctly.  That should be based on the git tag, though, so probably nothing you need to do in this PR.  It worked for alphas so I don't anticipate any trouble.

etc/helm/pachyderm/Chart.yaml
molinamelendezj: yea agreed, that would be a better scheme for nightly. but apologies, developing story here, looks like we want to use `alpha` instead.

PR #7872, opened 2022-07-07
[CORE-708] Support YAML Connector Configs

FahadBSyed:
## Description
This PR adds a `Config` field to `IDPConnector` objects so that they match the the `dex` documentation. Historically the `JsonConfig` field was use to configure the objects, the new `Config` field will be used going forward. `JsonConfig` is kept for backwards compatibility.

Due to a limitation with the [dexidp/dex library](https://github.com/dexidp/dex/blob/f49e7bc218ebffdc5bfb800c53c06d273958eef2/server/server.go#L562), we need to convert any YAML a user gives us back to JSON. `yaml.YAMLToJSON()` was used for this since it is a no op on existing JSON objects.

Whenever an `IDPConnector` object is created, the overall strategy is to check to see whether `Config` or `JsonConfig` is defined. In the case where both are defined, `Config` is defaulted to. Then the value is converted to JSON.


src/server/identity/cmds/cmds.go
```diff
@@ -6,6 +6,11 @@ import (
 	"io/ioutil"
 	"os"
 
+	"github.com/gogo/protobuf/types"
+
+	"github.com/ghodss/yaml"
+	"github.com/pachyderm/pachyderm/v2/src/server/identity/server"
```
acohen4, line 12: I think it would be better here to have a shared identityutil package that has the PickConfig function, rather than pull the server into the client

src/server/identity/server/dex_api_test.go
```diff
@@ -64,6 +68,45 @@ func TestConnectorCreateListGetDelete(t *testing.T) {
 	require.Equal(t, []*identity.IDPConnector{conn2}, connectors)
 }
 
+// TestCreateConnector tests that connects with valid configs can be created.
+func TestCreateConnector(t *testing.T) {
```
acohen4, line 72: 👍

src/server/identity/server/dex_api_test.go
```diff
@@ -215,7 +291,46 @@ func TestUpdateConnector(t *testing.T) {
...
+				Type:          "mockPassword",
+				Config:        &types.Any{Value: []byte("---\nusername: user\npassword: pass")},
+				ConfigVersion: 4,
+			},
+			expected: &identity.IDPConnector{
+				Id:            "conn",
+				Name:          "newname",
+				Type:          "mockPassword",
+				ConfigVersion: 4,
+				Config:        &types.Any{Value: []byte(`{"password":"pass","username":"user"}`)},
```
acohen4, line 309: it would be nice if the return value also came back in YAML. I don't think that's as critical though, and could be done in a separate PR

src/server/identity/server/dex_api.go
```diff
@@ -156,18 +167,24 @@ func (a *dexAPI) updateConnector(in *identity.UpdateIDPConnectorRequest) error {
...
 
+		// dexidp doesn't support unmarshalling from yaml, so we have to convert to json.
+		// If config is already json, then this is a no-op under the hood.
+		config, err := yaml.YAMLToJSON(PickConfig(in.Connector.Config, in.Connector.JsonConfig))
+		if err != nil {
+			return dex_storage.Connector{}, errors.EnsureStack(err)
+		}
+
+		if string(config) != "" && string(config) != "null" {
+			c.Config = config
```
acohen4, line 182: otherwise does c.Config == nil ?

src/server/identity/server/api_server.go
```diff
@@ -57,6 +58,21 @@ func (a *apiServer) EnvBootstrap(ctx context.Context) error {
 		if err := yaml.Unmarshal([]byte(a.env.Config.IdentityConnectors), &connectors); err != nil {
 			return errors.Wrapf(err, "unmarshal IdentityConnectors: %q", a.env.Config.IdentityConfig)
 		}
+		// IDPConnector.Config can't be unmarshalled to directly, so it has to be manually unmarshalled.
```
acohen4, line 61: " to **JSON** directly,"

src/server/identity/server/api_server.go
```diff
@@ -57,6 +58,21 @@ func (a *apiServer) EnvBootstrap(ctx context.Context) error {
...
+		var tmpConnectors []map[string]interface{}
```
acohen4, line 62: I think it's worth exploring if we can avoid this entirely by unmarshalling directly into map[string]interface{}

acohen4 (changes requested):
Overall this looks good! I made a few smaller comments. One bigger change to consider is whether we should represent config as a map instead of Any so that unmarshalling becomes trivial

etc/helm/pachyderm/values.yaml
```diff
@@ -695,32 +695,30 @@ oidc:
...
-  #           "issuer": "",
-  #           "clientID": "",
-  #           "clientSecret": "",
-  #           "redirectURI": "http://localhost:30658/callback",
-  #           "insecureEnableGroups": true,
-  #           "insecureSkipEmailVerified": true,
-  #           "insecureSkipIssuerCallbackDomainCheck": true,
-  #           "forwardedLoginParams": ["login_hint"]
-  #       }
+  #     Config:
```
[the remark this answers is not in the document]
FahadBSyed: Do you happen to know of any protobuf objects that will generate code into map[string]interface{}? I tried playing around with a few different objects but they either made the tests kind of clunky without solving the deserialization problem or they didn't generate valid code in the .pb.go file.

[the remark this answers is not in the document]
acohen4: yeah I see how that also may not be a clean solution. Maybe the simplest thing would be to just have it be of type `[]byte`? I think you suggested that at some point. Sorry for flip-flopping here - I feel like there should be a cleaner solution than the way `Any` deserializes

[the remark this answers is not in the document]
FahadBSyed: No worries, I can try that and see how it goes!!

src/server/identity/server/dex_api.go
FahadBSyed: Yes, I believe so

acohen4: I think it make sense to throw an error in the case that the client passes us an empty config

src/server/identityutil/identityutil.go
```diff
@@ -0,0 +1,37 @@
...
+		buf := bytes.Buffer{}
+		err := marshaler.Marshal(&buf, config)
+		if err != nil {
+			return nil, err
+		}
+		return buf.Bytes(), nil
+	case jsonConfig != "" && jsonConfig != "null":
+		return []byte(jsonConfig), nil
+	default:
+		return []byte(""), nil
```
acohen4, line 29: I think we should throw an error in this case

FahadBSyed: Ahh you mentioned this before, but I missed, that you're right!

src/server/identity/server/dex_api.go
FahadBSyed: Some of our tests omit config on updates (so they don't need to be aware of the current config). It seems like the intended behavior for those tests is that omitting the config should work on an update, but not on a create of a new connector.

src/server/identity/cmds/cmds.go
```diff
@@ -1,11 +1,18 @@
...
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 
+	"github.com/gogo/protobuf/jsonpb"
+
+	"github.com/pachyderm/pachyderm/v2/src/server/identityutil"
```
acohen4, line 12: super nit, but I think your editor is reformatting the imports

FahadBSyed: yeah that's a little weird, I can fix that though

src/server/identity/cmds/cmds.go
```diff
@@ -25,21 +29,32 @@ type connectorConfig struct {
...
-	if err := json.Unmarshal([]byte(conn.JsonConfig), &config.Config); err != nil {
-		return nil, errors.EnsureStack(err)
-	}
-	return &config, nil
+	return &connConfig, nil
 }
 
 func (c connectorConfig) toIDPConnector() (*identity.IDPConnector, error) {
-	jsonConfig, err := json.Marshal(c.Config)
+	configBytes, err := json.Marshal(c.Config)
```
acohen4, line 52: could you add a comment explaining the JSON/YAML shenanigans?

src/server/identity/server/api_server.go
```diff
@@ -53,8 +57,13 @@ func (a *apiServer) EnvBootstrap(ctx context.Context) error {
 	}
 	if a.env.Config.IdentityConnectors != "" {
 		a.env.Logger.Info("Started to configure identity connectors via environment")
-		var connectors []identity.IDPConnector
-		if err := yaml.Unmarshal([]byte(a.env.Config.IdentityConnectors), &connectors); err != nil {
+		var connectors identityutil.IDPConnectors
+		jsonBytes, err := yaml.YAMLToJSON([]byte(a.env.Config.IdentityConnectors))
```
acohen4, line 61: this still works if the src is JSON?

acohen4 (approved):
LGTM, left just a couple small comments

src/server/identity/cmds/cmds.go
FahadBSyed: Sure, can do.

src/server/identity/server/api_server.go
FahadBSyed: Yup, its a no-op if the source is JSON. https://github.com/ghodss/yaml/blob/d8423dcdf3440d0a5baffc6f90a11e4128545620/yaml.go#L105

PR #7879, opened 2022-07-07
Bootstrap Pachyderm before serving external traffic [CORE-585]

acohen4:


jrockway (approved):
Very nice.

Just for others looking at this PR... offline we had a long discussion about the enterprise server connecting back to the pachd that is registering.  We decided it's hard to guarantee that the customer's network topology would allow that, and starting a server early to respond to that request would be in the critical path of server startup for users that didn't want enterprise registration (potentially introducing a bug / flakiness to support a feature they don't care about).  For example, pachd would tell k8s and the load balancer that it was "ready" before it could serve PFS requests.  Seemed risky, so not doing this validation cleans up those two scenarios (at the cost of mistaken identities when registering with the enterprise server).

PR #7882, opened 2022-07-08
[CORE-811] Keep only Pachyderm logs

robert-uhl:
Drop logs unless the label `suite` is “pachyderm.”


jrockway (approved):
This seems like the right approach to me.  I'll test it out, but it looks good on paper.

PR #7884, opened 2022-07-08
alpha.0 fixes

tybritten:
- Fixes existing secret support for console oauth on pachd
- fix console readonlyrootfilesystem (add emptydir for `/home`)
- Fix auth URI settings (using `ingress.host` and proxy configures all the URIs properly while setting issuer to `http://pachd:30658/dex`

### DIFF
--- a/etc/helm/pachyderm/templates/console/deployment.yaml
+++ b/etc/helm/pachyderm/templates/console/deployment.yaml
@@ -110,6 +110,8 @@ spec:
         resources: {{ toYaml .Values.console.resources | nindent 10 }}
         {{- end }}
         volumeMounts:
+        - mountPath: /home/
+          name: tmp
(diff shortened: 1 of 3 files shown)


etc/helm/pachyderm/templates/console/deployment.yaml
```diff
@@ -110,6 +110,8 @@ spec:
         resources: {{ toYaml .Values.console.resources | nindent 10 }}
         {{- end }}
         volumeMounts:
+        - mountPath: /home/
```
acohen4, line 113: why do we need this as an emptyDir?

acohen4 (approved):
LGTM

etc/helm/pachyderm/templates/console/deployment.yaml
tybritten: npm writes a cache there

PR #7885, opened 2022-07-08
Give the internal auth user cluster admin perms [CORE-827]

acohen4:


jrockway (approved):
Looks good.

Do we need to do something like this for the metrics user as well?  I recently turned them on locally, but get `error:robot:metrics is not authorized to perform this operation - needs permissions [PIPELINE_LIST_JOB] on REPO edges`

PR #7886, opened 2022-07-08
Reduce the default cluster pool to 3

acohen4:


jrockway (approved):
Will be a good experiment.  Let's keep an eye on end-to-end time and flakiness.

PR #7888, opened 2022-07-11
[CORE-771] Ensure that ReplicationController name is ≤ 63 characters

robert-uhl:


src/internal/ppsutil/util.go
```diff
@@ -42,18 +42,32 @@ import (
...
+// manages a pipeline's workers.  Names are limited to 63 characters, so the
+// pipeline name is truncated if necessary in order to stay under the limit.  Of
+// note, collisions are _not_ currently handled.
 func PipelineRcName(name string, version uint64) string {
 	// k8s won't allow RC names that contain upper-case letters
 	// or underscores
 	// TODO: deal with name collision
 	name = strings.ReplaceAll(name, "_", "-")
+	var max = 63 - (len("pipeline") + 1) // pipeline-
+	max -= len(fmt.Sprint(version)) + 2  // -v$VERSION
```
PFedak, line 67: I feel like you might as well just use `len("pipeline-")` and `fmt.Sprint("-v%d",version)`, or maybe even a single line, rather than adding the fixed characters as numbers separately, makes it easier to see that things match up

src/internal/ppsutil/util.go
```diff
@@ -42,18 +42,32 @@ import (
...
+	if len(name) > max {
+		name = name[0:max]
```
PFedak, line 69: you can omit the zero

src/internal/ppsutil/util_test.go
```diff
@@ -0,0 +1,25 @@
...
+		result  string
+	}{
+		{"foo", 1, "pipeline-foo-v1"},
+		{"012345678901234567890123456789012345678901234567890123456789", 1, "pipeline-012345678901234567890123456789012345678901234567890-v1"},
+		{"012345678901234567890123456789012345678901234567890123456789", 1024, "pipeline-012345678901234567890123456789012345678901234567-v1024"},
+	} {
+		if result := ppsutil.PipelineRcName(test.name, test.version); result != test.result {
+			t.Errorf("PipelineRcName(%q, %d) = %q, not %q", test.name, test.version, result, test.result)
+		} else if len(result) > 63 {
+			t.Errorf("len(PipelineRcName(%q, %d)) = %d", test.name, test.version, len(result))
```
PFedak, line 22: maybe include "too long" or ">63" in here

PFedak (approved):
There's one potential issue here, and I can't decide how important it is. Potentially a user could have created a pipeline name longer than this limit which is working fine, but would theoretically stop working if the version was increased enough. Most of the system would keep working, but they would be unable to update the pipeline. It seems like the strictly correct option would be to permit too-long names in update requests.

I'll leave this up to you - I'm not sure we have a good way of assessing if any customer is in this state.

robert-uhl:
> Potentially a user could have created a pipeline name longer than this limit which is working fine, but would theoretically stop working if the version was increased enough.

I believe that this is incorrect: the name will be truncated by enough characters to fit the longer version number.

There _is_ an issue of collisions between different pipelines with common-prefix names at the same version number. This exacerbates the pre-existing issue with hyphens and underscores. I’ve opened CORE-829 to address collisions generally.

robert-uhl:
You’re right — and worse, that upgrade issue also applies to an existing pipeline which is upgraded past version 9,999. I need to close this PR and take a different tack entirely.

Thanks for the insightful review!

PR #7890, opened 2022-07-11
Fix copy url params

albscui:
Currently, every url param would get the last param's value due to the "Go for-loop pointer bug".


PFedak (approved):
LGTM, glad we caught this :)

PR #7893, opened 2022-07-12
Add runAsUser/Group to all containers

tybritten:
If OPA policies are applied up the stack (deployment, replication controller, etc) they will fail if they're only applied to the pod. This adds the runAsUser/Group to the containers explicitly as well.


jrockway (approved):
Very nice!

PR #7894, opened 2022-07-12
Remove liveness probes for pachd

tybritten:


etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -320,11 +320,6 @@ spec:
             command:
             - /pachd
             - --readiness
```
jrockway, line 322: I think we can keep the pachd liveness probe.

etc/helm/pachyderm/values.schema.json
```diff
@@ -439,6 +439,9 @@
                         "limits": {
                             "type": "object",
                             "properties": {
+                                "cpu": {
+                                    "type": "integer"
```
jrockway, line 443: This should be a string.  "100m" for example.

jrockway (changes requested):

etc/helm/pachyderm/values.schema.json
tybritten: yeah weird, that must have been from upstream. fixed

etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -325,13 +325,13 @@ spec:
...
+          failureThreshold: 10
+          timeoutSeconds: 30
         startupProbe:
           exec:
             command:
             - /pachd
             - --readiness
-          failureThreshold: 10
```
jrockway, line 333: We still need these.  Basically gives pachd 300 seconds to wait for postgres, apply migrations, and setup the config from the environment.

jrockway (approved):
1 nit ;)

PR #7897, opened 2022-07-12
[CORE-771] Support 63-character pipeline names

robert-uhl:
Pipeline pods are now identified by `pipelineName` and `pipelineVersion` instead of `app`.  There is a fallback check for the older `app`-style pods.


robert-uhl:
Of note: this PR changes all pipeline pods and replicationcontrollers to have `app=pipeline` and adds `pipelineName` and `pipelineVersion` labels. It also changes S3 gateway services to have `app=s3gateway` and adds `pipelineName` and `pipelineVersion` labels. Finally, it changes spout secrets to have `app=spout` and adds a `pipelineName` label.

robert-uhl:
While I was updating function signatures I unexported some functions not used outside their packages.

src/server/pps/server/api_server.go
```diff
@@ -3089,7 +3090,46 @@ func RepoNameToEnvString(repoName string) string {
...
+		},
+		LabelSelector: metav1.FormatLabelSelector(metav1.SetAsLabelSelector(map[string]string{"app": "pachd"})),
+	})
+	if err != nil {
+		return nil, errors.EnsureStack(err)
+	}
+	return podList.Items, nil
+}
+
+func (a *apiServer) rcPods(ctx context.Context, pipelineName string, pipelineVersion uint64) ([]v1.Pod, error) {
```
PFedak, line 3107: seems like this should be calling a helper that takes a kubeClient and a namespace, rather than duplicating it entirely, given the more extensive fallback logic

src/server/pps/server/api_server.go
```diff
@@ -3089,7 +3090,46 @@ func RepoNameToEnvString(repoName string) string {
...
+			APIVersion: "v1",
+		},
+		LabelSelector: metav1.FormatLabelSelector(metav1.SetAsLabelSelector(map[string]string{
+			"app":             "pipeline",
+			"pipelineName":    pipelineName,
+			"pipelineVersion": fmt.Sprint(pipelineVersion),
+		})),
+	})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
```
PFedak, line 3120: hmm, have you seen this? I'm confused when List would give not found

src/server/pps/server/infra_driver.go
```diff
@@ -129,7 +130,9 @@ func (d *mockInfraDriver) makeRC(pi *pps.PipelineInfo) *v1.ReplicationController
 				pachVersionAnnotation:        version.PrettyVersion(),
 			},
 			Labels: map[string]string{
-				pipelineNameLabel: pi.Pipeline.Name,
+				"app":                "pipeline",
+				pipelineNameLabel:    pi.Pipeline.Name,
+				pipelineVersionLabel: fmt.Sprint(pi.Version),
```
PFedak, line 135: maybe worth making this defined alongside `pipelineLabels` and `spoutLabels`

src/server/pps/server/s3g_sidecar.go
```diff
@@ -211,9 +211,12 @@ func (s *k8sServiceCreatingJobHandler) OnCreate(ctx context.Context, jobInfo *pp
 		return nm
 	}
 	selectorlabels := map[string]string{
-		"app":       ppsutil.PipelineRcName(jobInfo.Job.Pipeline.Name, jobInfo.PipelineVersion),
-		"suite":     "pachyderm",
-		"component": "worker",
+		// NOTE: this used to be ppsutil.PipelineRcName(jobInfo.Job.Pipeline.Name, jobInfo.PipelineVersion)
```
PFedak, line 214: Hmm, doesn't the selector need to match the worker pod labels? It seems like `app` would have to be `pipeline` here for that to work

src/server/pps/server/api_server.go
```diff
@@ -3089,7 +3090,46 @@ func RepoNameToEnvString(repoName string) string {
...
+		}
+		return nil, errors.EnsureStack(err)
+	}
+	if len(podList.Items) == 0 {
+		return a.legacyRCPods(ctx, pipelineName, pipelineVersion)
+	}
+	return podList.Items, nil
+}
+
+func (a *apiServer) legacyRCPods(ctx context.Context, pipelineName string, pipelineVersion uint64) ([]v1.Pod, error) {
```
PFedak, line 3131: I think this is worth a comment explaining in what sense it's legacy. I'm also not convinced we actually need this - this would be used when a new cluster is trying to iterate over pods from an out of date RC, but upgrading will trigger all of those to be deleted, which happens through the normal k8s propagation rather than directly via selector. Granted, this happens asynchronously, so it is possible to be in this state, so probably better safe than sorry.

PFedak (approved):
I'm a little concerned about the s3 gateway service selector, as it seems like it shouldn't be finding the right pods, but maybe I don't understand how that routing works.

Other than that, looks good, I'd just recommend a couple of comments about the legacy function and maybe consolidating it so there's a single source of truth for the right set of selector labels

PFedak (changes requested):
Given that the S3 tests aren't working, we probably do need to fix that the issue I pointed out. I'm not sure what labels we need on the s3 gateway service itself vs the selector, exactly

src/server/pps/server/api_server.go
robert-uhl: I refactored it into a `listPods` method.

src/server/pps/server/api_server.go
robert-uhl: It was a belt-and-suspenders approach, but no doubt you’re correct in thinking it unneeded. Removed.

PFedak (approved):
LGTM, glad we caught the test issue

src/server/pps/server/api_server.go
robert-uhl: I added a comment in the equivalent location.

PR #7901, opened 2022-07-13
pfs_file: return errors from close along with errors from callbacks

jrockway:
Fixes CORE-774.

This fixes an issue where pipelines that write filenames not representable in PFS (like `foo?bar`) fail with the error `EOF`.  

`EOF` comes from the logic in ModifyFileClient.  Files are uploaded by calling PutFileTAR on a ModifyFileClient.  ModifyFileClient implement PutFileTAR by writing `SendMsg(AddFile{...})` to the client streaming RPC call.  `SendMsg` can't return an error from the server at the time that you send an invalid record; it's a unidirectional stream, not a bidirectional stream.  What happens when you send an invalid record is that the server just closes the stream without telling anyone.  Subsequent calls to `SendMsg` succeed until a client-side buffer is full, at which point GRPC realizes the stream is closed, and returns `io.EOF`. 

Additionally, each write to the stream (PutFile/DeleteFile/CopyFile/etc.) is wrapped with a `maybeError` function that only calls the requested method if previous methods didn't return an error.  (When one of those methods returns an error, it retains that and returns it each time you call one of those methods.).  This is not a problem in and of itself, except that `Close` is also wrapped with the same logic (return the previous  error instead of calling the Close function), so that means once we hit an error we can no longer close the stream. 

Finally, `WithModifyFileClient`/`WithCreateFileSetClient` only call `CloseAndRecv` if the callback passed to those functions doesn't error.  Since, when this bug is in effect, we return EOF from that callback, we never call Close and get the chance to receive the real error message from the server.

Sometimes you do get the error message from the server despite all of this.  That's because `SendMsg` returns no error even if the filename is invalid; it's `Close` that returns the error from the server.  This case happens when the faulty file is uploaded before the stream buffer fills up (so, "zzz?bar" is likely to return the error, "000?bar" is likely not to because a subsequent file upload is likely to fill up the buffer).  If you have a pipeline that produces a single file and it's invalid, you'll get a good error message.  If you have a pipeline which produces hundreds of files, then you might not get the real error.   Thus, the bug was quite difficult to track down ;)

Anyway, the fix for this is twofold.  

First, we no longer gate closing the stream on the WithModifyFileClient (etc.) callback not returning an error.  We close no matter what.  The Close functions were gated by `maybeError`, which means even if we didn't have the don't-call logic in the With* functions, we'd still not call it.  I've rewritten the Close functions to Close no matter what, but also return the error that was retained inside the ModifyFileClient (to prevent any changes to the outward API; that error would be returned INSTEAD of closing before this change).  I used `multierr.Combine` to return both errors.  If Close succeeds but the stream broke earlier, you'll just get the raw retained error.  Similarly, if `Close` fails but there was no prior error, you'll only get the error from close.  So all code that consumes this API should receive similar behavior (err != nil never changes), but we always close the stream and always return the error from the server now.

The side effect of this is that most failures of the class described in CORE-744 end up with three error messages, two EOFs and one "hey that filename was invalid".  This is ugly, but it's the only way to not suppress errors to Close and still maintain the contract of our Close wrappers in the client.  (I could be persuaded to break this in 2.3, of course.)  One EOF error is the error returned from user code, the other EOF error is the error retained by ModifyFileClient, and the last error is the message from the server that we actually care about.

Regarding multierr... wrapping io.EOF is always scary because the standard library uses the raw non-wrapped value as a sentinel, and so does gRPC.  We never use ModifyFileClient as an io.Reader or io.Writer, and we do our wrapping after gRPC has inspected the errors, so this shouldn't be a problem.  But we should be careful about `multierr.Combine` in general (as with errors.Wrap and friends, of course; only EnsureStack maintains the io.EOF sentinel as a special case), because wrapping io.EOF can be dangerous in general.

### DIFF
--- a/src/client/pfs_file.go
+++ b/src/client/pfs_file.go
@@ -15,6 +15,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/grpcutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/renew"
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
+	"go.uber.org/multierr"
 )
 
 // PutFile puts a file into PFS from a reader.
@@ -83,9 +84,8 @@ func (c APIClient) WithModifyFileClient(commit *pfs.Commit, cb func(ModifyFile) 
 		return err
 	}
 	defer func() {
-		if retErr == nil {
-			retErr = mfc.Close()
-		}
+		err := mfc.Close()
+		retErr = multierr.Combine(retErr, errors.Wrap(err, "close"))
 	}()
 	return cb(mfc)
 }
@@ -129,7 +129,7 @@ func (mfc *modifyFileCore) PutFile(path string, r io.Reader, opts ...PutFileOpti
 	for _, opt := range opts {
 		opt(config)
 	}
-	return mfc.maybeError(func() error {
+	return mfc.maybeError("PutFile", func() error {
 		if !config.append {
 			if err := mfc.sendDeleteFile(&pfs.DeleteFile{
 				Path:  path,
@@ -161,14 +161,14 @@ func (mfc *modifyFileCore) PutFile(path string, r io.Reader, opts ...PutFileOpti
 	})
 }
 
-func (mfc *modifyFileCore) maybeError(f func() error) (retErr error) {
+func (mfc *modifyFileCore) maybeError(name string, f func() error) (retErr error) {
 	if mfc.err != nil {
 		return mfc.err
 	}
 	defer func() {
 		retErr = grpcutil.ScrubGRPC(retErr)
 		if retErr != nil {
-			mfc.err = retErr
+			mfc.err = errors.Wrap(retErr, name)
 		}
 	}()
 	return f()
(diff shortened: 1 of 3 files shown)


src/client/pfs_file.go
```diff
@@ -338,9 +336,11 @@ func (c APIClient) WithCreateFileSetClient(cb func(ModifyFile) error) (resp *pfs
 		return nil, err
 	}
 	defer func() {
-		if retErr == nil {
-			resp, retErr = ctfsc.Close()
+		r, err := ctfsc.Close()
+		if retErr == nil && err == nil {
+			resp = r
 		}
+		retErr = multierr.Combine(retErr, errors.Wrap(err, "close"))
```
molinamelendezj, line 343: Great fix! good read for sure. so here this can wrap `io.EOF `error when it occurs? Will the caller ever need to check for this error differently. You're right the standard lib says this should never be wrapped lol but since it's already inspected the glue should hold. The do have another error for the case [`ErrUnexpectedEOF` ](https://github.com/golang/go/blob/bf2ef26be3593d24487311576d85ec601185fbf4/src/io/io.go#L48)If the EOF occurs unexpectedly in a structured data stream. 

https://github.com/golang/go/blob/bf2ef26be3593d24487311576d85ec601185fbf4/src/io/io.go#L38

PR #7903, opened 2022-07-13
Rewrite Gatk make target to more closely match the Gatk README.md

FahadBSyed:
## Description
This PR rewrites the `gatk-data` make target so it more closely follows the instructions we provide the user in the readme. Its still not completely 1:1 but this should help for debuggability and dogfooding our process.


armaanv (approved):
looks good to me

PR #7910, opened 2022-07-14
re-structured coding conventions doc

lbliii:
Restructured the coding conventions contributor's guide; need to gather input on if any other changes should be made.


jrockway (approved):
I made some comments on the 1.13 version since it showed up first, but meant to do it in the "master" version.  Otherwise, looks good.

We do need to flesh this out significantly.  We don't really have any rules that we use internally, though.

nadegepepin (approved):
LGTM if ok with @jrockway

jrockway (approved):
Yup, this is totally fine with me.  Very nice work.

PR #7915, opened 2022-07-14
Add shallow checkout command

chainlink:


.circleci/main.yml
```diff
@@ -25,6 +25,13 @@ parameters:
     type: string
     default: "0"
 
+commands:
+  shallow-checkout:
+    steps:
+      - run:
+          name: Shallow clone repo
+          command: git clone --depth 1 "$CIRCLE_REPOSITORY_URL" --branch "$CIRCLE_BRANCH" .
```
molinamelendezj, line 33: @chainlink will this break if job is not off a branch and off a tag instead?

molinamelendezj (approved):
one comment otherwise great idea!

.circleci/main.yml
chainlink: Good catch! I'll revert it for the tag based ones, we can ship it incrementally

PR #7916, opened 2022-07-14
remove/archive gcloud cluster setup

lbliii:
- deleted/archived gcp cluster setup


nadegepepin (changes requested):
2 super quick feedbacks:
- We do not "archive" pages individually, Github takes care of keeping track of this. We archive a full version of the docs at a time when EOLife-ing a given version of Pachyderm.
- However, when a page is removed, just like when it is moved, adding a redirect might be a safe way to avoid 404s.

nadegepepin (approved):
LGTM

PR #7917, opened 2022-07-15
Fix swallowed error in PFS DeleteFile 

albscui:
This bug was discovered serendipitously while testing CronInput repo behaviour.

Currently, if you have a CronInput with `overwrite set to true`, and the existing cron timestamp falls behind PPS master's clock, such that the cron tick has to catch up quickly, then we see an accumulation of cron tick files in the repo. This is because while the parent commit is finishing (doing compaction), the child commit tries to delete the previous cron tick files, whereby the error gets swallowed, and commits the new file, outcompeting compaction. Although in the cron case, the system eventually goes back to a good state once compaction catches up, it's still an important bug to fix since other areas of the system also depends on this file deletion logic.

Thanks to @PFedak for pairing with me on this one.

### DIFF
--- a/src/client/pps.go
+++ b/src/client/pps.go
@@ -176,13 +176,14 @@ func NewCronInput(name string, spec string) *pps.Input {
 // It uses cron syntax to specify the schedule. The input will be exposed to
 // jobs as `/pfs/<name>/<timestamp>`. The timestamp uses the RFC 3339 format,
 // e.g. `2006-01-02T15:04:05Z07:00`. It includes all the options.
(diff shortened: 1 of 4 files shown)


brycemcanally (approved):
lgtm, good catch. It looks like the failing PFS test (GlobFile) is failing because it attempts a glob pattern delete which isn't supported in 2.x. Not returning the error and the followup root directory deletion is probably why the test was passing up to this point.

PFedak (approved):
seems bad that test with deleting glob patterns wasn't even checking that things got deleted

PR #7918, opened 2022-07-15
docs: document kind

jrockway:
`minikube tunnel` doesn't do the right thing on Linux.  So, suggest and document kind.

CORE-835


nadegepepin (approved):
LGTM

PR #7925, opened 2022-07-18
[CORE-883] Console should restart on helm upgrades

acohen4:


jrockway (approved):
Aha!  This makes total sense.

PR #7928, opened 2022-07-18
Fix lint and upgrade to go1.18.4

albscui:


.circleci/main.yml
```diff
@@ -138,7 +138,7 @@ jobs:
           path: /tmp/test-results
   helm-tests:
     docker:
-      - image: cimg/go:1.18.3
+      - image: cimg/go:1.18.4
```
jrockway, line 141: I noticed some things in this file are still go 1.17, and I'm not sure if that's for a good reason.  Would be good to try upgrading those at some point, maybe in a follow-up PR.

go.sum
```diff
@@ -371,7 +371,6 @@ github.com/coreos/go-semver v0.3.0/go.mod h1:nnelYz7RCh+5ahJtPPxZlU+153eP4D4r3Ee
 github.com/coreos/go-systemd v0.0.0-20161114122254-48702e0da86b/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
 github.com/coreos/go-systemd v0.0.0-20180511133405-39ca1b05acc7/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
 github.com/coreos/go-systemd v0.0.0-20190321100706-95778dfbb74e/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
-github.com/coreos/go-systemd v0.0.0-20190719114852-fd7a80b32e1f h1:JOrtw2xFKzlg+cbHpyrpLDmnN1HqhBfnX7WDiW7eG2c=
```
albscui, line 374: ran `go mod tidy` just for cleanup

jrockway (approved):
Looks good.  I would maybe SSH into the first build and just double check that "go" in the path is the right version, just so we don't make the same mistake twice.

PR #7931, opened 2022-07-19
Upgrade circle machine image

chainlink:


molinamelendezj (approved):
awesome, shipit! will keep and eye on the occasional TCP connection issue with this new version. have an open ticket for this on current image with circleci. hadn't bumped it until circle could get a reproduce setup.

PR #7934, opened 2022-07-19
Kill fallback cache for install.sh

chainlink:


jrockway (approved):
What's the background?  Albert was having some issues around this yesterday, I think.

chainlink:
Having a layered cache makes sense for package managers (ie npm) where grabbing older deps is fine because at least you get a subset instead of starting from scratch. It doesn't work in this case, because
1) It's never updated anywhere
2) Used in conjunction with out `install.sh` shell script, if you try to update a version of something in that script, it's already grabbed the old version in the cache fallback, and thus skips over installing the new one.

jrockway:
@chainlink Yup, the change totally makes sense to me.  Was just wondering what problem you ran into and if it's the same one as Albert (he's trying to make it use Go 1.18.  I think he eventually succeeded.)

chainlink:
cc @albscui for visibility

PR #7937, opened 2022-07-19
add helm and redhat jobs to release workflow

molinamelendezj:
since `master`'s circle config and its deps has diverged a bit. this PR is only for `2.2.x`. It will allows for helm and redhat release to run on the release workflow which has been refactored for `2.3.x+ `releases. only two more patch release are expected before `2.3.0` is cut so this will be a short lived patch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,13 @@ parameters:
   RELEASE_TYPE:
     type: string
     default: ""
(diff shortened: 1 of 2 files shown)


chainlink (approved):

molinamelendezj:
turns out this is not necessary. closing.

PR #7941, opened 2022-07-20
add buildinfo to version, and add parameters to circleci

seslattery:
This also removes some unused functions in server/version/client.go.  I chose not to explicitly enable a new flag to enable this behavior, as --raw is sufficient to see the new buildinfo if they are available.  

@molinamelendezj I did have a question - I think it makes sense to pass in the build_date as time of build from CI perspective with an Ldflag, not Git's status.  Do you have a preference?


molinamelendezj:
> 

id say let it do it and not pass it in, as that seems to be the[ whole point of this,](https://shibumi.dev/posts/go-18-feature/) not to have to pass things in

src/version/client.go
```diff
@@ -21,18 +21,57 @@ var (
...
 	Version = &pb.Version{
-		Major:      uint32(MajorVersion),
-		Minor:      uint32(MinorVersion),
-		Micro:      uint32(MicroVersion),
-		Additional: AdditionalVersion,
+		Major:           uint32(MajorVersion),
+		Minor:           uint32(MinorVersion),
+		Micro:           uint32(MicroVersion),
+		Additional:      AdditionalVersion,
+		GitCommit:       B.gitCommit,
```
molinamelendezj, line 32: if `debug.ReadBuildInfo()` returns error and b` `is empty won't this lead to `panic`? i guess `debug.ReadBuildInfo() `would be unlikely to have an issue?

molinamelendezj (approved):
one slight comment but over all pretty cool. jfyi, pachd actually runs this code at startup. and ive seen pachd crash if `panic`

seslattery:
> > 
> 
> id say let it do it and not pass it in, as that seems to be the[ whole point of this,](https://shibumi.dev/posts/go-18-feature/) not to have to pass things in

The reason I find passing in the timestamp compelling, is it allows us to look back to circleci during that time to help us narrow down what might have done wrong with a build.  Using the last modified timestamp isn't particuarily interesting as once I have the commit sha, it's trivial to see what went wrong.  The commit sha alone also isn't sufficient to trace a build back to circle ci, as we might rerun a build with an identical sha for a variety of reasons.  Passing in the timestamp gives us an additional bit of information.  An alternative might be to pass a build ID etc, but timestamp also gives us an idea of how old a release is which is also useful.  


> one slight comment but over all pretty cool. jfyi, pachd actually runs this code at startup. and ive seen pachd crash if `panic`

I think https://github.com/pachyderm/pachyderm/pull/7988 addresses that now.  (Thanks for fixing!)

src/version/client.go
seslattery: the default value for all of the fields in buildInfo is "", because that's the default for a string in go, not nil.  So there's no chance of a nil pointer here, since we're always returning a struct of buildinfo, it's just filled with the default values when it's empty in this case.

msteffen:
Closing this PR as obsolete

PR #7944, opened 2022-07-20
Add errcheck to golangci-lint

albscui:
The goal of this PR is to detect unchecked errors in our code, and prevent them in the future. To accomplish this, we enable `errcheck` in golangci-lint.

In general, these are the types of changes:
- basic `//nolint:errcheck` comment to ignore errcheck
- errors that obviously should be returned are returned now
- log previously swallowed errors where returning is impossible
- gingerly use  `_ = funcThatErrors()` to explicitly ignore the error with code, and let the reader know that we are intentionally swallowing the error

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,8 +5,6 @@ run:
 linters:
   enable:
     - wrapcheck
-  disable:
-    - errcheck
 linters-settings:
   wrapcheck:
     ignoreSigs:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,12 +9,9 @@ linters:
 linters-settings:
   errcheck:
      exclude-functions:
-      - io.Copy
-      - github.com/pachyderm/pachyderm/v2/src/internal/backoff.Retry
-      - github.com/pachyderm/pachyderm/v2/src/internal/backoff.RetryNotify
-      - github.com/pachyderm/pachyderm/v2/src/internal/backoff.RetryUntilCancel
       - (*database/sql.Tx).Rollback
       - (*github.com/spf13/cobra.Command).MarkFlagCustom
+      - (*github.com/spf13/cobra.Command).Usage
   wrapcheck:
     ignoreSigs:
       - github.com/pachyderm/pachyderm/v2/src/internal/errors.Errorf
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,6 +7,14 @@ linters:
     - wrapcheck
     - nolintlint
 linters-settings:
+  errcheck:
+     exclude-functions:
+      - io.Copy
+      - github.com/pachyderm/pachyderm/v2/src/internal/backoff.Retry
+      - github.com/pachyderm/pachyderm/v2/src/internal/backoff.RetryNotify
+      - github.com/pachyderm/pachyderm/v2/src/internal/backoff.RetryUntilCancel
+      - (*database/sql.Tx).Rollback
+      - (*github.com/spf13/cobra.Command).MarkFlagCustom
   wrapcheck:
     ignoreSigs:
       - github.com/pachyderm/pachyderm/v2/src/internal/errors.Errorf
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,6 +5,7 @@ run:
 linters:
   enable:
     - wrapcheck
+    - nolintlint
 linters-settings:
   wrapcheck:
     ignoreSigs:
@@ -19,3 +20,8 @@ linters-settings:
       - .WithStack(
     ignorePackageGlobs:
       - github.com/pachyderm/pachyderm/v2/src/*
+issues:
+  exclude-rules:
+    - path: _test\.go # disable some linters for test files
+      linters:
+        - errcheck
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -25,8 +25,3 @@ linters-settings:
       - .WithStack(
     ignorePackageGlobs:
       - github.com/pachyderm/pachyderm/v2/src/*
-issues:
-  exclude-rules:
-    - path: _test\.go # disable some linters for test files
-      linters:
-        - errcheck
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,6 +2,8 @@ run:
   timeout: 5m
   skip-dirs:
     - cached-deps
+  build-tags:
+    - k8s
 linters:
   enable:
     - wrapcheck
--- a/src/internal/metrics/segment.go
+++ b/src/internal/metrics/segment.go
@@ -23,6 +23,7 @@ func newSegmentClient() *analytics.Client {
 func reportClusterMetricsToSegment(client *analytics.Client, metrics *Metrics) {
 	// We're intentionally ignoring an error here because metrics code is
 	// non-critical
+	// nolint:errcheck
 	client.Track(&analytics.Track{
 		Event:       "cluster.metrics",
 		AnonymousId: metrics.ClusterID,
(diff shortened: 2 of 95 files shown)


src/internal/miscutil/miscutil.go
```diff
@@ -16,9 +16,7 @@ func WithPipe(wcb func(w io.Writer) error, rcb func(r io.Reader) error) error {
 	pr, pw := io.Pipe()
 	eg := errgroup.Group{}
 	eg.Go(func() error {
-		err := wcb(pw)
-		pw.CloseWithError(err)
-		return errors.EnsureStack(err)
+		return errors.EnsureStack(pw.CloseWithError(wcb(pw)))
```
jrockway, line 19: I'm not 100% convinced that PipeWriter.CloseWithError returns the error that you pass to it.  I think it might return nil actually.  (see src/io/pipe.go.)

src/server/cmd/pachd/main.go
```diff
@@ -208,7 +208,7 @@ func doEnterpriseMode(config interface{}) (retErr error) {
 	defer func() {
 		if retErr != nil {
 			log.WithError(retErr).Print("failed to start server")
-			pprof.Lookup("goroutine").WriteTo(os.Stderr, 2)
+			pprof.Lookup("goroutine").WriteTo(os.Stderr, 2) //nolint:errcheck // already logging error above
```
jrockway, line 211: I guess the error the linter is returning is the error from WriteTo.  Not much we can do if you can't write to stderr though.

src/server/pfs/fuse/loopback.go
```diff
@@ -168,7 +168,7 @@ func (n *loopbackNode) Mknod(ctx context.Context, name string, mode, rdev uint32
 	}
 	st := syscall.Stat_t{}
 	if err := syscall.Lstat(p, &st); err != nil {
-		syscall.Rmdir(p)
+		syscall.Rmdir(p) //nolint:errcheck
```
jrockway, line 171: These should also be multierr.

src/server/pfs/fuse/loopback.go
```diff
@@ -194,7 +194,7 @@ func (n *loopbackNode) Mkdir(ctx context.Context, name string, mode uint32, out
 	}
 	st := syscall.Stat_t{}
 	if err := syscall.Lstat(p, &st); err != nil {
-		syscall.Rmdir(p)
+		syscall.Rmdir(p) //nolint:errcheck // favour outter error instead
```
jrockway, line 197: "outer"

src/server/worker/common/common.go
```diff
@@ -43,14 +43,14 @@ func DatumID(inputs []*Input) string {
 	hash := pfs.NewHash()
 	for _, input := range inputs {
 		hash.Write([]byte(input.Name))
-		binary.Write(hash, binary.BigEndian, int64(len(input.Name)))
+		binary.Write(hash, binary.BigEndian, int64(len(input.Name))) //nolint:errcheck
```
jrockway, line 46: Yeah, this can't actually return an error.  binary.Write returns the underlying error from hash.Write, and hash.Write can't error.

jrockway (approved):
This is nice to have.  I think we're being a little liberal with suppressing errors, but we were already doing that without the lint check.  I would go to https://pachyderm.atlassian.net/browse/CORE-913 and add a task to clean up some of these.

I feel the same way about the hard-coded excludes, they should be removed later, but shouldn't block this PR.  Right now we're excluding every function (by not running the linter at all) so this is strictly better.

.golangci.yml
```diff
@@ -5,9 +5,16 @@ run:
 linters:
   enable:
     - wrapcheck
-  disable:
-    - errcheck
+    - nolintlint
 linters-settings:
+  errcheck:
+     exclude-functions:
+      - io.Copy
```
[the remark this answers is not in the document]
albscui: Oops on the io.Copy, I was just testing it out with a simple package name.

I'm aligned with you on the retry stuff. It's just that there are few places where we handle the error, and a lot more places where we don't, so I thought that was the default pattern.

For rollback, there's not much we can do if a database fails to rollback a commit.

src/internal/miscutil/miscutil.go
albscui: Omg good catch!

src/server/worker/pipeline/transform/registry.go
```diff
@@ -152,7 +152,7 @@ func (reg *registry) startJob(jobInfo *pps.JobInfo) (retErr error) {
 		if pj.ji.Details.JobTimeout != nil {
 			pj.logger.Logf("cancelling job at: %+v", afterTime)
 			timer := time.AfterFunc(afterTime, func() {
-				reg.killJob(pj, "job timed out")
+				reg.killJob(pj, "job timed out") //nolint:errcheck
```
PFedak, line 155: might be a good one to log

src/server/worker/worker.go
```diff
@@ -60,9 +60,14 @@ func NewWorker(
...
-		ppsutil.FailPipeline(env.Context(), env.GetDBClient(), driver.Pipelines(),
+		if err := ppsutil.FailPipeline(
+			env.Context(),
+			env.GetDBClient(),
+			driver.Pipelines(),
 			pipelineInfo.SpecCommit,
-			"nothing to run: no transform.cmd")
+			"nothing to run: no transform.cmd"); err != nil {
+			return nil, err
+		}
```
PFedak, line 70: this should probably have been returning regardless, the workers for the failed pipeline will get cleaned up, anyway

PFedak (approved):
Thanks for taking on this change! My comments are less important than getting this change in

src/internal/metrics/segment.go
```diff
@@ -21,8 +21,7 @@ func newSegmentClient() *analytics.Client {
 }
 
 func reportClusterMetricsToSegment(client *analytics.Client, metrics *Metrics) {
-	// We're intentionally ignoring an error here because metrics code is
-	// non-critical
+	//nolint:errcheck // ignoring error because metrics code is non-critical
```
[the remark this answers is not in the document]
albscui: it's a recommendation from: https://freshman.tech/linting-golang/#:~:text=//nolint%3Agosec%20//%20for%20faster%20performance

.golangci.yml
```diff
@@ -20,4 +27,9 @@ linters-settings:
       - .WithMessagef(
       - .WithStack(
     ignorePackageGlobs:
-      - github.com/pachyderm/pachyderm/v2/src/*
\ No newline at end of file
+      - github.com/pachyderm/pachyderm/v2/src/*
+issues:
+  exclude-rules:
+    - path: _test\.go # disable some linters for test files
```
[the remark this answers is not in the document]
albscui: Yea, originally I thought we don't have to be as strict with tests. But now I realize it's better to include them so the tests are better quality.

src/server/cmd/pachctl/shell/shell.go
```diff
@@ -101,7 +101,7 @@ func (s *shell) executor(in string) {
 	cmd.Stdin = strings.NewReader("pachctl " + in)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
-	cmd.Run()
+	cmd.Run() //nolint:errcheck
```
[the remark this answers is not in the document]
albscui: I tried this, but it just prints `exit status 1`, so I don't think this information is that useful.

src/server/pfs/s3/util_test.go
```diff
@@ -138,7 +138,7 @@ func testRunner(t *testing.T, pachClient *client.APIClient, group string, driver
 	require.NoError(t, err)
 
 	go func() {
-		server.Serve(listener)
+		server.Serve(listener) //nolint:errcheck
```
PFedak, line 141: this doesn't need to be in an anonymous function now, right?

PFedak (approved):

PR #7947, opened 2022-07-20
fix userAccessibleOauthIssuerHost

tybritten:


jrockway (approved):
I checked where we use this and you're correct that it shouldn't contain the URL scheme or anything.

We should probably add some validation to ensure that it doesn't contain :// anywhere, kind of like the true/false check at line 115 above.

PR #7952, opened 2022-07-20
Append custom Cluster RBAC to mock RBAC

acohen4:


jrockway (approved):
Good idea.  Does it work if you end up specifying the same key more than once (perhaps you want to give Mr. Trout some special permissions, so that username is mentioned in the mockIDP bindings, but then mentioned in the user-supplied bindings)?

PR #7955, opened 2022-07-20
[CORE-850] Ignore no-op updates during branch propagation.

PFedak:
Stopping and starting a pipeline could lead to new jobs in pipelines
downstream of that pipeline even if nothing actually changed. This is
one of the easier ways to get into an open meta commit state, and is
generally confusing behavior.

This change moves the check that a branch is consistent with its
provenance to the very beginning, ignoring unchanged branches entirely.
This means we can skip the check for subvenant branches (which
necessarily will have a new upstream commit).


src/server/pfs/server/driver.go
```diff
@@ -776,32 +794,14 @@ func (d *driver) propagateBranches(txnCtx *txncontext.TransactionContext, branch
...
-				return err
-			}
-			ids = append(ids, provOfSubvBI.Head.ID)
-		}
-
-		if allSameString(ids) {
-			if ids[0] == txnCtx.CommitSetID {
-				hasNewCommits = true
-			}
-			continue
```
acohen4, line 803: I may be mis-reading something, but in the case that we only update a branch by starting/stopping a pipeline, wouldn't  this `continue` stop `propagateBranches` from creating a new commit/job?

PFedak: in the old code, if we stop a pipeline, or start one while everything in provenance is still up to date, yes, this skips the commit/job. However, any downstream branches would still hit this check

acohen4: thank you for explaining to me over voice :)

acohen4 (approved):
LGTM!

msteffen:
Recording a slack conversation about this PR for posterity (the reason that this PR improves things was hard for me to understand initially):

> **Q** I still don’t see how this changes the end behavior of `propagateBranches`. Could you provide an example?
> **A** Sure, it's just an order of operations thing. In the specific case of a stopped pipeline (or in general, removing all provenance from a branch), it's still true that the pipeline output will not receive a commit. The difference is in downstream pipelines. Before, we computed the union of subvenance, then perform the [head commit = input commits] check for each, so downstream pipelines could still get a commit, even though nothing upstream of them changed. Now, if one of the propagation sources doesn't need a commit, we don't bother looking at its subvenance at all.
This wouldn't matter, except that the current alias commit logic leaves things in a sort of meta-stable state, where running propagation on the meta branch will cause bad things to happen unless we actually want a new commit.

Aside: the meta-stable state is one in which a meta branch, say `A.meta` has one commit ID, but its inputs as well as `A` all have a different commit ID (typically an alias commit), and if `propagateCommits` ever scans `A.meta`, a new commit will be created unifying `A.meta` and `A.meta.inputs`, but no job will run because, at the outset, `A` and `A`'s inputs (which are the same as `A.meta`'s inputs) matched. The new commit on `A.meta` will, therefore, never be closed. Subsequent jobs will have issues (won't run?)

**Q**: So downstream pipelines could still get a commit, even though nothing upstream of them changed? I think this is what I’m missing. Wouldn’t the redundant subvenance check always prevent the subvenant pipelines from receiving a commit in the case that the parent is a no-op? Or is it that there’s some weird interaction if a no-op’s subvenance overlaps with an op’s subvenance
**A**: You don't need an interaction like that, it's the normal state of, say, pipeline A which is an input to pipeline B. If we update B (or modify one of its other inputs), an alias commit is created on A's output branch, and A's provenance, but not the meta branch. Then the meta branch will satisfy the "not all the same ID" condition, even if nothing upstream actually changes.

msteffen (approved):
LGTM

PR #7956, opened 2022-07-21
[CORE-851] Skip jobless commits when determining base commit.

PFedak:
We require that a job exists when computing base job datums, so
currently once a system gets in this state the job will error in an
infinite loop, and the only recourse is squashing the offending commit
(which may not be possible).

While we both plan to prevent deleting jobs and fix the known provenance
bugs which can lead to this state, this also prevents any future issues.


acohen4 (approved):
LGTM

PR #7957, opened 2022-07-21
bump installed minikube version

chainlink:


etc/testing/circle/install.sh
```diff
@@ -35,7 +35,7 @@ fi
 # To get the latest minikube version:
 # curl https://api.github.com/repos/kubernetes/minikube/releases | jq -r .[].tag_name | sort -V | tail -n1
 if [ ! -f cached-deps/minikube ] ; then
-    MINIKUBE_VERSION=v1.25.2 # If changed, also do etc/kube/start-minikube.sh
+    MINIKUBE_VERSION=v1.26.0 # If changed, also do etc/kube/start-minikube.sh
```
molinamelendezj, line 38: is that comment to also bump it in `start-minikube.sh` relevant for this change. Id also ssh onto the circle agent to verify this. i remember having an issue this getting bumped, somehow the cached version was still be used back when i was testing larger circle agent machines. but i want to say that might not be an issue anymore.

molinamelendezj (approved):
👍🏼

etc/testing/circle/install.sh
chainlink: Fixed in https://github.com/pachyderm/pachyderm/commit/f0c8bdb8628454ffe0695a6e9c97307981134ea0 The comment was actually incorrect (referred to the kubernetes version not the minikube version, I've moved the comment, and synced the kubectl version with the currently used version of kube (1.25.2)

PR #7962, opened 2022-07-21
Deprecate enterprise root token and activateEnterprise helm values [CORE-893]

acohen4:


etc/helm/pachyderm/values.yaml
```diff
@@ -359,7 +359,7 @@ pachd:
     #apiGrpcPort:
     #  expose: true
     #  port: 30650
-  # instruct the config-pod to bootstrap enterprise on the cluster, even in upgrade releases
+  # DEPRECATED: activateEnterprise is no longer used.
```
jrockway, line 362: For the 2.3 release, can we just remove these?  I'm not sure what our guarantees about stability are.  (But I would say, removed fields are bad for our existing customers, but good for future users.  One less thing to be confused by.)

jrockway (approved):

PR #7965, opened 2022-07-21
Add docker executor for golang

chainlink:


molinamelendezj (approved):
LGTM but cc @seslattery on this one this _might_ collide with his PR for the go version.

chainlink:
Yeah, I like sean's pipeline variable for go version idea, I think we could use that on top of this.

PR #7966, opened 2022-07-21
Add circle validate job to makefile

chainlink:


Makefile
```diff
@@ -417,6 +417,10 @@ spellcheck:
 check-buckets:
 	./etc/testing/circle/check_buckets.sh
 
+validate-circle:
+	circleci config validate .circleci/main.yml
```
molinamelendezj, line 421: this assumes where ever this is executed from has the circleci cli installed. not a biggy, assume someone would know to install it once they see the error.

molinamelendezj (approved):

PR #7968, opened 2022-07-22
[Doc 392] Proxy and Enterprise Server

nadegepepin:


doc/docs/master/deploy-manage/deploy/deploy-w-proxy.md
```diff
@@ -553,3 +553,72 @@ The `pachyderm-proxy` service also routes Pachyderm's [**S3 gateway**](../../man
 **access Pachyderm's repo through the S3 protocol**) on port 80 (note the endpoint in the diagram below).
 
 ![Global S3 Gateway with Proxy](../../images/main-s3-gateway-with-proxy.png)
+
+## Changes to the Enterprise Server Setup
+
+Your enterprise server is [deployed in the same way as any regular cluster](#deploy-pachyderm-in-production-with-a-proxy) with a few differences (no object-store and one unique PostgreSQL database required: `dex`). The same applies when deploying an enterprise server with a proxy.
```
jrockway, line 559: I think we still require the pachyderm database.  https://pachyderm.atlassian.net/browse/FREQQ-70 refers to enterprise *members* not needing a dex database, but the enterprise server still needs to store some data in pachyderm like the license key, in addition to needing dex to serve the auth server.

I verified this by installing with `pachd.enabled=false` and `enterpriseServer.enabled=true` and looking at the database; pachd (in the `pach-enterprise-...` pod) applies the migrations, and it writes to tables like `license.clusters`.

doc/docs/master/deploy-manage/deploy/deploy-w-proxy.md
```diff
@@ -553,3 +553,72 @@ The `pachyderm-proxy` service also routes Pachyderm's [**S3 gateway**](../../man
...
+
+!!! Attention
+    Enabling an embedded enterprise server with your pachd as part of the same helm installation will not work with the proxy. 
+    You can use a standalone enterprise server instead.
+
+Follow your regular [enterprise server deployment and configuration instructions](../../../enterprise/auth/enterprise-server/setup){target=_blank}, but for those few steps:
+
+- [Section 1: Deploy an enterprise server](#1-deploy-an-enterprise-server), in the values.yaml provided as examples:
+
+    - Remove the `externalService` section and replace it with `proxy`:
```
jrockway, line 571: Technically `externalService` is one level deeper (child of the `pachd` key) than `proxy` (which is at the root).

doc/docs/master/deploy-manage/deploy/deploy-w-proxy.md
```diff
@@ -553,3 +553,72 @@ The `pachyderm-proxy` service also routes Pachyderm's [**S3 gateway**](../../man
...
+    - Set up the issuer in the idp config between the enterprise server and your cluster:
+    
+    ```shell
+    echo "issuer: http://<enterprise-external-IP-or-dns>" | pachctl idp set-config --config -
+    ```
+
+    - For each registered cluster, enable auth:
+
+        ```shell
+        pachctl auth activate --client-id <my-pachd-config-name> --redirect http://<pachd-rxternal-IP-or-DNS>:80/authorization-code/callback
```
jrockway, line 604: I know you already caught this, but `external-IP` not `rxternal-IP`.

The :80 can be omitted here if the scheme is http (and the :443 can be omitted if the scheme is https).

jrockway (approved):
Looks good!

doc/docs/master/deploy-manage/deploy/deploy-w-proxy.md
nadegepepin: Yes, you are correct https://pachyderm.slack.com/archives/CDWDPKPMY/p1658496327199979

Made the change to request the 2 databases for the Standalone Enterprise server.

PR #7969, opened 2022-07-22
[CORE-868] Make logs clearer

robert-uhl:
This works by attaching a method field to anything logged with
LogStep.

Example:

```
$ cat > /tmp/files <<EOF
> http://imgur.com/46Q8nDz.png
> http://imgur.com/8MN9Kg0.png
> http://imgur.com/g2QnNqa.png
> EOF
$ pachctl create repo images
$ pachctl put file images@master -i /tmp/files
$ pachctl logs
⋮
{"log":"{\"message\":\"\",\"method\":\"pfs.API.ModifyFile \",\"request\":{\"Body\":{\"set_commit\":{\"branch\":{\"repo\":{\"name\":\"images\",\"type\":\"user\"},\"name\":\"master\"}}}},\"severity\":\"info\",\"time\":\"2022-07-22T15:39:22.076019361Z\"}\n","stream":"stderr","time":"2022-07-22T15:39:22.076396414Z"}
{"log":"time=\"2022-07-22T15:39:22Z\" level=info msg=\"started UnorderedWriter.serialize during %s\" method=/pfs_v2.API/ModifyFile\n","stream":"stderr","time":"2022-07-22T15:39:22.541098193Z"}
{"log":"time=\"2022-07-22T15:39:22Z\" level=info msg=\"finished UnorderedWriter.serialize during %s\" duration=230.343875ms method=/pfs_v2.API/ModifyFile\n","stream":"stderr","time":"2022-07-22T15:39:22.771429479Z"}
{"log":"time=\"2022-07-22T15:39:22Z\" level=info msg=\"started directly compacting 1 file sets\" method=/pfs_v2.API/ModifyFile\n","stream":"stderr","time":"2022-07-22T15:39:22.771481892Z"}
{"log":"time=\"2022-07-22T15:39:22Z\" level=info msg=\"finished directly compacting 1 file sets\" duration=\"20.775µs\" method=/pfs_v2.API/ModifyFile\n","stream":"stderr","time":"2022-07-22T15:39:22.771484386Z"}
{"log":"{\"duration\":0.752235252,\"message\":\"\",\"method\":\"pfs.API.ModifyFile \",\"request\":{\"Body\":{\"set_commit\":{\"branch\":{\"repo\":{\"name\":\"images\",\"type\":\"user\"},\"name\":\"master\"}}}},\"response\":{},\"severity\":\"info\",\"time\":\"2022-07-22T15:39:22.784132086Z\"}\n","stream":"stderr","time":"2022-07-22T15:39:22.784205864Z"}

⋮
```


src/internal/storage/fileset/unordered_writer.go
```diff
@@ -93,11 +93,11 @@ func (uw *UnorderedWriter) validate(p string) error {
 
 // serialize will be called whenever the in-memory file set is past the memory threshold.
 // A new in-memory file set will be created for the following operations.
-func (uw *UnorderedWriter) serialize() error {
+func (uw *UnorderedWriter) serialize(ctx context.Context) error {
 	if uw.buffer.Empty() {
 		return nil
 	}
-	return miscutil.LogStep("UnorderedWriter.serialize", func() error {
+	return miscutil.LogStep(ctx, "UnorderedWriter.serialize during %s", func() error {
```
jrockway, line 100: This function doesn't apply format args, so the raw logs end up having %s in them.  (This ends up being the `name` argument to `entry.Infof("started %v", name)` etc.)

I think you will need to fmt.Sprintf whatever you want %s to be here just like you do at line 200 below.

src/internal/middleware/logging/interceptor.go
```diff
@@ -361,6 +361,7 @@ func isNilInterface(x interface{}) bool {
 }
 
 func (li *LoggingInterceptor) UnaryServerInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, retErr error) {
+	ctx = withMethodName(ctx, info.FullMethod)
```
jrockway, line 364: BTW, this is kind of built into grpc, but not in an easy to use way.

The full method name is written into the context here: https://github.com/grpc/grpc-go/blob/master/internal/transport/http2_server.go#L574

It is up to you to create a server that is initialized with a StatsHandler that actually puts this into the context.  Here are some implementations of that: https://pkg.go.dev/search?q=TagRPC&m=

TL;DR is that we can probably do all the logging and adding-the-name-to-the-context without an interceptor or custom stream type.  Maybe we should open up a ticket to improve this implementation in general.

jrockway (approved):
This is a nice improvement.

I am kind of torn on how helpful this is to the user; their URL never appears in the log message, so they're not going to connect these logs to the file download.  Whether this log ends up being useful depends on the size of the buffer that we read into and the speed of their network connection.  So kind of spammy for a ton of small files downloaded fast; not informative for a large download over a congested connection.

For a giant file, users won't be able to estimate how long they'll be waiting, and I think they still have to stay connected or we cancel the download, so it's a pretty poor user experience.  It basically means you can't use Pachyderm if you want to access large files.  

We should probably come back to this at some point, and have a data structure that tracks the downloads for the entire modify file call, and outputs a log line every 10 seconds or something like that `{"downloads": {"http://big-files.example/huge.tgz": {"downloaded": 123456, "size":1000000000}, "http://some-other-file": {...}}`.  We should also send this back over the stream, of course.

src/internal/storage/fileset/unordered_writer.go
robert-uhl: Argh, that was a leftover mistake that I did not catch. Thanks for the good eye.

robert-uhl:
> We should probably come back to this at some point, and have a data structure that tracks the downloads for the entire modify file call, and outputs a log line every 10 seconds or something like that `{"downloads": {"http://big-files.example/huge.tgz": {"downloaded": 123456, "size":1000000000}, "http://some-other-file": {...}}`. We should also send this back over the stream, of course.

That is a nifty idea.

PR #7972, opened 2022-07-22
[CORE-803] Disable service account token automounting for loki by default

FahadBSyed:


jrockway (approved):
Good typo fix as well 😅

PR #7975, opened 2022-07-22
fix ci remote docker image tagging

molinamelendezj:
remote docker in ci has been tagging and pushing incorrect images. low impact, as only the integration may have been effected with this uncaught issue. but now that the docker images are being used in CI load tests and iaas envs it bubbled up as a bug.


chainlink:
Wouldn't this affect all releases done through circle? Looks like the release job uses `setup_remote_docker` as well

goreleaser/docker.yml
```diff
@@ -110,6 +110,7 @@ dockers:
       - pachyderm/pachd
       - pachyderm/pachd:local
       - "pachyderm/pachd:{{.Version}}"
+      - pachyderm/pachd:{{ .FullCommit }}
```
chainlink, line 113: Nice find! although we may want to use `.Env.VERSION` for consistency

Makefile
```diff
@@ -126,12 +126,12 @@ docker-gpu: docker-build-gpu docker-push-gpu
 docker-gpu-dev: docker-build-gpu docker-push-gpu-dev
 
 docker-tag:
```
chainlink, line 128: I think we can get rid of the `docker-tag` make target entirely

chainlink (changes requested):
Couple comments

goreleaser/docker.yml
chainlink: Chatted about this offline, `FullCommit` makes sense, but we should probably add a note to explain why this is different than `.Version` / `.Env.VERSION`

Makefile
chainlink: As mentioned, this should move from a make task to directly in the circle yaml

Makefile
```diff
@@ -125,12 +125,6 @@ docker-gpu: docker-build-gpu docker-push-gpu
 
 docker-gpu-dev: docker-build-gpu docker-push-gpu-dev
 
-docker-tag:
-	docker tag pachyderm/pachd:$(PREVIOUS_VERSION) pachyderm/pachd:$(VERSION)
-	docker tag pachyderm/worker:$(PREVIOUS_VERSION) pachyderm/worker:$(VERSION)
-	docker tag pachyderm/pachctl:$(PREVIOUS_VERSION) pachyderm/pachctl:$(VERSION)
-	docker tag pachyderm/mount-server:$(PREVIOUS_VERSION) pachyderm/mount-server:$(VERSION)
-
 docker-push:
```
chainlink, line 128: `docker-tag` should be removed from the phony section of the makefile as well.

chainlink (approved):
One small change, otherwise looks good!

PR #7976, opened 2022-07-22
add jsonnet file linting ci tests 

molinamelendezj:
Adds jsonnet file linting ci tests using the jsonnet-lint tool from [go-jsonnet](https://github.com/google/go-jsonnet). Will search recursively for any jsonnet files depth 10. May or may not require a bit of investigation but at least flags down something is wrong.

sample failure:

```
#!/bin/bash -eo pipefail
jsonnet-lint src/templates/sql_ingest_cron.jsonnet
src/templates/sql_ingest_cron.jsonnet:18:20 Unterminated String

        name: 'cron',


Problems found!

Exited with code exit status 2
CircleCI received exit code 2
```

### DIFF
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -434,6 +434,17 @@ jobs:
       - checkout
       - run: sudo npm install -g prettier
       - run: prettier -c etc/helm/pachyderm/values.yaml .circleci/config.yml .circleci/main.yml
+  jsonnet-lint:
+    resource_class: small
+    machine:
+      image: << pipeline.parameters.machine_image >>
+    steps:
+      - checkout
+      - run:
+          name: Install go-jsonnet
+          command: |
+            curl -Lo - https://github.com/google/go-jsonnet/releases/download/v0.18.0/go-jsonnet_0.18.0_Linux_x86_64.tar.gz | sudo tar -C /usr/local/bin -xvzf - go-jsonnet
+      - run: prettier -c etc/helm/pachyderm/values.yaml .circleci/config.yml .circleci/main.yml
   test-envoy:
     docker:
       - image: envoyproxy/envoy:v1.22.0
@@ -565,6 +576,7 @@ workflows:
           version: $CIRCLE_SHA1
           upload: true
       - check-prettier
+      - jsonnet-lint
       - test-go
       - test-envoy
       - integration-tests:
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -443,7 +443,7 @@ jobs:
       - run:
           name: Install go-jsonnet
           command: |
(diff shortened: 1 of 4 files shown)


.circleci/main.yml
```diff
@@ -434,6 +434,20 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Install go-jsonnet
+          command: |
+            go install github.com/google/go-jsonnet/cmd/jsonnet-lint@latest
+      - run:
+          name: lint all jsonnet files
+          command: |
+            find . -maxdepth 10 -name \*.jsonnet -exec jsonnet-lint {} \;
```
jrockway, line 450: I think that find exits with success even if the exec command fails:

```
$ find . -exec /bin/false ';' ; echo $?
0
```
We might have to find some way to make this exit non-zero, otherwise we won't actually be testing anything.

jrockway (approved):
Excellent addition!

.circleci/main.yml
jrockway: LOL github ate one variant of this comment, now it's back.

molinamelendezj: yup. you were right. good catch. tried it with a pipefail instead.[ here is example run where I broke something on purpose and it failed as expected](https://app.circleci.com/pipelines/github/pachyderm/pachyderm/13402/workflows/9741d9ef-9ee4-4a8b-81dd-4fd552b363d2/jobs/190136)

albscui (commented):
Can we add this to `etc/testing/lint.sh`? This way you don't have to add another step in circleci.

molinamelendezj:
> Can we add this to `etc/testing/lint.sh`? This way you don't have to add another step in circleci.

I tried this which works fine. but if you don't mind i'm actually going to keep it out of there for now. a small executer is more explicit for now. those tests are embedded in the MISC bucket, which would needlessly run docker build and minikube to run. these test are about to get a make over with some work kyle and I have been talking about, so once that clean up is done, ill move it there. those test also look to needlessly run twice under helm test to, which is something we'll see if we can clean up.

albscui:
> > Can we add this to `etc/testing/lint.sh`? This way you don't have to add another step in circleci.
> 
> I tried this which will work fine. but if you don't mind i'm actually going to keep it out of there for now. a small executer is more explicit for now. those tests are embedded in the MISC bucket, which would needlessly run docker build and minikube to run. these test are about to get a make over with some work kyle and I have been talking about, so once that clean up is done, ill move it there. those test also look to needlessly run twice under helm test to, which is something we'll see if we can clean up.

Yea I agree, ideally we have an explicit linting step in circleci that just runs all of our linters.

albscui (approved):

PR #7982, opened 2022-07-25
Try not running CI on docs changes

chainlink:


molinamelendezj:
Nice! 

@nadegepepin might send you some cake if you land this lol dynamic config makes this possible.

chainlink:
It's close! I'm currently figuring out the logic of not running Pach CI when _only_ docs change. ie, if there's a docs change + a code change, it should run the full suite

PR #7984, opened 2022-07-25
POC: Auto upgrades with release channels

seslattery:
If we started using release channels based off of github tags such as "stable", and tagged our docker image accordingly, we could provide an opt in mechanism to customers to allow them to run a static image tag, but remove some of the downsides by having our application check once a day if an upgrade is available (current "stable" tag sha doesn't match the sha from buildinfo), and if so, triggers a signal that is user configurable for the calling application to handle and gracefully shutdown.  By restarting the pod, and having k8s set to have imagePullPolicy: Always, an application should be able to check for updates at least once a day.  

This does require us to be more disciplined around gracefully rolling out breaking changes.

### DIFF
--- a/src/internal/auto_upgrade/auto_upgrade.go
+++ b/src/internal/auto_upgrade/auto_upgrade.go
@@ -1,0 +1,171 @@
+package auto_upgrade
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+	"strings"
+	"syscall"
+	"time"
+)
+
+const timeFormat = "15:04"
+
+type Option func(*autoUpdater)
+
+// Register starts a goroutine which is meant to wake up once a day after the specified check time,
+// and query github to see if the current commit sha of the provided tag matches the provided commit sha.
+// If not, it triggers the defined signal which is then the responsibility of the caller to handle gracefully.
+func Register(org, repo, tag, commit string, opts ...Option) error {
+	o := autoUpdater{
+		org:       org,
+		repo:      repo,
+		tag:       tag,
+		checkTime: "01:00",
+		interupt:  syscall.SIGTERM,
+	}
+	for _, opt := range opts {
+		opt(&o)
+	}
+
+	//c := make(chan os.Signal, 1)
+	//signal.Notify(c, o.interupt)
+	//errors := make(chan error, 1)
+	log.Printf("input time: %v", o.checkTime)
+	check, err := time.Parse(timeFormat, o.checkTime)
+	if err != nil {
+		return err
+	}
+	log.Printf("parsed time: %v", check)
+	// TODO: Properly handle errors here
+	go func(check time.Time, o *autoUpdater) {
+		for {
+			// avoid's restarting rapidly during the 1min of the checkTime.
+			time.Sleep(time.Minute * 1)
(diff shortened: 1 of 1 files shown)


molinamelendezj:
this is an interesting idea, but what guarantee do we have` imagePullPolicy: Always` is being set by the user/maintainer? could we add a check to say if this is not set or gets changed some kind of alert bubbles up that this feature is disabled. 

@chainlink and I had discussed something a long the lines of using `latest` as a tag and coming to same predicament.

seslattery:
> this is an interesting idea, but what guarantee do we have` imagePullPolicy: Always` is being set by the user/maintainer? could we add a check to say if this is not set or gets changed some kind of alert bubbles up that this feature is disabled.
> 
> @chainlink and I had discussed something a long the lines of using `latest` as a tag and coming to same predicament.

I mean we can control that through the helmchart - could definitely make a flag - AutoUpgrade = True or something, it automatically set's ImagePullPolicy to Always (Could also just make that the default) and probably an env var to control this behavior for pachd.  Worst case if a user misconfigures it- this just causes a restart a day and they don't automatically upgrade.

chainlink:
I don't think running off a static tag is a good idea as it makes it more difficult to figure out exactly what version in running. Further, I think this is premature, as we don't (yet) have the rigor around testing to ensure an automated release won't break a client install / create a support headache.

seslattery:
> I don't think running off a static tag is a good idea as it makes it more difficult to figure out exactly what version in running. Further, I think this is premature, as we don't (yet) have the rigor around testing to ensure an automated release won't break a client install / create a support headache.

I have a complementary PR https://github.com/pachyderm/pachyderm/pull/7941  that addresses the versioning issue, but basically can leverage go's built in buildinfo to get the commit sha and plumb that through versioning / have that log line emitted on startup (which I need to add to that PR yet). Going forwards, it should be trivial to determine a version, and not rely on mutable tags for that information. Secondly, I 100% agree that this shouldn't be released or shown to customers yet. However, if this proposal seems useful, we could leverage it in a bunch of internal environments like  helium or pachyderm.pachops.com to start practicing / testing our release processes (ie if we had a nightly release channel) and upgrades to ensure that in a quarter or so, we could potentially start rolling this out to customers.  This also does not replace any of our existing release processes, and would only ever be opt-in when and if we are comfortable enough exposing this to customers.

chainlink:
I still think the issue here is pach automatically updating itself vs having an external process do it (helm upgrade, etc) How would this handle something getting added/removed from the helm chart?

seslattery:
> I still think the issue here is pach automatically updating itself vs having an external process do it (helm upgrade, etc) How would this handle something getting added/removed from the helm chart?

I think configuration changes are an outside concern, this should actually just slot in nicely to existing gitops tools.  We'll have to increase our maturity to not have breaking configuration changes without a deprecation cycle, i.e. deprecate certain fields / support old and new field for at least a release, then eventually remove that old field in a followup release. We could possibly even upgrade this tooling to check for certain values that need to be switched and warn/fix/not update too if we wanted. This does not alleviate the need for helm/gitops, but it enables customers to not have to think about upgrades in certain scenarios and be able to be upgraded faster.

msteffen:
Closing this PR as obsolete

PR #7986, opened 2022-07-26
[CORE-894] Filter ListDatum results

robert-uhl:
This PR implements filtering by datum state.


src/pps/filter.go
```diff
@@ -0,0 +1,22 @@
+package pps
+
+import (
+	context "context"
```
brycemcanally, line 4: This happens in my environment as well. It seems like it must be a goimports issue.

src/pps/filter.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func (r *ListDatumRequest_Filter) Allow(ctx context.Context, item *DatumInfo) bool {
+	if r == nil {
+		return true
+	}
+	match := false
+	for _, s := range r.State {
+		if s == item.State {
+			match = true
```
brycemcanally, line 14: You could just return true here and `len(r.State) == 0` below.

src/pps/pps.proto
```diff
@@ -568,6 +568,11 @@ message InspectDatumRequest {
 }
 
 message ListDatumRequest {
+  // Filter restricts returned DatumInfo messages to those which match
+  // all of the filtered attributes.
+  message Filter {
```
brycemcanally, line 573: :+1: I'm a fan of the nested message types.

src/pps/filter.go
```diff
@@ -0,0 +1,22 @@
+package pps
+
+import (
+	context "context"
+)
+
+func (r *ListDatumRequest_Filter) Allow(ctx context.Context, item *DatumInfo) bool {
```
brycemcanally, line 7: Why does this take a context?

src/server/pachyderm_test.go
```diff
@@ -9653,6 +9653,90 @@ func TestListDatum(t *testing.T) {
...
+				Pfs: &pps.PFSInput{
+					Repo: repo2,
+					Glob: "/*",
+				},
+			}},
+		},
+	})
+	require.NoError(t, err)
+
+	var i int
```
brycemcanally, line 9693: We should probably add a common utility function for this to the clientsdk package. You can use the utility functions for the other streaming rpcs as reference implementations.

src/pps/filter.go
robert-uhl: Looks like that is how the generated `.pb.go` files refer to ittoo; I daresay that `goimports` is getting fooled by that.

src/pps/filter.go
robert-uhl: I wanted to leave the right groundwork for multiple attributes, but you’re probably right: doing the simple thing now makes sense.

src/pps/filter.go
robert-uhl: The original reason was two-fold. One is practical: to enable more advanced filters that had to do more work to handle timeouts. The other is to be able to pull in request-scoped values when logging, for example the method name being called.

I think that in general every exported function should take an initial context argument, and omit one only by exception.

src/server/pachyderm_test.go
robert-uhl: I added a utility function to do this, and switched over to use it.

I am concerned that this kind of callback-based function is less idiomatic and immediately grokkable than just cycling through the client stream. It’s the sort of thing would would be completely the right thing to do in another language (e.g. in a Lisp macro), but feels like it’s not a great fit for Go.

Like I wrote though, I switched over as requested.

src/pps/filter.go
brycemcanally: I don't think it would make sense to have more advanced logic in a function that is being applied to every datum in a stream. I think of this as more of a utility function and more advanced logic (i.e. looking up indexes) would be better handled at a layer above. Also, I generally don't like adding complexity (context, more complicated match check) to code that could be easily changed later to support it. It can make sense when dealing with external APIs or code that has a strong probability of needing it, but I generally err on the side of simplicity.

src/server/pachyderm_test.go
brycemcanally: Callback-based iteration is a pretty standard iteration pattern in Go (i.e. https://pkg.go.dev/path/filepath#Walk). Having the wrappers makes it simpler imo by removing boilerplate. It also gives us an opportunity to handle errors in a standard way.

brycemcanally (approved):
lgtm

PR #7990, opened 2022-07-26
Break cache go 1.18

seslattery:


.circleci/config.yml
```diff
@@ -39,6 +39,10 @@ jobs:
       BUCKET: << parameters.bucket >>
     steps:
       - checkout
+      - restore_cache:
+          keys:
+            - pach-build-dependencies-v2-{{ checksum "etc/testing/circle/install.sh" }}
+            - pach-build-dependencies-v2-
```
chainlink, line 45: The `pach-build-dependencies-v2-` is the problem here, we got rid of this in master.

PR #7991, opened 2022-07-27
Delete env.sh, move to circle config

chainlink:
Eliminate more scripts on scripts


molinamelendezj (approved):
LGTM

just for clarity, we are still using the `cached-deps` in `etc/testing/circle/install.sh` right

PR #7993, opened 2022-07-27
[CORE-788] Disable Automount at the ServiceAccount Level

tybritten:
Disable ServiceAccount automounting for pachd at the SA level, enable at the pod level.


jrockway (approved):
This might be weird when people upgrade to this version; service account is applied, a new pipeline is added, workers are created, but then fail because of the missing token.  It probably eventually fixes itself when pachd rolls out.

acohen4 (approved):
LGTM. Out of curiosity, what's the driving use-case? Some user wants to apply the service account to another deployment but doesn't want the API token there?

PR #7997, opened 2022-07-28
Fix user configured resources for pgbouncer deployment

BOsterbuhr:


tybritten:
Add limits too

BOsterbuhr:
> Add limits too 

That adds everything under resources, only if they are defined. Pachd has the same thing [here](https://github.com/pachyderm/pachyderm/blob/0d49dd80f6974bfe16979f430ef65423810bca50/etc/helm/pachyderm/templates/pachd/deployment.yaml#L337-L339)

tybritten (approved):
LGTM

PR #7998, opened 2022-07-28
[CORE-929] Fix segmentation violation in pachctl enterprise get-state

robert-uhl:


albscui (approved):

src/server/license/cmds/cmds.go
```diff
@@ -253,7 +253,7 @@ func GetStateCmd() *cobra.Command {
 				fmt.Println("No Pachyderm Enterprise license is configured")
 				return nil
 			}
-			ts, err := types.TimestampFromProto(resp.Info.Expires)
+			ts, err := types.TimestampFromProto(resp.GetInfo().GetExpires())
```
acohen4, line 256: I think this still suffers from the same risk, since GetInfo() can return nil right?

robert-uhl: `GetExpires` also protects against `nil`:

```go
func (m *TokenInfo) GetExpires() *types.Timestamp {
	if m != nil {
		return m.Expires
	}
	return nil
}
```

And so does `TimestampFromProto`, so we should be good to go.

acohen4: Misread the autogenerated code, LGTM

acohen4 (approved):

PR #8002, opened 2022-07-29
Configure Additional Dex Clients via Helm [CORE-892]

acohen4:


etc/helm/pachyderm/templates/enterprise-server/deployment.yaml
```diff
@@ -140,6 +140,24 @@ spec:
             configMapKeyRef:
               name: "pachyderm-identity-config"
               key: "oidc-clients"
+        {{- if or .Values.oidc.additionalOIDCClientsSecretName .Values.oidc.additionalOIDCClients }}
+        - name: IDP_CLIENTS_EXTRAS
```
chainlink, line 144: perhaps we should stick with `additional_clients` here for consistency?

chainlink (approved):
Overall looks good modulo one comment, perhaps we should add a test here demoing what the input should look like?

etc/helm/pachyderm/templates/enterprise-server/deployment.yaml
acohen4: good call 👍

PR #8006, opened 2022-08-02
introduce preflight check toggle

molinamelendezj:
[CORE-928] is requesting a preflight check toggle around pps kube validations. (claims to be blocked by it) Defaults `true` for reverse compatibility but has a helm toggle to disable it. will result in a warning when disabled i.e
`
{"message":"Preflight checks are disabled. This is not recommended.","severity":"warning","time":"2022-08-02T03:40:18.897643513Z"}`

this toggle is more of a unblocker, needs to be confirmed for a real solution. perhaps adding the security context to the test RC spec so opa violation are not hit.

[CORE-928]: https://pachyderm.atlassian.net/browse/CORE-928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


molinamelendezj:
👀 @BOsterbuhr

robert-uhl (approved):
LGTM, but I think we need to confirm that this actually solves the problem with a custom build in the customer environment before merging.

molinamelendezj:
> LGTM, but I think we need to confirm that this actually solves the problem with a custom build in the customer environment before merging.

@robert-uhl well the customer is already on alpha which is cut from master and its where they saw they issue. ill have to merge this in to master to cut an alpha. customs for 2.3.0 dont exist yet since we dont have a 2.3.0 release branch. 

thats why i made it a toggle to keep same behavior for everyone else

PR #8007, opened 2022-08-02
Grab local image from Env if available

chainlink:
Enables tests to run based on a passed in SHA

### DIFF
--- a/src/internal/minikubetestenv/deploy.go
+++ b/src/internal/minikubetestenv/deploy.go
@@ -70,12 +70,10 @@ type DeployOpts struct {
 type helmPutE func(t terraTest.TestingT, options *helm.Options, chart string, releaseName string) error
 
 func getLocalImage() string {
-	sha := os.Getenv("TEST_IMAGE_SHA")
-	if sha == "" {
-		return localImage
-	} else {
+	if sha := os.Getenv("TEST_IMAGE_SHA"); sha != "" {
 		return sha
 	}
+	return localImage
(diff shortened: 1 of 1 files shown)


src/internal/minikubetestenv/deploy.go
```diff
@@ -69,6 +69,15 @@ type DeployOpts struct {
 
 type helmPutE func(t terraTest.TestingT, options *helm.Options, chart string, releaseName string) error
 
+func getLocalImage() string {
+	sha := os.Getenv("TESTSHA")
```
acohen4, line 73: super nit: WDYT about changing env var to `TEST_IMAGE_SHA`?

acohen4 (approved):
LGTM

src/internal/minikubetestenv/deploy.go
```diff
@@ -69,6 +69,15 @@ type DeployOpts struct {
 
 type helmPutE func(t terraTest.TestingT, options *helm.Options, chart string, releaseName string) error
 
+func getLocalImage() string {
+	sha := os.Getenv("TEST_IMAGE_SHA")
```
jrockway, line 73: Best style here is to keep the variable binding in the if statement, and to not indent the "default":

```go
if sha := os.Getenv("TEST_IMAGE_SHA"); sha != "" {
    return sha
}
return localImage
```

jrockway (approved):
Good idea!

PR #8013, opened 2022-08-03
[CORE-946] Plumb a context through cmdutil.Main

robert-uhl:
This is not a huge improvement on its own; it lays the groundwork for handling termination in a unified manner later.


acohen4 (approved):
LGTM!

PR #8014, opened 2022-08-04
Use prebuilt image in CI

chainlink:
This uses the docker image built centrally instead of building pach in every CI run.

### DIFF
--- a/etc/testing/circle/build.sh
+++ b/etc/testing/circle/build.sh
@@ -29,4 +29,6 @@ fi
 git config user.email "donotreply@pachyderm.com"
 git config user.name "anonymous"
 git tag -f -am "Circle CI test v$VERSION" v"$VERSION"
-make docker-build
+#make docker-build
+docker build -f etc/test-images/Dockerfile.testuser -t pachyderm/testuser:local .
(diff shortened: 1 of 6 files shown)


etc/testing/circle/build.sh
```diff
@@ -29,4 +29,6 @@ fi
 git config user.email "donotreply@pachyderm.com"
 git config user.name "anonymous"
 git tag -f -am "Circle CI test v$VERSION" v"$VERSION"
-make docker-build
+#make docker-build
+docker build -f etc/test-images/Dockerfile.testuser -t pachyderm/testuser:local .
+docker build --network=host -f etc/test-images/Dockerfile.netcat -t pachyderm/ubuntuplusnetcat:local .
```
jrockway, line 34: Can we just push this thing and stop building it in CI?  Looks like the "build" step in the bucket tests is now just building this, and it probably hasn't ever changed in its entire existence 😂

jrockway (approved):
Very nice.  Please let Jose take a look before merging.

etc/testing/circle/build.sh
chainlink: 💯 Planning on doing that in a future PR. I actually have a sister PR which eliminates one of these entirely

PR #8015, opened 2022-08-04
Add POSTGRES_DATABASE to pgbouncer liveness cmd

BOsterbuhr:
If `POSTGRES_USERNAME` and `POSTGRES_DATABASE` did not match the liveness probe would fail to connect and would restart pgbouncer


tybritten (approved):
LGTM. Odd that this is needed.. PG should use the default db if one isn't specified

PR #8019, opened 2022-08-04
[CORE-951] Factor out S3 server

robert-uhl:


src/server/cmd/pachd/main.go
```diff
@@ -948,6 +916,47 @@ func waitForError(name string, errChan chan error, required bool, f func() error
...
+			errCh <- errors.Wrapf(err, "couldn't load TLS cert for s3gateway: %v", err)
+		}
+		srv.TLSConfig = &gotls.Config{GetCertificate: cLoader.GetCertificate}
+		errCh <- errors.EnsureStack(srv.ListenAndServeTLS(certPath, keyPath))
+	}()
+	select {
+	case <-ctx.Done():
+		// NOTE: using context.Background here means that shutdown will
+		// wait until all requests terminate.
+		return errors.EnsureStack(srv.Shutdown(context.Background()))
```
acohen4, line 954: Could we add a log line that we're shutting down with an ctx.Err if one exists?

acohen4 (approved):
LGTM

PR #8033, opened 2022-08-05
[CORE-948] Handle signals uniformly

robert-uhl:


src/server/cmd/pachd/main.go
```diff
@@ -711,25 +712,25 @@ func doFullMode(ctx context.Context, config interface{}) (retErr error) {
...
-		log.Println("terminating; waiting for pachd server to gracefully stop")
-		var g, _ = errgroup.WithContext(ctx)
-		g.Go(func() error { externalServer.Server.GracefulStop(); return nil })
-		g.Go(func() error { internalServer.Server.GracefulStop(); return nil })
-		if err := g.Wait(); err != nil {
-			log.Errorf("error waiting for pachd server to gracefully stop: %v", err)
+	eg.Go(func() error {
+		<-ctx.Done() // wait for main context to complete
+		var eg = new(errgroup.Group)
+		log.Println("terminating; waiting for paused pachd server to gracefully stop")
```
acohen4, line 718: change message to (full) pachd server

src/server/cmd/pachd/main.go
```diff
@@ -711,25 +712,25 @@ func doFullMode(ctx context.Context, config interface{}) (retErr error) {
...
+		eg.Go(func() error { externalServer.Server.GracefulStop(); return nil })
+		eg.Go(func() error { internalServer.Server.GracefulStop(); return nil })
+		err := eg.Wait()
+		if err != nil {
+			log.Errorf("error waiting for paused pachd server to gracefully stop: %v", err)
```
acohen4, line 723: same as above comment

src/server/cmd/pachd/main.go
```diff
@@ -711,25 +712,25 @@ func doFullMode(ctx context.Context, config interface{}) (retErr error) {
...
-		<-c
-		log.Println("terminating; waiting for pachd server to gracefully stop")
-		var g, _ = errgroup.WithContext(ctx)
-		g.Go(func() error { externalServer.Server.GracefulStop(); return nil })
-		g.Go(func() error { internalServer.Server.GracefulStop(); return nil })
-		if err := g.Wait(); err != nil {
-			log.Errorf("error waiting for pachd server to gracefully stop: %v", err)
+	eg.Go(func() error {
+		<-ctx.Done() // wait for main context to complete
+		var eg = new(errgroup.Group)
```
acohen4, line 717: what do you think of having a timeout for Gracefully stopping?

src/server/cmd/pachd/main.go
```diff
@@ -711,25 +712,25 @@ func doFullMode(ctx context.Context, config interface{}) (retErr error) {
...
 	eg.Go(maybeIgnoreErrorFunc("Prometheus Server", requireNoncriticalServers, func() error { return prometheusServer{port: env.Config().PrometheusPort}.listenAndServe(ctx) }))
-	go func(c chan os.Signal) {
-		<-c
-		log.Println("terminating; waiting for pachd server to gracefully stop")
-		var g, _ = errgroup.WithContext(ctx)
-		g.Go(func() error { externalServer.Server.GracefulStop(); return nil })
-		g.Go(func() error { internalServer.Server.GracefulStop(); return nil })
-		if err := g.Wait(); err != nil {
-			log.Errorf("error waiting for pachd server to gracefully stop: %v", err)
+	eg.Go(func() error {
```
acohen4, line 715: should this logic be in the other modes?

src/server/cmd/pachd/main.go
```diff
@@ -86,7 +88,8 @@ func main() {
 	// set GOMAXPROCS to the container limit & log outcome to stdout
 	maxprocs.Set(maxprocs.Logger(log.Printf)) //nolint:errcheck
 	log.Infof("version info: %v", version.Version)
-	ctx := context.Background()
+	ctx, cancel := signal.NotifyContext(context.Background(), notifySignals...)
```
acohen4, line 91: is there a way we can get the log for the error that caused the shutdown?

src/server/cmd/pachd/main.go
robert-uhl: Done. And … yipes.

src/server/cmd/pachd/main.go
robert-uhl: Well, Kubernetes should stop it anyway, but … it’s not a bad idea. Do you think it should be configurable or just a hardcoded 10, 15 or 60 seconds?

src/server/cmd/pachd/main.go
robert-uhl: It was originally only in full & paused, not in enterprise or sidecar. I figured that I would leave things as they are, particularly because they’ll get unified in CORE-825.

src/server/cmd/pachd/main.go
robert-uhl: It gets logged in `cmdutil.Main`.

src/server/cmd/pachd/main.go
acohen4: I think hardcoding it is fine

acohen4: I think 10 seconds is reasonable

src/server/cmd/pachd/main.go
```diff
@@ -883,18 +881,20 @@ func doPausedMode(ctx context.Context, config interface{}) (retErr error) {
...
-		<-c
+	eg.Go(func() error {
+		<-ctx.Done() // wait for main context to complete
+		var eg = new(errgroup.Group)
 		log.Println("terminating; waiting for paused pachd server to gracefully stop")
-		var g, _ = errgroup.WithContext(ctx)
-		g.Go(func() error { externalServer.Server.GracefulStop(); return nil })
-		g.Go(func() error { internalServer.Server.GracefulStop(); return nil })
-		if err := g.Wait(); err != nil {
+		eg.Go(func() error { externalServer.Server.GracefulStop(); return nil })
```
acohen4, line 888: Out of curiosity, are we inverting the order on purpose?

acohen4 (approved):
Left a couple comments but LGTM

PR #8035, opened 2022-08-05
update to go 1.19

seslattery:
golangci-lint required an update to work with go 1.19, and the newest version of the linter errors out that we were still using ioutils package. So this fixes that by using the appropriate calls to os and io packages.  

I also had to slightly change the semantics of a small part of the TestMountDir test in fuse_test.go, as it was giving an interface conversion error.  Assuming my latest change works, this should be passing CI shortly (a run commenting out that line ended up passing fully).


jrockway (approved):
Very nice!

PR #8039, opened 2022-08-08
[CORE-825] Refactor pachd

robert-uhl:


src/internal/pachd/builder.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+	return nil
+}
+
+func (b *builder) initDexDB(ctx context.Context) error {
+	b.env.InitDexDB()
+	return nil
+}
+
+func (b *builder) maybeInitReporter(ctx context.Context) error {
```
FahadBSyed, line 151: I like this name 😆

src/internal/pachd/builder.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+	if b.loggingInterceptor == nil {
+		b.initLoggingInterceptor()
+	}
+	b.daemon.external, err = grpcutil.NewServer(
+		ctx,
+		true,
+		// Add an UnknownServiceHandler to catch the case where the user has a client with the wrong major version.
+		// Weirdly, GRPC seems to run the interceptor stack before the UnknownServiceHandler, so this is never called
+		// (because the version_middleware interceptor throws an error, or the auth interceptor does).
```
FahadBSyed, line 206: I think I might not fully understand this comment. If the unknownServiceHandler is never called, why are we adding it since it won't be able to catch the major upgrade error, right (because its never called)?

robert-uhl: I don’t know; it dates back to [this comment](https://github.com/pachyderm/pachyderm/commit/5ab8a0f7e632a31d41d080cc89913c1e262be2d6#diff-1aa4ab905e56432d62cd27077bfb5fbd35430854b2cbaab1de9cf2603e294c10R144-R160). Unfortunately Alysha is no longer here. Perhaps @acohen4 has some context?

src/internal/pachd/enterprise.go
```diff
@@ -0,0 +1,71 @@
...
+	builder
+}
+
+func (eb *EnterpriseBuilder) registerEnterpriseServer(ctx context.Context) error {
+	eb.enterpriseEnv = eprsserver.EnvFromServiceEnv(
+		eb.env,
+		path.Join(eb.env.Config().EtcdPrefix, eb.env.Config().EnterpriseEtcdPrefix),
+		eb.txnEnv,
+	)
+	apiServer, err := eprsserver.NewEnterpriseServer(
```
FahadBSyed, line 28: Is it worth having a helper function that calls the statements from lines 28 - 40 which takes a `enterpriseEnv` as a parameter, this way the code for creating and registering the server is not repeated?

src/internal/pachd/prometheus.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+type prometheusServer struct {
+	port uint16
+}
+
+// listenAndServe listens until ctx is cancelled; it then gracefully shuts down
+// the server, returning once all requests have been handled.
+func (ps prometheusServer) listenAndServe(ctx context.Context) error {
+	var (
```
FahadBSyed, line 22: Style question: any reason you chose to use a `var()` block instead of using the `:=` operator on these variables?

src/internal/pachd/s3.go
```diff
@@ -0,0 +1,55 @@
...
+		var (
+			certPath, keyPath, err = pachtls.GetCertPaths()
+		)
+		if err != nil {
+			log.Warnf("s3gateway TLS disabled: %v", err)
+			errCh <- errors.EnsureStack(srv.ListenAndServe())
+		}
+		cLoader := pachtls.NewCertLoader(certPath, keyPath, pachtls.CertCheckFrequency)
+		// Read TLS cert and key
+		err = cLoader.LoadAndStart()
```
FahadBSyed, line 39: Since this err doesn't seem to be used later, you could combine this into:
`if err := cLoader.LoadAndStart(); err != nil { // code... }`

src/internal/pachd/sidecar.go
```diff
@@ -0,0 +1,94 @@
...
+	builder
+}
+
+// NewSidecarBuilder returns an initialized SidecarBuilder.
+func NewSidecarBuilder(config any) *SidecarBuilder {
+	return &SidecarBuilder{newBuilder(config, "pachyderm-pachd-sidecar")}
+}
+
+func (sb *SidecarBuilder) registerAuthServer(ctx context.Context) error {
+
```
FahadBSyed, line 32: nit: extra newline here

src/server/cmd/pachd/main.go
```diff
@@ -2,70 +2,21 @@ package main
 
 import (
 	"context"
-	gotls "crypto/tls"
```
FahadBSyed, line 5: Awesome work refactoring this. Love seeing tons of red!

FahadBSyed (approved):
Just left a few nits. Since I am still the noob on the team, I would advise getting a second opinion from Alon or Sean, but overall looks awesome

src/internal/pachd/prometheus.go
robert-uhl: I prefer `var`: it is visually a separate section which emphasises that variables are being assigned.

src/internal/pachd/builder.go
```diff
@@ -0,0 +1,397 @@
...
+	transactionserver "github.com/pachyderm/pachyderm/v2/src/server/transaction/server"
+	"github.com/pachyderm/pachyderm/v2/src/transaction"
+	"github.com/pachyderm/pachyderm/v2/src/version"
+	"github.com/pachyderm/pachyderm/v2/src/version/versionpb"
+)
+
+// A bootstrapper is a type which needs to have some bootstrap code run after
+// initialization and before the dæmon starts listening.
+type bootstrapper interface {
+	EnvBootstrap(context.Context) error
```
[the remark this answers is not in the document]
robert-uhl: Renamed to `envBootstrapper` instead.

src/internal/pachd/builder.go
acohen4: I'm not sure - I agree it seems like it can be removed, but if we do so it should be in a follow up PR, so that this one can be interpreted as closer to a true refactor

acohen4 (approved):
Left some comments. Overall this is an awesome change! Really nice :)

src/internal/pachd/enterprise.go
```diff
@@ -0,0 +1,75 @@
...
+		eb.resumeHealth,
+		eb.daemon.serve,
+	)
+}
+
+// EnterpriseMode runs an enterprise-mode pachd.
+//
+// Enterprise mode is the enterprise server which is used to manage multiple
+// Pachyderm installations.
+func EnterpriseMode(ctx context.Context, config any) error {
```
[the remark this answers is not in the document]
robert-uhl: It’s shorter to type, and I think it’s clearer: the `config` value could be anything.

PR #8045, opened 2022-08-09
[CORE-612] Support Locking Repos Across Multiple PFS Masters Using Consistent Hashing Library

FahadBSyed:
## Description
This PR adds a `consistent hashing` library that we can use in distributed systems that use locks with a dynamic number of nodes.


src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,370 @@
...
+	"fmt"
+	"hash/crc32"
+	"path"
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/gogo/protobuf/types"
+	"github.com/pachyderm/pachyderm/v2/src/internal/collection"
+
```
brycemcanally, line 14: Why the arbitrary spaces?

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,370 @@
...
+	// LeaseTTL is used by nodes in a ring to refresh their liveness state.
+	LeaseTTL = int64(5) // number of seconds between lease renewals.
+)
+
+// Ring is a consistent hash Ring. Each process should only create one instance with a given prefix. Nodes can be
+// added locally to a Ring. A Ring watches for changes to the prefix to determine if new nodes have been added
+// or deleted.
+// A Ring should be initialized with a call to New.
+type Ring struct {
+	Client *etcd.Client   // Client will be used to make api calls against etcd.
```
brycemcanally, line 36: The field comments seem like overkill, and add some amount of maintenance overhead. I think non-trivial functionality / function comments make sense, but fields / variables should normally be self documenting based on there name and only needing comments in exceptional cases.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,370 @@
...
+	Ctx  context.Context // Ctx is the context returned by Lock.
+}
+
+// New is the default constructor for Ring. Creating a Ring starts a goroutine to watch prefix in etcd and creates a
+// node that continuously refreshes a lease in etcd.
+func New(ctx context.Context, prefix string, client *etcd.Client, logger *logrus.Logger, id ...string) (*Ring, error) {
+	if id == nil {
+		id = []string{uuid.New()}
+	}
+
```
brycemcanally, line 76: Nit, but I prefer not adding arbitrary spaces. If you feel a space is needed, then that probably means there is a logically separate block of code that could be separated by a comment explaining the block of code.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,370 @@
...
+	return index
+}
+
+// Get is a convenience wrapper around GetIndex that returns the Member that associates to a key.
+func (ring *Ring) Get(key string) Member {
+	return ring.members[ring.GetIndex(key)]
+}
+
+// Lock attempts to lock a key in etcd using consistent hashing. If key does not belong to the ring's node,
+// the call blocks until context is cancelled or expires.
```
brycemcanally, line 107: I would just refer to context canceled in the comments, rather than expires. The same mechanism is used for general cancelation and deadlines.

src/server/pfs/server/master.go
```diff
@@ -31,20 +33,24 @@ const (
...
-		defer masterLock.Unlock(masterCtx) //nolint:errcheck
-		eg, ctx := errgroup.WithContext(masterCtx)
+		eg, ctx := errgroup.WithContext(ctx)
 		trackerPeriod := time.Second * time.Duration(d.env.StorageConfig.StorageGCPeriod)
 		if trackerPeriod <= 0 {
 			d.log.Info("Skipping Storage GC")
 		} else {
-			d.log.Infof("Starting Storage GC with period=%v", trackerPeriod)
 			eg.Go(func() error {
+				lock := dlock.NewDLock(d.etcdClient, path.Join(d.prefix, masterLockPath, "gc"))
```
brycemcanally, line 43: Would you change the last component of the prefix to `storage-gc`. For the chunk gc, lets change it to `chunk-gc`.

src/server/pfs/server/master.go
```diff
@@ -31,20 +33,24 @@ const (
...
+				d.log.Infof("Starting Storage GC with period=%v", trackerPeriod)
+				gcCtx, err := lock.Lock(ctx)
```
brycemcanally, line 45: The gc run call isn't receiving this context. A common pattern with contexts is to overwrite `ctx` throughout a function so it is always handled correctly, or create a locally scoped `ctx` if you want to keep the outer `ctx` the same. In this case a locally scoped `ctx` would work, so change this to `ctx, err := ...`.

src/server/pfs/server/master.go
```diff
@@ -53,8 +59,18 @@ func (d *driver) master(ctx context.Context) {
 		if chunkPeriod <= 0 {
 			d.log.Info("Skipping Chunk Storage GC")
 		} else {
-			d.log.Infof("Starting Chunk Storage GC with period=%v", chunkPeriod)
 			eg.Go(func() error {
+				lock := dlock.NewDLock(d.etcdClient, path.Join(d.prefix, masterLockPath, "chunk"))
+				d.log.Infof("Starting Chunk Storage GC with period=%v", chunkPeriod)
+				gcCtx, err := lock.Lock(ctx)
```
brycemcanally, line 65: Same as above.

src/server/pfs/server/master.go
```diff
@@ -70,21 +86,32 @@ func (d *driver) master(ctx context.Context) {
 }
 
 func (d *driver) finishCommits(ctx context.Context) error {
+	ctx, cancel := context.WithCancel(ctx)
```
brycemcanally, line 89: The better pattern imo is to have the defer cancel immediately after it is created to ensure it is canceled upon exit. Having code sandwiched between its creation and the deferred cancel can result with issues over time if more logic is inserted between the two.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,370 @@
...
+// lockInfo instances are created for keys by a ring's node.
+// A context lockInfo.Ctx is stored so the proper context is used when unlocking lockInfo.Lock
+type lockInfo struct {
+	Lock dlock.DLock     // Lock is a dlock that locks a key with a prefix in etcd.
+	Ctx  context.Context // Ctx is the context returned by Lock.
+}
+
+// New is the default constructor for Ring. Creating a Ring starts a goroutine to watch prefix in etcd and creates a
+// node that continuously refreshes a lease in etcd.
+func New(ctx context.Context, prefix string, client *etcd.Client, logger *logrus.Logger, id ...string) (*Ring, error) {
```
brycemcanally, line 72: This is more of a stylistic thing in the codebase, but I think it improves readability. We generally define parameters ordered by decreasing scope, so this would look more like `func New(ctx context.Context, client *etcd.Client, logger *logrus.Logger, prefix string, id ...string) (*Ring, error)`

src/server/pfs/server/master.go
```diff
@@ -96,7 +123,21 @@ func (d *driver) finishCommits(ctx context.Context) error {
 		repos[key] = cancel
 		go func() {
 			backoff.RetryUntilCancel(ctx, func() error { //nolint:errcheck
-				return d.finishRepoCommits(ctx, compactor, key)
+				lockCtx, err := ring.Lock(ctx, lockPrefix)
```
brycemcanally, line 126: We should probably have a subcontext that gets canceled in a defer to ensure resources are cleaned up during a retry.

src/server/pfs/server/master.go
```diff
@@ -96,7 +123,21 @@ func (d *driver) finishCommits(ctx context.Context) error {
...
+				if err != nil {
+					return errors.Wrap(err, "error locking repo lock")
+				}
+				defer func() {
+					unlockErr := ring.Unlock(lockPrefix)
+					if err == nil {
```
brycemcanally, line 132: This isn't going to do what you expect. You would need a named return to check whether `finishRepoCommits` returned an error. The common pattern for this is `retErr` being the named return and checking that in the defer. We may want to take a look at the `multierr` package for this so we can capture the `finishRepoCommits` error and `Unlock` error.

src/server/pfs/server/master.go
```diff
@@ -70,21 +86,32 @@ func (d *driver) master(ctx context.Context) {
...
-	err := d.repos.ReadOnly(ctx).WatchF(func(ev *watch.Event) error {
+	err = d.repos.ReadOnly(ctx).WatchF(func(ev *watch.Event) error {
 		if ev.Type == watch.EventError {
 			return ev.Err
 		}
 		key := string(ev.Key)
+		lockPrefix := path.Join(d.prefix, masterLockPath, "repos", key)
 		if ev.Type == watch.EventDelete {
 			if cancel, ok := repos[key]; ok {
+				err = ring.Unlock(lockPrefix)
```
brycemcanally, line 110: We can just return without the cancel if err isn't nil. All of the repos will get canceled in that case.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,370 @@
...
+	})
+
+	if errs != nil {
+		return errors.New(fmt.Sprintf("errors unlocking locks while rebalancing: %v", errs))
+	}
+
+	return nil
+}
+
+func defaultRing(ctx context.Context, prefix string, client *etcd.Client, logger *logrus.Logger) *Ring {
```
brycemcanally, line 278: This functionality seems like it should just be in the `New` function.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,370 @@
...
+	ring.node.Locks.Delete(key)
+
+	ring.Logger.WithFields(ring.lockLogFields(key)).Info("released lock")
+
+	return nil
+}
+
+// start kicks off watch(), creates a node, and renews its lease.
+// todo: replace with collections stuff.
+func (ring *Ring) start(id string) error {
```
brycemcanally, line 167: I think we should restructure this such that the setup is in the `New` function, but the watch functionality can be in a function and the gorutine is kicked off by the `New` function. Also, the watch should take the passed in context and respect it.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,370 @@
...
+// watch watches etcd keys with prefix Ring.Prefix to determine if a node has been added or removed.
+// When a node is added, the ring calls rebalance to release any locks that no longer associate to its node. Rebalance
+// on delete happens by default since each Member attempts to get all locks. When a Member is deleted,
+// the pending calls to Lock by each Member will go through on the nodes that associates to given lock.
+func (ring *Ring) watch(col collection.EtcdCollection) {
+	if err := col.ReadOnly(ring.Ctx).WatchF(func(event *watch.Event) error {
+		nodeKey := string(event.Key)
+		switch event.Type {
+		case watch.EventDelete:
+			ring.removeMemberById(ring.etcdToMember(nodeKey))
```
brycemcanally, line 228: Are we not rebalancing when a member is removed?

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,370 @@
...
+
+// Get is a convenience wrapper around GetIndex that returns the Member that associates to a key.
+func (ring *Ring) Get(key string) Member {
+	return ring.members[ring.GetIndex(key)]
+}
+
+// Lock attempts to lock a key in etcd using consistent hashing. If key does not belong to the ring's node,
+// the call blocks until context is cancelled or expires.
+func (ring *Ring) Lock(ctx context.Context, key string) (context.Context, error) {
+	ticker := time.NewTicker(time.Second * 5)
```
brycemcanally, line 109: I would add a TODO comment here to make this event driven. This polling strategy might be fine for now, but it may be problematic at scale.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,370 @@
...
+
+	if err := ring.createNode(id, etcdCol); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// createNode creates a lease, inserts itself as a key to etcd, keeps the lease alive in the background.
+// it is not idempotent, so if two rings attempt to add the same 'id', the second add will fail. This should
```
brycemcanally, line 184: UUID collisions aren't a realistic concern, so we can simplify this function a bit. I think the setup can happen outside of the callback since we can be confident there isn't a collision, so we don't need the ready channel. I think the setup should be in `New`, similar to the other comments, and the claim can be in a function that receives a context and is kicked off as a goroutine in the `New` function. I generally prefer setup to be simple enough to be captured in the `New` function and goroutines kicked off there so it is obvious what goroutines are running when you create the abstraction.

src/internal/consistenthashing/etcd.go
FahadBSyed: Sorry, that's my IDE getting confused

src/server/pfs/server/master.go
FahadBSyed: Oh I totally missed, that good catch!. I'll just pass `gcCtx` to `gc.RunForever(gcCtx)`

src/internal/consistenthashing/etcd.go
FahadBSyed: I split the set up for testing purposes, but I can get around this by defining a different constructor in the test file.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,370 @@
...
+		Id:   id,
+		Hash: ring.hashFn([]byte(id)),
+	}
+
+	ring.insertMember(member)
+}
+
+// insertMember updates Ring.members, and sorts Ring.members by the Member hash.
+// If the Member already exists, insertMember is a no-op. The member can be mocked for testing purposes.
+func (ring *Ring) insertMember(member Member) {
```
brycemcanally, line 307: I think we need to lock modification to the members list. A concurrent lock could be problematic.

src/server/pfs/server/master.go
FahadBSyed: For this one I defined a context in the go func() but not inside the backoff, I was seeing some unnecessary cancelling when the context was defined in the backoff, but let me know if it should be changed.

brycemcanally: The context should be in the callback. We want to ensure the internal resources are cleaned up if finish repo commits and unlock error.

src/server/pfs/server/master.go
```diff
@@ -95,8 +123,18 @@ func (d *driver) finishCommits(ctx context.Context) error {
...
+			defer cancel()
 			backoff.RetryUntilCancel(ctx, func() error { //nolint:errcheck
-				return d.finishRepoCommits(ctx, compactor, key)
+				lockCtx, err := ring.Lock(ctx, lockPrefix)
+				if err != nil {
+					return errors.Wrap(err, "error locking repo lock")
+				}
+				defer func() {
+					err := ring.Unlock(lockPrefix)
+					retErr = multierror.Append(retErr, errors.Wrap(err, "error unlocking"))
```
brycemcanally, line 135: I don't think this is going to be correct for the nil case. I think it should be something like this:
```
if err := ring.Unlock(lockPrefix); err != nil {
  if retErr == nil {
    retErr = err
  } else {
    retErr = multierror.Append(retErr, errors.Wrap(err, "error unlocking"))
  }
}
```

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+
+	"github.com/sirupsen/logrus"
+
+	etcd "go.etcd.io/etcd/client/v3"
+)
+
+// Ring is a consistent hash Ring. Each process should only create one instance with a given prefix. Nodes can be
+// added locally to a Ring. A Ring watches for changes to the prefix to determine if new nodes have been added
+// or deleted.
+// A Ring should be initialized with a call to New.
```
brycemcanally, line 28: I don't really think this needs to be specified.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+// lockInfo instances are created for keys by a ring's node.
+// A context lockInfo.Ctx is stored so the proper context is used when unlocking lockInfo.Lock
+type lockInfo struct {
+	Lock dlock.DLock
+	Ctx  context.Context
+}
+
+// New is the default constructor for Ring. Creating a Ring starts a goroutine to watch prefix in etcd and creates a
+// node that continuously refreshes a lease in etcd.
+func New(ctx context.Context, client *etcd.Client, logger *logrus.Logger, prefix string, id ...string) (*Ring, error) {
```
brycemcanally, line 61: I would prefer us to not let the id be configurable. A client could shoot themselves in the foot if they set the same id for multiple instances.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+		client:  client,
+		hashFn:  crc32.ChecksumIEEE,
+		logger:  logger,
+		members: []Member{localMember},
+		node: node{
+			Member: localMember,
+			Locks:  map[string]lockInfo{},
+		},
+		prefix:    prefix,
+		stateLock: sync.Mutex{},
```
brycemcanally, line 79: The default value can be used for a mutex, so you don't need to initialize them.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+	if id == nil {
+		id = []string{uuid.New()}
+	}
+	localMember := Member{
+		Id:   id[0],
+		Hash: crc32.ChecksumIEEE([]byte(id[0])),
+	}
+	ring := &Ring{
+		client:  client,
+		hashFn:  crc32.ChecksumIEEE,
```
brycemcanally, line 71: I think we should just refactor this to a global const and only reference that in the code (including where we use it above for the local member). I don't realistically see this changing or being configurable.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// Lock attempts to lock a key in etcd using consistent hashing. If key does not belong to the ring's node,
+// the call blocks until context is cancelled.
+func (ring *Ring) Lock(ctx context.Context, key string) (context.Context, error) {
+	ring.stateLock.Lock()
+	defer ring.stateLock.Unlock()
+	ticker := time.NewTicker(time.Second * 5)
+	defer ticker.Stop()
+	// todo: this should be event driven instead of by polling for better scalability.
```
brycemcanally, line 114: I think you might need to change this to uppercase (`TODO`) to work with tooling for tracking todos.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+		logger:  logger,
+		members: []Member{localMember},
+		node: node{
+			Member: localMember,
+			Locks:  map[string]lockInfo{},
+		},
+		prefix:    prefix,
+		stateLock: sync.Mutex{},
+	}
+	etcdCol := collection.NewEtcdCollection(ring.client, path.Join(ring.prefix, "nodes"),
```
brycemcanally, line 81: We can just use empty type rather than bool, also the arbitrary newline is a bit weird. I am generally in favor of letting the lines run long to put pressure on reducing the number of parameters to a function when that occurs.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// GetIndex returns the index of the Member associated with key. A key belongs to the first Member whose hash is
+// greater than the key. If no Members are greater, then the Ring wraps around to the first Member.
+func (ring *Ring) GetIndex(key string) int {
+	keyHash := ring.hashFn([]byte(key))
+	index := sort.Search(len(ring.members), func(i int) bool {
+		return ring.members[i].Hash >= keyHash
+	})
+	// wrap around to the first node in the Ring.
```
brycemcanally, line 95: `node` -> `member`

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+// Lock attempts to lock a key in etcd using consistent hashing. If key does not belong to the ring's node,
+// the call blocks until context is cancelled.
+func (ring *Ring) Lock(ctx context.Context, key string) (context.Context, error) {
+	ring.stateLock.Lock()
+	defer ring.stateLock.Unlock()
+	ticker := time.NewTicker(time.Second * 5)
+	defer ticker.Stop()
+	// todo: this should be event driven instead of by polling for better scalability.
+	for {
+		if len(ring.members) != 0 {
```
brycemcanally, line 116: Can this occur because a delete event could be received for the member associated with this node?

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+	if err := info.Lock.Unlock(info.Ctx); err != nil {
+		return errors.EnsureStack(err)
+	}
+	delete(ring.node.Locks, key)
+	ring.logger.WithFields(ring.lockLogFields(key)).Info("released lock")
+	return nil
+}
+
+// createNode creates a lease, inserts itself as a key to etcd, keeps the lease alive in the background.
+func (ring *Ring) createNode(ctx context.Context, col collection.EtcdCollection, id string) {
```
brycemcanally, line 164: Would you move this and the watch function closer to the constructor. It improves the readability a bit imo. Generally, I like to have functions like this defined in order and close to where they are used.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+			return nil, errors.EnsureStack(ctx.Err())
+		case <-ticker.C:
+		}
+		ring.stateLock.Lock()
+	}
+	if info, alreadyExists := ring.node.Locks[key]; alreadyExists {
+		lockCtx := info.Ctx
+		return lockCtx, nil
+	}
+	nodeLock := dlock.NewDLock(ring.client, key)
```
brycemcanally, line 135: Should the prefix be a part of this?

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+		Lock: nodeLock,
+		Ctx:  lockCtx,
+	}
+	ring.logger.WithFields(ring.lockLogFields(key)).Info("claimed lock")
+	return lockCtx, nil
+}
+
+// Unlock attempts to unlock a key in etcd if and only if a ring's node owns a lock reference to it.
+// Remote Member instances must give up their own locks.
+func (ring *Ring) Unlock(key string) error {
```
brycemcanally, line 150: We need the state lock here.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+	info, localLock := ring.node.Locks[key]
+	if !localLock {
```
brycemcanally, line 152: Would you rename this to exists. That is a more common name for checking existence in a map.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+func (ring *Ring) rebalance() error {
+	var errs []error
+	for key, lockInfo := range ring.node.Locks {
+		if ring.Get(key).Id == ring.node.Id {
+			continue
+		}
+		if err := lockInfo.Lock.Unlock(lockInfo.Ctx); err != nil {
+			errs = append(errs, errors.EnsureStack(err))
```
brycemcanally, line 212: This looks like a good use for multierr.

src/internal/consistenthashing/etcd.go
FahadBSyed: I'm not sure. If it is, then two rings could lock the same key:
```go
ring1.Lock(ctx, "repo1")
ring2.Lock(ctx, "repo2")
``` 
where the lock creation would look like this
```
nodeLock := dlock.NewDLock(ring.client, ring1.Prefix + "/" + key)
nodeLock := dlock.NewDLock(ring.client, ring2.Prefix + "/" + key)
```
I don't think we want to be able to do that right?

src/internal/consistenthashing/etcd.go
FahadBSyed: good catch!

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,291 @@
...
+	ring.logger.WithFields(ring.memberLogFields()).Infof("deleting member %s from ring", member.Id)
+	ring.members = append(ring.members[:index], ring.members[index+1:]...)
+}
+
+// etcdToMember converts an etcd key in a ring's collection to a node ID.
+func (ring *Ring) etcdToMember(key string) string {
+	return path.Base(key)
+}
+
+// investigate tagged logger.
```
brycemcanally, line 276: Probably should put a `TODO` here.

src/internal/consistenthashing/etcd.go
FahadBSyed: If we have a ring with 1 member and blocking lock() calls, then the watch() function fails, the blocking lock() calls could attempt to access the members before checking the context and fail

src/internal/consistenthashing/etcd.go
FahadBSyed: Actually I ended up removing this comment

src/server/pfs/server/master.go
FahadBSyed: Gotcha, will move this back and re-test

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,331 @@
...
+	hashFn = crc32.ChecksumIEEE
+	idGen  = uuid.New
+)
+
+// Ring is a consistent hash Ring. Each process should only create one instance with a given prefix. Nodes can be
+// added locally to a Ring. A Ring watches for changes to the prefix to determine if new nodes have been added
+// or deleted.
+type Ring struct {
+	client    *etcd.Client
+	hashFn    func([]byte) uint32
```
brycemcanally, line 41: This field isn't needed any more if we are using the package variable.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,331 @@
...
+// node is a local member of a Ring. Upon creation, a node requests a lease from etcd and continuously refreshes it
+// until it is shutdown.
+type node struct {
+	member
+	locks map[string]lockInfo
+}
+
+// member contains metadata about members in a Ring.
+type member struct {
+	Id   string
```
brycemcanally, line 58: unexport

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,331 @@
...
+
+// lockInfo instances are created for keys by a ring's node.
+// A context lockInfo.ctx is stored so the proper context is used when unlocking lockInfo.lock
+type lockInfo struct {
+	lock dlock.DLock
+	ctx  context.Context
+}
+
+func (ring *Ring) Members() []member {
+	return ring.members
```
brycemcanally, line 70: state lock

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+func WithRing(ctx context.Context, client *etcd.Client, logger *logrus.Logger, prefix string, cb func(ctx context.Context, ring *Ring) error) error {
+	cancelCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	eg, ctx := errgroup.WithContext(cancelCtx)
+	id := idGen()
+	ring := ring(client, logger, prefix, id)
+	etcdCol := collection.NewEtcdCollection(ring.client, path.Join(ring.prefix, "nodes"), nil, nil, nil, nil)
+	eg.Go(func() error { return ring.watch(ctx, cancel, etcdCol) })
```
brycemcanally, line 80: The cancels should be removed from the two background goroutines. If they return an error, the errgroup context will be canceled. The cancel context we create is only for indicating successful completion of the client provided callback.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,331 @@
...
+			member: localMember,
+			locks:  map[string]lockInfo{},
+		},
+		prefix: prefix,
+	}
+}
+
+// createNode creates a lease, inserts itself as a key to etcd, keeps the lease alive in the background.
+func (ring *Ring) createNode(ctx context.Context, cancel context.CancelFunc, col collection.EtcdCollection, id string) error {
+	if err := col.Claim(ctx, id, &types.BoolValue{Value: true},
```
brycemcanally, line 118: Can just return here.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,331 @@
...
+		func(ctx context.Context) error {
+			// keep the lease alive until the context is canceled.
+			<-ctx.Done()
+			return errors.EnsureStack(ctx.Err())
```
brycemcanally, line 122: You could just return nil here and remove the check below for context canceled.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,331 @@
...
+			// Need to release locks that no longer associate to the ring's node.
+			if err := ring.rebalance(); err != nil {
+				ring.logger.WithField("ring", ring.prefix).Errorf("failed rebalancing: %v", err)
+				return err
+			}
+		}
+		return nil
+	}); err != nil {
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			ring.logger.WithFields(ring.memberLogFields()).Infof("shutting down ring")
```
brycemcanally, line 157: This should probably just be in a defer in the with ring function. Also, a setup log may be nice as well.

src/internal/consistenthashing/etcd.go
```diff
@@ -0,0 +1,331 @@
...
+// keyWithRingPrefix prepends the ring prefix to a key to prevent key collisions with other rings.
+func (ring *Ring) keyWithRingPrefix(key string) string {
+	return path.Join(ring.prefix, key)
+}
+
+// rebalance iterates through held locks and releases locks that no longer associate to a ring's node.
+func (ring *Ring) rebalance() error {
+	var errs error
+	for key, lockInfo := range ring.node.locks {
+		if ring.get(key).Id == ring.node.Id || lockInfo.lock == nil {
```
brycemcanally, line 181: When would the lock be nil if there is a lock info?

src/server/pfs/server/master.go
```diff
@@ -69,42 +87,67 @@ func (d *driver) master(ctx context.Context) {
...
+					backoff.RetryUntilCancel(ctx, func() error { //nolint:errcheck
+						ctx, cancel := context.WithCancel(ctx)
+						defer cancel()
+						lockCtx, err := ring.Lock(ctx, lockPrefix)
+						if err != nil {
+							return errors.Wrap(err, "error locking repo lock")
+						}
+						defer func() {
+							if err := ring.Unlock(lockPrefix); err != nil {
+								if retErr == nil {
```
brycemcanally, line 133: I just realized this can be simplified further to just `retErr = multierror.Append...`. It is fine if `retErr` is nil because append will only capture the unlock error in that case.

src/internal/consistenthashing/etcd_test.go
```diff
@@ -0,0 +1,294 @@
...
+	ctx            context.Context
+	cancel         context.CancelFunc
+	eg             *errgroup.Group
+	locks          int
+	keys           *sync.Map
+	workers        int
+	workerIds      map[int]string
+	workersReady   chan bool
+	ringConfig     testRingConfig
+	beginLocking   chan bool
```
brycemcanally, line 40: It is pretty common to use empty structs rather than booleans for a use case like this.

src/internal/consistenthashing/etcd_test.go
```diff
@@ -0,0 +1,294 @@
...
+		locksPerWorker[node]++
+
+		return true // return true acts as continue in a sync.Map.Range()
+	})
+	return locksPerWorker
+}
+
+func (test *lockTestConfig) runWorker(ctx context.Context, t *testing.T, id string) {
+	collection.DefaultTTL = 1
+	idGen = func() string { return id }
```
brycemcanally, line 251: This seems like an issue since this function is running in a goroutine. A different value could be set for idGen before it is used in `WithRing`.

src/internal/consistenthashing/etcd.go
FahadBSyed: We want this exported so we can return information about Members()

src/internal/consistenthashing/etcd_test.go
FahadBSyed: Any ideas on how I can do a dependency injection of an Id for testing?

src/internal/consistenthashing/etcd.go
FahadBSyed: this is a holdover from before when we were setting the lockInfo ahead of time and storing a cancel, this is not needed anymore.

brycemcanally (approved):
lgtm

PR #8050, opened 2022-08-10
Fix service pipelines not picking up new jobs.

PFedak:
A context wasn't respected after a refactor, and finish job would fail due to the absence of a meta repo. 

Also expand the existing test to hit the job switching logic, and skip `FINISHING` jobs like we do in normal transforms (maybe the subscribe logic should skip those automatically).

### DIFF
--- a/src/server/worker/pipeline/service/service.go
+++ b/src/server/worker/pipeline/service/service.go
@@ -54,6 +54,8 @@ func Run(driver driver.Driver, logger logs.TaggedLogger) error {
 				retErr = ppsutil.FinishJob(pachClient, jobInfo, pps.JobState_JOB_FINISHING, "")
 			}
 		}()
+		// now that we're actually running the datum, use a pachClient which is bound to the job-scoped context
+		pachClient := pachClient.WithCtx(ctx)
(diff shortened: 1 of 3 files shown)


src/server/worker/pipeline/service/service.go
```diff
@@ -85,6 +87,9 @@ func forEachJob(pachClient *client.APIClient, pipelineInfo *pps.PipelineInfo, lo
 	var cancel func()
 	var eg *errgroup.Group
 	return pachClient.SubscribeJob(pipelineInfo.Pipeline.Name, true, func(ji *pps.JobInfo) error {
+		if ji.State == pps.JobState_JOB_FINISHING {
```
acohen4, line 90: does this usually get hit if we just finished processing a job?

PFedak: I don't think it should generally, as we'd pick up the new job event before the finishing happens. But we do the same in the normal transform worker, and I felt like it was better to be defensive here, since there's nothing we'd want to do on a FINISHING job, anyway.

acohen4 (approved):
LGTM!

PR #8052, opened 2022-08-11
[Mount Server] Mount datums

smalyala:
Adds functionality to the mount server to take an input spec and mount the resulting datums.

Changes:
- Add endpoints for mounting datums, switching to a particular datum, and getting info back on the datums
- Update existing `/_mount` and `/_unmount` endpoints, request body, and response body
- Update response for `/repos` and `/mounts`
- Modify logic in `loopback.go` to allow only a specified list of files to be mounted

Jira: https://pachyderm.atlassian.net/browse/INT-663


lukemarsden (approved):
Looks great, thanks!

The API change looks backwardly incompatible: should we either add back in compatibility endpoints, otherwise carefully coordinate releasing a version of the frontend (jupyter plugin) that speaks the new API? Can I leave that with you to make sure we don't end up in a broken state?

smalyala:
Awesome, thanks for the review! Yes, I'll merge this PR when I merge the PR with the corresponding jupyter plugin changes.

PR #8055, opened 2022-08-11
[INT-614] Refactor to use serviceenv.GetPachClient

msteffen:
This is the 2.x version of #7633


bbonenfant (approved):

msteffen:
I got confused. This change is in 2.3.x, just not 2.2.x. We need this change for a user on 2.2.3, but I now see that, ideally, the user would just upgrade to 2.3

PR #8065, opened 2022-08-16
Adds Proxy.host 

tybritten:
For users leveraging the proxy only, they can configure `proxy.host` with the externally reachable name of the proxy.

Configuration combinations work in this order:
1. If ingress is enabled, an ingress hostname is required. This is the external address, whether the proxy is enabled or not. This matches the current behavior.
2. If the proxy is enabled but ingress isn't, `proxy.host` is used to set the external address. If the host isn't set, it still enabled the proxy but falls back to the localhost ports as it does today.
3. if neither are enabled, the current behavior of falling back to localhost ports and port-forward still applies.

This allows users to orderly transition from ingress to proxy while the proxy is optional and then prevents any breaking changes to proxy users when ingress is depricated.

### DIFF
--- a/etc/helm/pachyderm/templates/_helpers.tpl
+++ b/etc/helm/pachyderm/templates/_helpers.tpl
@@ -36,13 +36,23 @@ imagePullSecrets:
 {{- end }}
 {{- end -}}
 
-{{- define "pachyderm.ingressproto" -}}
+{{- define "pachyderm.hostproto" -}}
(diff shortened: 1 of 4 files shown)


jrockway (approved):
Not sure what is up with the EXAMPLES test.  I don't think it even ran against this code, but rather a previous version of this branch.

acohen4 (approved):
LGTM, thank you!

tybritten:
@lbliii this is going to require an update to the 2.3.x documentation on proxy- changing `ingress.host` to `proxy.host`

PR #8066, opened 2022-08-16
[core-975] Remove our fork of os/exec

seslattery:
In https://github.com/pachyderm/pachyderm/pull/2542  we forked the os/exec package from the go standard library.  However, we haven't updated that library since, we're now ~50+ commits behind.  After reading through the rationale given, and a handful of proposals on the go standard library on how to fix the given rationale, I don't think we were hitting the problem identified.  The documentation has significantly improved here over the years, and I am relatively confident that we can just leverage `cmd.Wait()` now directly.  My understanding is that we split out cmd.Wait() into two function calls, so that we could wait for the I/O loop to finish processing in the event of a context.Cancel().  However, this is the behavior of `cmd.Wait()`.  

```
cmd.Wait() documentation:

Wait waits for the command to exit and waits for any copying to stdin or copying from stdout or stderr to complete.

The command must have been started by Start.

The returned error is nil if the command runs, has no problems copying stdin, stdout, and stderr, and exits with a zero exit status.

If the command fails to run or doesn't complete successfully, the error is of type *ExitError. Other error types may be returned for I/O problems.

If any of c.Stdin, c.Stdout or c.Stderr are not an *os.File, Wait also waits for the respective I/O loop copying to or from the process to complete.

Wait releases any resources associated with the Cmd.
```

I also remove the utility function common.IsDone(), it wasn't idiomatic and I think it obfuscates more than it assists.

I am skipping 1 test that is particularly flaky for now, will create a JIRA ticket to de-flake it.


src/server/pachyderm_test.go
```diff
@@ -4538,6 +4538,9 @@ func TestDatumStatusRestart(t *testing.T) {
 		t.Skip("Skipping integration tests in short mode")
 	}
 
+	// TODO: Fix flaky test
```
jrockway, line 4541: If you do `TODO(CORE-XXX)` that would be helpful for the future reader.  Could also go in the message and save people looking at CI results a trip to the code, actually.

src/server/worker/driver/driver.go
```diff
@@ -361,24 +361,16 @@ func (d *driver) RunUserCode(
...
 	}
-	// A context with a deadline will successfully cancel/kill
-	// the running process (minus zombies)
-	state, err := cmd.Process.Wait()
-	if err != nil {
-		return errors.EnsureStack(err)
-	}
-	if common.IsDone(ctx) {
-		if err = ctx.Err(); err != nil {
+	select {
```
jrockway, line 364: I feel like Wait should respect the context and not require a quick check here to see if it's done.  What if it's done right after the chan recv fails and is done right before Wait runs?  Seems pointless to check.

src/server/worker/driver/driver.go
```diff
@@ -429,23 +421,16 @@ func (d *driver) RunUserErrorHandlingCode(
...
 	}
-	// A context w a deadline will successfully cancel/kill
-	// the running process (minus zombies)
-	state, err := cmd.Process.Wait()
-	if err != nil {
-		return errors.EnsureStack(err)
-	}
-	if common.IsDone(ctx) {
-		if err = ctx.Err(); err != nil {
+	select {
```
jrockway, line 424: Same story.  I don't know why we do this.  The "minus zombies" comment is interesting, we need to do more testing around that (and properly act as PID 1).

jrockway (approved):
I'd like to see two bugs filed; one to unskip this test, one to add a test that ensures that we clean up the datum even if we fork and exec something else.  (A place where I've seen zombies potentially not being reaped is when doing `go tool pprof -http :$RANDOM_PORT & curl http://localhost:$RANDOM_PORT`, btw.  In my pipeline that does this I wait for the pid of `go tool pprof` but it's an easy mistake to make and we should be able to shutdown the pod when a worker does this.)

seslattery:
> I'd like to see two bugs filed; one to unskip this test, one to add a test that ensures that we clean up the datum even if we fork and exec something else. (A place where I've seen zombies potentially not being reaped is when doing `go tool pprof -http :$RANDOM_PORT & curl http://localhost:$RANDOM_PORT`, btw. In my pipeline that does this I wait for the pid of `go tool pprof` but it's an easy mistake to make and we should be able to shutdown the pod when a worker does this.)

CORE-1015 and CORE-1016 have been created

PR #8069, opened 2022-08-17
use localhostIssuer=true even when network routing allows it to be false

jrockway:
This is due to the readiness check interfering with the ability to contact local pachd via k8s services.

CORE-968


acohen4 (approved):
LGTM

PR #8077, opened 2022-08-17
fix more spots using older go versions

seslattery:


robert-uhl:
You could run `go mod tidy` if you want, but otherwise LGTM.

robert-uhl (approved):

PR #8078, opened 2022-08-18
Add Projects as a Pachyderm resource [CORE-90]

acohen4:


src/server/cmd/pachctl/shell/completions.go
```diff
@@ -131,6 +132,22 @@ func BranchCompletion(flag, text string, maxCompletions int64) ([]prompt.Suggest
 	return result, samePart(part)
 }
 
+func ProjectCompletion(flag, text string, maxCompletions int64) ([]prompt.Suggest, CacheFunc) {
+	c := getPachClient()
+	resp, err := c.ListProject(context.Background(), &pfs.ListProjectRequest{})
```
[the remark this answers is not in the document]
acohen4: yeah this should in fact be c.Ctx() 👍

src/internal/pfsdb/pfsdb.go
```diff
@@ -175,6 +176,33 @@ func Branches(db *pachsql.DB, listener col.PostgresListener) col.PostgresCollect
...
+	return project.Name
+}
+
+func Projects(db *pachsql.DB, listener col.PostgresListener) col.PostgresCollection {
+	return col.NewPostgresCollection(
+		projectsCollectionName,
+		db,
+		listener,
+		&pfs.ProjectInfo{},
+		nil,
```
robert-uhl, line 189: I do mildly wonder if it is worthwhile having a key check here, but I don’t feel it strongly.

src/pfs/pfs.proto
```diff
@@ -21,6 +21,7 @@ message Repo {
   option (gogoproto.goproto_stringer) = false;
   string name = 1;
   string type = 2;
+  Project project = 3;
```
robert-uhl, line 24: Should this be part of this PR, or one implementing CORE-173?

src/pps/pps.proto
```diff
@@ -313,6 +313,7 @@ message Worker {
 
 message Pipeline {
   string name = 1;
+  string project = 2;
```
robert-uhl, line 316: If this goes in as part of this PR, should it be a `Project` rather than a raw string?

acohen4: Yeah I agree :)

src/internal/pfsdb/pfsdb.go
acohen4: Unlike the other keys that have some structure like <repo>.<type>, and <repo@branch>, project is just <project>, which suggests that we don't need to enforce it's structure, besides having the user input sanitized

src/pfs/pfs.proto
acohen4: i think it could be in either. I threw it in here to try to list out the general proto structure / relationships

src/client/pfs.go
```diff
@@ -367,6 +367,21 @@ func (c APIClient) DeleteBranch(repoName string, branchName string, force bool)
 	return grpcutil.ScrubGRPC(err)
 }
 
+// ListRepoByType returns info about Repos of the given type
```
brycemcanally, line 370: This comment needs to be updated.

src/client/pfs.go
```diff
@@ -367,6 +367,21 @@ func (c APIClient) DeleteBranch(repoName string, branchName string, force bool)
...
+// The if repoType is empty, all Repos will be included
+func (c APIClient) ListProject() (_ []*pfs.ProjectInfo, retErr error) {
+	ctx, cf := context.WithCancel(c.Ctx())
+	defer cf()
+	client, err := c.PfsAPIClient.ListProject(
+		ctx,
+		&pfs.ListProjectRequest{},
+	)
+	if err != nil {
+		return nil, grpcutil.ScrubGRPC(err)
```
brycemcanally, line 380: The scrub should probably be in a defer. The clientsdk function could return an unscrubbed error.

src/pfs/pfs.proto
```diff
@@ -284,6 +294,31 @@ message DeleteBranchRequest {
...
+  bool update = 3;
+}
+
+message CreateProjectResponse {}
+
+message InspectProjectRequest {
+  Project project = 1;
+}
+
+message InspectProjectResponse {
```
brycemcanally, line 309: This seems like it should be consistent with the other inspect rpcs which just return the Pachyderm object's info.

src/pfs/pfs.proto
```diff
@@ -284,6 +294,31 @@ message DeleteBranchRequest {
...
+}
+
+message ListProjectRequest {}
+
+message DeleteProjectRequest {
+  Project project = 1;
+  bool force = 2;
+}
+
+message DeleteProjectResponse {}
```
brycemcanally, line 320: Similar to above, delete calls don't normally have a response type. There could potentially be value in this, but I would favor consistency with the other rpcs personally.

src/pfs/pfs.proto
```diff
@@ -284,6 +294,31 @@ message DeleteBranchRequest {
   bool force = 2;
 }
 
+message CreateProjectRequest {
+  Project project = 1;
+  string description = 2;
+  bool update = 3;
+}
+
+message CreateProjectResponse {}
```
brycemcanally, line 303: This is inconsistent with other rpcs as well.

src/server/pfs/cmds/cmds.go
```diff
@@ -978,6 +979,150 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
 	shell.RegisterCompletionFunc(deleteBranch, shell.BranchCompletion)
 	commands = append(commands, cmdutil.CreateAliases(deleteBranch, "delete branch", branches))
 
+	projectDocs := &cobra.Command{
+		Short: "Docs for projects.",
+		Long: `Projects, are the top level organizational objects in Pachyderm.
```
brycemcanally, line 984: That comma seems unnecessary.

src/server/pfs/cmds/cmds.go
```diff
@@ -978,6 +979,150 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
...
+				&pfs.CreateProjectRequest{
+					Project:     &pfs.Project{Name: args[0]},
+					Description: description,
+					Update:      true,
+				})
+			return grpcutil.ScrubGRPC(err)
+
+		}),
+	}
+	updateProject.Flags().StringVarP(&description, "description", "d", "", "The description of the update project.")
```
brycemcanally, line 1034: `update` -> `updated`

src/server/pfs/cmds/cmds.go
```diff
@@ -978,6 +979,150 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
...
+	shell.RegisterCompletionFunc(updateProject, shell.ProjectCompletion)
+	commands = append(commands, cmdutil.CreateAliases(updateProject, "update project", projects))
+
+	inspectProject := &cobra.Command{
+		Use:   "{{alias}} <project>",
+		Short: "Create a new project.",
```
brycemcanally, line 1040: Wrong text.

src/server/pfs/cmds/cmds.go
```diff
@@ -978,6 +979,150 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
...
+			}
+			return writer.Flush()
+		}),
+	}
+	listProject.Flags().AddFlagSet(outputFlags)
+	commands = append(commands, cmdutil.CreateAliases(listProject, "list project", projects))
+
+	deleteProject := &cobra.Command{
+		Use:   "{{alias}} <project>",
+		Short: "Create a new project.",
```
brycemcanally, line 1105: Wrong text.

src/server/pfs/pretty/pretty.go
```diff
@@ -121,6 +123,17 @@ func PrintBranch(w io.Writer, branchInfo *pfs.BranchInfo) {
 	fmt.Fprintln(w)
 }
 
+// PrintProject pretty-prints a project.
+func PrintProject(w io.Writer, projectInfo *pfs.ProjectInfo) {
```
brycemcanally, line 127: `PrintProject` -> `PrintProjectInfo`.

src/server/pfs/server/driver.go
```diff
@@ -421,6 +424,57 @@ func (d *driver) deleteRepo(txnCtx *txncontext.TransactionContext, repo *pfs.Rep
 	return nil
 }
 
+func (d *driver) createProject(ctx context.Context, req *pfs.CreateProjectRequest) error {
+	if err := ancestry.ValidateName(req.Project.Name); err != nil {
+		return errors.Wrapf(err, "invalid project name")
+	}
+	if err := d.env.TxnEnv.WithWriteContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
```
brycemcanally, line 431: You could just return here.

src/server/pfs/server/driver.go
```diff
@@ -421,6 +424,57 @@ func (d *driver) deleteRepo(txnCtx *txncontext.TransactionContext, repo *pfs.Rep
...
+		projects := d.projects.ReadWrite(txnCtx.SqlTx)
+		if req.Update {
+			if err := projects.Get(pfsdb.ProjectKey(req.Project), &pfs.ProjectInfo{}); err != nil {
```
brycemcanally, line 434: Collections have an update function that will return a not exists error if it doesn't exist.

src/server/pfs/server/driver.go
```diff
@@ -421,6 +424,57 @@ func (d *driver) deleteRepo(txnCtx *txncontext.TransactionContext, repo *pfs.Rep
...
+	if err := d.projects.ReadOnly(ctx).Get(pfsdb.ProjectKey(project), pi); err != nil {
+		return nil, errors.EnsureStack(err)
+	}
+	return pi, nil
+}
+
+func (d *driver) listProject(ctx context.Context, cb func(*pfs.ProjectInfo) error) error {
+	projectInfo := &pfs.ProjectInfo{}
+	if err := d.projects.ReadOnly(ctx).List(projectInfo, col.DefaultOptions(), func(string) error {
+		return cb(proto.Clone(projectInfo).(*pfs.ProjectInfo))
```
brycemcanally, line 463: Is the clone necessary?

src/server/pfs/server/driver.go
```diff
@@ -421,6 +424,57 @@ func (d *driver) deleteRepo(txnCtx *txncontext.TransactionContext, repo *pfs.Rep
...
+	pi := &pfs.ProjectInfo{}
+	if err := d.projects.ReadOnly(ctx).Get(pfsdb.ProjectKey(project), pi); err != nil {
+		return nil, errors.EnsureStack(err)
+	}
+	return pi, nil
+}
+
+func (d *driver) listProject(ctx context.Context, cb func(*pfs.ProjectInfo) error) error {
+	projectInfo := &pfs.ProjectInfo{}
+	if err := d.projects.ReadOnly(ctx).List(projectInfo, col.DefaultOptions(), func(string) error {
```
brycemcanally, line 462: Can just return here.

src/pfs/pfs.proto
```diff
@@ -284,6 +294,31 @@ message DeleteBranchRequest {
   bool force = 2;
 }
 
+message CreateProjectRequest {
+  Project project = 1;
```
robert-uhl, line 298: Should this be a Project or a ProjectInfo?

src/pfs/pfs.proto
```diff
@@ -284,6 +294,31 @@ message DeleteBranchRequest {
...
+  string description = 2;
+  bool update = 3;
```
robert-uhl, line 300: What are the semantics of `update` here? Is updating a project name allowed? Are all repos in the project updated? Ditto all pipelines and pipeline inputs?

src/pfs/pfs.proto
```diff
@@ -284,6 +294,31 @@ message DeleteBranchRequest {
...
+
+message InspectProjectResponse {
+  ProjectInfo project_info = 1;
+}
+
+message ListProjectRequest {}
+
+message DeleteProjectRequest {
+  Project project = 1;
+  bool force = 2;
```
robert-uhl, line 317: If `force` is true, should deleting a project which does not exist succeed or fail?

src/pfs/pfs.proto
acohen4: to be consistent with other Request types this should be Project

src/pfs/pfs.proto
acohen4: I think you should have to pass `force` to delete a project's internal objects. Not passing force would only delete empty projects. WDYT?

src/server/pfs/server/driver.go
acohen4: Right now it is because we collect projects into a list during DeleteAll. List Repo works the same way. The client could also be responsible for copying but I think this is nice + consistent

src/pfs/pfs.proto
robert-uhl: That makes sense to me — I assume by internal objects you mean repos & pipelines, right?

How much of this behaviour should be connected up in this PR vice a follow-up?

acohen4: given that repos and pipelines don't know where they belong yet, that should be done in a follow up. My intention here was to describe the API surface

robert-uhl (approved):

src/server/pfs/server/api_server.go
```diff
@@ -348,6 +348,38 @@ func (a *apiServer) DeleteBranch(ctx context.Context, request *pfs.DeleteBranchR
...
+func (a *apiServer) CreateProject(ctx context.Context, request *pfs.CreateProjectRequest) (response *types.Empty, retErr error) {
+	if err := a.driver.createProject(ctx, request); err != nil {
+		return nil, err
+	}
+	return &types.Empty{}, nil
+}
+
+// InspectProject implements the protobuf pfs.InspectProject RPC
+func (a *apiServer) InspectProject(ctx context.Context, request *pfs.InspectProjectRequest) (response *pfs.ProjectInfo, retErr error) {
+	pi, err := a.driver.inspectProject(ctx, request.Project)
```
brycemcanally, line 361: you could just return this.

src/server/pfs/server/api_server.go
```diff
@@ -348,6 +348,38 @@ func (a *apiServer) DeleteBranch(ctx context.Context, request *pfs.DeleteBranchR
 	return &types.Empty{}, nil
 }
 
+// CreateProject implements the protobuf pfs.CreateProject RPC
+func (a *apiServer) CreateProject(ctx context.Context, request *pfs.CreateProjectRequest) (response *types.Empty, retErr error) {
```
brycemcanally, line 352: The named returns aren't needed for any of these methods.

src/server/pfs/server/driver.go
brycemcanally: That should probably happen in DeleteAll then. It seems like the cloning should only happen where it is needed. Although I don't think the performance will be a big deal since the number of projects will probably never be high enough for that to matter, a separate use case for list project would incur that overhead even if it doesn't need it.

brycemcanally (approved):
lgtm, just some nitpicky things left.

PR #8083, opened 2022-08-18
cronspec details

lbliii:
- restructured content
- added full list of "nonstandard" cron inputs (eg @hourly)
- added callouts
- added SQL ingest + jsonnet example of cronspec config

### DIFF
--- a/doc/docs/2.3.x/concepts/pipeline-concepts/pipeline/cron.md
+++ b/doc/docs/2.3.x/concepts/pipeline-concepts/pipeline/cron.md
@@ -1,69 +1,79 @@
 # Cron Pipeline
 
-Pachyderm triggers pipelines when new changes appear in the input repository.
-However, if you want to trigger a pipeline based on time instead of upon
-arrival of input data, you can schedule such pipelines to run periodically
-by using the Pachyderm built-in cron input type.
+A Cron pipeline is triggered by a set time interval *instead* of whenever new changes appear in the input repository.
 
-A standard pipeline with a PFS input might not satisfy
-the requirements of the following tasks:
+## About Cron Pipelines
 
-- Scrape websites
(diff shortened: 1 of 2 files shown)


doc/docs/2.3.x/concepts/pipeline-concepts/pipeline/cron.md
```diff
@@ -1,69 +1,79 @@
...
-    ```json
-      "input": {
-        "cron": {
-          "name": "tick",
-          "spec": "@every 10s",
-          "overwrite": true
-        }
-      }
-    ```
+| `"spec"`   | The interval between scheduled cron jobs; accepts [RFC 3339](https://www.ietf.org/rfc/rfc3339.txt) inputs as well as `never`, `@hourly`, `@daily`, `@weekly`, `@monthly`, `@yearly`, `@reboot`, and `@every ...`|
```
BOsterbuhr, line 30: What does `@reboot` do? The others are more obvious, but maybe we should add a little note on `@reboot` and also what can be used with `@every`.
For example, if I want it to be every 3 months, what would that be? `@every 3m` would be minutes, not months, right?

BOsterbuhr (changes requested):
Good work; things are clean and concise. However, I have one request regarding what can be used for `spec`.

doc/docs/2.3.x/concepts/pipeline-concepts/pipeline/cron.md
lbliii: Found some direct links to those supported formats in the cron.go pkg!

BOsterbuhr (approved):
LGTM

maguenza (approved):
LGTM - will share on the FREQQ ticket as well.

PR #8089, opened 2022-08-19
[FREQQ-143] [2.3.x] Remove Loki storageClassName default

BOsterbuhr:
Having `standard` as the default for `loki-stack.loki.persistence.storageClassName` would cause Loki to fail when deploying to AWS.
Turns out you don't actually need anything set for `storageClassName`; Kubernetes will assign the right `storageClass` itself.


tybritten (approved):

BOsterbuhr:
@jrockway good to merge?

PR #8090, opened 2022-08-19
[CORE-563] Update deleteExpiredTokens to Run on Startup

FahadBSyed:


src/server/auth/server/api_server.go
```diff
@@ -46,7 +46,7 @@ const (
 	clusterRoleBindingKey = "CLUSTER:"
 
 	// the length of interval between expired auth token cleanups
-	cleanupIntervalHours = 24
+	cleanupIntervalHours = 1
```
acohen4, line 49: If the problem is that we're concerned that pachd may never garbage collect tokens, maybe the right solution would be to run the deleteExpiredTokens immediately instead of waiting for the cleanupInterval to elapse. In that case I think the 24 hour interval would still be fine. Not married to 24 hours, but also not clear to me that shortening the interval to 1 hour makes sense

PR #8092, opened 2022-08-19
upgrade loki-stack

jrockway:
This fixes CORE-925, because the upstream chart deletes the faulty liveness checks.  I guess they had the same problem and just said "fuck this".  I added a more generous liveness check, tested against invalid and valid configurations, and the results are pretty ok.


tybritten (approved):
LGTM

PR #8096, opened 2022-08-19
Create aliases on DAG components.

PFedak:
Previous alias commit logic would frequently result in PFS states that
violated the key provenance invariant, which could lead to surprising
behavior. This change creates alias commits on entire connected
components of the provenance graph, rather than just on those branches
upstream of changes.

This results in more alias commits being created, which affects the
output of list commit <ID> and can lead to running a job after
restarting a pipeline in cases where that would not have happened
before.

Note that rewinding to states before this change is not prevented and
can still cause problems.


msteffen:
Closing this pr, as it's obsolete (after discussing with @acohen4).

PR #8098, opened 2022-08-19
Snowflake Read and Write connectors

albscui:
This PR introduces the idea of `Connectors`, as well as an implementation of Read and Write connectors for Snowflake.

Added:
- `src/integrations/connectors` directory for organizing various connectors
- A `Dockerfile` for building a Snowflake image containing `snowsql`, which is the entrypoint for interacting with Snowflake.
- Jsonnet files for both Read and Write pipelines
- An end to end integration test for Snowflake
- Run this integration test in CI

Changed:
- refactored our internal `sdata` package's test utilities into its own subpackage.

Outstanding questions:
- How should we version and tag the `pachyderm/snowflake` docker image?
- What version should we use in the Jsonnet templates? How would users get updates?
- Is there a better way to test this? Currently we use Snowflake's Go library for programmatically creating and destroying test objects, but we use `snowsql` in the actual docker image. This mismatch means we are two different tools for connecting to Snowflake.


albscui:
We will favour our built-in-house CLI tool over snowsql, because it gives us more flexibility wrt manipulating file systems. https://github.com/pachyderm/pachyderm/pull/8125

PR #8101, opened 2022-08-22
[CORE-800] Update inspectPipeline error message for non-existent pipelines

armaanv:


jrockway (approved):
I guess this is OK, but it will be weird when there is some I/O error instead of it actually not being found.  Probably clear from the error message though.

PR #8108, opened 2022-08-23
[CORE-989} create default project

robert-uhl:


src/pfs/pfs.proto
```diff
@@ -54,6 +54,7 @@ message RepoInfo {
     int64 size_bytes = 1;
   }
   Details details = 7;
+  Project project = 8;
```
acohen4, line 57: Since Project is stored inside the Repo Message, it shouldn't also be necessary here

src/pfs/pfs.go
```diff
@@ -18,6 +18,8 @@ const (
 	UserRepoType = "user"
 	MetaRepoType = "meta"
 	SpecRepoType = "spec"
+
+	DefaultProjectName = "default"
```
acohen4, line 22: nit: NBD, but I would call it `DefaultProject`, feeling that the type implies it's the name.

src/internal/clusterstate/v2.4.0.go
```diff
@@ -4,13 +4,26 @@ import (
...
 		return col.SetupPostgresCollections(ctx, env.Tx, pfsdb.CollectionsV2_4_0()...)
+	}).
+	Apply("Add default project", func(ctx context.Context, env migrations.Env) error {
+		var defaultProject = &pfs.ProjectInfo{
+			Project: &pfs.Project{
+				Name: pfs.DefaultProjectName,
+			},
+		}
+		if err := pfsdb.Projects(nil, nil).ReadWrite(env.Tx).Create(pfs.DefaultProjectName, defaultProject); err != nil {
+			return errors.Errorf("could not create default project: %w", err)
```
acohen4, line 24: We usually use `errors.Wrapf(err, message)`

src/server/pfs/server/testing/server_test.go
```diff
@@ -102,6 +102,68 @@ func finfosToPaths(finfos []*pfs.FileInfo) (paths []string) {
...
+		// and it should have taken effect
+		pp, err = env.PachClient.ListProject()
+		require.NoError(t, err)
+		for _, p := range pp {
+			if p.Project.Name == pfs.DefaultProjectName {
+				require.Equal(t, desc, p.Description)
+			}
+		}
+		// deleting the default project is allowed, too
+		require.NoError(t, env.PachClient.DeleteProject(pfs.DefaultProjectName, false))
```
acohen4, line 150: Loosley related to the test, but doesn't necessarily have to happen in this PR - I don't think we should let the user get into a state where they don't have any projects, and I think that it may make sense to prevent the user from deleting the Default project to achieve those means. Probably worth aligning with Product

src/pfs/pfs.go
robert-uhl: I started with `DefaultProject`, to be honest, but as I was using it I realised that it was a little less ergonomic than it could have been, because it is really a name rather than a project.

acohen4 (approved):
LGTM

src/server/pfs/server/testing/server_test.go
robert-uhl: Agreed. Left a note in the PRD for discussion.

PR #8111, opened 2022-08-23
enable nightly releases

molinamelendezj:
This PR accomplishes the following;

1. Triggers a nightly circle cron job every night (6AM UTC) (2AM EST) same time as load tests.
2. nightly tag on master created in format `v2.4.0-nightly.20220823`
4. that tag push will then trigger the existing release pipeline, where test are run and a soak test install is done on pulumi test envs where examples run. 
5. if that succeeds the release the release is rolled out automatically. otherwise if examples fail the release is not rolled out

⚠️ the version of pachyderm actively being developed is defined in the circle config file for tag creation, will need to be bumped every time we cut a minor release.

⚠️ there isn't a way to automatically retry CI flake tests yet, so a nightly release workflow may never be rolled out that night, we simply rerun from failed if flakes are seen in the morning to continue the roll out. 

⚠️ if a new nightly build needs to be run during the day, you can simply rerun the previous night's tag creation workflow


.circleci/main.yml
```diff
@@ -7,6 +7,11 @@ aliases:
       ignore: /.*/
     tags:
       only: /^v\d+\.\d+\.\d+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/
+  - &only-nightly-tags
+    branches:
+      ignore: /.*/
```
jrockway, line 12: Should this explicitly select master?  It would be annoying if someone tagged the 2.3.x branch and we got a 2.4.x nightly (which is hard-coded later).

.circleci/main.yml
```diff
@@ -542,6 +550,45 @@ jobs:
...
+      - run:
+          name: create nightly tag if new commits
+          command: |
+            changes=$( env -i git log --pretty=oneline --since='yesterday' | wc -l )
+            if [ $changes -gt 0 ]; then
+              TIMESTAMP=$(date +'%Y%m%d%T' | sed 's/://g')
+              WORKINGVERSION=<< pipeline.parameters.working-pachyderm-version >>
+              NIGHTLY=${WORKINGVERSION}-nightly.${TIMESTAMP}
+              git tag -f -am "nightly release tag ${NIGHTLY}" ${NIGHTLY}
+              git push -f origin ${NIGHTLY}
```
jrockway, line 574: I don't think we should push here, and especially not force push.  I think it should just be `git push --tags`.  If it already exists, the process should fail.  If code changes, that seems non-ideal as well.  It's certainly not worth overwriting commits with whatever we have here, which is what "git push -f" does.

.circleci/main.yml
```diff
@@ -542,6 +550,45 @@ jobs:
...
+          name: setup pachydermbuildbot git
+          command: |
+            git config --global user.email "buildbot@pachyderm.io"
+            git config --global user.name "pachydermbuildbot"
+      - run:
+          name: create nightly tag if new commits
+          command: |
+            changes=$( env -i git log --pretty=oneline --since='yesterday' | wc -l )
+            if [ $changes -gt 0 ]; then
+              TIMESTAMP=$(date +'%Y%m%d%T' | sed 's/://g')
```
jrockway, line 570: Do we need the timestamp (%T) in here?  If we release once a day, %Y%m%d should be sufficient?  Also, what time zone is the date in?  Does it depend on the worker machine at all?

jrockway (changes requested):
I think I'm OK with the process where one workflow creates a tag and another does the release.  Not ideal, but pretty reusable, so I don't have a problem with it.

You should double-check the tag regex and make sure that it only matches what you're expecting.  We should also not "git push -f" in this process.

.circleci/main.yml
```diff
@@ -7,6 +7,11 @@ aliases:
       ignore: /.*/
     tags:
       only: /^v\d+\.\d+\.\d+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/
+  - &only-nightly-tags
+    branches:
+      ignore: /.*/
+    tags:
+      only: /^v\d+\.\d+\.\d+-nightly+(\.[0-9A-Za-z-]+)*?$/
```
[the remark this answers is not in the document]
molinamelendezj: should we call them betas instead then?

.circleci/main.yml
```diff
@@ -542,6 +550,45 @@ jobs:
...
+          command: |
+            changes=$( env -i git log --pretty=oneline --since='yesterday' | wc -l )
+            if [ $changes -gt 0 ]; then
+              TIMESTAMP=$(date +'%Y%m%d%T' | sed 's/://g')
+              WORKINGVERSION=<< pipeline.parameters.working-pachyderm-version >>
+              NIGHTLY=${WORKINGVERSION}-nightly.${TIMESTAMP}
+              git tag -f -am "nightly release tag ${NIGHTLY}" ${NIGHTLY}
+              git push -f origin ${NIGHTLY}
+            else
+              echo "No new changes, skipping tonight's build."
```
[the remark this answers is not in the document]
molinamelendezj: okay, done. creating a tag every day on master date default to UTC from what I found.

.circleci/main.yml
```diff
@@ -542,6 +550,45 @@ jobs:
...
+            fi
+  auto-approve:
+    resource_class: small
+    docker:
+      - image: cimg/base:stable
+    steps:
+      - run:
+          name: auto sign-off release
```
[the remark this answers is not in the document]
molinamelendezj: no we built everything around the tag creation.

.circleci/main.yml
molinamelendezj: good catch, i have locked it to only master

.circleci/main.yml
molinamelendezj: good idea. using ` git push origin ${NIGHTLY} `now

jrockway (approved):
Looks good.

.circleci/main.yml
```diff
@@ -845,3 +888,76 @@ workflows:
           filters: *only-release-tags
           requires:
             - sign-off
+  nightly-release:
```
chainlink, line 891: Is the only difference between the `nightly-releases` workflow and the regular releases workflow is the `approval` step? If so, does the `filter` section in sign-off mean it will only run on the regular release tags? Alternatively, does adding a `when` to just the approval step work?

.circleci/main.yml
```diff
@@ -562,6 +577,26 @@ jobs:
       - store_artifacts:
           path: /tmp/debug-dump
           destination: debug-dump
+  create-nightly-tag:
```
chainlink, line 580: You might want to consider moving the `create-nightly-tag` job to to config.yml, that way, you can figure out if you just need to run that earlier in the process, and you may not need the downstream variables (`run_ci`, etc)

chainlink (commented):
Couple comments, happy to chat on it

.circleci/main.yml
molinamelendezj: yea thats the only difference. and no i've tried that and it just ignores it. plus even if it worked you would need to modify the requires down stream to conditionally wait on approval or not. so just easier to do a whole other pipeline definition.

chainlink: That's good to go this way, perhaps down the road we can use dynamic config for this.

.circleci/main.yml
molinamelendezj: good call. okay take a look made it simpler. will test on a test repo i have.

PR #8112, opened 2022-08-23
[CORE-824] Disallow configuring logs in text format

FahadBSyed:


etc/helm/pachyderm/templates/NOTES.txt
```diff
@@ -10,6 +10,10 @@ pachctl config set active-context local
 pachctl port-forward
 {{- end -}}
 
+{{- if eq .Values.pachd.logFormat "text" }}
+WARNING: setting log format to text is not supported. Defaulting to json
+{{- end -}}
```
jrockway, line 15: I think I'd warn that it's going away in 2.4 and not remove it in 2.3.  Seems odd for a patch release to make a change this large, even if "text" isn't that great.

jrockway (changes requested):

etc/helm/pachyderm/templates/NOTES.txt
FahadBSyed: Ahh okay, can do!

jrockway (approved):
Very nice!

PR #8115, opened 2022-08-24
Spark compatibility in the S3 gateway

lukemarsden:
These are the changes needed to the S3 Gateway to make it work with Spark, and an [example of how to use it](https://github.com/pachyderm/pachyderm/blob/luke-spark-s3g-fresh/examples/spark/s3gateway/README.md).

* Handle paths with a trailing slash as if they didn't have one.
* Don't error on trying to read the timestamp on new files in open commits in the S3 Gateway. Spark needs to be able to read them.
* Don't fail GetObject for paths that are simultaneously an imaginary directory and a real file. Instead, read the file. This requires dancing around the fact that `InspectFile` doesn't work correctly in this case, but `ListFile[0]` works just fine.

This is important for a customer, so it would be great if we could get it into 2.3.2 please :-)


src/server/pfs/s3/object.go
```diff
@@ -43,14 +42,36 @@ func (c *controller) GetObject(r *http.Request, bucketName, file, version string
...
+	if firstFile == nil {
+		// we never set it, probably zero results
+		return nil, s2.NoSuchKeyError(r)
+	}
+	fileInfo := firstFile
+
+	// the exact object named does not exist, but perhaps is a "directory".
+	// "directories" do not actually exist, and certainly cannot be read.
+	// ("seeker can't seek")
+	if fileInfo.File.Path[1:] != file {
```
albscui, line 68: hmm why is it `[1:]`? I would expect since we trim the suffix, it would be something like `[:len(file)-1]`?

src/server/pfs/s3/object.go
```diff
@@ -43,14 +42,36 @@ func (c *controller) GetObject(r *http.Request, bucketName, file, version string
 		commitID = version
 	}
 
-	fileInfo, err := pc.InspectFile(bucket.Commit, file)
+	// We use listFileResult[0] rather than InspectFile result since InspectFile
+	// on a path that has both a file and a directory in it returns the
+	// directory. However, ListFile will show it as a file, if it exists.
+	var firstFile *pfs.FileInfo
+	err = pc.ListFile(bucket.Commit, file, func(fi *pfs.FileInfo) (retErr error) {
+		if firstFile == nil {
```
albscui, line 50: So just to clarify my understanding, lets say we are in an open commit, and we have a situation like

```
NAME TYPE SIZE 
/0   file 1B   
/0/  dir  4B 
/0/foo file
```

The problem this is trying to solve is when Spark is trying to GetObject on `/0` and we want to return the `/0   file` instead of the dir.

Tho I'm a bit worried about the `/0/foo`, is it always guaranteed that the ListFile iteration always iterats on the `/0` first?

src/server/pfs/s3/object.go
```diff
@@ -43,14 +42,36 @@ func (c *controller) GetObject(r *http.Request, bucketName, file, version string
...
+			firstFile = fi
+		}
+		// TODO: would be nice to stop iterating after the first one
```
albscui, line 53: Haha, yea it really would be nice 😅

albscui (approved):
This all LGTM! Just a couple comments about not losing err values in debug message. My comments about the ListFile vs InspectFile is mainly for my own understanding.

albscui:
Also, is the plan to merge this into master later?

src/server/pfs/s3/object.go
albscui: what do you think about moving this check up to the `pc.ListFile` iteration's call back function?
Something like
```Go
if fi.File.Path[1:] == file {
    firstFile  = fi
}
```

lukemarsden: It's `[1:]` because Path starts with a `/` but file doesn't

src/server/pfs/s3/object.go
```diff
@@ -43,14 +42,36 @@ func (c *controller) GetObject(r *http.Request, bucketName, file, version string
...
+	// "directories" do not actually exist, and certainly cannot be read.
+	// ("seeker can't seek")
+	if fileInfo.File.Path[1:] != file {
+		return nil, s2.NoSuchKeyError(r)
+	}
 
 	modTime, err := types.TimestampFromProto(fileInfo.Committed)
 	if err != nil {
-		return nil, err
+		c.logger.Debugf("Warning: using nil timestamp (file probably in open commit)")
```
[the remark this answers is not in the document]
lukemarsden: done, thanks!

src/server/pfs/s3/object.go
lukemarsden: That's right. In practice I've always seen the `/0` show up first.

src/server/pfs/s3/object.go
lukemarsden: Is there any way to make the callback stop iterating without causing an error to bubble up?

lukemarsden:
> Also, is the plan to merge this into master later?

I'll make a PR into master now.

src/server/pfs/s3/object.go
albscui: You can try returning `errutil.ErrBreak` in your callback. See https://github.com/pachyderm/pachyderm/blob/master/src/client/pfs_file.go#L642

lukemarsden:
Here's the master PR: https://github.com/pachyderm/pachyderm/pull/8129

PR #8116, opened 2022-08-24
[CORE-448] Spread alias commits from output branches to meta branches…

acohen4:
…. (#8095)

* Spread alias commits from output branches to meta branches.

This eliminates a large class of bugs.

* Skip alias meta commits

* Update TestChainedPipelines

* fix TestChainedPipelinesNoDelay

Co-authored-by: acohen4 <alonco124@gmail.com>


molinamelendezj:
@acohen4 adding @armaanv as a reviewer too. I was wondering if we could try to reproduce the original edge case where meta commit doesn't close, and see if this avoids the issue. worth the effort. or have you already verified it?

molinamelendezj (approved):

PR #8117, opened 2022-08-24
fix circle ci nightly load test jobs

molinamelendezj:
Turns out scheduled nightly load test got broken when dynamic config was turned on because  [Scheduled workflows are not supported with Dynamic Configurations / Setup Workflows and, by the end of 2022, scheduled workflows will be phased out to be replaced by scheduled pipelines as the primary method of scheduling work.]( https://support.circleci.com/hc/en-us/articles/360060833331-Support-for-Scheduled-Workflows-in-Dynamic-Configurations-Setup-Workflows)

The schedules are now maintained in circle ci. Here is sample triggers that will need to be set up when this PR is merged.

Screen Shot 2022-08-24 at 6 54 44 PM

❗ This PR fixes this for nightly load tests. AND good news is we can actually use this for the nightly builds instead to avoid the double trigger approach in #8111, because we can store the working version in the trigger and we can bump that without source control changes and potentially toggle off the manual approval step when its a scheduled nightly release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,6 @@ parameters:
     type: boolean
     default: false
 
-
 # our defined job, and its steps
 jobs:
   setup:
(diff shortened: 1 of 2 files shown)


jrockway (approved):
Very nice!

PR #8120, opened 2022-08-25
[CORE-612] Add integration test for PFS

FahadBSyed:


src/server/pachyderm_test.go
```diff
@@ -10344,6 +10344,30 @@ func TestPipelineAncestry(t *testing.T) {
...
+	numTestRepos := 50
+	eg, _ := errgroup.WithContext(context.Background())
+	for i := 0; i < numTestRepos; i++ {
+		eg.Go(func() error {
+			require.NoErrorWithinT(t, time.Second*60, func() error {
+				repo := tu.UniqueString(t.Name())
+				require.NoError(t, c.CreateRepo(repo))
+				commit, err := c.StartCommit(repo, "master")
+				require.NoError(t, err)
+				require.NoError(t, c.PutFile(commit, "file", strings.NewReader("foo"), client.WithAppendPutFile()))
```
brycemcanally, line 10362: You can use a one-off put file for this. Just a put file without start / finish commit will automatically start and finish a commit for the put file operation. Also, you need to wait for the commit to be finished because finish commit only marks it as finishing. The PFS master needs to pick up the commit and finish it for it to transition to the finished state.

src/internal/minikubetestenv/client_factory.go
```diff
@@ -223,3 +231,38 @@ func AcquireCluster(t testing.TB, opts ...Option) (*client.APIClient, string) {
 	deleteAll(t, mc.client)
 	return mc.client, assigned
 }
+
+// set the coredns ttl to 0 to cut down on waiting time when tearing down and creating new pachd instances.
+func coreDNSConfigMap(t testing.TB) {
```
acohen4, line 236: I think this belongs in deploy.go. We try containing the k8s / infrastructure specific details there if possible

PR #8123, opened 2022-08-26
[CORE-1013] Ensure we collect errors from task chain goroutines

brycemcanally:
Errors that occur while a create task call is blocked can be lost since the errgroup context will be canceled. This PR ensures we 
collect the error in this case by creating a multierror with the error collected in the errgroup.

This PR also adds unit tests for the task chain and fixes an issue with incrementing the job restart counter.


src/internal/storage/chunk/task_chain_test.go
```diff
@@ -0,0 +1,83 @@
...
+	t.Run("Basic", func(t *testing.T) {
+		t.Parallel()
+		numTasks := 5
+		tc := NewTaskChain(ctx, semaphore.NewWeighted(int64(numTasks)))
+		doneParallel := make(chan int, numTasks)
+		doneSerial := make(chan int, numTasks)
+		for i := 0; i < numTasks; i++ {
+			i := i
+			require.NoError(t, tc.CreateTask(func(_ context.Context) (func() error, error) {
+				time.Sleep(time.Duration(numTasks-i-1) * time.Second)
```
acohen4, line 26: maybe make the sleep random?

src/internal/storage/chunk/task_chain_test.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	})
+	t.Run("Error", func(t *testing.T) {
+		t.Parallel()
+		tc := NewTaskChain(ctx, semaphore.NewWeighted(1))
+		errMsg := "task errored"
+		require.NoError(t, tc.CreateTask(func(_ context.Context) (func() error, error) {
+			return nil, errors.New(errMsg)
+		}))
+		// The error may not be acknowledged by the next create task call because of goroutine scheduling.
```
acohen4, line 69: Yeah this timing/scheduling related behavior definitely makes the abstraction trickier

acohen4: This behavior is actually a little surprising. Is it because the semaphore is released before the eg cancels the context?

acohen4 (approved):
LGTM

src/internal/storage/chunk/task_chain_test.go
brycemcanally: The sleep is intentionally longer for earlier tasks to ensure they ran in parallel. So, the order of sends to the done parallel channel should be the reverse of the task creation.

src/internal/storage/chunk/task_chain_test.go
brycemcanally: A semaphore acquire does a select on the semaphore ready channel and the context done channel. So, you cannot rely on one or the other first regardless of which was actually triggered first, although in our case we release the semaphore first.

PR #8124, opened 2022-08-26
[CORE-1012] Start sidecar PFS server in sidecar mode

robert-uhl:
Thanks to @brycemcanally  for spotting the bug.


brycemcanally (approved):
lgtm

PR #8125, opened 2022-08-27
Snowflake Connector (Read and Write)

albscui:
This PR introduces a new package at `src/integrations/connectors/snowflake`, which implements Snowflake Read and Write connectors.

Added:
- Dockerfile for building the Docker image used for both Read and Write connectors
- Jsonnet files for templating Read and Write connectors
- An integration test for testing Read and Write connectors end-to-end
- Run this integration test in CI

Changed:
- refactored our internal `sdata` package's test utilities into its own sub package.

Outstanding questions:
- How should we version and tag the `pachyderm/snowflake` docker image?
  - What version should we use in the Jsonnet templates? How would users get updates?
- Is there a better way to test this? We might want to move the Jsonnet templates out of the Core Pach repo so that it's more visible to the users. However, we also want the tests to be able to run on these Jsonnet templates, so that any changes won't break functionality.


msteffen (approved):
LGTM

PR #8126, opened 2022-08-28
[CORE-1014] Set AWS creds env vars in sidecar-s3-gateway-enabled pipelines

msteffen:
Sidecar-s3-gateway-enabled pipelines often trigger other, external pipeline systems. When auth is enabled, this is nearly impossible to do correctly, because the user code that is triggering the external pipeline doesn't have the auth token that the external pipeline would need in order to read/write data. This PR fixes the issue by setting the usual AWS env vars, which users can pass to their non-pachyderm pipelines, allowing them to read/write the sidecar S3 gateway.

Jira: [CORE-1014], [INT-735], [FREQQ-56]

[CORE-1014]: https://pachyderm.atlassian.net/browse/CORE-1014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[INT-735]: https://pachyderm.atlassian.net/browse/INT-735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FREQQ-56]: https://pachyderm.atlassian.net/browse/FREQQ-56?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/pps/server/s3g_sidecar_test.go
+++ b/src/server/pps/server/s3g_sidecar_test.go
@@ -26,7 +26,6 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/minikubetestenv"
 	"github.com/pachyderm/pachyderm/v2/src/internal/ppsutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/require"
-	"github.com/pachyderm/pachyderm/v2/src/internal/tarutil"
 	tu "github.com/pachyderm/pachyderm/v2/src/internal/testutil"
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
 	"github.com/pachyderm/pachyderm/v2/src/pps"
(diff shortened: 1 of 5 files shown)


bbonenfant (approved):
This all looks good to me

msteffen:
Update: there were many tests in `s3g_sidecar_test.go` where, in the test, many datums would output to the same file. This has been invalid since 2.0, but for some reason was overlooked until now (even though the test suite was printing warnings, like `    s3g_sidecar_test.go:525: Duplicate file paths from different datums no longer allowed`).

I'm not sure why, but these tests don't seem to have consistently failed until this PR. I'm not sure what Pachyderm does in this case (if you call `ListFile`, I believe you see multiple entries for the same file, but I'm not sure what happens if you call `GetFile`), but I've gone through and updated these tests to respect Pachyderm's new behavior.

PR #8127, opened 2022-08-28
add build-docker-image dependency for load tests

molinamelendezj:
nightly runs are still broken since now we are building arm64 images, and requires the binfmt_misc hack. So proposing we don't rebuild docker images in different places. and only use the same job and make nightly load depend on that.

nightly load will now use same image build job instead of rebuilding in its script. we should consider using the built in requires in circle ci and get rid of `wait_for_docker_images.sh` the ci tests too

### DIFF
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -291,21 +291,17 @@ jobs:
       GOOGLE_COMPUTE_REGION: us-east1
     steps:
       - checkout
-      - when:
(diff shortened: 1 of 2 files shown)


chainlink (approved):
Looks good! FYI We don't use don't require the docker-build to be complete before running the test suites so we can parallelize starting up minikube and building the images

PR #8128, opened 2022-08-29
[CORE-1011] Change liveness probe to not use URI

tybritten:
switches the liveness probe to using cli switches instead of URI. URI breaks if the username or password has an `@` in it.


jrockway (approved):
Looks good.  I always worry about spaces and weird characters in shell expansions, but I am not sure what features to help us are available in /bin/sh (versus bash).  I'd test this with quotes and spaces in the password if you feel like it.

tybritten:
Good point. We can do this for now but once we switch the pg bouncer image, we could just build a pgpassword file at start https://www.postgresql.org/docs/current/libpq-pgpass.html

PR #8129, opened 2022-08-30
Spark compatibility in S3G. (Squash all commits from PR#8115)

lukemarsden:
* Handle paths with a trailing slash as if they didn't have one.
* Don't error on trying to read the timestamp on new files in open commits in the S3 Gateway. Spark needs to be able to read them.
* Don't fail GetObject for paths that are simultaneously an imaginary directory and a real file. Instead, read the file. This requires dancing around the fact that InspectFile doesn't work correctly in this case, but ListFile[0] works just fine.


albscui (approved):
LGTM!

PR #8134, opened 2022-08-30
add additional meta alias commit test

molinamelendezj:
adds an additional meta alias commit test for #8095

### DIFF
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -4033,6 +4033,9 @@ func TestChainedPipelinesNoDelay(t *testing.T) {
 }
 
 // TestMetaAlias tracks https://github.com/pachyderm/pachyderm/pull/8116
+// This test is so we don't regress a problem; where a pipeline in a DAG is started,
+// the pipeline should not get stuck due to a missing Meta Commit. with 8116 the meta
+// repo should get the alias commit.
 // DAG:
 //
 // A
@@ -4040,21 +4043,18 @@ func TestChainedPipelinesNoDelay(t *testing.T) {
 // B
 // |
 // C
-func TestMetaAlias(t *testing.T) {
+func TestStartInternalPipeline(t *testing.T) {
 	if testing.Short() {
 		t.Skip("Skipping integration tests in short mode")
 	}
-
 	t.Parallel()
 	c, _ := minikubetestenv.AcquireCluster(t)
 	aRepo := tu.UniqueString("A")
 	require.NoError(t, c.CreateRepo(aRepo))
-
 	aCommit, err := c.StartCommit(aRepo, "master")
 	require.NoError(t, err)
 	require.NoError(t, c.PutFile(aCommit, "file", strings.NewReader("foo\n"), client.WithAppendPutFile()))
 	require.NoError(t, c.FinishCommit(aRepo, "master", ""))
-
 	bPipeline := tu.UniqueString("B")
 	require.NoError(t, c.CreatePipeline(
 		bPipeline,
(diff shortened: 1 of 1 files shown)


src/server/pachyderm_test.go
```diff
@@ -1784,11 +1784,14 @@ func TestProvenance(t *testing.T) {
 }
 
 // TestProvenance2 tests the following DAG:
-//   A
-//  / \
+//
```
acohen4, line 1787: the indentation is off with B+C

acohen4: can probably revert

src/server/pachyderm_test.go
```diff
@@ -4029,6 +4032,84 @@ func TestChainedPipelinesNoDelay(t *testing.T) {
...
+	cPipeline := tu.UniqueString("C")
+	require.NoError(t, c.CreatePipeline(
+		cPipeline,
+		"",
+		[]string{"sh"},
+		[]string{fmt.Sprintf("cp /pfs/%s/file /pfs/out/file", bPipeline)},
+		&pps.ParallelismSpec{
+			Constant: 1,
+		},
+		client.NewCrossInput(
```
acohen4, line 4081: don't think you need a cross inout here right? Unless I'm missing something, you should be able to use `basicPipelineReq()` to create these pipelines

src/server/pachyderm_test.go
```diff
@@ -4029,6 +4032,84 @@ func TestChainedPipelinesNoDelay(t *testing.T) {
...
+// A
+// |
+// B
+// |
+// C
+func TestMetaAlias(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration tests in short mode")
+	}
+
```
acohen4, line 4047: Just a code style thing - we should generally avoid newlines inside of functions. If it feels necessary, it probably means the space should be filled with a comment

src/server/pachyderm_test.go
```diff
@@ -4029,6 +4032,84 @@ func TestChainedPipelinesNoDelay(t *testing.T) {
...
 
+// TestMetaAlias tracks https://github.com/pachyderm/pachyderm/pull/8116
+// DAG:
+//
+// A
+// |
+// B
+// |
+// C
+func TestMetaAlias(t *testing.T) {
```
acohen4, line 4043: I think the name of this test should be a little more descriptive - maybe `TestStartInternalPipeline`? I think we should include some comments to describe that this was done in response to a problem where under some circumstances when a pipeline in the middle of a DAG is Started, the that Pipeline will get stuck in the future due to a missing Meta Commit, followed by a missing Output commit.

src/server/pachyderm_test.go
armaanv: +1 I think you use a regular PFS input here

src/server/pachyderm_test.go
```diff
@@ -4029,6 +4032,84 @@ func TestChainedPipelinesNoDelay(t *testing.T) {
...
+		"",
+		false,
+	))
+
+	cPipeline := tu.UniqueString("C")
+	require.NoError(t, c.CreatePipeline(
+		cPipeline,
+		"",
+		[]string{"sh"},
+		[]string{fmt.Sprintf("cp /pfs/%s/file /pfs/out/file", bPipeline)},
```
armaanv, line 4077: Btoh bPipeline and cPipeline have the same functionality (copying a file from input repo to output repo) but they have different commands. Not a big deal but just a bit odd. It would be nice to have uniformity here.

armaanv (approved):

PR #8135, opened 2022-08-31
Fix issue with pretty printing JobInfo.Started

albscui:
When the user inspects a job that hasn't started yet, they get something like  `Started: 52 years ago` or `Started: nil timestamp` when `full-timestamp` is set. This is because the `JobInfo.Started` field is nil, which gets interpreted as the unix time 0. This PR addresses this issue by checking for nil timestamp, if so, returning "-".

### DIFF
--- a/src/internal/pretty/pretty.go
+++ b/src/internal/pretty/pretty.go
@@ -32,7 +32,7 @@ func Since(timestamp *types.Timestamp) string {
 // human-readable string, and adds "ago" to the end.
 func Ago(timestamp *types.Timestamp) string {
 	if timestamp == nil {
-		return "N/A"
+		return "-"
 	}
 	since := Since(timestamp)
 	if since == "" {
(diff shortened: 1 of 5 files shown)


src/internal/pretty/pretty.go
```diff
@@ -31,6 +31,9 @@ func Since(timestamp *types.Timestamp) string {
 // Ago pretty-prints the amount of time that has passed since timestamp as a
 // human-readable string, and adds "ago" to the end.
 func Ago(timestamp *types.Timestamp) string {
+	if timestamp == nil {
+		return "N/A"
```
acohen4, line 35: just a thought: would "-" be more consistent in our CLI for when we have nothing to show?

armaanv (approved):

src/server/pachyderm_test.go
```diff
@@ -1784,11 +1784,14 @@ func TestProvenance(t *testing.T) {
 }
 
 // TestProvenance2 tests the following DAG:
-//   A
-//  / \
+//
+//	 A
```
acohen4, line 1788: is this a VSCode autoformatting thing? I saw the same thing on Jose's PR the other day

PR #8138, opened 2022-09-01
Use external postgres secret when requested

dingxiangfei2009:
This PR attempts to fix a problem with deploying enterprise servers, when `postgres` instances has other secret provisions. In addition, this addresses the issue of failing `helm template` when both `pachd` and `enterprise-server` are enabled.

### DIFF
--- a/etc/helm/pachyderm/templates/enterprise-server/deployment.yaml
+++ b/etc/helm/pachyderm/templates/enterprise-server/deployment.yaml
@@ -61,8 +61,8 @@ spec:
         - name: POSTGRES_PASSWORD
           valueFrom:
             secretKeyRef:
-              name: postgres # Must match secret setup by postgres subchart or postgres-secret.yaml if using external postgres
-              key: postgresql-password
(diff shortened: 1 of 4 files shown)


acohen4 (approved):
Looks great! Thanks for contributing @dingxiangfei2009!

acohen4:
> This time we noticed that a duplicated Secret is generated if both `pachd` and `enterprise-server` are enabled. `helm template` will fail because of the duplicated Secrets.

In practice we don't intend users to deploy the enterprise-server and pachd in the same namespace. This is something we should be enforcing with a validation in helm. 
Because of this, we shouldn't include the last commit in this PR changing the name of the postgres secret for the enterprise server.

dingxiangfei2009:
@acohen4 Got it. Reverting the last commit touching the said secret.

One question: does `enterprise-server` has S3 gateway? Its service definition does not mention `s3gateway` port.

dingxiangfei2009:
I also add a check to enforce the said rule.

PR #8145, opened 2022-09-06
CORE-814 Update Unit Tests to Work with GOMAXPROCS=8

FahadBSyed:


.circleci/main.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       - run:
           no_output_timeout: 20m
           command: |-
-            GOMAXPROCS=2 KUBECONFIG=/dev/null PACH_CONFIG=/dev/null \
+            GOMAXPROCS=8 KUBECONFIG=/dev/null PACH_CONFIG=/dev/null \
```
armaanv, line 58: did you try any larger values for GOMAXPROCS? we might be able to bump this up since we are using an xlarge machine

armaanv (approved):

.circleci/main.yml
FahadBSyed: Yup, I tried 16 but I didn't see any improvements in time. I think we're bound by our CI node here

PR #8149, opened 2022-09-06
add loki work around for load test envs

molinamelendezj:
load test env failing to create due to promtail health probs, so added the temp loki workaround

### DIFF
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -709,17 +709,17 @@ workflows:
             parameters:
               bucket:
                 - LOAD1
-                # - LOAD2
-                # - LOAD3
-                # - LOAD4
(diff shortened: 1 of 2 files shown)


jrockway:
I don't think this is needed after #8092.

molinamelendezj:
> I don't think this is needed after #8092.

True. But helium (load test envs) use latest helm chart (currently 2.3.2) with custom pachd images and that change hasn't gone out yet. do you plan to cherry pick #8092 for 2.3.3?

seslattery:
I think I'd prefer to upstream it into the general pulumi/helium usage, not just in the values here, because I think it'd apply to other environments too (though I could be wrong about that).

PR #8150, opened 2022-09-06
Add current project to pachctl config context

albscui:
This PR enables `pachctl config update context --project <project>`, which checks whether `<project>` exists, and sets it in the pach config file. This PR also adds some test clean up code to ensure idempotency since we share the same cluster across multiple runs of commands.


robert-uhl (approved):
LGTM!

PR #8154, opened 2022-09-07
[CORE-1039] Make PPS pipeline inspection project-aware

robert-uhl:
This is the first of several stories. The next story is to key all of the in-memory PPS structures by project name as well as pipeline name.


src/server/pps/iface.go
```diff
@@ -18,6 +18,6 @@ type APIServer interface {
 	StopJobInTransaction(*txncontext.TransactionContext, *pps_client.StopJobRequest) error
 	UpdateJobStateInTransaction(*txncontext.TransactionContext, *pps_client.UpdateJobStateRequest) error
 	CreatePipelineInTransaction(*txncontext.TransactionContext, *pps_client.CreatePipelineRequest) error
-	InspectPipelineInTransaction(*txncontext.TransactionContext, string) (*pps_client.PipelineInfo, error)
+	InspectPipelineInTransaction(txn *txncontext.TransactionContext, projectName, pipelineAncestry string) (*pps_client.PipelineInfo, error)
```
brycemcanally, line 21: Could we just use a pipeline proto here?

robert-uhl: I could, but that makes it unclear in to implementers of the interface if the string `Name` member should be treated like a name or a name-with-ancestry-syntax.

Also, by explicitly changing the signature to include the project name I force an update of all callers and implementations to ensure that the project name is taken into account. Otherwise it would be too easy to have a problem like these:

```go
a.InspectPipelineInTransaction(txn, &Pipeline{Name: pipelineName}) // forgot to include the project!
```

or:

```go
func (a someServer) InspectPipelineInTransaction(txn *txncontext.TransactionContext, p *pps.Pipeline) (*pps.PipelineInfo, error) {
    dbQuery(p.Name) // forgot to include the project!
}
```

I’m trying to use the type system to reduce human error.

src/client/pps.go
```diff
@@ -681,11 +681,11 @@ func (c APIClient) CreatePipeline(
 }
 
 // InspectPipeline returns info about a specific pipeline.
-func (c APIClient) InspectPipeline(pipelineName string, details bool) (*pps.PipelineInfo, error) {
+func (c APIClient) InspectPipeline(pipelineAncestry string, details bool) (*pps.PipelineInfo, error) {
```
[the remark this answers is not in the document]
brycemcanally: The ancestry syntax should be resolved as close to the gRPC boundary (or more generally the boundary where we need it resolved) as possible to prevent that rather than having to track them separately. This pattern makes sense generally for values that are valid in requests, but not internally. For example, we do this with path cleaning when working with files. This allows us to be more flexible with how users can interface with our API while not incurring additional complexity internally.

src/server/pps/iface.go
brycemcanally: What I said about the ancestry syntax above applies here. Also, I don't agree with using more parameters to potentially reduce bugs. It is less maintainable than a proto, forces clients to set fields that may not matter for the function (that is why config structs and functional options exist), and doesn't do much in terms of preventing bugs because it is still easy to just set the wrong value in the argument. For this particular case, we should ensure that a pipeline proto is created with all the fields correctly set when using it internally.

src/client/pps.go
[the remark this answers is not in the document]
robert-uhl: Oh, are you saying that this should just take a plain name and not allow the ancestry syntax? That makes sense. I’ll see if I can make that change work.

src/server/pps/iface.go
robert-uhl: > forces clients to set fields that may not matter for the function

In this case that none of the arguments is optional: a pipeline must never be referred to by pipeline name alone, but always by project name and pipeline name. Passing a Pipeline message permits an implicit missing Project message or an implicit empty Project name.

I changed the signature so that the interface takes a Pipeline message.

src/internal/testpachd/real_env.go
```diff
@@ -127,10 +128,10 @@ func NewRealEnv(t testing.TB, customOpts ...serviceenv.ConfigOption) *RealEnv {
 	realEnv.MockPPSTransactionServer = NewMockPPSTransactionServer()
 	realEnv.ServiceEnv.SetPpsServer(&realEnv.MockPPSTransactionServer.api)
 	realEnv.MockPPSTransactionServer.InspectPipelineInTransaction.
-		Use(func(txnctx *txncontext.TransactionContext, name string) (*pps.PipelineInfo, error) {
+		Use(func(txnctx *txncontext.TransactionContext, pipeline *pps.Pipeline) (*pps.PipelineInfo, error) {
 			return nil, col.ErrNotFound{
 				Type: "pipelines",
-				Key:  name,
+				Key:  fmt.Sprintf("%s/%s", pipeline.Project.GetName(), pipeline.Name),
```
brycemcanally, line 134: This seems like it should be a utility function.

src/server/pps/server/api_server.go
```diff
@@ -1161,12 +1161,12 @@ func (a *apiServer) GetLogs(request *pps.GetLogsRequest, apiGetLogsServer pps.AP
 
 		// 1) Lookup the PipelineInfo for this pipeline/job, for auth and to get the
 		// RC name
-		var pipelineInfo *pps.PipelineInfo
+		var (
+			pipelineInfo *pps.PipelineInfo
```
brycemcanally, line 1165: pipelineInfo could just be declared below with the inspect call.

src/server/pps/server/api_server.go
```diff
@@ -1951,7 +1951,7 @@ func (a *apiServer) CreatePipelineInTransaction(
 			if err := a.env.PFSServer.CreateRepoInTransaction(txnCtx,
 				&pfs.CreateRepoRequest{
 					Repo:        client.NewRepo(input.Cron.Repo),
-					Description: fmt.Sprintf("Cron tick repo for pipeline %s.", request.Pipeline.Name),
+					Description: fmt.Sprintf("Cron tick repo for pipeline %s/%s.", projectName, pipelineName),
```
brycemcanally, line 1954: If we had a utility function, we could just use the pipeline and not have the separate project name variable.

src/server/pps/server/api_server.go
```diff
@@ -2254,17 +2254,18 @@ func (a *apiServer) updatePipeline(
...
 func (a *apiServer) InspectPipeline(ctx context.Context, request *pps.InspectPipelineRequest) (response *pps.PipelineInfo, retErr error) {
-	return a.inspectPipeline(ctx, request.Pipeline.Name, request.Details)
+	return a.inspectPipeline(ctx, request.Pipeline, request.Details)
 }
 
 // inspectPipeline contains the functional implementation of InspectPipeline.
 // Many functions (GetLogs, ListPipeline) need to inspect a pipeline, so they
-// call this instead of making an RPC
-func (a *apiServer) inspectPipeline(ctx context.Context, name string, details bool) (*pps.PipelineInfo, error) {
+// call this instead of making an RPC.  PipelineAncestry may be either a bare
```
brycemcanally, line 2262: update comment.

src/server/pps/server/api_server.go
```diff
@@ -2316,24 +2320,25 @@ func (a *apiServer) InspectPipelineInTransaction(txnCtx *txncontext.TransactionC
 		return nil, errors.New("cannot inspect future pipelines")
 	}
 
-	key, err := ppsutil.FindPipelineSpecCommitInTransaction(txnCtx, a.env.PFSServer, name, "")
+	pipeline.Name = pipelineName
+	commit, err := ppsutil.FindPipelineSpecCommitInTransaction(txnCtx, a.env.PFSServer, pipeline, "")
 	if err != nil {
-		return nil, errors.Wrapf(err, "pipeline not found: couldn't find up to date spec for pipeline %q", name)
+		return nil, errors.Wrapf(err, "pipeline not found: couldn't find up to date spec for pipeline %q/%q", projectName, pipelineName)
```
brycemcanally, line 2326: same as above with utility function.

src/server/pps/server/api_server.go
```diff
@@ -2306,8 +2307,11 @@ func (a *apiServer) inspectPipeline(ctx context.Context, name string, details bo
 	return info, nil
 }
 
-func (a *apiServer) InspectPipelineInTransaction(txnCtx *txncontext.TransactionContext, name string) (*pps.PipelineInfo, error) {
-	name, ancestors, err := ancestry.Parse(name)
+// InspectPipelineInTransaction implements the APIServer interface.
+func (a *apiServer) InspectPipelineInTransaction(txnCtx *txncontext.TransactionContext, pipeline *pps.Pipeline) (*pps.PipelineInfo, error) {
+	projectName := pipeline.Project.GetName()
+	// The pipeline pipelineName arrived in ancestry format; need to turn it into a simple pipelineName.
+	pipelineName, ancestors, err := ancestry.Parse(pipeline.Name)
```
brycemcanally, line 2314: :+1:

src/internal/testpachd/real_env.go
robert-uhl: How do you feel about doing that in the other half of CORE-1039, in PR #8156?

src/pps/pps.go
```diff
@@ -0,0 +1,18 @@
...
+
+import fmt "fmt"
+
+func (j *Job) String() string {
+	return fmt.Sprintf("%s@%s", j.Pipeline.Name, j.ID)
+}
+
+func (p *Pipeline) String() string {
+	if p == nil {
+		return "nil"
```
brycemcanally, line 11: I think we should just use `""` similar to a proto getter for a string field.

src/server/pps/server/api_server.go
```diff
@@ -2316,24 +2315,25 @@ func (a *apiServer) InspectPipelineInTransaction(txnCtx *txncontext.TransactionC
 		return nil, errors.New("cannot inspect future pipelines")
 	}
 
-	key, err := ppsutil.FindPipelineSpecCommitInTransaction(txnCtx, a.env.PFSServer, name, "")
+	pipeline.Name = pipelineName
+	commit, err := ppsutil.FindPipelineSpecCommitInTransaction(txnCtx, a.env.PFSServer, pipeline, "")
 	if err != nil {
-		return nil, errors.Wrapf(err, "pipeline not found: couldn't find up to date spec for pipeline %q", name)
+		return nil, errors.Wrapf(err, "pipeline not found: couldn't find up to date spec for pipeline %q/%q", projectName, pipelineName)
```
brycemcanally, line 2321: Why are we not using a pipeline proto here?

brycemcanally: There are a few places below that seem like they should be operating on protos as well.

src/pps/pps.go
robert-uhl: The issue I see there is with something like:

```go
  var p *pps.Pipeline
  ⋮
  log.Println("processing pipeline", p)
```

The log entry would be `processing pipeline `, which is unlikely to be noticed; `processing pipeline nil` (or even something like `processing pipeline #<NIL>`) is a lot more likely to stick out as being wrong.

src/server/pps/server/api_server.go
robert-uhl: Because when I updated the places project & pipeline names were used together I searched for `%s/%s`, not `%q/%q`. Fixing …

src/pps/pps.go
brycemcanally: yeah, I considered that, but I prefer the consistency for the return values (default values) when using this pattern rather than potentially better logs. Either way, I would like us to use this pattern less, and expect the protos to not be nil when calling these functions.

brycemcanally (approved):
lgtm

PR #8155, opened 2022-09-07
[2.3.x Backport] liveness probe to not use URI (#8128)

tybritten:
Switches pg bouncer liveness probe to not use URI


jrockway (approved):
Very nice!

PR #8160, opened 2022-09-08
arm64 deploy tests

molinamelendezj:
[CORE-1020](https://pachyderm.atlassian.net/browse/CORE-1020)


src/testing/deploy/deploy_test.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/testutil"
 )
 
+var valueOverrides map[string]string
```
jrockway, line 21: I'd add a comment that these are defined per-architecture via build tags (the file named _arm64_ is really just a build tag).  I will note that you can just write `func init() { switch runtime.GOARCH { case "arm64": valueOverrides = ... ; case "amd64": ... } }`, which I think would be clearer.

jrockway (approved):
Wow, nice work.  Not many changes needed at all!

src/testing/deploy/deploy_test.go
molinamelendezj: ooo yes i like that. let me give it some more thought on the next PR, bob and I had come with this approach.

PR #8162, opened 2022-09-12
Additional storage layer unit tests

brycemcanally:
This PR adds a unit test for the new level-based compaction algorithm and a few unit tests for the metadata sharding.

### DIFF
--- a/src/internal/storage/fileset/compaction_test.go
+++ b/src/internal/storage/fileset/compaction_test.go
@@ -1,0 +1,88 @@
+package fileset
+
+import (
+	"context"
+	"io"
+	"math"
+	"math/rand"
+	"testing"
+	"time"
+
+	units "github.com/docker/go-units"
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/require"
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset/index"
+	"github.com/pachyderm/pachyderm/v2/src/internal/storage/track"
+	log "github.com/sirupsen/logrus"
+)
+
+func TestCompactLevelBasedFuzz(t *testing.T) {
+	ctx := context.Background()
+	s := newTestStorage(t)
+	s.shardSizeThreshold = units.KB
+	numFileSets := 1000
+	maxFanIn := 10
+	numSteps := 3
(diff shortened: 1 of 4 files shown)


src/internal/storage/fileset/compaction_test.go
```diff
@@ -0,0 +1,88 @@
...
+	// Check the compaction score of the final file set to ensure no file sets were lost.
+	prims, err := s.flattenPrimitives(ctx, []ID{*id})
+	require.NoError(t, err)
+	var checkScore int64
+	for _, prim := range prims {
+		checkScore += compactionScore(prim)
+	}
+	require.Equal(t, totalScore, checkScore)
+	// Check that compact was called the expected number of times.
+	numCompactions := (numFileSets - 1) / (maxFanIn - 1)
```
acohen4, line 86: why subtract 1 from both values?

src/internal/storage/fileset/index/index_test.go
```diff
@@ -122,6 +127,135 @@ func TestMultiLevel(t *testing.T) {
...
+	db := dockertestenv.NewTestDB(t)
+	tr := track.NewTestTracker(t, db)
+	_, chunks := chunk.NewTestStorage(t, db, tr)
+	cache := NewCache(chunks, 5)
+	// Create ten file sets and concatenate them into one file set.
+	var topIdxs []*Index
+	var totalNumFiles int
+	for i := 0; i < 10; i++ {
+		iw := NewWriter(ctx, chunks, uuid.NewWithoutDashes())
+		numFiles := rand.Intn(100000)
```
acohen4, line 141: would it be worth recording the sequence of rand values so that in case of failure we can run forensics?

acohen4 (approved):
LGTM. left a couple comments/questions

src/internal/storage/fileset/compaction_test.go
brycemcanally: The number of compactions is equal to the number of internal nodes in a perfect m-ary tree. This is a formula for computing that.

src/internal/storage/fileset/index/index_test.go
brycemcanally: I am planning on doing a larger set of changes in a separate PR for reporting the seeds of fuzz tests.

PR #8163, opened 2022-09-12
Expand RealEnv Mock and Convert Auth Integration Tests to Unit Tests

FahadBSyed:
## Description
This PR converts 45 of the auth package's integration tests into unit tests and increases the coverage of the auth package. It also replaces the PPS mock server in the realEnv package with a real PPS that uses the fake kubernetes API mock. In order to make that change, I needed to break a few import cycles which required exporting some functions that were previously unexported.


armaanv (approved):
This looks good to me. Nice changes to the auth tests. I had a discussion with Fahad offline and he said that adding a pps server backed by the fake kube clientset is needed for upcoming tests in the enterprise package. I think we should use something like https://github.com/pachyderm/fakeapiserver in the long term but this approach makes sense at the moment.

PR #8166, opened 2022-09-13
regenerate protos as they've gone out of sync

albscui:
The `config.proto` and compiled code went out of sync. Interestingly, this wasn't caught by `make test-proto-static`, because we were swallowing the output of the `test` command. This PR fixes that, so CI should error when protos need to be recompiled.


robert-uhl (approved):
LGTM

PR #8167, opened 2022-09-13
[CORE-859] Implement pachctl put file untar

brycemcanally:
This PR implements the untar flag for pachctl put file, which provides functionality for untarring the tar file(s) and putting each file within the tar stream(s) as a separate file in PFS.


FahadBSyed:
Logic looks good but think you have a few linting errors in integration-tests-MISC

FahadBSyed (approved):

PR #8169, opened 2022-09-13
update etcd to an image that has an arm64 build

jrockway:


molinamelendezj (approved):
@jrockway can we update it for the test suite too so we are testing with what is released. [start-minikube-vm.sh](https://github.com/pachyderm/pachyderm/blob/00fc79097c6a4adb295bc9662e86aa0dc0934c6c/etc/kube/start-minikube-vm.sh#L82)

jrockway:
@molinamelendezj We should fix that script in another way.  For anything we don't build before calling that script, it's just grabbing it from the Internet, which is what k8s will do anyway when you start a pod that refers to it.

I did this to pre-cache in my local setup: 
```shell
# Try to pre-load some images
for img in $(helm template pachyderm etc/helm/pachyderm -f $SCRIPT_DIR/hostname-values.yaml -f $SCRIPT_DIR/enterprise-key-values.yaml -f $SCRIPT_DIR/values.yaml | yq '.. | select(type|test("map")) | .image | select(. != null)' | grep -v -- --- | grep -v pachd); do
    (docker pull $img && kind load docker-image $img) || echo "Could not pre-load $img"
done
```

jrockway:
@molinamelendezj  I've updated to v3.5.5, which fixes some bugs around transactions, and also updated reset.py and start-minikube-vm.sh

PR #8172, opened 2022-09-14
[CORE-1042] Cache Symlink Commit Uploads

FahadBSyed:
## Description
This PR resolves a bug reported in CORE-1042 by expanding the cacheClient to wrap the ModifyFile client's copyFile(). Instead of copying files in a commit, a commit ID is put into a cache and a fileSet is generated for each commit. This saves time by bypassing calls to Postgres and Auth.

### DIFF
--- a/src/internal/pfssync/cache_client.go
+++ b/src/internal/pfssync/cache_client.go
@@ -4,6 +4,9 @@ import (
 	io "io"
 	"sync"
 
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pfsload"
+
 	"github.com/hashicorp/golang-lru/simplelru"
 	"github.com/pachyderm/pachyderm/v2/src/client"
 	"github.com/pachyderm/pachyderm/v2/src/internal/pfsdb"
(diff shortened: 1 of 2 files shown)


acohen4:
What do we think the main pathological cases here were? Is it the GetFileSet RPC (network + DB I/O)?

FahadBSyed:
> What do we think the main pathological cases here were? Is it the GetFileSet RPC (network + DB I/O)?

I think its the overhead of auth and postgres calls per copy() operation whereas with creating a fileSet once, we can just pull that from the cache over and over again.

src/internal/pfssync/cache_client.go
```diff
@@ -66,3 +69,45 @@ func (cc *CacheClient) put(key string, commit *pfs.Commit) {
 	defer cc.mu.Unlock()
 	cc.cache.Add(key, commit)
 }
+
+func (cc *CacheClient) WithCreateFileSetClient(cb func(client.ModifyFile) error) (*pfs.CreateFileSetResponse, error) {
+	pachClient := pfsload.NewPachClient(cc.APIClient)
```
brycemcanally, line 74: Why can't we just use the embedded APIClient?

src/server/worker/pipeline/transform/worker.go
```diff
@@ -264,16 +264,16 @@ func handleDatumSet(driver driver.Driver, logger logs.TaggedLogger, task *DatumS
 	stats := &datum.Stats{ProcessStats: &pps.ProcessStats{}}
 	if err := pachClient.WithRenewer(func(ctx context.Context, renewer *renew.StringSet) error {
 		// Setup file operation client for output meta commit.
```
brycemcanally, line 266: very minor nit, could we move this comment down to the output meta commit `WithCreateFileSetClient` call.

src/internal/pfssync/cache_client.go
FahadBSyed: We can but I noticed, the `WithCreateFileSetClient()` function for `pfsload.Client` seems to wrap the `WithCreateFileSetClient()` for `APIClient` and adds a metadata() call to ensure that the metadata is configured properly. If that step is unnecessary then, we can instead use the embedded `APIClient` instead

brycemcanally: That step is specific to the load testing tooling, the pach client available through the driver is already properly configured.

FahadBSyed: Gotcha, so I will swap back to the `APIClient`

brycemcanally (approved):
lgtm

PR #8173, opened 2022-09-14
fix unbound variable error on release test envs

molinamelendezj:
quick fix unbound variable error on release test env since` -u` is set


chainlink:
Adding this for posterity (TIL about default values in bash) https://stackoverflow.com/questions/15143863/bash-colon-operator-in-variable-substitution

chainlink (approved):
You might want to use a different value as the default. I found `gcp` a bit confusing initially, i think `default` might be more clear?

PR #8174, opened 2022-09-14
[CORE-1037] add startedTime marker for ListCommit

armaanv:
This PR adds support for marker based pagination for the ListCommit API. The number parameter is reused to limit the number of results returned from the API and a startedTime parameter is introduced as the marker. Consumers of the API can provide the commit started time of the last commit in their previous ListCommit call to get a new page of results.


src/client/pfs.go
```diff
@@ -234,9 +235,9 @@ func (c APIClient) inspectCommit(repoName string, branchName string, commitID st
 // If `to` and `from` are the same commit, no commits will be returned.
 // `number` determines how many commits are returned.  If `number` is 0,
 // all commits that match the aforementioned criteria are returned.
-func (c APIClient) ListCommit(repo *pfs.Repo, to, from *pfs.Commit, number int64) ([]*pfs.CommitInfo, error) {
```
robert-uhl, line 237: This changes the signature of client.APIClient.ListCommit, which is a breaking API change and hence requires a new major version.

I would recommend adding a new method like `ListCommitsSince` or somesuch _if_ you think you need to add anything to package client.

robert-uhl (changes requested):
Rather than add an alternate set of pagination-aware methods to the Go client library, I think that you should just make pagination available via the gRPC API. Looking at usage of the client library outside of package client, it looks like `ListCommit` is just used in tests, with a nil `startTime`.

I would just revert the changes to package client and leave it as-is for now.

src/client/pfs.go
armaanv: Makes sense. I've removed the changes to the go client and updated the tests to match

brycemcanally (approved):
lgtm, just address the random comment changes. We should eventually make this more performant by threading it through to the Postgres query.

armaanv:
> Why did several random comments get modified in this PR?

This is auto-formatting changes from my IDE. The PR touched a few more files before I removed the go client changes so that's why you see those other changes here. I'll remove them before merging

PR #8200, opened 2022-09-16
[CORE-860] Implement chunk prefetcher

brycemcanally:
This PR implements a file set layer chunk prefetcher that is used when getting small / medium (KB/MB) files from PFS. Prior to this, chunk prefetching would only be effective for larger files (~100MB+), which meant that getting large batches of small / medium files would be subject to serially downloading and processing the chunks.

### DIFF
--- a/src/internal/storage/chunk/storage.go
+++ b/src/internal/storage/chunk/storage.go
@@ -16,19 +16,21 @@ import (
 
 const (
 	// TrackerPrefix is the prefix used when creating tracker objects for chunks
-	TrackerPrefix   = "chunk/"
-	prefix          = "chunk"
-	defaultChunkTTL = 30 * time.Minute
+	TrackerPrefix        = "chunk/"
+	prefix               = "chunk"
+	defaultChunkTTL      = 30 * time.Minute
+	DefaultPrefetchLimit = 10
 )
 
 // Storage is the abstraction that manages chunk storage.
 type Storage struct {
-	objClient obj.Client
-	db        *pachsql.DB
-	tracker   track.Tracker
-	store     kv.Store
-	memCache  kv.GetPut
-	deduper   *miscutil.WorkDeduper
+	objClient     obj.Client
+	db            *pachsql.DB
+	tracker       track.Tracker
+	store         kv.Store
+	memCache      kv.GetPut
+	deduper       *miscutil.WorkDeduper
+	prefetchLimit int
 
 	createOpts CreateOptions
 }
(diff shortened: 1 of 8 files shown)


src/internal/storage/chunk/storage.go
```diff
@@ -19,7 +19,7 @@ const (
 	TrackerPrefix        = "chunk/"
 	prefix               = "chunk"
 	defaultChunkTTL      = 30 * time.Minute
-	defaultPrefetchLimit = 10
+	DefaultPrefetchLimit = 10
```
FahadBSyed, line 22: It seems like this const is only used below in `NewStorage()` is it exported because the plan is to reuse it in other packages in future PRs?

src/internal/storage/chunk/storage.go
```diff
@@ -67,6 +67,11 @@ func (s *Storage) NewDataReader(ctx context.Context, dataRef *DataRef) io.Reader
 	return newDataReader(ctx, client, s.memCache, s.deduper, dataRef, 0)
 }
 
+func (s *Storage) PrefetchData(ctx context.Context, dataRef *DataRef) error {
```
FahadBSyed, line 70: Sidenote, if you changed `NewDataReader()` to:

``` go
func (s *Storage) NewDataReader(ctx context.Context, dataRef *DataRef) *DataReader {
	client := NewClient(s.store, s.db, s.tracker, nil)
	return newDataReader(ctx, client, s.memCache, s.deduper, dataRef, 0)
}
```

You could change `PrefetchData()` to: 
```go
func (s *Storage) PrefetchData(ctx context.Context, dataRef *DataRef) error {
	return s.NewDataReader(ctx, dataRef).fetchData()
}
```

It seems like `DataReader` is already exported and implements the io.Reader interface, so this shouldn't break anything. Structuring it this may might help make `PrefetchData` resilient to changes made to `NewClient`'s signature.

src/internal/storage/fileset/prefetcher.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+		dataRefs := f.Index().File.DataRefs
+		// Handle empty files.
+		// It does not matter which chunk they get emitted with,
+		// just that they are emitted in the correct order.
+		if len(dataRefs) == 0 {
+			files = append(files, f)
+			return nil
+		}
+		// Handle first chunk.
```
FahadBSyed, line 92: This comment might be a little misleading. If we have a file that spans multiple chunks, once we prefetch all the chunks and emit the file on the last chunk, `ref` and `files` will both be `nil`. In that case this code will trigger again.
At the same time I'm not sure how to best phrase what it is actually doing.

src/internal/storage/fileset/prefetcher_test.go
```diff
@@ -0,0 +1,101 @@
...
+				require.Equal(t, i, p)
+				i++
+				return nil
+			}))
+			require.Equal(t, test.numFiles, i)
+		})
+	}
+}
+
+func BenchmarkNoPrefetcher(b *testing.B) {
```
FahadBSyed, line 63: Glad you included these benchmarks. Locally it took about 2.5 minutes without the prefetcher vs 30 seconds with it. Big difference!

FahadBSyed (approved):
I have some questions and nits, but code looks good to merge as-is

src/internal/storage/chunk/storage.go
brycemcanally: Yeah, I want to eventually merge the configurations for prefetching in the file set and chunk layer.

src/internal/storage/fileset/prefetcher.go
brycemcanally: I refactored the code a bit so we don't need to treat this as a special case.

PR #8208, opened 2022-09-19
[CORE-1041] Refactor realEnv code into a separate package

FahadBSyed:


src/server/pfs/fuse/fuse_test.go
```diff
@@ -579,7 +579,7 @@ func TestMountDir(t *testing.T) {
 		require.Equal(t, "bar", filepath.Base(files[0].Name()))
 		require.Equal(t, "foo", filepath.Base(files[1].Name()))
 
-		//require.ElementsEqualUnderFn(t, []string{"foo", "bar"}, []string{"foo", "bar"}, func(f interface{}) interface{} { return f.(iofs.FileInfo).Name() })
+		// require.ElementsEqualUnderFn(t, []string{"foo", "bar"}, []string{"foo", "bar"}, func(f interface{}) interface{} { return f.(iofs.FileInfo).Name() })
```
armaanv, line 582: Is this an auto formatting change from the IDE?

src/server/pps/server/pipeline_controller.go
```diff
@@ -434,10 +434,10 @@ func (pc *pipelineController) apply(ctx context.Context, pi *pps.PipelineInfo, r
 	return nil
 }
 
-// RcIsFresh returns a boolean indicating whether rc has the right labels
+// rcIsFresh returns a boolean indicating whether rc has the right labels
```
armaanv, line 437: I think this function was being exported before your changes. Do we need to lock it down?

armaanv (approved):
looks good to me. I left a couple of minor comments

src/server/pfs/fuse/fuse_test.go
FahadBSyed: Yeah it is, you are right

src/server/pps/server/pipeline_controller.go
FahadBSyed: I think it was unexported before based on the diff here:
https://github.com/pachyderm/pachyderm/blob/707a081ac581286d758ba4732aaedd409a926a8c/src/server/pps/server/pipeline_controller.go

PR #8211, opened 2022-09-20
Make ListJob project-aware

albscui:
This PR makes listing jobs project-aware, meaning jobs can now be scoped to a project. User can either get jobs across all projects, or jobs scoped to only a single project.

The filtering is on the server side. One implementation detail is that our protos, we don't have a dedicated flag for "all projects", we assume that if the user doesn't supply any filter, then they mean "all projects".


src/server/pps/cmds/cmds.go
```diff
@@ -51,6 +53,30 @@ const (
 	secrets   = "secrets"
 )
 
+// copied from Client, because need to extend for projects without breaking user facing API.
```
robert-uhl, line 56: I would change the signature to `(context.Context,  pps.APIClient, func(*pps.JobInfo) error) error`, letting the caller pull stuff out of the `*client.APIClient`.

src/server/pps/cmds/cmds.go
```diff
@@ -266,16 +309,29 @@ $ {{alias}} -p foo -i bar@YYY`,
 					})
 				} else {
 					// We are listing all sub-jobs, possibly restricted to a single pipeline
+					var pipeline *pps.Pipeline
+					if pipelineName != "" {
+						pipeline = &pps.Pipeline{Name: pipelineName}
```
robert-uhl, line 314: Don’t forget to set the project too: `Project: &pfs.Project{Name: project}` — this is another reason to only support a single project.

src/server/pps/cmds/cmds.go
```diff
@@ -307,6 +363,8 @@ $ {{alias}} -p foo -i bar@YYY`,
 		}),
 	}
 	listJob.Flags().StringVarP(&pipelineName, "pipeline", "p", "", "Limit to jobs made by pipeline.")
+	listJob.Flags().StringSliceVar(&projects, "project", []string{}, "Limit to jobs in a project.")
```
robert-uhl, line 366: Definitely think `--project` should take only a single project.

src/server/pps/server/api_server.go
```diff
@@ -593,6 +593,29 @@ func forEachCommitInJob(pachClient *client.APIClient, jobID string, wait bool, c
 	return nil
 }
 
+// For backwards compatibility, we want users who have objects that are not associated with projects
+// to belong to the "default" project. This means that both client and server should essentially treat
+// "default" and "" project names to be equivalent.
+func keepJobBasedOnProjects(job *pps.Job, projectsFilter []string) bool {
```
robert-uhl, line 599: Rather than hardcoding `"default"` use `pfs.DefaultProjectName`.

Also, this function is needlessly complex: you are iterating thrice over the projects filter for each job. Just turn the project list into a `map[string]bool` which returns true for any requested project.

robert-uhl: I think something like this would work:

```go
type stringFilter map[string]bool

func makeStringFilter(f []string) map[string]bool {
	var ff = make(map[string]bool)
	for _, s := range f {
		ff[s] = true
	}
	return ff
}

func (a *apiServer) ListJob(request *pps.ListJobRequest, resp pps.API_ListJobServer) (retErr error) {
	filter := makeStringFilter(request.GetProjects())
	return a.listJob(resp.Context(), request, func(ji *pps.JobInfo) error {
		if len(filter) > 0 && !filter[ji.GetJob().GetPipeline().GetProject().GetName()] {
			return nil
		}
		return errors.EnsureStack(resp.Send(ji))
	})
}


// ListJobSet implements the protobuf pps.ListJobSet RPC
func (a *apiServer) ListJobSet(request *pps.ListJobSetRequest, serv pps.API_ListJobSetServer) (retErr error) {
	pachClient := a.env.GetPachClient(serv.Context())

	// Track the jobsets we've already processed
	seen := map[string]struct{}{}

	// Return jobsets by the newest job in each set (which can be at a different
	// timestamp due to triggers or deferred processing)
	jobInfo := &pps.JobInfo{}
	filter := makeStringFilter(request.GetProjects())
	err := a.jobs.ReadOnly(serv.Context()).List(jobInfo, col.DefaultOptions(), func(string) error {
		if _, ok := seen[jobInfo.Job.ID]; ok {
			return nil
		}
		seen[jobInfo.Job.ID] = struct{}{}

		jobInfos, err := pachClient.InspectJobSet(jobInfo.Job.ID, request.Details)
		if err != nil {
			return err
		}
		
		var affectsProjects bool
		for _,ji := range jobInfos {
			if len(filter) > 0 && filter[ji.Job.Pipeline.Project.GetName()] {
				affectsProjects = true
			}
		}
		if !affectsProjects {
			return nil
		}

		return errors.EnsureStack(serv.Send(&pps.JobSetInfo{
			JobSet: client.NewJobSet(jobInfo.Job.ID),
			Jobs:   jobInfos,
		}))
	})
}
```

src/server/pps/server/api_server.go
```diff
@@ -2803,6 +2843,7 @@ func (a *apiServer) propagateJobs(txnCtx *txncontext.TransactionContext) error {
 
 		// Check if there is an existing job for the output commit
 		job := client.NewJob(pipelineInfo.Pipeline.Name, txnCtx.CommitSetID)
+		job.Pipeline.Project = pipelineInfo.Pipeline.Project // TODO remove this before merging
```
robert-uhl, line 2846: Does this need to be removed?

robert-uhl (changes requested):

src/server/pps/server/api_server.go
albscui: I assumed that you would be introducing this as part of your PR.

src/server/pps/server/api_server.go
robert-uhl: > I changed the proto field to a map directly

I think that leaving it as a list is probably cleaner from the API perspective. The fact that it’s a map from project name to boolean seems like an implementation detail, not something meaningful to a client of the API. Are there useful semantics to passing a mapping from a project name to _false_?

src/server/pps/cmds/cmds.go
```diff
@@ -235,6 +246,15 @@ $ {{alias}} -p foo -i bar@YYY`,
 				return errors.New("cannot set --output (-o) without --raw")
 			}
 
+			// To list jobs for all projects, user must be explicit about it.
+			// The --project filter takes precedence over everything else.
+			// By default use pfs.DefaultProjectName
+			projectsFilter := map[string]bool{project: true}
+			if allProjects {
+				projectsFilter = nil
+			} else if project == pfs.DefaultProjectName && project != pachCtx.GetProject() {
```
robert-uhl, line 255: Couldn’t this condition just be `project == pfs.DefaultProjectName`; if `project == pachCtx.GetProject()` then it doesn’t _hurt_ to reset `projectsFilter`, just wastes a few CPU cycles.

src/server/pps/server/api_server.go
```diff
@@ -617,18 +617,39 @@ func (a *apiServer) ListJobSet(request *pps.ListJobSetRequest, serv pps.API_List
...
-		seen[jobInfo.Job.ID] = struct{}{}
+		seen[id] = struct{}{}
 
-		jobInfos, err := pachClient.InspectJobSet(jobInfo.Job.ID, request.Details)
+		jobInfos, err := pachClient.InspectJobSet(id, request.Details)
 		if err != nil {
 			return err
 		}
 
+		logrus.Info("albscui len(jobInfos) given id ", id, len(jobInfos))
```
robert-uhl, line 631: Need to delete this debugging line☺

src/server/pps/server/api_server.go
```diff
@@ -703,13 +724,10 @@ func (a *apiServer) intersectCommitSets(ctx context.Context, commits []*pfs.Comm
 // ListJobStream.
 func (a *apiServer) listJob(
```
robert-uhl, line 725: I think that `listJob` has only one caller now; if so, you might as well merge it into `ListJob`.

src/server/pps/cmds/cmds.go
```diff
@@ -1563,3 +1600,25 @@ func ParsePipelineStates(stateStrs []string) (string, error) {
 	}
 	return validateJQConditionString(strings.Join(conditions, " or "))
 }
+
+// copied from Client, because need to extend for projects without breaking user facing API.
```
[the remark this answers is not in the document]
robert-uhl: Thanks for the explanation — makes perfect sense.

For the benefit of history: the reason is the need to pass the new filter into `ListJob`, and adding it to `ListProjectJobFilterF` would change the API surface.

src/pps/pps.proto
```diff
@@ -443,6 +443,7 @@ message InspectJobSetRequest {
 
 message ListJobSetRequest {
   bool details = 1;
+  map<string, bool> projects = 2;
```
robert-uhl, line 446: Might be worth adding a comment here about the semantics, but they are pretty obvious.

src/pps/pps.proto
```diff
@@ -453,6 +454,7 @@ message InspectJobRequest {
 }
 
 message ListJobRequest {
+  map<string, bool> projects = 7;
```
robert-uhl, line 457: https://github.com/pachyderm/pachyderm/pull/8211/files#r1015683677 ditto

robert-uhl (approved):

src/server/pps/cmds/cmds.go
albscui: Changed to `func listJobFilterF(ctx context.Context, client pps.API_ListJobClient, f func(*pps.JobInfo) error) error`

src/server/pps/server/api_server.go
albscui: Changed!

PR #8214, opened 2022-09-21
[CORE-1050] Add Auth Middleware to RealEnv and Convert Some Integration Tests to Unit Tests

FahadBSyed:
This PR is a little big, so apologies for that in advance:
- `NewMockEnv()` and `NewMockPachd()` now take an additional parameter that enables the `auth middleware interceptor`. This is needed for some of our integration tests to function. The logic is parameterized to prevent some callers of `MockEnv()` from having to set up auth in their tests. This is handy for tests unrelated to the auth functions.
- In order to enable the `auth middleware interceptor`, the `auth` mock in `MockPachd` had to be expanded to include missing methods of the `auth.APIServer` interface.
- `Enterprise` integration tests were converted into unit tests with the exception of tests that rely on `Pause()` behavior. In order to convert these, we'd have to spin up our `RealEnv` in a pauseable mode which requires more serious changes.
- `License` integration tests were all able to be converted into unit tests once the auth middleware was hooked up.
- The `!k8s` build flag was added to files where only unit tests were defined. This ensures that CircleCI integration jobs don't run the unit tests (which are already getting ran in the test-go job`. This required some helper functions in `pfs` tests to be refactored into separate files so they would be included whether the `k8s` build flag was on or not. The downside of structuring the flags this way is that we don't have a way to run the integration tests and unit tests at the same time. However, I assume that that should be okay given how we have our tests configured to run in CircleCI.


src/internal/testpachd/mock_env.go
```diff
@@ -23,7 +23,7 @@ type MockEnv struct {
 
 // NewMockEnv constructs a MockEnv for testing, which will be destroyed at the
 // end of the test.
-func NewMockEnv(t testing.TB) *MockEnv {
+func NewMockEnv(t testing.TB, interceptAuth bool) *MockEnv {
```
acohen4, line 26: WDYT about making this a functional option?

src/server/pfs/s3/worker_test.go
```diff
@@ -226,9 +228,6 @@ func workerListObjectsRecursive(t *testing.T, s *workerTestState) {
 }
 
 func TestWorkerDriver(t *testing.T) {
-	if testing.Short() {
```
armaanv, line 229: Why remove this?

src/server/pfs/pfs_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build !k8s
```
armaanv, line 1: Is the purpose of the !k8s flag to avoid running the tests twice?

src/server/pps/server/pipeline_controller_test.go
```diff
@@ -5,6 +5,9 @@ import (
 	"testing"
 	"time"
 
+	logrus "github.com/sirupsen/logrus"
```
armaanv, line 8: why change this?

src/server/pfs/fuse/helpers_test.go
```diff
@@ -0,0 +1,69 @@
+package fuse
```
armaanv, line 1: can we put these helper functions in the integration test file? It doesn't look like they are being called anywhere else

src/server/license/server/api_server.go
```diff
@@ -109,7 +109,8 @@ func (a *apiServer) activate(ctx context.Context, req *lc.ActivateRequest) (resp
 	// Allow request to override expiration in the activation code, for testing
 	if req.Expires != nil {
 		customExpiration, err := types.TimestampFromProto(req.Expires)
-		if err == nil && expiration.After(customExpiration) {
+		// TODO(fahad): Determine if the call to expiration.After() can be safely removed.
+		if err == nil /* && expiration.After(customExpiration) */ {
```
armaanv, line 113: Why does this check need to be removed?. It prevents the expiration time from being overridden to an later value

src/server/debug/shell/server.go
```diff
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	globlib "github.com/pachyderm/ohmyglob"
+
```
armaanv, line 22: nit: remove this change pls

src/server/enterprise/server/api_server_test.go
```diff
@@ -1,65 +0,0 @@
...
-	"time"
-
-	"github.com/pachyderm/pachyderm/v2/src/enterprise"
-	"github.com/pachyderm/pachyderm/v2/src/internal/dockertestenv"
-	"github.com/pachyderm/pachyderm/v2/src/internal/require"
-	"github.com/pachyderm/pachyderm/v2/src/internal/testpachd/realenv"
-	tu "github.com/pachyderm/pachyderm/v2/src/internal/testutil"
-)
-
-func TestEnterpriseDeactivate(t *testing.T) {
```
armaanv, line 16: do we have replacements for these tests? are they not needed anymore?

PR #8217, opened 2022-09-21
bump console version

molinamelendezj:


chainlink (approved):
Why -1 ?

molinamelendezj:
> Why -1 ?

@pleeko and I assume it's when you want to release a console patch only without another core release. Like we debated for this one, since there is not core patches.

PR #8226, opened 2022-09-26
Support yaml input spec in datum mounting

smalyala:
Currently, the input spec can only be JSON. Add support for YAML since PPS pipeline specs can be in YAML and it's more readable than JSON. After mounting datums, automatically format input specs to be YAML.

Jira: https://pachyderm.atlassian.net/browse/INT-696


pleeko:
Are we only allowing Yaml as input or is JSON also allowed?

smalyala:
> Are we only allowing Yaml as input or is JSON also allowed?

Both are allowed.

pleeko (approved):
nit: You should add a happy path case for YAML in the Datum tests.

smalyala:
Added!

PR #8227, opened 2022-09-27
Fix file set renewal in new compaction algorithm

brycemcanally:
This PR fixes file set renewal in the new compaction algorithm. The intermediary file sets during the level based compaction were not getting renewed, so they would get garbage collected during longer compactions. This PR also adds a test that simulates this scenario by running the garbage collector and lowering the ttl for file sets such that they are less than the run time for a test compactor.


FahadBSyed:
LGTM, it took me a while to understand why we needed to renew the intermediary filesets in both places, but its because the algorithm is recursive and we need the fileset generated by the compact callback to be renewed while we recurse through Ids.

FahadBSyed (approved):

FahadBSyed (approved):

PR #8235, opened 2022-09-30
Extend auth with project primitives, and implement CreateProject

albscui:
This PR does the following:
- updates auth protos to be project-aware, including various relevant permissions and project resource type
- create project related roles, but not all are actually hooked up or implemented (this is ongoing work)
- intercept `CreateProject` requests in our auth middleware, to check for `ProjectCreate` permission
- all users get `ProjectCreate` permission through a role binding that binds `allClusterUsers` to `ProjectCreator` (which contains the `ProjectCreate` permission), at the cluster level
- when a user creates a project, they automatically become the owner
- all users in the default project are automatically `ProjectWriter`


src/auth/auth.proto
```diff
@@ -218,8 +218,17 @@ enum Permission {
   REPO_ADD_PIPELINE_READER    = 212;
   REPO_REMOVE_PIPELINE_READER = 213;
   REPO_ADD_PIPELINE_WRITER    = 214;
+  REPO_CREATE = 215;
 
   PIPELINE_LIST_JOB     = 301;
+  PIPELINE_CREATE = 302;
```
acohen4, line 224: these pipeline permissions shouldn't be necessary. We use REPO perms to imply them

src/auth/auth.proto
```diff
@@ -218,8 +218,17 @@ enum Permission {
...
+  PIPELINE_MODIFY = 303;
+  PIPELINE_DELETE = 304;
+
+  PROJECT_CREATE = 400;
+  PROJECT_DELETE = 401;
+  PROJECT_LIST_REPO = 402;
+  PROJECT_LIST_PIPELINE = 403;
```
acohen4, line 231: similarly PROJECT_LIST_PIPELINE shouldn't be necessary

src/internal/clusterstate/v2.4.0.go
```diff
@@ -24,6 +26,25 @@ var state_2_4_0 migrations.State = state_2_3_0.
 			return errors.Wrap(err, "could not create default project")
 		}
 		return nil
+	}).
+	Apply("Grant ProjectCreator to all users at cluster level", func(ctx context.Context, env migrations.Env) error {
+		key := auth_server.ResourceKey(&auth.Resource{Type: auth.ResourceType_CLUSTER})
+		rb := &auth.RoleBinding{Entries: make(map[string]*auth.Roles)}
+		if err := auth_server.RoleBindingCollection(nil, nil).ReadWrite(env.Tx).Get(key, rb); err != nil {
```
acohen4, line 33: we should move src/server/auth/server/authdb.go ->  src/internal/pfsdb/authdb.go so that the migration package doesn't depend on the server package

src/internal/clusterstate/v2.4.0.go
```diff
@@ -24,6 +26,25 @@ var state_2_4_0 migrations.State = state_2_3_0.
...
+			if errors.Is(err, col.ErrNotFound{}) {
+				// auth is not activated
+				return nil
+			}
+			return errors.EnsureStack(err)
+		}
+		if _, ok := rb.Entries[auth.AllClusterUsersSubject]; !ok {
+			rb.Entries[auth.AllClusterUsersSubject] = &auth.Roles{Roles: make(map[string]bool)}
+		}
+		rb.Entries[auth.AllClusterUsersSubject].Roles[auth.ProjectWriter] = true
```
acohen4, line 43: they should be assigned ProjectCreator instead of ProjectWriter at the cluster level. Since the default project is created at this point, we should also set allClusterUsers ProjectWriter for the default project

src/server/auth/server/api_server.go
```diff
@@ -301,7 +301,7 @@ func (a *apiServer) getClusterRoleBinding(ctx context.Context) (*auth.RoleBindin
 			return nil, errors.New("cached cluster binding had unexpected type")
 		}
 
-		if bindings.Entries == nil {
+		if len(bindings.Entries) == 0 {
```
acohen4, line 304: why is this necessary?

src/server/auth/server/api_server.go
```diff
@@ -473,8 +473,9 @@ func (a *apiServer) activateInTransaction(ctx context.Context, txCtx *txncontext
 		roleBindings := a.roleBindings.ReadWrite(txCtx.SqlTx)
 		if err := roleBindings.Put(clusterRoleBindingKey, &auth.RoleBinding{
 			Entries: map[string]*auth.Roles{
-				auth.RootUser: {Roles: map[string]bool{auth.ClusterAdminRole: true}},
-				internalUser:  {Roles: map[string]bool{auth.ClusterAdminRole: true}},
+				auth.RootUser:               {Roles: map[string]bool{auth.ClusterAdminRole: true}},
+				internalUser:                {Roles: map[string]bool{auth.ClusterAdminRole: true}},
+				auth.AllClusterUsersSubject: {Roles: map[string]bool{auth.ProjectCreator: true}},
```
acohen4, line 478: we should also check if the default project exists, and if it does give `allClusterUsers` ProjectWriter for it

src/server/auth/server/api_server.go
```diff
@@ -676,6 +677,8 @@ func (a *apiServer) evaluateRoleBindingInTransaction(txnCtx *txncontext.Transact
 		return nil, err
 	}
 
+	// TODO get role bindings at Project level, and evaluate like above
```
acohen4, line 680: remove this TODO here?

src/server/auth/server/roles.go
```diff
@@ -171,6 +171,44 @@ func init() {
 		},
 	})
 
+	// Project related roles
+	projectViewer := registerRole(&auth.Role{
+		Name:          auth.ProjectViewer,
+		ResourceTypes: []auth.ResourceType{auth.ResourceType_PROJECT},
```
acohen4, line 177: all of these project roles should also be applicable for the CLUSTER resource type

src/server/auth/server/roles.go
```diff
@@ -185,6 +223,10 @@ func init() {
 			licenseAdminRole.Permissions,
 			secretAdminRole.Permissions,
 			pachdLogReaderRole.Permissions,
+			projectViewer.Permissions,
```
acohen4, line 226: you shouldn't need `projectViewer.Permissions, projectWriter.Permissions` here because they are implied by `projectOwner.Permissions`

src/server/auth/server/roles.go
```diff
@@ -207,14 +249,22 @@ func init() {
 				auth.Permission_CLUSTER_ENTERPRISE_DEACTIVATE,
 				auth.Permission_CLUSTER_DELETE_ALL,
 				auth.Permission_CLUSTER_ENTERPRISE_PAUSE,
+				auth.Permission_PROJECT_DELETE,
 			}),
 	})
 }
 
 func combinePermissions(permissions ...[]auth.Permission) []auth.Permission {
+	seen := make(map[auth.Permission]bool)
```
acohen4, line 258: with the above change I don't think this is necessary.

src/server/pachyderm_test.go
```diff
@@ -1785,11 +1785,14 @@ func TestProvenance(t *testing.T) {
 }
 
 // TestProvenance2 tests the following DAG:
-//   A
-//  / \
```
acohen4, line 1789: revert please

src/server/auth/server/testing/auth_test.go
```diff
@@ -2298,3 +2298,15 @@ func TestGetRobotTokenErrorNonAdminUser(t *testing.T) {
 	require.YesError(t, err)
 	require.Matches(t, "needs permissions \\[CLUSTER_AUTH_GET_ROBOT_TOKEN\\] on CLUSTER", err.Error())
 }
+
+func TestCreateProject(t *testing.T) {
```
acohen4, line 2302: We should also test the case where a user doesn't have PROJECT_CREATE permission

src/server/pfs/server/driver.go
```diff
@@ -429,6 +429,11 @@ func (d *driver) createProject(ctx context.Context, req *pfs.CreateProjectReques
 		return errors.Wrapf(err, "invalid project name")
 	}
 	return d.env.TxnEnv.WithWriteContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
+		whoAmI, err := txnCtx.WhoAmI()
```
acohen4, line 432: Since the authentication check is already happening at the interceptor, it would be cleaner to include this into the Create Role binding block:

`if whoAmI, err := txnCtx.WhoAmI(); err != nil {
....
} else if err !errors.Is(err, auth.ErrNotActivated) {
  return errors.Wrapf(err, "error authenticating (must log in to create project)")
}`

acohen4 (changes requested):

src/server/auth/server/api_server.go
albscui: to catch when `bindings.Entries` isn't nil but an empty map.
https://go.dev/play/p/_igsFVhZJGt

src/auth/auth.go
```diff
@@ -19,6 +19,11 @@ const (
 	// authenticated context
 	ContextTokenKey = "authn-token"
 
+	// clusterRoleBindingKey is a key in etcd, in the roleBindings collection,
+	// that contains the set of role bindings for the cluster. These are frequently
+	// accessed so we cache them.
+	ClusterRoleBindingKey = "CLUSTER:"
+
```
robert-uhl, line 26: Does this need to be part of a PR specific to projects? I can see that it might, but since it wasn't there earlier ...

src/server/pachyderm_test.go
albscui: I'm going  to wrap in `/* .. */` so that I don't have to worry about it anymore. I don't actually know how to configure my autoformatter to not format the `//`

src/auth/auth.go
albscui: Basically need to reference this in a lot of different places. Previously it was private in the auth server package, but really it should be here I think.

src/auth/auth.proto
```diff
@@ -218,8 +218,13 @@ enum Permission {
   REPO_ADD_PIPELINE_READER    = 212;
   REPO_REMOVE_PIPELINE_READER = 213;
   REPO_ADD_PIPELINE_WRITER    = 214;
+  REPO_CREATE = 215;
```
acohen4, line 221: this should be PROJECT_CREATE_REPO if you want to be consistent with PROJECT_LIST_REPO below

src/server/auth/server/testing/auth_test.go
```diff
@@ -1025,8 +1029,12 @@ func TestListAndInspectRepo(t *testing.T) {
 			auth.Permission_REPO_INSPECT_FILE,
 			auth.Permission_PIPELINE_LIST_JOB,
 		},
+		repoNone: {
+			auth.Permission_PROJECT_CREATE,
+		},
 	}
 	for _, info := range repoInfos {
+		fmt.Println(info.Repo.Name)
```
acohen4, line 1037: remove please!

src/auth/auth.proto
albscui: Hmm I prefer `REPO_CREATE` over `PROJECT_CREATE_REPO`, since since creating repos aren't tied to projects per se... what if I change `PROJECT_LIST_REPO` to just `LIST_REPO` or `REPO_LIST`?

acohen4: hmmm... 
`PROJECT_LIST_REPO` and `PROJECT_CREATE_REPO` seems to align more closely with our existing naming convention of specifying the most granular resource an action can be done against, and the action. For example, we have `REPO_LIST_FILE` which implies that for as granular as the REPO resource, we can do the action "LIST_FILE". 
WDYT?

albscui: True, I understand that reasoning, but I was coming from the perspective of using the fewest words to convey the meaning of each permission, since I didn't want to add too much noise to these names. But I think this is largely subjective. I'll update to `PROJECT_*`

src/server/auth/server/roles.go
```diff
@@ -171,6 +171,40 @@ func init() {
...
+		ResourceTypes: []auth.ResourceType{auth.ResourceType_PROJECT, auth.ResourceType_CLUSTER},
+		Permissions: combinePermissions(projectWriter.Permissions, []auth.Permission{
+			auth.Permission_PROJECT_DELETE,
+			auth.Permission_REPO_DELETE,
+		}),
+	})
+
+	projectCreator := registerRole(&auth.Role{
+		Name:          auth.ProjectCreator,
+		ResourceTypes: []auth.ResourceType{auth.ResourceType_CLUSTER, auth.ResourceType_CLUSTER},
```
acohen4, line 202: remove duplicate `auth.ResourceType_CLUSTER`

acohen4 (approved):
Suggested one small change. Otherwise LGTM!!

src/server/auth/server/roles.go
albscui: thanks for the catch!

src/auth/auth.go
robert-uhl: Cool. Should capitalise it in the comment too.

src/internal/authdb/authdb.go
```diff
@@ -12,11 +14,12 @@ const (
 	roleBindingsCollectionName = "role_bindings"
 	membersCollectionName      = "members"
 	groupsCollectionName       = "groups"
+	InternalUser               = auth.InternalPrefix + "auth-server"
```
robert-uhl, line 17: For this & all the other exported items, be sure to add some documentation which would be useful when using godoc.

robert-uhl (changes requested):
Looks pretty good from my point of view. Only change which leaps out at me is documenting stuff in package `authdb`.

robert-uhl (approved):
LGTM

PR #8237, opened 2022-10-03
remove docker version spec in ci

armaanv:
use the default docker version rather than specifying one per job.

### DIFF
--- a/etc/testing/circle/install.sh
+++ b/etc/testing/circle/install.sh
@@ -35,7 +35,7 @@ fi
 # To get the latest minikube version:
 # curl https://api.github.com/repos/kubernetes/minikube/releases | jq -r .[].tag_name | sort -V | tail -n1
 if [ ! -f cached-deps/minikube ] ; then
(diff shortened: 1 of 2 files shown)


molinamelendezj (approved):
we will always get the default version then?

etc/testing/circle/install.sh
```diff
@@ -35,7 +35,7 @@ fi
 # To get the latest minikube version:
 # curl https://api.github.com/repos/kubernetes/minikube/releases | jq -r .[].tag_name | sort -V | tail -n1
 if [ ! -f cached-deps/minikube ] ; then
-    MINIKUBE_VERSION=v1.26.0 
+    MINIKUBE_VERSION=v1.27.0
```
chainlink, line 38: We've had some issues with caching here, we should make sure 1.27 is actually getting installed.

chainlink (approved):
one small comment

msteffen:

PR #8242, opened 2022-10-04
[CORE-1074] Add Identity Server and Dex Web to RealEnv

FahadBSyed:
integration tests to unit tests.

### DIFF
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -67,7 +67,7 @@ jobs:
       - run:
           no_output_timeout: 20m
           command: |-
-            GOMAXPROCS=8 KUBECONFIG=/dev/null PACH_CONFIG=/dev/null \
+            GOMAXPROCS=8 CGO_ENABLED=0 KUBECONFIG=/dev/null PACH_CONFIG=/dev/null \
             gotestsum \
             --junitfile ${TEST_RESULTS}/gotestsum-report.xml \
             --rerun-fails \
--- a/Makefile
+++ b/Makefile
@@ -223,7 +223,7 @@ proto: docker-build-proto
 	./etc/proto/build.sh
 
 # Run all the tests. Note! This is no longer the test entrypoint for travis
-test: clean-launch launch-dev lint enterprise-code-checkin-test docker-build test-pfs-server test-cmds test-libs test-auth test-identity test-license test-enterprise test-worker test-admin test-pps
+test: clean-launch launch-dev lint enterprise-code-checkin-test docker-build test-pfs-server test-cmds test-libs test-auth test-license test-enterprise test-worker test-admin test-pps
 
 enterprise-code-checkin-test:
 	@which ag || { printf "'ag' not found. Run:\n  sudo apt-get install -y silversearcher-ag\n  brew install the_silver_searcher\nto install it\n\n"; exit 1; }
--- a/src/internal/testutil/identity.go
+++ b/src/internal/testutil/identity.go
@@ -37,22 +37,26 @@ func OIDCOIDCConfig(host, issuerPort, redirectPort string, local bool) *auth.OID
 
 // ConfigureOIDCProvider configures the identity service and the auth service to
 // use a mock connector.
-func ConfigureOIDCProvider(t *testing.T, c *client.APIClient, realEnv bool) error {
+func ConfigureOIDCProvider(t *testing.T, c *client.APIClient, unitTest bool) error {
 	adminClient := AuthenticateClient(t, c, auth.RootUser)
 
 	_, err := adminClient.IdentityAPIClient.DeleteAll(adminClient.Ctx(), &identity.DeleteAllRequest{})
 	require.NoError(t, err)
 
+	// In the case of integration tests, the following defaults are used to due to routing constraints
+	// imposed by minikube in integration tests. The address won't be reachable because the service external IP
+	// is not reachable without a tunnel on minikube's IP.
+	local := true
 	issuerHost := "pachd"
 	issuerPort := "1658"
 	redirectPort := "1657"
-	local := true
 
-	if realEnv {
+	if unitTest {
+		// In unit tests, all grpc servers listen on the peerPort which is randomly generated.
+		local = false
 		issuerHost = c.GetAddress().Host
 		issuerPort = strconv.Itoa(int(c.GetAddress().Port + 8))
 		redirectPort = strconv.Itoa(int(c.GetAddress().Port + 7))
-		local = false
 	}
 
 	_, err = adminClient.SetIdentityServerConfig(adminClient.Ctx(), &identity.SetIdentityServerConfigRequest{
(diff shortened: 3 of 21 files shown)


Makefile
```diff
@@ -223,7 +223,7 @@ proto: docker-build-proto
 	./etc/proto/build.sh
 
 # Run all the tests. Note! This is no longer the test entrypoint for travis
-test: clean-launch launch-dev lint enterprise-code-checkin-test docker-build test-pfs-server test-cmds test-libs test-auth test-identity test-license test-enterprise test-worker test-admin test-pps
+test: clean-launch launch-dev lint enterprise-code-checkin-test docker-build test-pfs-server test-cmds test-libs test-auth test-license test-enterprise test-worker test-admin test-pps
```
acohen4, line 226: I'm missing something, but where do identity-test get called now?

FahadBSyed: So I'm not sure if this is the best solution, but they aren't getting called anymore because they have no integration tests to run.

It looks like we get an error like this in Circle when we attempt to call it. It is strange because I figured it would just be a no-op.

https://app.circleci.com/pipelines/github/pachyderm/pachyderm/15250/workflows/a1dae3ed-4da9-4645-937b-11de8dd4111c/jobs/233829

src/server/identity/server/dex_web.go
```diff
@@ -60,6 +61,10 @@ var (
 // This is always /dex-assets in the docker image, but it can be overriden for testing
 var webDir = "/dex-assets"
 
+func WebDir(dir string) {
```
FahadBSyed, line 64: @acohen4 I wanted to get your thoughts on if this is an okay way to access and change the path to '/dex-assets'

.circleci/main.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       - run:
           no_output_timeout: 20m
           command: |-
-            GOMAXPROCS=8 KUBECONFIG=/dev/null PACH_CONFIG=/dev/null \
+            GOMAXPROCS=8 CGO_ENABLED=0 KUBECONFIG=/dev/null PACH_CONFIG=/dev/null \
```
FahadBSyed, line 70: @acohen4 I wanted to make sure this was okay as well. I think from a testing perspective it works fine, however I don't know if we actually do compile anything that relies on `cgo`. We have a file that inits dex only if cgo is excluded

src/server/auth/server/testing/oidc_test.go
```diff
@@ -70,12 +75,13 @@ func TestOIDCTrustedApp(t *testing.T) {
...
 func TestCannotAuthenticateWithExpiredLicense(t *testing.T) {
-	if testing.Short() {
-		t.Skip("Skipping integration tests in short mode")
-	}
-	c, _ := minikubetestenv.AcquireCluster(t)
-	tu.ActivateAuthClient(t, c)
-	require.NoError(t, tu.ConfigureOIDCProvider(t, c))
+	t.Parallel()
+	env := realenv.NewRealEnv(t, dockertestenv.NewTestDBConfig(t))
+	env.ActivateIdentity("../../../../../dex-assets")
```
acohen4, line 80: since this is used in a bunch of places, could we ActivateIdentity with a functional option to NewRealEnv?

src/server/identity/server/dex_web.go
acohen4: feels unideal. WDYT about figuring out the path to dex-assets by traversing up directories similar to what we do here: https://github.com/pachyderm/pachyderm/blob/master/src/internal/minikubetestenv/deploy.go#L88-L101 ?

src/internal/testutil/identity.go
```diff
@@ -22,28 +23,41 @@ import (
...
 	_, err := adminClient.IdentityAPIClient.DeleteAll(adminClient.Ctx(), &identity.DeleteAllRequest{})
 	require.NoError(t, err)
 
+	issuerHost := "pachd"
+	issuerPort := "1658"
+	redirectPort := "1657"
+	local := true
+
+	if realEnv {
+		issuerHost = c.GetAddress().Host
```
acohen4, line 52: what does this value typically evaluate to? I'm wondering if this could work for all cases

src/internal/testutil/identity.go
```diff
@@ -22,28 +23,41 @@ import (
...
+		issuerPort = strconv.Itoa(int(c.GetAddress().Port + 8))
+		redirectPort = strconv.Itoa(int(c.GetAddress().Port + 7))
+		local = false
```
acohen4, line 55: why does localhost issuer get set to false in the realEnv case?

acohen4 (changes requested):
This is awesome progress!! I left some questions / suggestions

src/internal/testutil/identity.go
FahadBSyed: Leaving this as "pachd" might be okay. I can test that out. But this is the host name generated by server that gets spun up in mockPachd. In my case `127.0.0.1` or localhost. The key is actually the port, since we listen on a randomly generated port which allows us to run multiple tests in parallel.

src/internal/testutil/identity.go
```diff
@@ -143,10 +157,16 @@ func GetOIDCTokenForTrustedApp(t testing.TB, testClient *client.APIClient) strin
 		return http.ErrUseLastResponse
 	}
 
+	redirectPort := "1657"
+
+	if realEnv {
+		redirectPort = strconv.Itoa(int(testClient.GetAddress().Port + 7))
```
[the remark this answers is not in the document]
FahadBSyed: Because we want to support running multiple integration tests concurrently, we need to configure the redirect port based on the dynamically generated port from mockPachd. Otherwise, we'll have collisions if multiple instances tried to claim port 1657.

The code that sets them is in real_env.go:

```
identityAddr := realEnv.PachClient.GetAddress().Host + ":" + strconv.Itoa(int(realEnv.PachClient.GetAddress().Port+8))
```

src/server/identity/server/dex_web.go
FahadBSyed: Yeah that could work!

src/server/auth/server/testing/oidc_test.go
FahadBSyed: So I tried to play around with this and unfortunately we can not. The best we can do is write an option to set one of the configuration variables to a magic value. Then if the value matches, we can call `ActivateIdentity()`. The underlying reason is that the config option allows us to change the `serviceenv.Configuration` struct, but we need to hook up changes to the realEnv as well. The solution I ended up going with for now is to define a new exported function for Identity being enabled.

acohen4: No problem. That makes sense. I'm giving this another pass and thinking through the existing code a bit more

acohen4 (approved):
This is awesome! LGTM

PR #8243, opened 2022-10-04
pgbouncer image release workflow

molinamelendezj:
[CORE-1022](https://pachyderm.atlassian.net/browse/CORE-1022)


Dockerfile.pgbouncer
```diff
@@ -0,0 +1,36 @@
...
+    make && \
+    # Manual install
+    cp pgbouncer /usr/bin && \
+    mkdir -p /etc/pgbouncer /var/log/pgbouncer /var/run/pgbouncer && \
+    # entrypoint installs the configuration, allow to write as postgres user
+    touch /etc/pgbouncer/userlist.txt && \
+    addgroup --gid 70 postgres 2>/dev/null && \
+    adduser --quiet --uid 70 --home /var/lib/postgresql  --shell /bin/bash --ingroup postgres --disabled-password postgres 2>/dev/null && \
+    chown -R postgres /var/run/pgbouncer /etc/pgbouncer && \
+    # Cleanup
```
jrockway, line 24: It would be better (make a smaller image) if we copied the build artifacts into a new container (FROM debian:bullseye AS build; <build>; FROM debian:bullseye; USER/ENTRYPOINT/CMD; COPY --from=build /usr/bin/pgbouncer ....).  It would also be good to use `gcr.io/distroless/base-debian11` instead of debian, because it contains the necessary support files for running rootless, etc.

jrockway (approved):
Looks good.  You should file a ticket to follow up on making the image size smaller, this one is going to be larger than necessary.

Dockerfile.pgbouncer
```diff
@@ -0,0 +1,36 @@
+FROM debian:bullseye
```
chainlink, line 1: Is this bespoke, or is this dockerfile copied from somewhere?

molinamelendezj: @chainlink comes from our pachyderm pgbouncer repo. just moving it to core. plan to release the new image, but do a round of testing before bumping it in our helm.

chainlink: Right, but we're not using that currently correct (currently still using bitnami?)

molinamelendezj: yea a bitnami image that was pulled and pushed. we are just going to rebuild same version of it. but need to do enough testing to feel this "new image" is good to update the helm value ref

PR #8244, opened 2022-10-04
PPS: Be more careful about k8s errors during RC updates

jrockway:
A customer was seeing the symptom of having work-in-progress canceled, with PIPELINE_FAILED and "error updating RC: context canceled" in the "reason" field.  This can only happen if the context is canceled while talking to k8s.  Why it's canceled is unclear; the most obvious point of cancelation is the PPS master losing its lease.  But I tried that, and the symptoms are different; when the master restarts, it scales workers down to 0 (as the transition into PIPELINE_STANDBY) at the k8s level without calling StopJob (which was happening in the customer's case).  To truly understand what's going on here, we should have a reproduction that causes "context canceled" while talking to k8s.  (We don't set a deadline or anything like that.)

This PR has some preemptive fixes.  I've made "context canceled" into an error that doesn't fail the pipeline outright.  That should prevent scaling down to 0 when the context gets canceled.  I also handle "deadline exceeded" there (with retry allowed) in case we add RPC deadlines around k8s in the future; we should, because we don't want to wait "forever" on a misbehaving k8s API server, which is what we do now.

I also added a ton of logging:

* All outgoing HTTP requests are logged when they complete, or if they're taking more than 10 seconds.  The completion is DEBUG level, because PPS polls RCs constantly and it would be too many logs.  Long requests are INFO though, so we should always have those in debug dumps.  The side effect of this change is that other stuff is logged, like reads and writes from object storage.  Things like that make me want to log at INFO level (it's quite informative), but I'll skip that for now.
* When a step in the pipeline controller fails to apply, we now return a description of the step in the error message ("KUBERNETES_RESOURCES(UP)" for example).  If we had that logging for this customer issue, it would be clear exactly what operation failed.  This involves implementing fmt.Stringer for side effects.  A lot of lines but straightforward code.
* Added a bunch of log messages around scaling up and enhanced the detail in the existing ones.
* Corrected some log messages that should be prefixed with "PPS master:" to include that prefix.

Right now, this isn't a full fix of all the problems, but will give us more debugging information next time.  The user will also see more detail in the "reason" field when a pipeline transitions into the failure state because of failing-to-apply side effects.

This is a difficult one to add automated tests for.  I tested manually extensively:

* To make k8s fail with "context canceled", I made losing the PPS master into an on-demand thing (`func init() { go http.ListenAndServe(func{ cancel() })`).  This didn't reproduce the customer's issue.
* I added another similar handler to make k8s return "context canceled" without causing the PPS master to exit.  This DOES reproduce the issue when it happens during autoscaling scale up, and this PR does fix that issue.   (Submit a job, wait for it to start, make k8s slow, submit another job, note failure when k8s call fails during autoscaling up.)
* I SIGSTOP'd my API server to test the logging around long HTTP requests.  It works.

### DIFF
--- a/src/server/pps/server/pipeline_controller.go
+++ b/src/server/pps/server/pipeline_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -60,6 +61,26 @@ const (
 	sideEffectName_CRASH_MONITOR        sideEffectName = 5
 )
 
+// String implements fmt.Stringer.
+func (s sideEffectName) String() string {
+	switch s {
+	case sideEffectName_KUBERNETES_RESOURCES:
+		return "KUBERNETES_RESOURCES"
+	case sideEffectName_FINISH_COMMITS:
+		return "FINISH_COMMITS"
+	case sideEffectName_RESTART:
+		return "RESTART"
+	case sideEffectName_SCALE_WORKERS:
+		return "SCALE_WORKERS"
+	case sideEffectName_PIPELINE_MONITOR:
+		return "PIPELINE_MONITOR"
+	case sideEffectName_CRASH_MONITOR:
+		return "CRASH_MONITOR"
+	default:
+		return "UNKNOWN"
+	}
+}
+
 type sideEffectToggle int32
 
 const (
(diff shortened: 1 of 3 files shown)


src/server/pps/server/pipeline_controller.go
```diff
@@ -60,6 +61,26 @@ const (
 	sideEffectName_CRASH_MONITOR        sideEffectName = 5
 )
 
+// String implements fmt.Stringer.
+func (s sideEffectName) String() string {
```
acohen4, line 65: 🙏

acohen4 (approved):
LGTM! This will definitely make it easier to debug :)

PR #8262, opened 2022-10-07
[CORE-872] Enable Memory-Backed Pachyderm Worker Volumes

FahadBSyed:
## Description

This PR adds a `bool` flag to the pipeline spec called `memory_volumes` when set to `true`,  pachyderm will set the `emptyDir.medium` to `"Memory"`. This increases performance for some workloads that rely on caching or are IO heavy.


src/internal/serviceenv/config.go
```diff
@@ -96,6 +96,7 @@ type PachdSpecificConfiguration struct {
 	ImagePullSecrets           string `env:"IMAGE_PULL_SECRETS,default="`
 	MemoryRequest              string `env:"PACHD_MEMORY_REQUEST,default=1T"`
 	WorkerUsesRoot             bool   `env:"WORKER_USES_ROOT,default=false"`
+	WorkerVolumesMemoryBacked  bool   `env:"WORKER_VOLUMES_MEMORY_BACKED,default=false"`
```
brycemcanally, line 99: Why is this needed?

src/pps/pps.proto
```diff
@@ -36,6 +36,7 @@ message Transform {
   string user = 11;
   string working_dir = 12;
   string dockerfile = 13;
+  bool memory_backed_volume = 14;
```
brycemcanally, line 39: Could we shorten this to `memory_volume`.

src/internal/serviceenv/config.go
FahadBSyed: Hey good catch, I initially put this here when I misread the JIRA's suggested solution. This should be removed.

src/pps/pps.proto
FahadBSyed: Sure, can do.

PR #8269, opened 2022-10-11
Use gotestsum for PPS tests and separate into their own run

chainlink:
This changes the PPS tests to use `gotestsum` which unlocks rerun on failure within a given run (instead of needing to manually hit re-run and rerun the entire bucket)

Example of a successful re-run https://app.circleci.com/pipelines/github/pachyderm/pachyderm/15459/workflows/93968c83-5301-4e6b-a322-ecc6e25e9625/jobs/238706

It also introduces some nice cleanup (no more need to check buckets, etc) and reduces indirection (between run_tests.sh, makefile, and main.yml)

This also removes `build.sh` from the circle config, which was just building `pachyderm/ubuntuplusnetcat:local` which ended up not being used in the tests anyway 

Uses circle parallelism, which unlocks some nice interface insights:
image


.circleci/main.yml
```diff
@@ -87,6 +87,112 @@ jobs:
...
+          no_output_timeout: 20m
+          command: |
+            PROM_PORT=$(kubectl --namespace=monitoring get svc/prometheus -o json | jq -r .spec.ports[0].nodePort)
+            export PROM_PORT
+
+            bucket_num=$CIRCLE_NODE_INDEX
+            num_buckets=$CIRCLE_NODE_TOTAL
+            package="./src/server"
+
+            tests=( $(go test -v -tags=k8s  "${package}" -list ".*" | grep -v '^ok' | grep -v '^Benchmark') )
```
molinamelendezj, line 161: thinking if we should pass in tags as params to reuse this task/job later on

.circleci/main.yml
```diff
@@ -87,6 +87,112 @@ jobs:
...
+              "start=bucket_size * (bucket_num-1)" \
+              "bucket_size+=bucket_num < num_buckets ? 0 : total_tests%num_buckets"
+
+            test_regex="$(IFS=\|; echo "${tests[*]:start:bucket_size}")"
+
+            echo $test_regex
+
+            gotestsum \
+              --junitfile ${TEST_RESULTS}/gotestsum-report.xml \
+              --rerun-fails \
```
molinamelendezj, line 180: Maximum rerun attempts defaults to 2, and can be changed with --rerun-fails=n good enough?

chainlink: Are we good with the default 2 for now?

PR #8272, opened 2022-10-12
Make repo role bindings aware of projects

albscui:


src/server/pfs/server/driver.go
```diff
@@ -219,7 +219,7 @@ func (d *driver) createRepo(txnCtx *txncontext.TransactionContext, repo *pfs.Rep
 				txnCtx,
 				whoAmI.Username,
 				[]string{auth.RepoOwnerRole},
-				&auth.Resource{Type: auth.ResourceType_REPO, Name: repo.Name},
+				&auth.Resource{Type: auth.ResourceType_REPO, Name: pfsdb.RepoKey(repo)},
```
robert-uhl, line 222: Should this be the database key of the repo? That feels like using package `pfsdb` for something unrelated to the PFS database layer. Maybe this should be the result of `repo.String` instead?

src/server/pfs/server/testing/server_test.go
```diff
@@ -310,6 +313,50 @@ func TestPFS(suite *testing.T) {
 		}
 	})
 
+	suite.Run("CreateRepoWithSameNameInDifferentProjects", func(t *testing.T) {
```
robert-uhl, line 316: … AndAuth☺

acohen4:
Reminder that this will need a migration as well

acohen4:
I actually think it would make a lot of sense to do the Auth migration as a part of this PR @albscui

albscui:
> I actually think it would make a lot of sense to do the Auth migration as a part of this PR @albscui

After some discussion, I think we agree that we will just do this in the big migrations work @robert-uhl is currently working on: https://github.com/pachyderm/pachyderm/pull/8245

robert-uhl (approved):

PR #8273, opened 2022-10-12
require scheme on userAccessibleOauthIssuerHost

seslattery:
Setting oidc.issuerURI: "http://pachd:30658/dex" but wanting a userAccessibleOauthIssuerHost with a scheme of https is currently incompatible, as `pachctl auth login -b` would use the scheme from the oidc.issuerURI.  This is a breaking config change, where we now require a scheme to be explicitly set when configuring userAccessibleOauthIssuerHost, and a guard clause has been added to try and prevent that misconfiguration.

### DIFF
--- a/etc/helm/pachyderm/templates/_helpers.tpl
+++ b/etc/helm/pachyderm/templates/_helpers.tpl
@@ -135,7 +135,7 @@ true
 {{- define "pachyderm.userAccessibleOauthIssuerHost" -}}
 {{- if .Values.oidc.userAccessibleOauthIssuerHost -}}
   {{- if not (hasPrefix "http" .Values.oidc.userAccessibleOauthIssuerHost) -}}
-    {{ fail (printf "oidc.userAccessibleOauthIssuerHost must begin with http:// or https://, not %s" .Values.oidc.userAccessibleOauthIssuerHost) }}
+    {{- printf "%s://%s" (include "pachyderm.hostproto" .) .Values.oidc.userAccessibleOauthIssuerHost -}}
   {{- end -}}
 {{ .Values.oidc.userAccessibleOauthIssuerHost }}
 {{- else if (include "pachyderm.host" .) -}}
--- a/etc/helm/pachyderm/templates/_helpers.tpl
+++ b/etc/helm/pachyderm/templates/_helpers.tpl
@@ -136,8 +136,9 @@ true
 {{- if .Values.oidc.userAccessibleOauthIssuerHost -}}
   {{- if not (hasPrefix "http" .Values.oidc.userAccessibleOauthIssuerHost) -}}
     {{- printf "%s://%s" (include "pachyderm.hostproto" .) .Values.oidc.userAccessibleOauthIssuerHost -}}
+  {{- else -}}
+  {{ .Values.oidc.userAccessibleOauthIssuerHost }}
   {{- end -}}
-{{ .Values.oidc.userAccessibleOauthIssuerHost }}
 {{- else if (include "pachyderm.host" .) -}}
 {{- printf "%s://%s" (include "pachyderm.hostproto" .) (include "pachyderm.host" .) -}}
 {{- else  -}}
--- a/etc/helm/pachyderm/templates/_helpers.tpl
+++ b/etc/helm/pachyderm/templates/_helpers.tpl
@@ -134,11 +134,14 @@ true
 
 {{- define "pachyderm.userAccessibleOauthIssuerHost" -}}
 {{- if .Values.oidc.userAccessibleOauthIssuerHost -}}
+  {{- if not (hasPrefix "http" .Values.oidc.userAccessibleOauthIssuerHost) -}}
+    {{ fail (printf "oidc.userAccessibleOauthIssuerHost must begin with http:// or https://, not %s" .Values.oidc.userAccessibleOauthIssuerHost) }}
+  {{- end -}}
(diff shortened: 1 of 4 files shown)


etc/helm/pachyderm/templates/_helpers.tpl
```diff
@@ -134,11 +134,14 @@ true
 
 {{- define "pachyderm.userAccessibleOauthIssuerHost" -}}
 {{- if .Values.oidc.userAccessibleOauthIssuerHost -}}
+  {{- if not (hasPrefix "http" .Values.oidc.userAccessibleOauthIssuerHost) -}}
```
BOsterbuhr, line 137: IIUC this is a breaking change because users don't currently have to have `http(s)://` but this would require that?
Would it be possible to use the same logic we have for [pachyderm.hostproto](https://github.com/pachyderm/pachyderm/blob/4a47712ebc3d6dafa61e3e86c33a05edb0e45e57/etc/helm/pachyderm/templates/_helpers.tpl#L39-L45) to avoid a breaking change?
Maybe something like this? btw I have no idea if this would work or is even possible.
```
  {{- if not (hasPrefix "https" .Values.oidc.userAccessibleOauthIssuerHost) -}}
    {{- if or .Values.ingress.tls.enabled .Values.proxy.tls.enabled .Values.ingress.uriHttpsProtoOverride -}}
    <add https:// prefix to userAccessibleOauthIssuerHost>
    {{- else if not (hasPrefix "http" .Values.oidc.userAccessibleOauthIssuerHost) -}}
    <add http:// prefix to userAccessibleOauthIssuerHost>
```

BOsterbuhr (commented):
Curious if we can avoid a breaking change.

etc/helm/pachyderm/templates/_helpers.tpl
acohen4: +1

tybritten: yeah, i don't think we should make a change to the format if `.Values.oidc.userAccessibleOauthIssuerHost`, just to `pachyderm.userAccessibleOauthIssuerHost`
so I think it shouldn't even check for http(s).. just automatically add `pachyderm.hostproto`

tybritten: there should never be a situation where `userAccessibleOauthIssuerHost` has a different proto than `pachyderm.hostproto`

seslattery:
Ok I pushed up a change making it explicitly non-breaking. PTAL

etc/helm/pachyderm/templates/_helpers.tpl
```diff
@@ -134,11 +134,15 @@ true
...
 {{- if .Values.oidc.userAccessibleOauthIssuerHost -}}
-{{ .Values.oidc.userAccessibleOauthIssuerHost }}
+  {{- if not (hasPrefix "http" .Values.oidc.userAccessibleOauthIssuerHost) -}}
+    {{- printf "%s://%s" (include "pachyderm.hostproto" .) .Values.oidc.userAccessibleOauthIssuerHost -}}
+  {{- else -}}
+  {{ .Values.oidc.userAccessibleOauthIssuerHost }}
+  {{- end -}}
 {{- else if (include "pachyderm.host" .) -}}
-{{- (include "pachyderm.host" .) -}}
+{{- printf "%s://%s" (include "pachyderm.hostproto" .) (include "pachyderm.host" .) -}}
```
jrockway, line 143: I don't think you should use this necessarily, but there is actually a URL building function built into helm: https://helm.sh/docs/chart_template_guide/function_list/#urljoin

jrockway (approved):
Looks good.

jrockway:
BTW @BOsterbuhr @seslattery just to be ultra up front about this, it's still potentially a breaking change for people that supply their own auth secret (pachyderm-auth).  I'm not sure what the semantics would look like if we accounted for that; on the one hand, generating a manifest with our chart and then expecting it to work forever is the wrong outlook, but on the other hand, people do do it.  That said, the one customer I've seen doing this does have a valid URL there, and like I said, I'm not sure what we would do without a scheme on that URL anyway.  Assume HTTP?

etc/helm/pachyderm/templates/_helpers.tpl
seslattery: Nice, going to leave it this way now for consistency, but might make sense to switch how we construct must url's to that in a followup PR.

seslattery:
> BTW @BOsterbuhr @seslattery just to be ultra up front about this, it's still potentially a breaking change for people that supply their own auth secret (pachyderm-auth). I'm not sure what the semantics would look like if we accounted for that; on the one hand, generating a manifest with our chart and then expecting it to work forever is the wrong outlook, but on the other hand, people do do it. That said, the one customer I've seen doing this does have a valid URL there, and like I said, I'm not sure what we would do without a scheme on that URL anyway. Assume HTTP?

Ah good callout on auth-secret.  I mean I could assume http, but even then it's still a potentially breaking change in the chance they actually wanted https.  I guess I could fallback to the old behavior and look at the issuerURI scheme?  But I really don't know that it's worth adding so many layers of logic indirection to try and prevent breakages, as I think it becomes much harder to support in the future.  Though we could support the old behavior here, and log.Warn an explicit message that field will require an explicit scheme set in 2.4.0+?

tybritten (approved):
I think its fine. I don't think many customers are using the secret directly.

PR #8277, opened 2022-10-13
[CORE-1123] Support External Loki Instance with Pachyderm

FahadBSyed:
## Description
This PR allows users to define two new helm values that will instantiate the lokiClient's address. This allows users to point to an external loki instance in the cluster.

This shouldn't be used if you're deploying your own pachyderm-loki instance. In the case that a user tries to do that, helm will display a warning.

### DIFF
--- a/doc/docs/2.3.x/deploy-manage/deploy/loki.md
+++ b/doc/docs/2.3.x/deploy-manage/deploy/loki.md
@@ -74,6 +74,15 @@ loki-stack:
        
          ![Container logs](../images/grafana_user_logs.png)
 
+## Using Loki in Another Namespace
+Instead of deploying a local loki instance in your pachyderm namespace, you can configure pachyderm to use a loki running in another namespace. To do so, you must
+set `lokiHost` and `lokiPort`. You should also set `lokiDeploy: false` to prevent the chart from deploying a local loki instance.:
+```yaml
+pachd:
+  lokiDeploy: false
+  lokiHost: "<loki-namespace>.<loki-service-name>.svc.cluster.local"
+  lokiPort: 3100
+```
 
     
 ## References
--- a/doc/docs/2.3.x/deploy-manage/deploy/loki.md
+++ b/doc/docs/2.3.x/deploy-manage/deploy/loki.md
@@ -74,15 +74,6 @@ loki-stack:
        
          ![Container logs](../images/grafana_user_logs.png)
 
-## Using Loki in Another Namespace
-Instead of deploying a local loki instance in your pachyderm namespace, you can configure pachyderm to use a loki running in another namespace. To do so, you must
-set `lokiHost` and `lokiPort`. You should also set `lokiDeploy: false` to prevent the chart from deploying a local loki instance.:
-```yaml
-pachd:
-  lokiDeploy: false
-  lokiHost: "<loki-namespace>.<loki-service-name>.svc.cluster.local"
-  lokiPort: 3100
-```
 
     
 ## References
--- a/doc/docs/master/deploy-manage/deploy/loki.md
+++ b/doc/docs/master/deploy-manage/deploy/loki.md
@@ -74,7 +74,15 @@ loki-stack:
        
          ![Container logs](../images/grafana_user_logs.png)
 
-
+## Using Loki in Another Namespace
+Instead of deploying a local loki instance in your pachyderm namespace, you can configure pachyderm to use a loki running in another namespace. To do so, you must
+set `lokiHost` and `lokiPort`. You should also set `lokiDeploy: false` to prevent the chart from deploying a local loki instance.:
+```yaml
+pachd:
+  lokiDeploy: false
+  lokiHost: "<loki-namespace>.<loki-service-name>.svc.cluster.local."
+  lokiPort: 3100
+```
     
 ## References
 
--- a/src/internal/serviceenv/service_env.go
+++ b/src/internal/serviceenv/service_env.go
@@ -12,8 +12,6 @@ import (
 	dex_storage "github.com/dexidp/dex/storage"
 	"github.com/dlmiddlecote/sqlstats"
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
-	"github.com/pachyderm/pachyderm/v2/src/identity"
-	loki "github.com/pachyderm/pachyderm/v2/src/internal/lokiutil/client"
 	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 	etcd "go.etcd.io/etcd/client/v3"
@@ -23,6 +21,9 @@ import (
 	"google.golang.org/grpc"
 	kube "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+
+	"github.com/pachyderm/pachyderm/v2/src/identity"
+	loki "github.com/pachyderm/pachyderm/v2/src/internal/lokiutil/client"
 
 	"github.com/pachyderm/pachyderm/v2/src/client"
 	"github.com/pachyderm/pachyderm/v2/src/internal/backoff"
@@ -172,7 +173,12 @@ func InitServiceEnv(config *Configuration) *NonblockingServiceEnv {
 		env.dbEg.Go(env.initDirectDBClient)
 	}
 	env.listener = env.newListener()
-	if lokiHost, lokiPort := os.Getenv(env.config.LokiHostVar), os.Getenv(env.config.LokiPortVar); lokiHost != "" && lokiPort != "" {
+	if lokiHost, lokiPort := env.config.LokiHost, env.config.LokiPort; lokiHost != "" && lokiPort != "" {
+		env.lokiClient = &loki.Client{
+			Address: fmt.Sprintf("http://%s", net.JoinHostPort(lokiHost, lokiPort)),
+		}
+	} else if lokiHost, lokiPort := os.Getenv(env.config.LokiHostVar), os.Getenv(env.config.LokiPortVar); lokiHost != "" && lokiPort != "" {
+		env.config.LokiHost, env.config.LokiPort = lokiHost, lokiPort
 		env.lokiClient = &loki.Client{
 			Address: fmt.Sprintf("http://%s", net.JoinHostPort(lokiHost, lokiPort)),
 		}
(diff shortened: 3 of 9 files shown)


src/server/pps/server/worker_rc.go
```diff
@@ -153,11 +177,11 @@ func (kd *kubeDriver) workerPodSpec(options *workerOptions, pipelineInfo *pps.Pi
 		Name:  client.PPSPipelineNameEnv,
 		Value: pipelineInfo.Pipeline.Name,
 	}, {
-		Name:  "LOKI_SERVICE_HOST_VAR",
-		Value: kd.config.LokiHostVar,
+		Name:  "LOKI_SERVICE_HOST",
```
FahadBSyed, line 180: Both sets of variables could be included, but I figured that it would be easier for debugging to just include whatever was actually being used.

src/internal/serviceenv/service_env.go
```diff
@@ -172,7 +173,12 @@ func InitServiceEnv(config *Configuration) *NonblockingServiceEnv {
 		env.dbEg.Go(env.initDirectDBClient)
 	}
 	env.listener = env.newListener()
-	if lokiHost, lokiPort := os.Getenv(env.config.LokiHostVar), os.Getenv(env.config.LokiPortVar); lokiHost != "" && lokiPort != "" {
+	if lokiHost, lokiPort := env.config.LokiHost, env.config.LokiPort; lokiHost != "" && lokiPort != "" {
```
FahadBSyed, line 176: This is probably an ugly way of defining this logic, but it works 😆

src/internal/serviceenv/service_env.go
```diff
@@ -172,7 +173,12 @@ func InitServiceEnv(config *Configuration) *NonblockingServiceEnv {
...
+		env.lokiClient = &loki.Client{
+			Address: fmt.Sprintf("http://%s", net.JoinHostPort(lokiHost, lokiPort)),
+		}
+	} else if lokiHost, lokiPort := os.Getenv(env.config.LokiHostVar), os.Getenv(env.config.LokiPortVar); lokiHost != "" && lokiPort != "" {
```
jrockway, line 180: It turns out that I don't know why we bother doing this environment variable evaluation in pachd.  It seems that environment variables in a k8s manifest can refer to existing environment variables, so we can make LOKI_SERVICE_HOST (etc.) the only configuration we take, and then have helm generate:

```yaml
env:
  - name: LOKI_SERVICE_HOST
    value:  {{ if .lokiServiceHost }} {{ .lokiServiceHost }} {{ else }} $({{ release.Name | ... }}_LOKI_SERVICE_HOST)
```

So that will evaluate to a value like `literal-hostname.whatever` or `$(FOOBAR_LOKI_SERVICE_HOST)` depending on external loki or built-in loki.  Then pachd doesn't have to do this work.

Of course, we should check and see if that actually works.  The service environment variables we're depending on to exist are pretty undocumented.  (I read the code in the kubelet, though, and don't see why it wouldn't work.)

doc/docs/2.3.x/deploy-manage/deploy/loki.md
```diff
@@ -74,6 +74,15 @@ loki-stack:
        
          ![Container logs](../images/grafana_user_logs.png)
 
+## Using Loki in Another Namespace
```
jrockway, line 77: Should this actually be in the `master` directory instead of the `2.3.x` directory?  I don't actually know, just assuming.

src/server/pps/server/worker_rc.go
jrockway, line 180: Why does the worker need to know about loki?

jrockway (commented):
This looks good, but I have a couple suggestions.

Regarding the worker RC... I feel like k8s is going to complain if you duplicate LOKI_SERVICE_HOST and LOKI_SERVICE_PORT in the manifest, and it seems like there are cases where this can happen.  But I wonder if we can just remove them entirely; I don't know why the worker needs to know about Loki.

src/internal/serviceenv/service_env.go
FahadBSyed: Oh this is way better. The variable indirection seemed unnecessary but I figured maybe it was in case the service IP or port changes during runtime. That said, if it changes during runtime, it doesn't matter since we only set the lokiClient one time anyways.

doc/docs/2.3.x/deploy-manage/deploy/loki.md
FahadBSyed: Oh yeah I think you're right, good catch!

src/internal/serviceenv/service_env.go
jrockway: It's possible that it doesn't work, or whoever added this feature didn't know k8s could substitute environment variables references.  The docs say it can, but I've never tried it, and they don't say what syntax to use.  I just guessed 😂

src/server/pps/server/worker_rc.go
FahadBSyed: My initial thought was that it is so `get logs -p <pipeline name>` works. But I guess that it doesn't matter because its pachd that is making the call out to loki and filtering for the pipeline. Yeah .. so not sure why these variables are needed unless the sidecar also needs to talk to loki.

jrockway: Logs are sent to loki by promtail reading the container logs from disk.  Logs are read from loki over its http API via the GetLogsLoki call in pachd.  So it's not obvious to me why this was ever passed thru.

src/internal/serviceenv/service_env.go
FahadBSyed: It works!

Helm deployment.yaml
```yaml
        - name: LOKI_SERVICE_HOST
        {{- if .Values.pachd.lokiServiceHost }}
          value:  {{ .Values.pachd.lokiServiceHost }}
        {{ else }}
          value: $({{ snakecase .Release.Name | upper }}_LOKI_SERVICE_HOST)
        {{- end }}
```
```bash
kubectl exec -it pachd-699bd4cd9d-wmb5f -- env | grep LOKI
LOKI_SERVICE_PORT=3100
LOKI_SERVICE_HOST=10.100.47.53
LOKI_SERVICE_HOST_VAR=PACHYDERM_LOKI_SERVICE_HOST
LOKI_SERVICE_PORT_VAR=PACHYDERM_LOKI_SERVICE_PORT
LOKI_LOGGING=true
PACHYDERM_LOKI_PORT_3100_TCP_ADDR=10.100.47.53
PACHYDERM_LOKI_PORT_3100_TCP_PROTO=tcp
PACHYDERM_LOKI_PORT=tcp://10.100.47.53:3100
PACHYDERM_LOKI_SERVICE_PORT_HTTP_METRICS=3100
PACHYDERM_LOKI_SERVICE_HOST=10.100.47.53
PACHYDERM_LOKI_PORT_3100_TCP_PORT=3100
PACHYDERM_LOKI_SERVICE_PORT=3100
PACHYDERM_LOKI_PORT_3100_TCP=tcp://10.100.47.53:3100
```

src/server/pps/server/worker_rc.go
FahadBSyed: Also I deduped the `LOKI_SERVICE_HOST` and `LOKI_SERVICE_PORT` in the manifest

src/server/pps/server/worker_rc.go
```diff
@@ -153,11 +154,11 @@ func (kd *kubeDriver) workerPodSpec(options *workerOptions, pipelineInfo *pps.Pi
...
+		Value: kd.config.LokiHost,
 	}, {
-		Name:  "LOKI_SERVICE_PORT_VAR",
-		Value: kd.config.LokiPortVar,
+		Name:  "LOKI_SERVICE_PORT",
+		Value: kd.config.LokiPort,
```
jrockway, line 161: I think this is fine but be aware that we change the semantics here.  Before, LOKI_SERVICE_HOST_VAR (etc.) was set at RC creation time, and each pachyderm replica being created by the RC would look up the environment variable at startup time (since pachd did the resolution, not anything in k8s).  

Typically the variable this was pointing at was a variable based on services in the cluster (a normal service named `foo` causes FOO_SERVICE_... variables to be set at pod creation time, by the kubelet).  What this means is that each replica could have its own unique value if you edited the service between replica creations; i.e. it was 1.2.3.4:1234 for the first replica, then you edit the service to point at 10.0.0.123:1234, then a day passes and someone submits two jobs at the same time and autoscaling is turned on, so the second replica gets 10.0.0.123:1234 as the loki address.

With your change, this doesn't happen.  pachd looks this up at startup, and applies the same value to every RC it creates.

The old behavior is honestly confusing and hard to debug, since we go our of our way to make a new RC when the environment changes (or actually, I think whenever a new PPS master is elected/created), and this is better.  But, it is a change!

doc/docs/master/deploy-manage/deploy/loki.md
```diff
@@ -74,7 +74,15 @@ loki-stack:
...
          ![Container logs](../images/grafana_user_logs.png)
 
-
+## Using Loki in Another Namespace
+Instead of deploying a local loki instance in your pachyderm namespace, you can configure pachyderm to use a loki running in another namespace. To do so, you must
+set `lokiHost` and `lokiPort`. You should also set `lokiDeploy: false` to prevent the chart from deploying a local loki instance.:
+```yaml
+pachd:
+  lokiDeploy: false
+  lokiHost: "<loki-namespace>.<loki-service-name>.svc.cluster.local"
```
jrockway, line 83: oh btw, I like to have all k8s DNS addresses in documentation end with `.`, like `loki.monitoring.svc.cluster.local.`.  Do a quick search for "ndots" at k8s.af.  (TL;DR: ending the domain name with a period shows that the name is in the root namespace.  without it, the dns search path is consulted, resulting in extra queries that can overwhelm the DNS service in large clusters.  will this be a problem for our use of loki?  never.  but something to keep in mind, given the number of public postmortems involving ndots problems.)

jrockway (approved):
Love it!  Very nice work.

doc/docs/master/deploy-manage/deploy/loki.md
FahadBSyed: Sure, I can update the documentation to reflect that

src/server/pps/server/worker_rc.go
FahadBSyed: That's true. I am not sure how to best call this out, but it might be a good thing to mention in our support docs

PR #8281, opened 2022-10-13
set pachd's $GOMEMLIMIT from k8s requests or limits

jrockway:
This should help us take maximum advantage of the memory that the user has given to pachd.  If we have a lot of extra RAM, then we do GC less.  If we're close to running out of memory and being killed, we do GC more.  Of course, it's still possible to run out of memory.

At startup we log what we're doing with the memory limit, like:

```
memlimit: setting GOMEMLIMIT to 3.8 GiB, 95% of the kubernetes limit
memlimit: setting GOMEMLIMIT to 486 MiB, 95% of the kubernetes request
memlimit: using configured GOMEMLIMIT of 256 MiB
```

We also allow GOMEMLIMIT and GC_PERCENT to be overridden from the helm chart.

We used to set GOGC=50 as a way to make pachd use less memory at the cost of spending more time doing GC.  I changed that back to the default of 100 because it GOMEMLIMIT should do a better job balancing cost and time (increasing cost as heap size approaches GOMEMLIMIT).  Users can still change both if they feel like it.

Should it become necessary to disable completely, set it to math.MaxInt64 (9223372036854775807).

While testing, I noticed that the Minio client calls `runtime.GC()` for every file upload; meaning that if we're uploading files to Minio, it doesn't matter what our GC settings are, it is just running constantly.  Apparently some random open source user was running out of memory, added calls to `debug.FreeOSMemory()` in the function to reduce memory usage, and they accepted the PR.  A few months later, they realized it was a problem, and deleted the calls.  We started using the Minio client somewhere in the middle, so we were stuck with the broken version, which I've upgraded in this PR. 

Finally, we don't bother with GOMEMLIMIT in the worker.  The requests and limits users have set up on the worker don't really account for how much memory our Go code can use, so aren't a good configuration source.  (They set them to how much they want their own code to use, so we can't tell the Go runtime it's all available.)

Testing:
I have tested with:
*  no requests or limits set (k8s sets the limit to the amount of allocatable memory on the machine, which is fine with me though surprising)
* only limits (works; prints "using requests" even though it's using the limits; k8s sets the environment variable for requests to the value for limits unless you say "memory: 0" in the resource requests, in which case we print "using limits")
* only requests (works)
* only goMemLimit in the values (works)
* requests and limits (uses the request)
* requests and limits and goMemLimit
* limits and goMemLimit

I also did some performance testing, comparing master@e4c1988 (plus the minio client upgrades) with this PR.  I expected there to be a small impact on performance because GC would be called less, and indeed I did.  The performance test was this manifest:
```yaml
count: 30
modifications:
    - count: 1
      putFile:
          count: 10
          source: "random"
fileSources:
    - name: "random"
      random:
          sizes:
              - min: 9000000
                max: 10000000
                prob: 100
validator:
    frequency:
        count: 30
```
and each number is averaged across 3 runs.  I also set GODEBUG=gctrace=1 and STORAGE_MEMORY_CACHE_SIZE=400 (to keep some objects in memory that can't be collected, increasing the work the GC has to do).  I chose varying amounts of memory to see what happens when the GC needs to run frequently to stay under the limit, vs. not really needing to do GC at all.  Turns out it doesn't make much of a difference.  Varying the CPUs lets you see what happens when the runtime can't use spare cycles on another core to do GC; more of a difference.

The full results, generated by the script in https://github.com/pachyderm/pachyderm/pull/8319:

| CPU | Memory | Time on master | Time on this PR |
| --- | --- | --- | --- |
| 8 | 32G | 32.16s | 31.80s |
| 8 | 8G | 31.93s | 32.17s | 
| 8 | 1G | 32.66s | 32.23s |
| 1 | 32G | 65.42s | 62.16s |
| 1 | 8G | 65.61s |  60.67s  |
| 1 | 1G | 64.56s | 61.51s |

Basically, this doesn't make anything worse, and makes things better when CPU constrained.


.circleci/main.yml
```diff
@@ -59,6 +59,8 @@ jobs:
       TEST_RESULTS: /tmp/test-results
     steps:
       - checkout
+      - go/install:
```
jrockway, line 62: Apparently our pure go tests were using go 1.18, now they're using go 1.19 like everything else :/

PR #8284, opened 2022-10-14
proxy: set the default stream idle timeout to 1h instead of 10m

jrockway:
This is necessary for the load test.  I think we tested this with "pachctl wait commit" a while back and it never timed out, but that's also a unary RPC, so I'm not sure what the difference is.

How this was tested:
* Ran load test, after 10 minutes, got Unavailable "stream idle timeout" and SI in the envoy logs.
* Made this change, after 1200s, the load test completed.

Maybe we want an even longer timeout than 3600s though.


acohen4 (approved):
LGTM

PR #8289, opened 2022-10-17
CODEOWNERS Init

brendoncarroll:
This is the first attempt at introducing a CODEOWNERS file to the repo.

There are 3 groups referenced in the owners file: `@pachyderm/core @pachyderm/pfs @pachyderm/int`.  They match the project names in JIRA.

The proposal is to give default ownership to the CORE team, and then carve out exceptions to that general rule throughout the codebase.  e.g. `/src/server/pfs/ @pachyderm/pfs`.

This is intended to be a collaborative PR; I figured it would work better than a Google Doc for this use case.  Please add comments where you would like yourself or others to be added or removed.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -37,7 +37,7 @@
 
 # Integrations
 /src/integrations/                         @pachyderm/INT
-/src/integrations/connectors/snowflake/    @robert-uhl
+/src/integrations/connectors/snowflake/    @pachyderm/INT @albscui
 /src/templates/                            @pachyderm/INT
 
 # Internal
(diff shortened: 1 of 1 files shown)


molinamelendezj:
[CORE-420](https://pachyderm.atlassian.net/browse/CORE-420)

.github/CODEOWNERS
```diff
@@ -0,0 +1,71 @@
...
+/src/server/pfs/         @pachyderm/PFS
+
+/src/pps/                @brycemcanally @acohen4
+/src/server/pps/         @brycemcanally @acohen4
+
+/src/task/               @brycemcanally
+
+# Integrations
+/src/integrations/                         @pachyderm/INT
+/src/integrations/connectors/snowflake/    @robert-uhl
```
albscui, line 40: This should be `@pachyderm/INT` and `@albscui`

brendoncarroll: Line 40, 39 or both?

.github/CODEOWNERS
```diff
@@ -0,0 +1,71 @@
...
+/src/internal/collection/            @brycemcanally @acohen4 @robert-uhl
+/src/internal/clusterstate/          @jrockway @brycemcanally
+/src/internal/migrations/            @brendoncarroll
+/src/interna/consistenthashing/      @FahadBSyed
+/src/internal/dockertestenv/         @brendoncarroll
+/src/interna/minikubetestenv/        @acohen4
+/src/internal/obj/                   @pachyderm/PFS
+/src/interna/pachhash/               @pachyderm/PFS
+/src/interna/pachsql/                @albscui @brendoncarroll
+/src/interna/pachtmpl/               @albscui @brendoncarroll @pachyderm/INT
```
robert-uhl, line 57: These should all read `/src/interna*l*/`, right?

jrockway (approved):
Looks good.  Do we want to add pachyderm/INT to any s3 gateway areas?  Also, do we want to add any more owners to the examples directory?

brendoncarroll:
> we should just have teams
That would be the least complicated solution to the immediate problem (cross team reviews). 

But I also think we need to do a better job of assigning ownership (vision and maintenance) throughout the codebase.  IME individuals tend to improve things they own, in a way that groups do not.  I would rather work in a codebase with libraries which are crafted by some_one_ rather than tweaked without vision by a dozen different people as they went about their work.

PR #8291, opened 2022-10-18
[CORE-1152] Separate core & Jupyter extension jobs in CircleCI config

robert-uhl:


chainlink (approved):
This will work for now, but we'll want to revisit and include more folders. Singling our `src` is potentially confusing here.

PR #8292, opened 2022-10-18
Migration load testing

brycemcanally:
This PR implements load testing functionality that enables migration load testing. PPS and PFS load tests now persist internal state in a state file set between runs. The ID for the state file set is returned after a load test run and can be passed in as a parameter for a subsequent run to be used as the base state. A few noteworthy details about the semantics are:
- PFS state corresponds to the full set of commits generated by the run as well as prior runs if a base state was used for the run.
- PPS state corresponds to the source branch that was used for the run as well as the PFS state.
- Validation is rerun on the base state before generating the state for the run. The validation frequency defined in the spec applies to this step as well. For example, if the validator is set to run every 5 commits with a base of 10 commits, the validation will run on the 5th and 10th commit, then every 5th commit in the run.


src/internal/pfsload/validator.go
```diff
@@ -66,12 +72,21 @@ func (v *Validator) Validate(client Client, commit *pfs.Commit) (retErr error) {
 		if ci.Commit.Branch.Repo.Type != pfs.UserRepoType || ci.Origin.Kind == pfs.OriginKind_ALIAS {
 			return nil
 		}
-		return validate(client, ci.Commit, v.hash)
+		hash, err := computeCommitHash(client, ci.Commit)
+		if err != nil {
+			return err
+		}
+		if !bytes.Equal(hash, v.hash) {
```
FahadBSyed, line 79: Any reason why this was moved out of computeCommitHash? Do you want potentially reuse computeCommitHash elsewhere?

src/internal/ppsload/pipeline.go
```diff
@@ -93,3 +116,32 @@ func createPipeline(pachClient *client.APIClient, repo *pfs.Repo, inputRepos []s
 	)
 	return errors.EnsureStack(err)
 }
+
+const stateFileName = "state"
+
+func serializeState(pachClient *client.APIClient, state *State) (string, error) {
```
FahadBSyed, line 122: This code looks the same as serialize/deserialize in pfs but I guess it can't be cleanly extracted into a single function since `State` is defined differently in `pfs` and `pps`

FahadBSyed (approved):
Approved with a few nonblocking questions

src/internal/pfsload/validator.go
brycemcanally: yes, I think it will be more useful for later work and it makes more sense to me to separate the compute and check functionality.

src/internal/ppsload/pipeline.go
brycemcanally: Yeah, we have the same issue with the tasks in the task service. I am very open to smarter ways to handle this.

PR #8295, opened 2022-10-19
[CORE-1096] ListFile pagination

armaanv:
This PR adds three new parameters to ListFile for pagination of the results -

    Number -number of results to return
    paginationMarker - marker to start the page of results from. This should be a File object.
    reverse - flag to indicate that files should be returned in reverse order i.e. higher to lower filepaths in lexicographical order


src/pfs/pfs.proto
```diff
@@ -376,7 +376,16 @@ message ListFileRequest {
   // repo, the commit/branch, and path prefix of files we're interested in
   // If the "path" field is omitted, a list of files at the top level of the repo
   // is returned
-  File file = 1;
+  File parentDirectory = 1;
```
brycemcanally, line 379: Changing the name of this field would be a breaking change and this isn't an accurate name regardless of the above comment. This field can refer to a regular file.

src/server/pfs/server/driver_file.go
```diff
@@ -263,13 +268,38 @@ func (d *driver) listFile(ctx context.Context, file *pfs.File, cb func(*pfs.File
...
-			return cb(fi)
+			fis = append(fis, fi)
+			return nil
 		}
 		return nil
-	})
+	}); err != nil {
+		return errors.EnsureStack(err)
+	}
+	for i, fi := range fis {
```
brycemcanally, line 291: Doing this generally is going to be problematic for large lists. We need to change this algorithm to scale. I think we could do this by separating the forward from the reverse case, and have a separate callback that gets used for each. The forward case would just execute the callback above normally until the number has been emitted. The reverse case would collect a list up to a fixed size until the iteration completes and then return the number of files from that once the iteration completes. If the reverse list size gets too large, then we should return an error to prevent clients from OOM killing pachd. Also, it is worth noting the reverse performance can be O(n) towards the end of the lexicographical ordering of the file paths.

src/pfs/pfs.proto
armaanv: I didn't realize this would be a breaking change. I get your point about the name not being accurate though. I've reverted the name change

PR #8299, opened 2022-10-20
Check for proxy.host if proxy is enabled

tybritten:


jrockway (approved):
I've seen the thread about `proxy.host` but I don't have any objections.  I cannot reason about what would happen if you omit this.  I don't think console or auth would work, though, and that seems pretty bad.

tybritten:
technically you can omit it if you're doing community edition (and we omit it in CI, hence all the failures). Maybe it needs to be if proxy.enabled & pachd.activateauth or something

PR #8309, opened 2022-10-21
Improve random data generation speed

jrockway:
This adjusts randutil.Bytes/BytesReader to use a faster reduction algorithm than `random.Intn`.  random.Intn does a very slow divide to avoid any bias; we do `len(letters)*uint16(byte) >> 8` which is much faster for the CPU to actually execute, and doesn't introduce too much bias.  See https://lemire.me/blog/2016/06/30/fast-random-shuffling for details.  I added some tests to make sure each letter does show up in the output with enough rolls of the dice.

Before this change, BytesReader could generate data at about 70MB/s.  Now it does 530MB/s.  (Without doing any mapping from bytes to letter, you can get about 630MB/s.)

I also got rid of the `miscutil.MinInt64` function, and made `miscutil.Min` generic.

One downside is that some tests depend on randutil.Bytes always producing the same output; this algorithm changes the output for a given seed.  I think it's just one set of tests, though, so I ported the old code into that test for it to use in isolation.

### DIFF
--- a/src/internal/miscutil/math.go
+++ b/src/internal/miscutil/math.go
@@ -1,15 +1,9 @@
 package miscutil
 
-// Min returns the smaller of a and b.
-func Min(a, b int) int {
-	if a < b {
-		return a
-	}
-	return b
-}
+import "golang.org/x/exp/constraints"
 
-// MinInt64 returns the smaller of a and b.
-func MinInt64(a, b int64) int64 {
+// Min returns the smaller of a and b.
+func Min[T constraints.Ordered](a, b T) T {
 	if a < b {
 		return a
 	}
--- a/src/internal/randutil/rand_test.go
+++ b/src/internal/randutil/rand_test.go
@@ -24,6 +24,11 @@ func BenchmarkReader(b *testing.B) {
 	}
 }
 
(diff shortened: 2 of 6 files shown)


src/internal/miscutil/math.go
```diff
@@ -1,15 +1,9 @@
...
-		return a
-	}
-	return b
-}
+import "golang.org/x/exp/constraints"
 
-// MinInt64 returns the smaller of a and b.
-func MinInt64(a, b int64) int64 {
+// Min returns the smaller of a and b.
+func Min[T constraints.Ordered](a, b T) T {
```
brendoncarroll, line 6: 🧞‍♂️

src/internal/randutil/rand_test.go
```diff
@@ -23,3 +23,27 @@ func BenchmarkReader(b *testing.B) {
 		require.NoError(b, err)
 	}
 }
+
+func testLetters(t *testing.T, buf []byte) {
```
brendoncarroll, line 27: This looks like it works probabilistically based on the size of buf. Maybe leave a comment about that?

brendoncarroll (approved):
LGTM

PR #8315, opened 2022-10-25
[CORE-549]remove duplicae pps env variables

djanicekpach:
## Problem Summary:
https://pachyderm.atlassian.net/browse/CORE-549

Two environment variables, `PPS_PROJECT_NAME` and `PPS_PIPELINE_NAME` are being set in multiple places causing warnings from Kubernetes. You can see this issue in the pachd logs and in the pod description:
image

## Fix:
`getStorageEnvVars` and `getWorkerOptions` set these variables even though they are also set in the `commonEnv`.

These functions are only called in this file, and both are appended to commonEnv before usage, so deleting them _shouldn't_ have any side effects, and testing looks good, but this seems like the part to double check.

## Testing:

Re-ran `example/opencv` after the build and confirmed the variables are only listed once, and there was no warning in the pachd logs.

The pipeline passed, but only after a retry. I'll look more into the failure, but I believe this was flake in the load test.

After:
image


acohen4 (approved):
LGTM!

PR #8317, opened 2022-10-25
unused code part II

seslattery:


jrockway (approved):
This is very surprising but probably true.  I wonder what the "cars table" was supposed to be for.  Was added in a commit called "git mv" :D

lukemarsden (approved):

PR #8320, opened 2022-10-26
Add 'make docker-build-amd' to speed up local development

acohen4:


jrockway (approved):
Good idea.  I think it's going to be a big pain to maintain all of these goreleaser manifests that are basically the same but slightly different.  jsonnet could be very good here.

PR #8323, opened 2022-10-27
[CORE-1097] ListDatum pagination

armaanv:
This PR adds three new parameters to ListDatum for pagination of the results - 
- Number -number of results to return
- paginationMarker - marker to start the page of results from. This should be a datumID. 
- reverse - flag to indicate that datums should be returned in reverse order i.e. higher to lower datumIDs

The tests exercise getting the first, next, previous and last page of results.


src/server/pps/server/api_server.go
```diff
@@ -1065,23 +1066,55 @@ func (a *apiServer) InspectDatum(ctx context.Context, request *pps.InspectDatumR
...
+		}); err != nil {
+			return errors.EnsureStack(err)
+		}
+	}
+	number := request.Number
+	if number == 0 {
+		number = math.MaxInt64
 	}
-	return a.collectDatums(server.Context(), request.Job, func(meta *datum.Meta, _ *pfs.File) error {
+	for i, meta := range metas {
```
brycemcanally, line 1092: This is a similar problem to what I mentioned in the list file PR. I think the same strategy I suggested there would work here as well.

src/server/pps/server/pipeline_controller_test.go
```diff
@@ -36,7 +37,7 @@ func ppsMasterHandles(t *testing.T) (*mockStateDriver, *mockInfraDriver, *testpa
 		GetPachClient: func(ctx context.Context) *client.APIClient {
 			return mockEnv.PachClient.WithCtx(ctx)
 		},
-		Config:      newConfig(t),
+		Config:      *serviceenv.ConfigFromOptions(),
```
brycemcanally, line 40: What is going on here?

brycemcanally (approved):
lgtm

PR #8329, opened 2022-10-27
[CORE-1198] loki timeout and line num change

armaanv:


src/internal/lokiutil/client/client.go
```diff
@@ -64,6 +64,9 @@ func (c *Client) doQuery(ctx context.Context, path string, query string, quiet b
 }
 
 func (c *Client) doRequest(ctx context.Context, path, query string, quiet bool, out interface{}) error {
+	// Loki requests can hang if the size of the log lines is too big, so we set a timeout
+	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
+	defer cancel()
```
jrockway, line 69: I think it would be best to apply this timeout only to the debug dump code.  This code here is also called in interactive requests, and since a user is at the terminal in that case, I'd rather let them choose how long to wait for logs.  If there's ever a case where the Loki query they do is slow but actually completable, I don't want to give up early.  For the debug dumps where we have to, in an unattended manner, get a ton of logs, it makes sense to give up if we think it's hanging.

Thus I'd suggest that you set the timeout in src/server/debug/server/server.go in the collectPipelineDumpFunc, in the section that reaches out to loki.  Basically there's this code:

```go
                if s.env.Config().LokiHost != "" {
                        if err := s.forEachWorkerLoki(ctx, pipelineInfo, func(pod string) error {
                                workerPrefix := join(podPrefix, pod)
                                if len(prefix) > 0 {
                                        workerPrefix = join(prefix[0], workerPrefix)
                                }
                                userPrefix := join(workerPrefix, client.PPSWorkerUserContainerName)
                                if err := s.collectLogsLoki(ctx, tw, pod, client.PPSWorkerUserContainerName, userPrefix); err != nil {
                                        return err
                                }
                                sidecarPrefix := join(workerPrefix, client.PPSWorkerSidecarContainerName)
                                return s.collectLogsLoki(ctx, tw, pod, client.PPSWorkerSidecarContainerName, sidecarPrefix)
                        }); err != nil {
                                name := "loki"
                                if len(prefix) > 0 {
                                        name = join(prefix[0], name)
                                }
                                return writeErrorFile(tw, err, name)
                        }
                }
```
I'd set the timeout for each call of the callback passed to forEachWorkerLoki.  1 minute to get both user logs and storage logs.

jrockway (changes requested):
Sorry that I didn't clarify this up front.  I think it's too aggressive to set the timeout for ALL loki calls; rather we should only do it during the debug dump.  This lets interactive users wait a long time, but ensures that the debug dump always makes progress.

src/internal/lokiutil/client/client.go
armaanv: Makes sense. I've moved the timeout to the debug dump server

jrockway (approved):
Looks good!

PR #8330, opened 2022-10-28
[CORE-1066] Refactor PachctlBashCommand

robert-uhl:
The existing `PachctlBashCommand` has implicit semantics: it only runs in a test; it can panic or fail the test; the config file is always stored under a name based on the test; state matters between runs.

This update adds new `PachctlBash` type which can be used to run scripts using `pachctl` in Bash. It can be run independently of a test; it always returns errors and never panics; the config file is stored in a caller-configurable path; state is preserved as long as the instance is not closed. A `Cmd` type which embeds `exec.Cmd` to provide easy access to buffered stdout/stdin is also provided.


src/internal/testutil/pachctl.go
```diff
@@ -0,0 +1,185 @@
...
+	b, err := io.ReadAll(cmd.stderr)
+	if err != nil {
+		panic(err)
+	}
+	return string(b)
+}
+
+func newCmd(ctx context.Context, name string, args []string, stdin io.Reader) Cmd {
+	var cmd Cmd
+	cmd.Cmd = exec.CommandContext(ctx, "/bin/bash", args...)
```
smalyala, line 154: Should "/bin/bash" be `name`?

src/internal/testutil/pachctl.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return errors.Wrap(err, "could not create new template")
+	}
+	if err := tmpl.Execute(w, data); err != nil {
+		return errors.Wrap(err, "could not execute script template")
+	}
+	return nil
+}
+
+// bashTemplate interprets scriptTemplate as an indented Go template for a Bash script; it returns an io.Reader containing
```
smalyala, line 112: Is this comment unfinished?

src/internal/testutil/cmds.go
```diff
@@ -133,51 +136,37 @@ func BashCmd(cmd string, subs ...string) *exec.Cmd {
 	res.Env = os.Environ()
 	return res
 }
-func PachctlBashCmd(t *testing.T, c *client.APIClient, cmd string, subs ...string) *exec.Cmd {
-	t.Helper()
 
-	buf := new(bytes.Buffer)
-	_, err := buf.ReadFrom(bashCmd(cmd, subs...))
-	require.NoError(t, err)
+func subsToTemplateData(subs ...string) (map[string]string, error) {
```
smalyala, line 140: Maybe we could have `bashCmd()` call this function instead of repeating the same steps.

smalyala (approved):
LGTM! Left a few comments

src/internal/testutil/pachctl.go
robert-uhl: 🤦‍♂️

src/server/pfs/fuse/server_test.go
```diff
@@ -745,7 +745,7 @@ func TestDeletingMountedRepo(t *testing.T) {
 		_, err := put("_mount", b)
 		require.NoError(t, err)
 
-		env.PachClient.DeleteProjectBranch(pfs.DefaultProjectName, "repo", "b1", false)
+		require.NoError(t, env.PachClient.DeleteProjectBranch(pfs.DefaultProjectName, "repo", "b1", false))
```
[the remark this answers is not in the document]
robert-uhl: It triggered `golangci-lint`, so I fixed it. Which raises the question of why the scripts which run `golangci-lint` didn’t error … maybe I should open another story for that, and then include this fix there.

armaanv (approved):

FahadBSyed (approved):

molinamelendezj (approved):

PR #8332, opened 2022-10-31
[2.4.x port][CORE-1198] Add a timeout to loki requests from debug dump (#8329)

armaanv:
* loki timeout and line num change

* move timout to debug dump server


molinamelendezj (approved):

jrockway (approved):
Good catch!

PR #8333, opened 2022-10-31
add fail body to curl helium workspace create

molinamelendezj:


seslattery (changes requested):
For dumb reasons, you'll want `-F disableNotebooks="True"`  I think :/

seslattery (approved):

djanicekpach (approved):

PR #8334, opened 2022-11-01
use unique repo name to deflake proxy tests

armaanv:
I noticed a number of test failures which were due to conflicting repo names. This should fix that case

### DIFF
--- a/src/server/proxy_test.go
+++ b/src/server/proxy_test.go
@@ -33,7 +33,6 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/minikubetestenv"
 	"github.com/pachyderm/pachyderm/v2/src/internal/require"
 	"github.com/pachyderm/pachyderm/v2/src/internal/testutil"
-	tu "github.com/pachyderm/pachyderm/v2/src/internal/testutil"
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
 )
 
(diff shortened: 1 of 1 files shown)


src/server/proxy_test.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/minikubetestenv"
 	"github.com/pachyderm/pachyderm/v2/src/internal/require"
 	"github.com/pachyderm/pachyderm/v2/src/internal/testutil"
+	tu "github.com/pachyderm/pachyderm/v2/src/internal/testutil"
```
jrockway, line 36: I think this is getting imported twice.  Just call it "testutil" below and don't import the aliased version.

src/server/proxy_test.go
```diff
@@ -96,10 +97,11 @@ func proxyTest(t *testing.T, httpClient *http.Client, c *client.APIClient, secur
 	// Test GRPC API.
 	t.Run("TestGRPC", func(t *testing.T) {
 		require.NoErrorWithinTRetry(t, 60*time.Second, func() error {
-			if err := c.CreateProjectRepo(pfs.DefaultProjectName, "test"); err != nil {
+			testRepo := tu.UniqueString("TestGRPC")
+			if err := c.CreateProjectRepo(pfs.DefaultProjectName, testRepo); err != nil {
 				return errors.Errorf("create repo: %w", err)
```
jrockway, line 102: It would be good if these errors contained the name of the repo, so when the test fails you can look at the pachd logs and correlate them with this test run.  (One more case below.)

jrockway (approved):
Excellent find.

PR #8340, opened 2022-11-02
[CORE-1206] Fix list datum input with auth

brycemcanally:
This PR fixes the issue with the list datum input functionality with auth enabled. The auth token was not being used by the background worker that was creating the datums based on the request. This PR adds propagation of auth tokens through the datum creation code.


acohen4 (approved):
LGTM

jrockway (approved):

PR #8341, opened 2022-11-02
[Mount Server] Prevent mounting non-existent branch in a read-only mount

smalyala:
Prior to this change, mounting a non-existent branch "works", creating the corresponding repo folder in the pfs mount directory. Now, throw an error, except in the case where the mount is a read-write mount (since users can use this to add data on a new branch).

Ticket: https://linear.app/pachyderm/issue/INT-559/mounting-nonexistent-branch-silently-fails


chainlink (approved):
LGTM

PR #8344, opened 2022-11-04
[Mount Server] Mount latest non-alias commit on branch

smalyala:
As of right now, we mount the [latest commit](https://github.com/pachyderm/pachyderm/blob/master/src/server/pfs/fuse/loopback.go#L744) on a repo's branch. This can be an alias commit, which complicates user experience (see the linked ticket for more info on how). This PR changes mounting behavior to only mount the latest non-alias commit.

Ticket: https://linear.app/pachyderm/issue/INT-754/obfuscate-alias-commits-from-notebooks

### DIFF
--- a/src/server/pfs/fuse/server.go
+++ b/src/server/pfs/fuse/server.go
@@ -27,6 +27,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/client"
 	"github.com/pachyderm/pachyderm/v2/src/internal/config"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/errutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/grpcutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/ppsutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/progress"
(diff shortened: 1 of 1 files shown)


src/server/pfs/fuse/server.go
```diff
@@ -1401,10 +1391,31 @@ func mountingState(m *MountStateMachine) StateFn {
 			Write:    m.Mode == "rw",
 		}
 		m.manager.root.branches[m.Name] = m.Branch
-	}()
+
+		// Get the latest non-alias commit on branch
+		// TODO: notebooks support
```
chainlink, line 1396: What's the TODO here?

smalyala: Just a reminder to myself to update this code when I add projects support to the mount server. Here, it'll be passing the appropriate project in the `InspectBranch` rpc.

src/server/pfs/fuse/server.go
```diff
@@ -1180,29 +1180,18 @@ func (m *MountStateMachine) RefreshMountState() error {
...
 	}
+	m.LatestCommit = commitInfos[0].Commit.ID
+
 	// reverse slice
 	for i, j := 0, len(commitInfos)-1; i < j; i, j = i+1, j-1 {
 		commitInfos[i], commitInfos[j] = commitInfos[j], commitInfos[i]
 	}
 
-	// iterate over commits in branch, counting how many are behind LatestCommit
+	// iterate over non-alias commits in branch, calculating how many commits behind LatestCommit ActualMountedCommit is
```
msteffen, line 1194: It looks like the comment changed, but the code didn't—just checking that that's intentional? This always skips alias commits?

msteffen (approved):
Overall LGTM, with some random nits.

One big question, though—are we cutting 2.4 from `master` or from the `2.4.x` branch? If it's the latter, we should change this PR (or prepare another) to go into that branch.

src/server/pfs/fuse/server.go
smalyala: Yep it's intentional. It's to help clarify that the `ListCommit` rpc above returns non-alias commits by default.

PR #8346, opened 2022-11-07
update pgbouncer image release

molinamelendezj:


.circleci/config.yml
```diff
@@ -29,10 +29,12 @@ parameters:
...
     type: boolean
     default: false
-  pgbouncer-image-version:
+  pgbouncer-base-image-version:
     type: string
     default: "1.16.1"
-
+  pgbouncer-image-tag:
+    type: string
+    default: "1.16.1" #may not always be in sync with pgbouncer-base-image-version, for entrypoint.sh wrapper updates
```
jrockway, line 37: I'd make this note a paragraph style comment (between lines 36 and 37, spanning multiple lines).  But actually, why isn't entrypoint.sh something we just vendor in etc?

jrockway (approved):

.circleci/config.yml
molinamelendezj: would be better place for it. ill move it.

seslattery (approved):

PR #8348, opened 2022-11-08
[CORE-1072] Plumb contextual project through pachctl commands

robert-uhl:


src/server/pfs/cmds/cmds.go
```diff
@@ -74,7 +75,16 @@ or type (e.g. csv, binary, images, etc).`,
...
+	cfg, err := config.Read(false, true)
+	if err != nil {
+		return nil, err
+	}
+	_, pachCtx, err := cfg.ActiveContext(true)
+	if err != nil {
+		return nil, err
+	}
+	var description string
+	project := pachCtx.Project
```
acohen4, line 87: what if the Project isn't set in the context?

src/server/pfs/cmds/cmds.go
```diff
@@ -52,7 +53,7 @@ const (
 )
 
 // Cmds returns a slice containing pfs commands.
-func Cmds() []*cobra.Command {
+func Cmds() ([]*cobra.Command, error) {
```
acohen4, line 56: why not just panic if we hit an error?

acohen4 (commented):
Couple Qs

src/server/pfs/cmds/cmds.go
robert-uhl: [config.ActiveContext](https://github.com/pachyderm/pachyderm/blob/f9a8dc63e8d32a5fbe198ac97de5d02c799d5419/src/internal/config/config.go#L70-L72) (called on line 82) sets Project in the context.

src/server/pfs/cmds/cmds.go
robert-uhl: Better to return an error and let the caller determine how to handle the error. In this case the [caller](https://github.com/pachyderm/pachyderm/blob/f9a8dc63e8d32a5fbe198ac97de5d02c799d5419/src/server/cmd/pachctl/cmd/cmd.go#L320) _does_ just return an error, but it could return an error too, and print a useful-to-the-user error message.

I could add that to this PR …

acohen4: It seems odd to me that we're only applying it to pfs and pps cmds in this PR - besides updating a potentially preferable signature, I don't think it's providing value in this PR. I would be more in favor of just panicking or refactoring more broadly. panicking seems fine in this context because in a CLI the program terminates anyway.
WDYT?

robert-uhl: I’d rather refactor more broadly than panic. IMHO a panic indicates a bug, such as a nil memory access. Getting a stack trace can be useful to a developer with a panic; getting a stack trace is useless to en end user. In this case an error would be returned if the user’s `~/.pachyderm/config` has a syntax error — that is not necessarily a bug, but could be caused by the user manually editing the JSON. I think it would be more helpful to print _that_ then a panic trace.

acohen4: Ok I would prefer to leave it out of this Pr, but if you want to commit as is I encourage prioritizing a ticket to finish the job.

acohen4 (approved):

src/server/pfs/cmds/cmds.go
```diff
@@ -74,7 +75,16 @@ or type (e.g. csv, binary, images, etc).`,
 	}
 	commands = append(commands, cmdutil.CreateDocsAliases(repoDocs, "repo", " repo$", repos))
 
-	var description, project string
+	cfg, err := config.Read(false, true)
```
brycemcanally, line 78: Could we do this at a higher level, maybe even the main function? It seems like it would be better to do this in one place and pass around a config and / or context to be used by the commands. That would also let us avoid having to handle errors when creating the commands.

robert-uhl: Refactoring for _just_ PFS & PPS commands turned out to be extremely simple. I started to do the same for the rest of the codebase (e.g. for transactions), but it gets pretty hairy because some parts of the code expect a read-only config, others a writable one, others expect an error when getting the active context and others don’t. Resolving all that is a bigger story. How do you feel about me opening an issue on the backlog for that?

brycemcanally: Yeah, that makes sense.

brycemcanally (approved):
lgtm

src/server/pfs/cmds/cmds.go
robert-uhl: Thanks, I created CORE-1281 for that.

molinamelendezj (approved):

PR #8349, opened 2022-11-08
testutils and password tests

djanicekpach:
### Tickets: 

CORE-1100
CORE-714

### Summary

I'm adding project awareness to auth tests. This doesn't handle adding new tests, but sets us up to do so in the future.

I'm adding tests to check for weird characters(specifically ") in passwords by just making it the default pg password in the auth tests.

I added project awareness to the makefile for the word_count example as a bonus to help with tesitng as that project moves forward.

**I changed cmds.go since I think there was a typo, but let me know if I misunderstood.**

### Testing:

Ran both the auth_test.go and auth_integration_test.go locally and confirmed all passing.
Ran the word count example and confirmed passing.

### DIFF
--- a/examples/word_count/Makefile
+++ b/examples/word_count/Makefile
@@ -3,6 +3,7 @@ DOCKER_ACCOUNT := pachyderm
 CONTAINER_NAME := example-wordcount
 CONTAINER_VERSION := 2.0.3
 CONTAINER_TAG := $(DOCKER_ACCOUNT)/$(CONTAINER_NAME):$(CONTAINER_VERSION)
+PROJECT ?= default
 
 docker-image:
 	@docker buildx build --platform linux/amd64 -t $(CONTAINER_TAG)-amd64 .
@@ -15,20 +16,20 @@ docker-image:
 
 
 wordcount:
-	pachctl create repo urls
-	cd data && pachctl put file urls@master -f Wikipedia
-	pachctl create pipeline -f pipelines/scraper.json
-	pachctl create pipeline -f pipelines/map.json
-	pachctl create pipeline -f pipelines/reduce.json
+	pachctl create repo urls --project $(PROJECT)
+	cd data && pachctl put file urls@master -f Wikipedia --project $(PROJECT)
+	pachctl create pipeline -f pipelines/scraper.json --project $(PROJECT)
+	pachctl create pipeline -f pipelines/map.json --project $(PROJECT)
+	pachctl create pipeline -f pipelines/reduce.json --project $(PROJECT)
 
 clean:
-	pachctl delete pipeline reduce
-	pachctl delete pipeline map
-	pachctl delete pipeline scraper
-	pachctl delete repo urls
-	pachctl delete repo scraper	
-	pachctl delete repo map	
-	pachctl delete repo reduce
+	pachctl delete pipeline reduce --project $(PROJECT)
+	pachctl delete pipeline map --project $(PROJECT)
+	pachctl delete pipeline scraper --project $(PROJECT)
+	pachctl delete repo urls --project $(PROJECT)
+	pachctl delete repo scraper	--project $(PROJECT)
+	pachctl delete repo map	--project $(PROJECT)
+	pachctl delete repo reduce --project $(PROJECT)
 
 
 .PHONY:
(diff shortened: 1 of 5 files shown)


examples/word_count/Makefile
```diff
@@ -15,20 +16,20 @@ docker-image:
 
 
 wordcount:
-	pachctl create repo urls
-	cd data && pachctl put file urls@master -f Wikipedia
-	pachctl create pipeline -f pipelines/scraper.json
-	pachctl create pipeline -f pipelines/map.json
-	pachctl create pipeline -f pipelines/reduce.json
+	pachctl create repo urls --project $(PROJECT)
```
jrockway, line 19: Realistically, it feels like we got the UI wrong for pachctl if every command is going to need a `--project` flag.  Kubernetes puts the namespace in the config file.  We probably should too.  Nobody would tolerate typing `kubectl get pods --namespace my-app` every single time.

src/internal/testutil/auth.go
```diff
@@ -144,16 +144,16 @@ func CommitCnt(t *testing.T, c *client.APIClient, repo *pfs.Repo) int {
...
-func PipelineNames(t *testing.T, c *client.APIClient) []string {
+// ListPipeline in the specified project. "" project name means get all pipelines.
+func PipelineNames(t *testing.T, c *client.APIClient, project string) []string {
 	t.Helper()
 	ps, err := c.ListPipeline(false)
 	require.NoError(t, err)
 	result := make([]string, len(ps))
 	for i, p := range ps {
-		result[i] = p.Pipeline.Name
+		if project == "" || project == p.Pipeline.Project.Name {
```
jrockway, line 154: Where are we documenting that `""` means "all projects"?

I recommend being paranoid when accessing nested fields in protocol buffers messages; p.GetPipeline().GetProject().GetName().  This returns `""` if any of p, pipeline, or project are `nil`.   The former panics.

src/internal/testutil/auth.go
```diff
@@ -144,16 +144,16 @@ func CommitCnt(t *testing.T, c *client.APIClient, repo *pfs.Repo) int {
...
+			result[i] = p.Pipeline.Name
```
jrockway, line 155: I am not convinced that this algorithm is correct.  If only a few pipelines survive the filtering, then random elements are equal to `""`.  Prefer:

```go
var result []string
for _, p := range ps {
  if project == ... {
      result = append(result, p.GetPipeline().GetName())
  }
}
return result
```

src/server/auth/server/testing/auth_integration_test.go
```diff
@@ -32,6 +32,11 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/pps"
 )
 
+// works with global.postgresql.postgresqlAuthType: md5, which is default in values file.
+var default_test_opts = minikubetestenv.WithValueOverrides(map[string]string{
```
jrockway, line 36: Idiomatic Go doesn't use `snake_case` variable names.  Name this defaultTestOptions.  (Abbreviation is also of unclear value, unless the type name is Opt or Opts, maybe.  But it's "Options" here.)

jrockway (changes requested):
Looking good.  A few comments:

examples/word_count/Makefile
robert-uhl: We store the current project in the config file, too: the user may set the default project with `pachctl config update context --project $whatever`.

djanicekpach: We could update the default with `pachctl config update context --project`, but I don't know if we want to update someones default project in an example makefile. It probably wouldn't be an issue though. If you prefer setting it in the context let me know and I'll do that.

src/server/auth/server/testing/auth_integration_test.go
djanicekpach: woops, I was thinking in the wrong language. fixed.

src/internal/testutil/auth.go
djanicekpach: good catch, fixed.

src/internal/testutil/auth.go
djanicekpach: The function comment mentioned it, but initially I thought I was modelling this off how ListJobs was working. Looking again I misunderstood. I've changed it so that  "" means default since that seems to be the standard behavior elsewhere.

jrockway (approved):
Looks good.

PR #8350, opened 2022-11-09
[CORE-1169] ListJob pagination

armaanv:


src/server/pps/server/api_server.go
```diff
@@ -760,6 +767,16 @@ func (a *apiServer) listJob(
 	jobs := a.jobs.ReadOnly(ctx)
 	jobInfo := &pps.JobInfo{}
 	_f := func(string) error {
+		if number == 0 {
+			return errutil.ErrBreak
+		}
+		if paginationMarker != nil {
+			createdAt := time.Unix(int64(jobInfo.Created.GetSeconds()), int64(jobInfo.Created.GetNanos())).UTC()
+			fromTime := time.Unix(int64(paginationMarker.GetSeconds()), int64(paginationMarker.GetNanos())).UTC()
+			if createdAt.Equal(fromTime) || !reverse && createdAt.After(fromTime) || reverse && createdAt.Before(fromTime) {
```
acohen4, line 776: so we don't want to include records that were created at `fromTime`?

src/server/pps/server/api_server.go
```diff
@@ -760,6 +767,16 @@ func (a *apiServer) listJob(
...
+				return nil
```
acohen4, line 777: should this return errutil.ErrBreak?

src/server/pps/server/api_server.go
acohen4: nvm that makes sense based on the usage

acohen4 (approved):
LGTM

src/server/pps/server/api_server.go
armaanv: yeah, the usage pattern is that users will get the first page of results and then request the next page by specifying the paginationTimestamp as the createdAt time of the last job in the first page. We don't want to return that last job again so we exclude matching timestamps

src/server/pps/server/api_server.go
armaanv: hmm, I think you're right. errutil.ErrBreak will probably save some iterations over records that we know don't need to be included

PR #8352, opened 2022-11-09
[CORE-1225] [CORE-1158] Fix input reordering with cross inputs

brycemcanally:
This PR fixes an issue with cross inputs where the inputs could get reordered in certain cases. Reordering the inputs would change the datum hashes, so the input order needs to be preserved when computing the datums. 

This PR also includes a few subtle changes that address issues I noticed while fixing the input reordering issue:
- Ensure that every file set creates at least one shard (empty file sets did not).
- Sort the inputs for list datum input as well as pipeline creation. Not sorting the inputs for list datum input meant that different datums could be created for the same input used during pipeline creation.
- Only use the meta commit in datum creation when it is finished, rather than finishing. This prevents a performance issue where the meta commit file set has a large number of primitive file sets prior to compaction.

This PR also adds an upgrade load test to CI which uses the new upgrade load testing functionality.


acohen4 (approved):
LGTM!

PR #8354, opened 2022-11-09
Make ModifyRoleBinding project aware

albscui:
This PR extends `ModifyRoleBinding` to be project-aware, by checking for the `PROJECT_MODIFY_BINDINGS` permission on the subject.

Notable changes:

- Moved shared logic across `[CheckClusterIsAuthorizedInTransaction, CheckProjectIsAuthorizedInTransaction, CheckRepoIsAuthorizedInTransaction]` to `checkResourceIsAuthorizedInTransaction`.
- Implemented "inheritance" in our resource hierarchy for the set of `Check<Resource>IsAuthorizedInTransaction()` APIs. This is achieved by passing the requested permissions up a resource's parent first. These calls get chained, and eventually gets to the root resource, which is the `cluster`.


src/server/auth/server/testing/auth_test.go
```diff
@@ -2346,17 +2346,46 @@ func TestDeleteAll(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Bob can now remove Alice as project owner
+	_, err = bobClient.ModifyRoleBinding(bobClient.Ctx(), &auth.ModifyRoleBindingRequest{
+		Principal: alice,
+		Roles:     []string{},
+		Resource:  &auth.Resource{Type: auth.ResourceType_PROJECT, Name: projectName},
+	})
+	require.NoError(t, err)
+
```
jrockway, line 2384: Any tests that auth.ProjectOwner can't be applied to repos and things like that?  Though it does seem like we don't do these checks for other things, perhaps mistakenly.  (You can `pachctl auth set enterprise repoOwner user:kilgore@trout` without getting an error.)

jrockway (approved):
Looks good to me.  Not sure which of the listed reviewers are also required.

src/server/auth/server/roles.go
```diff
@@ -174,26 +174,27 @@ func init() {
...
 		Permissions: combinePermissions(projectViewer.Permissions, []auth.Permission{
-			auth.Permission_PROJECT_LIST_REPO,
+			auth.Permission_PROJECT_CREATE_REPO,
 		}),
 	})
 
 	projectOwner := registerRole(&auth.Role{
 		Name:          auth.ProjectOwner,
-		ResourceTypes: []auth.ResourceType{auth.ResourceType_PROJECT, auth.ResourceType_CLUSTER},
+		ResourceTypes: []auth.ResourceType{auth.ResourceType_CLUSTER, auth.ResourceType_PROJECT},
```
acohen4, line 193: should project owners be able to modify repo role bindings within their project?

Could you add a test that iterates over the main project, repo and cluster admin roles and tries to modify the role bindings for different resources to verify the behavior?

src/server/auth/server/testing/auth_test.go
albscui: Good question. We don't have code that prevents binding roles to resource types that are incompatible with the role's `auth.Role.ResourceTypes`. Currently, this is just used to tell the user, which roles allows the user to access a resource type.

src/server/auth/server/roles.go
albscui: Currently the ProjectOwner role can't modify repo role bindings because it doesn't have the `REPO_MODIFY_BINDINGS` permission. Should we make ProjectOwner inherit RepoOwner?

And yeap I'll add more tests!

src/server/auth/server/roles.go
```diff
@@ -243,11 +244,11 @@ func init() {
 				auth.Permission_CLUSTER_ENTERPRISE_DEACTIVATE,
 				auth.Permission_CLUSTER_DELETE_ALL,
 				auth.Permission_CLUSTER_ENTERPRISE_PAUSE,
-				auth.Permission_PROJECT_DELETE,
 			}),
 	})
 }
 
+// TODO Deduplicate permissiosn in result
 func combinePermissions(permissions ...[]auth.Permission) []auth.Permission {
```
robert-uhl, line 252: I think it would be easy to add the deduplication code here: build up a map, then turn it into a slice and return that.

robert-uhl (approved):

src/server/auth/server/api_server.go
```diff
@@ -704,7 +727,7 @@ func (a *apiServer) evaluateRoleBindingInTransaction(txnCtx *txncontext.Transact
 }
 
 // AuthorizeInTransaction is identical to Authorize except that it can run
-// inside an existing etcd STM transaction.  This is not an RPC.
+// inside an existing STM transaction. This is not an RPC.
```
acohen4, line 730: we should update this comment to reference a normal psql txn and not an STM

src/server/auth/server/testing/auth_test.go
```diff
@@ -2358,5 +2367,60 @@ func TestCreateProject(t *testing.T) {
...
+	require.NoError(t, bobClient.CreateProjectRepo(project1, repo2))
+	require.NoError(t, bobClient.CreateProject(project2))
+	// auth resources
+	clusterResource := auth.Resource{Type: auth.ResourceType_CLUSTER}
+	aliceProject := auth.Resource{Type: auth.ResourceType_PROJECT, Name: project1}
+	bobProject := auth.Resource{Type: auth.ResourceType_PROJECT, Name: project2}
+	aliceRepoInAliceProject := auth.Resource{Type: auth.ResourceType_REPO, Name: fmt.Sprintf("%s/%s", project1, repo1)}
+	bobRepoInAliceProject := auth.Resource{Type: auth.ResourceType_REPO, Name: fmt.Sprintf("%s/%s", project1, repo2)}
+
+	tests := map[string]struct {
```
acohen4, line 2396: nice test :)

acohen4 (approved):
LGTM!

PR #8356, opened 2022-11-10
[CORE-1231] pass pachctl args to server for audit logging

armaanv:
This PR adds os.args to the client's grpc request metadata which is then fetched by the server and logged. I saw the log getting filled with messages from the pachd readiness check so I added a check for that.


jrockway (approved):
Very nice!

PR #8364, opened 2022-11-11
bump envoy proxy stream idle timeout to 24hrs

molinamelendezj:
nightly load test env recently started using the proxy, envs had been broken for a bit and seems we missed these errors, `rpc error: code = Unavailable desc = stream timeout` which require a longer timeout

related to #8284


jrockway (approved):
Looks good.  Can you also make this one 86400 for maximum paranoia?  https://github.com/pachyderm/pachyderm/blob/dd5fb90d416d744a74ccb60e9d4c5a42658c677b/etc/generate-envoy-config/envoy.libsonnet#L108

PR #8365, opened 2022-11-14
[CORE-1116] S3 gateway and projects

robert-uhl:


src/server/pfs/s3/driver.go
```diff
@@ -52,13 +53,31 @@ func NewMasterDriver() *MasterDriver {
...
+
 func (d *MasterDriver) listBuckets(pc *client.APIClient, r *http.Request, buckets *[]*s2.Bucket) error {
+	var isProjectAware = mux.Vars(r)["isProjectAware"] == "yes"
 	repos, err := pc.ListRepoByType("") // get repos of all types
 	if err != nil {
 		return err
 	}
 
 	for _, repo := range repos {
+		if !isProjectAware && repo.Repo.Project.Name != pfs.DefaultProjectName {
```
bbonenfant, line 78: Should this be `isProjectAware` rather than `!isProjectAware`?

bbonenfant (approved):
This looks OK to me. Just to clarify, this is a breaking change and the user must always specify the project even if its the default project?

robert-uhl:
> This looks OK to me. Just to clarify, this is a breaking change and the user must always specify the project even if its the default project?

Nope, it is backwards-compatible. If the user sets the access key ID and secret access key to the same value (as is current), then the existing non-project-aware code is in effect; if the user prepends `"PAC1"` to the access key ID, then the new project-aware code is in effect.

This way users can opt in each S3 gateway client one at a time.

src/server/pfs/s3/driver.go
robert-uhl: No: the logic here is that if the client is not project aware and if the repo is not in the default project, then the repo is skipped. So only repos in the default project will be returned to non-project-aware clients.

PR #8367, opened 2022-11-15
[Jupyter] Fix datums-related error message when notebooks starts up

smalyala:
Make datums request at notebooks startup only when connected to a cluster and logged in if auth enabled (when mount server response is `200`).

Ticket: https://linear.app/pachyderm/issue/INT-760/error-message-when-starting-up-notebooks

JIRA: [INT-782]

[INT-782]: https://pachyderm.atlassian.net/browse/INT-782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/jupyter-extension/src/plugins/mount/mount.tsx
+++ b/jupyter-extension/src/plugins/mount/mount.tsx
@@ -424,15 +424,17 @@ export class MountPlugin implements IMountPlugin {
           this._poller.status.code !== 200,
       );
 
-      try {
-        const res = await requestAPI<CurrentDatumResponse>('datums', 'GET');
-        if (res['num_datums'] > 0) {
-          this._keepMounted = true;
-          this._currentDatumInfo = res;
-          await this.setShowDatum(true);
+      if (this._poller.status.code === 200) {
+        try {
+          const res = await requestAPI<CurrentDatumResponse>('datums', 'GET');
+          if (res['num_datums'] > 0) {
+            this._keepMounted = true;
+            this._currentDatumInfo = res;
+            await this.setShowDatum(true);
+          }
+        } catch (e) {
+          console.log(e);
         }
-      } catch (e) {
-        console.log(e);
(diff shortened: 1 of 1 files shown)


jupyter-extension/src/plugins/mount/mount.tsx
```diff
@@ -424,15 +424,17 @@ export class MountPlugin implements IMountPlugin {
           this._poller.status.code !== 200,
       );
 
-      try {
-        const res = await requestAPI<CurrentDatumResponse>('datums', 'GET');
-        if (res['num_datums'] > 0) {
-          this._keepMounted = true;
-          this._currentDatumInfo = res;
-          await this.setShowDatum(true);
+      if (this._poller.status.code === 200) {
```
seslattery, line 427: Would there ever be a case where `requestAPI<ListMountsResponse>`  finishes with a non-200 response, but we'd still expect `requestAPI<CurrentDatumResponse>` to provide useful info?  If so, we could move the conditional into the catch statement, and still resolve this issue based on my understanding of the ticket.

seslattery (approved):
One minor question, but this looks good to me

jupyter-extension/src/plugins/mount/mount.tsx
smalyala: No. The `/datums` request is to verify if we currently have datums mounted. For datums to be currently mounted, that means we must have successfully connected to the cluster (and logged into auth if auth is activated) and made a mount datums request at some point prior. 
A 200 response means we're successfully connected to a cluster (and logged into auth if enabled). If we're getting a non-200 response, that effectively means we don't have datums mounted locally, defeating the need to use the response from the `/datums` request to check if datums are mounted or not.

PR #8370, opened 2022-11-15
[CORE-1219] Update S3 sidecar to be project-aware

robert-uhl:
The S3 buckets do not change for the sidecar, but this ensures that project-aware credentials work.

Jira: [INT-778]

[INT-778]: https://pachyderm.atlassian.net/browse/INT-778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/server/pfs/s3/s3.go
```diff
@@ -64,6 +64,10 @@ func (c *controller) requestClient(r *http.Request) *client.APIClient {
 	vars := mux.Vars(r)
 	if vars["s3gAuth"] != "disabled" {
 		accessKey := vars["authAccessKey"]
+		if strings.HasPrefix(accessKey, "PAC1") {
```
msteffen, line 67: To be honest, I'm offhand skeptical of this logic from a product perspective, so I want to make sure I understand and see what you think of a possible alternative approach.

- My skepticism is rooted in the fact that the auth token has been a point of confusion for users in the past (e.g. with auth off, the ACCESS_KEY_ID and SECRET_ACCESS_KEY _must_ be nonempty and _must_ match, but otherwise can be any string, which users find confusing). If we additionally have to explain that it _must_ contain this fairly opaque prefix, I worry that it'll cause even more confusion
- An alternative approach could be to put the project in the bucket name. Currently, in the global S3 gateway, buckets are identified as `<branch>.<repo>`, e.g. `master.myrepo`. What could work is to implement the convention that if S3G bucket names contain a single dot, then they refer to branches in the default project. If they contain two dots, then the identity becomes `<branch>.<repo>.<project>`. That way, existing keys and bucket names continue to work as they did previously, and things happen in the default project by default, but support for other projects is present.
- For the sidecar S3 gateway, buckets exist in the context of a pipeline, and I believe buckets are identified with the name of the PFS input they correspond to. So `/pfs/foo` becomes `s3://foo`. I believe this convention is already robust in the presence of projects (where inputs from other projects still have an input name, which is the repo name by default) so nothing needs to change.

WDYT? Do you prefer the auth access-key based approach or are you ok with the alternative? And let us know if the issue with the alternative approach is bandwidth

msteffen (changes requested):

src/server/pfs/s3/s3.go
msteffen: I see. I think what may make sense is to merge this PR to the end of unblocking projects, but for Integrations to plan on reverting the `PAC1` access key behavior and implementing the bucket name approach described blow (with a solution to the ambiguity you raise) before the 2.5 release. If we're quick, we'll have the solution proposed below, but if we run out of time, we'll still be able to release 2.5 with project support.

> Unfortunately, the currently accepted bucket name pattern is actually [<commitID>.][<branch>. | <branch>.<type>.]<repoName>

I think we should change the scheme to replace `<repo>.<type>` with `<repo>_<type>` to make the three-dots approach work (IIRC, types apply to repos rather than branches, e.g. spec and meta. If they do apply to branches, the fix still works, but will be slightly different). I know it's an API-incompatible change, but a) I'd bet money that zero people have ever accessed a non-user repo via the S3 gateway, so b) foisting a breaking change on those people seems basically way better than foisting a magic constant on the (I expect) significant number of users interested in projects.

> Unfortunately, our s2 library doesn’t just use SecretKey to deal with access key IDs; it also passes it in the authAccessKey request variable. This bug was revealed when testing a project-aware S3 sidecar client.

This I would like to understand--why is it a bug for S2 to pass the secret key in authAccessKey? (separately, we'd like to bring S2 into the main pach repo and archive the S2 repo, among other S2 changes).

I'm approving the PR, but please also get approval from Ben or Sahith before merging.

msteffen (approved):
Approving per my prior comment, but please also get approval from Ben or Sahith before merging

src/server/pfs/s3/s3.go
```diff
@@ -64,6 +64,10 @@ func (c *controller) requestClient(r *http.Request) *client.APIClient {
...
+			vars["isProjectAware"] = "yes"
```
acohen4, line 68: does the `isProjectAware` param get used anywhere?

acohen4: nvm 👍 read over CORE-1116

acohen4:
Ok I might be missing something so bare with me lol. 

The user shouldn't have to do anything different to use the S3 Sidecar right? We expose the S3_ENDPOINT env var for them to read from, and we already know the input names. For output, we always know the output project so s3://out should still be good.

I understand that projects do matter for the Global S3 Gateway because it's outside the context of a pipeline.

src/server/pps/server/s3g_sidecar_test.go
```diff
@@ -493,27 +524,38 @@ func TestFullS3(t *testing.T) {
...
-	name := t.Name()
-
 	c, userToken, ns := initPachClient(t)
+	if userToken == "" {
+		userToken = "abc123"
+	}
+	t.Run("ProjectUnaware", func(t *testing.T) {
+		testFullS3(t, c, userToken, userToken, ns)
+	})
+	t.Run("ProjectAware", func(t *testing.T) {
```
acohen4, line 538: Should we make these ProjectAware test cases use non-default project?

src/server/pfs/s3/s3.go
robert-uhl: Yup, it gets set here and in [auth.go](https://github.com/pachyderm/pachyderm/blob/f7f5088b2f5f2a43722c49f09e8b5a9ff540376a/src/server/pfs/s3/auth.go#L20), and it gets used in [listBuckets](https://github.com/pachyderm/pachyderm/blob/f7f5088b2f5f2a43722c49f09e8b5a9ff540376a/src/server/pfs/s3/driver.go#L71) and [bucket](https://github.com/pachyderm/pachyderm/blob/f7f5088b2f5f2a43722c49f09e8b5a9ff540376a/src/server/pfs/s3/driver.go#L163).

It would be nice to only set it in one place, but that would require a change in the s2 package.

robert-uhl:
> The user shouldn't have to do anything different to use the S3 Sidecar right? We expose the S3_ENDPOINT env var for them to read from, and we already know the input names. For output, we always know the output project so s3://out should still be good.

Yup. This story reduced down to ensuring that project-aware access key IDs work with the sidecar.

src/server/pps/server/s3g_sidecar_test.go
robert-uhl: Good idea. Done.

src/server/pps/server/s3g_sidecar_test.go
```diff
@@ -772,3 +822,22 @@ func TestS3SkippedDatums(t *testing.T) {
...
+		userToken = "abc123"
+	}
+	t.Run("ProjectUnaware", func(t *testing.T) {
+		testS3SkippedDatums(t, c, userToken, userToken, ns, pfs.DefaultProjectName, func(p, r string) string { return r })
+	})
+	t.Run("ProjectAware", func(t *testing.T) {
+		accessKeyID := "PAC1" + userToken
+		projectName := tu.UniqueString("project")
+		require.NoError(t, c.CreateProject(projectName))
+		testS3SkippedDatums(t, c, accessKeyID, userToken, ns, projectName, func(p, r string) string { return r + "." + p })
```
smalyala, line 841: Is this convention of `r + "." + p` just used for the sake of testing here and is isolated from the discussion with Matthew regarding how to name buckets with projects moving forward?

robert-uhl: It’s due to this particular test using _both_ the global _and_ the sidecar S3 gateways; the global gateway bucket name schema is different for project-aware clients. The bucket names when used as a sidecar do not change.

smalyala (approved):
LGTM

src/server/pfs/s3/s3.go
acohen4, line 68: ok one more nit: should the expected value be "true" instead of "yes"?

robert-uhl: It’s a raw string, so I don’t think that the value matters so much. _But_ you make a good point: this ends up being a magic string and someone might set it to "true" or "YES" or "AFFIRMATIVE" or "yeah" or whatever. I replaced the magic strings with some constants.

brendoncarroll (approved):

PR #8374, opened 2022-11-16
[PFS-10] Fix Misleading Version Check Logic

FahadBSyed:
This PR updates the version check logic:

- Previously, if an error was returned from inspect cluster and the error was scrubbed, pachctl would display an error saying the version is different, which was incorrect.
- Pachctl now attempts to get the version of pachd and display the actual version mismatch

### DIFF
--- a/src/client/admin.go
+++ b/src/client/admin.go
@@ -2,15 +2,16 @@ package client
 
 import (
 	"github.com/gogo/protobuf/types"
+
 	"github.com/pachyderm/pachyderm/v2/src/admin"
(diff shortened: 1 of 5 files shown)


robert-uhl:
Is there any way to detect this situation with a typed error and/or gRPC status code? It seems extremely fragile to depend on the contents of a human-readable string.

robert-uhl (approved):
LGTM. IMHO we should consider migrating from `grpcutil.ScrubGRPC` but that is not part of this story.

PR #8376, opened 2022-11-18
[CORE-1078] Update debug service to be project-aware

robert-uhl:


src/internal/testutil/debug.go
```diff
@@ -20,7 +21,7 @@ func DebugFiles(t testing.TB, dataRepo string) (map[string]*globlib.Glob, []stri
 	}
 	// Record glob patterns for expected source repo files.
 	for _, file := range []string{"commits.json", "commits-chart**"} {
-		pattern := path.Join("source-repos", dataRepo, file)
+		pattern := path.Join("source-repos", pfs.DefaultProjectName, dataRepo, file)
```
jrockway, line 24: I don't know about this.  We basically get no test coverage for dumps that involve projects; probably a bad idea to not be absolutely sure that we can debug projects.

src/server/debug/server/server.go
```diff
@@ -212,7 +212,7 @@ func (s *debugServer) handlePipelineRedirect(
 	collectWorker collectWorkerFunc,
 	redirect redirectFunc,
 ) (retErr error) {
-	prefix := join(pipelinePrefix, pipelineInfo.Pipeline.Name)
+	prefix := join(pipelinePrefix, pipelineInfo.Pipeline.Project.Name, pipelineInfo.Pipeline.Name)
```
jrockway, line 215: This pattern in general worries me; if somehow things are in a bad state and project is nil, then we panic.  I would prefer using the accessors throughout: pipelineInfo.GetPipeline().GetProject().GetName().  These will not panic when nil, even when chained.

jrockway (commented):
I think it would be good if we had the tests use the default project and a non-default project.  It would be most annoying if we somehow missed descending into projects in the debug dump; we wouldn't be able to fix bugs in the newest feature.

src/internal/testutil/debug.go
robert-uhl: I’ve updated the tests to test all four options of auth/no-auth and default-project/non-default-project.

src/server/auth/server/testing/auth_integration_test.go
```diff
@@ -274,6 +268,18 @@ func TestDebug(t *testing.T) {
...
 
+func TestDebug(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration tests in short mode")
+	}
+	t.Parallel()
+	c, _ := minikubetestenv.AcquireCluster(t, defaultTestOptions)
+	tu.ActivateAuthClient(t, c)
+	for _, projectName := range []string{pfs.DefaultProjectName, tu.UniqueString("project")} {
+		testDebug(t, c, projectName, tu.UniqueString("repo"))
```
jrockway, line 279: I would prefer `t.Run(projectName, func(t *testing.T) { testDebug(t, c, projectName, tu.UniqueString("repo")) }`.  Makes it clearer that we're running variants of the same test.

src/server/debug/server/server.go
jrockway: I like your fixup here.

src/server/debug/server/server.go
```diff
@@ -822,7 +832,7 @@ func (s *debugServer) getWorkerPodsLoki(ctx context.Context, pipelineInfo *pps.P
 	// logs for further inspection.  The alternative would be to get every worker that existed
 	// in some time interval, but that results in too much data to inspect.
 
-	queryStr := `{pipelineName="` + pipelineInfo.Pipeline.Name + `"}`
+	queryStr := fmt.Sprintf(`{pipelineProject="%s", pipelineName="%s"}`, pipelineInfo.Pipeline.Project.Name, pipelineInfo.Pipeline.Name)
```
jrockway, line 835: So, I know your quoting algorithm is correct based on the constraints on project and pipeline names, but here's Google's style guide's take: https://google.github.io/styleguide/go/decisions#use-q

To me it's a wash on whether you should rewrite this, but it does save you 4 bytes ;)

src/server/pachyderm_test.go
```diff
@@ -9966,45 +9966,38 @@ func TestListDatumFilter(t *testing.T) {
 	require.Equal(t, 25, i)
 }
 
-func TestDebug(t *testing.T) {
-	if testing.Short() {
-		t.Skip("Skipping integration tests in short mode")
-	}
-	t.Parallel()
-	c, _ := minikubetestenv.AcquireCluster(t)
+func testDebug(t *testing.T, c *client.APIClient, projectName, repoName string) {
```
jrockway, line 9969: t.Helper()

jrockway (approved):
Looking very good.  Some minor nits but nothing blocking.

src/server/debug/server/server.go
robert-uhl: My reasoning here is that this particular example is creating a LogQL string, and while LogQL and Go both assign similar meanings to strings within double quotes, it is a coincidence. If project or pipeline names could have any characters which would be invalid in LogQL, I would pass them to a QuoteLogQL function …

… I should do that anyway, shouldn’t I? That way it’s obvious when one reads it, and it is correct in the future. Although it’ll cost a lot more than four bytes of code☺

src/server/auth/server/testing/auth_integration_test.go
```diff
@@ -207,7 +201,7 @@ func TestDebug(t *testing.T) {
 	dataRepo := tu.UniqueString("TestDebug_data")
 	require.NoError(t, aliceClient.CreateProjectRepo(pfs.DefaultProjectName, dataRepo))
 
-	expectedFiles, pipelines := tu.DebugFiles(t, dataRepo)
+	expectedFiles, pipelines := tu.DebugFiles(t, pfs.DefaultProjectName, dataRepo)
```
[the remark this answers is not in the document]
robert-uhl: Yup, that was a mistake. I should probably have pulled the PR back into draft status while I was still working? Or just not responded to your original comment until I was fully done.

src/server/auth/server/testing/auth_integration_test.go
robert-uhl: Yeah, although it does make for longer tests, it probably makes the test _output_ more legible, which is a win.

src/server/debug/server/server.go
robert-uhl: I ended up doing exactly that: adding `quoteLogQLStreamSelector`, which just prints using Go’s `%q`.

src/server/debug/server/server.go
```diff
@@ -815,14 +831,25 @@ func (s *debugServer) forEachWorkerLoki(ctx context.Context, pipelineInfo *pps.P
...
 
+// quoteLogQLStreamSelector returns a string quoted as a LogQL stream selector.
+// The rules for quoting a LogQL stream selector are documented at
+// https://grafana.com/docs/loki/latest/logql/log_queries/#log-stream-selector
+// to be the same as for a Prometheus string literal, as documented at
+// https://prometheus.io/docs/prometheus/latest/querying/basics/#string-literals.
+// This happens to be the same as a Go string, with single or double quotes or
+// backticks allowed as enclosing characters.
+func quoteLogQLStreamSelector(s string) string {
+	return fmt.Sprintf("%q", s)
```
jrockway, line 842: strconv.Quote

Apparently "github.com/prometheus/prometheus/model/labels" is the actual library used to parse these selectors; it prints them with strconv.Quote as well.  I suppose if we already depend on that library you could use that directly though.  (labels.NewMatcher(labels.Equal, "pipeline", whatever)).

src/server/debug/server/util.go
```diff
@@ -132,3 +134,59 @@ func copyTar(tw *tar.Writer, tr *tar.Reader, prefix ...string) error {
 		}
 	}
 }
+
+func validateProject(p *pfs.Project) error {
```
brendoncarroll, line 138: Don't we have a `ppsutil` or `pfsutil` where these could be exported.  It seems like they would get a lot of use?

robert-uhl: There is a `src/internal/ppsutil`, but no `pfsutil` (yet). If we did end up needing them anywhere else in the codebase then it would probably make sense to put them in a new `src/internal/pfsutil`, but since we don’t yet I think it’s better not to add more entities to the system.

src/server/auth/server/testing/auth_integration_test.go
```diff
@@ -274,6 +269,20 @@ func TestDebug(t *testing.T) {
...
 
+func TestDebug(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration tests in short mode")
+	}
+	t.Parallel()
+	c, _ := minikubetestenv.AcquireCluster(t, defaultTestOptions)
+	tu.ActivateAuthClient(t, c)
+	for _, projectName := range []string{pfs.DefaultProjectName, tu.UniqueString("project")} {
+		t.Run("projectName", func(t *testing.T) {
```
smalyala, line 280: I'm assuming "projectName" here wasn't intended to be hardcoded?

smalyala (approved):
LGTM. Left one question.

PR #8378, opened 2022-11-18
support for projects in dag file

djanicekpach:
CORE-1063 - make load tests project aware

This adds the ability to switch the specified project for the `pachctl run pps-load-test` command. You can now specify a project by setting `projectName/repoName` in the dagSpec file rather than just specifying `repoName`. 

You can see an exmaple in the newly added dag-5.txt spec. If no project is specified it will use the default project like it did before, and there should be no change in behavior. I ran this against each present dag spec to test.


djanicekpach:
Closing. See CORE-1063 for discussion on why.

PR #8381, opened 2022-11-21
refactor fsck to make it easier to test

zmajeed:
Logic should not have changed


zmajeed:
tests do not run on fork branches because non-free machines are needed

PR #8382, opened 2022-11-21
ListRepo - filter on projects and check permissions

albscui:
The result of `ListRepo` can now be filtered based on both projects and permissions. The projects filter allows the user to list repos that are only part of the project specified. After that, our auth checks kicks in, and we check whether the user has `PROJECT_LIST_REPO` permission on the repo's project or `REPO_READ` on the specific repo.


src/pfs/pfs.proto
```diff
@@ -191,6 +191,7 @@ message ListRepoRequest {
   // type is the type of (system) repos that should be returned
   // an empty string requests all repos
   string type = 1;
+  repeated string projects = 2;
```
robert-uhl, line 194: Should probably document that no projects means all projects, while some projects means just those projects.

src/server/auth/server/util.go
```diff
@@ -32,34 +32,18 @@ func (a *apiServer) CheckRepoIsAuthorizedInTransaction(txnCtx *txncontext.Transa
...
-	if errors.Is(err, auth.ErrNotActivated) {
-		return nil
-	}
-
-	req := &auth.AuthorizeRequest{Resource: resource, Permissions: p}
-	resp, err := a.AuthorizeInTransaction(txnCtx, req)
 	if err != nil {
+		// FIXME stop swallowing ErrNotActivated
+		// Why are we swallowing? So that code that calls into this can not worry about auth activeness.
+		// However, we should let the code calling this explicitly handle it.
```
robert-uhl, line 38: Are you sure about that? That would mean that each call site would end up looking like:

```go
if err := srv.checkResourceIsAuthorizedInTransaction(…); err != nil {
  if errors.Is(err, auth.ErrNotActivated) {
    return nil
  }
}
```

which seems a bit much. And error-prone too. Maybe `WhoAmI` should not return an error when auth is inactive? That sounds iffy too …

src/server/auth/server/util.go
```diff
@@ -72,18 +56,39 @@ func (a *apiServer) CheckRepoIsAuthorized(ctx context.Context, r *pfs.Repo, p ..
...
+}
+
+func (a *apiServer) checkResourceIsAuthorized(ctx context.Context, resource *auth.Resource, p ...auth.Permission) error {
 	me, err := a.WhoAmI(ctx, &auth.WhoAmIRequest{})
-	if auth.IsErrNotActivated(err) {
-		return nil
+	if err != nil {
+		// FIXME stop swallowing ErrNotActivated
+		// Why are we swallowing? So that code that calls into this can not worry about auth activeness.
+		// However, we should let the code calling this explicitly handle it.
```
robert-uhl, line 78: See above.

src/server/pfs/server/driver.go
```diff
@@ -256,46 +251,65 @@ func (d *driver) inspectRepo(txnCtx *txncontext.TransactionContext, repo *pfs.Re
...
 
 	if repoType == "" {
 		// blank type means return all
-		return errors.EnsureStack(d.repos.ReadOnly(ctx).List(repoInfo, col.DefaultOptions(), processFunc))
-	} else {
-		err := d.repos.ReadOnly(ctx).GetByIndex(pfsdb.ReposTypeIndex, repoType, repoInfo, col.DefaultOptions(), processFunc)
-		return errors.EnsureStack(err)
+		return errors.Wrap(d.repos.ReadOnly(ctx).List(repoInfo, col.DefaultOptions(), processFunc), "could not list repos of all types")
 	}
+	return errors.Wrap(d.repos.ReadOnly(ctx).GetByIndex(pfsdb.ReposTypeIndex, repoType, repoInfo, col.DefaultOptions(), processFunc), "could not to list repos")
```
robert-uhl, line 312: `could not list repos`

Although: I prefer to let the caller say ‘could not list repos’; I would return something like ‘could not get repos of type spec: ERROR FROM GetByIndex’. That way each function returns what went wrong from its point of view.

src/server/auth/server/testing/auth_test.go
```diff
@@ -2423,3 +2423,82 @@ func TestModifyRoleBindingAccess(t *testing.T) {
 		})
 	}
 }
+
+func TestListRepoWithProjectAccessControl(t *testing.T) {
```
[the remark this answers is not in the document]
albscui: In general, our "integration tests" are really tests that involve pachyderm pipelines, which depends K8s. If the logic we are testing doesn't involve pipelines, then we don't really need or rather should avoid "integration-style" tests.

In this case, our logic just involves creating some repos and testing access controls, so no need for integration tests.

albscui:
> Is this supposed to update the pachctl command as well? I don't actually see a ticket for updating the commands, so I assumed it was here, but I'm not sure. It'd be nice for testing to have this as well.

Good catch! I'll extend the pachctl part as well. I thought we had a ticket for extending pachctl to be project aware for all commands, but I can't find that ticket anymore. However, I now think it is just better to include both the client and server side features as a single deliverable unit with tests.

src/server/pfs/server/driver.go
```diff
@@ -256,46 +251,65 @@ func (d *driver) inspectRepo(txnCtx *txncontext.TransactionContext, repo *pfs.Re
...
 			}
 		}
+		if authSeemsActive && includeAuth {
+			permissions, roles, err := d.getPermissions(ctx, repoInfo.Repo)
+			if err != nil {
+				if errors.Is(err, auth.ErrNotActivated) {
+					authSeemsActive = false
+					return cb(proto.Clone(repoInfo).(*pfs.RepoInfo))
+				} else {
+					return errors.Wrapf(err, "error getting access level for %q", repoInfo.Repo)
```
djanicekpach, line 300: I'm hitting this error when running opencv as pach:root if I have configured a project other than default as my default project. Do I need to do anything to set up auth for a specific repo? (If this is an extension of the client changes, feel free to ignore)
```
pachctl list repo
rpc error: code = Unknown desc = could not get repos of type "user": ERROR FROM GetByIndex: error getting access level for "NON_DEFAULT-1/images": no role binding exists for REPO NON_DEFAULT-1/images
```
T
he project does exist:
```
pachctl create project NON_DEFAULT-1
projects NON_DEFAULT-1 already exists
```

albscui: Does `NON_DEFAULT-1/images` repo exist? We create the role binding at repo creation time.

djanicekpach: It should, at least when I run the example with auth turned off I can see it in `list repo` I tried running the example both before and after activating enterprise auth and it seems to exist without the auth.

albscui: We figured out the issue. When projects are created in an auth inactive Pachyderm instance, they don't get role bindings created. Once auth is active, auth checks are going to check for the project's role bindings and through an error that they don't existing. However, currently there's a bug where given a repo, we also check the repo's project's role binding, which doesn't exist, but we through an error that is specific to the repo instead of the project.
Fixed here https://github.com/pachyderm/pachyderm/pull/8382/commits/2d360e6cc2fd2b288983bb2365816bc38ee886d2

Issue #2544, opened 2017-11-27
Handle Zombie Processes


sjezewski:
If a user's code creates subprocesses and dies, we can get zombie processes.

These will not get reaped when the user code completes / the context is cancelled / etc.

Instead, we should write the pid of the worker process, and kill all other processes in the container. The alternative is nsenter / namespace the PIDs ... but that seems more complicated.

This will also strengthen our datum/job timeouts.

Issue #8375, opened 2022-11-16
Service pipeline stops serving static files after new data committed


dsgibbons:
**What happened?**:

I'm using a pipeline that generates artifacts which are then served by a Flask app embedded in a service pipeline. Some of these artifacts are `png` images that need to be served from `static`. To achieve this, I use the following service pipeline:

**service.json**
```json
{
    "pipeline": {
      "name": "dummy_service"
    },
    "input": {
        "pfs": {
            "glob": "/",
            "repo": "do_stuff"
        }
    },
    "service": {
        "external_port": 30887,
        "internal_port": 8887
    },
    "transform": {
        "cmd": [ "/bin/bash" ],
        "stdin": [
            "ln -s /pfs/do_stuff static",
            "python app.py" 
        ],
        "image": "dummy:0.0.1"
    }
}
```

where `dummy:0.0.1` is created from the following Dockerfile:

**Dockerfile**
```dockerfile
FROM python
RUN pip install flask flask-cors --no-cache-dir 
WORKDIR /app/
COPY app.py .
```

and `app.py` is as follows:

**app.py**
```python
from flask import Flask, url_for, redirect
from flask_cors import CORS


def create_app():
    app = Flask(__name__) 

    CORS(app)

    @app.route("/")
    def home():
        return "home"

    @app.route("/<file>")
    def view_file(file):
        return redirect(url_for("static", filename=f"{file}.png"))

    return app


if __name__ == "__main__":
    app = create_app()
    app.run(host="0.0.0.0", port=8887)
```

Let's say the `do_stuff` pipeline processes a datum and produces the output `image1.png`. Then, I can deploy the above pipeline using:

`pachctl create pipeline -f service.json`

I can then navigate to `127.0.0.1:30887/image1` and it will successfully display `127.0.0.1:30887/static/image1.png`. However, if `do_stuff` processes a new datum and produces `image2.png`, then all `static` locations end up breaking (including `image1`). Running 

`pachctl update pipeline -f service.json --reprocess`

fixes the problem and allows both files to be served from `static`. 

While broken, the logs show (let `pachderm=image1` and `elephant=image2`):

**log**
```
 * Serving Flask app 'app'
 * Debug mode: off
Address already in use
Port 8887 is in use by another program. Either identify and stop that program, or start the server with a different port.
Address already in use
 * Serving Flask app 'app'
 * Debug mode: off
Port 8887 is in use by another program. Either identify and stop that program, or start the server with a different port.
[...]

PR #8385, opened 2022-11-23
Deal with subprocesses cleanly in the worker

jrockway:
This fixes 2 things.

## Zombies (CORE-1279)

If user code spawns a background process and it exits without the user code calling `waitpid` on it before the user code exits, we end up with a zombie process that takes up a slot in the process table forever.  We now run the worker through dumb-init to reap these zombies, preventing the process table from filling up when user code behaves like this.   Fixing this is not a huge deal in practice, but it's good for correctness.  Setting DUMB_INIT_DEBUG=1 will print something whenever a zombie is reaped.  A pipeline that runs a bash script like `sleep 30 & touch /pfs/out/foo` is an example of how to produce a zombie prior to this change. 

## Subprocesses that don't exit (CORE-1016, CORE-1280):

If user code spawns a background process that DOES NOT exit, then we have big problems.  It could hold stdout open forever, which means we wait forever reading that inside `cmd.Wait()`, even if the user code is killed because of a context timeout.  We had a workaround prior to #8066, but it didn't go far enough.  A case it ignores is where a child of the user code doesn't read STDOUT but uses some other resource on the system, like a TCP port.  Issue #8375 is an example of something that does this.  This PR fixes that case by running the user code in a process group, and sending SIGKILL to the process group when the user code exits (leaving around background processes) or is killed (because of an expiring context).  This means that everything the user code spawned gets SIGKILL'd and is guaranteed to terminate, releasing any file descriptors we were reading (STDOUT) or other resources (TCP ports).  

We make a best effort to log when we're about to do this (find /proc/... races with processes exiting on their own before we call kill); so the logs look something like this running the opencv example as a bash script like `sleep infinity & python3 /edges.py`:

```
Nov 22 22:43:24 beginning to run user code pipelineName:↑ projectName:↑ workerId:↑ jobId:↑ data:[{"hash":"q4zXe15LED6ufrTyMlNYCP+dN/coyuQPAQ/KoY7n9iQ=","path":"/1VqcWw9.jpg"}] datumId:aa7e5b5ee2d9e78cfd714209d5d591078f86671e53e1710f1e658e353813cc25
Nov 22 22:43:24 /usr/local/lib/python3.4/dist-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment. pipelineName:↑ projectName:↑ workerId:↑ jobId:↑ data:↑ datumId:↑ user:true
Nov 22 22:43:24   warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.') pipelineName:↑ projectName:↑ workerId:↑ jobId:↑ data:↑ datumId:↑ user:↑
Nov 22 22:43:24 /usr/local/lib/python3.4/dist-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment. pipelineName:↑ projectName:↑ workerId:↑ jobId:↑ data:↑ datumId:↑ user:↑
Nov 22 22:43:24   warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.') pipelineName:↑ projectName:↑ workerId:↑ jobId:↑ data:↑ datumId:↑ user:↑
Nov 22 22:43:24 note: about to kill unexpectedly-remaining subprocess of the user code: pid=35 comm=(sleep) cmdline=sleep infinity  pipelineName:↑ projectName:↑ workerId:↑ jobId:↑ data:↑ datumId:↑
Nov 22 22:43:24 finished running user code after 598.008993ms pipelineName:↑ projectName:↑ workerId:↑ jobId:↑ data:↑ datumId:↑
```                                                                                                              

A process can leave the process group and they will have the same problems (stdout stuck open) as before.  You have to go out of your way to leave the process group, but if a user does that, we have no way of fixing that.  (For that, we'd have to create a PID namespace with unshare, but that requires root or at least CAP_SYSADMIN.)

Anyway, this bug is fairly common; if you've ever run "stop job" and it didn't stop, or if you see something taking longer than the DatumTimeout, you hit this bug.  I've seen it quite a bit but didn't recognize the cause until looking into this in detail.  `sleep infinity & touch /pfs/out/foo` is an example of a script that used to break, but now works.

`TestService` in pachyderm_test would be broken by this bug, but they found a workaround for this issue: `exec python ...` instead of `python ...`.  I've made that test do it "wrong" now, acting as a test for this fix.  Now users don't have to be UNIX wizards to have a working web server for files in a repo!

One reason this bug is problematic in general is that we use SIGKILL (via a hard-coded default in exec.Cmd); if the user code has cleanup code to terminate it subprocesses, it can't run it because SIGKILL is not something it can catch.  SIGTERM + grace period + SIGKILL would be more polite and less likely to cause problems; a well-behaved app would catch SIGTERM and kill its children.  The ideal solution here is to run the user code in a new PID namespace and let it be pid 1; then user code could use dumb-init to remap SIGKILL to SIGTERM if it felt like it.  But creating a PID namespace requires root, which we try to avoid, so I didn't pursue that route.

Finally, *this is technically a breaking change*.  If you had a pipeline like `do_actual_work & sleep 10` and `do_actual_work` runs longer than 10 seconds, that would work in the past, but won't work now.  Worse, the job will be marked successful because bash will exit 0 after the sleep 10.  In the past, we would wait indefinitely for `do_actual_work` to close stdout, but now we SIGKILL it instantly.  Something we'll have to watch out for as users upgrade.

## How I tested this

I ran TestPipelineWithSubprocess with and without this PR -- without, it hangs forever; with, it passes.

I ran (this PR's version of) TestService with and without this PR -- without, the first version of the server never gets killed and the test times out; with, it works as expected.  Some logs:
```
Nov 22 21:22:23 beginning to run user code master:true pipelineName:pipelineservice47fafd305644 projectName:default workerId:pipeline-default-pipelineservice47fafd305644-v1-5jpzk data:[{"hash":"tzUlQiTOsvaBavGE0GHMNpn5FVUqne/23Cvboctq8pw=","path":"/"}] datumId:17f6ef2602f33a7faa8e4fe5eca9730c4851f2132fb97dd67c1d48cecf82a1c7

Nov 22 21:22:28 10.244.0.1 - - [23/Nov/2022 02:22:28] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:true
Nov 22 21:22:28 canceling previous service, new job ready master:true pipelineName:↑ projectName:↑ workerId:↑
Nov 22 21:22:29 10.244.0.1 - - [23/Nov/2022 02:22:29] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:[{"hash":"tzUlQiTOsvaBavGE0GHMNpn5FVUqne/23Cvboctq8pw=","path":"/"}] datumId:17f6ef2602f33a7faa8e4fe5eca9730c4851f2132fb97dd67c1d48cecf82a1c7 user:true
Nov 22 21:22:29 10.244.0.1 - - [23/Nov/2022 02:22:29] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:22:30 10.244.0.1 - - [23/Nov/2022 02:22:30] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:22:32 10.244.0.1 - - [23/Nov/2022 02:22:32] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:22:33 10.244.0.1 - - [23/Nov/2022 02:22:33] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:22:35 10.244.0.1 - - [23/Nov/2022 02:22:35] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:22:40 10.244.0.1 - - [23/Nov/2022 02:22:40] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:22:43 10.244.0.1 - - [23/Nov/2022 02:22:43] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:22:46 10.244.0.1 - - [23/Nov/2022 02:22:46] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:22:49 10.244.0.1 - - [23/Nov/2022 02:22:49] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:22:53 10.244.0.1 - - [23/Nov/2022 02:22:53] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:22:58 10.244.0.1 - - [23/Nov/2022 02:22:58] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:23:05 10.244.0.1 - - [23/Nov/2022 02:23:05] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:23:08 10.244.0.1 - - [23/Nov/2022 02:23:08] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:23:13 10.244.0.1 - - [23/Nov/2022 02:23:13] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:23:17 10.244.0.1 - - [23/Nov/2022 02:23:17] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:23:22 10.244.0.1 - - [23/Nov/2022 02:23:22] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:23:26 10.244.0.1 - - [23/Nov/2022 02:23:26] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
Nov 22 21:23:30 10.244.0.1 - - [23/Nov/2022 02:23:30] "GET /TestService_data9eabed90410e/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:↑
```
(Note how we keep serving the wrong content after "canceling previous service, new job ready"; the requests for file1 at increasing intervals is the backoff loop waiting for the content to change.)

With this change, the logs are much more in line with what we expect:
```
Nov 22 21:29:51 beginning to run user code master:true pipelineName:pipelineservice6c2a4335d762 projectName:default workerId:pipeline-default-pipelineservice6c2a4335d762-v1-pljn6 data:[{"hash":"tzUlQiTOsvaBavGE0GHMNpn5FVUqne/23Cvboctq8pw=","path":"/"}] datumId:f69008bc94305836b09b32cedafbbc3544a97d1d1ac08da93cef0b8b4c93dafc
Nov 22 21:29:56 10.244.0.1 - - [23/Nov/2022 02:29:56] "GET /TestService_data23806d8940d0/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:true
Nov 22 21:29:56 canceling previous service, new job ready master:true pipelineName:↑ projectName:↑ workerId:↑
Nov 22 21:29:56 killing remaining children master:↑ pipelineName:↑ projectName:↑ workerId:↑ data:[{"hash":"tzUlQiTOsvaBavGE0GHMNpn5FVUqne/23Cvboctq8pw=","path":"/"}] datumId:f69008bc94305836b09b32cedafbbc3544a97d1d1ac08da93cef0b8b4c93dafcNov 22 21:29:56 errored running user code after 5.124852286s: signal: killed master:↑ pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑
Nov 22 21:29:56 starting new service, job: 073c5aa44d0f4d959f342f16dadae57b master:↑ pipelineName:↑ projectName:↑ workerId:↑
Nov 22 21:29:57 beginning to run user code master:true pipelineName:pipelineservice6c2a4335d762 projectName:default workerId:pipeline-default-pipelineservice6c2a4335d762-v1-pljn6 data:[{"hash":"NXAbkRl+Dg9yqP1d97rKGS/hekrlWOTEN8Tqj8pi+dg=","path":"/"}] datumId:f69008bc94305836b09b32cedafbbc3544a97d1d1ac08da93cef0b8b4c93dafc
Nov 22 21:29:57 10.244.0.1 - - [23/Nov/2022 02:29:57] "GET /TestService_data23806d8940d0/file1 HTTP/1.1" 200 - pipelineName:↑ projectName:↑ workerId:↑ data:↑ datumId:↑ user:true
```

### DIFF
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -7,7 +7,12 @@ COPY LICENSE /LICENSE
 COPY licenses /licenses
 
 WORKDIR /app
-COPY worker . 
+# etc/worker/init.go copies the dumb-init for the right architecture into the user container and
+# makes it executable.  This is to avoid needing a shell in this container to run
+# `if [ $TARGETPLATFORM = "linux/amd64 "]; ...`
+ADD --chown=1000:1000 https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_x86_64 /app/dumb-init-amd64
+ADD --chown=1000:1000 https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_aarch64 /app/dumb-init-arm64
+COPY worker .
 COPY worker_init init
 COPY pachctl .
 COPY pachtf .
--- a/etc/worker/init.go
+++ b/etc/worker/init.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 )
 
 func cp(src, dst string) error {
@@ -31,13 +32,12 @@ func cp(src, dst string) error {
 	}
 	defer out.Close()
 
-	_, err = io.Copy(out, in)
-	if err != nil {
+	if _, err := io.Copy(out, in); err != nil {
 		return err
 	}
 
 	// make the file executable
-	if err = out.Chmod(os.ModePerm); err != nil {
+	if err := out.Chmod(os.ModePerm); err != nil {
 		return err
 	}
 
@@ -45,10 +45,15 @@ func cp(src, dst string) error {
 }
 
 func main() {
+	// Copy over pachyderm binaries.
 	for _, bin := range []string{"worker", "pachctl", "pachtf"} {
 		src, dst := fmt.Sprintf("/app/%s", bin), fmt.Sprintf("/pach-bin/%s", bin)
 		if err := cp(src, dst); err != nil {
 			panic(err)
 		}
+	}
+	// Copy over the correct variant of dumb-init.
+	if err := cp(fmt.Sprintf("/app/dumb-init-%s", runtime.GOARCH), "/pach-bin/dumb-init"); err != nil {
+		panic(err)
 	}
 }
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -201,12 +201,24 @@ func TestPipelineWithSubprocesses(t *testing.T) {
 			DatumTimeout:    types.DurationProto(5 * time.Second),
 		},
 	)
-	require.NoError(t, err, "should create pipeline")
+	require.NoError(t, err, "should create pipeline ok")
 
-	_, err = c.WaitCommitSetAll(commit1.ID)
-	require.NoError(t, err, "should wait for commitset")
-	require.NoError(t, c.GetFile(client.NewProjectCommit(projectName, pipeline, "master", commit1.ID), "foo", io.Discard), "should get foo without error")
-	require.NoError(t, c.GetFile(client.NewProjectCommit(projectName, pipeline, "master", commit1.ID), "bar", io.Discard), "should get bar without error")
+	commitInfo, err := c.InspectProjectCommit(projectName, pipeline, "master", "")
+	require.NoError(t, err, "should inspect commit ok")
+	commitInfos, err := c.WaitCommitSetAll(commitInfo.Commit.ID)
+	require.NoError(t, err, "should wait commit ok")
+
+	var output *pfs.CommitInfo
+	for _, info := range commitInfos {
+		if proto.Equal(info.Commit.Branch.Repo, client.NewProjectRepo(projectName, pipeline)) {
+			output = info
+			break
+		}
+	}
+	require.NotNil(t, output, "output commit should have been found (got: %#v)", commitInfos)
+
+	require.NoError(t, c.GetFile(output.Commit, "foo", io.Discard), "should get foo without error")
+	require.NoError(t, c.GetFile(output.Commit, "bar", io.Discard), "should get bar without error")
 }
 
 func TestCrossProjectPipeline(t *testing.T) {
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -173,7 +173,7 @@ func TestPipelineWithSubprocesses(t *testing.T) {
 	c, _ := minikubetestenv.AcquireCluster(t)
 	c = c.WithDefaultTransformUser("1000")
 
-	projectName := tu.UniqueString("project")
+	projectName := tu.UniqueString("p")
 	require.NoError(t, c.CreateProject(projectName))
 	dataRepoName := tu.UniqueString("TestPipelineWithSubprocesses_data")
 	require.NoError(t, c.CreateProjectRepo(projectName, dataRepoName))
@@ -184,7 +184,7 @@ func TestPipelineWithSubprocesses(t *testing.T) {
 	require.NoError(t, c.PutFile(commit1, "bar", strings.NewReader("bar"), client.WithAppendPutFile()))
 	require.NoError(t, c.FinishProjectCommit(projectName, dataRepoName, commit1.Branch.Name, commit1.ID))
 
-	pipeline := tu.UniqueString("TestPipelineWithSubprocesses")
+	pipeline := tu.UniqueString("TestPipelineWS")
 	_, err = c.PpsAPIClient.CreatePipeline(
 		c.Ctx(),
 		&pps.CreatePipelineRequest{
@@ -217,8 +217,12 @@ func TestPipelineWithSubprocesses(t *testing.T) {
 	}
 	require.NotNil(t, output, "output commit should have been found (got: %#v)", commitInfos)
 
-	require.NoError(t, c.GetFile(output.Commit, "foo", io.Discard), "should get foo without error")
-	require.NoError(t, c.GetFile(output.Commit, "bar", io.Discard), "should get bar without error")
+	buf := new(bytes.Buffer)
+	require.NoError(t, c.GetFile(output.Commit, "foo", buf), "should get foo without error")
+	require.Equal(t, "foo", buf.String(), "content should be correct")
+	buf.Reset()
+	require.NoError(t, c.GetFile(output.Commit, "bar", buf), "should get bar without error")
+	require.Equal(t, "bar", buf.String(), "content should be correct")
 }
 
 func TestCrossProjectPipeline(t *testing.T) {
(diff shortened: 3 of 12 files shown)


msteffen:
Fix #2544

PR #8386, opened 2022-11-23
worker_rc: set default requests if resource_requests and resource_limits are both empty

jrockway:
I tested this by upgrading to this version on a cluster that had some pipelines with no resources setup; they all restarted with 1 CPU.  I then edited a pipeline to explicitly say:

```json
"resource_requests": {},
```
and that worker was restarted with a request for 0 CPUs.  (ResourceSpec.Cpu is a float64, so 0 == empty.  Fortunately, the containing message can distinguish between empty and unset!) 

Since the original version of this PR, I've added defaults for memory and ephemeral storage, made those defaults configurable via helm, added a log message to warn the user that defaults are being set for them (with instructions on how to opt out), and completely replaced our previous version of this hack (cpu: 0, mem: 64M) with this method.  (You can now really disable all resource requests; that wasn't actually possible before.)

Nothing changes about the storage container, or pipelines that specify any sort of resource request or limit.

### DIFF
--- a/etc/helm/pachyderm/templates/pachd/deployment.yaml
+++ b/etc/helm/pachyderm/templates/pachd/deployment.yaml
@@ -342,6 +342,18 @@ spec:
         - name: GC_PERCENT
           value: {{ .Values.pachd.gcPercent | quote }}
         {{- end }}
+        {{- if ne .Values.pachd.defaultPipelineCPURequest "" }}
+        - name: PIPELINE_DEFAULT_CPU_REQUEST
+          value: {{ .Values.pachd.defaultPipelineCPURequest | quote }}
+        {{- end }}
+        {{- if ne .Values.pachd.defaultPipelineMemoryRequest "" }}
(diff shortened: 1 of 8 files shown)


jrockway:
OK, as expected, the CI machine doesn't have enough CPUs.  I might go and make this default configurable at the pachd level.

acohen4 (approved):
LGTM. Not sure what's up with the bad CI run though

jrockway:
I'm not 100% sure, but it looks like a lot of tests time out each run.  I reduced the number of parallel tests by 1 (to 2), but that doesn't help.  I think what happens is that some test creates a bunch of pipelines and then we can't schedule anything else.

jrockway:
@acohen4 Please take another look.  I've made all the defaults configurable, so that CI/local installs don't have to deal with having sufficient CPU (etc.).  I also made cmdutil.Populate able to parse things into structs (useful for k8s resource.Quantity 🎉).

robert-uhl (approved):

PR #8390, opened 2022-11-23
[CORE-1080] Support for projects in extended traces

robert-uhl:


src/internal/tracing/extended/extended_trace.proto
```diff
@@ -11,6 +11,7 @@ message TraceProto {
   // (trace ID, span ID, sampled) tuple, basically)
   map<string, string> serialized_trace = 1;
 
+  string project = 3;
```
acohen4, line 14: i guess it probably would have been nicer if we just had the Pipeline proto in here

acohen4 (approved):
LGTM

src/internal/tracing/extended/extended_trace.proto
robert-uhl: Agreed, but since this already had the string name instead of the message I figured that it was better to make the small change than a larger one. Could have gone either way, TBH.

src/internal/tracing/extended/extended_trace.go
```diff
@@ -100,7 +102,7 @@ func PersistAny(ctx context.Context, c *etcd.Client, pipeline string) {
 	}
 	if _, err := col.NewSTM(ctx, c, func(stm col.STM) error {
 		tracesCol := TracesCol(c).ReadWrite(stm)
-		return errors.EnsureStack(tracesCol.PutTTL(pipeline, traceProto, int64(duration.Seconds())))
+		return errors.EnsureStack(tracesCol.PutTTL(pipeline.String(), traceProto, int64(duration.Seconds())))
 	}); err != nil {
 		log.Errorf("could not persist extended trace for pipeline %q to etcd: %v",
 			pipeline, err)
```
smalyala, line 108: Do you want `pipeline.String()` here too?

smalyala (approved):
LGTM

src/internal/tracing/extended/extended_trace.go
robert-uhl: It’s called when using `%q`, and https://github.com/pachyderm/pachyderm/blob/3fc355fc4d979f42c77b1763247b227223ebe898/src/pps/pps.go#L11 will include the project name if any.

PR #8391, opened 2022-11-23
Enable a bunch of linters

jrockway:
I was mostly in here to mandate gofmt, but also enabled some other easy-to-enable linters and opened a ticket to do the rest.

I thought I was going crazy because I've always assumed these linters were turned on, but it was in Hub that I turned them on.  Now we can have them too.

This is CORE-1314.


albscui (approved):
OMG, this is awesome! 🥳

PR #8306, opened 2022-10-21
Provenance Re-Model

acohen4:
User visible changes:
- behavior of what is in a commit set, as per the commit set API may change, because aliases are not immediately present
- when commits come back from the API server, their branch field will not be populated.
- triggered commits will now be present in a new commit set

Open items
- [x] Rethink InspectCommitSet implementation
- [x] Migrate fake alias edge case
- [x] Figure out whether we need to handle "un-migrated" commit structs stored in meta repos
- [x] Blocking error for Branchless Commit Migration
- [x] Make decision on which server_test.go tests to disable

Add Tests
- [x] Trigger Migration (tests branchless commit migration)
- [x] Migration after squashing user commits
- [x] Complicated Squash

PR #8393, opened 2022-11-28
[CORE-1121] Correct pachctl delete repo --all for projects

robert-uhl:
Currently `pachctl delete repo --all` calls `DeleteAll`, which deletes all repos and all projects. This changes it to be project-aware, deleting all repos _within_ a project.

API-wise, it adds a `project` member to the `DeleteRepoRequest` message. An alternative solution would have been to add an entirely new `DeleteRepos` API; I chose not to implement that out of a desire to keep the API surface small.


src/pfs/pfs.proto
```diff
@@ -196,6 +196,9 @@ message ListRepoRequest {
 message DeleteRepoRequest {
   Repo repo = 1;
   bool force = 2;
+  // If project present, and if repo is absent, then all repos in
+  // project will be deleted.
+  Project project = 3;
```
acohen4, line 201: since Repo already contains project, shouldn't we rely on that? If we want to express deleting an entire project, I think that should be its own RPC

robert-uhl: > since Repo already contains project, shouldn't we rely on that?

My first attempt did that: a request to delete a repo with an empty repo name would delete all repos in that repo’s project. But on consideration that seemed like a footgun, so I figured it was better to be more explicit.

There’s already an RPC to delete a project, `DeleteProject`; `pachctl delete repo --all` is specifically intended to delete all repos within a single project. Are you advocating a separate RPC `DeleteRepos`, which takes a request containing a project?

src/server/pfs/server/driver.go
```diff
@@ -2109,6 +2109,27 @@ func (d *driver) addBranchProvenance(txnCtx *txncontext.TransactionContext, bran
...
+	if err := d.listRepo(ctx, !includeAuth, "", func(repoInfo *pfs.RepoInfo) error {
+		if repoInfo.GetRepo().GetProject().GetName() == project.GetName() {
+			repoInfos = append(repoInfos, repoInfo)
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+	return d.txnEnv.WithWriteContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
+		// the list does not use the transaction
```
brendoncarroll, line 2123: :+1:

Dockerfile.pachd
```diff
@@ -1,11 +1,11 @@
 FROM alpine:3.10 AS pachyderm_build
 
 RUN mkdir -p /tmp/to-copy/tmp && chmod -R 777 /tmp/to-copy 
-RUN apk add -U ca-certificates
+RUN apk add -U ca-certificates && update-ca-certificates
 
 FROM scratch
 
-MAINTAINER jdoliner@pachyerm.io
+MAINTAINER support@pachyderm.io
```
brendoncarroll, line 8: https://y.yarn.co/2c88b45d-e612-4d2d-9265-c88e4c08e7d3_text.gif

src/pfs/pfs.proto
brendoncarroll: Can we do a one_of, specify a Repo or a Project?  If you give a Repo, we use that, if you specify a project we'll use that and find all the Repos?  There would be no ambiguous behavior, and force could be outside the one_of.

brendoncarroll (changes requested):
Still need consensus on the external gRPC changes @robert-uhl @acohen4.  My 2 cents is the `one_of`, but if you guys both want something else, that's fine.

src/pfs/pfs.proto
robert-uhl: > Can we do a one_of, specify a Repo or a Project?

Oh, that's interesting. The [protobuf docs do specify](https://developers.google.com/protocol-buffers/docs/proto3#oneof) ‘you can safely move a single field into a new oneof.’

robert-uhl: Ah, but the Go gRPC bindings will change incompatibly.

Maybe it’s cleaner just to add the separate RPC `DeleteRepos` instead and not try to cram multi-repo deletion into `DeleteRepo` after all?

brendoncarroll: That makes sense to me.

src/server/pfs/server/driver.go
```diff
@@ -2109,6 +2109,27 @@ func (d *driver) addBranchProvenance(txnCtx *txncontext.TransactionContext, bran
...
+			repos = append(repos, repoInfo.Repo)
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+	return d.txnEnv.WithWriteContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
+		// the list does not use the transaction
+		for _, repo := range repos {
+			if err := d.deleteRepo(txnCtx, repo, true); err != nil && !auth.IsErrNotAuthorized(err) {
```
albscui, line 2125: A bit out of scope, but I kinda prefer `errors.As(err, &auth.ErrNotAuthorized{})`, since `auth.IsErrNotAuthorized(err)` is using string matching. However, maybe we ought to leave this work to the overall error handling improvement work planned.

robert-uhl:
My latest commit adds a new RPC `DeleteRepos` which takes a `DeleteReposRequest` specifying the project(s) to delete repos from. It returns a list of deleted repos.

src/pfs/pfs.proto
```diff
@@ -578,6 +589,8 @@ service API {
   rpc ListRepo(ListRepoRequest) returns (stream RepoInfo) {}
   // DeleteRepo deletes a repo.
   rpc DeleteRepo(DeleteRepoRequest) returns (google.protobuf.Empty) {}
+  // DeleteRepos deletes more than one repo at once.
+  rpc DeleteRepos(DeleteReposRequest) returns (DeleteReposResponse) {}
```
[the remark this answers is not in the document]
acohen4: I think the semantics should be "delete all repos that I have permissions to delete"? Not sure where we landed on "who gets to delete repos" but if it's a very privileged user it might be that they will always be able to delete all the repos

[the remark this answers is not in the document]
robert-uhl: @brendoncarroll, I believe that that the list of repos returned is exactly the list of repos that are deleted. I’ve added a test for this, but am I missing something?

@acohen4, are you saying that an auth failure when deleting a repo should not be an error? I think that makes sense, but I can also imagine that it should be all-or-nothing. This might be a non-technical question though?

robert-uhl:
This version of `DeleteRepos` deletes all the repos the user is permitted to delete, and properly only returns the actually-deleted repos. There are tests for both the auth and unauth versions.

src/pfs/pfs.proto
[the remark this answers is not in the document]
brendoncarroll: It looks like another caller could delete a repo in between listing them and performing the deletion.
- Either the internal deleteRepo is idempotent and returns nil for already deleted repos, causing us to return repos that weren't deleted by the call, but which existed at the start of the call.
- Or the internal deleteRepo errors on non-existent repos, and the function returns an error (perhaps unnecessarily).

The actual behavior is arguably less important than what we promise with the API.  We can modify the behavior later within the bounds of whatever we promise.  If it happens to be all or nothing, but we don't want to guarantee then let's add a comment here explaining that.

I don't really have strong opinions on the semantics, just that we document what they are.  And consumers of the PFS API don't have to look at the PFS implementation to reason about what's happening.

brendoncarroll (approved):

src/pfs/pfs.proto
```diff
@@ -200,6 +200,17 @@ message DeleteRepoRequest {
   bool force = 2;
 }
 
+// DeleteReposRequest is used to delete more than one repo at once.
+message DeleteReposRequest {
```
acohen4, line 204: this might be obvious, but why not call this DeleteProjects?

src/pfs/pfs.proto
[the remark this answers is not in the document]
acohen4: @brendoncarroll is this still an issue with the updated API semantics? The request doesn't contain the list of repos now, so there won't be an inconsistency between request.repos and response.repos

src/server/pfs/server/api_server.go
```diff
@@ -170,6 +175,21 @@ func (a *apiServer) DeleteRepo(ctx context.Context, request *pfs.DeleteRepoReque
...
+// DeleteRepos implements the pfs.DeleteRepo RPC.  It deletes more than one repo at once.
+func (a *apiServer) DeleteRepos(ctx context.Context, request *pfs.DeleteReposRequest) (resp *pfs.DeleteReposResponse, err error) {
+	if len(request.Projects) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "DeleteRepos must specify project(s) whose repos should be deleted.")
+	}
+	repos, err := a.driver.deleteProjectsRepos(ctx, request.Projects)
+	if err != nil {
+		return nil, err
+	}
+
```
acohen4, line 187: super nit - remove newline

src/server/pfs/server/driver.go
```diff
@@ -2119,6 +2121,44 @@ func (d *driver) addBranchProvenance(txnCtx *txncontext.TransactionContext, bran
...
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	if len(repos) == 0 {
+		return nil, nil
+	}
+
+	if err := d.txnEnv.WithWriteContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
+		// the list does not use the transaction
```
acohen4, line 2145: Ok I'm missing something. Why not do the whole operation in a transaction?

src/pfs/pfs.proto
robert-uhl: Because it deletes repos, not projects☺

src/internal/testpachd/realenv/real_env.go
```diff
@@ -242,6 +243,9 @@ func linkServers(mockServerPtr interface{}, realServer interface{}) {
 			mockPtrValue.Set(mock.Addr())
 
 			useFn := mockPtrValue.MethodByName("Use")
+			if useFn.Kind() == reflect.Invalid {
```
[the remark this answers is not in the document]
robert-uhl: Oh, this was a fun one. So, when adding a new RPC one has to add a type implementing `Use` for the mocks to work. The argument to `Use` varies by the RPC being mocked. Without this check the error message was very opaque; with this error message it should be more obvious to a developer what he needs to do.

The longer-term fix is to revamp how mocks work, but … that is a whole project in itself.

src/server/pfs/server/api_server.go
robert-uhl: I kinda liked it: do some stuff, take a breath, return. But no big deal removing it☺

src/server/pfs/server/driver.go
robert-uhl: It’s the same pattern as `deleteAll` below.

acohen4: I think this should be in a single transaction to support all-or-none semantics

src/pfs/pfs.proto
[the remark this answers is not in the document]
brendoncarroll: All or nothing and partial completion are both fine.  It looks like all or nothing is difficult to implement.  I just think we should write down the guarantees here on the interface in a comment, so autocomplete shows the comment with the semantics to developers.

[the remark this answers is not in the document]
robert-uhl: I added comments spelling out the semantics.

brendoncarroll (approved):
LGTM

src/server/pfs/server/driver.go
acohen4: so I actually re-implemented this function in the provenance PR:
https://github.com/pachyderm/pachyderm/pull/8306/files#diff-a40f20848be655d50f1da1f38f4232133ba02eb4e63e01099055003740955d14R343
In short, we don't need the pachyderm API to support transactions to make operations transactional as long as our internal APIs support transactions.

I'm going to have to merge these changes in anyway, so it's fine if we go with this for now

acohen4 (approved):
LGTM

PR #8394, opened 2022-11-29
Distributed put / get file url

brycemcanally:
This PR distributes the processing of put / get file url requests through the task service. The objects / files to be imported / exported are split into batches that can be processed as tasks through the task service. For put file url, the object prefix is walked to determine the batches. For get file url, the files are iterated through to get the batches. Put file url batches are specified as a list of paths to import. Get file url batches are specified as a file and a path range to export. Both put and get file url use the task service streaming interface to mitigate the impact of stragglers. This implementation has a couple shortcomings that I intend to address in a followup PR that will change our object storage client interface (potentially to this: https://github.com/google/go-cloud). The shortcomings are:
- No path range specification for put file url tasks, which means all of the paths are stored in the task description.
- Not as much benefit for large files since they will still be downloaded serially. We would ideally be able to split large files up into byte ranges and spread them across multiple tasks.

### DIFF
--- a/src/internal/storage/fileset/unordered_writer.go
+++ b/src/internal/storage/fileset/unordered_writer.go
@@ -20,7 +20,6 @@ type UnorderedWriter struct {
 	memAvailable, memThreshold int64
 	fileThreshold              int64
 	buffer                     *Buffer
-	subFileSet                 int64
 	ttl                        time.Duration
 	renewer                    *Renewer
 	ids                        []ID
@@ -136,7 +135,6 @@ func (uw *UnorderedWriter) withWriter(cb func(*Writer) error) error {
 	// Reset fileset buffer.
 	uw.buffer = NewBuffer()
 	uw.memAvailable = uw.memThreshold
-	uw.subFileSet++
 	return nil
 }
 
@@ -188,6 +186,17 @@ func (uw *UnorderedWriter) Copy(ctx context.Context, fs FileSet, datum string, a
 		}
 		return nil
 	}, opts...)
+}
+
+func (uw *UnorderedWriter) AddFileSet(ctx context.Context, id ID) error {
+	if err := uw.serialize(ctx); err != nil {
+		return err
+	}
+	uw.ids = append(uw.ids, id)
+	if uw.renewer != nil {
+		return uw.renewer.Add(uw.ctx, id)
+	}
+	return nil
 }
 
 // Close closes the writer.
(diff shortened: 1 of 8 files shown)


src/server/pfs/server/url.go
```diff
@@ -0,0 +1,207 @@
...
+	defaultURLTaskSize = 1000
+)
+
+func putFileURL(ctx context.Context, taskService task.Service, uw *fileset.UnorderedWriter, dstPath, tag string, src *pfs.AddFile_URLSource) (n int64, retErr error) {
+	url, err := url.Parse(src.URL)
+	if err != nil {
+		return 0, errors.EnsureStack(err)
+	}
+	switch url.Scheme {
+	case "http":
```
FahadBSyed, line 30: [nitpick] I know most of this is copied from `api_server.go`, but these cases could also be combined 
```go
case "https", "http": 
```

FahadBSyed (approved):

src/internal/storage/fileset/unordered_writer.go
```diff
@@ -20,7 +20,6 @@ type UnorderedWriter struct {
 	memAvailable, memThreshold int64
 	fileThreshold              int64
 	buffer                     *Buffer
-	subFileSet                 int64
```
FahadBSyed, line 23: I'm curious to know what this was used for before

src/server/pfs/server/url_worker.go
```diff
@@ -0,0 +1,189 @@
...
+
+func deserializeGetFileURLTask(taskAny *types.Any) (*GetFileURLTask, error) {
+	task := &GetFileURLTask{}
+	if err := types.UnmarshalAny(taskAny, task); err != nil {
+		return nil, errors.EnsureStack(err)
+	}
+	return task, nil
+}
+
+func serializeGetFileURLTaskResult(task *GetFileURLTaskResult) (*types.Any, error) {
```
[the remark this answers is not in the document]
brycemcanally: The linter complained about it being unused, so I went ahead and removed it. We can add it when / if we need it.

src/internal/storage/fileset/unordered_writer.go
brycemcanally: This is a vestige from when we needed a unique and ordered value for each primitive file set. That was how we knew the order of the file set layers. The current model, that has been in place for a while, uses lists both in-memory and in the serialization format.

PR #8395, opened 2022-11-29
[CORE-1227] Rename files and references from v2.4.0 to v2.5.0

robert-uhl:


jrockway:
What's the backstory behind this?  We haven't been doing this every release; only changes that touch the database have been changing this, but I'm not sure that's right either.

robert-uhl:
> What's the backstory behind this?

These migrations add project support; they were originally intended for the v2.4.0 release, hence the original names. The projects feature was stripped out of v2.4.0 and will now be released as part of v2.5.0, so the names were no longer correct.

brendoncarroll (approved):
LGTM.  Took me a second to realize it was a mv.

PR #8398, opened 2022-11-30
envoy: switch to JSON; log listener events

jrockway:
This adds more details to the access log, and they should be self-describing now.

I enabled listener logs to help with requests like https://pachyderm.slack.com/archives/CEY70V55G/p1669756711368999.  If you do something like `echo -e '\r\n' | socat tcp-connect:localhost:80 stdio`, we'll log something like: `{"tx_bytes":0,"duration_ms":0,"response_code":0,"request_attempt_count":0,"downstream_remote_address":"10.244.0.1","connection_id":113,"protocol":"HTTP/1.1","message":"listener log","rx_bytes":3,"response_flags":"-","severity":"info","request_headers_bytes":0,"listener":"proxy-http","timestamp":"2022-11-30T05:21:40.397Z"}`, which might be helpful.


tybritten:
Can we filter out the kube probes from logging?

robert-uhl (approved):

jrockway:
@tybritten Yeah, I'll add a filter to the admin access logger to remove requests from kube-probe.  Will send you a PR for that.

PR #8404, opened 2022-12-01
script to benchmark putfile

jrockway:
This is a small go program that uploads data to Pachyderm and prints how long it takes.

It has expanded in scope to test grpc, http (#8403 protocol), console, s3, and s3-multipart uploads.  It can also negotiate h2c to compare http/1.1 and http/2 locally, talk to TLS servers, and do an integrity check on the upload.


etc/benchmark-uploads/main.go
```diff
@@ -0,0 +1,381 @@
...
+			var idInfo struct {
+				UploadId string `json:"uploadId"`
+			}
+			if err := json.Unmarshal(body, &idInfo); err != nil {
+				return errors.Wrap(err, "unmarshal start reply")
+			}
+
+			// Upload
+			ckSize := 50 * units.MB
+			nCk := int(math.Ceil(float64(length) / float64(ckSize)))
```
msteffen, line 237: not really important, but if you want `ceil` without the float conversion, you can do `(length + ckSize - 1)/ckSize`

If you're interested: http://blog.pkh.me/p/36-figuring-out-round%2C-floor-and-ceil-with-integer-division.html and in particular `⌈n/m⌉ = ⌊(n-1)/m⌋+1 = ⌊(n+m-1)/m⌋`

msteffen (approved):
One question for myself and one piece of trivia; LGTM

etc/benchmark-uploads/main.go
```diff
@@ -0,0 +1,381 @@
...
+		if x, err := humanize.ParseBytes(*size); err != nil {
+			log.Fatalf("parse size: %v", err)
+		} else {
+			r.Len = x
+			length = x
+		}
+
+		n, err := uuid.NewV4()
+		if err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
jrockway: Oversight!  Updated.

etc/benchmark-uploads/main.go
jrockway: That is a really good tip.  I always feel dirty doing the float conversion and never stopped to think if there was a better way of doing that.  I'll have to add this to my muscle memory!

PR #8407, opened 2022-12-02
s3gateway: remove limit on file uploads; make upload timeout less aggressive

jrockway:
People have had trouble uploading files with multipart as it's very slow, but we previously set a limit on single-part uploads.  This removes that limit (but be aware that we buffer the entire file in memory, so people will probably OOM kill pachd when they upload large files, which will be confusing).

We also had a limit of 10 seconds for the entire HTTP request.  That makes no sense to me, so I made it 5 minutes.  I didn't update the timeout on the proxy (which is set to 1 week for S3) just in case I'm misunderstanding this.


smalyala (approved):
LGTM

armaanv (approved):

PR #8409, opened 2022-12-02
Create project role bindings upon auth activation

albscui:


djanicekpach (approved):
looked through the code and tested locally and no longer got the error.

robert-uhl (approved):

PR #8410, opened 2022-12-02
Authorize DeleteProject

albscui:
This PR ensures that only users with the valid permissions to delete a project can delete the project. Note that deleting a project also deletes its role binding.


src/server/pfs/server/driver.go
```diff
@@ -2132,7 +2140,7 @@ func (d *driver) deleteAll(ctx context.Context) error {
 			}
 		}
 		for _, projectInfo := range projectInfos {
-			if err := d.deleteProject(txnCtx, projectInfo.Project, true); err != nil {
+			if err := d.deleteProject(ctx, &pfs.DeleteProjectRequest{Project: projectInfo.Project, Force: true}); err != nil {
```
robert-uhl, line 2143: Hmmm, the change to create the transaction with `deleteProject` means that deleting N projects will create N+1 transactions. I think it would be good to continue to pass in the transaction. It makes the signature unwieldy (`context.Context, txncontext.TransactionContext, …`), but that is life.

src/server/pfs/server/driver.go
```diff
@@ -471,11 +471,19 @@ func (d *driver) listProject(ctx context.Context, cb func(*pfs.ProjectInfo) erro
 }
 
 // TODO: delete all repos and pipelines within project
-func (d *driver) deleteProject(txnCtx *txncontext.TransactionContext, project *pfs.Project, force bool) error {
-	if err := d.projects.ReadWrite(txnCtx.SqlTx).Delete(pfsdb.ProjectKey(project)); err != nil {
-		return errors.Wrapf(err, "delete project %q", project.Name)
+func (d *driver) deleteProject(ctx context.Context, req *pfs.DeleteProjectRequest) error {
```
robert-uhl, line 474: Not a huge fan of passing in the `DeleteProjectRequest` here rather than destructuring to the important bits. On the one hand there is a benefit to using a struct as an argument to act kinda like keyword args, but on the other a function call really ought to read naturally-ish.

PR #8411, opened 2022-12-03
Implement the `log` package

jrockway:
This implements a logger as per the Pachyderm Telemetry design doc.  All existing code has been ported over, and the linter enforces a ban on any future use of "log" or "logrus".

Detailed godoc is available: https://github.com/pachyderm/pachyderm/blob/38042515d79b6abae9415dd45fc72c08c90db99b/src/internal/log/doc.go

Overall, this went smoothly.  Some details:

A lot of tests now take contexts instead of using context.Background(); the upside is that a lot of RealEnv-based tests now have useful log messages.

The gRPC interceptor has improved; logging the start of a rpc, even if a message hasn't been received.  (So if the client connects but fails to send something, we have log evidence.)  This means that streaming RPCs will log a maximum of 4 messages now; stream opened, first message received, first message sent, stream closed.  Previously, it was 2 messages; first message received, last message sent + stream closed.  I think my way is better, but this is all very provisional; the next step is to combine all of our middleware into one interceptor and move it into the pctx package.

The worker logs changed the most.  Most code that could parse them before can parse them now, but there are more fields, so jsonpb specifically chokes.  That's been updated for log (non-Loki and Loki) parsing, but there may be third party tools that can't handle it.  The protobuf spec https://developers.google.com/protocol-buffers/docs/proto3#json says it's optional to allow unknown fields, and I guess we're taking advantage of it.

I tried rewriting the worker to use contexts correctly, instead of swallowing them into the driver, pendingJob, and pachClient objects.  This was not a success.  I spent about 30 hours trying to make it work (some long nights!), and it was very very fragile.  99.9999% of things worked fine, but a couple tests (TestAutoscaling, and the sequence TestJsonnetPipelineTemplate -> TestJsonnetPipelineTemplateMulti) were hard to fix.  If I replicated the exact quirks of the old code, then the Autoscaling tests would fail, but Jsonnet tests would pass.  Remove the hack, Autoscaling passes, Jsonnet fails.  Given how much stuff is going into the next release, we just can't risk it.  So TaggedLogger wraps a context just for logging.  It was a sad end to a fun project, but reality set in and I had to keep things working the old pre-context way.  We will clean up the worker soon.

(BTW, the Jsonnet test failures have nothing to do with jsonnet.  They both use the same input repo and pipeline names, but the pipelines have different behavior, and the test checks the content of the output the pipeline produces.  If the non-multi test is scheduled first, then the multi version starts running, and the foo-pipeline from the non-multi test is still running and happily processes the work.  When the multi test calls wait commit on the output, the output is there, but it's actually from the wrong pipeline.  It detects this and correctly fails.  With the "old" worker code, it properly detects that (how, I have no idea), and doesn't process any work after "pachctl delete all".  The new worker code somehow didn't do that, and we got super lucky this combination of tests tested this undocumented but correct behavior.  The underlying problem is that project-name-version is the only pipeline identifier in the system, so the worker can misunderstand that it's for `foo-pipeline-v1` and not the new `foo-pipeline-v1`.)

The worker now uses the gRPC interceptor to log requests and responses to the worker, but it's at DEBUG level instead of INFO level, so nobody will notice by default.  The worker didn't distinguish between log levels before, but they all work now. 

The server's LOG_LEVEL is propagated to workers now, so debug logging can be enabled.  (I plan to have an RPC to do this at some point in the near future as well.  I also opened a ticket to move some debug logs in the worker to info; spans replaced some explicit log messages in the old code, those were always printed before, but now they're only on in debug mode.)

The hardest part of this PR was rate-limiting logs, which is one of the main reasons we wanted to switch loggers.  There is one place in the code that must never rate limit on logging; printing the output of user code. Zap didn't have a way to switch rate limiting on and off at will.  (You can replace the "core" with one that's not rate-limiting, but you lose all the fields when you do that!  So in the worker logging user code, that's a big problem, because by the time we're there we've added the jobID, pipelineID, data, datumID, etc.)  So, I copied the sampling "zap core" out of zap, and it's now in log/core.go.  Very simple code, so I don't feel bad about forking it, and now the log option `WithoutRateLimit` can turn rate limiting off at any point someone wants.  (You can also turn it back on, but I didn't expose a public method for that.  If it ever comes to that, something else has gone wrong ;)

The biggest quirk of the rewrite is our error handling.  We wrap errors and attach a stack trace in a lot of places.  Apparently, nowhere in pachd did we ever look at the chain of stack traces, but zap logs them by default as the `errorVerbose` key, and they can get HUGE.  It's actually very nice debug info to have, so I'm not super upset, but I have a feeling we'll have to turn them off.  That's a project on my backlog (manual annotation instead of automatic stack traces), but it will be a liiiitle messy until then.

All in all, I'm really happy about the current state.  The API is super easy to use.  The common span-tracking methods don't add another stack frame, which makes stack traces and the code itself easier to read.  The performance is excellent, 2x better than logrus in the worst case (20x faster in the best case).  Especially in code where you want to add a ton of debug messages, the rate limiting really reduces the compute and I/O burden, so it should be very safe to log every iteration of a for loop that we're particularly afraid of.  If the loop is fast, there are 2 log messages.  If it's slow, then you get a log message after the slow thing has caused the rate limit to expire.  Good stuff, I think.

If I were starting a new project, I would definitely want to copy this code over without any reservations.  Although it's new, it feels very battle-tested to me since so much code is now using it.


src/internal/log/span.go
```diff
@@ -0,0 +1,138 @@
...
+		z.Error(msg, fields...)
+		return
+	}
+	z.DPanic("log: internal error: unknown level in Level.log call", zap.Int("level", int(l)))
+	z.Debug(msg, fields...)
+}
+
+// EndSpanFunc is a function that ends a span.
+type EndSpanFunc = func(fields ...Field)
+
```
robert-uhl, line 38: I notice that there are `ErrorL` (error and a level), `Errorp` (error pointer) and `ErrorpL` (error pointer and a level), but no error-without-a-level option, which is a bit surprising — wouldn’t that be the most commonly desired version?

src/internal/log/span.go
```diff
@@ -0,0 +1,138 @@
...
+// span ends.  See ErrorL.
+func ErrorpL(err *error, level Level) Field {
+	// If you're wondering why this ends in "pL" with that capitalization; zap uses the
+	// convention 'lowercase p = pointer', and we use the convention 'uppercase L = level'.
+	f := zap.Reflect("error", err)
+	f.Integer = int64(level)
+	return f
+}
+
+// Errorp is a Field that marks a span as failed.  See ErrorpL.
```
robert-uhl, line 59: Honestly, I expect `Errorp` to be a predicate returning true if its argument is an error, but I get that that’s niche nowadays.

src/internal/log/span.go
```diff
@@ -0,0 +1,138 @@
...
+	f.Integer = int64(level)
+	return f
+}
+
+// ErrorpL is a Field that marks a span as failed if *err is a non-nil error at the time when the
+// span ends.  See ErrorL.
+func ErrorpL(err *error, level Level) Field {
+	// If you're wondering why this ends in "pL" with that capitalization; zap uses the
+	// convention 'lowercase p = pointer', and we use the convention 'uppercase L = level'.
+	f := zap.Reflect("error", err)
```
robert-uhl, line 54: FWIW, I noticed that the [`zap` docs for `Reflect`](https://pkg.go.dev/go.uber.org/zap#Reflect) state ‘outside tests, `Any` is always a better choice *[than `Reflect`]*.’

src/internal/log/span.go
jrockway: Yeah, it's built into zap, so zap.Error.  I'm debating whether or not to alias all of these into our own package or not.  It can't be this one because `Error` is already the logging function.  (Avoided in zap itself by making it a method on the *zap.Logger object.)

src/internal/log/span.go
jrockway: Again, this is zap's convention: 

```
$ go doc zap.Field | egrep 'p\('
func Boolp(key string, val *bool) Field
func Complex128p(key string, val *complex128) Field
func Complex64p(key string, val *complex64) Field
func Durationp(key string, val *time.Duration) Field
func Float32p(key string, val *float32) Field
func Float64p(key string, val *float64) Field
func Int16p(key string, val *int16) Field
func Int32p(key string, val *int32) Field
func Int64p(key string, val *int64) Field
func Int8p(key string, val *int8) Field
func Intp(key string, val *int) Field
func Skip() Field
func StackSkip(key string, skip int) Field
func Stringp(key string, val *string) Field
func Timep(key string, val *time.Time) Field
func Uint16p(key string, val *uint16) Field
func Uint32p(key string, val *uint32) Field
func Uint64p(key string, val *uint64) Field
func Uint8p(key string, val *uint8) Field
func Uintp(key string, val *uint) Field
func Uintptrp(key string, val *uintptr) Field
```

src/internal/log/span.go
jrockway: Basically Reflect decides the type at log message marshaling time, and Any decides at the time you call Any.  We sneak in there and change the internals of the field before this is passed anywhere.  It might be clearest if we `zapcore.Field{Type: ErrorField, Interface: err}` or something.

brendoncarroll:
It looks like we are introducing imports to zap throughout the codebase.  What do you think about aliasing things like `zap.String` in our `log` package as `log.String`?

jrockway:
@brendoncarroll I thought about it, but it would have to be a different package, because `zap.Error` is a field type, and we already have `log.Error` to log an error, so that would be annoying.  I wouldn't want to give the error field a longer name to avoid the collision.  The additional problem is that the `Span` functionality introduces a few new field types, ErrorL (for changing the log level of the "span failed" log message), and there would be some cross-package action at a distance if they're created in a fields package but read in the log package.  So for now, I don't really care if people have to import zap to get at the fields types.  It also saves having to keep our aliases updated if Zap adds new types, though that seems unlikely.

I think the underlying desire you have here is to completely abstract away which logging package we're actually using.  I've taken great care to avoid leaking `*zap.Logger` out of the package for that reason, so we can more easily switch at some point.  The fields being in `zap` do conflict with that goal, but there are so many fields that they will always interfere with the switch.  (I think slog has like Int, Int64, and String or something like that.  zap has like 80 the last time I counted, and implementing the subset of fields we use will be a requirement to switch logging backends.)

The `pctx` stuff you see right now is kind of a work in progress.  It wasn't clear to me what would be needed before I went into the refactor, and the refactor is exposing the cases that are necessary.  So far I've tried to avoid a simple `WithLogger(extractLogger(ctx).Named("XXX"))` function because I feel like that's not actually the API surface that consumers want, but it's very much in flux.

`Span` exists for a quick oneliner to show XXX started / XXX done ok / XXX done with error.  If `Span` returned a context and an end function, then you couldn't `defer Span()()` for a quick span.

`SpanContext` exists because some spans have additional logging, so you can't do the quick and easy `defer` thing.

Both of those functions are intended for relatively short-lived spans; within a single function.

`pctx.BackgroundProcess` and `pctx.WithBackgroundProcess` are a little bit weird and are for longer-lived processes.  There are cases where sometimes a background process / span has no logging relationship with the context that it was created against (`startup` calls `serve`, the servers are their own background processes not related to the `startup` span.)  We have cases in the code where these background processes either use context.Background() or they use some other context (or none at all; a common pattern in the S3 gateway and mount server).

This not the final API or fully fleshed-out yet, but rather a way to make some progress getting contexts injected to where code is doing logging right now, in a way that yields both correct log messages and the correct controlling context.  Once things are passed though I will make another pass at these methods.

The cases that make sense to me are:

1. background processes gated by context.Background (the worker) -> pctx.BackgroundProcess
1. background processes gated by some other context (pachd) -> pctx.WithBackgroundProcess (bad name)
1. combining contexts with servers and requests; an rpc server will have a context to kill it and a background process id (this happens to be a uuid containing a 2 for the 12th byte right now) to track the run, requests against that RPC server will have that context as the parent and their own per-request context (which has a uuid request-id containing a 4 or 1 for the 12th byte); logs from the request will have the server name, server id, and request id on each message.  (no function for this yet.)

The cases that don't make sense to me yet are:

1. combing contexts with log spans; DLock is the most interesting case (Lock takes a context to gate on waiting for the lock, then returns a child context gated on the distributed lock.  the log messages should be `<caller of Lock>.<Lock prefix> want lock`, `<caller of Lock>.<Lock prefix> got lock`, `<caller of Lock>.<Lock prefix> lost lock with reason`, but the returned context should not log as a child logger of `<Lock prefix>`.  this then feeds into things like the PPS master, that want a unique background task id for each "run" of the master (it can exit and restart in the same pachd, or move to another pachd).  not sure how to deal with that yet, but there will be a solution in this PR.
1. long-lived background tasks that can move between pachd servers; tasks basically.  these outlive any context but have an implicit database-managed context that we don't have a data structure for. haven't gotten to these yet, but probably won't be making any dramatic changes in this PR.

src/internal/log/span.go
jrockway: I actually ended up making this its own field type, so there should be no confusion now.  The implementation is a little bit hack (we just choose a number much larger than any existing zapcore field type), but it probably won't be a problem.

brendoncarroll:
RE: the zap Fields.  That makes sense.

> Span exists for a quick oneliner to show XXX started / XXX done ok / XXX done with error. If Span returned a context and an end function, then you couldn't defer Span()() for a quick span.
> SpanContext exists because some spans have additional logging, so you can't do the quick and easy defer thing.

I think Span is orthogonal to the named context tree.  In my mind it exists to time something, not to create a logical subgroup.

I agree with this:
> Both of those functions are intended for relatively short-lived spans; within a single function.


I don't think we should be calling WithBackgroundProcess in many places.  I would do a single call to context.Background() in pachd, then WithLogger, then call my proposed `pctx.Child` for each component that pachd sets up which runs in the background.

For the cases that don't make sense:
1. I would make this explicit rather than trying to pick the one true context.  There is a context for cancelling the call.  That should be the first argument `ctx`, then there is a context which will be wrapped for cancelling the lease.  That could be `bgCtx`.  If you don't have a timeout on the call to DLock then those could be the same, no harm in passing the same thing in 2 places if that's really what's intended.  The semantics would be something like this
```
// DLock blocks until a distributed lock is acquired, or ctx is cancelled. 
// Once the lock is acquired, a child context is spawned from bgCtx using `pctx.Child` with a uniquely generated name.
// That new child context is returned.
func DLock(ctx, bgCtx context.Context) (context.Context, context.CancelFunc, error)
```

2. I'm not sure I understand what you're referring to exactly, but operations which span instances don't need to be represented in the context tree.  Whatever identifier we use can be included as a field on the logger.  One background process in pachd might be a loop that processes tasks.  It could use a child context labeled "worker", for all its calls, and then create additional child contexts for tasks.  The namespacing should be based on how the software is written and how a single instances runs.  Other information representing a global Job/Task tree that spans multiple instances can be stashed as fields in the logger.

jrockway:
There are some unverified commits in there because apparently `git filter-repo` doesn't know how to sign commits.  I ran that to remove my first attempt at refactoring the repo, for a smoother merge if the worker code changes at all between now and when we merge this.

src/internal/pctx/http.go
```diff
@@ -0,0 +1,9 @@
+package pctx
+
+// // ConnContextFn returns a function suitable for use as an (http.Server).ConnContext function.  It
+// // creates a new context containing pctx information and a logging span for each incoming request.
+// func ConnContextFn(ctx context.Context) func(l net.Listener) context.Context {
```
brendoncarroll, line 5: What's going on here?

src/internal/pctx/testing.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/log"
+)
+
+// TestContext returns a context suitable for use as the root context of tests.
+func TestContext(t testing.TB) context.Context {
```
brendoncarroll, line 11: 👍

src/internal/log/testing.go
```diff
@@ -0,0 +1,264 @@
...
+func Test(t testing.TB, opts ...zaptest.LoggerOption) context.Context {
+	l := zaptest.NewLogger(t, opts...)
+	t.Cleanup(zap.ReplaceGlobals(l))
+	t.Cleanup(zap.RedirectStdLog(l))
+	return withLogger(context.Background(), l)
+}
+
+// Test returns a new Context appropriate for use in parallel tests, at the cost of not logging
+// messages sent to the global logger.
+func TestParallel(t testing.TB, opts ...zaptest.LoggerOption) context.Context {
```
brendoncarroll, line 31: My understanding is that the preferred way to access this is through `pctx.TestContext`.  Can we add a comment here saying that.

brendoncarroll:
This looks good.
I think it's forward-looking to replace calls to `log.Child` with `pctx.Child`.  pctx can call `log.Child`, and eventually other things as well.  If you want to do logging you use `log`, if you want to manage contexts you use `pctx`.  I think what you did with `pctx.TestContext` is the right approach and we should take it for `Child` and `Background` as well.  I left a comment about discouraging direct use of `log.TestParallel` and that applies to log.Child and log.Background as well.

I want to minimize the things developers have to understand to write code that is correctly instrumented.  I don't want us to have to make calls to `ptelem.Child` and `ptelem.Background` with every call to `log.Child` and `log.Background` if telemetry becomes its own thing.
Maintaining and iterating on the logging solution will probably require some amount of ramp up on specific knowledge.  But adding structure to the software in the form of contexts trees will part of everyone's day-to-day, and we should make that API as stable as possible.  The best practice could be `pctx.Child` for a long time, but `log.Child`, `telem.Child` `otherthing.Child` `thing2.Child` is a moving target that will lead to code missing some component of instrumentation because the developer at the time wasn't up to date on all the context objects floating around.

src/internal/pctx/http.go
jrockway: Very early in the refactor I wasn't quite sure how I was going to get loggable contexts into the bare HTTP servers.  Ended up doing it in `log` though that will probably move to `pctx` some day.

src/internal/log/testing.go
jrockway: That's correct.  I've updated the existing uses of Test/TestParallel to prefer TestContext, and have updated the docs here appropriately.

jrockway:
@brendoncarroll Child/Background/TODO have been moved into pctx.  Realistically, I don't think telemetry will create any new contexts, but we're somewhat safe from API changes now, so I think that's OK.  Eventually the gRPC interceptor should move into pctx (or sub-package) as well.

jrockway:
All tests remain passing after the pctx refactor.  Based on slack conversation, it sounds like you wanted to hit approve on and there are like 3 changes I need to rebase on top of this, so I'm going to submit.

PR #8412, opened 2022-12-05
add simple metric collector

molinamelendezj:
simple metric collector pipeline that gets added to alpha release tests and records data points in bigquery to compare across runs. this may be a temp solution as we explore other tooling.


etc/testing/circle/workloads/metric-collector/app.py
```diff
@@ -0,0 +1,48 @@
...
+
+    pipelines = list(pachClient.list_pipeline())
+
+    for p in pipelines:
+
+        jobs = list(pachClient.list_job(p.pipeline.name))
+
+        for j in jobs:
+            started = float((str(j.started.seconds) + "." + str(j.started.nanos)))
+            finished = float((str(j.finished.seconds) + "." + str(j.finished.nanos)))
```
djanicekpach, line 35: I'm not sure this is completely right. For example, if nanos is 10(an int32, so no leading 0s), and seconds is 50 you get 50.10 seconds, but .10 seconds is way more than 10 nanos. The protobuf timestamp data type has an example with Python's Timestamp class which seems more reliable [https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp)

djanicekpach: Actually it looks like JobInfo has stats.processTime(as well as upload and download time). You could potentially just use those. Up to you if you want to do this calculation instead, either way works for me.

djanicekpach (changes requested):

etc/testing/circle/workloads/metric-collector/app.py
molinamelendezj: @djanicekpach good catch. actually after thinking a bit more, I want make this collector super simple. so i'm going to simply collect and do calculations/visualization/stats elsewhere. I want to focus back on getting the env creations back to even start running this.

djanicekpach (approved):

PR #8414, opened 2022-12-05
Kill use of gogoproto empty

chainlink:
TODO

- [ ] Double check the diff between gogo empty and google empty
- [ ] Test newer version of pachctl against older one (to verify protos read correctly)


jrockway (approved):
Looks good.  I don't think pachctl will behave differently; the wire format does not change with this (or future) changes.

PR #8417, opened 2022-12-06
[CORE-1347] Fix pachctl verbose logging

tybritten:
picks the correct string that has the error level


tybritten:
Closed- not needed due to https://github.com/pachyderm/pachyderm/pull/8411

PR #8421, opened 2022-12-07
Djanicek/performance pipeline

djanicekpach:
This change adds a CI pipeline and sh script to trigger the Locust performance tests in circle-ci. It runs the tests against a minikube deployment in the circle-ci runner. Once this is merged, I'll set up a nightly test to run these everynight.

to trigger the performance tests, set the pipeline environment variable `run-perf-tests` to true. Here's an example pipeline job: https://app.circleci.com/pipelines/github/pachyderm/pachyderm/16553/workflows/7a0a53ca-91b1-4a2e-bb6f-a994be4137fa

### DIFF
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -1218,61 +1218,7 @@ jobs:
             index=${#prefix}
             additional=${v:index}
             ./etc/build/update_homebrew.sh ${CIRCLE_TAG:1} ${additional}
-  api-performance-test:
-    resource_class: xlarge
-    parameters:
-      version:
-        type: string
-    machine:
-      image: << pipeline.parameters.machine_image >>
-    environment:
-      TEST_RESULTS: /tmp/test-results
-      PACHD_VERSION: << parameters.version >>
-    parallelism: 1
-    steps:
-      - checkout
-      - run:
-          name: Collect node stats
-          command: sar 10 -BbdHwzS -I SUM -n DEV -q -r ALL -u ALL -h
-          background: true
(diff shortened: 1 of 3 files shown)


.circleci/main.yml
```diff
@@ -41,6 +41,9 @@ parameters:
   run-core-jobs:
     type: boolean
     default: false
+  run-perf-tests:
+    type: boolean
```
molinamelendezj, line 45: guessing you can remove all changes from here now

etc/testing/circle/run_api_performance_tests.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+pachctl_tag=$PACHD_VERSION 
+if [ "$PACHD_VERSION" == "latest" ]; then
+    pachctl_tag="v2.5.0-alpha.1" #$(git tag --sort=taggerdate | tail -1) Hard code for the moment until nightly docker images are deploying
```
molinamelendezj, line 7: would be nice to be configurable version?

djanicekpach: It should be using that latest tag code, this was a temp thing. Do you know if/when the images are going to be published to dockerhub in the future? That's the real blocker here.

djanicekpach: It deployed last night, so this should be good to go! I removed the stop gap.

molinamelendezj (approved):

PR #8422, opened 2022-12-08
[CORE-1335] Correct nil pointer dereference from ListFile

robert-uhl:


src/server/auth/server/testing/auth_integration_test.go
```diff
@@ -912,3 +912,56 @@ func TestDeleteAllAfterDeactivate(t *testing.T) {
...
+		t.Skip("Skipping integration tests in short mode")
+	}
+	t.Parallel()
+	c, _ := minikubetestenv.AcquireCluster(t, defaultTestOptions)
+	tu.ActivateAuthClient(t, c)
+	alice := tu.Robot(tu.UniqueString("alice"))
+	aliceClient := tu.AuthenticateClient(t, c, alice)
+	repo := "foo"
+	require.NoError(t, aliceClient.CreateProjectRepo(pfs.DefaultProjectName, repo))
+	var commit *pfs.Commit
```
jrockway, line 927: This is fine to check in now, but I think you'd enjoy this as a table-driven test:
```go
testData := []struct{name string; commit *pfs.Commit}
   {name: "nil commit"},
   {name: "empty commit", commit: &pfs.Commit{}},
    ...
} 
for _, test := range testData {
   t.Run(test.name, func(t *testing.T) { 
       err := aliceClient.ListFile(test.commit, "/", func(_ *pfs.FileInfo) { t.Fatal("unreachable") })
       if err == nil { t.Error("unexpected success") }
   }
}
```
It would also be good to extend this to any other RPC that takes a commit; maybe you can kill a lot of birds with one stone here.  Doesn't need to be done now, but consider it for the backlog.

jrockway (approved):

src/server/auth/server/testing/auth_integration_test.go
robert-uhl: It’s a good idea, and quick enough to do!

jrockway (approved):
Nice.  We will get you on the test case name train someday ;)

PR #8424, opened 2022-12-08
Run Pachd in 'PachW' Mode and Autoscale To Handle Tasks

FahadBSyed:
Today every PFS worker instance runs a worker routine to process background storage tasks from the task service. This is wasteful as PFS workers could be scheduled on expensive nodes that have GPUs. Instead of coupling the background storage routines with the PFS workers, this proposal suggests defining a new pachd mode ‘pachw’ that would process background storage tasks from the task service. The master pachd instance would deploy pachw mode workers as separate pods and scale them up and down to match the volume of tasks.

https://docs.google.com/document/d/1GF1zA6XyOKWldQ6QQ9raE7k4lnI5AZs_3jHpWLtHF-M/edit?usp=sharing

### DIFF
--- a/etc/helm/pachyderm/templates/pachd/rbac/role.yaml
+++ b/etc/helm/pachyderm/templates/pachd/rbac/role.yaml
@@ -53,8 +53,6 @@ rules:
   resources:
   - deployments
   - deployments/scale
-  - replicasets
-  - replicasets/scale
   verbs:
   - get
   - list
--- a/etc/helm/pachyderm/templates/pachd/rbac/role.yaml
+++ b/etc/helm/pachyderm/templates/pachd/rbac/role.yaml
@@ -53,6 +53,8 @@ rules:
   resources:
   - deployments
   - deployments/scale
+  - replicasets
+  - replicasets/scale
   verbs:
   - get
   - list
--- a/src/internal/minikubetestenv/deploy.go
+++ b/src/internal/minikubetestenv/deploy.go
@@ -243,7 +243,6 @@ func withEnterprise(host, rootToken string, issuerPort, clientPort int) *helm.Op
 
 func withEnterpriseServer(image, host string) *helm.Options {
 	return &helm.Options{SetValues: map[string]string{
-		"pachw.enabled":                      "false",
 		"pachd.enabled":                      "false",
 		"enterpriseServer.enabled":           "true",
 		"enterpriseServer.image.tag":         image,
--- a/src/internal/minikubetestenv/deploy.go
+++ b/src/internal/minikubetestenv/deploy.go
@@ -501,8 +501,8 @@ func putRelease(t testing.TB, ctx context.Context, namespace string, kubeClient 
 	} else {
 		helmOpts = union(helmOpts, withPachd(version))
 		// TODO(acohen4): apply minio deployment to this namespace
-		helmOpts = union(helmOpts, withMinio())
 	}
+	helmOpts = union(helmOpts, withMinio())
 	if opts.PortOffset != 0 {
 		pachAddress.Port += opts.PortOffset
 		helmOpts = union(helmOpts, withPort(namespace, pachAddress.Port, opts.TLS))
--- a/src/internal/minikubetestenv/deploy.go
+++ b/src/internal/minikubetestenv/deploy.go
@@ -243,6 +243,7 @@ func withEnterprise(host, rootToken string, issuerPort, clientPort int) *helm.Op
 
 func withEnterpriseServer(image, host string) *helm.Options {
 	return &helm.Options{SetValues: map[string]string{
+		"pachw.enabled":                      "false",
 		"pachd.enabled":                      "false",
 		"enterpriseServer.enabled":           "true",
 		"enterpriseServer.image.tag":         image,
@@ -497,12 +498,13 @@ func putRelease(t testing.TB, ctx context.Context, namespace string, kubeClient 
 	}
 	if opts.EnterpriseServer {
 		helmOpts = union(helmOpts, withEnterpriseServer(version, pachAddress.Host))
+		helmOpts = union(helmOpts, withMinio())
 		pachAddress.Port = uint16(31650)
 	} else {
 		helmOpts = union(helmOpts, withPachd(version))
 		// TODO(acohen4): apply minio deployment to this namespace
+		helmOpts = union(helmOpts, withMinio())
 	}
-	helmOpts = union(helmOpts, withMinio())
 	if opts.PortOffset != 0 {
 		pachAddress.Port += opts.PortOffset
 		helmOpts = union(helmOpts, withPort(namespace, pachAddress.Port, opts.TLS))
--- a/src/internal/miscutil/math.go
+++ b/src/internal/miscutil/math.go
@@ -9,3 +9,11 @@ func Min[T constraints.Ordered](a, b T) T {
 	}
 	return b
 }
+
+// Max returns the larger of a and b.
+func Max[T constraints.Ordered](a, b T) T {
+	if a > b {
+		return a
+	}
+	return b
+}
--- a/src/internal/serviceenv/config.go
+++ b/src/internal/serviceenv/config.go
@@ -28,9 +28,9 @@ type GlobalConfiguration struct {
 	LokiPort                       string `env:"LOKI_SERVICE_PORT"`
 	OidcPort                       uint16 `env:"OIDC_PORT,default=1657"`
 	IsPachw                        bool   `env:"IS_PACHW,default=false"`
-	PachwEnabled                   bool   `env:"PACHW_ENABLED,default=true"`
-	PachwMinReplicas               int    `env:"PACHW_MIN_REPLICAS,default=0"`
-	PachwMaxReplicas               int    `env:"PACHW_MAX_REPLICAS"`
+	PachwInSidecars                bool   `env:"PACHW_IN_SIDECARS,default=false"`
+	PachwMinReplicas               int    `env:"PACHW_MIN_REPLICAS"`
+	PachwMaxReplicas               int    `env:"PACHW_MAX_REPLICAS,default=1"`
 	PGBouncerHost                  string `env:"PG_BOUNCER_HOST,required"`
 	PGBouncerPort                  int    `env:"PG_BOUNCER_PORT,required"`
 	PostgresSSL                    string `env:"POSTGRES_SSL,default=disable"`
--- a/src/internal/serviceenv/config.go
+++ b/src/internal/serviceenv/config.go
@@ -28,13 +28,16 @@ type GlobalConfiguration struct {
 	LokiPort                       string `env:"LOKI_SERVICE_PORT"`
 	OidcPort                       uint16 `env:"OIDC_PORT,default=1657"`
 	IsPachw                        bool   `env:"IS_PACHW,default=false"`
+	PachwEnabled                   bool   `env:"PACHW_ENABLED,default=true"`
+	PachwMinReplicas               int    `env:"PACHW_MIN_REPLICAS,default=0"`
+	PachwMaxReplicas               int    `env:"PACHW_MAX_REPLICAS"`
 	PGBouncerHost                  string `env:"PG_BOUNCER_HOST,required"`
 	PGBouncerPort                  int    `env:"PG_BOUNCER_PORT,required"`
 	PostgresSSL                    string `env:"POSTGRES_SSL,default=disable"`
 	PostgresHost                   string `env:"POSTGRES_HOST"`
 	PostgresPort                   int    `env:"POSTGRES_PORT"`
-	PostgresDBName                 string `env:"POSTGRES_DATABASE"` //made this non-required, hopefully it doesn't explode
-	PostgresUser                   string `env:"POSTGRES_USER"`     //made this non-required, hopefully it doesn't explode
+	PostgresDBName                 string `env:"POSTGRES_DATABASE"`
+	PostgresUser                   string `env:"POSTGRES_USER"`
 	PostgresPassword               string `env:"POSTGRES_PASSWORD"`
 	PostgresMaxOpenConns           int    `env:"POSTGRES_MAX_OPEN_CONNS,default=10"`
 	PostgresMaxIdleConns           int    `env:"POSTGRES_MAX_IDLE_CONNS,default=10"`
--- a/src/internal/serviceenv/config.go
+++ b/src/internal/serviceenv/config.go
@@ -27,13 +27,14 @@ type GlobalConfiguration struct {
 	LokiHost                       string `env:"LOKI_SERVICE_HOST"`
 	LokiPort                       string `env:"LOKI_SERVICE_PORT"`
 	OidcPort                       uint16 `env:"OIDC_PORT,default=1657"`
+	IsPachw                        bool   `env:"IS_PACHW,default=false"`
 	PGBouncerHost                  string `env:"PG_BOUNCER_HOST,required"`
 	PGBouncerPort                  int    `env:"PG_BOUNCER_PORT,required"`
 	PostgresSSL                    string `env:"POSTGRES_SSL,default=disable"`
 	PostgresHost                   string `env:"POSTGRES_HOST"`
 	PostgresPort                   int    `env:"POSTGRES_PORT"`
-	PostgresDBName                 string `env:"POSTGRES_DATABASE,required"`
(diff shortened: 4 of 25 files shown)


src/internal/minikubetestenv/deploy.go
```diff
@@ -494,6 +495,7 @@ func putRelease(t testing.TB, ctx context.Context, namespace string, kubeClient
 	}
 	if opts.EnterpriseServer {
 		helmOpts = union(helmOpts, withEnterpriseServer(version, pachAddress.Host))
+		helmOpts = union(helmOpts, withMinio())
```
FahadBSyed, line 498: I want to check with Alon when he comes back to see if this is okay. We currently need this to set 'deployTarget' when running in enterprise server mode. The deployTarget gets set in 'withMinio()'

src/server/pachw/server/pachw.go
```diff
@@ -0,0 +1,152 @@
...
+
+func NewMaster(ctx context.Context, env *Env) {
+	master := &pachW{
+		env: *env,
+	}
+	go master.master(ctx)
+}
+
+func (p *pachW) master(ctx context.Context) {
+	ctx = auth.AsInternalUser(ctx, "pfs-master")
```
brycemcanally, line 37: pfs-master -> pachw-master, I guess.

src/server/pachw/server/pachw.go
```diff
@@ -0,0 +1,152 @@
...
+	backoff.RetryUntilCancel(ctx, func() error { //nolint:errcheck
+		eg, ctx := errgroup.WithContext(ctx)
+		eg.Go(func() error {
+			lock := dlock.NewDLock(p.env.EtcdClient, path.Join(p.env.EtcdPrefix, "pachw-master-lock"))
+			defer func() {
+				if err := lock.Unlock(ctx); err != nil {
+					p.env.Logger.Errorf("error unlocking in pfs master: %v", err)
```
brycemcanally, line 44: pfs -> pachw

src/server/pachw/server/pachw.go
```diff
@@ -0,0 +1,152 @@
...
+	tickerSeconds    = 5
+	samplesPerMinute = 60 / tickerSeconds
+)
+
+type pachW struct {
+	env Env
+}
+
+func NewMaster(ctx context.Context, env *Env) {
+	master := &pachW{
```
brycemcanally, line 30: We should create a new logger with default fields for this component.

src/testing/deploy/deploy_test.go
```diff
@@ -43,7 +43,7 @@ func TestInstallAndUpgradeEnterpriseWithEnv(t *testing.T) {
 	// set new root token via env
 	opts.AuthUser = ""
 	token := "new-root-token"
-	opts.ValueOverrides = map[string]string{"pachd.rootToken": token}
+	opts.ValueOverrides = map[string]string{"pachd.rootToken": token, "pachw.enabled": "false"}
```
brycemcanally, line 46: Why are we using false here?

src/internal/serviceenv/service_env.go
```diff
@@ -194,7 +194,7 @@ func (env *NonblockingServiceEnv) Config() *Configuration {
 }
 
 func (env *NonblockingServiceEnv) isWorker() bool {
-	return env.config.PPSPipelineName != ""
+	return env.config.PPSPipelineName != "" || env.config.IsPachw
```
brycemcanally, line 197: We cannot use the pipeline name field for this?

src/internal/serviceenv/config.go
```diff
@@ -25,13 +25,17 @@ type GlobalConfiguration struct {
 	LokiHost                       string `env:"LOKI_SERVICE_HOST"`
 	LokiPort                       string `env:"LOKI_SERVICE_PORT"`
 	OidcPort                       uint16 `env:"OIDC_PORT,default=1657"`
+	IsPachw                        bool   `env:"IS_PACHW,default=false"`
+	PachwEnabled                   bool   `env:"PACHW_ENABLED,default=true"`
+	PachwMinReplicas               int    `env:"PACHW_MIN_REPLICAS,default=0"`
```
brycemcanally, line 30: I don't think setting the default to zero is needed since that is the default value for an int.

src/internal/serviceenv/service_env.go
FahadBSyed: We could, but since its not a pipeline, I felt like that would be hacky or misleading and opted to use a new config variable.

src/testing/deploy/upgrade_test.go
```diff
@@ -48,6 +49,7 @@ func upgradeTest(suite *testing.T, ctx context.Context, fromVersions []string, p
 					// overrides.  They became the default (instead of random)
 					// in the 2.3 alpha cycle.
 					ValueOverrides: map[string]string{
+						"pachw.enabled":                                "false",
```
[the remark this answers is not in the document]
FahadBSyed: this was a hack to work around an error in the upgrade load test. I think I'm going to dig into this today and see why it causes an EOF if this isn't supplied.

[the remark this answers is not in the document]
FahadBSyed: So, I rebased off of master and that seems to have fixed whatever was causing this problem

src/internal/serviceenv/service_env.go
brycemcanally: Ohh, right. We want it to be considered a worker to avoid a direct connection to postgres. That makes sense.

src/server/pachw/server/env.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/sirupsen/logrus"
+	etcd "go.etcd.io/etcd/client/v3"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/serviceenv"
+	"github.com/pachyderm/pachyderm/v2/src/internal/task"
+)
+
+// Env is the dependencies needed to run the pachW Master
```
brycemcanally, line 15: controller

src/server/pachw/server/env.go
```diff
@@ -0,0 +1,46 @@
...
+type Env struct {
+	EtcdPrefix        string
+	EtcdClient        *etcd.Client
+	TaskService       task.Service
+	KubeClient        kubernetes.Interface
+	Namespace         string
+	Enabled           bool
```
brycemcanally, line 22: Is this field needed?

FahadBSyed: Good catch, yeah I don't think its actually used by the pachw code

FahadBSyed: Actually, I use this in the builder code to determine whether to spin up a master or not in maybeInitPachwController(). For some reason my IDE was showing me 0 usages.

FahadBSyed: So instead of using enabled, I can just check the pachwEnabled flag on the Env struct.

src/server/pachw/server/pachw.go
```diff
@@ -0,0 +1,154 @@
...
+	backoff.RetryUntilCancel(ctx, func() error { //nolint:errcheck
+		eg, ctx := errgroup.WithContext(ctx)
+		eg.Go(func() error {
+			lock := dlock.NewDLock(p.env.EtcdClient, path.Join(p.env.EtcdPrefix, "pachw-controller-lock"))
+			defer func() {
+				if err := lock.Unlock(ctx); err != nil {
+					p.log.Errorf("error unlocking in pachw controller: %v", err)
+				}
+			}()
+			totalInstances := 0
```
brycemcanally, line 50: Could we change this to look more like this:
```
var replicas int
var scaleDownCount int
ticker := time.NewTicker(period)
defer ticker.Stop()
for {
	numTasks := p.countTasks(ctx, []string{server.URLTaskNamespace, server.StorageTaskNamespace})
	scaleDownCount++
	if numTasks > replicas/2 {
		scaleDownCount = 0
	}
	if numTasks > replicas {
		replicas = math.Min(numTasks, p.env.MaxReplicas)
		p.log.Infof("scaling up pachw workers to %d replicas", replicas)
		if err := p.scalePachw(ctx, int32(replicas)); err != nil {
			p.log.Errorf("failed to scale up pachw: %v", err)
		}
	} else if scaleDownCount == scaleDownThreshold {
		replicas = math.Max(replicas / 2, p.env.MinReplicas)
		p.log.Infof("scaling down pachw workers to %d replicas", replicas)
		if err := p.scalePachw(ctx, int32(replicas)); err != nil {
			p.log.Errorf("failed to scale down pachw: %v", err)
		}
	}
	select {
	case <-ticker.C:
	case <-ctx.Done():
		return errors.EnsureStack(ctx.Err())
	}
}
```
I would like this code to be as simple / readable as possible.

src/server/pachw/server/pachw.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/internal/backoff"
+	"github.com/pachyderm/pachyderm/v2/src/internal/dlock"
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/middleware/auth"
+	"github.com/pachyderm/pachyderm/v2/src/internal/miscutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/task"
+	"github.com/pachyderm/pachyderm/v2/src/server/pfs/server"
+)
+
+const (
```
brycemcanally, line 21: I think this would be clearer:
```
const (
	period             = 5 * time.Second
	scaleDownThreshold = 12
)
```

src/internal/miscutil/math.go
```diff
@@ -9,3 +9,11 @@ func Min[T constraints.Ordered](a, b T) T {
 	}
 	return b
 }
+
+// Max returns the larger of a and b.
+func Max[T constraints.Ordered](a, b T) T {
```
brycemcanally, line 14: lol

etc/helm/pachyderm/templates/pachd/rbac/role.yaml
```diff
@@ -53,6 +53,8 @@ rules:
   resources:
   - deployments
   - deployments/scale
+  - replicasets
```
brycemcanally, line 56: Is this still needed after the move to deployments?

src/server/pachw/server/pachw.go
```diff
@@ -0,0 +1,117 @@
...
+		env: *env,
+		log: env.Logger,
+	}
+	go controller.run(ctx)
+}
+
+func (p *pachW) run(ctx context.Context) {
+	ctx = auth.AsInternalUser(ctx, "pachw-controller")
+	backoff.RetryUntilCancel(ctx, func() error { //nolint:errcheck
+		eg, ctx := errgroup.WithContext(ctx)
```
brycemcanally, line 43: Why is this errgroup needed?

brycemcanally: Also, it looks like we are not returning errors in a few places, just logging them. We should return errors within the backoff callback and lean on the logging in the backoff.

src/server/pachw/server/pachw.go
```diff
@@ -0,0 +1,117 @@
...
+		return errors.Wrap(err, "could not get pachw replicaSet")
+	}
+	pachwRs.Spec.Replicas = &replicas
+	_, err = p.env.KubeClient.AppsV1().Deployments(p.env.Namespace).Update(
+		ctx,
+		pachwRs,
+		v1.UpdateOptions{},
+	)
+	if err != nil {
+		return errors.Wrap(err, "could not scale pachw replicaSet")
```
brycemcanally, line 114: replicaSet -> deployment

src/server/pachw/server/pachw.go
```diff
@@ -0,0 +1,117 @@
...
+		totalTasks += int(tasks)
+	}
+	return totalTasks
+}
+
+func (p *pachW) scalePachw(ctx context.Context, replicas int32) error {
+	pachwRs, err := p.env.KubeClient.AppsV1().Deployments(p.env.Namespace).Get(ctx,
+		"pachw", v1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(err, "could not get pachw replicaSet")
```
brycemcanally, line 105: Same as above, or below if looking at the code.

src/server/pfs/server/server.go
```diff
@@ -22,5 +24,21 @@ func NewSidecarAPIServer(env Env) (pfsserver.APIServer, error) {
...
+	return newValidatedAPIServer(a, env.AuthServer), nil
+}
+
+// NewPachwAPIServer is used for Pachw Mode. Pachw worker pool instances to process background tasks from the task service.
+func NewPachwAPIServer(env Env) (pfsserver.APIServer, error) {
+	a, err := newAPIServer(env)
+	if err != nil {
+		return nil, err
+	}
+	go func() { pfsload.Worker(env.GetPachClient(env.BackgroundContext), env.TaskService) }() //nolint:errcheck
```
brycemcanally, line 40: Shouldn't there be a URL worker here? Also, the above comment reads a bit weird.

src/server/pachw/server/pachw.go
FahadBSyed: I think this is prone to cause issues if we change the period, I can see a bug happening where we forget to change the scaleDownThreshold

etc/helm/pachyderm/templates/pachd/rbac/role.yaml
FahadBSyed: no its not, good catch

src/server/pachw/server/pachw.go
FahadBSyed: I think we want an error group here so we can return the error from the goroutine. 

Not returning errors inside is a good point, I forgot that this was happening in a backoff.

src/server/pfs/server/server.go
FahadBSyed: Good catch

src/server/pachw/server/pachw.go
FahadBSyed: This is partially resolved: Errors are returned, but we still are using an error group.

src/server/pachw/server/pachw.go
brycemcanally: Okay, if you want to keep the period and scale down time independent, then what you have is fine.

src/server/pachw/server/pachw.go
brycemcanally: I don't think it needs to be in a goroutine, you can just run the code in the backoff callback. Also, we should capture the error during unlock with a multierror and the second scale call is missing a error return.

FahadBSyed: That's fair, I guess right now the pachw controller doesn't need to process anything else, so it can just run this routine in the main thread.

src/server/pachw/server/pachw.go
```diff
@@ -0,0 +1,117 @@
...
+	go controller.run(ctx)
+}
+
+func (p *pachW) run(ctx context.Context) {
+	ctx = auth.AsInternalUser(ctx, "pachw-controller")
+	backoff.RetryUntilCancel(ctx, func() (retErr error) { //nolint:errcheck
+		lock := dlock.NewDLock(p.env.EtcdClient, path.Join(p.env.EtcdPrefix, "pachw-controller-lock"))
+		defer func() {
+			if err := lock.Unlock(ctx); err != nil {
+				unlockErr := errors.Wrap(err, "error unlocking in pachw controller")
```
brycemcanally, line 46: Could we use this pattern instead: https://github.com/pachyderm/pachyderm/blob/80582159e2e13ba0cb354c0d1a40fc60afd63e02/src/server/pfs/server/master.go#L131. I think the pattern of just returning errors in the callback and using the backoff for logging is a bit cleaner. This would also apply to the places below where we log as well.

src/server/pachw/server/pachw.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil
+	})
+}
+
+func (p *pachW) countTasks(ctx context.Context, namespaces []string) int {
+	totalTasks := 0
+	for _, ns := range namespaces {
+		tasks, _, err := task.Count(ctx, p.env.TaskService, ns, "")
+		if err != nil {
+			p.log.Errorf("error counting tasks: %v", err)
```
brycemcanally, line 94: We should just return an error here.

brycemcanally (approved):
lgtm

PR #8425, opened 2022-12-08
Update version of jaeger all-in-one image

msteffen:
Also clean up manifest comments

### DIFF
--- a/etc/deploy/tracing/jaeger-all-in-one.yaml
+++ b/etc/deploy/tracing/jaeger-all-in-one.yaml
@@ -1,14 +1,19 @@
 # copied from
 # https://github.com/jaegertracing/jaeger-kubernetes/blob/bcca10c297b38c01d9d7c554172572ffa11944ce/all-in-one/jaeger-all-in-one-template.yml
 # with minor changes:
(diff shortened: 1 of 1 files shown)


etc/deploy/tracing/jaeger-all-in-one.yaml
```diff
@@ -1,14 +1,19 @@
 # copied from
 # https://github.com/jaegertracing/jaeger-kubernetes/blob/bcca10c297b38c01d9d7c554172572ffa11944ce/all-in-one/jaeger-all-in-one-template.yml
```
jrockway, line 2: Should probably upgrade to the latest version of this manifest.

jrockway (approved):

etc/deploy/tracing/jaeger-all-in-one.yaml
msteffen: Ah, I'll change this comment, but I did pull the latest as part of creating this PR and nothing had changed!

jrockway: Oh excellent.  You can submit this without waiting for tests, I don't think the tests test this.

PR #8434, opened 2022-12-09
helm: enable proxy by default; add NOTE about externalService deprecation

jrockway:
This is CORE-1307.

The NOTE might be too long, but it seems helpful to me.  If you are migrating from external service, it will print:

```
WARNING: pachd.externalService is deprecated in favor of proxy.enabled.  You can transparently
migrate by exposing the proxy and removing the external service:

  pachd:
    externalService: false

  proxy:
    enabled: true
    host: 192.168.1.70
    service:
      type: LoadBalancer
      loadBalancerIP: "<the IP provisioned for the load balancer; not setting this is OK if you don't know it>"
      legacyPorts:
        grpc: 30650
        s3Gateway: 30600
        oidcPort: 0
        identityPort: 0

Please see values.yaml for a full list of configuration options, including SSL/TLS/HTTPS support.

The legacy ports are not strictly necessary except for compatability with your existing
configuration.  The proxy inspects traffic arriving on normal http or https ports (80 and 443) and
automatically routes it to the right service.  pachd, console, S3, oidc, and identity will be
available on port 80 (grpc://<proxy.host>:80, http://<proxy.host>) if TLS is disabled, or port 443
otherwise (grpcs://<proxy.host>:443, https://<proxy.host>).

As of 2.5.0, the proxy is enabled by default, but not exposed to the Internet.  You can test it out
with:

    kubectl port-forward svc/pachyderm-proxy 80

and then visiting http://localhost in the browser (or grpc://localhost:80 with pachctl).

Your deployment's security posture will improve by enabling the proxy.  It is hardened against
malicious traffic, and writes out extensive audit logs for all requests to Pachyderm.
```

(Everything that can be pre-filled from existing values is pre-filled there.  We hard code LoadBalancer because the old externalService did.)

After you adjust your configuration to have a working proxy setup, you get:

```
Start using your Pachyderm deployment by visiting the Console at http://192.168.1.70 in your browser.

Connect pachctl by running:

    echo '{"pachd_address":"grpc://192.168.1.70:80"}' | pachctl config set context pachyderm --overwrite && pachctl config set active-context pachyderm
```

If your helm release name is `$(rm / -rf)` i guess you're gonna have a bad time if you copypaste that, but I'm sure helm won't let you do that 😛 

The messages adjust depending on `proxy.tls.enabled`.

One sticky bit here is that if users manually configure issuerURI and the URL doesn't end in /dex, their working configuration will stop working (rather they won't be able to helm upgrade).  Any thoughts on that @tybritten?

### DIFF
--- a/etc/helm/pachyderm/templates/NOTES.txt
+++ b/etc/helm/pachyderm/templates/NOTES.txt
@@ -18,7 +18,7 @@ WARNING: Your cluster is configured with a default login of admin:password
 WARNING: You have pachd.enterpriseRootToken or pachd.enterpriseRootTokenSecretName set. These values are being replaced by pachd.enterpriseServerToken and pachd.enterpriseServerTokenSecretName which you can set now. The deprecated values will be removed in version 2.4.0.
 {{- end -}}
 
-{{ if .Values.pachd.externalService.enabled and (eq .Values.proxy.service.type "ClusterIP")}}
+{{ if and .Values.pachd.externalService.enabled (eq .Values.proxy.service.type "ClusterIP")}}
 WARNING: pachd.externalService is deprecated in favor of proxy.enabled.  You can transparently
 migrate by exposing the proxy and removing the external service:
 
--- a/etc/helm/pachyderm/templates/NOTES.txt
+++ b/etc/helm/pachyderm/templates/NOTES.txt
@@ -64,3 +64,20 @@ and then visiting http://localhost in the browser (or grpc://localhost:80 with p
 Your deployment's security posture will improve by enabling the proxy.  It is hardened against
 malicious traffic, and writes out extensive audit logs for all requests to Pachyderm.
 {{- end -}}
+
+
+{{- if and .Values.proxy.enabled (ne .Values.proxy.service.type "ClusterIP") (ne .Values.proxy.host "")}}
+{{- if .Values.proxy.tls.enabled }}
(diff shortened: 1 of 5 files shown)


etc/helm/pachyderm/templates/NOTES.txt
```diff
@@ -18,3 +18,66 @@ WARNING: Your cluster is configured with a default login of admin:password
...
+{{- end -}}
+
+
+{{- if and .Values.proxy.enabled (ne .Values.proxy.service.type "ClusterIP") (ne .Values.proxy.host "")}}
+{{- if .Values.proxy.tls.enabled }}
+Start using your Pachyderm deployment by visiting the Console at https://{{.Values.proxy.host}} in you browser.
+
+Connect pachctl by running:
+
+    echo '{"pachd_address":"grpc://{{ .Values.proxy.host }}:443"}' | pachctl config set context {{ .Release.Name }} --overwrite && pachctl config set active-context {{ .Release.Name }}
```
tybritten, line 75:
```suggestion
    echo '{"pachd_address":"grpcs://{{ .Values.proxy.host }}:443"}' | pachctl config set context {{ .Release.Name }} --overwrite && pachctl config set active-context {{ .Release.Name }}
```
grpcs

tybritten (changes requested):
LGTM, just that one typo

jrockway:
PTAL

tybritten (approved):
LGTM

PR #8437, opened 2022-12-12
2.4.2 tag bump for notebooks-user

lbliii:
this is a temporary fix for the example used in the docs. let me know when we're ready to replace it!


msteffen (approved):
LGTM—thanks!

PR #8440, opened 2022-12-13
move pfs cmds tests off kubernetes

djanicekpach:
The only thing preventing the pfs command tests from running was the fact that pachctl calls inspectCluster, which normally needs a cluster. This PR adds a mockfor inspectCluster that just returns the IDs it expects so the rest of the test can run. 

It took 90-120 seconds to create a minikube cluster for testing this file, now on my machine the entire file runs in 30 seconds, so this seems worthwhile.

I left TestMount using minikube, and moved it to a file called cmds_integration_test.go. I tried to convert this, but it could not find branches after mounting without kubernetes. I suspect it either needs real storage or there's some permissions thing going on, but I timeboxed my investigation since this was starting to take a while to dig into.

### DIFF
--- a/src/server/pfs/cmds/cmds_integration_test.go
+++ b/src/server/pfs/cmds/cmds_integration_test.go
@@ -1,0 +1,130 @@
+//go:build k8s
+
+package cmds
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/cenkalti/backoff"
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/minikubetestenv"
+	"github.com/pachyderm/pachyderm/v2/src/internal/require"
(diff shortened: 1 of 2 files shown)


src/server/pfs/cmds/cmds_integration_test.go
```diff
@@ -0,0 +1,130 @@
...
+package cmds
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/cenkalti/backoff"
```
jrockway, line 12: This is the wrong backoff library.  I think you wanted src/internal/backoff.

jrockway (approved):
LGTM

src/server/pfs/cmds/cmds_integration_test.go
djanicekpach: Turns out that is the one the test was using before. The code didn't compile with the internal library either :/ 
I tried to adopt the internal library with a few tweaks, since I assume we prefer internal over external libraries. It's small but, I'd appreciate it if you could sanity check this since I really haven't worked with go Contexts before.

PR #8441, opened 2022-12-13
automate console release from core repo

molinamelendezj:
proposing this circle ci job in the the release workflow. this simply automates the steps we manually take today. with the assumption console team follows the same git branch strategy as core. the `sign-off` job is not dependent on it for now, should there be failures, manual intervention can still be done.


.circleci/main.yml
```diff
@@ -1163,7 +1163,61 @@ jobs:
...
+                  echo "regular release."
+                  parentBranch=${CIRCLE_TAG:1}
+              fi
+            git clone https://github.com/pachyderm/haberdashery.git haberdashery
+            git checkout -b "release-${CIRCLE_TAG:1}"
+            tmp=$(mktemp)
+            jq --arg val "${CIRCLE_TAG:1}-${CIRCLE_SHA1}" '.pachyderm = $val' version.json > "$tmp" && mv "$tmp" version.json
+            jq --arg val "${CIRCLE_SHA1}" '.pachReleaseCommit = $val' version.json > "$tmp" && mv "$tmp" version.json
+            jq --arg val "${CIRCLE_TAG:1}" '.pachctl = $val' version.json > "$tmp" && mv "$tmp" version.json
+            jq --arg val "${CIRCLE_SHA1}" '.proto = $val' version.json > "$tmp" && mv "$tmp" version.json
```
pleeko, line 1189: you might not want to bump the proto. I think this field should be left alone and only updated by developers when we need new fields or functions from core.

pleeko: you are also bumping without building the protos in the script.

molinamelendezj: good catch!

.circleci/main.yml
```diff
@@ -1163,7 +1163,60 @@ jobs:
...
+  console-release-draft:
+    docker:
+      - image: cimg/base:stable
+    steps:
+      - gh/setup:
+          version: 2.13.0
+      - run:
+          name: create github console release draft
+          command: |
+            if [[ $CIRCLE_TAG == *"-"* ]];
```
djanicekpach, line 1175: Is there any reason you put this command inline instead of in an sh script? I don't have a preference, just asking for my own knowledge.

.circleci/main.yml
```diff
@@ -1163,7 +1163,60 @@ jobs:
...
+                  echo "regular release."
+                  parentBranch=${CIRCLE_TAG:1}
+              fi
+            git clone https://github.com/pachyderm/haberdashery.git haberdashery
+            git checkout -b "release-${CIRCLE_TAG:1}"
+            tmp=$(mktemp)
+            jq --arg val "${CIRCLE_TAG:1}-${CIRCLE_SHA1}" '.pachyderm = $val' version.json > "$tmp" && mv "$tmp" version.json
+            jq --arg val "${CIRCLE_SHA1}" '.pachReleaseCommit = $val' version.json > "$tmp" && mv "$tmp" version.json
+            jq --arg val "${CIRCLE_TAG:1}" '.pachctl = $val' version.json > "$tmp" && mv "$tmp" version.json
+            git commit -m "bump pachyderm version"
```
djanicekpach, line 1189: I think there needs to be a `git add` somewhere between the `clone` and `commit`? Are we in the haberdashery folder/repo as well, or are we supposed to be in the `pachyderm` repo still?

.circleci/main.yml
```diff
@@ -1163,7 +1163,60 @@ jobs:
...
+            git push
+            PR=$(gh pr create --base $parentBranch --title "bump pach version for release ${CIRCLE_TAG:1}" --repo pachyderm/haberdashery)
+            gh pr merge ${PR} --auto --repo pachyderm/haberdashery
+            gh pr checks ${PR} --10 --watch --repo pachyderm/haberdashery
```
djanicekpach, line 1193: this is pretty slick! What is the `--10`? I was hoping to find a timeout setting in the docs, but I couldn't find one.

.circleci/main.yml
```diff
@@ -1163,7 +1163,60 @@ jobs:
...
+                  echo "git tag has - assuming prerelease."
+                  gh release create --draft ${CIRCLE_TAG:1}-1 --tittle ${CIRCLE_TAG:1}-1 --target $parentBranch --generate-notes --prerelease --repo pachyderm/haberdashery
+              else
+                  echo "regular release."
+                  gh release create --draft ${CIRCLE_TAG:1}-1 --tittle ${CIRCLE_TAG:1}-1 --target $parentBranch --generate-notes --repo pachyderm/haberdashery
+              fi
+            SECONDS=0
+            while :; do
+                docker pull pachyderm/haberdashery:${CIRCLE_TAG:1}-1 2>/dev/null && break
+                if (( $SECONDS > 1 )); then
```
djanicekpach, line 1206: how does `SECONDS` get incremented here?

djanicekpach (changes requested):

.circleci/main.yml
molinamelendezj: good question. I really don't know what the preference has been, endless shell scripts or endless lines of yaml 😆 ideally would love to break this config out and leverage shell scripts to further break it down.

.circleci/main.yml
molinamelendezj: in the haberdashery repo, good catch. I was able to test most of this. unfortunately not really all testable until a release is done. but made it to where we can still intervene manually since no other job depends on it.

.circleci/main.yml
molinamelendezj: I think I ate the `-i` flag. yea its the only glue I could find https://cli.github.com/manual/gh_pr_checks

.circleci/main.yml
molinamelendezj: `SECONDS` is a special variable in bash whose value is incremented once per second.

djanicekpach: I had no idea! I wonder if this has broken any scripts I've ever written thinking SECONDS was just a normal variable.... that's horrifying

.circleci/main.yml
djanicekpach: Yeah, I figured this is one of those things that's hard to test until you use it. I think it still needs a `cd haberdashery` somewhere though

molinamelendezj: @djanicekpach I think you have a good point. I think, maybe I should wait to merge this, run these steps on the next release as a test and then merge it once I confirm they all work. 

I believe the PR will never close as their might not be an approver. So might need to add that step too.

pleeko: @molinamelendezj did the steps work for the last release?

pleeko (approved):
It looks good to me, but I am not that confident in my circle ci scripting knowledge. Giving this a 👍 with the condition we discussed of manually testing the docker image until we have that step automated.

djanicekpach (approved):

PR #8443, opened 2022-12-14
[PFS-32] Datum batching

brycemcanally:
This PR contains the implementation of datum batching.


src/pps/pps.proto
```diff
@@ -708,7 +709,9 @@ message CreatePipelineRequest {
   Metadata metadata = 28;
   string reprocess_spec = 29;
   bool autoscaling = 30;
+  // TODO: Why is there a field number gap here?
```
brendoncarroll, line 712: We may never know, but good catch, we should add `reserved` for the field.

src/pps/pps.proto
```diff
@@ -708,7 +709,9 @@ message CreatePipelineRequest {
...
   repeated Toleration tolerations = 34;
+  bool batching = 35;
```
brendoncarroll, line 714: So batching is a property of the Pipeline, not the Transform?

src/server/pps/cmds/cmds.go
```diff
@@ -1337,6 +1338,25 @@ All jobs created by a pipeline will create commits in the pipeline's output repo
 	runLoadTest.Flags().StringVar(&stateID, "state-id", "", "The ID of the base state to use for the load.")
 	commands = append(commands, cmdutil.CreateAlias(runLoadTest, "run pps-load-test"))
 
+	var errStr string
+	nextDatum := &cobra.Command{
```
brendoncarroll, line 1342: can we mark this as Hidden?

src/server/worker/server/api_server.go
```diff
@@ -48,3 +49,15 @@ func (a *APIServer) Cancel(ctx context.Context, request *CancelRequest) (*Cancel
 	success := a.workerInterface.Cancel(request.JobID, request.DataFilters)
 	return &CancelResponse{Success: success}, nil
 }
+
+func (a *APIServer) NextDatum(ctx context.Context, request *NextDatumRequest) (*NextDatumResponse, error) {
+	var err error
+	if request.Error != "" {
+		err = errors.New(request.Error)
```
brendoncarroll, line 56: This value of error is never used.

src/server/worker/server/api_server.go
```diff
@@ -48,3 +49,15 @@ func (a *APIServer) Cancel(ctx context.Context, request *CancelRequest) (*Cancel
...
+	}
+	env, err := a.workerInterface.NextDatum(ctx, err)
```
brendoncarroll, line 58: It's shadowed here.

src/server/worker/server/api_server.go
```diff
@@ -15,6 +15,7 @@ import (
 type WorkerInterface interface {
```
brendoncarroll, line 15: Where is this used?  I think we should be moving away from defining veneer interfaces to avoid using the gRPC interfaces.  I would rather see us write utility functions (maybe generic in `grpcutil`), to make dealing with the gRPC interfaces less annoying.

src/pps/pps.proto
brendoncarroll: I expanded on this in the design doc.
> I think datum batching needs to be coupled to the Pipeline Transform as a concept.  We are going to have a situation where some docker images are for batching, and some for non-batching.  This needs to make it into our model.  I believe we call the docker image, and the other parameters that go into running it the Transform, so that seems like the best place.

src/server/worker/pipeline/transform/worker.go
```diff
@@ -268,63 +269,19 @@ func handleDatumSet(driver driver.Driver, logger logs.TaggedLogger, task *DatumS
...
 					datum.WithStats(stats),
 				}
-				// TODO: Can this just be refactored into the datum package such that we don't need to specify a storage root for the sets?
-				// The sets would just create a temporary directory under /tmp.
-				storageRoot := filepath.Join(driver.InputDir(), client.PPSScratchSpace, uuid.NewWithoutDashes())
-				userImageID, err := driver.GetContainerImageID(pachClient.Ctx(), "user")
-				if err != nil {
-					return errors.Wrap(err, "could not get user image ID")
+				if driver.PipelineInfo().Details.Batching {
+					return handleDatumSetBatching(ctx, driver, logger, task, status, cacheClient, di, opts)
```
brendoncarroll, line 280: 🔥

src/server/worker/pipeline/transform/status.go
```diff
@@ -98,3 +101,20 @@ func (s *Status) Cancel(jobID string, datumFilter []string) bool {
 	}
 	return false
 }
+
+func (s *Status) NextDatum(ctx context.Context, err error) ([]string, error) {
+	if s.nextChan == nil {
```
brendoncarroll, line 106: It seems like this could race with the assignment in the other file.

src/server/worker/pipeline/transform/worker.go
```diff
@@ -347,6 +304,125 @@ func handleDatumSet(driver driver.Driver, logger logs.TaggedLogger, task *DatumS
...
+		}
+	})
+}
+
+type datumCallback = func(ctx context.Context, logger logs.TaggedLogger, env []string) error
+
+// TODO: There are way too many parameters here. Consider grouping them into a reasonable struct or storing some of these in the driver.
+func forEachDatum(ctx context.Context, driver driver.Driver, logger logs.TaggedLogger, task *DatumSetTask, status *Status, cacheClient *pfssync.CacheClient, di datum.Iterator, setOpts []datum.SetOption, cb datumCallback) error {
+	// TODO: Can this just be refactored into the datum package such that we don't need to specify a storage root for the sets?
+	// The sets would just create a temporary directory under /tmp.
```
brendoncarroll, line 376: I don't think we should be building things that assume they can write to a path without being passed that path as a parameter.  It makes it difficult to test multiple copies of the thing concurrently, and it makes the component more difficult to reason about because it has secret global state from the filesystem.

src/server/worker/server/api_server.go
brycemcanally: Refer to the comment above.

src/server/worker/server/api_server.go
brycemcanally: This is more like a driver interface for the implementation of the worker gRPC server, although the current implementation of it is not called a driver. This extra level of indirection may not be worth it.

src/server/worker/pipeline/transform/status.go
```diff
@@ -98,3 +101,20 @@ func (s *Status) Cancel(jobID string, datumFilter []string) bool {
...
+		return nil, errors.New("datum batching not enabled")
+	}
+	select {
+	case s.nextChan <- err:
+	case <-ctx.Done():
+		return nil, errors.EnsureStack(ctx.Err())
+	}
+	select {
+	case env := <-s.setupChan:
+		return env, nil
```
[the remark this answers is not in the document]
brycemcanally: yeah

src/server/worker/pipeline/transform/worker.go
brycemcanally: This was just a copied over TODO. Generally making the root directory / prefix configurable makes sense.

src/server/worker/pipeline/transform/worker.go
```diff
@@ -347,6 +304,125 @@ func handleDatumSet(driver driver.Driver, logger logs.TaggedLogger, task *DatumS
 	})
 }
 
+func handleDatumSetBatching(ctx context.Context, driver driver.Driver, logger logs.TaggedLogger, task *DatumSetTask, status *Status, cacheClient *pfssync.CacheClient, di datum.Iterator, setOpts []datum.SetOption) error {
+	nextChan, setupChan := make(chan error), make(chan []string)
+	status.nextChan, status.setupChan = nextChan, setupChan
```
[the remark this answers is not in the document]
brycemcanally: I refactored this a bit. Status now has a method that will provide the channels in a callback. The channels are created and niled within a lock. The problematic scenarios that should now be handled are:
- Multiple goroutines attempting to set up a batch will be detected and produce a panic.
- User code code running / calling next datum before the batch is set up will return an error.
- Ensuring that the next datum call is terminated before niling the channels. Prior to this, we just ensured that the user code has terminated, but that doesn't mean that the next datum call on the worker server side has terminated. The locking on cleanup now ensures this.

I don't think the first two could happen with the current code, but we are safe against them in the future. The third could happen, but is handled now.

src/server/worker/pipeline/transform/status.go
brycemcanally: Refer to the comment above.

src/server/worker/pipeline/transform/status.go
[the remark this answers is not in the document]
brycemcanally: It could be, but it doesn't need to be now. This is an internal detail that we can easily change later.

src/server/worker/pipeline/transform/worker.go
```diff
@@ -347,6 +304,125 @@ func handleDatumSet(driver driver.Driver, logger logs.TaggedLogger, task *DatumS
 	})
 }
 
+func handleDatumSetBatching(ctx context.Context, driver driver.Driver, logger logs.TaggedLogger, task *DatumSetTask, status *Status, cacheClient *pfssync.CacheClient, di datum.Iterator, setOpts []datum.SetOption) error {
```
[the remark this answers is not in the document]
brycemcanally: There is probably a better way to organize the parameters than just grouping them into a struct for this specific function.

src/pps/pps.proto
brycemcanally: I moved it to the transform.

src/server/worker/pipeline/transform/status.go
```diff
@@ -98,3 +102,41 @@ func (s *Status) Cancel(jobID string, datumFilter []string) bool {
...
+	defer s.withBatchLock(func() {
+		s.nextChan, s.setupChan = nil, nil
+	})
+	return cb(s.nextChan, s.setupChan)
+}
+
+func (s *Status) withBatchLock(cb func()) {
+	s.batchMutex.Lock()
+	cb()
+	s.batchMutex.Unlock()
```
brendoncarroll, line 122: I think it's better practice to defer here, because that unlocks during a panic.  Not that it really matters for us.

brendoncarroll:
This looks good to me, just nits leftover.

brendoncarroll (approved):

src/server/worker/pipeline/transform/status.go
brycemcanally: yeah, I chose this for consistency with the other with lock method. I went ahead and changed both to the defer unlock pattern.

PR #8445, opened 2022-12-14
Djanicek/validation test conversion

djanicekpach:
This converts validation tests to use RealEnv instead of minikube to test. Additionally, the tests also use a non-default project to get some extra coverage, and I added a validation test to check project names.

Side fix - while debugging some test flake I noticed the `kubectl describe pod` was not outputting info since it didn't have namespaces, so I added the 6 namespaces that we also output logs for.


src/server/pachyderm_validation_test.go
```diff
@@ -88,38 +88,80 @@ func TestPipelineThatUseNonexistentInputs(t *testing.T) {
...
 			Constant: 1,
 		},
-		client.NewProjectPFSInput(pfs.DefaultProjectName, dataRepo, "/*"),
+		client.NewProjectPFSInput(projectName, dataRepo, "/*"),
 		"",
 		false,
 	))
 }
+
+func TestProjectNameValidation(t *testing.T) {
```
jrockway, line 129: I think a RealEnv is too heavy for testing a validation routine.  pfs.CreateProject calls driver.CreateProject; that's where we should add the tests.

It looks like you can make a mock service env and use that to create a pfs/server.driver, then createProject on the driver object. 

Now as it turns out as I look into this, we already have this exact test in src/server/pfs/server/testing/server_test; called "InvalidProject".  It uses a realenv 😂 

But please go add these extra test cases there, I think that makes the most sense.  (They very happy-pathed those tests so your additional cases will prove helpful!)

jrockway: Last tangent, the validation logic really lives in src/pfs/pfs.go#ValidateName.

I am not sure why nobody writes tests for our proto accoutrements (which is what src/X usually is).  That's where I'd put this test though.  

Finally, that method delegats to src/internal/ancestry which does have tests.  ">.<" is the test data I most enjoyed in there.

jrockway (approved):
Good cleanup.  Move those validation tests over to src/server/pfs/server/testing tho.

src/server/pachyderm_validation_test.go
djanicekpach: Ah cool, I'll move it. 
I actually like that NewRealEnv goes all the way to the DB. I wasn't aware of pfs/server/testing, but I did know about ValidateName. In theory I agree that tests at the lower level are better, but most of the issues I'm seeing with names are things that work as expected with `ValidateName` but the DB chokes, or there's an error with pipelines or something else downstream, so I was trying to exercise that.

Since we're already using realenv in /pfs/server/testing I'd like to defer coming up with a stripped down pfs mock environment. I see something similar for pipelines in pipeline_controller_test.go. I'll add a ticket to CORE-1188 to loop back and convert these to a PFS specific mock instead.

PR #8446, opened 2022-12-15
Make `pachctl list pipelines` filter on projects

albscui:
Also updated the pretty printer to be aware of projects.
```
pachctl list pipelines --project myProject
PROJECT   NAME                   VERSION INPUT                             CREATED        STATE   / LAST_JOB DESCRIPTION 
myProject pipeline2-e6e72c989a56 1       default/pipeline1-0df9cddae8e9:/* 14 minutes ago running / success

pachctl list pipelines --all-projects
PROJECT   NAME                   VERSION INPUT                             CREATED        STATE   / LAST_JOB DESCRIPTION 
myProject pipeline2-e6e72c989a56 1       default/pipeline1-0df9cddae8e9:/* 14 minutes ago running / success              
default   pipeline1-0df9cddae8e9 1       default/input:/*                  14 minutes ago running / success             
```


src/pps/pps.proto
```diff
@@ -690,6 +690,9 @@ message ListPipelineRequest {
 
   // If non-nil, will return all the pipeline infos at this commit set
   pfs_v2.CommitSet commit_set = 5;
+
+  // Projects to filter on. Empty list means no filter, so return all pipelines.
+  repeated string projects = 6;
```
acohen4, line 695: It's probably a better practice to use the Project proto as the type here

src/server/pps/pretty/pretty.go
```diff
@@ -24,7 +24,7 @@ import (
 
 const (
 	// PipelineHeader is the header for pipelines.
-	PipelineHeader = "NAME\tVERSION\tINPUT\tCREATED\tSTATE / LAST JOB\tDESCRIPTION\t\n"
+	PipelineHeader = "PROJECT\tNAME\tVERSION\tINPUT\tCREATED\tSTATE / LAST JOB\tDESCRIPTION\t\n"
```
acohen4, line 27: idk how we've done it elsewhere, but should PROJECT be the first column? My naive UX opinion is that the NAME is the most significant piece of information and should therefore come first, even though the PIPELINE fits into a PROJECT

albscui: Seems like this still needs some design work. Should I keep my current changes in, since it's consistent with listing repos? We can change this once we decide on a proper design.

robert-uhl: On the one hand, projects contain repos & pipelines, so it makes sense to have them leftmost. OTOH, Alon is right that the individual names are more significant in a list — especially if they are all from the same project. Maybe it would make sense to only display the project column when items from multiple projects are listed?

This probably needs more design work and input from non-engineers. For the time being, I think that it’s probably okay to move forward with this pattern. @maguenza, any objection?

robert-uhl (approved):

PR #8449, opened 2022-12-15
Add timeout to queryLoki

albscui:
A lot of the recent debug dump hanging issues is due to Loki queries hanging, which is a bug in Loki itself, but for now, we want to handle this by gracefully timing out so that the user can get back some results.

### DIFF
--- a/src/server/debug/server/server_test.go
+++ b/src/server/debug/server/server_test.go
@@ -147,7 +147,7 @@ func TestQueryLoki(t *testing.T) {
 			name: "all logs",
 			buildEntries: func() []loki.Entry {
 				var entries []loki.Entry
-				for i := -99; i >= 0; i++ {
+				for i := -99; i <= 0; i++ {
 					entries = append(entries, loki.Entry{
 						Timestamp: time.Now().Add(time.Duration(-1) * time.Second),
 						Line:      fmt.Sprintf("%v", i),
@@ -157,7 +157,7 @@ func TestQueryLoki(t *testing.T) {
 			},
 			buildWant: func() []int {
 				var want []int
-				for i := -99; i >= 0; i++ {
+				for i := -99; i <= 0; i++ {
 					want = append(want, i)
 				}
 				return want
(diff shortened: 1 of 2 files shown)


src/server/debug/server/server_test.go
```diff
@@ -241,34 +257,81 @@ func TestQueryLoki(t *testing.T) {
...
 				env: &serviceenv.TestServiceEnv{
 					LokiClient: &loki.Client{Address: s.URL},
 				},
 			}
 
 			var got []int
-			out, err := d.queryLoki(context.Background(), `{foo="bar"}`)
+			ctx, cancel := context.WithTimeout(context.Background(), queryTimeout)
+			defer cancel()
+			log.Println("querying loki")
```
jrockway, line 329: t.Log

jrockway (approved):
Excellent tests.  Did the output end up making sense when it failed?  I spent a long time tweaking that table and it might have worked at some point :D

src/server/debug/server/server_test.go
```diff
@@ -208,7 +220,7 @@ func TestQueryLoki(t *testing.T) {
 			buildWant: func() []int {
 				var want []int
 				want = append(want, -2)
-				for i := 0; i < 4999; i++ {
+				for i := 0; i < serverMaxLogs-1; i++ {
```
albscui, line 223: I actually don't understand why this has to be 1 less than the `serverMaxLogs`

jrockway: Log 1 is "-2", logs 2-serverMax are "-1", and then we seek over 40,000-ish lines and just return "0".  We seek over them because there is no way to address them with the Loki API, so we expect serverMaxLogs + 1 log.  It's the best we can do with their API :(

albscui:
> Excellent tests. Did the output end up making sense when it failed? I spent a long time tweaking that table and it might have worked at some point :D

The table is great! Just had to change some tabs to spaces for the header to align.

PR #8450, opened 2022-12-16
project names must start with an alphanumeric character [CORE-1343]

acohen4:


robert-uhl (approved):
Future-proofing against the day we allow more characters in names … I like it☺

robert-uhl (changes requested):
Need to fix the test errors though …

robert-uhl (approved):

PR #8452, opened 2022-12-16
Djanicek/api perf data export

djanicekpach:
Creates a pipeline script to export performance test data to big query. I ran in the pipeline and was able to successfully export data.


etc/testing/circle/run_api_performance_tests.sh
```diff
@@ -38,3 +38,12 @@ locust -f locustfile.py --headless --users 50 --spawn-rate 1 --run-time 3m \
...
     --exit-code-on-error 0 # errors are reported in the artifacts, if the test finishes the pipeline ran successfully
+cd ..
+
+# Collect and export stats to bigquery
+export PACHD_PERF_VERSION="$image_tag"
+export API_PERF_RESULTS_FOLDER="/tmp/test-results"
+cd etc/testing/circle/workloads/api-perf-collector
+pip3 install -r requirements.txt
+python3 app.py
+cd ..
```
molinamelendezj, line 49: eof

molinamelendezj (approved):
neat! I like it. we can start building some visualizations soon.  at some point, we will have to think about how much data we keep for how long, so big query doesn't cost us to much.

PR #8453, opened 2022-12-16
Add number of tokens revoked to RevokeToken response

albscui:
This PR adds the feature where `pachctl revoke token` prints out how many tokens were revoked. This gives the user a better UX.

```
pachctl auth revoke --token <token>
1 auth token revoked

pachctl auth revoke --token <token>
0 auth tokens revoked


pachctl auth revoke --user <user>
3 auth tokens revoked

pachctl auth revoke --user <user>
0 auth tokens revoked
```


robert-uhl:
It would be a bit nicer for humans to print ‘N auth tokens revoked’; [plural](https://github.com/pachyderm/pachyderm/pull/8448) could handle changing between ‘token’ & ‘tokens’ given the number.

Not a huge deal.

robert-uhl (approved):

src/auth/auth.proto
```diff
@@ -392,7 +392,9 @@ message RevokeAuthTokenRequest {
   string token = 1;
 }
 
-message RevokeAuthTokenResponse {}
+message RevokeAuthTokenResponse {
+  int64 num = 1;
```
acohen4, line 396: is num a good name here? I know we use Number in other places. Not sure what the right answer is

acohen4 (approved):
one comment on the proto field name we should align on. LGTM

src/auth/auth.proto
albscui: `Number` sounds good!

PR #8454, opened 2022-12-16
minikubetestenv: add a flag to use whatever you have currently setup

jrockway:
I'm not sure if this duplicates existing functionality.  I hacked it in one evening and it works great for me (yeah, some tests that use NodePorts or that inspect k8s fail; TestAutoscaling, TestService, etc. but it wasn't too burdensome to temporarily modify those as I was working on them).


acohen4 (approved):
LGTM

PR #8456, opened 2022-12-16
s3: log net/http errors at Debug level

jrockway:


seslattery (approved):

jrockway:
I am not sure what's going on with govulncheck in this branch.  I have the same version of Go and govulncheck as the CI machine, and don't get any warnings.

PR #8457, opened 2022-12-16
[Backport 2.4.x] Add number of tokens revoked to RevokeToken response (#8453)

albscui:
* Add number of tokens revoked to RevokeToken response

* handle plurals properly

* rename proto field from num to number

* fix lint


jrockway (approved):
I like the pluralization!

PR #8460, opened 2022-12-20
GetPermissionsResponse.Permissions should respect the requested resource type [CORE-1404]

acohen4:


albscui (approved):
Awesome that we finally caught this bug!

PR #8463, opened 2022-12-20
Djanicek/projects pachctl testing

djanicekpach:
This includes test changes to help reduce flake and test Projects. Related tickets:
CORE-1409 - We didn't have a test for job or inspect job, so we had a small bug. This fixes the bug and adds a cmd test.
CORE-1015 - The only test I could find for datum restart was flaky. I added projects and de-flaked it. I beleieve the cap for the exponential backoff was too high and was causing test failures.

Additionally, for increased test coverage I updated TestPFS/SubscribeCommit to include projects.

### DIFF
--- a/src/internal/require/require.go
+++ b/src/internal/require/require.go
@@ -446,8 +446,21 @@ func NoErrorWithinT(tb testing.TB, t time.Duration, f func() error, msgAndArgs .
 }
 
 // NoErrorWithinTRetry checks that 'f' finishes within time 't' and does not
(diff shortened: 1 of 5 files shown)


albscui (approved):
LGTM, finding and fixing flakes is not trivial, nice work! Just some renaming issues.
On a relatled noted, perhaps its worth looking at https://github.com/pachyderm/pachyderm/blob/2cdb2cf2b8e68a6290e6fef6a0d304537f534987/src/internal/backoff/exponential.go#L130 and see if that's causing similar flakes.

PR #8465, opened 2022-12-20
[CORE-1371] Disallow changing project or pipeline name in pachctl edit pipeline

robert-uhl:


src/server/pps/cmds/cmds.go
```diff
@@ -881,6 +881,27 @@ All jobs created by a pipeline will create commits in the pipeline's output repo
...
+			// then it is considered unchanged.
+			project := request.Pipeline.GetProject()
+			projectName := project.GetName()
+			if projectName == "" {
+				if project == nil {
+					project = new(pfs.Project)
+				}
+				project.Name = pipelineInfo.Pipeline.GetProject().GetName()
+				request.Pipeline.Project = project
+			} else if projectName != pipelineInfo.Pipeline.GetProject().GetName() {
```
albscui, line 894: Can we move the exceptional cases up?
```Go
if projectName != pipelineInfo.Pipeline.GetProject().GetName() {
    return errors.New("may not change project name")
}
if pipelineName != pipelineInfo.Pipeline.GetName() {
     return errors.New("may not change pipeline name")
}

// business logic ...
```

albscui: Also given that we don't allow the pipeline/project name to change, I think we can safely assign it to the request without conditionals.

acohen4 (approved):
LGTM

src/server/pps/cmds/cmds.go
acohen4: +1

PR #8468, opened 2022-12-20
Lock the pachw-controller-lock 🔒

FahadBSyed:


jrockway:
I took care of this in #8411

PR #8472, opened 2022-12-21
Increase reliability of debug dumps

jrockway:
Fixes CORE-1193 and CORE-1294.

This PR does a bunch of stuff to make debug dumps more reliable, at least without burning the whole thing down and starting over.

`pachctl debug dump` can now specify a timeout; it defaults to 30m.

The timeout is adjusted down on the server side to about 90% of the client timeout.  That means the debug dumper has some time to handle `context deadline exceeded` and start producing output before the RPC is totally aborted.  I've had good results with timeouts as low as 100ms; you don't get everything, but you get some files.  At 30m it should be Really Good (tm).

Every multi-step operation that the dumper does now continues in the face of errors, if the error doesn't affect the next thing.  Every for loop or function that does two+ things now uses `multierr.Append` to collect all the errors.  That means if we hit an issue where we try to do something silly like InspectPipeline an input repo, we just continue doing the whole debug dump anyway.  At the very end, an error will be returned, but we can still write all the other files.

I fixed the thing where we did InspectPipeline on an input repo; there was a missing `continue` statement.  I also tried to fix PPS's error message for a pipeline not being found, but it's actually not relevant to this PR.  (I don't think the code can ever hit the case I "fixed", but in case it does, hey now the error type is correct.  We still don't return grpc.status = NotFound from PPS under any of these circumstances though.)

I added some arbitrary timeouts around things I don't think will be too slow, like we did for Loki.

I noticed that the Pod Describer from the k8s library can't take a context.  That means it could run forever, so I put it in a background goroutine; the foreground goroutine tries to get its output until the context expires, and then it just abandons it and moves on.  This will leak memory if it runs forever, but hey, after we review the debug dump we'll probably tell you to restart pachd anyway.  In the future we'll have to just collect pod YAMLs instead of "describe" output.  Or fork k8s.io/client-go to make the silly thing take a context.

As an example, here's what a run with an aggressive timeout looks like now:
```
$ rm dump.tgz; /usr/bin/time pachctl debug dump dump.tgz  --timeout=1s ; tar tzvf dump.tgz; du -h dump.tgz
rpc error: code = Unknown desc = listPipelines: context deadline exceeded; appLogs: context deadline exceeded; collectDatabaseDump: collectDatabaseTables: list tables: context deadline exceeded
Command exited with non-zero status 1
0.09user 0.02system 0:01.04elapsed 11%CPU (0avgtext+0avgdata 66060maxresident)k
0inputs+2176outputs (0major+2005minor)pagefaults 0swaps
-rwxrwxrwx 0/0            6214 1969-12-31 19:00 source-repos/default/benchmark-upload/commits.json
-rwxrwxrwx 0/0           52020 1969-12-31 19:00 source-repos/default/benchmark-upload/commits-chart.png
-rwxrwxrwx 0/0            8083 1969-12-31 19:00 source-repos/default/images/commits.json
-rwxrwxrwx 0/0           45961 1969-12-31 19:00 source-repos/default/images/commits-chart.png
-rwxrwxrwx 0/0              17 1969-12-31 19:00 pachd/pachd-84f6794987-74hf2/pachd/version.txt
-rwxrwxrwx 0/0            7612 1969-12-31 19:00 pachd/pachd-84f6794987-74hf2/pachd/describe.txt
-rwxrwxrwx 0/0         8690350 1969-12-31 19:00 pachd/pachd-84f6794987-74hf2/pachd/logs.txt
-rwxrwxrwx 0/0              80 1969-12-31 19:00 pachd/pachd-84f6794987-74hf2/pachd/logs-previous/error.txt
-rwxrwxrwx 0/0         8640042 1969-12-31 19:00 pachd/pachd-84f6794987-74hf2/pachd/logs-loki.txt
-rwxrwxrwx 0/0           22422 1969-12-31 19:00 pachd/pachd-84f6794987-74hf2/pachd/go_info.txt
-rwxrwxrwx 0/0           11559 1969-12-31 19:00 pachd/pachd-84f6794987-74hf2/pachd/goroutine
-rwxrwxrwx 0/0           84444 1969-12-31 19:00 pachd/pachd-84f6794987-74hf2/pachd/heap
-rwxrwxrwx 0/0              26 1969-12-31 19:00 database/activities/error.txt
-rwxrwxrwx 0/0              26 1969-12-31 19:00 database/row-counts/error.txt
-rwxrwxrwx 0/0              26 1969-12-31 19:00 database/table-sizes/error.txt
1.1M    dump.tgz
```
We end up with data (and a long chain of error messages) even if we hit timeouts.

[CORE-1193]: https://pachyderm.atlassian.net/browse/CORE-1193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CORE-1294]: https://pachyderm.atlassian.net/browse/CORE-1294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/debug/server/server.go
+++ b/src/server/debug/server/server.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/wcharczuk/go-chart"
+	"go.uber.org/multierr"
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


src/server/debug/server/server.go
```diff
@@ -901,7 +990,12 @@ type lokiLog struct {
 	Entry  *loki.Entry
 }
 
-func (s *debugServer) queryLoki(ctx context.Context, queryStr string) ([]lokiLog, error) {
+func (s *debugServer) queryLoki(ctx context.Context, queryStr string) (retResult []lokiLog, retErr error) {
+	ctx, finishSpan := log.SpanContext(ctx, "queryLoki", zap.String("queryStr", queryStr))
+	defer func() {
```
albscui, line 995: I noticed that in other places where we end the span, we do something like:
```go
ctx, end := log.SpanContext(ctx, "queryLoki", zap.String("queryStr", queryStr))
defer end(log.Errorp(&retErr), zap.Int("logs", len(retResult)))
```
Just curious why we do it differently here.

src/server/debug/server/server.go
```diff
@@ -361,11 +386,11 @@ func (s *debugServer) Profile(request *debug.ProfileRequest, server debug.Debug_
 		pachClient,
 		server,
 		request.Filter,
-		collectProfileFunc(request.Profile),
-		nil,
-		nil,
-		redirectProfileFunc(request.Profile),
-		collectProfileFunc(request.Profile),
+		collectProfileFunc(request.Profile),  // collectPachd
+		nil,                                  // collectPipeline
```
albscui, line 390: nit: for annotating naked parameters, I kinda prefer C-style comments.

```suggestion
		nil /* collectPipeline */,
```

But I can see an argument in favor of aligning the comments like this since every parameter is on a new line.

src/server/debug/server/server.go
```diff
@@ -530,19 +565,24 @@ func (s *debugServer) collectInputRepos(ctx context.Context, tw *tar.Writer, pac
...
 			if errutil.IsNotFoundError(err) {
 				repoPrefix := join("source-repos", repoInfo.Repo.Project.Name, repoInfo.Repo.Name)
-				return s.collectCommits(ctx, tw, pachClient, repoInfo.Repo, limit, repoPrefix)
+				if err := s.collectCommits(ctx, tw, pachClient, repoInfo.Repo, limit, repoPrefix); err != nil {
+					multierr.AppendInto(&errs, errors.Wrapf(err, "collectCommits(%s:%s)", repoInfo.GetRepo(), repoPrefix))
+				}
+				continue
 			}
-			return err
+			multierr.AppendInto(&errs, errors.Wrapf(err, "inspectPipeline(%s)", repoInfo.GetRepo()))
```
albscui, line 582: Nice! Now we capture all the input repos!

src/server/debug/server/util.go
```diff
@@ -95,9 +92,7 @@ func collectDebugStream(tw *tar.Writer, r io.Reader, prefix ...string) (retErr e
 		return errors.EnsureStack(err)
 	}
 	defer func() {
-		if err := gr.Close(); retErr == nil {
-			retErr = err
-		}
+		multierr.AppendInto(&retErr, gr.Close())
```
albscui, line 95: Curious why in most places we first initialize an error with `var errs error` before calling  `multierr.AppendInto(&errs, ...)`, but here we reuse the `retErrr`?

albscui (approved):
LGTM! Most of my comments are just nits and curiosities, so not blockers for merging.

src/server/debug/server/util.go
jrockway: There are two cases. 1 is where we are creating the raw return error from the sub-task, that's the `var errs error` case.  We just need to allocate a pointer to an interface to allocate the multi-error into.  2 is where we do some cleanup while returning from a function, in which case `func X() (retErr error)` allocates `retErr` and it's either the main error or nil; we append into that object.  (It may seem weird to write into a nil pointer, but the interface is not nil; it's an error that points to a nil object, as opposed to `var error *error` which is a nil pointer to an interface, not an interface pointing to a nil object.)

src/server/debug/server/server.go
jrockway: Yeah they wanted to write `break` there initially, not `return` basically.

src/server/debug/server/server.go
jrockway: I don't really have a preference.

src/server/debug/server/server.go
```diff
@@ -94,76 +95,98 @@ func (s *debugServer) handleRedirect(
...
 						// Collect the user container.
 						if err := collect(ctx, tw, pachClient, client.PPSWorkerUserContainerName); err != nil {
-							return err
+							multierr.AppendInto(&errs, errors.Wrap(err, "collect user container"))
 						}
 						// Redirect to the storage container.
 						r, err := redirect(ctx, s.sidecarClient.DebugClient, filter)
 						if err != nil {
-							return err
+							multierr.AppendInto(&errs, errors.Wrap(err, "collect user container"))
```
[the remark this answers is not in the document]
jrockway: Done.

src/server/debug/server/server.go
jrockway: finishSpan instead of end?  Because we use `end` for the end time of the query :/

Oh, and error instead of errorp?  That's because we also want to record len(retResult) on the way out, and we don't have a utility function that takes a pointer to a slice and prints the length, but we do have one for errors (since that's the more common case).

PR #8473, opened 2022-12-21
Make a 'pachyderm-tap/pachctl' formula represent the latest stable release

acohen4:


jrockway (approved):
This looks like it will work!

etc/build/update_homebrew.sh
```diff
@@ -20,6 +21,11 @@ pushd homebrew-tap
     git checkout -b "$BRANCH" || git checkout "$BRANCH"
     VERSION=$version ./update-formula.sh
     git add "pachctl@$MAJOR_MINOR.rb"
+    if [[ $stable ]]; then
+        cp "pachctl@$MAJOR_MINOR.rb" pachctl.rb
```
molinamelendezj, line 25: Can't remember, but cp will over write file with out error if file exist right.

molinamelendezj (approved):
Looks good one small comment. Please cherry pick this too.

etc/build/update_homebrew.sh
acohen4: nah it will create <dest> if it doesn't exist

PR #8475, opened 2022-12-21
[CORE-1246] WalkFile pagination

armaanv:


src/server/pfs/server/driver_file.go
```diff
@@ -356,17 +362,69 @@ func (d *driver) walkFile(ctx context.Context, file *pfs.File, cb func(*pfs.File
...
 		return cb(fi)
 	})
-	if p == "" && pacherr.IsNotExist(err) {
+	if (p == "" && pacherr.IsNotExist(err)) || err == errutil.ErrBreak {
 		err = nil
 	}
 	return err
 }
 
+func (d *driver) walkFileReverse(ctx context.Context, s Source, number int64, cb func(*pfs.FileInfo) error) error {
```
brycemcanally, line 392: We should reuse `listFileReverse` since it is the same code. Maybe rename it to `reverseFiles`. It would just be a utility function for reversing a certain number of files from a source.

src/server/pfs/server/driver_file.go
```diff
@@ -356,17 +362,69 @@ func (d *driver) walkFile(ctx context.Context, file *pfs.File, cb func(*pfs.File
...
+	}
 	err = s.Iterate(ctx, func(fi *pfs.FileInfo, f fileset.File) error {
+		if number == 0 {
+			return errutil.ErrBreak
+		}
+		number--
 		return cb(fi)
 	})
-	if p == "" && pacherr.IsNotExist(err) {
+	if (p == "" && pacherr.IsNotExist(err)) || err == errutil.ErrBreak {
```
brycemcanally, line 386: This should probably be an `errors.Is` check.

src/server/pfs/server/driver_file.go
```diff
@@ -347,6 +347,12 @@ func (d *driver) walkFile(ctx context.Context, file *pfs.File, cb func(*pfs.File
 	if p == "/" {
 		p = ""
 	}
+	if number == 0 {
```
brycemcanally, line 350: This should have the same requirements as a list file pagination, so we should probably just refactor the check into a utility function (i.e. number == 0 && reverse is an error and number > 100000 is an error).

src/server/pfs/server/driver_file.go
armaanv: I've moved the circular buffer into a struct with helper functions which removes a lot of the re-used code.

src/server/pfs/server/driver_file.go
```diff
@@ -283,7 +290,19 @@ func (d *driver) listFile(ctx context.Context, file *pfs.File, paginationMarker
...
+		fis := newCircularList(number)
+		if err := s.Iterate(ctx, func(fi *pfs.FileInfo, _ fileset.File) error {
+			if pathIsChild(name, pfsfile.CleanPath(fi.File.Path)) {
+				fis.add(fi)
+			}
+			return nil
+		}); err != nil {
+			return errors.EnsureStack(err)
+		}
+		if err := fis.iterateReverse(cb); err != nil {
```
brycemcanally, line 302: nit, we could just return the error from the call.

src/server/pfs/server/driver_file.go
```diff
@@ -356,12 +392,41 @@ func (d *driver) walkFile(ctx context.Context, file *pfs.File, cb func(*pfs.File
...
+	}
+	if reverse {
+		fis := newCircularList(number)
+		if err := s.Iterate(ctx, func(fi *pfs.FileInfo, _ fileset.File) error {
+			fis.add(fi)
+			return nil
+		}); err != nil {
+			return errors.EnsureStack(err)
+		}
+		if err := fis.iterateReverse(cb); err != nil {
```
brycemcanally, line 417: same as above.

src/server/pfs/server/driver_file.go
```diff
@@ -300,45 +319,62 @@ func (d *driver) listFile(ctx context.Context, file *pfs.File, paginationMarker
...
+	// next index to be populated
+	index int
+	// size is the number of file infos in the ring
+	size int
+}
+
+func newCircularList(size int64) *circularList {
+	r := &circularList{}
+	r.buffer = make([]*pfs.FileInfo, size)
+	r.index = 0
```
brycemcanally, line 334: These index and size don't need to be set since the default is zero.

src/server/pfs/server/driver_file.go
```diff
@@ -300,45 +319,62 @@ func (d *driver) listFile(ctx context.Context, file *pfs.File, paginationMarker
...
+	// a circular buffer of file infos
+	buffer []*pfs.FileInfo
+	// next index to be populated
+	index int
+	// size is the number of file infos in the ring
+	size int
+}
+
+func newCircularList(size int64) *circularList {
+	r := &circularList{}
```
brycemcanally, line 332: It is a bit cleaner to just do this:
```
return &circularList{
  buffer: make([]*pfs.FileInfo, size),
}
```

src/server/pfs/server/driver_file.go
```diff
@@ -300,45 +319,62 @@ func (d *driver) listFile(ctx context.Context, file *pfs.File, paginationMarker
...
+	r.buffer = make([]*pfs.FileInfo, size)
+	r.index = 0
+	r.size = 0
+	return r
+}
+
+func (r *circularList) add(fi *pfs.FileInfo) {
+	r.buffer[r.index] = fi
+	r.index++
```
brycemcanally, line 341: We could just have a condition for the size increment, rather than increment then set it back to the length of the buffer. Something like this:
```
r.index++
// if we are at the end of the buffer, wrap around
if r.index == len(r.buffer) {
	r.index = 0
}
if r.size < len(r.buffer) {
	r.size++
}
	
```

src/server/pfs/server/driver_file.go
```diff
@@ -300,45 +319,62 @@ func (d *driver) listFile(ctx context.Context, file *pfs.File, paginationMarker
...
-	if index == 0 {
-		// no files were found
+	if r.size > len(r.buffer) {
+		r.size = len(r.buffer)
+	}
+}
+
+// returns the last element added to the ring
+func (r *circularList) get() *pfs.FileInfo {
+	if r.size == 0 {
```
brycemcanally, line 354: Is this check useful? It seems like it isn't needed for the current use case, and it wouldn't be helpful if get was called on a size zero buffer since it is returning a nil file info, rather than a nil file info and an error.

src/server/pfs/server/driver_file.go
```diff
@@ -300,45 +319,62 @@ func (d *driver) listFile(ctx context.Context, file *pfs.File, paginationMarker
...
-			index = int(number) - 1
+	r.index--
+	// if we are at the beginning of the buffer, wrap around
+	if r.index < 0 {
+		r.index = len(r.buffer) - 1
+	}
+	return r.buffer[r.index]
+}
+
+func (r *circularList) iterateReverse(cb func(*pfs.FileInfo) error) error {
```
brycemcanally, line 365: I think we would probably be better off not having a separate get method and just using a local index that starts at r.index.

src/server/pfs/server/driver_file.go
armaanv: makes sense, I've removed it and merged the get function into the iteration

src/server/pfs/server/driver_file.go
```diff
@@ -300,45 +316,56 @@ func (d *driver) listFile(ctx context.Context, file *pfs.File, paginationMarker
...
+func (r *circularList) add(fi *pfs.FileInfo) {
+	r.buffer[r.index] = fi
+	r.index++
+	// if we are at the end of the buffer, wrap around
+	if r.index == len(r.buffer) {
+		r.index = 0
+	}
+	if r.size < len(r.buffer) {
+		r.size++
+	} else {
```
brycemcanally, line 343: Is this else needed?

brycemcanally (approved):
lgtm, just one more question / change.

src/server/pfs/server/driver_file.go
armaanv: not needed actually. I'm removing it

PR #8476, opened 2022-12-21
Fix several goroutine leaks when closing a service environment

brycemcanally:
This PR fixes several goroutine leaks when closing a service environment. This is particularly problematic for the PFS tests since they set up a new environment for each test. This hasn't been an issue up to now because the PFS tests did not set up the PPS, Auth, and Enterprise services, which contain the leaks.


FahadBSyed (approved):

FahadBSyed:
The code change looks good, but there are a bunch of flakey tests that all failed at once, which seems suspicious

PR #8478, opened 2022-12-21
Warn about outdated pachctl

jrockway:
This implements some compatibility checking between pachctl (actually any Go client) and pachd.

Compatibility is defined as:

* Either the client or the server are 0.0.0 as developer builds are.
* If either the client or the server are a pre-release (nightly/alpha/beta/rc), then the server and client versions have to be an exact match.
* Otherwise, the major and minor number have to be the same.

See the test cases in `version_test.go`.

The client always calls InspectCluster before connecting, so this PR modifies InspectCluster to take the client's version as a parameter.  The server then checks that version against it's own version, and returns warnings in the reponse.  There is also a `version_warnings_ok` flag.  Old servers won't set that (since it's not in the message version they have), so the client can detect a way-too-old server.  If there are any warnings set, the client will log them at level error.  Technically this would be intrusive to users of the go client, but since the `pctx.TODO()` logger points to no-op logger until someone calls `InitPachctlLogger()` and that's a symbol they can't import, only pachctl users will ever see this.

It is a little weird to change InspectCluster from taking Empty to taking a message type, but it seems perfectly safe to me.  The Go client API doesn't change; only people that directly generate stubs and call methods on it (like some of our tests) are affected.  Users of the Go client that want to send a version have the option to do so with a new function in the client, InspectClusterWithVersion. 

The server logs at INFO level whenever an incompatible client is detected, so even if users miss the warnings, administrators can know.

Here's an example of what it looks like in pachctl (runs for every command, can't be turned off).  In this particular case, a "released" client is talking to a nightly build, which requires an exact version match between client and server: 

Capture

And the server logs:
Capture

The server log is the same for every case (modulo the `error` field), but the client message varies based on the constants in `admin/api_server.go`.  Feel free to wordsmith them.


src/version/versionpb/version.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	if rawServer == nil {
+		return errors.New("server version is nil")
+	}
+	if rawClient.IsDevelopment() || rawServer.IsDevelopment() {
+		return nil
+	}
+	s, c := rawServer.Canonical(), rawClient.Canonical()
+	if semver.Prerelease(s) != "" || semver.Prerelease(c) != "" {
+		// If either the client or the server are a prerelase, then the verions must match
```
robert-uhl, line 51: - s/prerelase/prerelease/
- s/verions/version/

src/version/versionpb/version.go
```diff
@@ -0,0 +1,67 @@
...
+	return v.Major == 0 && v.Minor == 0 && v.Micro == 0
+}
+
+var (
+	ErrClientTooOld        = errors.New("client is outdated")
+	ErrServerTooOld        = errors.New("server is outdated")
+	ErrIncompatiblePreview = errors.New("client and server versions must match exactly")
+)
+
+// IsCompatible determines if two versions are compatible.
```
robert-uhl, line 38: Is it worth spelling out the compatibility rules in English here, or is that duplicating the code (and hence liable to get out of date)?

src/version/versionpb/version_test.go
```diff
@@ -0,0 +1,102 @@
...
+		{
+			client:  Version{Major: 2, Minor: 5, Micro: 0},
+			server:  Version{Major: 2, Minor: 5, Micro: 0, Additional: "-nightly.20230101"},
+			wantErr: ErrIncompatiblePreview,
+		},
+		{
+			client:  Version{Major: 2, Minor: 5, Micro: 0, Additional: "-nightly.20230101"},
+			server:  Version{Major: 2, Minor: 5, Micro: 0, Additional: "-nightly.20230102"},
+			wantErr: ErrIncompatiblePreview,
+		},
```
robert-uhl, line 93: Is it worth adding an explicit _compatible_ preview test?

robert-uhl (approved):
This is good. I am personally concerned that it is a bit too strict — e.g. I think that a 2.5.0 server _ought_ to work fine with a 2.4.3 client — but this is the level of strictness we have decided to go with for now.

JimmyWhitaker:
This is great! Should we look at putting a link to the docs on how to upgrade? Our docs seem to slightly change between versions, so I'm not sure what the best way to link it is. (E.g. should we link to the docs for the pachd version or the most recent version). 

[Upgrade Docs](https://docs.pachyderm.com/2.3.x/deploy-manage/manage/upgrades/#3-upgrade-pachctl-version)

jrockway:
> This is great! Should we look at putting a link to the docs on how to upgrade? Our docs seem to slightly change between versions, so I'm not sure what the best way to link it is. (E.g. should we link to the docs for the pachd version or the most recent version).
> 
> [Upgrade Docs](https://docs.pachyderm.com/2.4.x/deploy-manage/manage/upgrades/#how-to-upgrade-pachyderm)

Yeah, I'm not sure how we can link stably; our redirect for /latest/deploy-manage/manage/upgrades redirects to the wrong place.  The docs for older versions may be gone by the time a message is being produced.  I think we'll just delegate this task to Google.

src/version/versionpb/version.go
jrockway: Done.

src/version/versionpb/version.go
jrockway: Done.

src/version/versionpb/version_test.go
jrockway: Done.

PR #8481, opened 2023-01-03
Djanicek/add coverage projects

djanicekpach:
For the most part, this PR has test changes to finish exercising pachctl commands with projects.

So, most of the changes are changing pfs.DefaultProjectName to a dynamic project name. This PR should make it so all of the non-auth commands are covered at the cli or api level.

There were a few changes for test stability that I made while I was running the tests and touching the code. Those changes were a bit different, so here's the explanation of what was changed with those tests and why:

- internal/testutil/kubeclient.go was not checking for pod readiness correctly, only for _running_ status which occurs much sooner than _readiness_. In particular, pod delete wasn't properly waiting for the pod to go down before checking that it was up, so it immediately found a running pod and assumed it was back online, when it had never gone down in the first place. Additionally, a new function was added to wait for any pod to be in a running and ready state, not just pachd.
- TestDatumSetCache in pachyderm_test.go uses the new pod delete waits to ensure etcd has a few seconds of uptime before it's deleted. Before, if etcd took too long to come up(~45 seconds) then very few datums would be processed. Since there are 90 datums, this could cause the test to take much longer than the 10 minute timeout. Now that it waits, we should guarantee a more consistent amount of up before being re-deleted.
- TestRapidUpdatePipelines in pachyderm_test.go was changed to use the new utility to wait for the pipeline to start before updating, rather than using sleep(). The pipeline name was changed to not be so long(it wasn't coming up the first time before  since the name was too long). Finally, I changed to use the constant retry rather than exponential so the tests finishes within 10 seconds of success, rather than 1 minute. I'm less sure these changes fully resolve issues on this test but they seem to help.

### DIFF
--- a/src/internal/testutil/kubeclient.go
+++ b/src/internal/testutil/kubeclient.go
@@ -2,6 +2,7 @@ package testutil
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"testing"
 	"time"
@@ -69,9 +70,11 @@ func DeletePod(t testing.TB, app, ns string) {
 					map[string]string{"app": app, "suite": "pachyderm"},
 				)),
 			})
+		waitForPodEvent(t, app, ns, watch.Deleted)
 		if err != nil {
 			return errors.EnsureStack(err)
 		}
+
 		if len(podList.Items) == 0 {
 			return nil
 		}
@@ -99,7 +102,7 @@ func DeletePod(t testing.TB, app, ns string) {
 		return nil
 	})
 
-	require.NoErrorWithinTRetry(t, 30*time.Second, func() error {
+	require.NoErrorWithinTRetry(t, 120*time.Second, func() error {
 		podList, err := kubeClient.CoreV1().Pods(ns).List(
 			context.Background(),
 			metav1.ListOptions{
@@ -159,8 +162,12 @@ func DeletePipelineRC(t testing.TB, pipeline, namespace string) {
 // PachdDeployment finds the corresponding deployment for pachd in the
 // kubernetes namespace and returns it.
 func PachdDeployment(t testing.TB, namespace string) *apps.Deployment {
+	return GetDeployment(t, "pachd", namespace)
+}
+
+func GetDeployment(t testing.TB, appName string, namespace string) *apps.Deployment {
 	k := GetKubeClient(t)
-	result, err := k.AppsV1().Deployments(namespace).Get(context.Background(), "pachd", metav1.GetOptions{})
+	result, err := k.AppsV1().Deployments(namespace).Get(context.Background(), appName, metav1.GetOptions{})
 	require.NoError(t, err)
 	return result
 }
@@ -173,14 +180,34 @@ func podRunningAndReady(e watch.Event) (bool, error) {
 	if !ok {
 		return false, errors.Errorf("unexpected object type in watch.Event")
 	}
-	return pod.Status.Phase == v1.PodRunning, nil
+	containersReady := true
+	for _, cs := range pod.Status.ContainerStatuses {
+		if !cs.Ready {
+			containersReady = false
+			break
+		}
+	}
+	podReady := true
+	for _, c := range pod.Status.Conditions {
+		if c.Type == v1.PodReady && c.Status != v1.ConditionTrue {
+			podReady = false
+			break
+		}
+	}
+	return pod.Status.Phase == v1.PodRunning && containersReady && podReady, nil
 }
 
 // WaitForPachdReady finds the pachd pods within the kubernetes namespace and
 // blocks until they are all ready.
+// WaitForPachdReady finds the pachd pods within the kubernetes namespace and
+// blocks until they are all ready.
 func WaitForPachdReady(t testing.TB, namespace string) {
+	WaitForDeploymentReady(t, "pachd", namespace)
+}
+
+func WaitForDeploymentReady(t testing.TB, appName string, namespace string) {
 	k := GetKubeClient(t)
-	deployment := PachdDeployment(t, namespace)
+	deployment := GetDeployment(t, appName, namespace)
 	for {
 		newDeployment, err := k.AppsV1().Deployments(namespace).Get(context.Background(), deployment.Name, metav1.GetOptions{})
 		require.NoError(t, err)
@@ -190,7 +217,7 @@ func WaitForPachdReady(t testing.TB, namespace string) {
 		time.Sleep(time.Second * 5)
 	}
 	watch, err := k.CoreV1().Pods(namespace).Watch(context.Background(), metav1.ListOptions{
-		LabelSelector: "app=pachd",
+		LabelSelector: fmt.Sprintf("app=%s", appName),
 	})
 	defer watch.Stop()
 	require.NoError(t, err)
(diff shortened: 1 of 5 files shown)


src/internal/testutil/kubeclient.go
```diff
@@ -210,3 +244,36 @@ func WaitForPachdReady(t testing.TB, namespace string) {
...
 	}
 }
+
+// Wait deployment only works for deployments and checks every pod.
+// This function is simpler and just unblocks as soon as 1 pod is read.
+// It can be used on non-Deployment objects like StatefulSets.
+func WaitForFirstPodReady(t testing.TB, appName string, namespace string) {
+	k := GetKubeClient(t)
+	watch, err := k.CoreV1().Pods(namespace).Watch(context.Background(), metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("app=%s", appName),
```
jrockway, line 254: This should probably be refactored to take a selector instead of hard-coding app.  The tests for pipelines below really want to supply pipelineName=<genreated name> and not app=pipeline (which applies to all pipelines, even ones that were started by a previous test when -clusters.reuse is true).

src/server/pachyderm_test.go
```diff
@@ -2370,12 +2374,12 @@ func TestDeletePipeline(t *testing.T) {
 
 	createPipelines()
 
-	deletePipeline := func(pipeline string) {
-		require.NoError(t, c.DeleteProjectPipeline(pfs.DefaultProjectName, pipeline, false))
+	deletePipeline := func(project, pipeline string) {
+		require.NoError(t, c.DeleteProjectPipeline(project, pipeline, false))
 		time.Sleep(5 * time.Second)
```
jrockway, line 2379: Why sleep?  Just let the retry logic retry.  (I think this is an effort to make the first attempt succeed, but we should be using the testing backoff, which should just try constantly every 1 second or so.  I don't think backoff.Retry is ever the right thing to use in tests, though.  I recall that you might have fixed the testing version of that, though.)

src/server/pachyderm_test.go
```diff
@@ -10676,14 +10701,23 @@ func TestDatumSetCache(t *testing.T) {
...
 	go func() {
-		ticker := time.NewTicker(time.Minute)
+		ticker := time.NewTimer(50 * time.Second)
 		defer ticker.Stop()
 		for {
 			select {
 			case <-ctx.Done():
 				return
 			case <-ticker.C:
+				// ticker is too smart and will adjust to always go at 1 minute, even if etcd is slow.
```
jrockway, line 10711: Why not time.After then?  We create a new ticker every iteration, seems like it would be easier to use time.After which will only tick once.

Comment is also slightly outdated; Minute changed to 50 seconds here.

src/server/pachyderm_test.go
```diff
@@ -10676,14 +10701,23 @@ func TestDatumSetCache(t *testing.T) {
...
+				// So, we pause the ticker while the pod restart occurs so we know how long the pod is executing for.
 				tu.DeletePod(t, "etcd", ns)
+				tu.WaitForFirstPodReady(t, "etcd", ns)
```
jrockway, line 10714: This seems like the only place where we call DeletePod and there is a lot of action at a distance here (it doesn't merely delete the pod, it also waits for a replacement to be scheduled, but not ready).  I think it would be better to do this logic somewhere in this test file and do something more reliable than DeletePod; scale the etcd StatefulSet to 0, wait for there to be no matching etcd pods, scale etcd to 1, wait for matching pod to be ready.  Overall DeletePod chooses an unreliable and abstraction-breaking algorithm for the sake of being general, but this is the only place that calls it, so it doesn't seem like a good tradeoff to me.

src/server/pfs/server/testing/server_test.go
```diff
@@ -2788,31 +2791,32 @@ func TestPFS(suite *testing.T) {
...
 
-		branchInfos, err := env.PachClient.ListProjectBranch(pfs.DefaultProjectName, repo)
+		branchInfos, err := env.PachClient.ListProjectBranch(project, repo)
 		require.NoError(t, err)
 		require.Equal(t, len(expectedBranches), len(branchInfos))
 		for i, branchInfo := range branchInfos {
 			// branches should return in newest-first order
 			require.Equal(t, expectedBranches[len(branchInfos)-i-1], branchInfo.Branch.Name)
-
+			require.Equal(t, project, branchInfo.Branch.Repo.Project.GetName())
```
jrockway, line 2819: It's preferred to use the accessors here; branchInfo.GetBranch().GetRepo().GetProject().GetName().

jrockway (commented):
It looks like there is a lot that can be deleted from testutil/kubeclient, which I think we should do here; or rather, just submit the changes that don't require refactoring kubeclient, and then approach that in a follow-up?  My concern with kubeclient is that it's being generalized in a flaky way, but most of the methods are only called once, so doing it specifically in a better way would end up being better.

src/internal/testutil/kubeclient.go
```diff
@@ -190,7 +224,7 @@ func WaitForPachdReady(t testing.TB, namespace string) {
 		time.Sleep(time.Second * 5)
 	}
 	watch, err := k.CoreV1().Pods(namespace).Watch(context.Background(), metav1.ListOptions{
-		LabelSelector: "app=pachd",
+		LabelSelector: fmt.Sprintf("app=%s", appName),
```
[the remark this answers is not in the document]
djanicekpach: Huh, I really thought it was used, but I must have gotten confused there. That changes a lot. 

Would these changes work for you?
- In kubeclient.go delete everything except `GetKubeClient`, and `DeletePipelineRC`(I'll make a ticket and assign to myself to revisit `DeletePipelineRC`, I think it has a similar issue to DeletePods)
- Move the delete pod and readiness wait logic into the tests
- For logic on readiness, I was planning to follow the logic of minikubeTestEnv.waitForPachd, but I think it does still use the readiness check instead of sending a request to a new port: `p.Status.Phase == v1.PodRunning && strings.HasSuffix(p.Spec.Containers[0].Image, ":"+version) && p.Status.ContainerStatuses[0].Ready && len(pachds.Items) == 1`

src/server/pfs/server/testing/server_test.go
djanicekpach: changed

src/server/pachyderm_test.go
djanicekpach: changed. I didn't know time.After was an option.

src/server/transaction/server/testing/server_test.go
```diff
@@ -237,22 +260,38 @@ func TestTransactions(suite *testing.T) {
...
 		require.NoError(t, err)
-		err = txnClient.FinishProjectCommit(pfs.DefaultProjectName, "foo", "master", "")
+		err = txnClient.FinishProjectCommit(project, "foo", "master", "")
 		require.NoError(t, err)
 
-		_, err = txnClient.StartProjectCommit(pfs.DefaultProjectName, "foo", "master")
+		_, err = txnClient.StartProjectCommit(project, "foo", "master")
 		require.YesError(t, err)
 		require.Matches(t, "already has a commit in this transaction", err.Error())
+		//Delete and verify deletion occurs as well
```
jrockway, line 283: There should be a space between // and Delete.  Comments for humans have a space, comments for tooling (`//nolint:whatever`, `//go:build`, etc.) omit the space.

src/server/transaction/server/testing/server_test.go
```diff
@@ -119,25 +120,29 @@ func TestTransactions(suite *testing.T) {
...
+		project := testutil.UniqueString("p-")
 		txn, err := env.PachClient.StartTransaction()
 		require.NoError(t, err)
 
 		txnClient := env.PachClient.WithTransaction(txn)
 
+		_, err = txnClient.PfsAPIClient.CreateProject(txnClient.Ctx(), &pfs.CreateProjectRequest{
+			Project: client.NewProject(project), // projects is not transactional, so it shouldn't be added to inspect transaction request count
+		})
+		require.NoError(t, err)
```
jrockway, line 132: It would be nice to supply a message with each assertion.  `require.NoError(t, err, "CreateProject(%s) should succeed", project)`.

src/server/pachyderm_test.go
```diff
@@ -104,6 +105,56 @@ func createTestCommits(t *testing.T, repoName, branchName string, numCommits int
 	}
 }
 
+func deleteEtcd(t *testing.T, ctx context.Context, namespace string) {
```
jrockway, line 108: This should call `t.Helper()` so that assertion failures appear from the caller's perspective.  Same with waitForOnePodReady.

src/server/pachyderm_test.go
```diff
@@ -104,6 +105,56 @@ func createTestCommits(t *testing.T, repoName, branchName string, numCommits int
...
+	label := "app=etcd"
+	etcd, err := kubeClient.CoreV1().Pods(namespace).
+		List(ctx, metav1.ListOptions{LabelSelector: label})
+	require.NoError(t, err, "Error attempting to find etcd pod.")
+	require.Equal(t, 1, len(etcd.Items), "etcd did not have the correct number of pods.")
+
+	watcher, err := kubeClient.CoreV1().Pods(namespace).
+		Watch(ctx, metav1.ListOptions{LabelSelector: label})
+	defer watcher.Stop()
+	require.NoError(t, err)
```
jrockway, line 119: After you add t.Helper, you'll want to make sure that every failure is annotated.  (`"starting pod watch should succeed"` etc.)

jrockway (approved):
Looks good!

PR #8485, opened 2023-01-04
Gracefully handle old pipelines where output commit is ALIAS and meta commit is AUTO

acohen4:
In #8116 we fixed an issue where ALIAS output commits would get created with corresponding AUTO meta commits. This didn't handle previously faulty DAGs. In this PR, we gracefully handle DAGs already containing the faulty structure.


src/server/worker/pipeline/transform/pending_job.go
```diff
@@ -107,7 +107,8 @@ func (pj *pendingJob) load() error {
 				return errors.EnsureStack(err)
 			}
 			// both commits must have succeeded - a validation error will only show up in the output
-			if metaCI.Error == "" && outputCI.Error == "" {
+			// the commit must also not be of type ALIAS, to ensure that we have a corresponding job
+			if metaCI.Error == "" && outputCI.Error == "" && outputCI.Origin.Kind != pfs.OriginKind_ALIAS {
```
jrockway, line 111: Do we want to add a log message around line 117 so we know which commit we picked as the base commit?  Maybe it gets recorded elsewhere?

jrockway (approved):

src/server/worker/pipeline/transform/pending_job.go
acohen4: good idea!

PR #8486, opened 2023-01-04
Ensure we only create jobs for output branch commits [PFS-35]

brycemcanally:
This PR ensures we only create jobs for output branch commits.


acohen4 (approved):
LGTM

PR #8501, opened 2023-01-09
[CORE-1395] Move PostgreSQL collection v2.5.0 migration tests

robert-uhl:


albscui (approved):

robert-uhl:
> We could also just remove `src/internal/collection/v2.5.0.go` altogether right?

It would probably have helped had I merged & pushed … fixed. Sorry about that.

PR #8502, opened 2023-01-09
[PFS-36] Fix the locking in the consistent hashing package

brycemcanally:
This PR fixes the locking in the consistent hashing package. The prior locking had a few issues that could result with us using the wrong context and getting into a bad state in certain situations, particularly with regards to locks being lost at inopportune times. The changes in this PR ensure that we are always using the lock context, which can be canceled due to the user context being canceled or the lock being lost. When releasing the lock through unlock or rebalance, we ensure that either the context is canceled or the unlock call succeeds. If the context is canceled, the underlying session resources are cleaned up which will result with the backing lease eventually timing out.

### DIFF
--- a/src/internal/consistenthashing/etcd.go
+++ b/src/internal/consistenthashing/etcd.go
@@ -9,13 +9,13 @@ import (
 	"time"
 
 	"github.com/gogo/protobuf/types"
-	"github.com/hashicorp/go-multierror"
 	etcd "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/client/v3/concurrency"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 	"golang.org/x/sync/errgroup"
 
+	"github.com/pachyderm/pachyderm/v2/src/internal/backoff"
 	"github.com/pachyderm/pachyderm/v2/src/internal/collection"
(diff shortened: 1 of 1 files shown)


FahadBSyed (approved):

src/internal/consistenthashing/etcd.go
```diff
@@ -208,6 +205,27 @@ func (ring *Ring) getIndex(key string) int {
...
+	// The lock is released when either:
+	// - The lock context is canceled.
+	// - The unlock call completes successfully.
+	ctx := lockInfo.ctx
+	return backoff.RetryNotify(func() error {
+		if errors.Is(ctx.Err(), context.Canceled) {
+			return nil
+		}
+		return lockInfo.lock.Unlock(ctx)
+	}, backoff.NewInfiniteBackOff(), func(err error, d time.Duration) error {
```
acohen4, line 223: is there a concern with having an infinite backoff here? If we're already in the case where we're failing to unlock, shouldn't we be retrying with more urgency lol

brycemcanally: We are using an exponential backoff because that is our standard strategy for transient errors and we use the infinite backoff because we have to ensure that the lock is released.

PR #8503, opened 2023-01-09
fix typo

djanicekpach:
typo found while testing. Quick one-minute fix so I'm just submitting it now.
https://pachyderm.atlassian.net/browse/CORE-1443


albscui (approved):
oops my bad 😅

PR #8506, opened 2023-01-09
Migrate PutFileURL and GetFileURL to Go-Cloud-SDK

FahadBSyed:
To increase maintainability, this change migrates to using the [go-cloud-sdk](https://gocloud.dev/) for interfacing with cloud storage providers. This PR addresses PutFileURL and GetFileURL and leaves the existing object client code in place for the storage instances. Future PRs will fully migrate to the go-cloud-sdk once we have battle tested it and are confident that using it does not introduce any strange bugs.


src/server/pfs/server/url_worker.go
```diff
@@ -53,23 +55,28 @@ func (d *driver) URLWorker(ctx context.Context) {
 	})
 }
 
-func (d *driver) processPutFileURLTask(ctx context.Context, task *PutFileURLTask) (*types.Any, error) {
+func (d *driver) processPutFileURLTask(ctx context.Context, task *PutFileURLTask) (any *types.Any, retErr error) {
```
brycemcanally, line 58: `any` can just be `_`

src/server/pfs/server/url_worker.go
```diff
@@ -90,7 +97,7 @@ func (d *driver) processPutFileURLTask(ctx context.Context, task *PutFileURLTask
 	return serializePutFileURLTaskResult(result)
 }
 
-func (d *driver) processGetFileURLTask(ctx context.Context, task *GetFileURLTask) (*types.Any, error) {
+func (d *driver) processGetFileURLTask(ctx context.Context, task *GetFileURLTask) (any *types.Any, retErr error) {
```
brycemcanally, line 100: Same as above.

src/server/pfs/server/storage_bucket.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, errors.EnsureStack(errors.Wrapf(err, "error creating session s", url.Bucket))
+	}
+	bucket, err = s3blob.OpenBucket(ctx, sess, parts[1], nil)
+	if err != nil {
+		return nil, errors.EnsureStack(errors.Wrapf(err, "error opening bucket %s", url.Bucket))
+	}
+	return bucket, nil
+}
+
+func importObj(ctx context.Context, w io.Writer, name, bucketName string, bucket *blob.Bucket) (retErr error) {
```
brycemcanally, line 108: I'm not a fan of passing in a parameter just for having it in the errors. I think the errors in this and export should be generic and the bucket name can be reported by the caller.

src/server/pfs/server/url_worker.go
```diff
@@ -53,23 +55,28 @@ func (d *driver) URLWorker(ctx context.Context) {
...
+			retErr = multierror.Append(retErr, errors.Wrapf(err, "error closing bucket %s", url.Bucket))
+		}
+	}()
 	prefix := strings.TrimPrefix(url.Object, "/")
 	result := &PutFileURLTaskResult{}
 	if err := log.LogStep(ctx, "putFileURLTask", func(ctx context.Context) error {
 		return d.storage.WithRenewer(ctx, defaultTTL, func(ctx context.Context, renewer *fileset.Renewer) error {
 			id, err := d.withUnorderedWriter(ctx, renewer, func(uw *fileset.UnorderedWriter) error {
 				for _, path := range task.Paths {
 					if err := miscutil.WithPipe(func(w io.Writer) error {
```
brycemcanally, line 78: This pipe isn't needed with the new interface. We can simplify this and just pass the bucket reader to the put call.

src/server/pfs/server/storage_bucket.go
brycemcanally: We can just get rid of import / export since the pipes are no longer needed.

src/server/pfs/server/url.go
```diff
@@ -95,10 +105,15 @@ func putFileURLRecursive(ctx context.Context, taskService task.Service, uw *file
...
 			return errors.Wrapf(err, "error parsing url %v", src)
 		}
-		objClient, err := obj.NewClientFromURLAndSecret(ctx, url, false)
+		bucket, err := openBucket(ctx, url)
 		if err != nil {
 			return err
 		}
+		defer func() {
+			if err := bucket.Close(); err != nil {
+				retErr = multierror.Append(retErr, errors.Wrapf(err, "error closing bucket %s", url.Bucket))
```
brycemcanally, line 114: This needs to be operating on the error returned by the function being run in the error group, not the outer function.

src/server/pfs/server/url.go
```diff
@@ -117,17 +132,24 @@ func putFileURLRecursive(ctx context.Context, taskService task.Service, uw *file
 			}
 			return nil
 		}
-		if err := objClient.Walk(ctx, url.Object, func(p string) error {
+		list := bucket.List(&blob.ListOptions{Prefix: url.Object})
+		var listObj *blob.ListObject
+		for {
+			listObj, err = list.Next(ctx)
+			if err != nil {
+				if err == io.EOF {
```
brycemcanally, line 140: We might want this to be an `errors.Is` check.

src/server/pfs/server/url.go
```diff
@@ -117,17 +132,24 @@ func putFileURLRecursive(ctx context.Context, taskService task.Service, uw *file
...
+					break
+				}
+				return errors.Wrapf(err, "error iterating through list for bucket %s", url.Bucket)
+			}
+			log.Debug(ctx, "object stats", zap.String("object", listObj.Key), zap.Int64("size", listObj.Size))
 			if len(paths) >= defaultURLTaskSize {
 				if err := createTask(); err != nil {
-					return err
+					return errors.Wrapf(err, "error from callback on key %s in bucket %s", listObj.Key, url.Bucket)
```
brycemcanally, line 148: I think this message could generally be misleading since the key / bucket will likely not have anything to do with an error here. I think we should just return the error as is.

brycemcanally: Creating a logger with a bucket field for this section of code would probably be a cleaner way to handle this.

etc/helm/pachyderm/templates/pachd/storage-secret.yaml
```diff
@@ -15,6 +15,7 @@ data:
   {{- if eq (include "pachyderm.storageBackend" . )  "GOOGLE" }}
   GOOGLE_BUCKET: {{ required "Google bucket required" .Values.pachd.storage.google.bucket | b64enc | quote }}
   GOOGLE_CRED: {{ .Values.pachd.storage.google.cred | b64enc | quote }}
+  GOOGLE_APPLICATION_CREDENTIALS: {{ "/pachyderm-storage-secret/GOOGLE_CRED" | b64enc | quote }}
```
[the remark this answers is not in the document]
FahadBSyed: No, as far as I can tell the GCP configuration used by go-cdk requires us to supply a path to a file rather than the raw value

src/server/pfs/server/url.go
FahadBSyed: The example in the [godocs ](https://pkg.go.dev/gocloud.dev/blob@v0.27.0#example-Bucket.List) for `.List()` uses io.EOF so I think we should leave it this way

brycemcanally: That isn't the recommended way to perform error checks anymore (https://pkg.go.dev/errors). We've had errors in the past using equality checks, so it would be better to future proof against that.

etc/helm/pachyderm/templates/pachd/storage-secret.yaml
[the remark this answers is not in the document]
brycemcanally: Okay, we should probably also add a note that this is for go-cdk similar to what you added below.

src/internal/obj/factory.go
```diff
@@ -418,36 +417,45 @@ type ObjectStoreURL struct {
...
+
 // ParseURL parses an URL into ObjectStoreURL.
 func ParseURL(urlStr string) (*ObjectStoreURL, error) {
 	u, err := url.Parse(urlStr)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error parsing url %v", urlStr)
 	}
 	switch u.Scheme {
-	case "s3", "gcs", "gs", "local":
+	case "azblob", "gcs", "gs", "local", "s3":
```
brycemcanally, line 443: Is `gs` valid?

brycemcanally: It looks like `gs` is the standard scheme and I cannot find a reference to `gcs`.

brycemcanally: Basically, I don't think we should be supporting schemes that are not actually supported by the backend. If we could do that replacement in one place and only operate on valid schemes afterwards, that would be ideal.

src/server/pfs/server/url_worker.go
```diff
@@ -99,20 +112,41 @@ func (d *driver) processGetFileURLTask(ctx context.Context, task *GetFileURLTask
...
 		err := src.Iterate(ctx, func(fi *pfs.FileInfo, file fileset.File) error {
 			if fi.FileType != pfs.FileType_FILE {
 				return nil
 			}
-			return miscutil.WithPipe(func(w io.Writer) error {
-				return errors.EnsureStack(file.Content(ctx, w))
-			}, func(r io.Reader) error {
-				return errors.EnsureStack(objClient.Put(ctx, filepath.Join(url.Object, fi.File.Path), r))
-			})
+			exists, err := bucket.Exists(ctx, fi.File.Path)
```
brycemcanally, line 129: I don't think we want this check. Object storage puts are idempotent and this prevents us from overwriting a file in a subsequent get file url. We may want to consider a more proper sync later though, because we won't be deleting files with either approach.

src/server/pfs/server/storage_bucket.go
```diff
@@ -0,0 +1,103 @@
...
+	_ "gocloud.dev/blob/gcsblob"
+	_ "gocloud.dev/blob/memblob"
+	_ "gocloud.dev/blob/s3blob"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/obj"
+)
+
+// openBucket swaps out client backend implementation for s3, gcp, and azure with go-cloud-sdk.
+func openBucket(ctx context.Context, url *obj.ObjectStoreURL) (bucket *blob.Bucket, err error) {
```
brycemcanally, line 27: I don't think we need the named returns here.

src/server/pfs/server/storage_bucket.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		return bucket, nil
+	case "test-minio", "minio":
+		return handleMinio(ctx, url)
+	default:
+		return nil, errors.Errorf("unrecognized object store: %s", url.Scheme)
+	}
+}
+
+func handleMinio(ctx context.Context, url *obj.ObjectStoreURL) (bucket *blob.Bucket, err error) {
```
brycemcanally, line 60: Same as above.

src/server/pfs/server/storage_bucket_test.go
```diff
@@ -0,0 +1,125 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30000)
+	defer cancel()
+	bucket, err := openBucket(ctx, url)
+	defer func() {
+		require.NoError(t, bucket.Close())
+	}()
+	require.NoError(t, err, "should be able to open bucket")
+	objName := randutil.UniqueString("test-object-")
+	writeToObjStorage(ctx, t, bucket, objName)
+	buf := bytes.NewBuffer(make([]byte, 0))
```
brycemcanally, line 93: You can create a valid bytes buffer with `&bytes.Buffer{}`

src/server/pfs/server/storage_bucket_test.go
```diff
@@ -0,0 +1,125 @@
...
+	readFromObjStorage(ctx, t, bucket, objName, buf)
+	require.Equal(t, buf.String(), objName)
+	err = bucket.Delete(ctx, objName)
+	require.NoError(t, err, "should be able to delete object")
```
brycemcanally, line 97: You could just inline the delete call.

src/server/pfs/server/storage_bucket_test.go
```diff
@@ -0,0 +1,125 @@
...
+	// We may have to check if we have valid ssl for our ECS configuration.
+	require.NoError(t, os.Setenv("DISABLE_SSL", "true"))
+	bucketName := os.Getenv("ECS_CLIENT_BUCKET")
+	url, err := obj.ParseURL("s3://" + bucketName)
+	require.NoError(t, err, "should be able to parse url")
+	writeReadDelete(t, url)
+}
+
+func writeReadDelete(t *testing.T, url *obj.ObjectStoreURL) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30000)
```
brycemcanally, line 84: That is a long timeout.

src/server/pfs/server/url.go
```diff
@@ -117,17 +136,24 @@ func putFileURLRecursive(ctx context.Context, taskService task.Service, uw *file
...
+				if err == io.EOF {
+					break
+				}
+				return errors.Wrapf(err, "error iterating through list for bucket %s", url.Bucket)
+			}
+			log.Debug(ctx, "object stats", zap.String("object", listObj.Key), zap.Int64("size", listObj.Size))
 			if len(paths) >= defaultURLTaskSize {
 				if err := createTask(); err != nil {
-					return err
+					return errors.Wrap(err, "error from callback")
```
brycemcanally, line 152: This wrap certainly seems excessive.

src/server/pfs/server/storage_bucket_test.go
FahadBSyed: whoops, I must have left that after introspecting with my debugger

src/server/pfs/server/storage_bucket_test.go
FahadBSyed: What do you mean by `inline` here?

FahadBSyed: Ahh I see

src/internal/obj/factory.go
```diff
@@ -418,51 +417,71 @@ type ObjectStoreURL struct {
...
+			Params: u.RawQuery,
+		}
+	default:
+		// return nil, errors.Errorf("unrecognized object store: %s", u.Scheme)
+		return nil, errors.Errorf("unrecognized object store: %s", u.Scheme)
+	}
+	switch objStoreUrl.Scheme {
+	case "wasb":
+		objStoreUrl.Scheme = "azblob"
+	case "gcs": // assuming 'gcs' is an alias for 'gs'
```
brycemcanally, line 481: Would you move this to immediately after the parse above with a TODO to get rid of it when we are okay with removing support for that non-existent scheme. I don't think this would work as is since gcs was removed from the above switch. Also, could we just handle the wasb -> azblob in the wasb case above.

src/server/pfs/server/url.go
```diff
@@ -54,21 +58,31 @@ func putFileURL(ctx context.Context, taskService task.Service, uw *fileset.Unord
...
+			if err := r.Close(); err != nil {
+				retErr = multierror.Append(retErr, errors.Wrapf(err, "error closing reader for bucket %s", url.Bucket))
+			}
+		}()
+		return 0, uw.Put(ctx, dstPath, tag, true, r)
 	}
 }
 
-func putFileURLRecursive(ctx context.Context, taskService task.Service, uw *fileset.UnorderedWriter, dst, tag string, src *pfs.AddFile_URLSource) error {
+func putFileURLRecursive(ctx context.Context, taskService task.Service, uw *fileset.UnorderedWriter, dst, tag string, src *pfs.AddFile_URLSource) (retErr error) {
```
brycemcanally, line 85: retErr isn't needed.

src/server/pfs/server/url_worker.go
```diff
@@ -90,29 +102,43 @@ func (d *driver) processPutFileURLTask(ctx context.Context, task *PutFileURLTask
...
+			w, err := bucket.NewWriter(ctx, fi.File.Path, nil)
+			if err != nil {
+				return errors.EnsureStack(err)
+			}
+			defer func() {
+				if err := w.Close(); err != nil {
+					retErr = multierror.Append(retErr, errors.Wrapf(err, "error closing writer for bucket %s", url.Bucket))
+				}
+			}()
+			err = file.Content(ctx, w)
```
brycemcanally, line 137: nit, you could just return the result of the call.

src/internal/obj/factory.go
```diff
@@ -418,51 +417,69 @@ type ObjectStoreURL struct {
...
 	}
+	// TODO: remove once we remove support for non-existing schemes.
+	if u.Scheme == "gcs" {
+		u.Scheme = "gs"
+	}
+	var objStoreUrl *ObjectStoreURL
 	switch u.Scheme {
-	case "s3", "gcs", "gs", "local":
-		return &ObjectStoreURL{
+	// gcs seems to be unused, but is left here in case it is used by customers.
```
brycemcanally, line 448: This comment is no longer needed.

brycemcanally (approved):
lgtm, you just need to remove a comment.

PR #8507, opened 2023-01-10
Make `pachctl copy file` able to copy files between projects

albscui:
New command help page.

```
pachctl copy file --help                                                                        
Copy files between pfs paths.

Usage:
  pachctl copy file <src-repo>@<src-branch-or-commit>:<src-path> <dst-repo>@<dst-branch-or-commit>:<dst-path> [flags]

Aliases:
  file, files

Examples:

# copy between repos within the current project defined by the pachyderm context
# defaults to the "default" project
$ pachctl copy file srcRepo@master:/file destRepo@master:/file

# copy within a specified project
$ pachctl copy file srcRepo@master:/file destRepo@master:/file --project myProject

# copy from the current project to a different project
# here, srcRepo is in the current project, while destRepo is in myProject
$ pachctl copy file srcRepo@master:/file destRepo@master:/file --dest-project myProject

# copy from a different project to the current project
# here, srcRepo is in myProject, while destRepo is in the current project
$ pachctl copy file srcRepo@master:/file destRepo@master:/file --src-project myProject

# copy between repos across two different projects
# here, srcRepo is in project1, while destRepo is in project2
$ pachctl copy file srcRepo@master:/file destRepo@master:/file --src-project project1 --dest-project project2


Flags:
  -a, --append                Append to the existing content of the file, either from previous commits or previous calls to 'put file' within this commit.
      --dest-project string   Project in which the destination repo is located. This overrides --project.
  -h, --help                  help for file
      --project string        Project in which both source and destination repos are located. (default "default")
      --src-project string    Project in which the source repo is located. This overrides --project.
```

### DIFF
--- a/src/server/pfs/cmds/cmds.go
+++ b/src/server/pfs/cmds/cmds.go
@@ -1362,8 +1362,7 @@ $ {{alias}} srcRepo@master:/file destRepo@master:/file --src-project myProject
 
 # copy between repos across two different projects
 # here, srcRepo is in project1, while destRepo is in project2
-$ {{alias}} srcRepo@master:/file destRepo@master:/file --src-project project1 --dest-project project2
-`,
+$ {{alias}} srcRepo@master:/file destRepo@master:/file --src-project project1 --dest-project project2`,
 		Run: cmdutil.RunFixedArgs(2, func(args []string) (retErr error) {
 			if srcProject == "" {
 				srcProject = project
--- a/src/server/pfs/cmds/cmds.go
+++ b/src/server/pfs/cmds/cmds.go
@@ -1339,16 +1339,23 @@ $ {{alias}} repo@branch -i http://host/path`,
 		})
 	commands = append(commands, cmdutil.CreateAliases(putFile, "put file", files))
 
+	var srcProject, destProject string
 	copyFile := &cobra.Command{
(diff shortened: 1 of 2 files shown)


src/server/pfs/cmds/cmds.go
```diff
@@ -1371,6 +1378,8 @@ $ {{alias}} repo@branch -i http://host/path`,
 	}
 	copyFile.Flags().BoolVarP(&appendFile, "append", "a", false, "Append to the existing content of the file, either from previous commits or previous calls to 'put file' within this commit.")
 	copyFile.Flags().StringVar(&project, "project", project, "Project in which repo is located.")
```
djanicekpach, line 1380: Can you add some more info in help text about how these 3 project flags interact. The behavior seems good for advanced usage, but I'm not sure I'd know how this fallback behavior works starting out. Maybe for each, add a line about what the default is like:
"Project in which the source repo is located. Uses the --project flag if not supplied"
and for --project "Project in which repo**s are** located. Uses the context default if not supplied."

I'm trying to keep it short, but also explicit, so feel free to change if you have a better idea for help text.

djanicekpach (changes requested):

src/server/pfs/cmds/cmds.go
albscui: I updated the command with examples as described above.

djanicekpach: Awesome!

djanicekpach (approved):

PR #8510, opened 2023-01-10
[CORE-1444] Make DeleteRepos and DeletePipelines similar

robert-uhl:
This PR makes repo & pipeline deletion similar:

- `DeleteRepo` & `DeletePipeline` are intended to delete a single repo or pipeline. `DeletePipeline`’s `all` member is still supported, but deprecated — it only deletes all pipelines in the default project (for backwards compatibility, since the default project is the only one visible to clients ≤ v2.4).
- `pachctl delete {repo,pipeline} --all` delete all repos or pipelines in the current project.
- `pachctl delete {repo,pipeline} --all --all-projects` delete all repos or pipelines in all projects.


loganripplinger (commented):
LGTM

src/server/pps/server/api_server.go
```diff
@@ -2955,6 +2959,46 @@ func (a *apiServer) deletePipelineInTransaction(txnCtx *txncontext.TransactionCo
...
+		if !request.GetAll() && !projects[pipelineInfo.GetPipeline().GetProject().GetName()] {
+			return nil
+		}
+		dr.Pipeline = pipelineInfo.Pipeline
+		err := a.deletePipeline(ctx, dr)
+		if err == nil {
+			deleted[pipelineInfo.Pipeline.String()] = struct{}{}
+			pp = append(pp, pipelineInfo.Pipeline)
+		}
+		return err
```
albscui, line 2995: Maybe aggregate this error into a `multierror` as well?

robert-uhl: Ditto here — I think that it is probably better to terminate early rather than continue.

albscui (approved):
LGTM

PR #8514, opened 2023-01-13
Add Pachctl connect command

chainlink:
This adds a new sub command to pachctl called `connect` which is a quick way to connect to a pachyderm cluster

Old Method:

```
echo '{"pachd_address": "grpcs://aws-namespace-002.workspace.pachyderm.com:443"}' | pachctl config set context aws-namespace-002 --overwrite && pachctl config set active-context aws-namespace-002
```

New Method:

```
pachctl connect grpc://localhost:80 
```

TODO
- [x] Maybe set context name based on the address? 
- [x] Add Tests
- [x] Add CLI docs


src/server/config/cmds.go
```diff
@@ -53,6 +53,38 @@ func deduceActiveEnterpriseContext(cfg *config.Config) (string, error) {
...
+				if !errors.Is(err, grpcutil.ErrNoPachdAddress) {
+					return err
+				}
+			} else {
+				context.PachdAddress = pachdAddress.Qualified()
+			}
+			name := "local"
+			cfg.V2.Contexts[name] = &context
+			cfg.V2.ActiveContext = name
+			return cfg.Write()
```
jrockway, line 80: It would be nice if we did a quick sanity check.  My initial thought with this feature would be to run "auth login", but I'm not sure what happens if you "auth login" when auth is off (probably a confusing error message).  Ignoring that, I think it would be nice to do an InspectCluster call just to check that the connection works before writing the file.  Maybe that's annoying though; interested in your thoughts there.

jrockway (approved):
This will be nice!

src/server/config/cmds.go
chainlink: I changed things around here a bit and added some more guardrails and output. I also set the context name to the passed in address vs the fully qualified one. It also checks to see if there's a matching context first and if so, doesn't modify it (so it doesn't blow away any auth or other stuff that might be saved there)

chainlink: w.r.t connecting sanity checks, I think I'd like to leave those out for now, I'd like to keep this command "offline" for now, but we could potentially add those checks in a second pass at this command.

jrockway: OK, this looks good to me.  Switching context by address is a really good idea.

jrockway (approved):
Very nice!

PR #8515, opened 2023-01-13
[Jupyter] Qualify unmounted repos keys by project

smalyala:
There were a couple spots where project wasn't added as a qualifier to repos in the unmounted repositories list. This led to funky behavior with repos of the same name in different projects. When one of those repos was mounted, the "Mount" button for all other repos sharing the same name was disabled if the selected branch in the dropdown matched the branch name of the mounted repo's branch.


djanicekpach (approved):
Tested and looks good!

PR #8516, opened 2023-01-14
[INT-802] Implement the S3 Gateway bucket name proposal

msteffen:
Design doc: https://docs.google.com/document/d/1fZOMtcS0onT2AKjd3qqP9FsssjIa_uvNpJc9oJ1ADI0/edit#heading=h.od8ud5qhgnpc

* Remove meta repos from the S3 gateway
* ListBucket returns exclusively bucket names of the form branch.repo.project (even when branch="master" and project="default")
* Otherwise, preserve backwards compatibility when resolving bucket names (so "s3://repo" and "s3://branch.repo" still resolve correctly as well)

### DIFF
--- a/src/server/pfs/s3/driver.go
+++ b/src/server/pfs/s3/driver.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/gorilla/mux"
 	"github.com/pachyderm/pachyderm/v2/src/client"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/grpcutil"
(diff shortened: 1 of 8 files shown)


src/server/pfs/s3/driver.go
```diff
@@ -54,47 +53,20 @@ func NewMasterDriver() *MasterDriver {
...
-	if b.Repo.Type == pfs.UserRepoType {
-		return fmt.Sprintf("%s.%s.%s", b.Name, b.Repo.Name, b.Repo.Project.Name)
-	}
-	return fmt.Sprintf("%s.%s.%s.%s", b.Name, b.Repo.Type, b.Repo.Name, b.Repo.Project.Name)
-}
-
 func (d *MasterDriver) listBuckets(pc *client.APIClient, r *http.Request, buckets *[]*s2.Bucket) error {
-	var isProjectAware = mux.Vars(r)[isProjectAwareVar] == isProjectAwareValue
-	repos, err := pc.ListRepoByType("") // get repos of all types
+	repos, err := pc.ListRepo() // get repos of all types
```
robert-uhl, line 57: `ListRepo` only returns user repos, which I think is what you intend — but the comment is now incorrect.

src/server/pfs/s3/driver.go
```diff
@@ -54,47 +53,20 @@ func NewMasterDriver() *MasterDriver {
...
-			var name string
-			if isProjectAware {
-				name = projectBranchBucketName(branch)
-			} else {
-				name = branchBucketName(branch)
-			}
+		for _, b := range repo.Branches {
 			*buckets = append(*buckets, &s2.Bucket{
-				Name:         name,
+				Name:         fmt.Sprintf("%s.%s.%s", b.Name, b.Repo.Name, b.Repo.Project.Name),
```
robert-uhl, line 69: You might wish to either use `GetRepo().GetProject().GetName()` &c., or check for nils.

src/server/pfs/s3/driver.go
```diff
@@ -103,10 +75,12 @@ func (d *MasterDriver) listBuckets(pc *client.APIClient, r *http.Request, bucket
...
-	branch := "master"
-	var repo *pfs.Repo
+func bucketNameToCommit(bucketName string) (*pfs.Commit, error) {
+	var (
+		id     string
+		branch string
+		repo   *pfs.Repo
+	)
 
 	// the name is [commitID.][branch. | branch.type.]repoName
```
bbonenfant, line 85: I believe this comment is now out of date

src/server/pfs/s3/driver.go
```diff
@@ -115,64 +89,31 @@ func bucketNameToCommit(bucketName string) *pfs.Commit {
...
-		return nil, errors.Errorf("bad bucket name %s", bucketName)
+		// e.g. s3://myrepo
+		repo, branch = client.NewProjectRepo(pfs.DefaultProjectName, parts[0]), "master"
 	case 2:
-		repo = client.NewProjectRepo(parts[1], parts[0])
+		// e.g. s3://mybranch.myrepo
+		repo, branch = client.NewProjectRepo(pfs.DefaultProjectName, parts[1]), parts[0]
 	case 3:
-		repo = client.NewSystemProjectRepo(parts[2], parts[1], parts[0])
-	default:
```
bbonenfant, line 156: Should the default case have been removed?

bbonenfant (approved):
LGTM! It appears that this doesn't include the "user nudging" towards using fully qualified paths. That makes sense given the scope, but should a follow on ticket be created?

src/server/pfs/s3/driver.go
msteffen: Done!

src/server/pfs/s3/driver.go
msteffen: Fixed, thanks

src/server/pfs/s3/driver.go
msteffen: Nope, thanks! I fixed it and added a test (and, actually, changed the `SplitN(bucketName, 4)` call above to `SplitN(bucketName, 5)` to fix the test—bucket names could have 4 components, so splitting the name into only 4 pieces could mask a too-long-bucket-name error by parsing the name so that the project has a '.' in it. But they can't have 5)

msteffen:
@robert-uhl is on PTO until the end of the week, and since Ben has already looked through this PR, I wanted to merge it so we could start using it in testing sessions and such. Bob, if you notice anything that you want addressed, just let me know and I'll make a follow up!

msteffen:
> LGTM! It appears that this doesn't include the "user nudging" towards using fully qualified paths. That makes sense given the scope, but should a follow on ticket be created?

Just created https://github.com/pachyderm/pachyderm/pull/8516#pullrequestreview-1264226028 to track that.

PR #8517, opened 2023-01-17
Allow pipeline spec to directly specify k8s tolerations

jrockway:
People were using pod patches to allow pipelines to tolerate node taints; because, for example, they have dedicated nodes just for Pachyderm workloads.  This change allows you to directly specify the tolerations in the pipeline spec.  Something like:

```yaml
tolerations:
  - key: dedicated
    operator: EXISTS
    effect: NO_SCHEDULE
```
Would allow the workload to tolerate a node tainted like: `kubectl taint node whatever dedicated:NoSchedule`.

You can have as many tolerations as you want, including 0.

The protobuf API directly clones the k8s API, except that enums like "Exists" or "NoSchedule" in Go are "EXISTS" and "NO_SCHEDULE" in proto.  (Thought about keeping them the same, but decided not to have MixedCase enums.)

We do some preliminary validation at create/update, so completely invalid tolerations won't be accepted.

I added a test for a lot of the RC generation in worker_rc_test.go.  The cases mostly cover tolerations, but I remember adding some default resource allocations there a while back, so I added a test for that as well.  I also tested this by tainting a node, submitting a pipeline that tolerates that taint, and checked that it ran.  It works!  "pachctl edit" also works correctly.  I decided not to test this against live k8s in CI because we only have 1 node there, and we run multiple tests in parallel, so it would be pretty invasive (adding a taint would prevent other tests from scheduling, unless we had a bucket only for this test, and adjusted minikubetestenv to apply the tolerations to all pachd workloads; maybe a good idea at some point, but not today).  I think the unit test alone is adequate, though.


src/server/pps/server/worker_rc.go
```diff
@@ -839,9 +857,42 @@ func (kd *kubeDriver) getWorkerOptions(ctx context.Context, pipelineInfo *pps.Pi
 		schedulingSpec:        pipelineInfo.Details.SchedulingSpec,
 		podSpec:               pipelineInfo.Details.PodSpec,
 		podPatch:              pipelineInfo.Details.PodPatch,
+		tolerations:           tolerations,
 	}, nil
 }
 
+// transformToleration transofrms a pps.Toleration into a k8s Toleration.  It's used while creating
```
albscui, line 864:
```suggestion
// transformToleration transforms a pps.Toleration into a k8s Toleration.  It's used while creating
```

src/server/pps/server/worker_rc.go
```diff
@@ -839,9 +857,42 @@ func (kd *kubeDriver) getWorkerOptions(ctx context.Context, pipelineInfo *pps.Pi
...
+		out.Effect = v1.TaintEffectNoExecute
+	case pps.TaintEffect_NO_SCHEDULE:
+		out.Effect = v1.TaintEffectNoSchedule
+	case pps.TaintEffect_PREFER_NO_SCHEDULE:
+		out.Effect = v1.TaintEffectPreferNoSchedule
+	}
+	switch in.GetOperator() { //exhaustive:enforce
+	case pps.TolerationOperator_EMPTY:
+		return out, errors.New("cannot omit key/value comparison operator; specify EQUAL or EXISTS")
+	case pps.TolerationOperator_EXISTS:
```
albscui, line 888: Based on the docs, it says `If the operator is Exists, the value should be empty, otherwise just a regular string.` Should we enforce that here by setting `out.Value = ""` just to be safe?

jrockway: I'd rather let k8s handle that logic.  It's something that could change, and we shouldn't be enforcing it.  In general, where the API allows users to represent garbage, we should pass it through.  (A possible case is an admission controller that uses that field for some undocumented use; might as well not break that even though it's a terrible design on their part.)

PR #8518, opened 2023-01-18
[CORE-1233] Add kubernetes event logs to pachctl

armaanv:
This PR adds a new pachctl command `pachctl kube-events` which returns the logs from the kube-event-tail pod that is deployed along with pachyderm


armaanv:
I updated this to go through Loki. The RPC now takes in a LokiRequest message and outputs LokiLogMessages. These can be used for future additions to the API. It's also a streaming RPC now. I've also added in some formatting for the output along with a couple of flags.

src/server/pps/cmds/cmds.go
```diff
@@ -539,6 +539,42 @@ each datum.`,
...
+		Long:  "Return the kubernetes events.",
+		Run: cmdutil.RunFixedArgs(0, func(args []string) error {
+			client, err := pachdclient.NewOnUserMachine("user")
+			if err != nil {
+				return err
+			}
+			defer client.Close()
+			since, err := time.ParseDuration(since)
+			if err != nil {
+				return errors.Wrapf(err, "error parsing since(%q)", since)
```
jrockway, line 555: Errors need not contain the word "error".  `parse since(%q)` is fine.

jrockway (approved):
Looks good.  Does the tablewriter thing treat \t specially?  I feel like there isn't enough space in 8 characters to print any of that stuff.

src/server/pps/cmds/cmds.go
armaanv: will change this

armaanv:
> Looks good. Does the tablewriter thing treat \t specially? I feel like there isn't enough space in 8 characters to print any of that stuff.

Yeah, it uses tabs to delimit the columns in the output

PR #8520, opened 2023-01-18
Update PutFileURL Task Model to Utilize PathRanges and Offsets 

FahadBSyed:
## Description
This PR adds a `startPath`, `endPath`, `startOffset`, and `endOffset` to `putFileURL` tasks, allowing us to take advantage of the go-cdk library's `RangeReader()` which can read segments of a file. Reading a byte range of a file enables us to split the work up when importing objects with a prefix from object storage. We can divide the work into batches based on size and process them in parallel.


src/server/pfs/server/url_test.go
```diff
@@ -0,0 +1,115 @@
...
+				EndPath:     endPath,
+				StartOffset: startOffset,
+				EndOffset:   endOffset,
+			}
+			tasks = append(tasks, task)
+			return nil
+		}))
+	processedFiles := processTasks(ctx, t, tasks, dir, bucket, objStoreDir)
+	for file, data := range files {
+		require.Equal(t, data, processedFiles[file], "files should match")
```
FahadBSyed, line 67: The files being compared here are between 0 and 10 bytes, so use of `require.Equal()` is probably fine.

src/server/pfs/server/pfsserver.proto
```diff
@@ -53,7 +53,11 @@ message PutFileURLTask {
   string dst = 1;
   string datum = 2;
   string URL = 3;
-  repeated string paths = 4;
+  reserved 4;
```
brycemcanally, line 56: This isn't needed, task service protos can be changed arbitrarily across versions.

src/server/pfs/server/pfsserver.proto
```diff
@@ -53,7 +53,11 @@ message PutFileURLTask {
...
+  string start_path = 5;
```
brycemcanally, line 57: nit, it seems like it would read better if the fields were ordered as: start path, start offset, end path, end offset.

src/server/pfs/server/url_worker.go
```diff
@@ -144,6 +176,7 @@ func (d *driver) processGetFileURLTask(ctx context.Context, task *GetFileURLTask
 }
 
 func serializePutFileURLTask(task *PutFileURLTask) (*types.Any, error) {
+	fmt.Printf("task: %+v\n", task)
```
brycemcanally, line 179: remove

src/internal/task/util.go
```diff
@@ -3,16 +3,46 @@ package task
...
+			if !ok {
+				return taskChain.Wait()
+			}
+			if err := taskChain.CreateTask(func(context.Context) (serCB func() error, err error) {
+				result, err := DoOne(ctx, doer, input)
+				if err != nil {
+					return nil, errors.EnsureStack(err)
+				}
+				return func() error {
+					out <- result
```
brycemcanally, line 34: We need a select with a context done check here.

src/internal/task/util.go
```diff
@@ -3,16 +3,46 @@ package task
...
+	taskapi "github.com/pachyderm/pachyderm/v2/src/task"
+
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/log"
-	"golang.org/x/sync/errgroup"
+	"github.com/pachyderm/pachyderm/v2/src/internal/taskchain"
 )
 
+// DoOrdered returns objects of type Any in order on channel out.
+func DoOrdered(ctx context.Context, doer Doer, inputs, out chan *types.Any, parallelism int) error {
```
brycemcanally, line 20: nit, out -> outputs

src/internal/task/util.go
```diff
@@ -3,16 +3,46 @@ package task
...
+	taskChain := taskchain.NewTaskChain(ctx, semaphore.NewWeighted(int64(parallelism)))
```
brycemcanally, line 21: If we are going to use taskchain as the package name, then we would probably be better off with the constructor just being `New`.

src/server/pfs/server/url.go
```diff
@@ -84,14 +88,45 @@ func putFileURL(ctx context.Context, taskService task.Service, uw *fileset.Unord
...
+			return err
+		}
+		close(inputChan)
+		return nil
+	})
+	// Order output from input channel.
+	eg.Go(func() error {
+		// TODO: Add cache?
+		defer close(outputChan)
+		return task.DoOrdered(ctx, doer, inputChan, outputChan, 100)
```
brycemcanally, line 125: `DoOrdered` could just return the output through a callback. That is the underlying model of the task chain, which would make the outputChan and additional goroutine unnecessary.

src/server/pfs/server/url.go
```diff
@@ -84,14 +88,45 @@ func putFileURL(ctx context.Context, taskService task.Service, uw *fileset.Unord
...
 	eg.Go(func() error {
-		// TODO: Add cache?
-		doer := taskService.NewDoer(URLTaskNamespace, uuid.NewWithoutDashes(), nil)
-		err := doer.Do(ctx, inputChan, func(_ int64, output *types.Any, err error) error {
+		if err := coordinateTasks(ctx, src.URL, func(startOffset, endOffset int64, startPath, endPath string) error {
+			input, err := serializePutFileURLTask(&PutFileURLTask{
+				Dst:         dst,
+				Datum:       tag,
+				URL:         src.URL,
+				StartOffset: startOffset,
```
brycemcanally, line 101: nit, I think we should generally order these fields / parameters across this code as start path, start offset, end path, end offset.

src/server/pfs/server/url.go
```diff
@@ -100,70 +135,82 @@ func putFileURLRecursive(ctx context.Context, taskService task.Service, uw *file
...
+			retErr = multierror.Append(retErr, errors.EnsureStack(err))
 		}
-		bucket, err := openBucket(ctx, url)
-		if err != nil {
-			return err
+	}()
+	startPath, endPath := "", ""
+	list := bucket.List(&blob.ListOptions{Prefix: url.Object})
+	listObj, err := list.Next(ctx)
+	if err != nil {
```
brycemcanally, line 166: This doesn't seem right. A real error isn't returned and an EOF seems like shouldn't be an error.

src/server/pfs/server/url.go
```diff
@@ -26,6 +24,10 @@ const (
 	defaultURLTaskSize = 1000
 )
 
+var (
+	threshold = int64(1000000) // threshold is overridden when testing. Default is 1 Mb
```
brycemcanally, line 28: 1MB is too low for a task, the overhead of managing the tasks will outweigh the work being done. Also, a file set for every MB would be excessive in terms of compaction. I think between 100MB - 1GB would be the right range for this, so that the task is on the order of seconds. I would lean towards 1GB to give it parity with the sharding we do in compaction.

src/server/pfs/server/url.go
```diff
@@ -100,70 +129,82 @@ func putFileURLRecursive(ctx context.Context, taskService task.Service, uw *file
...
-			paths = append(paths, listObj.Key)
+			endOffset := pos + threshold
+			if err := cb(startPath, pos, endPath, endOffset); err != nil {
+				return errors.EnsureStack(err)
+			}
+			pos = endOffset
+		}
+		return nil
+	}
+
```
brycemcanally, line 177: nit, arbitrary space.

src/internal/task/util.go
```diff
@@ -3,16 +3,45 @@ package task
...
+// DoOrdered returns objects of type Any in order via the provided callback cb.
+func DoOrdered(ctx context.Context, doer Doer, inputs chan *types.Any, parallelism int, cb CollectFunc) error {
+	taskChain := taskchain.New(ctx, semaphore.NewWeighted(int64(parallelism)))
+	for {
+		select {
+		case input, ok := <-inputs:
+			if !ok {
+				return taskChain.Wait()
+			}
+			if err := taskChain.CreateTask(func(context.Context) (serCB func() error, err error) {
```
brycemcanally, line 28: The named returns are not needed.

src/internal/task/util.go
```diff
@@ -3,16 +3,45 @@ package task
...
+
+	taskapi "github.com/pachyderm/pachyderm/v2/src/task"
+
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/log"
-	"golang.org/x/sync/errgroup"
+	"github.com/pachyderm/pachyderm/v2/src/internal/taskchain"
 )
 
+// DoOrdered processes tasks in parallel, but returns objects in order via the provided callback cb.
```
brycemcanally, line 19: nit, objects -> outputs

src/server/pfs/server/url.go
```diff
@@ -22,8 +21,10 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
 )
 
-const (
-	defaultURLTaskSize = 1000
+var (
+	// these are overridden when testing.
+	defaultNumObjectsThreshold = 1000
+	defaultSizeThreshold       = int64(units.GB * 1)
```
brycemcanally, line 27: This can just be `units.GB`

src/internal/task/util.go
brycemcanally: We should probably also have a note about the index in the collect function not being supported.

src/server/pfs/server/url.go
```diff
@@ -135,37 +109,70 @@ func putFileURLRecursive(ctx context.Context, taskService task.Service, uw *file
...
+		size += listObj.Size
+		if len(paths) >= defaultNumObjectsThreshold || size >= defaultSizeThreshold {
+			if err := cb(paths, 0, -1); err != nil {
+				return errors.EnsureStack(err)
+			}
+			paths = nil
+			size = 0
+		}
+	}
+	if len(paths) != 0 {
```
brycemcanally, line 170: nit, > is a bit more clear than != here.

brycemcanally (approved):
lgtm

PR #8524, opened 2023-01-19
Improve TestUpgradeSimple by leveraging OpenCV example

albscui:
Previously, we found a bug (https://github.com/pachyderm/pachyderm/pull/8513) in the migration code that wasn't caught by our upgrade tests. It was found manually by running OpenCV example locally. 

This PR updates the test to leverage OpenCV so that we can have better coverage.


djanicekpach (approved):
Looks good to me, thanks fro doing this!

PR #8528, opened 2023-01-20
[PFS-46] Fix directory creation with path range requests

brycemcanally:
This PR fixes directory creation when the first file in the directory is the upper bound in a path range request. The source abstraction is responsible for emitting PFS directories / files and supports path range requests. The issue is that a path range is an inclusive, exclusive range, so a directory with its first file as the upper bound would not get generated since the file would not be emitted by the internal index iterators. To ensure that the correct directories are emitted in this case, the path range that is passed to the indexing layer is tweaked to iterate past the upper bound. The source iterator will exit early when the upper bound is reached or passed. This ensures that the correct directories will be emitted within the path range provided to the source. This PR also includes some additional testing for this.

### DIFF
--- a/src/server/pfs/server/source.go
+++ b/src/server/pfs/server/source.go
@@ -58,10 +58,14 @@ func NewSource(commitInfo *pfs.CommitInfo, fs fileset.FileSet, opts ...SourceOpt
 		// and the first file that shouldn't get created.
 		// For example, the files /d1/f1 and /d2/f2 with a path range of [/d1/f1, /d2/f2) should
 		// emit /d1/f1 and /d2/.
-		s.fileIndexOpts = append(s.fileIndexOpts, index.WithRange(&index.PathRange{
+		pr := &index.PathRange{
 			Lower: sc.pathRange.Lower,
-			Upper: sc.pathRange.Upper + string(0),
-		}))
+			Upper: sc.pathRange.Upper,
+		}
+		if pr.Upper != "" {
+			pr.Upper += string(0)
+		}
+		s.fileIndexOpts = append(s.fileIndexOpts, index.WithRange(pr))
 		s.upper = sc.pathRange.Upper
 	}
 	if sc.filter != nil {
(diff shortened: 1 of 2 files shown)


src/server/pfs/server/source.go
```diff
@@ -51,10 +53,16 @@ func NewSource(commitInfo *pfs.CommitInfo, fs fileset.FileSet, opts ...SourceOpt
...
 		}))
+		// The upper for the index path range is set to be past the provided upper
+		// to ensure that directories between the last file that should be emitted
+		// and the first file that shouldn't get created.
+		// For example, the files /d1/f1 and /d2/f2 with a path range of [/d1/f1, /d2/f2) should
+		// emit /d1/f1 and /d2/.
 		s.fileIndexOpts = append(s.fileIndexOpts, index.WithRange(&index.PathRange{
 			Lower: sc.pathRange.Lower,
-			Upper: sc.pathRange.Upper,
+			Upper: sc.pathRange.Upper + "_",
```
robert-uhl, line 63: Is there any value to using `"\0"` here?

robert-uhl (approved):

src/server/pfs/server/source.go
brendoncarroll, line 63: If we don't explode on invalid paths, I would prefer to add a null byte here.  I prefer the guarantee that there is no path between `sc.pathRange.Upper` and the index.PathRange upper bound. I believe the guarantee here is that there is no currently valid path.  If we change the allowed character set, this could break.

brendoncarroll (approved):
Small suggestion, up to you. Otherwise good to merge.

src/server/pfs/server/source.go
brycemcanally: It doesn't affect the results, but it is probably clearer what is being done if the null byte is used.

brycemcanally: It wouldn't break with an allowable character set change because the upper bound is still used in the source iteration, but using the null byte is probably clearer anyways.

PR #8529, opened 2023-01-20
Don't bail if the cluster deployment ID of the cluster doesn't match

chainlink:
Completely [ripping out](https://github.com/pachyderm/pachyderm/pull/8527/files) cluster deployment ID is hard and potentially contentious. What we can do instead is just remove the check which errors out the client if the cluster deployment ID's don't match.


jrockway (approved):
This will be nice.  I think we should delete the entire block starting at line 642, though we can't remove the deployment id field entirely (because it will break parsing of existing configs, which isn't worth it).

chainlink:
I think we still need to save it for now for some enterprise context stuff.

PR #8533, opened 2023-01-23
helm: proxy: allow values.yaml to declare the proxy service external traffic policy

jrockway:
We have to default to `""` because this field cannot exist on non-NodePort/LoadBalancer services.

This is tracked in CORE-1473.


BOsterbuhr (approved):
LGTM

PR #8534, opened 2023-01-23
catch failed login attempts

djanicekpach:
[CORE-1422 ](https://pachyderm.atlassian.net/browse/CORE-1422)

This does not fix the test flake, however it should prevent the token timeout error from swallowing login failures. This will hopefully give us some better insight into what is actually failing in the deploy tests.


acohen4 (approved):
LGTM!

PR #8535, opened 2023-01-24
CORE-961 Don't print warning about default login config when a real ID provider is configured for helm deployment

zmajeed:
fixed by setting new `oidc.hasRealIDP` field in `_helpers.tpl` - then checking it in `NOTES.txt` before printing the warning


etc/helm/pachyderm/templates/_helpers.tpl
```diff
@@ -163,6 +163,16 @@ http://localhost:30658
 {{- end }}
 {{- end }}
 
+{{- define "oidc.hasRealIDP" -}}
+{{- if .Values.oidc.upstreamIDPs -}}
+true
+{{- else if .Values.oidc.mockIDP -}}
```
acohen4, line 169: simplify to `else false`?

zmajeed: it's redundant but I wanted to call out `mockIDP` - I'll remove it if you prefer

etc/helm/pachyderm/templates/NOTES.txt
```diff
@@ -10,8 +10,10 @@ pachctl config set active-context local
 pachctl port-forward
 {{- end -}}
 
-{{ if and (.Values.oidc.mockIDP) (.Values.pachd.enterpriseLicenseKey) (.Values.pachd.activateAuth) }}
+{{ if not (include "oidc.hasRealIDP" .) }}
```
acohen4, line 13: I think this would be simpler if we don't add `oidc.hasRealIDP` and instead include `not .Values.oidc.upstreamIDPs`  here.

zmajeed: Yeah - that's simpler and makes the relationship between real and mock IDPs clear

acohen4 (approved):
LGTM!

zmajeed:
amd64 deploy tests failed twice - specifically TestUpgradeLoad/UpgradeFrom_2.3.9 - circleci dashboard shows this test failing on many branches - master, acohen4/projects-migrate-tracker, rau/core-1476-address-getlogsloki-project-todo

PR #8537, opened 2023-01-24
[PFS-47] Add support for affinity and tolerations

FahadBSyed:


etc/helm/pachyderm/templates/pachw/deployment.yaml
```diff
@@ -41,7 +41,9 @@ spec:
       {{- if .Values.pachd.priorityClassName }}
       priorityClassName: {{ .Values.pachd.priorityClassName }}
       {{- end }}
-      {{-  if .Values.pachd.affinity }}
+      {{-  if .Values.pachw.affinity }}
```
jrockway, line 44: What do we do about the case where the user wants this off for pachw but on for pachd?

FahadBSyed: That's a good question. I hadn't actually thought about that case. What would you recommend as the best way to handle that? We could capture that with an extra flag, but I figure that's a little finnicky. We could alternatively ditch using pachd as a default for tolerations and affinity.

BOsterbuhr:
LGTM, I really like the `inheritFromPachd` idea!

BOsterbuhr (approved):

PR #8539, opened 2023-01-24
Implement ProjectViewer

albscui:
> Note albert/CORE-1408/projects-auth should track the head commit of the master branch prior to 2.5.0 release. We won't be releasing ProjectViewer and ProjectWriter in 2.5.0. Therefore, we will use albert/CORE-1408/projects-auth as a semi long lived branch to merge new projects related features to. Once 2.5.0 is released, we can merge this branch into master.

> Update: now that 2.5.0 is released, we can just merge this into master.

Updated `ListRepo` and `ListPipeline` APIs to check for the requried permissions. Both APIs have similar semantics:
- check for PROJECT_LIST_REPO permission on the repo/pipeline's project before returning it to the user
  - if that fails, then check for REPO_READ access on the individual repo or pipeline's output repo.

Since permissions at the project level applies to all repos within the project, we cache the result of each project-level permission check. We use a local in-memory LRU cache provided by a library from Hashicorp. 
> I had to import the v2 version of Hashicorp's golang-lru library, because it supports Go generics. Meanwhile we actually use v1 in other places.
> TODO standardize usage of the v2 library across our code.


src/server/pfs/server/driver.go
```diff
@@ -280,7 +280,22 @@ func (d *driver) listRepo(ctx context.Context, includeAuth bool, repoType string
...
+					}
+					return errors.Wrap(err, "could not check user is authorized to access repo")
+				}
+				// Here we know that the user does not have permission to list repos at the project level,
+				// but they should still see the repo because they have read permission.
+			} else {
+				return errors.Wrap(err, "could not check user is authorized to access project")
+			}
+		}
+
```
acohen4, line 298: nit: remove newline

src/server/auth/server/roles.go
```diff
@@ -208,6 +219,8 @@ func init() {
 			licenseAdminRole.Permissions,
 			secretAdminRole.Permissions,
 			pachdLogReaderRole.Permissions,
+			projectViewerRole.Permissions,
```
acohen4, line 222: I don't think these two are necessary since they're implied by project owner

src/server/pfs/server/driver.go
```diff
@@ -280,7 +280,22 @@ func (d *driver) listRepo(ctx context.Context, includeAuth bool, repoType string
 		}
 		repoInfo.SizeBytesUpperBound = size
 
-		// TODO CORE-1111 check whether user has PROJECT_LIST_REPO on project or REPO_READ on repo.
+		// Check if the user can list repo within the project, otherwise check if they can read the repo.
+		if err := d.env.AuthServer.CheckProjectIsAuthorized(ctx, repoInfo.Repo.Project, auth.Permission_PROJECT_LIST_REPO); err != nil {
```
acohen4, line 284: I think it's worth optimizing this, because as it is we may be running the same auth queries for every repo in a project. One way to make this more performant is to first collect the user's auth info for each project and check that in processFunc, before running the individual REPO_READ query

src/server/auth/server/testing/auth_test.go
```diff
@@ -1027,10 +1031,16 @@ func TestListAndInspectRepo(t *testing.T) {
...
+			auth.Permission_PROJECT_CREATE_REPO,
+		},
+		repoNone: {
+			auth.Permission_PROJECT_LIST_REPO,
+			auth.Permission_PROJECT_CREATE_REPO,
 		},
-		repoNone: {},
 	}
 	for _, info := range repoInfos {
+		fmt.Println("qqq", info.Repo)
```
acohen4, line 1043: remove

src/server/auth/server/testing/auth_test.go
```diff
@@ -2532,3 +2542,82 @@ func TestDeleteRepos(t *testing.T) {
...
+	project1Repos := []string{project1Repo1, project1Repo2}
+	project2Repos := []string{project2Repo1, project2Repo2}
+	allRepos := append(defaultRepos, append(project1Repos, project2Repos...)...)
+
+	tests := map[string]struct {
+		user          *client.APIClient
+		filterBy      []*pfs.Project
+		expectedRepos []string
+	}{
+		"admin no filter":                            {admin, nil, allRepos},
```
acohen4, line 2599: nice test 🔥

acohen4 (changes requested):
Few comments. overall looks good!

robert-uhl (approved):
LGTM, pending @acohen4’s comments.

src/server/auth/server/roles.go
```diff
@@ -172,22 +172,33 @@ func init() {
 	})
 
 	// Project related roles
+	projectViewerRole := registerRole(&auth.Role{
+		Name:          auth.ProjectViewerRole,
+		ResourceTypes: []auth.ResourceType{auth.ResourceType_PROJECT, auth.ResourceType_REPO},
```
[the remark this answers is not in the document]
acohen4: sorry for late reply. The semantics are the other way around. Reading over the existing implementation, I think that it could be better and more clear. The `ResourceTypes` mean the Resources (auth levels) that this role can be applied to.

I thought that we were validating this at `[Create/Modify]RoleBinding` but turns out that we aren't...

As far as API surface it currently gets exposed in the semantics of `GetPermissions(resource) -> (permissions, roles)` where `roles` is all the roles the user has related to that resource that satisfied these permissions. So for example, if I have RepoReader assigned to me at the CLUSTER auth level, I should get RepoReader back in the response of `GetPermissions("repoFoo", "user:alon") -> (REPO_READ, RepoReader)`

acohen4 (changes requested):
a few more comments

src/internal/clusterstate/v2.6.0.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+var state_2_6_0 migrations.State = state_2_5_0.
+	Apply("Grant all users ProjectWriter role for the default project", func(ctx context.Context, env migrations.Env) error {
+		roleBindingsCol := authdb.RoleBindingCollection(nil, nil).ReadWrite(env.Tx)
+		if !authIsActive(roleBindingsCol) {
+			return nil
+		}
+		rb := &auth.RoleBinding{}
+		if err := roleBindingsCol.Upsert("PROJECT:default", rb, func() error {
```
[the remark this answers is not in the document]
albscui: As part of v2.5.0, we already created an empty role binding for the default project here https://github.com/pachyderm/pachyderm/blob/master/src/internal/clusterstate/v2.5.0/auth.go#L62. Maybe I should've used "Create" back then, but now "Upsert" makes more sense right?

src/server/pfs/server/driver.go
```diff
@@ -264,23 +265,52 @@ func (d *driver) getPermissions(ctx context.Context, repo *pfs.Repo) ([]auth.Per
...
+			// Allow access if user has the right permissions at the individual Repo-level.
+			return d.env.AuthServer.CheckRepoIsAuthorized(ctx, repo, auth.Permission_REPO_READ)
+		}
+		return nil
 	}
 
 	processFunc := func(string) error {
-		// Assume the user meant all projects by not providing any projects to filter on.
-		if len(projectsFilter) > 0 && !projectsFilter[repoInfo.Repo.Project.String()] {
+		if !keep(repoInfo.Repo) {
```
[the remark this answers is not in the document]
albscui: I wanted to put the exclude case on top so that we don't wrap majority of the actual logic in an if-block.

In terms of semantics, here `keep()` returns true if the repo's project satisfies the user's text-based filter via `--project`, otherwise we can return early to avoid unnecessary auth calls since we know the user doesn't want to see this repo at all. Similarly, `checkAccess()` is returns an error if the user doesn't have access to this repo. Returning `nil` is our way of skipping this repo since it won't be returned to the client.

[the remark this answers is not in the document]
acohen4: ahh totally misread the keep() semantics 👍

acohen4 (approved):
LGTM!!

PR #8540, opened 2023-01-25
[CORE-1462] re-introduce explicit audit log

armaanv:
This PR adds an explicit log line for pachctl commands for audit purposes. This was previously logged as 'user command' before the logging refactor

### DIFF
--- a/src/internal/middleware/logging/interceptor.go
+++ b/src/internal/middleware/logging/interceptor.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"io"
 	"reflect"
+	"strings"
 	"time"
 
 	"go.uber.org/zap"
@@ -168,6 +169,12 @@ func (li *LoggingInterceptor) UnaryServerInterceptor(ctx context.Context, req in
 
 	service, method := parseMethod(info.FullMethod)
 	ctx = getCommonLogger(ctx, service, method)
+
+	md, ok := metadata.FromIncomingContext(ctx)
+	if ok && len(md.Get("command")) > 0 {
+		command := md.Get("command")
+		dolog(ctx, log.InfoLevel, "audit log: %s"+strings.Join(command, ""))
+	}
 
 	// NOTE(jonathan): We use service/method in the log messages so that rate limiting applies
 	// per-RPC instead of for all RPCs.  (Rate limiting algorithm looks at the message and the
--- a/src/internal/middleware/logging/interceptor.go
+++ b/src/internal/middleware/logging/interceptor.go
@@ -98,7 +98,11 @@ func getCommonLogger(ctx context.Context, service, method string) context.Contex
 		}
 		if command := md.Get("command"); command != nil {
 			f = append(f, zap.Strings("command", command))
-			log.Info(ctx, "audit log: ", f...)
+			if method != "InspectCluster" {
+				// InspectCluster is called by the client to get the deployment ID, so
+				// we don't want to log that.
+				log.Info(ctx, "audit log", f...)
+			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/internal/middleware/logging/interceptor.go
```diff
@@ -437,6 +438,12 @@ func (li *LoggingInterceptor) UnaryServerInterceptor(ctx context.Context, req in
 	service, method := parseMethod(info.FullMethod)
 	ctx = getCommonLogger(ctx, service, method)
 
+	md, ok := metadata.FromIncomingContext(ctx)
```
jrockway, line 441: This only handles unary requests; most of our requests are streaming, so you'll need to log there as well.   Do keep in mind that we already log the command with every message logged during a request (line 368).  (The common logger is used for every line the RPC handler logs, including the request and response.  The request logger logs the request; the response logger logs the response.)  If there needs to be a log line that contains "audit log" with the text of the command, I'd do it in getCommonLogger.  Using the context passed to that function, extract the request id (to tie the audit line to the rest of the request), username, and command.  then `log.Info(ctx, "pachctl command: " + command, zap.String("x-request-id", requestID), zap.String("user", user))`.  It all seems a little redundant to me; it's an extra log line with no additional information.

jrockway (commented):
The only blocker here is adding logging to the streaming calls, but I do think we should take a step back and understand why the existing logging is inadequate.  It includes the command with every message logged during the RPC, not just the request ;)

src/internal/middleware/logging/interceptor.go
armaanv: The audit log line is to provide the users a better way to list all the pachctl commands that have been run against their cluster. At the moment it's difficult to obtain this list from the logs because there are so many log lines printed for just one command. We could parse out the command field from the other log lines and then group them into unique pachctl commands for an audit log but that is a lot more computation. Eventually, I think we should remove this line from the logs and add it as an event log. 

Moving this to getCommonLogger makes sense to me

jrockway: Fair enough.  Yeah, basically treat it as an unrelated log in getCommonLogger, I think that will work fine.

src/internal/middleware/logging/interceptor.go
```diff
@@ -98,6 +98,7 @@ func getCommonLogger(ctx context.Context, service, method string) context.Contex
 		}
 		if command := md.Get("command"); command != nil {
 			f = append(f, zap.Strings("command", command))
+			log.Info(ctx, "audit log: ", f...)
```
jrockway, line 101: That makes sense.  I'd omit the trailing ": ".  That marshals as something like `{"message":"audit log: "}` which looks kind of weird.

I also think that the user and request id are interesting items to add here; the command on its own doesn't seem that useful unless it's related somehow to the rest of the data stream.  (Remember that a busy pachd instance can be executing hundreds of RPCs in parallel, so the position of the log line won't necessarily make sense under those circumstances.)

jrockway (approved):
One other thing; pachctl makes at least 2 RPCs; InspectCluster when the client is created, and then whatever RPCs it wanted to make.  That means, you see this log line 2x per pachctl command.  Maybe do a check that the method is not InspectCluster (please test in case I'm forgetting the name), and only log this for the "real" command it's running?

src/internal/middleware/logging/interceptor.go
armaanv: will remove the colon. user and request id should be captured in f already.

jrockway: Oh of course!

PR #8541, opened 2023-01-25
Add codegen to generate MarshalLogObject for protos

jrockway:
I'd like to avoid reflection for logging gRPC requests and responses, so I added a code generator that creates MarshalLogObject for each proto.  They end up in XXX.pb.zap.go for each XXX.proto.  This PR adds the generator in etc/proto/protoc-gen-zap, and it is setup to run during "make proto", so no workflow needs to change.  I also updated log.Proto to take advantage of this code.  (The generated code also handles embedded well-known types; google.protobuf.* cleanly... log.Proto does the same transformations now.)

This a fork of some upstream code; I made it aware of gogo and well-known types.  The codegen is using the newest proto compiler author's API, so we aren't tied to gogo going forward (hence; a copy of gogo.proto generated with the new proto compiler is also embedded here).  

The change to handle well-known types is so that Timestamps, Durations, etc. are marshaled as native objects, so you see `"2023-01-24T08:51:00Z"` instead of `{seconds: 1674611511, nanos: 123}`.  BytesValues are treated specially; we emit the first 32 bytes as base64 and the length if there are more than 32 bytes.  This is important for gRPC logging because we frequently return large objects as bytes; this gives you the important part and saves dumping 20MB to the logs ;)  This actually works really well.  Here's a log line for GetFile:

```json
{"severity":"info","time":"2023-01-25T01:47:12.217600153Z","logger":"grpc.pfs_v2.API/GetFile","caller":"logging/interceptor.go:261","message":"first message sent for pfs_v2.API/GetFile","service":"pfs_v2.API","method":"GetFile","user":"pach:root","x-request-id":["2d9e420d-4660-4144-8514-589aff98b1fe"],"command":["pachctl get file images@master:/46Q8nDz.jpg"],"peer":"10.244.0.8:52212","response":{"len":58644,"firstBytes":"/9j/4AAQSkZJRgABAQAAAQABAAD/7gAOQWRvYmUAZMA="},"grpc.code":0}
```  
You can then dump this into `file` to figure out what the file is:
```
$ base64 -d  | file -
/9j/4AAQSkZJRgABAQAAAQABAAD/7gAOQWRvYmUAZMA=
/dev/stdin: JPEG image data, JFIF standard 1.01, aspect ratio, density 1x1, segment length 16
```
(The file has the extension .png though!!!)

I also added the ability to `mask` or `half` fields during log marshaling by applying an extension in the proto definition.  A field annotated with `[(log.mask) = true]` marshals as `[MASKED]` (even if empty/default; for now).  half replaces the value with the first half of the string, `.../`, and the total length in bytes (algorithm copied from oidc code that did this manually for various fields).  No special effort is required to get the masking, once the code is generated, any time you `log.Proto` or `zap.Object` any instance of the message, the redaction is applied.  Of course, it also applies to embedded messages; if you embed a message with a masked field in a new message, it's still masked when you log the container.  I've therefore removed the custom masking code from the GRPC interceptor.  I also changed a few deletions to `half`, mostly the enterprise license key.  The name and expiry happen to end up there, but not a valid signature.  Now looking at the logs, you can see what's going on with heartbeats and whatnot, while still not letting anyone reading the logs go setup their own Pachyderm enterprise instance for free.  Perfect.

I added some benchmarks for this code.  It's about 25% slower than reflection (which yields ugly messages like `"&Int32Value{Value:42,XXX_unrecognized:[],}"`),  200% slower than `proto.Marshal` (which is unreadable by humans), 2500% faster than jsonpb.Marshal (!), and 300% faster than `proto.String()` which is what we currently use.  I'm unhappy that it's slower than reflection, but it is nicer looking, so that's nice.  It should really be as fast as proto.Marshal though.  Future improvement, I guess.

One last quirk; our logs are usually JSON, but the protos fields are actually marshaled with their "native" names and not their JSON names.  This is a quirk of the upstream code that I didn't want to change.  Because of the transformations we do on well-known types, you wouldn't necessarily be able to copy a CreatePipelineRequest out of the logs and `echo <that> | pachctl create pipeline -f -` anyway, but I have a feeling that people will think this is weird and annoying.  It is straightforward, though tedious, to change.


etc/proto/protoc-gen-zap/gogoproto/gogo.pb.go
```diff
@@ -0,0 +1,1298 @@
+// Protocol Buffers for Go with Gadgets
```
brendoncarroll, line 1: Does this grow our dependency on gogo?  I thought we were trying to reduce that.  Otherwise it looks like the new stuff depends on the official go protobuf packages, which is good.

etc/proto/protoc-gen-zap/main.go
```diff
@@ -0,0 +1,302 @@
...
+
+	for _, m := range file.Messages {
+		generateMessage(g, m)
+	}
+
+	return g
+}
+
+func main() {
+	protogen.Options{}.Run(func(plugin *protogen.Plugin) error {
```
brendoncarroll, line 292: This is a neat way to do the entrypoint.

etc/proto/protoc-gen-zap/main.go
```diff
@@ -0,0 +1,302 @@
...
+	return strings.HasPrefix(fullName, "datum.") ||
+		strings.HasPrefix(fullName, "pfsload.") ||
+		strings.HasPrefix(fullName, "taskapi.")
+}
+
+func generateMessage(g *protogen.GeneratedFile, m *protogen.Message) {
+	ident := g.QualifiedGoIdent(m.GoIdent)
+	g.P("func (x *", ident, ") MarshalLogObject(enc ", g.QualifiedGoIdent(zapcorePkg.Ident("ObjectEncoder")), ") error {")
+	g.P("if x == nil {")
+	g.P("return nil")
```
brendoncarroll, line 234: This must go through gofmt after?

etc/proto/protoc-gen-zap/gogoproto/gogo.pb.go
jrockway: Yeah, this is the gogo proto generated with the new protobuf compiler/library.  Necessary to read the annotations from the new protobuf api (they get registered in a global data structure at import time and standard/gogo use different places :/).  When we stop using gogo, we just delete this file, remove the references in main.go to gogoproto.E_Customname and friends, and it's gone.

etc/proto/protoc-gen-zap/main.go
jrockway: yup :)

brendoncarroll (approved):
This looks reasonable to me.  The log.mask stuff looks like it's in all the right places, the codegen covered all the protos that I'm aware of, including internal ones.

Not sure how this fits into our long term protobuf strategy, but you probably have a better handle on that than me.

jrockway:
This doesn't change our long-term strategy to move off of gogo.  When we do that, we simply remove the support for customname, move our libraries from gogo's well-known types to google's, and all this continues to work.  When I started writing this I had a ton of trouble getting the gogo extensions to be read and wanted to do the gogo removal first... but I really want these improved log messages in 2.5, but thought it would be awfully disruptive to change the proto generation right before release.  So it ended up being more net work, but with better timing.

etc/proto/protoc-gen-zap/main.go
jrockway: Copied and pasted from the godoc 😂

PR #8542, opened 2023-01-25
Fix logger generation of pfsserver.*

jrockway:
I missed this because it only showed up for the array case, which didn't print a warning.  Now all 3 cases (array, map, primitive) print warnings, and pfsserver.* is treated as Object instead of possibly falling back to reflection.

@FahadBSyed I'll rebase and merge after your PutFileURL tasks PR.


brendoncarroll (approved):
LGTM

PR #8544, opened 2023-01-25
Modify GetFileURL Coordination Algorithm to Batch Files into Tasks by Size

FahadBSyed:


src/server/pfs/server/url_worker.go
```diff
@@ -176,7 +175,6 @@ func (d *driver) processGetFileURLTask(ctx context.Context, task *GetFileURLTask
 }
 
 func serializePutFileURLTask(task *PutFileURLTask) (*types.Any, error) {
-	fmt.Printf("task: %+v\n", task)
```
brendoncarroll, line 179: Not sure if you forgot this while debugging.  Remove or, if you want a log line, use our internal `log` package.

brendoncarroll:
I'm not sure if we have logging in the task.Doer.  If we don't can you add some logging using our internal log package.
- every time we submit a task on the `inputChan`.  Total paths and total bytes would be good to include.
- Once all the tasks have been submitted and we are just waiting for the results.

src/server/pfs/server/url_worker.go
FahadBSyed: Yeah I think this is part of PFS-25, so I will remove that from there and this diff should vanish.

src/server/pfs/server/url.go
```diff
@@ -186,7 +186,7 @@ func (d *driver) getFileURL(ctx context.Context, taskService task.Service, URL s
 		err := doer.Do(ctx, inputChan, func(_ int64, output *types.Any, err error) error { return err })
 		return errors.EnsureStack(err)
 	})
-	var bytesWritten int64
+	var size int64
```
brycemcanally, line 189: This isn't correct. Bytes written is separate from the sharding criteria. We should revert this back to bytes written and collect it independent of the size variable. Also, we should rename count to numObjects for consistency.

src/server/pfs/server/testing/server_test.go
```diff
@@ -5385,21 +5385,23 @@ func TestPFS(suite *testing.T) {
 		for _, path := range paths {
 			require.NoError(t, env.PachClient.PutFile(commit, path, strings.NewReader(path)))
 		}
+		objC := dockertestenv.NewTestObjClient(ctx, t)
+		bucketURL := objC.BucketURL().String()
 		check := func() {
-			objC := dockertestenv.NewTestObjClient(ctx, t)
```
brycemcanally, line 5389: Why was this change needed?

FahadBSyed: I wanted to add a case to test importing a directory and not individual files. My goal was to structure the code to reduce the amount of redunant lines, so I moved some of the steps out of check(). This way I could reset after testing individual file imports and test importing a directory.

src/server/pfs/server/url.go
FahadBSyed: Good catch, size is getting reset, but bytesWritten shouldn't be.

src/server/pfs/server/url.go
```diff
@@ -211,23 +211,24 @@ func (d *driver) getFileURL(ctx context.Context, taskService task.Service, URL s
 			}
 			return nil
 		}
-		var count int64
+		var numObjects, size int64
 		if err := src.Iterate(ctx, func(fi *pfs.FileInfo, file fileset.File) error {
 			if fi.FileType != pfs.FileType_FILE {
 				return nil
 			}
+			numObjects++
```
brycemcanally, line 219: I think the update should be after the check. Path ranges are inclusive / exclusive.

src/server/pfs/server/testing/server_test.go
brycemcanally: The check function pattern in the server code is intended to ensure that the same access patterns work before and after finishing a commit. This sort of removes that since it now changes what is being checked against (files vs directory) before and after finishing the commit. I think we should just add the directory case to the check. So, it could be structured like: check files, clear object store, check directories.

brycemcanally (approved):
lgtm

PR #8555, opened 2023-01-31
[CORE-1099] Handle overlong service names

robert-uhl:
This PR addresses the issue of overlong service names by simply not creating services; they are (apparently) not actually used or documented.


src/server/pps/server/worker_rc.go
```diff
@@ -1007,41 +1006,6 @@ func (kd *kubeDriver) createWorkerSvcAndRc(ctx context.Context, pipelineInfo *pp
 			return errors.EnsureStack(err)
 		}
 	}
-	serviceAnnotations := map[string]string{
```
msteffen, line 1010: Nit: maybe also change the name of the function to just `createWorkerRc()`?

msteffen (approved):
One minor nit, but overall LGTM!

As discussed [in Slack](https://pachyderm.slack.com/archives/CDWDPKPMY/p1675207101883669?thread_ts=1675190573.055919&cid=CDWDPKPMY), this shouldn't affect the sidecar/single-job S3 gateway, or S3G-enabled pipelines at all.

robert-uhl:
Handled in #8559.

PR #8557, opened 2023-02-01
[Jupyter] Basic PPS UI

bbonenfant:
This is a very simple (first draft) of the UI for the PPS plugin for Jupyterlab. This doesn't include any backend calls. This my first PR for front end work so I'm sure a lot of it is incorrect, please tear it apart!

Question for @smalyala : I'm not sure that we want this merged into master yet (I don't know how the plugin releases are currently being done). Do you have thoughts about where this line of work should live?

Jira: [INT-665](https://pachyderm.atlassian.net/browse/INT-665)

[INT-665]: https://pachyderm.atlassian.net/browse/INT-665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
+++ b/jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
@@ -3,7 +3,6 @@ import {closeIcon} from '@jupyterlab/ui-components';
 import {usePipeline} from './hooks/usePipeline';
 
 type PipelineProps = {
-  showPipeline: boolean;
   setShowPipeline: (shouldShow: boolean) => void;
 };
 
@@ -12,9 +11,9 @@ const placeholderInputSpec = `pfs:
   branch: dev
   glob: /*
 `;
-const placeholderRequirements = './requirements.txt"';
+const placeholderRequirements = './requirements.txt';
 
-const Pipeline: React.FC<PipelineProps> = ({showPipeline, setShowPipeline}) => {
+const Pipeline: React.FC<PipelineProps> = ({setShowPipeline}) => {
   const {
     loading,
     pipelineName,
@@ -27,7 +26,7 @@ const Pipeline: React.FC<PipelineProps> = ({showPipeline, setShowPipeline}) => {
     setRequirements,
     callCreatePipeline,
     errorMessage,
-  } = usePipeline(showPipeline);
+  } = usePipeline();
 
   return (
     <div className="pachyderm-mount-pipeline-base">
@@ -159,6 +158,8 @@ const Pipeline: React.FC<PipelineProps> = ({showPipeline, setShowPipeline}) => {
           data-testid="Pipeline__specPreview"
           name="specPreview"
           value={`name: ${pipelineName}
+transform:
+  image: ${imageName}
 input:
 ${inputSpec
   .split('\n')
--- a/jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
+++ b/jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
@@ -12,18 +12,19 @@ const placeholderInputSpec = `pfs:
   branch: dev
   glob: /*
 `;
-const placeholderRequirements = 'dependency=="1.2.3"';
-const placeholderSecrets = 'MyPassword: ********';
+const placeholderRequirements = './requirements.txt"';
 
 const Pipeline: React.FC<PipelineProps> = ({showPipeline, setShowPipeline}) => {
   const {
     loading,
+    pipelineName,
+    setPipelineName,
+    imageName,
+    setImageName,
     inputSpec,
     setInputSpec,
     requirements,
     setRequirements,
-    secrets,
-    setSecrets,
     callCreatePipeline,
     errorMessage,
   } = usePipeline(showPipeline);
@@ -32,7 +33,7 @@ const Pipeline: React.FC<PipelineProps> = ({showPipeline, setShowPipeline}) => {
     <div className="pachyderm-mount-pipeline-base">
       <div className="pachyderm-mount-pipeline-back">
         <button
-          data-testid="Datum__back"
+          data-testid="Pipeline__back"
           className="pachyderm-button-link"
           onClick={async () => {
             setShowPipeline(false);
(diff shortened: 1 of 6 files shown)


smalyala:
Let's start a feature branch for the pipeline extension, which we merge into master once the pipeline extension is ready.

jupyter-extension/src/plugins/mount/mount.tsx
```diff
@@ -394,6 +444,7 @@ export class MountPlugin implements IMountPlugin {
       this._mountBrowser.setHidden(false);
     }
     this._datum.setHidden(true);
+    this._pipeline.setHidden(true);
```
smalyala, line 447: Add this to `setShowFullPageError` below as well

jupyter-extension/src/plugins/mount/mount.tsx
```diff
@@ -203,7 +217,22 @@ export class MountPlugin implements IMountPlugin {
     );
     this._datum.addClass('pachyderm-mount-datum-wrapper');
 
+    this._pipeline = ReactWidget.create(
```
smalyala, line 220: We touched upon this during the FE walkthrough with @pleeko in Miami, but we probably don't need to use signals here since the pipeline component doesn't have any properties/data that changes like the other components does. Signals are useful in the case where we see that some data has changed so we re-render the components that depend on that data. But since the pipeline component doesn't take any input that could change, we can remove signals.

Now, how this would look with no signals, I defer to Peter.

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
```diff
@@ -0,0 +1,70 @@
...
+  setImageName: (input: string) => void;
+  inputSpec: string;
+  setInputSpec: (input: string) => void;
+  requirements: string;
+  setRequirements: (input: string) => void;
+  callCreatePipeline: () => Promise<void>;
+  errorMessage: string;
+};
+
+export const usePipeline = (showPipeline: boolean): usePipelineResponse => {
```
smalyala, line 19: Probably don't need `showPipeline` here

jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
```diff
@@ -0,0 +1,175 @@
...
+  showPipeline: boolean;
+  setShowPipeline: (shouldShow: boolean) => void;
+};
+
+const placeholderInputSpec = `pfs:
+  repo: images
+  branch: dev
+  glob: /*
+`;
+const placeholderRequirements = './requirements.txt"';
```
smalyala, line 15: There's a trailing double quotes here

jupyter-extension/src/plugins/mount/mount.tsx
```diff
@@ -377,6 +409,24 @@ export class MountPlugin implements IMountPlugin {
...
+    if (shouldShow) {
+      this._pipeline.setHidden(false);
+      this._mountedList.setHidden(true);
+      this._unmountedList.setHidden(true);
+    } else {
+      this._pipeline.setHidden(true);
+      this._mountedList.setHidden(false);
+      this._unmountedList.setHidden(false);
+    }
+    this._mountBrowser.setHidden(false);
```
smalyala, line 422: We don't need the file browser in the bottom of the extension when we're in pipeline mode right? If not, we can hide it to give more space to the pipeline UI.

jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
```diff
@@ -0,0 +1,175 @@
...
+      <div className="pachyderm-pipeline-spec-preview pachyderm-pipeline-textarea-wrapper">
+        <label className="pachyderm-pipeline-preview-label">
+          Pipeline Spec Preview: {'  '}
+        </label>
+        <textarea
+          className="pachyderm-pipeline-spec-preview-textarea"
+          style={{backgroundColor: '#80808080'}}
+          data-testid="Pipeline__specPreview"
+          name="specPreview"
+          value={`name: ${pipelineName}
```
smalyala, line 161: Are we not going to add the image and command to the pipeline spec preview?

jupyter-extension/src/plugins/mount/components/Pipeline/__tests__/Pipeline.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+describe('PPS screen', () => {
+  let setShowPipeline = jest.fn();
+  const mockRequestAPI = requestAPI as jest.Mocked<typeof requestAPI>;
+
+  beforeEach(() => {
+    setShowPipeline = jest.fn();
+    mockRequestAPI.requestAPI.mockImplementation(mockedRequestAPI({}));
+
+    // IntersectionObserver isn't available in test environment
+    const mockIntersectionObserver = jest.fn();
```
smalyala, line 19: You can remove these lines related to `mockIntersectionObserver`. This is used for tests that determine whether an HTML element is visible on the screen (different from when it's rendered but hidden behind another element).

jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
```diff
@@ -0,0 +1,175 @@
...
+};
+
+const placeholderInputSpec = `pfs:
+  repo: images
+  branch: dev
+  glob: /*
+`;
+const placeholderRequirements = './requirements.txt"';
+
+const Pipeline: React.FC<PipelineProps> = ({showPipeline, setShowPipeline}) => {
```
smalyala, line 17: You can remove `showPipeline` as a prop since it's not used anywhere

jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
bbonenfant: I don't think we need to show the user the command, since it will not be useful to them, but I'll add the image here.

jupyter-extension/src/plugins/mount/mount.tsx
bbonenfant: I removed the signal in a way that made sense to me, but I'll wait for Peter's feedback

jupyter-extension/src/plugins/mount/mount.tsx
bbonenfant: Nice I was wondering how to get rid of that, haha

pleeko (approved):
LGTM

PR #8558, opened 2023-02-01
[CORE-1099] Address long names compatibly

robert-uhl:
This is a combination of suggestions from @jrockway and @brendoncarroll.


src/internal/ppsutil/util.go
```diff
@@ -46,19 +49,85 @@ import (
...
+		//
+		// TODO(CORE-1099): deal with name collision & too-long names
+		pipelineName := strings.ReplaceAll(pipelineName, "_", "-")
+		return fmt.Sprintf("pipeline-%s-v%d", strings.ToLower(pipelineName), version), nil
+	case "v2.5":
+		version := strconv.FormatUint(version, 10)
+		charsAvailable := 63 - (len("pipeline---v") + len(version))
+		pipeline := &pps.Pipeline{Project: &pfs.Project{Name: projectName}, Name: pipelineName}
+		h := sha1.New()
+		io.WriteString(h, pipeline.String()) //nolint:errcheck
```
brendoncarroll, line 82: `pipeline.String` seems very likely to change.  Do we depend on this elsewhere?  I would be more confident in a` `fmt.Sprintf` here which accesses the fields of pipeline directly.

src/internal/ppsutil/util.go
```diff
@@ -46,19 +49,85 @@ import (
...
+		// or underscores
+		//
+		// TODO(CORE-1099): deal with name collision & too-long names
+		pipelineName := strings.ReplaceAll(pipelineName, "_", "-")
+		return fmt.Sprintf("pipeline-%s-v%d", strings.ToLower(pipelineName), version), nil
+	case "v2.5":
+		version := strconv.FormatUint(version, 10)
+		charsAvailable := 63 - (len("pipeline---v") + len(version))
+		pipeline := &pps.Pipeline{Project: &pfs.Project{Name: projectName}, Name: pipelineName}
+		h := sha1.New()
```
brendoncarroll, line 81: SHA1 will definitely work here, but it doesn't feel great to use a hash function which is no longer secure against collisions.
https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html

 We have the `pachhash` package which is just BLAKE2b, we use that throughout the storage layer.  You can truncate the output.  `pachhash.Sum(input)[:20]`

brendoncarroll: I saw your comment on the design doc.  Given the switch to base32, I think 16 bytes is probably a better compromise.

src/internal/ppsutil/util.go
```diff
@@ -46,19 +49,85 @@ import (
...
-	pipelineName := strings.ReplaceAll(pi.Pipeline.Name, "_", "-")
-	if projectName := pi.Pipeline.Project.GetName(); projectName != "" {
-		projectName = strings.ReplaceAll(projectName, "_", "-")
-		return fmt.Sprintf("pipeline-%s-%s-v%d", strings.ToLower(projectName), strings.ToLower(pipelineName), pi.Version)
+// manages a pipeline's workers.
+func PipelineRcName(pi *pps.PipelineInfo) (string, error) {
+	return pipelineRcName(pi.GetDetails().GetK8SNameVersion(), pi.GetPipeline().GetProject().GetName(), pi.GetPipeline().GetName(), pi.GetVersion())
+}
+
+func pipelineRcName(nameVersion, projectName, pipelineName string, version uint64) (string, error) {
```
brendoncarroll, line 68: Is `nameVersion` something we already track?  Otherwise I think the version of Pachyderm the pipeline was created with would be more valuable than starting a separate format with versions just for the k8s name.

robert-uhl:
Handled in #8559.

PR #8559, opened 2023-02-01
[CORE-1099] Address the issue of overlong Kubernetes resource names

robert-uhl:
Replaces `pipeline-` with `$PROJECT-` for all projects; constrains project names to 51 characters.


src/server/pps/server/api_server.go
```diff
@@ -79,6 +79,10 @@ const (
 	// default we return logs from up to 24 hours ago.
 	DefaultLogsFrom  = time.Hour * 24
 	ppsTaskNamespace = "/pps"
+
+	// kubernetesNameLimit is the maximum length of a ReplicationController
+	// or Service name.
+	kubernetesNameLimit = 63
```
jrockway, line 85: This is a slightly inaccurate name.  It's really services that are to blame, which then blames some ancient RFCs: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

RFC1123, incidentally, says that clients MUST understand 63 character DNS names, but SHOULD understand 255.  RFC1035 mandates 63 without explanation, and mandates 255 for entire DNS paths with the explanation that it's "for simplicity".  RFC1035 is from 1987 😂

src/server/pps/server/api_server.go
```diff
@@ -1770,6 +1774,11 @@ func (a *apiServer) validatePipelineRequest(request *pps.CreatePipelineRequest)
 		return errors.Errorf("pipeline name is %d characters long, but must have at most 63: %q",
```
jrockway, line 1774: Didn't we want to change this to 51?

Also for these comparisons that yield an error message, I recommend writing:

```go
if got, want := len(request.Pipeline.Name), 63; got != want {
        return errors.Errorf("pipeline name is %d characters long, but must have at most %d", got, want)
}
```

(Though I don't bother making it read that well in general.  "pipeline name too long; got %d, want %d".)

src/server/pps/server/api_server.go
```diff
@@ -1770,6 +1774,11 @@ func (a *apiServer) validatePipelineRequest(request *pps.CreatePipelineRequest)
...
 			len(request.Pipeline.Name), request.Pipeline.Name)
 	}
+	// TODO(CORE-1489): Remove dependency of name length on Kubernetes
+	// resource naming convention.
+	if max := kubernetesNameLimit - len("-v99"); len(request.Pipeline.Name)+len(request.Pipeline.Project.GetName())+1 > max {
```
jrockway, line 1779: We should figure out some way here to invoke ppsutil.PipelineRCName here, and then compare the length of the result to kubernetesNameLimit directly.  The name of the service/RC should also end up in the error message, as well.

I'd just `pps.PipelineInfo{Pipeline: request.Pipeline, Version: 99}` to get something you call PipelineRCName with.

As written, this is too action-at-a-distance-y.  The +1 must be the - separating project and pipeline, but nobody is going to realize that when they go to change the names.

src/server/pps/server/worker_rc_test.go
```diff
@@ -19,7 +19,7 @@ import (
 
 func mergeDefaultOptions(project, name string, add *workerOptions) *workerOptions {
 	base := &workerOptions{
-		rcName: strings.Join([]string{"pipeline", project, name, "v0"}, "-"),
+		rcName: strings.Join([]string{project, name, "v0"}, "-"),
```
jrockway, line 22: I'm pretty sure this doesn't matter (could be called "foobar" for all the tests care), but let's call pps.PipelineRCName here as well.

jrockway (approved):

src/server/pps/server/api_server.go
jrockway: dnsNameLimit?

src/server/pps/server/api_server.go
robert-uhl: I think that may have been verbal but not written, but I am okay doing that; the next check will neck it down further anyway.

How do you feel about messages of the form `pipeline name %q is %d characters longer than %d max`?

src/server/pachyderm_test.go
```diff
@@ -216,6 +216,38 @@ func TestSimplePipeline(t *testing.T) {
 			require.Equal(t, "foo", buf.String())
 		}
 	}
+
+	pipeline = strings.Repeat("x", 59-len(projectName)-1)
```
jrockway, line 220: This is fine to test here, but the most ideal situation would be a unit test for validateCreatePipelineRequest or if you want to be more careful, send a request to the API server setup with mockEnv.  There isn't a huge need to boot up pachyderm to have it reject the name.

src/server/pps/server/api_server.go
```diff
@@ -1766,9 +1772,14 @@ func (a *apiServer) validatePipelineRequest(request *pps.CreatePipelineRequest)
...
-		return errors.Errorf("pipeline name is %d characters long, but must have at most 63: %q",
-			len(request.Pipeline.Name), request.Pipeline.Name)
+	if len(request.Pipeline.Name) > maxPipelineNameLength {
+		return errors.Errorf("pipeline name %q is %d characters longer than the %d max",
+			request.Pipeline.Name, maxPipelineNameLength-len(request.Pipeline.Name), maxPipelineNameLength)
+	}
+	// TODO(CORE-1489): Remove dependency of name length on Kubernetes
+	// resource naming convention.
+	if k8sName := ppsutil.PipelineRcName(&pps.PipelineInfo{Pipeline: &pps.Pipeline{Project: &pfs.Project{Name: request.Pipeline.GetProject().GetName()}, Name: request.Pipeline.Name}, Version: 99}); len(k8sName) > dnsLabelLimit {
+		return errors.Errorf("Kubernetes name %q is %d characters longer than the %d max", k8sName, dnsLabelLimit-len(k8sName), dnsLabelLimit)
```
jrockway, line 1782: Very good message.

PR #8564, opened 2023-02-03
Djanicek/nightly perf stats collecting

djanicekpach:
CORE-1365 
This allows us to collect more metrics from the performance tests and include them in big query. 
I also update run_api_performance_test.sh to hopefully have a more reliable means for getting the right pachctl and pachd versions

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,7 +142,7 @@ jobs:
       - run: mkdir -p ${TEST_RESULTS}
       - run:
           name: Collect node stats
-          command: sar 10 -BbdHwzS -I SUM -n DEV -q -r ALL -u ALL -h | tee "${TEST_RESULTS}/sar_stats.tsv"
+          command: sar 10 -BbdHwzS -I SUM -n DEV -q -r ALL -u ALL -h -o "${TEST_RESULTS}/sar_stats"
           background: true
       - run:
           name: Clone Locust tests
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -47,6 +48,9 @@ parameters:
   pachd-latest-version:
     type: string
     default: ""
+  go-version:
+    type: string
+    default: "1.19"
```
molinamelendezj, line 53: what is this for?

djanicekpach: It's for line 156 in config.yaml. But, the real reason is that `make install` requires go 1.19 so for the fallback in run_api_performance_test.sh we need go 1.19+ and the image comes with 1.18.

.circleci/config.yml
```diff
@@ -236,6 +242,8 @@ workflows:
           matrix:
             parameters:
               version:
+                - v2.0.0
+                - v2.3.9
                 - v2.4.0
-                - v2.5.0-alpha.1
+                - v2.5.0-alpha.4
```
molinamelendezj, line 248: is this something we would have to keep changing?

molinamelendezj (approved):

.circleci/config.yml
djanicekpach: no, once GA is done we can leave it. The changes in etc/testing/circle/run_api_performance_tests.sh should let us automatically pick up the latest version more reliably.  So we only need to add new major versions after they're released. We could probably have an improvement to get the latest patch for each version, but I think that is out of scope for this.

PR #8569, opened 2023-02-09
pachhash: KVFingerprinter

brendoncarroll:
KVFingerprinter is for creating a unique fingerprint from the set of key value pairs.


src/internal/pachhash/fingerprinter.go
```diff
@@ -0,0 +1,49 @@
...
+// The zero value of KVFingerprinter is ready to use.
+type KVFingerprinter struct {
+	sum  Output
+	last []byte
+}
+
+// Absorb adds a key, value pair to the fingerprinter, and sets the last key to key.
+// keys must be added in ascending order.
+func (kv *KVFingerprinter) Absorb(key, value []byte) error {
+	if kv.last != nil && bytes.Compare(kv.last, value) >= 0 {
```
jrockway, line 21: I feel like if last is nil you can still pass it to Compare, and it will be less than everything except nil.

jrockway (approved):
tests necessary

PR #8570, opened 2023-02-09
stream: generic Iterator interface

brendoncarroll:


src/internal/grpcutil/iter.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/stream"
+	"google.golang.org/grpc"
+)
+
+type ClientStream[T any] interface {
+	Recv() (*T, error)
+	grpc.ClientStream
```
jrockway, line 12: I think this can be omitted.  If some other object also has Recv(*T, error), then let them be an iterator too :D

src/internal/grpcutil/iter.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"context"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/stream"
+	"google.golang.org/grpc"
+)
+
+type ClientStream[T any] interface {
+	Recv() (*T, error)
```
jrockway, line 11: Can this be T instead of *T (then the pointer nature is encoded by the type constraint above, and iterator[t] will return slices of t instead of *t)?

src/internal/grpcutil/iter.go
```diff
@@ -0,0 +1,37 @@
...
+
+func NewIterator[T any](cs ClientStream[T]) Iterator[T] {
+	return Iterator[T]{cs: cs}
+}
+
+func (it Iterator[T]) Next(ctx context.Context, x *T) error {
+	return it.cs.RecvMsg(x)
+}
+
+func ForEach[T any](cs ClientStream[T], fn func(x T) error) error {
```
jrockway, line 27: Yeah, this is why I think the constraint should be a pointer; fn should take a *T if Recv returns *T.

src/internal/stream/doc.go
```diff
@@ -0,0 +1,2 @@
+// package stream contains interfaces and helper functions for managing iterators that can block.
```
jrockway, line 1: package -> Package.  They made the convention "Package x ..." so that the sentence can start with an uppercase letter instead of the lowercase letter starting the package name :D

src/internal/stream/heap/heap.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	return Peek(h.x), true
+}
+
+func (h *Heap[T]) Len() int {
+	return len(h.x)
+}
+
+// Push adds x to the heap h
+func Push[E any, S ~[]E](h S, x E, lt func(a, b E) bool) S {
```
jrockway, line 36: Why do push and pop take lt, but the h also contains an lt added at New time?

src/internal/stream/merger.go
```diff
@@ -0,0 +1,79 @@
...
+type Merger[T any] struct {
+	its     []Peekable[T]
+	heap    heap.Heap[mergeEntry[T]]
+	isSetup bool
+}
+
+// NewMerger creates an iterator which merges the entries from its into a single iterator.
+// The entries will come out in ascending order.
+// The iterators slice can be thought of as layers, with higher layers masking the value of lower layers when the entries are equal.
+func NewMerger[T any](its []Peekable[T], lt func(a, b T) bool) *Merger[T] {
```
jrockway, line 29: This is a good interview question as well.

src/internal/grpcutil/iter.go
brendoncarroll: We need Context(), and I don't know if I'm ready for this to be used for things that aren't gRPC, "some other object" should implement stream.Iterator.

src/internal/grpcutil/iter.go
brendoncarroll: Does that work with `RecvMsg`?  I think we would get a double pointer.

Would it be better if`grpcutil.Iterator[T]` implements `stream.Iterator[*T]`?

src/internal/stream/heap/heap.go
brendoncarroll: Push and Pop are for using a slice as a heap directly.  The Heap type is an opaque container type.

src/internal/stream/merger.go
brendoncarroll: Lol it took me more than an hour.  You can't copy, because the caller providing the destination means we can't do copy or assignment of values, the only way to write something is to call `Peek` or `Next`

src/internal/grpcutil/iter.go
jrockway: So I guess my thought is that the type variable should directly match the return type of RecvMsg; right now we are assuming that exists, so the person creating an iterator has to know that they need to use a different type than what's actually in the API if that makes sense.  i.e if RecvMsg returns (Foo, error), they can't use the iterator.  It never does, of course.

BTW, I wonder if `proto.Message` is a better type constraint than `any`.

src/internal/stream/merger.go
jrockway: ahh, interesting.  i had to do this on the whiteboard at my google interview.  might have handwaved it as "oh it's just the merge from merge sort" tho.

src/internal/grpcutil/iter.go
brendoncarroll: I agree with `proto.Message` vs `any`, but I think that means `grpcutil.Iterator[T]` implements `stream.Iterator[*T]`, which means that it's Next method looks like this `Next(ctx, **T)`.  I'm not really sure if that's a good idea yet; maybe it should be private for now.

I'm more confident about the following:
- `stream.Iterator` taking a pointer to T for Next is the right API.
- functions in grpcutil should work with pointers to protocol buffer objects because that's how all the other protocol buffer APIs work.   e.g. `Collect[T](cs ClientStream, max int) ([]*T, error)`

src/internal/grpcutil/iter.go
jrockway: Fair.

src/internal/grpcutil/iter.go
brendoncarroll: That seams reasonable to me.  The big thing with `ForEach` is that the closure can't retain the object.  I guess it doesn't matter if it gets passed a pointer or not?

src/internal/stream/merger.go
brendoncarroll: If we had gone with the nicer, but less performant `Next(context.Context) (T, error)` API, then I think it get's easier.  Maybe easy enough to squeeze into an interview.

src/internal/stream/heap/heap.go
```diff
@@ -0,0 +1,99 @@
...
+		if !lt(h[j], h[i]) {
+			break
+		}
+		swap(h, i, j)
+		i = j
+	}
+	return i > i0
+}
+
+func swap[E any, S ~[]E](x S, i, j int) {
```
jrockway, line 97: Does this need the generic type signature?

src/internal/stream/iterator.go
```diff
@@ -0,0 +1,174 @@
...
+		if err := it.Next(ctx, &buf[i]); err != nil {
+			return n, err
+		}
+		n++
+	}
+	return n, nil
+}
+
+// Collect reads at most max from the iterator into a buffer and returns it.
+func Collect[T any](ctx context.Context, it Iterator[T], max int) (ret []T, _ error) {
```
jrockway, line 72: Does something like this perform any better (only 1 allocation, but potentially creates garbage):

```go
result := make([]T, max)
n, err := Read(ctx, it, result)
// handle err, overly-long read
return result[:n]
```

jrockway (approved):
LGTM

src/internal/grpcutil/iter.go
```diff
@@ -0,0 +1,53 @@
...
+
+type iterator[T proto.Message] struct {
+	cs ClientStream[T]
+}
+
+func newIterator[T proto.Message](cs ClientStream[T]) iterator[T] {
+	return iterator[T]{cs: cs}
+}
+
+func (it iterator[T]) Next(ctx context.Context, dst *T) error {
```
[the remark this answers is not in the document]
jrockway: I see the friendlier API below and this makes sense.  It's especially nice for the pre-allocated slice case.

For whatever reason I have never associated much cost with "return <pointer>" vs *result = pointer, but I don't have any actual evidence to back that up.

src/internal/stream/heap/heap.go
brendoncarroll: You could also do a callback like with the pre 1.18 sorting methods.  But if you want to do the swap on the slice then you have to take the type of the slice.

jrockway: Got it.  []any as a type is too loose, and doesn't allow passing []int for example.  The next level up `[S []any]` is also too broad, because the elements of the slice have to be all of the same type.  Makes sense!

src/client/pfs.go
```diff
@@ -178,7 +177,7 @@ func (c APIClient) ListProjectRepo(r *pfs.ListRepoRequest) ([]*pfs.RepoInfo, err
 	if err != nil {
 		return nil, grpcutil.ScrubGRPC(err)
 	}
-	return clientsdk.ListRepoInfo(client)
+	return grpcutil.Collect[*pfs.RepoInfo](client, 1000)
```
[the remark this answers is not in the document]
brendoncarroll: Lol that's the point.  Every call site of `clientsdk.List` was just rolling the dice with getting OOM'd. That produces a much less friendly error message than Collect.  Getting an escalation with the Collect error message is an easy fix: increase the value, or switch to a better iteration strategy.  Sometimes we don't get logs for OOM, so it's much longer to find the root cause.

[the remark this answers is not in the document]
jrockway: That's fair.  I have no objection to this.  Nobody is forcing anyone to upgrade to a new version of the Go client anyway.

src/internal/grpcutil/iter.go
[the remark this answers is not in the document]
brendoncarroll: Yeah this is not a performance boost.  This is just the most straightforward mapping of the gRPC stuff to the generic iterators.

The next steps would be experimenting with calling `Reset()` on a single gRPC object.  That would allow us to reuse a single gRPC and any of it's backing buffers for new data.

src/internal/stream/iterator.go
brendoncarroll: It's fewer allocations, but I was thinking that max would be some significant fraction of the available memory in some cases, and we wouldn't want to allocate e.g. 1GB everytime just because that's what the worst case plan is.

PR #8571, opened 2023-02-09
Add an RPC to change the log level at runtime

jrockway:
This lets a cluster admin change the log level for all pachyderm pods with  pachctl command.  The log level change is gated with a duration; so without further input, the system will converge back to its original state.  The pachd that receives the initial request changes its log level, then finds all other pachyderm pods, and sends the RPC on to those pods.  pachd replicas set their level and return.  pachw can't serve the DebugServer, so doesn't participate (I tried adding the DebugServer, but it needs a working activated EnterpriseServer to authenticate the request).  workers get a request in the user container, and then propagate the request to the storage container.  Each pod's (actually container's) status is reported in the RPC, and the set of success/failures is relayed to pachctl.  A failure of one pod to switch does not return an RPC error, only if the serving pachd returned an error does propagation stop.  A pod failing to switch does not revert other changes, and we continue to attempt other pods after a failure.  The goal is a best effort switch of the log level throughout pachyderm that automatically reverts after a selected duration.

You can also change the level of the GRPC logger; that works the same way.

This is part of CORE-1368


src/server/debug/server/log.go
```diff
@@ -0,0 +1,51 @@
...
+		case debug.SetLogLevelRequest_DEBUG:
+			log.SetLevel(log.DebugLevel)
+			log.Debug(ctx, "set log level to debug")
+		case debug.SetLogLevelRequest_INFO:
+			log.SetLevel(log.InfoLevel)
+			log.Info(ctx, "set log level to info")
+		case debug.SetLogLevelRequest_ERROR:
+			log.SetLevel(log.ErrorLevel)
+			log.Error(ctx, "set log level to error")
+		default:
```
armaanv, line 46: can we add an error case for SetLogLevelRequest_OFF here?

armaanv (approved):
looks good to me!

src/server/debug/server/log.go
jrockway: I think the generic error is as good as any custom error we'd generate here.  SetLevel won't accept OFF (it's not a log.Level), so the message "cannot set log level to OFF" seems fine to me.

src/internal/log/loggers.go
```diff
@@ -57,6 +64,41 @@ func SetGRPCLogLevel(l zapcore.Level) {
...
+func revertLogLevel(tPtr *atomic.Pointer[time.Timer], levelVar atomicLeveler, originalLevel, newLevel zapcore.Level, d time.Duration, msg string) {
+	wasSet := make(chan struct{})
+	oldTimer := tPtr.Swap(time.AfterFunc(d, func() {
+		<-wasSet
+		cur := levelVar.Level()
+		levelVar.SetLevel(originalLevel)
+		zap.L().Info(msg, zap.Stringer("from", cur), zap.Stringer("to", originalLevel))
+	}))
+	if oldTimer != nil {
+		oldTimer.Stop()
```
brendoncarroll, line 84: Do we have to drain a channel here?

jrockway: Nope.  time.NewTimer requires that, but the channel is not part of a time.Timer itself, it's something NewTimer adds.

src/internal/log/loggers.go
```diff
@@ -57,6 +64,41 @@ func SetGRPCLogLevel(l zapcore.Level) {
...
+// revertLogLevel sets up a log level change that is reverted to an original level after the
+// duration, logging the revert with the text of msg, ensuring that even with multiple concurrent
+// calls, the most recent call cancels the actions of prior calls.
+func revertLogLevel(tPtr *atomic.Pointer[time.Timer], levelVar atomicLeveler, originalLevel, newLevel zapcore.Level, d time.Duration, msg string) {
+	wasSet := make(chan struct{})
+	oldTimer := tPtr.Swap(time.AfterFunc(d, func() {
+		<-wasSet
+		cur := levelVar.Level()
+		levelVar.SetLevel(originalLevel)
+		zap.L().Info(msg, zap.Stringer("from", cur), zap.Stringer("to", originalLevel))
```
brendoncarroll, line 81: I think this should be like max level, always print no matter what.  Otherwise this becomes difficult to reason about.  

The user set's it to something above info and now we can't see the changes.  Use of the global logger isn't great.

src/internal/log/loggers.go
```diff
@@ -57,6 +64,41 @@ func SetGRPCLogLevel(l zapcore.Level) {
...
+		oldTimer.Stop()
+	}
+	levelVar.SetLevel(newLevel)
+	close(wasSet) // prevent the revert from happening before we have the chance to set the new level
+}
+
+// SetLevelFor changes the global logger level for the set duration, and then reverts to the log
+// level at process startup.  Subsequent calls before expiration completely override previous calls;
+// the new level takes effect immediately and the previous scheduled revert is canceled.
+func SetLevelFor(l Level, d time.Duration) {
```
brendoncarroll, line 93: Discussed this some offline, but I think these could be methods on an object.  Maybe a `LogManager`, which get's a reference to the logger setup by pachd in the pachd constructor, and exposes an API to manipulate it.

src/internal/log/loggers.go
jrockway: That makes sense to me.  That's actually what we do when changing it.  I used Info here because Error logs alarm people and this isn't an error, but on the other hand, changing the log level to Error also isn't an error but that's what it's logged at.

src/internal/log/loggers.go
brendoncarroll: Down below, the RPC is on the debug server, I think that's a good an object as any to own the timer and any state associated with reverting a level.

src/internal/log/loggers.go
jrockway: I changed the API to let the caller pass in the logging function to use; this way, the logs also correspond to the request-id that caused the level to be changed.

src/internal/log/loggers.go
jrockway: I added a resettableLevel object. Much cleaner.  Also added extensive tests.

src/internal/log/level.go
```diff
@@ -0,0 +1,89 @@
...
+		cur:   new(atomic.Int32),
+		reset: new(atomic.Pointer[time.Timer]),
+	}
+	rl.orig.Store(int32(l))
+	rl.cur.Store(int32(l))
+	return rl
+}
+
+// Level returns the current level.
+func (rl resettableLevel) Level() zapcore.Level {
```
brendoncarroll, line 37: 2 pointers copied by value. I don't think there's anything incorrect about this.

A `*resettableLevel` with non-pointer values in the struct would probably have fewer de-references and better memory locality.  Not a blocker.

src/debug/debug.proto
```diff
@@ -42,8 +42,30 @@ message DumpRequest {
   int64 limit = 2;
 }
 
+message SetLogLevelRequest {
+    enum LogLevel {
+      UNKNOWN = 0;
+      DEBUG = 1;
```
brendoncarroll, line 48: Do these have to be sequential? Should we match the constants that zap provides for these.  In `slog` there is a power of 2s thing going on.

src/server/debug/server/log.go
```diff
@@ -0,0 +1,188 @@
...
+		// If not recursive mode, return now.
+		return result, nil
+	}
+	req.Recurse = false
+
+	// Recurse to other pachyderm processes.
+	pods := map[string]string{}
+	apps := map[string]string{
+		"pach-enterprise": strconv.Itoa(int(s.env.Config().Port)),
+		//"pachw": strconv.Itoa(int(s.env.Config().PeerPort)),
```
brendoncarroll, line 118: Commented?

brendoncarroll:
Transplanting our slack conversation:

Consider the situation
1. The initial log level is A
2. I set a temporary log level to B
3. I set the permanent log level to C before B has reverted.

Reverting to A doesn't make any sense.

The two options that do make sense are:
1. Clear the temporary log level.  The log level is now C, and there will be no automatic change without more API calls.
2. Keep the log level at B, when B is reverted, revert to C.

I think we decided on **1** and it looks like that's what's implemented, but just wanted to confirm.

src/server/debug/server/log.go
jrockway: I will delete.  I really want pachw to participate in this, but we can't authenticate our calls because pachw doesn't have an enterprise server to check validation status.  (Though honestly, pachw needs more thought.  If pachd spins it up while in extra-debugging mode, should pachw produce extra debug logs?  Probably yes.  But then we'd need extra startup environment variables; base log level, enhanced log level, remaining duration on the enhanced log level.  Well, to really get it right we'd need that.  I suppose workers suffer from the same problem, too.  Will need a mini-design.)

jrockway:
@brendoncarroll Yup, we do 1.  SetLevel and SetLevelFor both delete the revert timer and set the new log level; SetLevel additionally changes the "original" log level to the new one; SetLevelFor starts a new revert timer.

src/internal/log/level.go
jrockway: We talked about this a little in Slack.  I kind of got "talked into" this pattern by zap.AtomicLevel, which takes non-pointer receivers and has a pointer to the *atomic.Int32 that stores the level.

brendoncarroll:
This looks good, especially with the `resettableLevel` object encapsulating the mutable state.

The global state in the log package is still a concern.  Maybe it's fine to have one logger in the log package, but I think the server implementation (the debug server) should take a reference to a log level or something for it to manipulate.  The push with the `Env` structs everywhere was to break any sort of cross package shared global state, and make dependencies easy to configure.
Any object that has package level dependencies is inherently difficult to test.  Instead of intervening when the object is constructed, a developer would have to intervene at link time with a substitute package.  That's not easy or practical.

I guess the argument "one stderr/stdout => one logger" is persuasive, but it doesn't follow that we should add package level state as a dependency to one of the servers.

src/debug/debug.proto
jrockway: I'm not sure why slog uses powers of 2.  Powers of 2 always seem like that pattern where you `ONE | TWO | FOUR` to set some bits.  Maybe they mark each entry with ERROR|WARN|INFO|DEBUG and then do level & DEBUG == DEBUG to test whether or not to log at level DEBUG?  But why do that when < works just as well.

jrockway:
@brendoncarroll I removed the package-level SetLevelFor calls in favor of sharing the LogLevel and GRPCLogLevel objects with external packages, with a note that it's only for the debug package.  (The "public" versions are `interface { SetLevelFor(...) }` so encapsulation can't be violated too badly.) The SetLogLevel RPC now calls SetLevelFor on server.logLevel and server.grpcLogLevel, and the tests demonstrate the non-global nature.

src/server/debug/server/server.go
```diff
@@ -63,6 +63,8 @@ type debugServer struct {
 	sidecarClient *client.APIClient
 	marshaller    *jsonpb.Marshaler
 	database      *pachsql.DB
+
+	logLevel, grpcLevel log.LevelChanger
```
brendoncarroll, line 67: 👍

brendoncarroll (approved):
LGTM

PR #8572, opened 2023-02-11
miscutil: generic WorkDeduper

brendoncarroll:
Adds type parameter `K` for the type of key used when deduping requests.


jrockway (approved):
LGTM

jrockway:
Also while reviewing this I looked into (typed) sync.Map alternatives.  This is worth reading: https://puzpuzpuz.dev/so-long-syncmap.

PR #8577, opened 2023-02-14
Schema migration for auth_tokens table to extend subject column length

albscui:
As part of adding projects, we are extending existing auth tokens with the project name. However, this could lead to breakages because our auth token table has a limit of 64 characters. This PR double this limit to 128.


djanicekpach (approved):
looks good to me, I'll re-run the upgrade test once it's in.

PR #8579, opened 2023-02-15
CORE-1463 tool to check postgres connection

zmajeed:
Usage: pach_postgres_check -connection_string str [-timeout number]
  -connection_string string
        postgres database connection string
  -timeout float
        connection attempt timeout in seconds (default 10)

Example:
```
pach_postgres_check --connection_string "host=localhost port=34265 user=pachyderm dbname=pachyderm" --timeout 1
```


etc/postgres_connect_check/pach_postgres_check.go
```diff
@@ -0,0 +1,36 @@
...
+        fmt.Fprintf(os.Stderr, "error: database connection string is required with -connection_string")
+        os.Exit(1)
+    }
+
+    ctx, cancel := context.WithTimeout(context.Background(), time.Duration(*timeout * float64(time.Second)))
+    defer cancel()
+
+    conn, err := pgx.Connect(ctx, *connectionString)
+    if err != nil {
+        fmt.Fprintf(os.Stderr, "error: database connection failed with connection string \"%s\": %v\n", *connectionString, err)
```
jrockway, line 31: Use %q to quote strings.

etc/postgres_connect_check/pach_postgres_check.go
```diff
@@ -0,0 +1,36 @@
...
+    "os"
+    "flag"
+    "time"
+
+    "github.com/jackc/pgx/v5"
+)
+
+func main() {
+
+    connectionString := flag.String("connection_string", "", "postgres database connection string")
```
jrockway, line 17: Does this connection string want to be in pgx's format, or in pachyderm's format (see service_env.go/initDBClient; we build it from environment variables)?

It looks like pgbouncer is configured in our setup with $POSTGRESQL_USERNAME, $POSTGRESQL_DATABASE, $POSTGRESQL_HOST, and $POSTGRESQL_PASSWORD and it would be convenient for this tool to take it in that form as well.

etc/postgres_connect_check/pach_postgres_check.go
```diff
@@ -0,0 +1,36 @@
...
+    timeout := flag.Float64("timeout", 10, "connection attempt timeout in seconds")
```
jrockway, line 18: This should be a flag.Duration.  This will allow the flag to be in this format:

```
    ParseDuration parses a duration string. A duration string is a possibly
    signed sequence of decimal numbers, each with optional fraction and a unit
    suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns",
    "us" (or "µs"), "ms", "s", "m", "h".
```

etc/postgres_connect_check/pach_postgres_check.go
```diff
@@ -0,0 +1,36 @@
...
+
+    ctx, cancel := context.WithTimeout(context.Background(), time.Duration(*timeout * float64(time.Second)))
+    defer cancel()
+
+    conn, err := pgx.Connect(ctx, *connectionString)
+    if err != nil {
+        fmt.Fprintf(os.Stderr, "error: database connection failed with connection string \"%s\": %v\n", *connectionString, err)
+        os.Exit(2)
+    }
+    conn.Close(ctx)
```
jrockway, line 34: Close is apparently an RPC (it sends a close message to postgres and waits for a response), so what do we want to do if this fails?  I think we probably just want to exit without closing; this timing out will cause a hard-to-detect error.

jrockway (commented):
What's the plan to get this binary into the pg bouncer container?  Should the Go code live in https://github.com/pachyderm/pg-bouncer so it's easy to build the binary into our pgbouncer fork?

Either way, if we want to distribute this tool with pachyderm, we should put it in the src/ directory, with the entrypoint in src/server/cmd/postgres-check/main.go.  Code in etc/ doesn't get linted and can't use src/internal/... libraries.  So there is some testing code and codegen code in there, but nothing we ship to users.  (Really, there's no good reason why codegen is in etc/.)

Finally, we should have some sort of test that checks that this works.

PR #8581, opened 2023-02-15
[PFS-57] Expose Option for Tuning Max Task Concurrency in PutFileURL 

FahadBSyed:


src/pfs/pfs.proto
```diff
@@ -346,6 +346,7 @@ message AddFile {
   message URLSource {
     string URL = 1;
     bool recursive = 2;
+    int64 concurrency = 3;
```
brendoncarroll, line 349: I was a little unclear on where this should go, but I think it makes sense here.  Any URLSource is something we could import using our new Import process.

brendoncarroll: Can you make this an `uint32` or `uint64`, that way we don't have to worry about the negative case?

FahadBSyed: Sure can do

src/client/option.go
```diff
@@ -24,6 +25,13 @@ func WithDatumPutFile(datum string) PutFileOption {
 	}
 }
 
+// WithConcurrency configures the maximum number of tasks in flight created by PutFileURL.
+func WithConcurrency(concurrency int64) PutFileOption {
```
brendoncarroll, line 29: `WithConcurrency` might be too overloaded of a term, but I guess this use case got to it first.

`WithImportConcurrency` might be less overloaded for a `PutFileOption`.

Not a blocker.

brendoncarroll (approved):
LGTM

etc/proto/test_no_changes.sh
```diff
@@ -18,6 +18,10 @@ find src -regex ".*\.pb\.go" \
 )"
 
 make proto
+git status
+git add src/pps/pps.pb.go
+git commit -m "[CORE-57] protogen hack"
+git push origin PFS-57
```
brendoncarroll, line 24: What's going on here?

FahadBSyed: I have a weird problem with my proto generation where a specific pps proto file gets generated slightly differently (I think something to do with whitespace) locally when I run make proto. This wasn't a problem in the past, but we added a check to validate the protos match by hash a few months ago. So I work around it by having circleCI commit the proto it generated for that file. Then I pull down the commit and force push over it with this file removed. 

Its annoying, but its an easy workaround.

src/client/option.go
FahadBSyed: That makes sense, WithConcurrency is overloaded, I agree.

etc/proto/test_no_changes.sh
brendoncarroll: Gotcha, just remember to pull it out before merging.

PR #8591, opened 2023-02-17
Get the file sets in the concat step in parallel

brycemcanally:
This PR changes the concat step such that we get the file sets in parallel. Certain workloads with large volumes of data and / or high database latency may notice a performance improvement with this change.

### DIFF
--- a/src/internal/storage/fileset/storage.go
+++ b/src/internal/storage/fileset/storage.go
@@ -14,6 +14,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset/index"
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/renew"
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/track"
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
 )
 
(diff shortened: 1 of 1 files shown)


src/internal/storage/fileset/storage.go
```diff
@@ -240,18 +241,37 @@ func (s *Storage) getPrimitives(ctx context.Context, ids []ID) ([]*Primitive, er
...
-	var size int64
-	additive := index.NewWriter(ctx, s.ChunkStorage(), "additive-index-writer")
-	deletive := index.NewWriter(ctx, s.ChunkStorage(), "deletive-index-writer")
-	for _, id := range ids {
-		md, err := s.store.Get(ctx, id)
-		if err != nil {
+	cancelCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	eg, ctx := errgroup.WithContext(cancelCtx)
+	sem := semaphore.NewWeighted(10)
```
brendoncarroll, line 247: The semaphore + errgroup might be simplified with SetLimit.
https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit

brendoncarroll (approved):

src/internal/storage/fileset/storage.go
brycemcanally: The semantics aren't ideal with that functionality. We want to exit immediately if the context is canceled, not go through all of the goroutines.

brendoncarroll (approved):

PR #8592, opened 2023-02-17
[CORE-1501] handle different formats of loki logs for kube events

armaanv:
Loki logs can be formatted differently based on the container runtime being used and the loki configuration. This PR updates the `pachctl kube-events` command to accept all three possible formats


jrockway (approved):
Very nice!

jrockway:
BTW, I tested thsi branch and it works great.

PR #8593, opened 2023-02-17
Migrate storage packages to stream.Iterator

brendoncarroll:


brendoncarroll:
Only -300 LOC net, but we avoid a lot of casts from `interface{}` and remove some of the awkwardness around comparing a stream to another stream.

PR #8594, opened 2023-02-17
[PFS-55] Implement efficient file concatenation

brycemcanally:
This PR implements efficient file concatenation. Large files that are spread across multiple file sets (i.e. due to append) will no longer need to be fully read during compaction in the majority of cases. This is supported by a new copy algorithm in the chunk uploader which will reuse stable data references in the provided list while creating new chunks (in parallel) for the unstable data references. This is primarily enabled by a new alignment algorithm that detects when we can start reusing stable data references after encountering a sublist of unstable data references. This will enable us to better support the distributed upload of large files since we won't be bottlenecked by the file being read by one machine during compaction.


src/internal/storage/chunk/storage.go
```diff
@@ -62,7 +62,8 @@ func NewStorage(objC obj.Client, memCache kv.GetPut, db *pachsql.DB, tracker tra
 // NewReader creates a new Reader.
 func (s *Storage) NewReader(ctx context.Context, dataRefs []*DataRef, opts ...ReaderOption) *Reader {
 	client := NewClient(s.store, s.db, s.tracker, nil)
-	return newReader(ctx, client, s.memCache, s.deduper, s.prefetchLimit, dataRefs, opts...)
+	defaultOpts := []ReaderOption{WithPrefetchLimit(s.prefetchLimit)}
```
brendoncarroll, line 65: Shouldn't the defaults be what we initialize the reader with?  Maybe we define them in a constant, and then we refer to the constant when we create the `reader` in the structure.

src/internal/storage/chunk/uploader.go
```diff
@@ -76,6 +81,97 @@ func (u *Uploader) Upload(meta interface{}, r io.Reader) error {
 }
 
 func (u *Uploader) Copy(meta interface{}, dataRefs []*DataRef) error {
```
brendoncarroll, line 83: Can we add a comment to this? What is meta allowed to be?  What does this copy?

src/internal/storage/chunk/uploader.go
```diff
@@ -76,6 +81,97 @@ func (u *Uploader) Upload(meta interface{}, r io.Reader) error {
...
+			splitBytesLeft -= chunkBytesLeft
+			return nil
+		})
+	}); err != nil && !errors.Is(err, errutil.ErrBreak) {
+		return 0, err
+	}
+	return i, nil
+}
+
+func (u *Uploader) CopyByReference(meta interface{}, dataRefs []*DataRef) error {
```
brendoncarroll, line 174: Should this be public?  If it's possible to misuse this, then it should be private and we can call into it in Copy

src/internal/storage/chunk/uploader.go
```diff
@@ -76,6 +81,97 @@ func (u *Uploader) Upload(meta interface{}, r io.Reader) error {
...
+			stableDataRefs[len(stableDataRefs)-1].Ref.Edge = true
+		}
+		return func() error {
+			return u.cb(meta, stableDataRefs)
+		}, nil
+	})
+}
+
+// A data reference is stable if it does not reference an edge chunk and references the full chunk.
+func stableDataRef(dataRef *DataRef) bool {
```
brendoncarroll, line 140: nit: `isStableDataRef` communicates the purpose better IMO

src/internal/storage/chunk/uploader.go
```diff
@@ -76,6 +81,97 @@ func (u *Uploader) Upload(meta interface{}, r io.Reader) error {
...
+		nextDataRefs = append(nextDataRefs, dataRefs[i])
+	}
+	if err := appendDataRefs(nextDataRefs); err != nil {
+		return err
+	}
+	return u.taskChain.CreateTask(func(_ context.Context) (func() error, error) {
+		if err := taskChain.Wait(); err != nil {
+			return nil, err
+		}
+		if len(dataRefs) > 0 {
```
brendoncarroll, line 129: This refers to dataRefs potentially after Copy has returned.  I think that's fine because the slice header (including the length) is passed by value.  Accessing this would not be okay. 

I think a comment on the function explaining how dataRefs is or isn't retained would be helpful.

src/internal/storage/fileset/fileset_test.go
```diff
@@ -291,6 +299,14 @@ func testStableHash(ctx context.Context, t *testing.T, data, expected []byte, ms
 		ids = append(ids, *id)
 	}
 	getHash := func() []byte {
+		if compact {
+			compactFunc := func(ctx context.Context, ids []ID, ttl time.Duration) (*ID, error) {
```
brendoncarroll, line 303: This literal seems unnecessary.  Am I missing something?  Can we pass `storage.Compact` instead?

src/internal/storage/chunk/uploader.go
brendoncarroll: This comment might be misplaced:  Somewhere in here there needs to be a call to client in order to create the dependency between the chunks.  That is also the mechanism that prevents dangling references.

src/internal/storage/fileset/writer.go
```diff
@@ -138,6 +138,9 @@ func (w *Writer) Copy(file File, datum string) error {
 				Datum: datum,
 			},
 		}
+		if _, ok := file.(*FileReader); ok {
+			return w.uploader.CopyByReference(copyIdx, idx.File.DataRefs)
```
brendoncarroll, line 142: I like having CopyByReference available to us here.  It definitely simplifies things.  I don't think it's safe enough to be public though, see my other comment above.

src/internal/storage/chunk/uploader.go
```diff
@@ -76,6 +81,97 @@ func (u *Uploader) Upload(meta interface{}, r io.Reader) error {
...
+		}, nil
+	})
+}
+
+// A data reference is stable if it does not reference an edge chunk and references the full chunk.
+func stableDataRef(dataRef *DataRef) bool {
+	return !dataRef.Ref.Edge && dataRef.OffsetBytes == 0 && dataRef.SizeBytes == dataRef.Ref.SizeBytes
+}
+
+func (u *Uploader) align(ctx context.Context, dataRefs []*DataRef, i int, cb func([]byte) error) (int, error) {
```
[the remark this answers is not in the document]
brendoncarroll: We could omit i, by slicing dataRefs in the caller.  Assume i starts at 0

src/internal/storage/chunk/storage.go
brycemcanally: Not sure I follow. The default mentioned here is the default based on the storage configuration, not the default for the reader. A storage instance can be created with a default prefetch limit that will be used for readers. This can be overridden by clients for particular use cases with the functional options provided to `NewReader`.

src/internal/storage/chunk/uploader.go
brycemcanally: We discussed this offline. This can only be misused in the context of the client (i.e. incorrect file hashes). A client can choose to either perform a normal copy when they want to ensure stable data references or use copy by reference when they just want to reuse the references as is.

src/internal/storage/chunk/uploader.go
brycemcanally: This has been updated to use stableDataRefs, which it should have been using anyways.

src/internal/storage/fileset/fileset_test.go
brycemcanally: No, the signatures are different. `storage.Compact` has functional options.

src/internal/storage/chunk/uploader.go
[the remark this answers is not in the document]
brycemcanally: I refactored this to iterate based on slicing.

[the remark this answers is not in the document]
brycemcanally: And added some documentation.

src/internal/storage/fileset/writer.go
brycemcanally: Refer to the relevant thread above.

brendoncarroll (approved):

PR #8606, opened 2023-02-21
[PFS-37] Add an RPC Endpoint to PFS to Search for a File in Branch 

FahadBSyed:
## Description
This PR implements the RPC endpoint described in the [File Version History ](https://docs.google.com/document/d/1mQ941ZQxbrjjfBkXO_4fgfHO6HkoIieiBuA9QbqH4VA/edit?usp=sharing)design document.

This endpoint takes a starting commit and a filepath and serially searches through the branch commit-by-commit. If a reference to the filepath, we add the commit to a list to return to the user. If the search request hits its deadline, it returns what it found so far as well as the last commit it searched. This allows a user to start a subsequent searching from that commit. 

This endpoint is primarily designed to be used by Console to display a list of commits to a user that contain a modification to a specific file. The goal is to have elements in the user interface that allow a user to continue a search (or make it easy for Console to continue a search in the background as it updates the list.)

```bash
$ make opencv
$ pachctl delete file images@master:/46Q8nDz.jpg
$ pachctl put file images@master -i "."/images.txt

$ pachctl find commit images@master:/46Q8nDz.jpg
FOUND IN COMMITS
0a44eadc08b2434eb526694f8933c171
5df154f78594467988f66825bad1e87e
e2208004c6df44dc8d3e21835009b512

LAST SEARCHED COMMIT
09476fef7242411bb320c8f7c5d0ac8d
$ pachctl find commit images@5df154f78594467988f66825bad1e87e:/46Q8nDz.jpg
FOUND IN COMMITS
5df154f78594467988f66825bad1e87e
e2208004c6df44dc8d3e21835009b512

LAST SEARCHED COMMIT
09476fef7242411bb320c8f7c5d0ac8d
$ pachctl find commit images@09476fef7242411bb320c8f7c5d0ac8d:/46Q8nDz.jpg
FOUND IN COMMITS
no commits found.

LAST SEARCHED COMMIT
09476fef7242411bb320c8f7c5d0ac8d
```


src/pfs/pfs.proto
```diff
@@ -300,6 +300,20 @@ message CreateBranchRequest {
   bool new_commit_set = 5; // overrides the default behavior of using the same CommitSet as 'head'
 }
 
+message SearchForFileInBranchRequest {
+  Commit start = 1;
+  string file_path = 2;
+  int32 limit = 3; // Does it make more sense to define a limit in terms of raw commits to search or in terms of the length of the returned 'found commits' array?
```
brendoncarroll, line 306: Limit usually refers to the amount returned.  In SQL for example the user doesn't specify the maximum number of rows to search.

Maybe we should expose the maximum number to search as well, but the user probably cares more about a deadline for the rpc, than the amount of commits checked.  gRPC has the ability to send that over as metadata and we can get at it with the `context.Context` that is provided.

src/pfs/pfs.proto
```diff
@@ -300,6 +300,20 @@ message CreateBranchRequest {
...
+  google.protobuf.Duration timeout = 4;
+}
+
+message SearchForFileInBranchResponse {
+  repeated Commit found_commits = 1;
+  Commit last_searched_commit = 2;
+  google.protobuf.Duration duration = 3;
```
brendoncarroll, line 313: The user can time duration themselves and it will will be more accurate from their POV than anything we time on our end.

src/pfs/pfs.proto
FahadBSyed: Okay cool. Thanks for the input. I think based your comment, we'll leave the limit logic the way it is now. We can also pass in a timeout via the context instead of defining it as a parameter, that makes more sense.

src/pfs/pfs.proto
```diff
@@ -300,6 +300,19 @@ message CreateBranchRequest {
...
+  string file_path = 2;
+  int32 limit = 3;
+  google.protobuf.Duration timeout = 4;
+}
+
+message SearchForFileInBranchResponse {
+  repeated Commit found_commits = 1;
+  Commit last_searched_commit = 2;
+  int32 commits_searched = 4;
+}
```
pleeko, line 314: Looks good

src/pfs/pfs.proto
```diff
@@ -631,6 +644,8 @@ service API {
   rpc SquashCommitSet(SquashCommitSetRequest) returns (google.protobuf.Empty) {}
   // DropCommitSet drops the commits of a CommitSet and all data included in the commits.
   rpc DropCommitSet(DropCommitSetRequest) returns (google.protobuf.Empty) {}
+  // SearchForFileInBranch searches for commits that reference a supplied file being modified in a branch.
+  rpc SearchForFileInBranch(SearchForFileInBranchRequest) returns (SearchForFileInBranchResponse) {}
```
pleeko, line 648: Are we not returning a stream here anymore? TBH the response object looks good and we are not really using streams correctly in console ATM (we read until stream is closed in backend and return result 😆).

pleeko (approved):

src/pfs/pfs.proto
FahadBSyed: That's correct. Its a non-streamed response

pleeko: works for me

src/pfs/pfs.proto
brendoncarroll: Our client wrapper isn't really idiomatic Go.  You can set the context, but not as an argument.  I've talked with @jrockway about plans to deprecate the client wrapper.  If we continue to support it, it needs to be reworked.  I wouldn't worry too much about this package/method.  Anyone seriously using the API will use gRPC directly, Hub (maybe the heaviest consumers of the APIs ever) is a good example.

src/pfs/pfs.proto
```diff
@@ -300,6 +300,19 @@ message CreateBranchRequest {
...
+message SearchForFileInBranchRequest {
+  Commit start = 1;
+  string file_path = 2;
+  int32 limit = 3;
+  google.protobuf.Duration timeout = 4;
+}
+
+message SearchForFileInBranchResponse {
+  repeated Commit found_commits = 1;
+  Commit last_searched_commit = 2;
```
brendoncarroll, line 312: Can this just be a Commit ID instead of the whole Commit?  Is there any benefit to sending over the whole thing?
The downsides I can see are:
- Using more bandwidth
- Being mistaken for a relevant commit

src/server/pfs/cmds/cmds.go
```diff
@@ -1017,6 +1021,59 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
...
+			if err != nil {
+				return err
+			}
+			commit := file.Commit
+			c, err := client.NewOnUserMachine("user")
+			if err != nil {
+				return err
+			}
+			defer c.Close()
+			resp, err := c.SearchForFileInBranch(&pfs.SearchForFileInBranchRequest{
```
brendoncarroll, line 1039: If you are using are client, you need to either get out the PFS gRPC and call it with a context that you configure with the timeout.  Or you can call the method on our client which clones it and sets the context.

src/server/pfs/cmds/cmds.go
```diff
@@ -1017,6 +1021,59 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
...
+				Start:    commit,
+				FilePath: file.Path,
+				Timeout:  types.DurationProto(timeout),
+				Limit:    limit,
+			})
+			if err != nil {
+				return grpcutil.ScrubGRPC(err)
+			}
+			writer := tabwriter.NewWriter(os.Stdout, "FOUND IN COMMITS\n")
```
brendoncarroll, line 1048: +1 to `os.Stdout` and `tabwriter` instead of fmt.Println

brendoncarroll: Ideally we would use the writer that cobra gives us (which enables better testing), but our `cmdutil` package throws that out so os.Stdout is fine.

src/pfs/pfs.proto
brendoncarroll, line 648: The Unary rpc makes the protocol simpler (since we only need one message type).  But the other methods that list commits return streams.  We should probably be consistent with the rest of the API, unless there is a plan for migration to unary RPCs for iteration.

I don't think I'm up to date on the latest thinking there, so I'd defer to @armaanv or someone else on CORE

src/server/pfs/server/api_server.go
```diff
@@ -398,6 +401,101 @@ func (a *apiServer) DeleteBranch(ctx context.Context, request *pfs.DeleteBranchR
...
+	if err = diffFileSet.Iterate(ctx, func(file fileset.File) error {
+		if file.Index().Path == filePath {
+			found = true
+			return errutil.ErrBreak
+		}
+		return nil
+	}); err != nil && !errors.Is(err, errutil.ErrBreak) {
+		return false, err
+	}
+	// a file cannot be in the additive and deletive set, so if its found, skip iteration over deletive.
```
brendoncarroll, line 483: Is this true?  I thought you do a put with a delete + add

src/server/pfs/server/testing/server_test.go
```diff
@@ -4274,6 +4274,72 @@ func TestPFS(suite *testing.T) {
 		//	}
 	})
 
+	suite.Run("SearchFileInBranch", func(t *testing.T) {
```
brendoncarroll, line 4277: This looks good ✅

src/pfs/pfs.proto
brendoncarroll: Separately, this is a descriptive name, but it might be worth making this more general.

In the design doc you had proposed `FindCommits`, which searches for commits matching arbitrary criteria.  I don't think we should deviate too far from that model/level of generality.

brendoncarroll:
I'm going through the design doc we had for this again and making notes:
https://docs.google.com/document/d/1mQ941ZQxbrjjfBkXO_4fgfHO6HkoIieiBuA9QbqH4VA/edit#heading=h.9cbbj7wdvlvz

- [x] No need to do distributed or concurrent processing for v0.  We can revisit if there are perf issues. 
- [x] Searching linearly is okay, we can return a stream of X items and the API can have a commit to start searching from
A deadline is okay
- [ ] Design API as a stream
- [x ] Diff fileset compaction will be done beforehand. (Done in another PR)
- [x ] Path based searching is okay
- [x] Also want to return the number of commits searched, ID of first searched commit and ID of last commit

src/pfs/pfs.proto
FahadBSyed: Returning the commit is more convenient for the client since they can use the reference to begin a new search without having to make another API call to get the commit from the ID.

src/server/pfs/cmds/cmds.go
FahadBSyed: I figured this would be a better setup if we wanted to add more columns to the output later as well.

src/pfs/pfs.proto
FahadBSyed: Gotcha, `FindCommits` is probably a good candidate for being general enough while retaining intention/description

src/server/pfs/server/api_server.go
FahadBSyed: Oh interesting, Is that the case?
Actually this comment should be reworked, because if a file is found in the additive, we can skip iterating over deletive because we don't currently care about the operation, just that the file was found.

src/pfs/pfs.proto
FahadBSyed: Also, we did agree on using streams which is something I forgot about while doing the implementation, so that should be easy enough to swap to

src/pfs/pfs.proto
brendoncarroll: Yeah that makes sense.  I got Commit confused with CommitInfo.

src/server/pfs/server/api_server.go
brendoncarroll: For the modified case, a match in either would be sufficient.  You can check both in parallel or just one and possibly skip the other.

src/pfs/pfs.proto
```diff
@@ -300,6 +300,19 @@ message CreateBranchRequest {
...
+  int32 commits_searched = 4;
```
brendoncarroll, line 313: This can be a uint32

src/internal/clientsdk/pfs.go
```diff
@@ -147,6 +147,48 @@ func ListCommit(client pfs.API_ListCommitClient) ([]*pfs.CommitInfo, error) {
...
+			if errors.Is(err, pacherr.ErrBreak) {
+				err = nil
+			}
+			return err
+		}
+	}
+	return nil
+}
+
+type FindCommitsResponseUser struct {
```
brendoncarroll, line 169: This seems useful to me, probably worth reviving `clientsdk` for.

src/server/pfs/cmds/cmds.go
```diff
@@ -1017,6 +1022,68 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
...
+			findCommitClient, err := c.PfsAPIClient.FindCommits(ctx, &pfs.FindCommitsRequest{
+				Start:    commit,
+				FilePath: file.Path,
+				Limit:    limit,
+			})
+			if err != nil {
+				return grpcutil.ScrubGRPC(err)
+			}
+			writer := tabwriter.NewWriter(os.Stdout, "FOUND IN COMMITS\n")
+			return grpcutil.ScrubGRPC(clientsdk.ForEachFindCommitsResponse(findCommitClient, func(x *pfs.FindCommitsResponse) error {
```
brendoncarroll, line 1055: This can be changed to `grpcutil.ForEach`

src/server/pfs/server/driver.go
```diff
@@ -498,6 +498,94 @@ func (d *driver) inspectProject(ctx context.Context, project *pfs.Project) (*pfs
 	return pi, nil
 }
 
+func (d *driver) findCommits(ctx context.Context, request *pfs.FindCommitsRequest, cb func(response *pfs.FindCommitsResponse) error) error {
```
brendoncarroll, line 501: Is there an advantage to having a callback function here rather than a gRPC stream?

src/internal/clientsdk/pfs.go
```diff
@@ -147,6 +147,48 @@ func ListCommit(client pfs.API_ListCommitClient) ([]*pfs.CommitInfo, error) {
 	return results, nil
 }
 
+func ForEachFindCommitsResponse(client pfs.API_FindCommitsClient, cb func(response *pfs.FindCommitsResponse) error) error {
```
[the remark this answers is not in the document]
FahadBSyed: Yup, I saw it when I pulled in changes from master

src/internal/clientsdk/pfs.go
FahadBSyed: This can probably live in the client package.

src/server/pfs/cmds/cmds.go
```diff
@@ -1017,6 +1022,68 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
...
+				switch x.Result.(type) {
+				case *pfs.FindCommitsResponse_FoundCommit:
+					_, err := writer.Write([]byte(x.GetFoundCommit().ID + "\n"))
```
brendoncarroll, line 1058: nit: since this writer is buffered, it's probably faster to let it do the concatenation than creating a new string and passing it to the writer.

src/server/pfs/server/driver.go
FahadBSyed: I'm not sure actually. I was attempting to stay consistent with the pattern used for streaming RPCs based on our existing implementations in our ListCommit/ListBranch/ListRepo endpoints. They seem to pass in the server.Send() as a callback which is used by the driver to stream back data to the client.

src/server/pfs/cmds/cmds.go
FahadBSyed: Could you elaborate more on what you mean here by letting the writer do the concatenation?

src/pfs/pfs.proto
```diff
@@ -300,6 +300,20 @@ message CreateBranchRequest {
...
 }
 
+message FindCommitsRequest {
+  Commit start = 1;
+  string file_path = 2;
+  uint32 limit = 3; // a limit of 0 means there is no upper bound on the limit.
+}
+
+message FindCommitsResponse {
+  oneof result {
```
armaanv, line 310: why do we return only one of these?

FahadBSyed: This message is returned as a stream. If we find 5 commits, there will be 1 message per commit. On the client side, we stitch the response together into a flat struct that we can present to the user.

src/server/pfs/cmds/cmds.go
```diff
@@ -1016,6 +1021,66 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
...
+			writer := tabwriter.NewWriter(os.Stdout, "FOUND IN COMMITS\n")
+			return grpcutil.ScrubGRPC(grpcutil.ForEach[*pfs.FindCommitsResponse](findCommitClient, func(x *pfs.FindCommitsResponse) error {
+				switch x.Result.(type) {
+				case *pfs.FindCommitsResponse_FoundCommit:
+					_, err := writer.Write([]byte(x.GetFoundCommit().ID + "\n"))
+					if err != nil {
+						return errors.EnsureStack(err)
+					}
+				case *pfs.FindCommitsResponse_LastSearchedCommit:
+					if _, err = writer.Write([]byte("\n")); err != nil {
```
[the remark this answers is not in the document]
FahadBSyed: That's a good idea. I think Kubernetes has a not found message as well.

src/server/pfs/cmds/cmds.go
```diff
@@ -1016,6 +1021,66 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
 	shell.RegisterCompletionFunc(deleteBranch, shell.BranchCompletion)
 	commands = append(commands, cmdutil.CreateAliases(deleteBranch, "delete branch", branches))
 
+	FindCommits := &cobra.Command{
+		Use:   "{{alias}} <repo>@<branch-or-commit>:<path/in/pfs> [flags]",
+		Short: "search for commits with reference to <filePath> within a branch starting from <repo@commitID>",
```
brendoncarroll, line 1026: This is user facing.  Let's change "search for" to find for consistency.

src/server/pfs/server/driver.go
```diff
@@ -498,6 +498,94 @@ func (d *driver) inspectProject(ctx context.Context, project *pfs.Project) (*pfs
...
+	diffID, err := d.commitStore.GetDiffFileSet(ctx, commit)
+	if err != nil {
+		return false, err
+	}
+	diffFileSet, err := d.storage.Open(ctx, []fileset.ID{*diffID})
+	if err != nil {
+		return false, err
+	}
+	found := false
+	if err = diffFileSet.Iterate(ctx, func(file fileset.File) error {
```
brendoncarroll, line 564: You need an `index.Option` with the path here for this to be O(1) not O(n)

src/server/pfs/cmds/cmds.go
```diff
@@ -1016,6 +1021,66 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
...
+		Long:  "search for commits with reference to <filePath> within a branch starting from <repo@commitID>",
+		Run: cmdutil.RunFixedArgs(1, func(args []string) error {
+			file, err := cmdutil.ParseFile(project, args[0])
```
brendoncarroll, line 1029: What do you think about making this another flag?  Since `FindCommit` is general enough to be extended with other matching criteria, we shouldn't assume a single argument as one of those criteria.

Maybe `--modifies`, `--modifies-path, `--changes-path`?

src/server/pfs/cmds/cmds.go
brendoncarroll: `writer.Write([]byte(x.GetFoundCommit().ID))`
and then 
`writer.Write("\n")`

Because it's buffered this should be efficient.  If the write made a system call than that would be a different story.

src/server/pfs/server/driver.go
```diff
@@ -498,6 +498,94 @@ func (d *driver) inspectProject(ctx context.Context, project *pfs.Project) (*pfs
...
+		}
+		resp.CommitsSearched = commitsSearched
+		return resp
+	}
+	for {
+		if searchDone {
+			return errors.EnsureStack(cb(makeResp(nil, commitsSearched, commit)))
+		}
+		if err := log.LogStep(ctx, "searchingCommit", func(ctx context.Context) error {
+			inspectCommitResp, err := d.inspectCommit(ctx, commit, pfs.CommitState_FINISHED)
```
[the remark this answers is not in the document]
FahadBSyed: That's a good question. I probably need to investigate that by checking the behavior against an open commit.

src/server/pfs/server/driver.go
```diff
@@ -498,6 +498,94 @@ func (d *driver) inspectProject(ctx context.Context, project *pfs.Project) (*pfs
...
+		}, zap.String("commit", commit.ID), zap.String("branch", commit.Branch.String()), zap.String("target", request.FilePath)); err != nil {
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+				return errors.EnsureStack(cb(makeResp(nil, commitsSearched, commit)))
+			}
+			return errors.EnsureStack(err)
+		}
+	}
+}
+
+func (d *driver) isFileInCommit(ctx context.Context, commit *pfs.Commit, filePath string) (bool, error) {
```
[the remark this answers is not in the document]
FahadBSyed: Yup, that's a much better name! Thanks

src/server/pfs/cmds/cmds.go
FahadBSyed: I think that change might be best left to do once we introduce other criteria for findCommit

src/server/pfs/cmds/cmds.go
```diff
@@ -1016,6 +1021,66 @@ Any pachctl command that can take a commit, can take a branch name instead.`,
...
+						return errors.EnsureStack(err)
+					}
+				case *pfs.FindCommitsResponse_LastSearchedCommit:
+					if _, err = writer.Write([]byte("\n")); err != nil {
+						return errors.EnsureStack(err)
+					}
+					if err = writer.Flush(); err != nil {
+						return errors.EnsureStack(err)
+					}
+					writer = tabwriter.NewWriter(os.Stdout, "LAST SEARCHED COMMIT\n")
```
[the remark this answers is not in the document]
FahadBSyed: I think the best thing to do would be to have a `--json` flag that just returns the response marshalled as json and do parsing that way

src/server/pfs/server/driver.go
```diff
@@ -498,6 +498,94 @@ func (d *driver) inspectProject(ctx context.Context, project *pfs.Project) (*pfs
...
+		return false, err
+	}
+	found := false
+	if err = diffFileSet.Iterate(ctx, func(file fileset.File) error {
+		if file.Index().Path == filePath {
+			found = true
+			return errutil.ErrBreak
+		}
+		return nil
+	}); err != nil && !errors.Is(err, errutil.ErrBreak) {
```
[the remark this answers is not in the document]
FahadBSyed: So it looks like they don't drop `ErrBreak` when its thrown as part of the callback (unless I'm using the interface wrong)

```
 server_test.go:4307: No error is expected but got BREAK
    require.go:588: github.com/pachyderm/pachyderm/v2/src/internal/grpcutil.ScrubGRPC
        	C:/Users/Deadb/Desktop/Pachyderm/pachyderm/src/internal/grpcutil/error.go:16
    require.go:588: github.com/pachyderm/pachyderm/v2/src/client.APIClient.FindCommits
        	C:/Users/Deadb/Desktop/Pachyderm/pachyderm/src/client/pfs.go:465
    require.go:588: github.com/pachyderm/pachyderm/v2/src/server/pfs/server/testing.TestPFS.func105
        	C:/Users/Deadb/Desktop/Pachyderm/pachyderm/src/server/pfs/server/testing/server_test.go:4306
    require.go:588: testing.tRunner
        	C:/Program Files/Go/src/testing/testing.go:1446
    require.go:588: runtime.goexit
        	C:/Program Files/Go/src/runtime/asm_amd64.s:1594
    server_test.go:4307: current stack:
        goroutine 206 [running]:
        runtime/debug.Stack()
```

src/server/pfs/server/driver.go
```diff
@@ -498,6 +498,94 @@ func (d *driver) inspectProject(ctx context.Context, project *pfs.Project) (*pfs
...
+			if inCommit {
+				log.Info(ctx, "found target", zap.String("commit", commit.ID), zap.String("branch", commit.Branch.String()), zap.String("target", request.FilePath))
+				found = commit
+				foundCommits++
+				if err := cb(makeResp(found, commitsSearched, nil)); err != nil {
+					return err
+				}
+			}
+			commitsSearched++
+			if (foundCommits == request.Limit && request.Limit != 0) || inspectCommitResp.ParentCommit == nil {
```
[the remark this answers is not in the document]
FahadBSyed: I don't think there is a situation where `commitsSearched` can be 0 when this conditional runs. We need to check the commit that was passed as part of the input request as minimum. My assumption is that if `inspectCommit()` is given a nil or invalid commit, it should throw and error before we get here.

src/server/pfs/server/driver.go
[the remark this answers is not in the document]
FahadBSyed: So upon further investigation, it does look like we block until the timeout if we run across a commit that is not in `FINISHED`. I added code to handle that along with a test case.

src/server/pfs/server/driver.go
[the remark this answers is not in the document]
brendoncarroll: If it doesn't that's fine.

src/server/pfs/server/driver.go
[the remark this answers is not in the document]
brendoncarroll: I was thinking more of the case where we have 2^32 commits in an rpc.

src/server/pfs/cmds/cmds.go
brendoncarroll: That would be a breaking change.  We don't really have strong compatibility guarantees though, so it's probably fine for now.

armaanv: +1 on using a flag for this

armaanv (approved):
looks good!

brendoncarroll (approved):

PR #8607, opened 2023-02-21
Fix branch provenance test

brycemcanally:
This PR fixes two issues in the PFS BranchProvenance test. The first issue was referencing a for loop variable in a goroutine and the second was checking the wrong field (branch name rather than repo name) when validating the provenance / subvenance.


jrockway (approved):
LGTM

PR #8608, opened 2023-02-21
fix: apply current project to list datum from file

albscui:
This fixes the broken behaviour of list datum from file, where a user does not provid the project in the PPS Input section, and we wrongly infer the default project. Now, we use the current project in the local context to infer the project.

Solves CORE-1507, which unblocks the 2.5 GA release.


BOsterbuhr (approved):
It works as expected; thanks for the quick turnaround, @albscui!

PR #8609, opened 2023-02-21
[Jupyter] Restore python pachyderm in integration tests

smalyala:
When adding projects support to the mount extension, we switched to using a bash script to do some setup for integration tests since python pachyderm wasn't updated yet to be compatible with projects. We know revert those changes since python pachyderm supports projects.

Jira: [INT-859]

[INT-859]: https://pachyderm.atlassian.net/browse/INT-859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bbonenfant (approved):
Looks good to me. What was breaking without project support? It doesn't look like you are using any project related code

PR #8611, opened 2023-02-22
[PFS-61] Fix datum logger

brycemcanally:
This PR fixes an issue with the datum logger which caused it to accumulate fields for all of the datums in a datum set. This issue was caused by a subtle semantic change to the way new loggers are created, which causes the new loggers to inherit the fields from the old logger. In the previous implementation of this, the datum fields were just overwritten rather than inherited.


jrockway (approved):
Good find!

PR #8615, opened 2023-02-23
upgrade_test: fix test to run for all listed versions; reduce version list

jrockway:
This makes the upgrade tests actually run all of the "fromVersion"s that are listed; a bug was causing only the last version to run for every test.  Because this makes some old versions actually run, I've removed the ones that don't pass; the tests assume that the proxy exists, so versions of pachyderm that didn't have proxy have been removed.  I checked on Slack and no customers use these versions, so this just speeds up the test.

### DIFF
--- a/src/testing/deploy/upgrade_test.go
+++ b/src/testing/deploy/upgrade_test.go
@@ -31,13 +31,17 @@ const (
 var skip bool
 
 // runs the upgrade test from all versions specified in "fromVersions" against the local image
-func upgradeTest(suite *testing.T, ctx context.Context, fromVersions []string, preUpgrade func(*testing.T, *client.APIClient), postUpgrade func(*testing.T, *client.APIClient)) {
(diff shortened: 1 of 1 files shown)


src/testing/deploy/upgrade_test.go
```diff
@@ -31,13 +31,17 @@ const (
 var skip bool
 
 // runs the upgrade test from all versions specified in "fromVersions" against the local image
-func upgradeTest(suite *testing.T, ctx context.Context, fromVersions []string, preUpgrade func(*testing.T, *client.APIClient), postUpgrade func(*testing.T, *client.APIClient)) {
+func upgradeTest(suite *testing.T, ctx context.Context, parallelOK bool, fromVersions []string, preUpgrade func(*testing.T, *client.APIClient), postUpgrade func(*testing.T, *client.APIClient)) {
 	k := testutil.GetKubeClient(suite)
 	for _, from := range fromVersions {
+		from := from // suite.Run runs in a background goroutine if parallelOK is true
```
brendoncarroll, line 37: ✅

brendoncarroll (approved):
LGTM

PR #8618, opened 2023-02-24
Upgrade CI dependencies

jrockway:


jrockway:
@molinamelendezj Added you as a reviewer, but let's wait and see if the newest versions even install and work :)

jrockway:
OK, prometheus and gatekeeper manifests need to be upgraded.  I'll do that.

PR #8605, opened 2023-02-20
pachctl: add "misc" section to pachctl for random tools; add some network tools

jrockway:
This adds `pachctl misc`, and is hidden by default.

```
$ pachctl misc help
Miscellaneous utilities unrelated to Pachyderm itself.  These utilities can be removed or changed in minor releases; do not rely on them.

Usage:
  pachctl misc [command]

Available Commands:
  dial        Dials a network server and then disconnects.
  dns-lookup  Do a DNS lookup on a hostname.
  http-head   Make an HTTP HEAD request.

Flags:
  -h, --help   help for misc

Global Flags:
      --no-color   Turn off colors.
  -v, --verbose    Output verbose logs
```

`pachctl misc dial`: to dial a network address and print debugging info:

```
$ pachctl misc dial tcp 192.168.1.1:53 -v
2023-02-20T18:19:48.523-0500    DEBUG   cmds/cmds.go:31 created socket for connection   {"network": "tcp4", "address": "192.168.1.1:53", "rawConn": {}, "rawConn.type": "*net.rawConn", "fd": 9}
OK: 192.168.1.70:54588 -> 192.168.1.1:53
```

`pachctl misc dns-lookup`: to lookup a host using go's DNS stack:
```
$ pachctl misc dns-lookup dns.google. -v
2023-02-20T18:21:00.048-0500    DEBUG   DialDNS cmds/cmds.go:38 DialDNS: span start     {"network": "udp", "address": "8.8.8.8:53", "timeout": "4.999976399s"}
2023-02-20T18:21:00.048-0500    DEBUG   cmds/cmds.go:31 created socket for connection   {"network": "udp4", "address": "8.8.8.8:53", "rawConn": {}, "rawConn.type": "*net.rawConn", "fd": 9}
2023-02-20T18:21:00.048-0500    DEBUG   DialDNS cmds/cmds.go:38 DialDNS: span start     {"network": "udp", "address": "8.8.8.8:53", "timeout": "4.999996599s"}
2023-02-20T18:21:00.048-0500    DEBUG   DialDNS cmds/cmds.go:39 DialDNS: span finished ok       {"network": "udp", "address": "8.8.8.8:53", "spanDuration": "93.402µs"}
2023-02-20T18:21:00.048-0500    DEBUG   cmds/cmds.go:31 created socket for connection   {"network": "udp4", "address": "8.8.8.8:53", "rawConn": {}, "rawConn.type": "*net.rawConn", "fd": 10}
[...]

PR #8619, opened 2023-02-24
[2.5.x] Add some miscellaneous debugging tools to "pachctl misc" 

jrockway:
Backport of #8605

### DIFF
--- a/src/server/misc/cmds/cmds_1.20.go
+++ b/src/server/misc/cmds/cmds_1.20.go
@@ -1,0 +1,22 @@
+//go:build go1.20
+
+package cmds
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"syscall"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/log"
+	"go.uber.org/zap"
+)
+
+func setupControl(d *net.Dialer) {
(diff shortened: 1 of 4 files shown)


src/server/misc/cmds/cmds_1.20.go
```diff
@@ -0,0 +1,22 @@
...
+	"syscall"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/log"
+	"go.uber.org/zap"
+)
+
+func setupControl(d *net.Dialer) {
+	d.ControlContext = func(ctx context.Context, network, address string, c syscall.RawConn) error {
+		var fd uintptr
+		c.Control(func(f uintptr) { fd = f }) //nolint:errcheck
```
acohen4, line 18: why's this implementation 1.20 specific?

acohen4 (approved):
LGTM

src/server/misc/cmds/cmds_1.20.go
jrockway: Hahah brendon asked the same thing on the original PR.  ControlContext is newly-added in 1.20.  There is just Control in 1.19, which doesn't interact with our logger as nicely.  In retrospect, it's not worth 2 files just to print the FD number (but the FD number is nice when stracing, which maybe someone will do sometime, which is why I added it) 😂

PR #8623, opened 2023-02-27
[PFS-59] Compact Diff Filesets For New Commits

FahadBSyed:
As part of this work, I need to revisit whether we need to read the compacted diff fileset as part of the total fileset compaction (or if that should already be happening under the hood due to the structure of the compaction code).

### DIFF
--- a/src/server/pfs/server/master.go
+++ b/src/server/pfs/server/master.go
@@ -21,6 +21,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/pfsdb"
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/chunk"
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset"
+	"github.com/pachyderm/pachyderm/v2/src/internal/task"
 	"github.com/pachyderm/pachyderm/v2/src/internal/transactionenv/txncontext"
 	"github.com/pachyderm/pachyderm/v2/src/internal/watch"
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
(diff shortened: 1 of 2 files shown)


src/server/pfs/server/master.go
```diff
@@ -143,82 +144,106 @@ func (d *driver) finishCommits(ctx context.Context) (retErr error) {
...
 
 func (d *driver) finishRepoCommits(ctx context.Context, repoKey string) error {
-	err := d.commits.ReadOnly(ctx).WatchByIndexF(pfsdb.CommitsRepoIndex, repoKey, func(ev *watch.Event) error {
+	commitInfo := &pfs.CommitInfo{}
+	return d.commits.ReadOnly(ctx).WatchByIndexF(pfsdb.CommitsRepoIndex, repoKey, func(ev *watch.Event) error {
 		if ev.Type == watch.EventError {
 			return ev.Err
 		}
 		var key string
-		commitInfo := &pfs.CommitInfo{}
```
brycemcanally, line 151: Lets go ahead and revert this.

src/server/pfs/server/master.go
```diff
@@ -216,9 +215,48 @@ func (d *driver) finishRepoCommits(ctx context.Context, repoKey string) error {
 				log.Error(ctx, "error finishing commit", zap.Error(err), zap.Duration("retryAfter", d))
 				return nil
 			})
-		}, zap.Bool("finishing", true), log.Proto("commit", commit), zap.String("repo", key))
+		}, zap.Bool("finishing", true), log.Proto("commit", commitInfo.Commit), zap.String("repo", key))
 	}, watch.WithSort(col.SortByCreateRevision, col.SortAscend), watch.IgnoreDelete)
-	return errors.EnsureStack(err)
+}
+
+func (d *driver) compactDiffFileset(ctx context.Context, commit *pfs.Commit, compactor *compactor, renewer *fileset.Renewer, doer task.Doer) error {
```
brycemcanally, line 222: nit, would you change the ordering to ctx, compactor, taskDoer, renewer, commit. Ordering by scope generally reads better imo. Also, would you do this for the total file set method as well.

src/server/pfs/server/master.go
```diff
@@ -216,9 +215,48 @@ func (d *driver) finishRepoCommits(ctx context.Context, repoKey string) error {
...
+	diffId, err := d.commitStore.GetDiffFileSet(ctx, commit)
```
brycemcanally, line 223: nit, this reads a bit better as id being what you get here, and diffId being the result of compact.

src/server/pfs/server/master.go
```diff
@@ -216,9 +215,48 @@ func (d *driver) finishRepoCommits(ctx context.Context, repoKey string) error {
...
+	if err != nil {
+		if pfsserver.IsCommitNotFoundErr(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	if err := renewer.Add(ctx, *id); err != nil {
+		return nil, err
+	}
+	// Compact the commit. The totalID is a composed fileset that contains the diff fileset.
```
brycemcanally, line 251: I would remove this comment. I don't think there should be a consideration for the diff file set here, that should just be an optimization handled by the calling function. This is just a general way to compact the total file set of a commit.

src/server/pfs/server/master.go
```diff
@@ -216,9 +215,48 @@ func (d *driver) finishRepoCommits(ctx context.Context, repoKey string) error {
...
-		}, zap.Bool("finishing", true), log.Proto("commit", commit), zap.String("repo", key))
+		}, zap.Bool("finishing", true), log.Proto("commit", commitInfo.Commit), zap.String("repo", key))
 	}, watch.WithSort(col.SortByCreateRevision, col.SortAscend), watch.IgnoreDelete)
-	return errors.EnsureStack(err)
+}
+
+func (d *driver) compactDiffFileset(ctx context.Context, commit *pfs.Commit, compactor *compactor, renewer *fileset.Renewer, doer task.Doer) error {
+	diffId, err := d.commitStore.GetDiffFileSet(ctx, commit)
+	if err != nil {
+		if pfsserver.IsCommitNotFoundErr(err) {
```
brycemcanally, line 225: This should be handled by the caller, we want to propagate the error here. We should do the same in the total method.

src/server/pfs/server/master.go
```diff
@@ -143,55 +145,52 @@ func (d *driver) finishCommits(ctx context.Context) (retErr error) {
...
-					if err := renewer.Add(ctx, *id); err != nil {
+					// Compacting the diff before getting the total allows us to compose the
+					// total fileset so that it includes the compacted diff.
+					if err := log.LogStep(ctx, "compactDiffFileset", func(ctx context.Context) error {
+						return d.compactDiffFileset(ctx, commit, compactor, renewer, taskDoer)
+					}); err != nil {
 						return err
 					}
-					// Compact the commit.
+					// The totalID is a composed fileset that contains the compacted diff fileset.
```
brycemcanally, line 186: I would remove this comment and change the log name to `compactTotalFileSet`

src/server/pfs/server/master.go
```diff
@@ -143,55 +145,52 @@ func (d *driver) finishCommits(ctx context.Context) (retErr error) {
...
-					}
-					return err
-				}
 				compactor := newCompactor(d.storage, d.env.StorageConfig.StorageCompactionMaxFanIn)
 				taskDoer := d.env.TaskService.NewDoer(StorageTaskNamespace, commit.ID, cache)
 				var totalId *fileset.ID
 				start := time.Now()
 				if err := d.storage.WithRenewer(ctx, defaultTTL, func(ctx context.Context, renewer *fileset.Renewer) error {
-					if err := renewer.Add(ctx, *id); err != nil {
+					// Compacting the diff before getting the total allows us to compose the
```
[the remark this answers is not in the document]
FahadBSyed: Gotcha. I think I'm used to trying it in code which is why I wrote it as one word.

src/server/pfs/server/master.go
```diff
@@ -143,55 +145,57 @@ func (d *driver) finishCommits(ctx context.Context) (retErr error) {
...
+					// total file set so that it includes the compacted diff.
+					if err := log.LogStep(ctx, "compactDiffFileset", func(ctx context.Context) error {
```
brycemcanally, line 181: Fileset -> FileSet here and below.

src/server/pfs/server/master.go
```diff
@@ -143,55 +145,57 @@ func (d *driver) finishCommits(ctx context.Context) (retErr error) {
...
+						return d.compactDiffFileset(ctx, compactor, taskDoer, renewer, commit)
+					}); err != nil {
+						if pfsserver.IsCommitNotFoundErr(err) {
+							return nil
+						}
 						return err
 					}
-					// Compact the commit.
-					return log.LogStep(ctx, "compactCommit", func(ctx context.Context) error {
+					return log.LogStep(ctx, "compactTotalFileset", func(ctx context.Context) error {
```
brycemcanally, line 189: same as above.

brycemcanally (approved):
lgtm

PR #8628, opened 2023-03-06
[CORE-1513] Correct project-already-exists error message

robert-uhl:
Turns out that the error message came from the collection package and indicated that the item in the “projects” table already existed.


src/server/pfs/server/driver.go
```diff
@@ -484,10 +484,18 @@ func (d *driver) createProject(ctx context.Context, req *pfs.CreateProjectReques
 		} else if !errors.Is(err, auth.ErrNotActivated) {
 			return errors.Wrap(err, "could not get caller's username")
 		}
-		return errors.EnsureStack(projects.Create(pfsdb.ProjectKey(req.Project), &pfs.ProjectInfo{
+		if err := projects.Create(pfsdb.ProjectKey(req.Project), &pfs.ProjectInfo{
 			Project:     req.Project,
 			Description: req.Description,
-		}))
+		}); err != nil {
+			if errors.Is(err, col.ErrExists{}) {
```
albscui, line 491: Should this use `errors.As` instead? In the case where a function returns a none default `ErrExists` with extra values filled in, `errors.Is` would return false.
```suggestion
			if errors.As(err, &col.ErrExists{}) {
```

albscui (commented):
Maybe add a test?

src/server/pfs/server/driver.go
robert-uhl: You’re right, although it happens to work due to https://github.com/pachyderm/pachyderm/blob/master/src/internal/collection/errors.go#L50-L53.

Changed.

robert-uhl:
> Maybe add a test?

In this case, it’s just about the English which the user sees, so I figure not such a big deal.

albscui (approved):
LGTM!

PR #8631, opened 2023-03-06
[CORE-1500] Return error instead of deleting a project containing repos

robert-uhl:


jrockway (approved):
LGTM.

We discussed on Slack a little bit... I wouldn't go to this much effort to make the error message nice.  Just multierr.AppendInto for each repo existence, return err.  An errors.SetGRPCStatus(&err, code) would be a nice utility function, however.

PR #8637, opened 2023-03-07
[CORE-1505] Add a warning message when deleting the current project

robert-uhl:


src/server/pfs/cmds/cmds.go
```diff
@@ -1212,7 +1212,13 @@ Projects contain pachyderm objects such as Repos and Pipelines.`,
 					Project: &pfs.Project{Name: args[0]},
 					Force:   force,
 				})
-			return grpcutil.ScrubGRPC(err)
+			if err != nil {
+				return grpcutil.ScrubGRPC(err)
+			}
+			if args[0] == pachCtx.Project {
+				fmt.Fprintf(os.Stderr, "warning: deleted current project %s; update context by running:\n   pachctl config update context --project PROJECT\n", pachCtx.Project)
```
albscui, line 1219: nit: a small worry is that if we change the command for updating the context's project, we need to remember to update it here as well.

albscui (approved):
LGTM!

src/server/pfs/cmds/cmds.go
robert-uhl: True, but I don’t think that there’s a way to autogenerate the string — is there?

PR #8639, opened 2023-03-08
Djanicek/core 1481/debug fix

djanicekpach:
**There are 2 issues** with the upgrade load tests(there are more, but keeping the scope to these two EOF errors) 

1. The auth interceptor fails without a registered Enterprise server
2. Pachw can connect to the old postgres if the tests start running before the helm upgrade is complete. Any DB connections from pachw then fail if this happens.

**To solve this we**

1.  register the enterprise server when we buildAndRun pachw
2. We wait for postgres to be ready before running the load test. This follows what we do with other pods on start up and potentially also helps stabilize other upgrade tests.

**Considered but not done:**

I looked into adding retries into the load tests or upgrade tests somewhere to address issue 2

- It is difficult to restart in the upgrade_test itself because the load tests don't fail atomically(they start a commit and potentially make repos), so a retry from the same state never works.
- Retrying in the load test means we have to restart the worker process to get a new DB client, which is slower and I'm worried about the effects of adding retries to a test that is designed to measure test duration, and that is not primarily designed for the upgrade test. I did get this working by using panic() in the pfsload worker on error, but simply waiting for postgres seems like the cleanest fix to me at the moment.

**Idiosyncrasies in the PR**

- Wait seconds is now 0. It was 10, probably because someone notice the DB errors, but didn't know what was up. Now that we wait correctly we can toss the artificial wait.
- Cleanup after is false. This lets the CI pipeline actually dump the k8s logs and pods on failure which makes debugging easier, and it does not appear to have any functional impact(and it shouldn't based on code flow). Better debugging is nice.
- I synced registerAuthServer a little bit with pachd, but stopped short of setting public=true since it wasn't necessary to fix issue 1

### DIFF
--- a/src/internal/minikubetestenv/deploy.go
+++ b/src/internal/minikubetestenv/deploy.go
@@ -410,6 +410,21 @@ func waitForPgbouncer(t testing.TB, ctx context.Context, kubeClient *kube.Client
 	}, backoff.RetryEvery(5*time.Second).For(5*time.Minute)))
 }
 
+func waitForPostgres(t testing.TB, ctx context.Context, kubeClient *kube.Clientset, namespace string) {
+	require.NoError(t, backoff.Retry(func() error {
+		pbs, err := kubeClient.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{LabelSelector: "app.kubernetes.io/name=postgresql"})
+		if err != nil {
+			return errors.Wrap(err, "error on pod list")
+		}
+		for _, p := range pbs.Items {
+			if p.Status.Phase == v1.PodRunning && p.Status.ContainerStatuses[0].Ready && len(pbs.Items) == 1 {
+				return nil
+			}
+		}
+		return errors.Errorf("deployment in progress")
+	}, backoff.RetryEvery(5*time.Second).For(5*time.Minute)))
+}
+
 func pachClient(t testing.TB, pachAddress *grpcutil.PachdAddress, authUser, namespace string, certpool *x509.CertPool) *client.APIClient {
 	var c *client.APIClient
 	// retry connecting if it doesn't immediately work
@@ -544,6 +559,7 @@ func putRelease(t testing.TB, ctx context.Context, namespace string, kubeClient 
 		waitForLoki(t, pachAddress.Host, int(pachAddress.Port)+9)
 	}
 	waitForPgbouncer(t, ctx, kubeClient, namespace)
+	waitForPostgres(t, ctx, kubeClient, namespace)
 	if opts.WaitSeconds > 0 {
 		time.Sleep(time.Duration(opts.WaitSeconds) * time.Second)
 	}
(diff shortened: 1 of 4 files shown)


src/internal/minikubetestenv/deploy.go
```diff
@@ -396,8 +396,16 @@ func waitForLoki(t testing.TB, lokiHost string, lokiPort int) {
...
 
 func waitForPgbouncer(t testing.TB, ctx context.Context, kubeClient *kube.Clientset, namespace string) {
+	waitForLabeledPod(t, ctx, kubeClient, namespace, "app=pg-bouncer")
+}
+
+func waitForPostgres(t testing.TB, ctx context.Context, kubeClient *kube.Clientset, namespace string) {
+	waitForLabeledPod(t, ctx, kubeClient, namespace, "app.kubernetes.io/name=postgresql")
+}
+
+func waitForLabeledPod(t testing.TB, ctx context.Context, kubeClient *kube.Clientset, namespace string, label string) {
```
jrockway, line 406: btw, it would be nice if this t.Log'd every retry; which label selector it's waiting on, and the status of the matches (pod phase if != running, which container is not ready, etc.).

src/internal/pachd/pachw.go
```diff
@@ -59,6 +85,7 @@ func (pachwb *pachwBuilder) buildAndRun(ctx context.Context) error {
 		pachwb.initInternalServer,
 		pachwb.registerAuthServer,
 		pachwb.registerPFSServer, //PFS seems to need a non-nil auth server.
+		pachwb.registerEnterpriseServer,
```
jrockway, line 88: The other pachs register enterprise before auth and I think we should do that here.  auth also requires an identity server, but I guess we don't call methods that actually dereference the identity server?  dunno.

jrockway (approved):
Looks good.  Some comments.

src/internal/minikubetestenv/deploy.go
```diff
@@ -396,8 +396,16 @@ func waitForLoki(t testing.TB, lokiHost string, lokiPort int) {
 }
 
 func waitForPgbouncer(t testing.TB, ctx context.Context, kubeClient *kube.Clientset, namespace string) {
+	waitForLabeledPod(t, ctx, kubeClient, namespace, "app=pg-bouncer")
+}
+
+func waitForPostgres(t testing.TB, ctx context.Context, kubeClient *kube.Clientset, namespace string) {
```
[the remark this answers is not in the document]
djanicekpach: Yeah, I was aware of this potential but I think

- all of the other waitFor functions are affected by this as well, so it's not a new issue. I made a ticket to address this holistically in the future https://pachyderm.atlassian.net/browse/CORE-1536
- postgres goes into terminating very quickly, it just takes a while to spin down and back up. Since we check postgres last(after loki, pachd and pg-bouncer) the chance that postgres isn't at least going down is really small. Watching locally at least, we aren't even close to hitting this race condition.

I could attempt to find a way to get the `controller-revision-hash` and confirm it's good, but I think it's pragmatic to get this in and handle that in bulk with CORE-1536.

src/internal/minikubetestenv/deploy.go
djanicekpach: Edit: it looks like all this info is already logged in require.noErrorWithinTRetry, so I will refrain from doing it a second time.

PR #8642, opened 2023-03-09
retry during mock IDP flake

djanicekpach:
One of the most common deploy test failures is the mock IDP returning a 503. [Example](https://app.circleci.com/pipelines/github/pachyderm/pachyderm/18339/workflows/f9aa266f-26df-4903-b71a-4c1cf4b25d70/jobs/292746/tests#failed-test-0)

Testing suggest this generally works after a retry. Although it's hard to reproduce that's what we'd expect from a 503. To help make the tests pass more consistently, we are retrying the login. 

We are retrying the whole login and not just the part of it that returns 503 because different parts of the workflow can cause the error. For example I have seen a bad login url returned on the initial GET, which leads to an error on the form POST.

Here's an example of the fix working 
https://app.circleci.com/pipelines/github/pachyderm/pachyderm/18344/workflows/1ec5467b-96b6-4d5f-917f-6e7235cb7900/jobs/292869

> 2023-03-09T19:34:53     deploy.go:444: Success connecting to pachd on port: 30660, in namespace: test-cluster-1
2023-03-09T19:35:06     require.go:472: retryable: Could not retrieve the mock login page. Expected: 200, Recieved: 503
2023-03-09T19:35:11 --- PASS: TestInstallAndUpgradeEnterpriseWithEnv (163.95s)

### DIFF
--- a/src/testing/deploy/deploy_test.go
+++ b/src/testing/deploy/deploy_test.go
@@ -142,7 +142,7 @@ func mockIDPLogin(t testing.TB, c *client.APIClient) {
 		expectedResponse := "^(You are now logged in).*$"
 		matches, err := regexp.Match(expectedResponse, postBody)
 		require.NoError(t, err, "Regex matching failed") // a regex compile error should be deterministic, so fail immediately
-		if matches {
+		if !matches {
 			return errors.EnsureStack(errors.Errorf("Recieved an unexpected response body from mock IDP login form. Expected: %s, Recieved: %s", expectedResponse, string(postBody)))
 		}
 
--- a/src/testing/deploy/deploy_test.go
+++ b/src/testing/deploy/deploy_test.go
@@ -121,7 +121,7 @@ func mockIDPLogin(t testing.TB, c *client.APIClient) {
 			return errors.EnsureStack(err)
 		}
 		if 200 != getResp.StatusCode {
-			errors.EnsureStack(errors.Errorf("Could not retrieve the mock login page. Expected: 200, Recieved: %d", getResp.StatusCode))
+			return errors.EnsureStack(errors.Errorf("Could not retrieve the mock login page. Expected: 200, Recieved: %d", getResp.StatusCode))
 		}
 
 		vals := make(url.Values)
@@ -129,7 +129,7 @@ func mockIDPLogin(t testing.TB, c *client.APIClient) {
 		vals.Add("password", "password")
 		postResp, err := hc.PostForm(getResp.Request.URL.String(), vals)
 		if err != nil {
-			return err
+			return errors.EnsureStack(err)
 		}
 		if 200 != postResp.StatusCode {
 			return errors.EnsureStack(errors.Errorf("POST to perform mock login failed. Expected: 200, Recieved: %d", postResp.StatusCode))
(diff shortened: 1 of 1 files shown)


src/testing/deploy/deploy_test.go
```diff
@@ -40,7 +43,7 @@ func TestInstallAndUpgradeEnterpriseWithEnv(t *testing.T) {
 	c.SetAuthToken("")
 	mockIDPLogin(t, c)
 	// Test Upgrade
-	opts.CleanupAfter = true
+	opts.CleanupAfter = false
```
molinamelendezj, line 46: ive seen this get toggled a lot. should we clean up after test?

molinamelendezj (approved):

src/testing/deploy/deploy_test.go
djanicekpach: I reverted it because I haven't tested enough, but I'm becoming increasingly convinced that `cleanupAfter=true` doesn't make a functional difference and really only makes the pipelines harder to debug because the `dump k8s info` step can't dump the cluster state.

src/testing/deploy/deploy_test.go
```diff
@@ -102,36 +105,62 @@ func TestEnterpriseServerMember(t *testing.T) {
...
+		}
+		postBody, err := io.ReadAll(postResp.Body)
+		if err != nil {
+			return errors.EnsureStack(errors.Errorf("Could not read login form POST response: %s", err.Error()))
+		}
+		// There is a login failure case in which the response returned is a redirect back to the login page that returns 200, but does not log in
+		expectedResponse := "^(You are now logged in).*$"
+		matches, err := regexp.Match(expectedResponse, postBody)
+		require.NoError(t, err, "Regex matching failed") // a regex compile error should be deterministic, so fail immediately
+		if !matches {
```
jrockway, line 145: After you switch to HasPrefix, do this all in one line (`if !strings.HasPrefix(...) { error }`)

jrockway (approved):
This will be nice.

I made some style comments, however.

src/testing/deploy/deploy_test.go
```diff
@@ -102,36 +105,62 @@ func TestEnterpriseServerMember(t *testing.T) {
...
+			return errors.EnsureStack(errors.Errorf("Could not retrieve the mock login page. Expected: 200, Recieved: %d", getResp.StatusCode))
+		}
 
-	vals := make(url.Values)
-	vals.Add("login", "admin")
-	vals.Add("password", "password")
+		vals := make(url.Values)
+		vals.Add("login", "admin")
+		vals.Add("password", "password")
+		postResp, err := hc.PostForm(getResp.Request.URL.String(), vals)
```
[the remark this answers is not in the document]
djanicekpach: I checked and these are different URLs. The GET gets the test login url specifically, so I'll leave this for now.

src/testing/deploy/deploy_test.go
```diff
@@ -102,36 +105,62 @@ func TestEnterpriseServerMember(t *testing.T) {
...
 
-	// Get the initial URL from the grpc, which should point to the dex login page
-	getResp, err := hc.Get(loginInfo.LoginURL)
-	require.NoError(t, err)
-	require.Equal(t, 200, getResp.StatusCode, "Mock login could not retrieve the login page.")
+		// Get the initial URL from the grpc, which should point to the dex login page
+		getResp, err := hc.Get(loginInfo.LoginURL)
+		if err != nil {
+			return errors.EnsureStack(err)
+		}
```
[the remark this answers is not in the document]
djanicekpach: Style changes are in a new PR since this is merged already. https://github.com/pachyderm/pachyderm/pull/8647

PR #8643, opened 2023-03-10
storage/kv: port obj.Client utilities

brendoncarroll:
We would like to eventually migrate away from our `internal/obj` package in favor of the [go-cloud sdk](https://pkg.go.dev/gocloud.dev) interfaces, which are maintained by people-who-aren't-us.  We already use their clients for imports and export done with PutFileURL and GetFileURL, but we would also like to use them for the main storage backend.  The original motivation for this was better object metadata when listing the contents of a bucket; it was easier to switch to this for import/export than to add that functionality to our interfaces for all the implementations.

This PR ports a lot of the extra functionality that we built adhering to the `obj.Client` interface to the `kv.Store` interface.   The `kv.Store` interface is used throughout the storage layer since we only deal in chunks.

After this PR the last thing to do will be to replace the object client adapter in the `kv` package with one for the go-cloud SDK.


src/internal/storage/kv/filesystem.go
```diff
@@ -0,0 +1,206 @@
...
+		return err
+	}
+	err := os.Remove(s.finalPathFor(key))
+	if os.IsNotExist(err) {
+		err = nil
+	}
+	return err
+}
+
+func (c *FSStore) stagingPathFor(k []byte) string {
```
FahadBSyed, line 92: [nit] the method receiver here is `c` but `s` is used elsewhere

src/internal/storage/kv/filesystem.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (s *FSStore) ensureInit(ctx context.Context) (err error) {
+	if s.initDone.Load() {
+		return nil
+	}
+	if err := s.initSem.Acquire(ctx, 1); err != nil {
+		return errors.EnsureStack(err)
+	}
+	defer s.initSem.Release(1)
+	if s.initDone.Load() {
```
FahadBSyed, line 148: why is this checked twice?

src/internal/storage/kv/kv.go
```diff
@@ -1,32 +1,103 @@
...
 // Store is a key-value store
 type Store interface {
-	GetPut
-	Delete(ctx context.Context, key []byte) error
+	Getter
+	Putter
+	Deleter
 	Exists(ctx context.Context, key []byte) (bool, error)
-	Walk(ctx context.Context, prefix []byte, cb func(key []byte) error) error
+
```
FahadBSyed, line 49: Could you use `KeyIterable` instead here?

src/internal/storage/kv/filesystem.go
brendoncarroll: To prevent us from initializing twice.

- `initDone` tracks whether or not we are done initializing.  (We should only initialize once).
- `initSem` ensures that only one go-routine is trying to initialize at a given time.

It's designed so that if initialization fails, it could succeed in a subsequent call rather than being failed forever.

src/internal/storage/kv/kv.go
brendoncarroll: Yes, good call.

src/internal/storage/kv/lru_cache.go
```diff
@@ -0,0 +1,107 @@
...
+func (c *LRUCache) NewKeyIterator(span Span) stream.Iterator[[]byte] {
+	return c.slow.NewKeyIterator(span)
+}
+
+func (c *LRUCache) putFast(ctx context.Context, key, value []byte) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.ctx = ctx
+	defer func() { c.ctx = nil }()
+	if err := c.fast.Put(ctx, key, value); err != nil {
```
brendoncarroll, line 87: I think we probably wan't to do this put before we get the lock.  Or maybe after, but definitely not with it.

src/internal/storage/kv/lru_cache.go
```diff
@@ -0,0 +1,107 @@
...
+		log.Error(ctx, "kv.LRUCache could not put in fast layer", zap.Error(err))
+	}
+	return n, nil
+}
+
+func (c *LRUCache) Put(ctx context.Context, key, value []byte) error {
+	return c.slow.Put(ctx, key, value)
+}
+
+func (c *LRUCache) Exists(ctx context.Context, key []byte) (bool, error) {
```
brendoncarroll, line 58: I think this is the correct behavior for an Exists method, but we need to double check that we use postgres as the source of truth for whether we can skip an upload.

brendoncarroll: We do check with postgres in `chunk/client.go`

src/internal/storage/kv/pool.go
```diff
@@ -0,0 +1,44 @@
...
+		return err
+	}
+	return cb((*buf)[:n])
+}
+
+func (p *Pool) acquire() *[]byte {
+	return (p.pool.Get().(*[]byte))
+}
+
+func (p *Pool) release(x *[]byte) {
```
FahadBSyed, line 42: release is more straight forward semantically than `put` (should really be PutBack in this case), so thanks for doing this.

src/internal/storage/chunk/storage.go
```diff
@@ -81,9 +82,12 @@ func (s *Storage) PrefetchData(ctx context.Context, dataRef *DataRef) error {
 
 // List lists all of the chunks in object storage.
 func (s *Storage) List(ctx context.Context, cb func(id ID) error) error {
-	return errors.EnsureStack(s.store.Walk(ctx, nil, func(key []byte) error {
-		return cb(ID(key))
-	}))
+	it := s.store.NewKeyIterator(kv.Span{})
+	return stream.ForEach(ctx, it, func(key []byte) error {
+		var id ID
+		copy(id[:], key)
```
FahadBSyed, line 88: Is the key the chunk itself? I am assuming that the key is an index and the value is the chunk when reading this. Do we want to be executing the callback against the key or the value?

src/internal/storage/kv/filesystem.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	if len(value) > s.maxValueSize {
+		return errors.Errorf("max value size %d exceeded. len(value)=%d", s.maxKeySize, len(value))
+	}
+	if err := s.ensureInit(ctx); err != nil {
+		return err
+	}
+	staging := s.stagingPathFor(key)
+	final := s.finalPathFor(key)
+	defer s.cleanupFile(ctx, &retErr, staging)
```
FahadBSyed, line 54: This is a clever way of handling an error from a deferred function, but won't the error get overridden if os.WriteFile or os.Rename fail? If we care enough about those errors, we could use the `multierr` package to handle combining multiple errors and append the error from `s.cleanupFile` onto a potential error.

src/internal/storage/kv/filesystem.go
```diff
@@ -0,0 +1,206 @@
...
+				return errors.EnsureStack(err)
+			}
+			keys = append(keys, key)
+		}
+		it.keys = keys
+	}
+	for it.pos < len(it.keys) {
+		k := it.keys[it.pos]
+		if it.span.Contains(k) {
+			*dst = append((*dst)[:0], k...)
```
FahadBSyed, line 130: This looks like a complicated way of copying the contents of `k` into `*dst` by value. It may be more readable to do something like the following:
``` Go
*dst = make([]byte, len(key))
copy(*dst, key)
```

src/internal/storage/kv/filesystem.go
FahadBSyed: Nevermind, I see the way this is handled is by logging an error if `retErr` is not nil.

src/internal/storage/kv/lru_cache.go
```diff
@@ -0,0 +1,111 @@
...
+	"go.uber.org/zap"
+)
+
+var _ Store = &LRUCache{}
+
+type LRUCache struct {
+	slow, fast Store
+
+	mu      sync.RWMutex
+	cache   simplelru.LRU[string, struct{}]
```
FahadBSyed, line 19: I'm a little confused on the purpose of `cache` here. I see we already have a `slow` and `fast` store. The LRU process when applied to slow and fast makes sense. But it looks like we are also updating the internal cache when we do a put and delete. I assume that the simple lru cache handles updating the eviction logic for us, but if we have cache, why do we need fast?

FahadBSyed (approved):
Left some non-blocking comments with some minor suggestions. Otherwise, looks good to me

src/internal/storage/kv/filesystem.go
brendoncarroll: ^That will allocate a new buffer for dst, rather than reusing the existing buffer.

src/internal/storage/kv/kv.go
```diff
@@ -1,32 +1,102 @@
...
+func SpanFromPrefix(prefix []byte) Span {
+	return Span{
+		Begin: prefix,
+		End:   PrefixEnd(prefix),
+	}
+}
+
+// PrefixEnd returns the key > all the keys with prefix p, but < any other key
+func PrefixEnd(prefix []byte) []byte {
```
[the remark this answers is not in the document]
brendoncarroll: If you look at the `Span` type in the other file, it has `Begin` and `End` fields which define a range including the lower bound, but excluding the upper bound.  The first key when listing a prefix is just the prefix itself, so that's what you set `Begin` to.  The key just after the end of the prefix is what you set `End` to.  `PrefixEnd` gives you the `End` key for the prefix that you pass in.

src/internal/storage/chunk/storage.go
brendoncarroll: Yeah you're right.  It's not the chunk id, it's the whole key (which includes the chunk prefix).  I need to fix this.  We have to parse the key and discard the chunk part.

PR #8648, opened 2023-03-11
Add a metrics API

jrockway:
This adds the new metrics API.  The gist is that it's pretty simple; at any time, anywhere you have a context, you can `Set`, `Inc` (rement), or `Sample` some data.  These end up being structured log messages.

You can make it more complicated.  Sometimes you don't really want to generate as many log lines as the changing data would require, so there are aggregations; `Gauge`, `Counter`, and `Delta`.  You can set a certain metric to be aggregated with `pctx`; `pctx.Child(ctx, ..., pctx.WithGauge("name"))`.  Once the aggregation is set up, any calls to `Set(ctx, "name", ...)` get aggregated according to the settings passed to WithGauge, defaulting to outputting the metric every 10 seconds (at write time, no background goro).  So if you call Set a million times on `"name"` in 10 seconds, you only get 1 log line.  Additionally, the metric is flushed when the underlying context expires, if it has changed since last being printed to the logs.

Gauge + Set is for absolute values ("amount of disk left on device").  Counter/Inc is for cases where you only know the 1st derivative of the absolute value, like you read some bytes and can Inc "bytes_read" by that many bytes.  Sample is for sampling data; probably to be visualized as a histogram.  (Sample doesn't have a matching aggregation yet; log sampling handles reducing the volume of frequently-written metrics.)  Finally, there is Delta/Set which is a Gauge, but that is output incrementally.  The first Set of a Delta records the absolute value in the logs; subsequent calls to Set just print the difference.  I added these for monitoring user code and values like cpu seconds used; you only know the absolute value, but you want to print a log line that says "10 seconds of CPU used since the last log line".  

There are probably other types like Delta lurking in the metrics world that we want to implement.  We'll do it when it comes up.

I tested the API by adding some OS level metrics for the pachyderm suite (`self_`) and for user code.  Memory, CPU, and I/O are now monitored, and thanks to how Delta works, ends up not being very noisy.  As Pachyderm or the user code allocates memory, uses CPU time, or does IO, progress information is incrementally reported.

Left to do is to convert over some of our old monitoring; WriterWithThroughput and that stuff.  Will do that in this PR.


src/internal/m/m.go
```diff
@@ -0,0 +1,441 @@
...
+// Done, a log line is also emitted.  From an analysis standpoint, nothing changes; the emitted
+// aggregated metrics are indistinguishable from immediately emitted metrics.
+//
+// Aggregated metrics can only be one type of metric with one type of value; if you create a counter
+// named tx_bytes and then call Set on it, no aggregation will be done on that data.  (Inc calls
+// continue to be aggregated; it doesn't "break" the aggregation to accidentally call the wrong
+// value-changing function.)  Similarly, the type of the value is set at registration time; calls
+// writing values of a different type will not be aggregated, but again, do not break any ongoing
+// aggregation.
+package m
```
brendoncarroll, line 69: I appreciate the value of this  being short because it is used often, but I think it goes against the guidelines of not choosing package names, which would be good, short variable names.

brendoncarroll:
## Client APIs
The client API (`Set`, `Inc`) looks good.

At some point we had talked about having units as a required parameter, but I don't feel strongly about that, and the metric names will probably be descriptive enough.

## MetricSpace
The main suggestion I have for the model is a concept of a `MetricSpace` which is required to collect metrics.  I'm thinking something like
```go
type MetricSpace1 interface {
  // Get or Create a Gauge in the space
  Gauge(k string) Gauge
  // Get or Create a Counter in the space
  Counter(k string) Counter
}
```

or this
```go
type MetricSpace2 interface {
  // Increments the metric called k in the  space.
  Inc(k string, v Monoid)
  // Sets the gauge called k in the space
  Set(k string, v any)
}
```

To perform an operation on a metric, first we get the `MetricSpace` out of the context, if it doesn't exist, then we do nothing.   If it does, we get or create the metric in that space, and perform the operation.  If the caller wants to batch metrics they can create a batching space

There are a few things, which that abstraction would accomplish:

### Explicit Collection
If there is an instrumented function, which increments a bunch of metrics, and it's called by a function that doesn't know anything about the metrics package, the called function should not have any detectable difference in behavior.  It shouldn't add noise to the logs, and it shouldn't interact with any metrics singletons.  In order to collect metrics, there should be an explicit step where some top level metrics object is setup and attached to the context.  That could be the MetricSpace

### Batching/Flushing
Overriding this collector per context fits in nicely with the batching and flushing.  It looks like this code is already doing something like that.

### Backend implementation
Right now we are turning metrics into logs, but we should be able to use prometheus with the same API.  Right now the top level metric space would be a `LogBackedMetricSpace` or `LoggingSpace` or a better name.   At some point it could also be prometheus, or some other metrics solution.  That should be the only connection to the logging system.

 ### Namespacing
Namespacing can be accomplished with a MetricSpace which prefixes all it's metric names.  `pctx.Child` would use a `PrefixMetricSpace`.

### Point of extension
If someone wants a metric sampled or aggregated in a bizarre way, they could write a MetricSpace which has buckets, or batching with a specific watermarking strategy, or whatever else.  They add it to the context, and everything they call with it is collected how they like, then it is passed up the chain to the parent MetricSpace.


## What Happens to Prometheus?
There are a few places in the code where we have prometheus singletons that we interact with to collect metrics.  `internal/dbutil` and `internal/obj` come to mind.  I hate singletons, and I would like to get rid of them, but there are a few things to consider:
  1. Is the plan to stop using prometheus entirely?
  2. Will anyone miss that data showing up over HTTP; is anyone scraping it now?

src/internal/m/m.go
```diff
@@ -0,0 +1,441 @@
+// Package m implements lightweight metrics.
+
+// A metric is a name, a set of key/value pairs (set by
+// the underlying logger), and a value.  A value can be an absolute value ("gauge"), an incremental
+// value ("counter"), or a set of samples (like "histogram", though lossless in this
+// implementation).  Metrics are most useful when additional code analyzes the entire log of a
+// program's run; this is called "analysis code" or "anaylsis software" throughout the
+// documentation.
+//
+// Values are stored by writing the set of operations to build the value to the logs.  For example,
```
brendoncarroll, line 10: See my comment about a `MetricSpace`  I think this section describes a particular implementation of that and doesn't need to be included in the mental model of the metrics system.

src/server/cmd/pachd/main.go
```diff
@@ -37,6 +38,8 @@ func main() {
 	logMode := func(mode string) {
 		log.Info(ctx, "pachd: starting", zap.String("mode", mode))
 	}
+	go proc.MonitorSelf(ctx)
+
```
brendoncarroll, line 42: This is where I think we should create the logging MetricSpace and attach it to the context.  I actually think we should call `zap.NewProduction` here too, but that's a discussion for another day.

brendoncarroll (approved):
LGTM

src/internal/m/m.go
```diff
@@ -0,0 +1,441 @@
...
+// the set of key=value metadata; a metric foo with field foo=bar is logically a different value
+// from the metric foo with field foo=baz.  The metadata is defined by the fields applied to the
+// underlying logger.  For example, if you declare a counter `tx_bytes` on each HTTP request you
+// handle, each time the count of transmitted bytes is updated, the log message will include fields
+// inherited from the default HTTP logger, which includes a per-request ID.  Therefore, analysis
+// code can calculate a tx_byte count for every request handled by the system by observing the value
+// of the x-request-id field on log lines matching the tx_byte counter format.  And, of course, it
+// can ignore the extra fields and add everything up to show a program-wide count of bytes
+// transmitted.
+//
```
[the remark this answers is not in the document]
jrockway: Good point.  I added documentation about not needing to ambiguate this because pctx.Child is the place to do it.  I also fleshed out the documentation on Child to describe how it works, and what naming conventions should be used there.

PR #8650, opened 2023-03-14
[CORE-1475] Add prompts for deleting repos & pipelines when deleting project

robert-uhl:


albscui (commented):
Shouldn't the order of operations be 1) delete all pipelines 2) delete all repos?

Also why can't we do this on the server side? This implementation means we also need to mirror it in python-pachyderm.

robert-uhl:
> Shouldn't the order of operations be 1) delete all pipelines 2) delete all repos?

It only deletes the user repos, then lets pipeline deletion delete the system repos.

> Also why can't we do this on the server side? This implementation means we also need to mirror it in python-pachyderm.

How would the server prompt? Yeah, this could be a fancier stream-based protocol where the server would send a message for each action it’s about to take and the client confirms or forbids the action, but … that’s signicantly more complex and requires a fair amount of API changes.

robert-uhl:
Yup. It tries to do the obvious thing when deleting a project, but no more than that. If there’s still something to be untangled … don’t do anything (so the system remains in a good state) and let the user determine the right course of action.

Better, I think, than for project deletion to start deleting things in other projects.

albscui:
Gotcha, maybe add a pachctl test?

albscui (approved):

PR #8652, opened 2023-03-15
[CORE-1494] Support sidecar_resource_requests in pipeline spec

armaanv:


src/server/pps/pretty/pretty.go
```diff
@@ -286,6 +286,9 @@ ParallelismSpec: {{.Details.ParallelismSpec}}
 {{ if .Details.SidecarResourceLimits }}SidecarResourceLimits:
   CPU: {{ .Details.SidecarResourceLimits.Cpu }}
   Memory: {{ .Details.SidecarResourceLimits.Memory }} {{end}}
+{{ if .Details.SidecarResourceRequests }}SidecarResourceRequests:
```
jrockway, line 289: We should print requests before limits, to be consistent with the non-sidecar stuff above.

jrockway (commented):
There should be a way to set the defaults from the helm chart.

armaanv:
> There should be a way to set the defaults from the helm chart.

yup, meant to add this in the original PR. It's in the latest commit

jrockway (approved):
Very nice!

PR #8655, opened 2023-03-15
[Python] datum batching in the Python SDK

bbonenfant:
Implements a `batch_datums` extension method. This implementation follows what `pachctl` does and attempts to establish a gRPC channel with the worker without the user explicitly doing so. A side-effect of this is that you can call the `batch_datums` method on an uninstantiated `pachyderm_sdk.Client` object. The `_classproperty.py` file is to enable this.

Example:
```python
from pachyderm_sdk import Client

while Client.worker.batch_datums():
    # process datum
    pass
```  

Jira: [INT-983]

[INT-983]: https://pachyderm.atlassian.net/browse/INT-983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python-sdk/pachyderm_sdk/api/worker/extension.py
```diff
@@ -0,0 +1,54 @@
...
+        This context manager expects to be called within an infinite while
+        loop -- see the examples section. This context can only be entered
+        from within a Pachyderm worker and the worker will terminate your
+        code when all datums have been processed.
+
+        Note: The API stub must have an open gRPC channel with the worker
+        for NextDatum to function correctly. The ``Client`` object
+        should automatically do this for the user.
+
+        Examples
```
kevyang, line 30: it slipped my mind on the datum batching example you wrote - it could be good to add an example in object detection for batch_datum() as well as the decorator

python-sdk/pachyderm_sdk/datum_batching.py
```diff
@@ -0,0 +1,44 @@
...
+    this decorator will call the NextDatum endpoint within the worker
+    and set any environment variables specified by the worker.
+
+    Any exceptions raised during the execution of the wrapped function
+    will be reported back to the worker. See the pachyderm documentation
+    for more information on how the datum batching feature works.
+
+    Note: This can only be used within a Pachyderm worker.
+
+    Examples
```
kevyang, line 21: maybe also include a reference to the object detection example you wrote.

kevyang (approved):
lgtm, with a couple of minor comments

python-sdk/pachyderm_sdk/api/worker/extension.py
bbonenfant: Only having an example for `batch_all_datums` was purposeful as we want to guide users towards that workflow. `batch_datums` is provided just in case users need the added flexibility.

PR #8659, opened 2023-03-16
Add more context to the worker's logging when user code context is ca…

albscui:
…nceled

For example, when the user code's context gets cancelled due to timeout, we now log with:
```
errored running user code after <duration>: signal: killed because context deadline exceeded
```

If the context wasn't cancelled, then the log message will remain the same as before.


jrockway (approved):
Really good cleanup.  Thank you!

PR #8661, opened 2023-03-17
[CORE-1545] error handling for retrieveCallInfo

armaanv:
adds error checks for the calls to the runtime library


zmajeed (approved):

jrockway:
The other thing I'd do here is pass `hex.EncodeToString(key)` instead of `string(key)` to the error message.

Error text shouldn't end up in Prometheus as the cardinality is too high; we should figure out where that works its way in and remove that feature, I think.  But even if we can't find that code (!) the hex-encoding of the key is an improvement.  (You can also consider base64; if a proto were being logged, it would be base64'd.  But things like global IDs are hex strings, so that's where that idea comes from.)

src/internal/storage/metrics/metrics.go
```diff
@@ -81,14 +89,21 @@ type callInfo struct {
 	line        int
 }
 
-func retrieveCallInfo(skip ...int) *callInfo {
+func retrieveCallInfo(skip ...int) (*callInfo, error) {
```
jrockway, line 92: BTW, another potential refactoring here is to just make retrieveCallInfo take the name of the function that's calling it, and then upstream you end up with something like `MonitorThroughput("GetFoo", func () ...)` instead 😂   You don't have to fix that here, I like what you have.

jrockway (approved):
LGTM.  I think that the hex.EncodeToString will prevent any panics in WithLabelValues.  I don't think that was the underlying issue for this escalation, but it will be nice to have it fixed!

PR #8663, opened 2023-03-20
Expose logging options for mount server

kevyang:
Users often have trouble sending us /tmp/mount-server.log because it’s too long to copy/paste, but it’s trapped in the notebook server/sidecar container. If we could log to the container’s stdout, users could capture the logs with their existing logging solution, or at least download the logs with kubectl logs.

This change exposes the mount server log location as an ENV variable. When the ENV variable is not specified, mount-server will log to stdout/stderr, which is captured within the Jupyter extension when running as a subprocess. By default, we will still log to /tmp/mount-server.log in order to not change the behavior in an unexpected way.

### DIFF
--- a/jupyter-extension/Dockerfile
+++ b/jupyter-extension/Dockerfile
@@ -2,6 +2,7 @@ FROM jupyter/scipy-notebook:lab-3.3.3
 # https://github.com/pachyderm/docker-stacks/pull/1/commits/fde4beb9ff1afb404f0e34828adc1f311f4bf2d7
 
 ENV PFS_MOUNT_DIR=/pfs
+ENV MOUNT_SERVER_LOG_DIR=/tmp/mount-server.log
 
 # TODO: use ARG TARGETPLATFORM to support arm builds, downloading pachctl arm64
 # binary below (instead of ..._linux_amd64.tar.gz below). See:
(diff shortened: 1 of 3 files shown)


msteffen:
> When the ENV variable is not specified, mount-server will log to stdout/stderr, which is captured within the Jupyter extension when running as a subprocess.

@chainlink I can check with Brody too, but do you know how Shakudo runs the mount-server (I know they use a Combined configuration, but I don't know if they use our image—my guess is no). I think that means that their container will start logging to stdout when they upgrade to the 2.6 `mount-server` (which I think is good, but just to check).

Also, do you know if there's a way in `docker run` to unset an environment variable? Specifically, if a user wants to use our all-in-one image, and wants it to log to stdout, can they do that? Will `docker run -e MOUNT_SERVER_LOG_DIR=  pachyderm/notebooks-user:2.6.0` (or some such) work?

chainlink:
Good thought @msteffen! It looks like this fails when passing a blank env var, so we should cover that case.

```
[I 2023-03-21 15:05:19.358 ServerApp.jupyterlab_pachyderm] Starting mount server...
bash: -c: line 0: syntax error near unexpected token `2'
bash: -c: line 0: `set -o pipefail; mount-server --mount-dir /pfs >>  2>&1'
```

Additionally, we should discuss what the default is here. It's actually kind of nice to have access to it in the container (most users have no idea where their jupyter server logs are)

kevyang:
Thanks, I didn't realize that an empty env variable would get picked up and be different than None. As for the default, I am open to removing the env var from the Dockerfile. I had opted to include /tmp/mount-server.log because I didn't want to change the way things were working when running things with the Dockerfile with no changes. With the new change I just made, setting the file to an empty string will log to stdout/stderr as well.

chainlink (approved):

PR #8664, opened 2023-03-21
Python-sdk CI

chainlink:
This is the first pass at running the python sdk tests in CI. It just adds the tests themselves. I'm ok with shipping them passing as is, or bringing back the failing tests and we can deal with that upstream in the python-sdk branch.

TODO
Tests
- [x] Create variable to flag based on changes to directory
- [x] Get Poetry running in CI
- [x] Create Circle job to run python tests

Future Work
- [ ] Output results in circle consumable format
  Publish
- [ ] Command to package up python
- [ ] Publish to Test pypi 
- [ ] Weave Jobs into Pachyderm release pipeline
- [ ] Publish to Pypi
- [ ] Docker container? - Default python container for users? - Testing datum batching and other features?
- [ ] Run Black formatter?

Future work - Tests we may want to write
- [ ] Testing with/without auth
- [ ] Testing with/without SSL
- [ ] Testing with/without enterprise

Proto-gen
- [ ] Flag when core protos change / are out of sync (We do this for the go protos I believe)
- [ ] Auto gen protos when core protos change?

Other changes
- [ ] Change from kind to mini-kube to match other tests in core 
- [ ] Use proxy instead of port-forward
- [ ] Change port defaults for proxy? (forthcoming proxy by default change)


chainlink:
Requires: https://github.com/pachyderm/pachyderm/pull/8670

.circleci/config.yml
```diff
@@ -207,6 +207,7 @@ workflows:
           mapping: |
             ([^/]+|(.circleci|src|etc|examples|goreleaser|licenses|dex-assets)/.*) run-core-jobs true
             ((.circleci|jupyter-extension)/.*) run-jupyter-jobs true
+            python-sdk/.* run-python-sdk-jobs true
```
bbonenfant, line 210: should this also be checking changes to `.circleci`?

.circleci/main.yml
```diff
@@ -1320,6 +1324,35 @@ jobs:
...
+            mkdir -p /home/circleci/.local/poetry/venv
+            export POETRY_HOME=/home/circleci/.local/poetry/venv
+            python3 -m venv $POETRY_HOME
+            $POETRY_HOME/bin/pip install poetry==1.4.1
+            $POETRY_HOME/bin/poetry --version
+            ln -s $POETRY_HOME/bin/poetry /home/circleci/bin/poetry
+      - run: #Note: --no-ansi https://github.com/python-poetry/poetry/issues/7184
+          name: Install Python Dependencies with Poetry
+          command: poetry --no-ansi install
+      - install-jupyter-test-deps
```
bbonenfant, line 1347: Do we need this?

bbonenfant (approved):
Looks good to me

.circleci/main.yml
chainlink: This is what boots up kube and such, I just borrowed it from the jupyter tests

bbonenfant: Ah, should we rename it something generic and/or extract the generic parts into a separate command?

PR #8665, opened 2023-03-21
distribute deploy tests round-robin to executors

djanicekpach:
This adds 4 executors for the deploy tests. This should lead to 5+ minute speed up for the pipelines. from ~18-20 minutes to ~12-15 minutes. This should mean the bottleneck for ci pipeline performance is now another job.
image

this unfortunately does not split sub-tests since `go test -list` can't calculate them without starting the execution.

I also updated the api performance test versions while I was there since they hadn't been updated since before the 2.5 release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -242,8 +242,6 @@ workflows:
           matrix:
             parameters:
               version:
-                - v2.0.0
-                - v2.3.9
-                - v2.4.0
-                - v2.5.0-alpha.4
+                - v2.4.6
+                - v2.5.1
                 - latest
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -482,6 +482,7 @@ jobs:
         type: string
         default: amd64
     resource_class: << parameters.resource_class >>
+    parallelism: 4
     machine:
       image: << pipeline.parameters.machine_image >>
     steps:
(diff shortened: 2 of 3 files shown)


molinamelendezj (approved):

.circleci/config.yml
```diff
@@ -242,8 +242,6 @@ workflows:
           matrix:
             parameters:
               version:
-                - v2.0.0
-                - v2.3.9
-                - v2.4.0
-                - v2.5.0-alpha.4
```
chainlink, line 248: Probably unrelated: Should the version numbers for the API perf tests be in the various version branches instead of in a matrix against master?

chainlink:
I think this is a good first pass, but we're definitely going to want to look at splitting based on test timing, as the current buckets are pretty unbalanced.

.circleci/config.yml
djanicekpach: Maybe? I kinda like defining what runs all in one place since this is a nightly-only run. We could have each version branch have it's own trigger in circle-ci and maybe that makes more sense organizationally and maintenance wise. I don't have a strong opinion, so I'm happy to switch it if you think it's better, but I think it's a bigger change than I should add to this PR. I'll make a ticket to circle back and re-evaluate.

PR #8669, opened 2023-03-23
[Jupyter] Hookup "Create Pipeline" button

bbonenfant:
* Implement callCreatePipeline function in the frontend
* Update python backend to handle real same config body sent from the frontend

Future work
- [ ] Add visual feedback pipeline has started to run (modal?)
- [ ] Display nicer error if pipeline fails (currently a stack trace)
- [ ] Check if notebook saved (dirty) before running pipeline
- [ ] SAME is extremely slow, perhaps this will be fixed when we interact with pachyderm directly, may need to add loading spinner
- [ ] Input spec - handle JSON in call cases (ie, saving notebook)
- [ ] e2e tests

### DIFF
--- a/jupyter-extension/jupyterlab_pachyderm/pps_client.py
+++ b/jupyter-extension/jupyterlab_pachyderm/pps_client.py
@@ -6,8 +6,11 @@ from pathlib import Path
 from tempfile import NamedTemporaryFile
 
 from tornado.httpclient import AsyncHTTPClient
+from tornado.web import HTTPError
 
 from .log import get_logger
+
+SAME_METADATA_FIELDS = ('apiVersion', 'environments', 'metadata', 'run', 'notebook')
 
 
 class PPSClient:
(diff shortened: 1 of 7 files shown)


jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
```diff
@@ -41,68 +44,54 @@ export const usePipeline = (
...
+    },
+    notebook: {
+      requirements: requirements,
+    },
+    run: {
+      name: pipelineName,
+      input: JSON.stringify(input),
+    },
+  };
+
```
bbonenfant, line 77: I moved these out of the functions to avoid repeated code. I _think_ this is OK because any changes to these values cause a re-render of the Pipeline extension. @pleeko could you do a quick sanity check?

jupyter-extension/jupyterlab_pachyderm/pps_client.py
```diff
@@ -29,38 +29,36 @@ async def generate(self, path, config):
...
-        return json.dumps(dict())
+            Path(temp_config.name).write_text(json.dumps(config))
+            command = [sys.executable, "-m", "sameproject.main", "run"]
+            command.extend(["--same-file", temp_config.name])
+            command.extend(["--target", "pachyderm"])
+            command.extend(["--input", input_spec])
+            call = subprocess.run(command, capture_output=True)
+            error = call.stderr.decode() if call.returncode else None
+
+        return json.dumps(dict(error=error))
```
bbonenfant, line 61: Currently this is always replying with a 200 code. @smalyala Is this what the mount extension does or is there a way to return specific error codes?

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
chainlink: Huh interesting, I confirmed `usePipeline` is getting called every time the metadata changes, but I'm curious if we actually want this behavior. Just to check my understanding: The `metadata` prop is used to set the metadata on initial component load, but from then on, the `set` state methods and the `useEffect` hook are used to track component state, and the component shouldn't get re-rendered for these internal state changes.

PR #8670, opened 2023-03-24
Change regex for core, remove need for boolean check on core jobs

chainlink:
This makes the path filtering logic more clear by updating the core path filtering regex to include all core code. This means future modules (like the forthcoming python sdk) can be added much more cleanly.

Questions
-  [x] Do we want to trigger a full CI run on any circle config changes? (Right now you need to manually dirty the module you're trying to test, see testing PRs below
- [ ] Arguably we don't need to run CI on license changes


chainlink:
Testing PRs
- https://github.com/pachyderm/pachyderm/pull/8671
- https://github.com/pachyderm/pachyderm/pull/8672

djanicekpach (approved):

PR #8674, opened 2023-03-24
Remove proxy from python-sdk (internal)

chainlink:


jrockway:
Yeah, proxy is internal.  It works around a pgbouncer limitation regarding subscribing to Postgres events; pgbouncer can't do that, so pachd does it for the workers and exposes it via this API.  @bbonenfant

PR #8675, opened 2023-03-24
Add makefile to python-sdk

chainlink:


bbonenfant (approved):
While we are making this file, do we want to move the `generate-protos.sh` script here as a `protos` command?

PR #8677, opened 2023-03-24
[CORE-1533] GetKubeEvents client accepts more than 1000 messages 

armaanv:
The new stream iterator has a limit of 1000 elements. We can get around this for `pachctl kube-events` by printing the events or flushing them to the writer as we iterate through the results

### DIFF
--- a/src/client/pps.go
+++ b/src/client/pps.go
@@ -858,7 +858,14 @@ func (c APIClient) GetKubeEvents(since time.Duration) ([]*pps.LokiLogMessage, er
 	if err != nil {
 		return nil, grpcutil.ScrubGRPC(err)
 	}
-	return grpcutil.Collect[*pps.LokiLogMessage](client, 1000)
+	var resp []*pps.LokiLogMessage
+	if err := grpcutil.ForEach[*pps.LokiLogMessage](client, func(msg *pps.LokiLogMessage) error {
+		resp = append(resp, msg)
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	return resp, nil
 }
 
 // CreatePipeline creates a new pipeline, pipelines are the main computation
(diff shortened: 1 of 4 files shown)


jrockway (changes requested):
I would prefer that the client side be fixed to take an unlimited number of messages vs making the server return a magic number worth of messages.  In the implementation of the command, do: 
```
logs, err := client.PpsAPIClient.GetKubeEvents(ctx, req)
grpcutil.ForEach(logs, func(x *pps.LokiLogMessage) {
...
})
```

src/client/pps.go
```diff
@@ -858,7 +858,14 @@ func (c APIClient) GetKubeEvents(since time.Duration) ([]*pps.LokiLogMessage, er
 	if err != nil {
 		return nil, grpcutil.ScrubGRPC(err)
 	}
-	return grpcutil.Collect[*pps.LokiLogMessage](client, 1000)
+	var resp []*pps.LokiLogMessage
+	if err := grpcutil.ForEach[*pps.LokiLogMessage](client, func(msg *pps.LokiLogMessage) error {
```
jrockway, line 862: BTW, are you sure that Go can't infer the type in `[*pps.LokiLogMessage]` from the type of the callback passed in?  Worth a try if you haven't already.

armaanv: hmm, will try this

src/client/pps.go
```diff
@@ -858,7 +858,14 @@ func (c APIClient) GetKubeEvents(since time.Duration) ([]*pps.LokiLogMessage, er
...
+		resp = append(resp, msg)
```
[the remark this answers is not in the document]
armaanv: I moved this code to the commands package

jrockway (approved):

PR #8679, opened 2023-03-27
[CORE-1202] Adjust pachctl logs --job to take into account job finish time

robert-uhl:


jrockway (commented):
So my worry here is that looking for the "finished" event is going to be confusing.  Say you have a job that takes 2 days to run, and it stalled, and you want to grab logs for the beginning of the event (maybe your user code logs "warning: this is slow and you should change your data).  With this change, you're used to seeing 24 hours of job history when looking at completed jobs, but now since you have a hanging job, the semantics of the flag change; it's only logs from 24 hours ago.

Maybe I could be persuaded by a warning: "job is not finished, showing the last 24 hours of logs, there may be more"?

robert-uhl:
Maybe the right solution is to hit the log system twice, once to get the timestamp of the last matching log and then once to get all logs within the `since` window prior to that timestamp? This `--since 24h` with a job which has been hung for 2 days would return the 24 hours of logs prior to the hang.

robert-uhl:
This version:

- If `--since` is explicitly set, respect it.
- If `--since` is not set, this is a job log request and there is a job creation timestamp, then return logs from job creation to the present.
- Otherwise return logs from the last 24 hours.

This is implemented for both Loki and non-Loki logs.

src/server/pps/server/api_server.go
```diff
@@ -1387,7 +1383,17 @@ func (a *apiServer) GetLogs(request *pps.GetLogsRequest, apiGetLogsServer pps.AP
 		rcName, containerName string
 		pods                  []v1.Pod
 		err                   error
+		sinceSeconds          *int64
```
jrockway, line 1386: Why represent this as an integer with units and not a *time.Duration?

robert-uhl: It was an integer seconds in the original code. But you’re right: it should be a duration. Refactored.

src/server/pps/server/api_server.go
```diff
@@ -1417,6 +1423,14 @@ func (a *apiServer) GetLogs(request *pps.GetLogsRequest, apiGetLogsServer pps.AP
 			if err != nil {
 				return errors.Wrapf(err, "could not get job information for \"%s\"", request.Job.ID)
 			}
+			if created := jobInfo.GetCreated(); sinceSeconds == nil && created != nil {
```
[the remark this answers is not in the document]
robert-uhl: I added an hour skew, and also created CORE-1647 to address the max query length issue (the short version is: no, `QueryLoki` does not currently work around the query length limit).

PR #8681, opened 2023-03-28
[Jupyter] replace SAME with python_pachyderm

bbonenfant:
Tasks:
 - [x] Remove SAME
 - [x] Create and upload files to companion repo
 - [x] Update body of request sent by frontend
 - [x] update tests

Note: This changes the format of the wrapper python which we use to execute the notebook code.

To check (kyle)
- [ ] Save/load vs run pipeline formats

Future Work:
 - [ ] Include hash of file in body of request and pull config directly from notebook file.
   - [ ] Update save-metadata code to have a simpler structure
   - [ ] Update backend config parsing to parse notebook metadata
 - [ ] Auth


jupyter-extension/jupyterlab_pachyderm/pps_client.py
```diff
@@ -41,72 +39,142 @@ async def create(self, path: str, config: dict):
...
+    entrypoint_script = (
+        f'{dedent(getsource(entrypoint))}\n\n'
+        'if __name__ == "__main__":\n'
+        '    entrypoint()\n'
+    )
+
+    with client.commit(repo, "master") as commit:
+        # Remove the old files.
+        for item in client.list_file((repo, "master"), "/"):
+            client.delete_file(commit, item.file.path)
```
chainlink, line 167: Why not make a new commit on top?

jupyter-extension/src/plugins/mount/types.ts
```diff
@@ -86,6 +86,13 @@ export interface IMountPlugin {
...
 }
 
+export type PpsConfig = {
+  pipeline_name: string;
+  image: string;
+  requirements: string | null;
+  input_spec: any;
+};
+
 export type SameMetadata = {
```
chainlink, line 96: Can you fully eliminate the Same metadata and other related code?

bbonenfant: Yes, I can do it here, but I figured it would be nice to do this incrementally in another PR

jupyter-extension/jupyterlab_pachyderm/pps_client.py
bbonenfant: Not sure what you mean by "commit on top". This _is_ a little heavy handed, but it's to make sure that no files linger in the context when the pipeline is updated. A example which could cause trouble is a user removing their "requirements.txt" file and we want to make sure its deleted in the companion repo. This will also be more important when we allow users to include arbitrary files in their "context"

jupyter-extension/src/plugins/mount/types.ts
chainlink: Cool, I'll add a future work section

jupyter-extension/jupyterlab_pachyderm/pps_client.py
chainlink: This makes sense, sorry my mental model was wrong here (was thinking of separate branches, not commits, which are of course additive)

chainlink:
May want to change the `METADATA_KEY` in mount.tsx

jupyter-extension/jupyterlab_pachyderm/pps_client.py
```diff
@@ -41,72 +39,142 @@ async def create(self, path: str, config: dict):
...
-
-        + 'root_module = sys.argv[1]; '
-        + f'context_dir = Path("/pfs", "{companion_repo}"); '
-        + 'reqs = context_dir / "requirements.txt"; '
-        + 'reqs.exists() and run(["pip", "--disable-pip-version-check", "install", "-r", reqs.as_posix()]); '
-        + 'sys.path.append(context_dir.as_posix()); '
-        + 'script = import_module(root_module); '
-        + 'script.root()'
+@dataclass
+class PpsConfig:
```
chainlink, line 71: Not sure of the python convention here w.r.t. separate classes, but should this be in it's own file?

bbonenfant: It's generally ok to keep related classes in the same file.

jupyter-extension/jupyterlab_pachyderm/tests/test_handlers.py
```diff
@@ -540,7 +540,17 @@ async def test_health(mock_client, jp_fetch):
 
 @pytest.mark.skipif(sys.version_info < (3, 7), reason="requires python3.7 or higher")
```
chainlink, line 541: Future work: Can get rid of this skip tests

jupyter-extension/jupyterlab_pachyderm/tests/test_integrations.py
```diff
@@ -401,24 +400,21 @@ def test_pps(dev_server, simple_pachyderm_env):
...
+        image=image,
+        input_spec=dict(pfs=dict(repo=repo_name, glob="/*"))
     )
     r = requests.put(f"{BASE_URL}/pps/_create/{path}", data=json.dumps(data))
     assert r.status_code == 200
     assert next(client.inspect_pipeline(pipeline_name))
 
 
-@pytest.mark.parametrize('excluded_field', SAME_METADATA_FIELDS)
+@pytest.mark.parametrize('excluded_field', ('pipeline_name', 'image', 'input_spec'))
```
chainlink, line 413: Future work: These fields should still be captured in a variable to reduce duplication

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
```diff
@@ -82,12 +81,28 @@ export const usePipeline = (
...
+        throw e;
+      }
+    }
+
+    // const sameMetadata = createSameMetadata();
     try {
       const response = await requestAPI<CreatePipelineResponse>(
         `pps/_create/${notebookPath}`,
         'PUT',
-        sameMetadata as ReadonlyJSONObject,
```
chainlink, line 90: Future work: Should use PPSConfig object defined below

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
```diff
@@ -82,12 +81,28 @@ export const usePipeline = (
...
+      input = YAML.parse(inputSpec);
+    } catch (e) {
+      if (e instanceof YAML.YAMLParseError) {
+        input = JSON.parse(inputSpec);
+      } else {
+        throw e;
+      }
+    }
+
+    // const sameMetadata = createSameMetadata();
```
chainlink, line 95: Future work: Remove commented out code here

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
```diff
@@ -82,12 +81,28 @@ export const usePipeline = (
     setLoading(true);
     setErrorMessage('');
 
-    const sameMetadata = createSameMetadata();
+    let input: string;
```
chainlink, line 84: Future work: We should encapsulate this yaml/json parsing to it's own method

chainlink:
Just to check my own understanding. This doesn't change the format of the data saved/loaded from the notebook, but does change the format of what's being sent to the server, right?

jupyter-extension/jupyterlab_pachyderm/pps_client.py
chainlink: Future work: We also might want to consider separating concerns here. Generic PPSConfig class which returns an error, and then the caller can decide what to do with it (wrap in http, etc)

jupyter-extension/jupyterlab_pachyderm/pps_client.py
```diff
@@ -41,72 +39,142 @@ async def create(self, path: str, config: dict):
...
-    pipeline_name: str,
-    image: str,
-    input_spec: dict,
-    script_name: str,
-) -> dict:
-    """Generates the pipelines spec. Currently copied from """
-    companion_repo = f"{pipeline_name}__context"
-    micro_entrypoint = (
-        'print("Greetings from the Pachyderm PPS Extension"); '
-        + 'import sys; '
```
chainlink, line 91: It looks like we've changed how we're running the script? (This wrapper doesn't look present in the new version)

bbonenfant:
> Just to check my own understanding. This doesn't change the format of the data saved/loaded from the notebook, but does change the format of what's being sent to the server, right?

It does not. That will be addressed in a subsequent PR

jupyter-extension/jupyterlab_pachyderm/pps_client.py
bbonenfant: It has changed. This method that was deleted basically defines our entrypoint in as a series of python statements and is hacky, not very maintainable, and would most likely be confusing for the user when looking at their pipeline spec. However, when moving that code over from SAME, I realized we should just upload that entrypoint as a script within the companion repo and actually doing so removes a lot of the complication surrounding it.

jupyter-extension/jupyterlab_pachyderm/pps_client.py
bbonenfant: Oh I actually did exactly this in #8696

chainlink (approved):

PR #8678, opened 2023-03-24
auth: setup default project role binding where we set up all the others

jrockway:
This avoids a transaction conflict that can't be resolved.

PR #8683, opened 2023-03-29
[2.5.x] Fix auth activation from the environment

jrockway:
Backport of #8678.

Relative to the version in master, I removed a log line that would log sensitive information in 2.5 (no log redaction yet).  I also didn't fix any TODOs in the nearby code; 2.6 adds some new roles around projects, we don't add that one here.

* auth: setup default project role binding where we set up all the other role bindings; to avoid a txn conflict

* auth: add test for the transaction problems; adjust old tests that rely on half-activated auth; add log messages to auth activation flow

* fix logging typos and incorrect fields; make sure the test fails when the bug is re-introduced (it does)

* auth: activateInTransaction: avoid nested transaction

* auth: we rely on transactions being committed too early; fixing that opened a new issue; so now the transaction context also caches whether or not auth is enabled

* lint

* improve comment slightly

* tyop

### DIFF
--- a/src/internal/transactionenv/txncontext/context.go
+++ b/src/internal/transactionenv/txncontext/context.go
@@ -2,6 +2,7 @@ package txncontext
 
 import (
 	"context"
+	"sync/atomic"
 	"time"
 
 	"github.com/gogo/protobuf/types"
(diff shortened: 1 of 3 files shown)


acohen4 (approved):
LGTM

PR #8689, opened 2023-03-29
Update pachctl stop job to use job's actual project instead of default

albscui:


src/server/pps/cmds/cmds.go
```diff
@@ -405,7 +405,7 @@ $ {{alias}} -p foo -i bar@YYY`,
 				}
 				if _, err := client.RunBatchInTransaction(func(tb *pachdclient.TransactionBuilder) error {
 					for _, jobInfo := range jobInfos {
-						if err := tb.StopProjectJob(pfs.DefaultProjectName, jobInfo.Job.Pipeline.Name, jobInfo.Job.ID); err != nil {
+						if err := tb.StopProjectJob(jobInfo.Job.Pipeline.Project.Name, jobInfo.Job.Pipeline.Name, jobInfo.Job.ID); err != nil {
```
jrockway, line 408: Are we not being paranoid about pipeline.project being nil here?

albscui: I think at this point, we are pretty sure that all pipelines have projects. However, even if we are not sure, I think we should just let this panic and fix retroactively.

jrockway (approved):

PR #8696, opened 2023-03-30
[Jupyter] Read PPS config from notebook

bbonenfant:
Tasks
- [x] Backend reads config from notebook metadata
- [x] Use last_modified_time to verify correct notebook version
  - [x] Access last_modified_time in frontend
  - [x] Feed last_modified_time to Pipeline react widget at render time
  - [x] Rerender Pipeline widget on save to disk.
    - [x] Bundle relevant frontend data into a PpsContext object.
- [x] Remove request body from pps/_create GET call.
- [x] Remove logging statements.

Future Tasks
- [ ] Restructure PPS config
- [ ] Don't stringify input spec on save

Test cases
- Confirmed tripping of last_modified timestamp check by: create a pipeline enabled notebook and saving it, using `touch` to bump the timestamp on the file, then clicking `Create Pipeline` in the UI.
- Tried creating a pipeline with a deeply nested notebook `./test/something/new/InFolder.ipynb` which was created correctly.

### DIFF
--- a/jupyter-extension/jupyterlab_pachyderm/pps_client.py
+++ b/jupyter-extension/jupyterlab_pachyderm/pps_client.py
@@ -1,9 +1,11 @@
-import os.path
 import json
+import os.path
 from dataclasses import dataclass
+from datetime import datetime
 from inspect import getsource
+from pathlib import Path
 from textwrap import dedent
-from typing import Optional
+from typing import Optional, Union
 
 import python_pachyderm
 from nbconvert import PythonExporter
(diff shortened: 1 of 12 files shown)


chainlink:
Do you have a different linter running here? Noticing a bunch of stylistic changes

jupyter-extension/jupyterlab_pachyderm/pps_client.py
```diff
@@ -70,32 +97,42 @@ async def create(self, path: str, config: dict):
...
+        """Parses a config from the metadata of a notebook file.
 
-        Raises HTTPError with 500 code if required field is not specified.
+        Raises ValueError if required field is not specified.
         """
-        pipeline_name = config.get('pipeline_name', None)
+        notebook_path = Path(notebook_path)
+        notebook_data = json.loads(notebook_path.read_bytes())
+
+        config = notebook_data['metadata'].get('same_config')
```
chainlink, line 115: We should figure out a way to share the config key name so the FE and BE use the same value

chainlink: future work note

jupyter-extension/jupyterlab_pachyderm/pps_client.py
```diff
@@ -18,35 +20,60 @@ class PPSClient:
     def __init__(self):
         self.nbconvert = PythonExporter()
 
-    async def generate(self, path, config):
+    @staticmethod
```
chainlink, line 23: Just for my own understanding, now come this changed to a staticmethod?

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
```diff
@@ -34,20 +33,56 @@ export const usePipeline = (
...
+  let callCreatePipeline: () => Promise<void>;
+  if (ppsContext?.notebookModel) {
+    const notebook = ppsContext.notebookModel;
+    callCreatePipeline = async () => {
+      setLoading(true);
+      setErrorMessage('');
+      setResponseMessage('');
+      try {
+        const response = await requestAPI<CreatePipelineResponse>(
+          `pps/_create/${notebook.path}`,
```
chainlink, line 60: Is this path URL safe?

chainlink (approved):
Looks good! Can merge pending getting the tests running

chainlink:
Future work: We may want to consider moving the notebook path to inside the request body. While the contents API has the path in the URL, they do a bunch of munging on it to make sure it's valid.

PR #8703, opened 2023-03-30
logging middleware: log grpc metadata, including user-agent

jrockway:
This helps with `pachyderm-sdk`'s desire to send some headers with each request, for client version number and whatnot.

`log.Metadata` already knows how to redact the auth token, and does.

### DIFF
--- a/src/internal/log/field.go
+++ b/src/internal/log/field.go
@@ -65,14 +65,17 @@ func RetryAttempt(i int, max int) Field {
 	return zap.Inline(&attempt{i: i, max: max})
 }
 
-// Metadata is a Field that logs the provided metadata (canonicalizing keys, and collapsing
-// single-element values to strings).
+// Metadata is a Field that logs the provided metadata (canonicalizing keys, collapsing
+// single-element values to strings, and removing the Pachyderm auth token).
 func Metadata(name string, md metadata.MD) Field {
 	return zap.Object(name, zapcore.ObjectMarshalerFunc(func(enc zapcore.ObjectEncoder) error {
 		keys := maps.Keys(md)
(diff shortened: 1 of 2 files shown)


robert-uhl (changes requested):
I don’t see anything in `log.Metadata` which sanitizes the auth token; I do see that there is a `log.OutgoingMetadata` field type; perhaps there should be a corresponding `log.IncomingMetadata`, or perhaps `log.Metadata` could detect `constants.ContextTokenKey`?

jrockway:
@robert-uhl Good catch.  I've adjusted Metadata to handle removing the auth token (and changed it to "[MASKED]" which is what protoc-gen-zap does).  PTAL.

src/internal/log/field.go
```diff
@@ -65,14 +65,17 @@ func RetryAttempt(i int, max int) Field {
...
+// single-element values to strings, and removing the Pachyderm auth token).
 func Metadata(name string, md metadata.MD) Field {
 	return zap.Object(name, zapcore.ObjectMarshalerFunc(func(enc zapcore.ObjectEncoder) error {
 		keys := maps.Keys(md)
 		sort.Strings(keys)
 		for _, k := range keys {
 			v := md[k]
+			if k == constants.ContextTokenKey {
+				v = []string{"[MASKED]"}
+			}
```
robert-uhl, line 78: Right now it’s only possible to have a single token, so this is okay. It _might_ be worth iterating through the value, setting each item to `"MASKED"`, in order to preserve an a congruent array, in case zero or multiple tokens someday make it into a request.

robert-uhl (approved):

PR #8704, opened 2023-03-31
Compact Diff File Sets in Old Commits When Calling Find Commits

FahadBSyed:
## Description
This PR implements a simple backfill strategy for handling reads on uncompacted diff file sets. These file sets will be compacted at read time by the PFS master's driver. It is possible that multiple concurrent client calls could lead to the driver attempting to recompact the diff file set for the same commit at the same time and waste work, but we think its a worthy trade off for keeping the master code simpler.

To make this more efficient, the PFS team revisited the IsCompacted() logic which involved making optimizations to Flatten() so that it would stream over a set of primitives rather than holding all the primitives in memory. Some helper functions were also modified to reduce duplicate code.

### DIFF
--- a/src/internal/storage/fileset/storage.go
+++ b/src/internal/storage/fileset/storage.go
@@ -7,25 +7,22 @@ import (
 	"time"
 
 	units "github.com/docker/go-units"
+	"golang.org/x/sync/semaphore"
+
 	"github.com/pachyderm/pachyderm/v2/src/internal/dbutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/errutil"
 	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/chunk"
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/fileset/index"
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/renew"
 	"github.com/pachyderm/pachyderm/v2/src/internal/storage/track"
-	"golang.org/x/sync/semaphore"
 )
 
 const (
 	// DefaultMemoryThreshold is the default for the memory threshold that must
 	// be met before a file set part is serialized (excluding close).
 	DefaultMemoryThreshold = units.GB
-	// DefaultCompactionFixedDelay is the default fixed delay for compaction.
-	// This is expressed as the number of primitive filesets.
-	// TODO: Potentially remove this configuration.
-	// It is easy to footgun with this configuration.
-	DefaultCompactionFixedDelay = 1
 	// DefaultCompactionLevelFactor is the default factor that level sizes increase by in a compacted fileset.
 	DefaultCompactionLevelFactor = 10
 	DefaultPrefetchLimit         = 10
@@ -65,7 +62,7 @@ type Storage struct {
 }
 
 type CompactionConfig struct {
-	FixedDelay, LevelFactor int64
+	LevelFactor int64
 }
 
 // NewStorage creates a new Storage.
@@ -81,7 +78,6 @@ func NewStorage(mds MetadataStore, tr track.Tracker, chunks *chunk.Storage, opts
 			SizeBytes: index.DefaultShardSizeThreshold,
 		},
 		compactionConfig: &CompactionConfig{
-			FixedDelay:  DefaultCompactionFixedDelay,
 			LevelFactor: DefaultCompactionLevelFactor,
 		},
 		filesetSem:    semaphore.NewWeighted(math.MaxInt64),
@@ -126,7 +122,7 @@ func (s *Storage) newReader(id ID) *Reader {
 // Open opens a file set for reading.
 func (s *Storage) Open(ctx context.Context, ids []ID) (FileSet, error) {
 	var err error
-	ids, err = s.Flatten(ctx, ids)
+	ids, err = s.FlattenAll(ctx, ids)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


src/internal/storage/fileset/storage.go
```diff
@@ -185,39 +181,55 @@ func (s *Storage) CloneTx(tx *pachsql.Tx, id ID, ttl time.Duration) (*ID, error)
...
 		md, err := s.store.Get(ctx, id)
 		if err != nil {
-			return nil, err
+			return err
 		}
 		switch x := md.Value.(type) {
 		case *Metadata_Primitive:
-			flattened = append(flattened, id)
+			if err := cb(id); err != nil {
+				if err != errutil.ErrBreak {
```
brycemcanally, line 207: nit, this check is usually structured like this:
```
if errors.Is(err, errutil.ErrBreak) {
  return nil
}
return err
```

src/internal/storage/fileset/storage.go
```diff
@@ -185,39 +181,55 @@ func (s *Storage) CloneTx(tx *pachsql.Tx, id ID, ttl time.Duration) (*ID, error)
...
 			ids, err := x.Composite.PointsTo()
 			if err != nil {
-				return nil, err
+				return err
 			}
-			ids2, err := s.Flatten(ctx, ids)
-			if err != nil {
-				return nil, err
+			if err := s.flatten(ctx, ids, cb); err != nil {
+				if err != errutil.ErrBreak {
```
brycemcanally, line 218: same as above.

src/server/pfs/server/commit_store.go
```diff
@@ -22,21 +22,21 @@ type commitStore interface {
 	AddFileSet(ctx context.Context, commit *pfs.Commit, filesetID fileset.ID) error
 	// AddFileSetTx is identical to AddFileSet except it runs in the provided transaction.
 	AddFileSetTx(tx *pachsql.Tx, commit *pfs.Commit, filesetID fileset.ID) error
-	// SetTotalFileSet sets the total fileset for the commit, overwriting whatever is there.
+	// SetTotalFileSet sets the total file set for the commit, overwriting whatever is there.
```
brycemcanally, line 25: :+1:

src/server/pfs/server/driver.go
```diff
@@ -632,6 +633,45 @@ func (d *driver) isPathModifiedInCommit(ctx context.Context, commit *pfs.Commit,
...
+		}
+		if diff != nil {
+			compactedDiff = diff
+			return nil
+		}
+		return errors.Errorf("diff file set for commit %s is not compacted yet", commit.ID)
+	}, backoff.NewConstantBackOff(time.Second), backoff.NotifyCtx(ctx, "getCompactedDiffFileSet for "+commit.ID))
+}
+
+func (d *driver) getDiffFileSetIfCompacted(ctx context.Context, commit *pfs.Commit) (*fileset.ID, error) {
```
brycemcanally, line 651: I think this code should just be a part of `getCompactedDiffFileSet`. I don't see us needing the no-op event separate from the backoff, and the nil, nil return seems like an anti-pattern.

src/server/pfs/server/driver.go
```diff
@@ -632,6 +633,45 @@ func (d *driver) isPathModifiedInCommit(ctx context.Context, commit *pfs.Commit,
...
+		diff, err := d.getDiffFileSetIfCompacted(ctx, commit)
+		if err != nil {
+			return err
+		}
+		if diff != nil {
+			compactedDiff = diff
+			return nil
+		}
+		return errors.Errorf("diff file set for commit %s is not compacted yet", commit.ID)
+	}, backoff.NewConstantBackOff(time.Second), backoff.NotifyCtx(ctx, "getCompactedDiffFileSet for "+commit.ID))
```
brycemcanally, line 648: We should discuss the backoff configuration more. Polling every second seems excessive.

FahadBSyed: I think last time we discussed this, we acknowledged that 1 second was probably too aggressive generally, but it would lead to a better user experience from the Console perspective. Happy to adjust it to something less aggressive if we think its necessary.

src/internal/storage/fileset/storage.go
```diff
@@ -185,39 +181,56 @@ func (s *Storage) CloneTx(tx *pachsql.Tx, id ID, ttl time.Duration) (*ID, error)
 	}
 }
 
-// Flatten takes a list of IDs and replaces references to composite FileSets
-// with references to all their layers inplace.
-// The returned IDs will only contain ids of Primitive FileSets
-func (s *Storage) Flatten(ctx context.Context, ids []ID) ([]ID, error) {
+// FlattenAll is like Flatten, but collects the primitives to return to the user.
```
brycemcanally, line 184: nit, could we move this below the `Flatten` definition. It just reads better imo.

src/server/pfs/server/master.go
```diff
@@ -212,15 +212,25 @@ func (d *driver) finishRepoCommits(ctx context.Context, repoKey string) error {
 func (d *driver) compactCommit(ctx context.Context, compactor *compactor, doer task.Doer, commit *pfs.Commit) (*fileset.ID, error) {
 	var totalId *fileset.ID
 	if err := d.storage.WithRenewer(ctx, defaultTTL, func(ctx context.Context, renewer *fileset.Renewer) error {
+		var err error
```
brycemcanally, line 215: Why not define this closer to where it is used?

src/server/pfs/server/master.go
```diff
@@ -180,27 +180,52 @@ func (d *driver) finishRepoCommits(ctx context.Context, repoKey string) error {
 				compactor := newCompactor(d.storage, d.env.StorageConfig.StorageCompactionMaxFanIn)
 				taskDoer := d.env.TaskService.NewDoer(StorageTaskNamespace, commit.ID, cache)
 				var totalId *fileset.ID
```
brycemcanally, line 182: I don't think this is needed anymore.

src/server/pfs/server/master.go
```diff
@@ -180,27 +180,52 @@ func (d *driver) finishRepoCommits(ctx context.Context, repoKey string) error {
...
-				}); err != nil {
-					return err
-				}
-				details.ValidatingTime = types.DurationProto(time.Since(start))
-				// Finish the commit.
-				return d.finalizeCommit(ctx, commit, validationError, details, totalId)
+					// A commit may have a compacted total file set, but an uncompacted diff file set.
+					// If a total exists, this commit has been through total file set compaction before
+					// and the rest of the compaction steps can be skipped.
+					totalId, err = d.commitStore.GetTotalFileSet(ctx, commit)
```
brycemcanally, line 196: I think both of these variables can be declared here, and the check for a non-nil error can probably be removed. We only need to handle the nil case, and proceed otherwise.

src/server/pfs/server/driver.go
```diff
@@ -773,6 +773,31 @@ func (d *driver) isPathModifiedInCommit(ctx context.Context, commit *pfs.Commit,
 	return found, nil
 }
 
+func (d *driver) getCompactedDiffFileSet(ctx context.Context, commit *pfs.Commit) (*fileset.ID, error) {
+	var diff *fileset.ID
+	var err error
+	diff, err = d.commitStore.GetDiffFileSet(ctx, commit)
```
brycemcanally, line 779: nit, you could just use a short declaration.

brycemcanally (approved):
lgtm, just one minor nit.

PR #8706, opened 2023-03-31
CORE-1479 Add new cron expression @never to disable scheduled runs of cron pipeline but still allow manual runs

zmajeed:
Adds support for `@never` cron spec for cron pipelines - still adding unit tests but there is a script to test the feature at https://github.com/pachyderm/zartaj/blob/main/pachyderm_bugs/core_1479/core_1479.sh - the script creates a cron pipline with spec “@every 31s” to run every 31 seconds. The pipeline is allowed to run twice - then the cron spec for the pipeline is updated to `@never` and two periods of 31 seconds are allowed to pass and pipeline is checked for any activity - finally the pipeline is run manually with pachctl run cron - again 2 times and runs and output files are checked

The frequency and the number of runs can be changed for the test script

The main change is to wrap the cron library robfig/cron in a new module `cronutil` that checks for the enhancement of `@never` - and returns a timepoint far into the future (year 9999) to act like forever so scheduled runs are practically disabled - any other expression is straightaway passed to robfig/cron - this should preserve all non-scheduled behavior of a cron pipeline even in the "disabled" `@never` case

I picked year 9999 to keep within 4-digit years even though time support should extend much further out


brendoncarroll (approved):
LGTM

zmajeed:
Adding unit tests for new cronutil module

PR #8707, opened 2023-03-31
CORE-1474 Fix cron pipelines to honor an updated start time

zmajeed:
Tweaked logic in `monitor.go` 

Return bigger of start time and last filename timestamp from cron repo if both are valid
Otherwise return start time if available
Otherwise return error

Bug was start time was ignored if timestamp files existed in cron repo

There's a demo script at https://github.com/pachyderm/zartaj/blob/main/pachyderm_bugs/core_1474/core_1474.sh but I'll look into adding unit tests


src/server/pps/server/monitor.go
```diff
@@ -370,22 +370,43 @@ func getLatestCronTime(ctx context.Context, env Env, in *pps.Input) (retTime tim
...
 		if err != nil {
-			return latestTime, errors.EnsureStack(err)
+			return latestTime, err
 		}
+		// get cron start time to compare if previous start time was updated
+		startTime, err := types.TimestampFromProto(in.Cron.Start)
+
+		// return latest time from filename if start time cannot be determined
+		if err != nil {
+			return latestTime, nil
```
acohen4, line 396: should return `err`

acohen4 (changes requested):
A few comments

acohen4 (approved):
LGTM. Only nit is that the test may run long

PR #8719, opened 2023-04-06
[CORE-1585] Apply projects filter after pagination for ListJob and ListJobSet

armaanv:
Currently, these APIs don't respect page size API parameter when multiple projects exist. This fixes that by applying the projects filter before pagination count is decremented


albscui (approved):
LGTM!

PR #8722, opened 2023-04-06
Djanicek/core 1603/auth migrate

djanicekpach:
2.6 version of https://github.com/pachyderm/pachyderm/pull/8717

We can't cherry pick since 2.6 migrates already exist, so the boilerplate change is slightly different.

Design: https://www.notion.so/2023-04-05-Projects-Auth-migrate-hot-fix-patch-fix-migrate-5a3dc29e9b384889995a053e9e30031d

Adding design/slack commenters as reviewers for visibility, although this should just match the design.

### DIFF
--- a/src/internal/clusterstate/v2.5.4.go
+++ b/src/internal/clusterstate/v2.5.4.go
@@ -1,0 +1,18 @@
+// DO NOT MODIFY THIS STATE
+// IT HAS ALREADY SHIPPED IN A RELEASE
+package clusterstate
+
+import (
+	"context"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/migrations"
+)
+
+var state_2_5_4 migrations.State = state_2_5_0.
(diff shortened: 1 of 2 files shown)


albscui (changes requested):
I wouldn't inject the 2.5.4 migration here, since 2.6 and 2.5 now live on separate branches and should not influence each other. The only reason we sometimes backport changes in 2.6 back to 2.5 is to fix bugs.

I think we should just add a step in the 2.6 migration like I did here. https://github.com/pachyderm/pachyderm/pull/8716/files

albscui (approved):
Actually you are right 😅 
We need to include the 2.5.4 migration here, in order to syncrhonize with users who already ran the 2.5.4 migration.

djanicekpach:
> Actually you are right sweat_smile We need to include the 2.5.4 migration here, in order to syncrhonize with users who already ran the 2.5.4 migration.

I updated the doc to include info on the decision 
https://www.notion.so/2023-04-05-Projects-Auth-migrate-hot-fix-patch-fix-migrate-5a3dc29e9b384889995a053e9e30031d?pvs=4#67d7af1744434f8c911dd25a7e99e688

PR #8723, opened 2023-04-06
Update docker run command

chainlink:
with port forwarding, and add instructions on re-use


jupyter-extension/README.md
```diff
@@ -74,7 +73,20 @@ Within container run:
...
+Note: Once run above, you can start the dev container again in future sessions using
+
+```
+docker start <container id>
+```
+
+and then running
+
+```
+docker exec -it a40 bash
```
kevyang, line 85: what is a40?

chainlink: Updated!

kevyang (approved):

PR #8725, opened 2023-04-06
[PFS-70] Only define GOOGLE_APPLICATION_CREDENTIALS if GOOGLE_CREDS is also defined

FahadBSyed:
A bug was introduced as part of the `go-cdk` object storage client migration where GCP users were forced to use environment variables for authentication. This is because `GOOGLE_APPLICATION_CREDENTIALS` was always being defined, even in cases where no credentials were being supplied (like when using workload identity). The solution is to only set `GOOGLE_APPLICATION_CREDENTIALS` when `GOOGLE_CREDS` is also defined by the user.

### DIFF
--- a/etc/helm/pachyderm/templates/pachd/storage-secret.yaml
+++ b/etc/helm/pachyderm/templates/pachd/storage-secret.yaml
@@ -15,7 +15,7 @@ data:
   {{- if eq (include "pachyderm.storageBackend" . )  "GOOGLE" }}
   GOOGLE_BUCKET: {{ required "Google bucket required" .Values.pachd.storage.google.bucket | b64enc | quote }}
   GOOGLE_CRED: {{ .Values.pachd.storage.google.cred | b64enc | quote }}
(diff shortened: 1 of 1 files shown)


etc/helm/pachyderm/templates/pachd/storage-secret.yaml
```diff
@@ -15,8 +15,10 @@ data:
   {{- if eq (include "pachyderm.storageBackend" . )  "GOOGLE" }}
   GOOGLE_BUCKET: {{ required "Google bucket required" .Values.pachd.storage.google.bucket | b64enc | quote }}
   GOOGLE_CRED: {{ .Values.pachd.storage.google.cred | b64enc | quote }}
-  # This path is used by go-cdk for gcp authentication
+  {{- if not (empty .Values.pachd.storage.google.cred) -}}
```
brycemcanally, line 18: nit, but can this not just be `if .Values.pachd.storage.google.cred`.

brycemcanally (approved):
lgtm

PR #8727, opened 2023-04-06
CreatePipeline validates input commit and errors accordingly

albscui:


jrockway:
I'm really unconvinced that this is the right fix.  We read from and write to this field internally, and if it's only for internal use, I guess that should be documented.   For example, it's read to create some internal structures for the S3 gateway, so there must have been some intent to use it, right?

src/server/pps/server/api_server.go
```diff
@@ -212,6 +212,8 @@ func (a *apiServer) validateInput(pipeline *pps.Pipeline, input *pps.Input) erro
 				return errors.Errorf("input cannot specify both 's3' and " +
 					"'empty_files', as 's3' requires input data to be accessed via " +
 					"Pachyderm's S3 gateway rather than the file system")
+			case input.Pfs.Commit != "":
+				return errors.Errorf("input commit is not supported at the PPS Spec level")
```
jrockway, line 216: "input cannot come from a commit; make a branch at that commit and set the branch"?

jrockway (approved):

albscui:
> I'm really unconvinced that this is the right fix. We read from and write to this field internally, and if it's only for internal use, I guess that should be documented. For example, it's read to create some internal structures for the S3 gateway, so there must have been some intent to use it, right?

Great point, we seem to be leaking private fields. However, the validation applied here only runs at pipeline creation time, so shouldn't affect how we use these field internally.

That said, the long term solution could be to either remove this field from the PPS spec, or actually implement the feature that actually keeps the input at a fixed commit as the user expects.

PR #8732, opened 2023-04-11
Make PFS DeleteProject & DeleteAll Transactional

acohen4:


src/server/pfs/server/driver.go
```diff
@@ -266,9 +266,25 @@ func (d *driver) getPermissions(ctx context.Context, repo *pfs.Repo) ([]auth.Per
...
+	resp, err := d.env.AuthServer.GetPermissionsInTransaction(txnCtx, &auth.GetPermissionsRequest{Resource: repo.AuthResource()})
+	if err != nil {
+		return nil, nil, errors.EnsureStack(err)
+	}
+
+	return resp.Permissions, resp.Roles, nil
+}
+
+// NOTE if includeAuth = true, it is expected that auth is active
+func (d *driver) processListRepoInfo(
```
FahadBSyed, line 279: Why did we choose to model some of the logic as function parameters instead of putting that logic inside this method itself? Is that for unit testing purposes?

src/server/pfs/server/driver.go
```diff
@@ -533,44 +616,48 @@ func (d *driver) canDeleteRepo(txnCtx *txncontext.TransactionContext, repo *pfs.
 }
 
 func (d *driver) createProject(ctx context.Context, req *pfs.CreateProjectRequest) error {
+	return d.txnEnv.WithWriteContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
+		return d.createProjectInTransaction(txnCtx, req)
+	})
+}
+
+func (d *driver) createProjectInTransaction(txnCtx *txncontext.TransactionContext, req *pfs.CreateProjectRequest) error {
```
FahadBSyed, line 624: I'm curious to know, is there a reason why we aren't doing this in a write transaction anymore?

FahadBSyed (approved):
lgtm, just had some questions for context

src/server/pfs/server/driver.go
acohen4: we are still doing it in a write transaction. The Write transaction gets set up in createProject(), and it passes the transaction context into this function

PR #8735, opened 2023-04-12
Don't download data if input is S3 input

msteffen:
Fix [INT-897]

[INT-897]: https://pachyderm.atlassian.net/browse/INT-897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/pps/server/s3g_sidecar_test.go
+++ b/src/server/pps/server/s3g_sidecar_test.go
@@ -860,11 +860,12 @@ func TestDontDownloadData(t *testing.T) {
 		Transform: &pps.Transform{
 			Cmd: []string{"bash", "-x"},
 			Stdin: []string{
-				"if find /pfs/.scratch -type f | xargs grep --with-filename \"This is a test\"; then",
-				"  exit 1", // The data shouldn't be downloaded; this means it was
-				"else",
-				"  exit 0", // No data == success
-				"fi",
+				`if find -L /pfs -type f | xargs grep --with-filename "This is a test"; then`,
+				`  exit 1`, // The data shouldn't be downloaded; this means it was
+				`fi`,
+				`if ls /pfs/input_repo; then`,
+				`  exit 1`, // The input dir shouldn't be linked either; this means it was
+				`fi`,       // success
 			},
 		},
 		ParallelismSpec: &pps.ParallelismSpec{Constant: 1},
--- a/src/server/pps/server/s3g_sidecar_test.go
+++ b/src/server/pps/server/s3g_sidecar_test.go
@@ -15,8 +15,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"io"
-	"math/rand"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


src/server/worker/datum/datum.go
```diff
@@ -258,6 +258,9 @@ func (d *Datum) downloadData(downloader pfssync.Downloader) error {
 				return nil
 			}),
 		}
+		if input.S3 {
+			opts = append(opts, pfssync.WithS3())
```
brycemcanally, line 262: Is the functional option necessary? It seems like you could just continue here?

src/server/pps/server/s3g_sidecar_test.go
```diff
@@ -837,3 +839,57 @@ func TestS3SkippedDatums(t *testing.T) {
...
+	if testing.Short() {
+		t.Skip("Skipping integration tests in short mode")
+	}
+	c, _, _ := initPachClient(t)
+
+	repo := tu.UniqueString(t.Name() + "_data")
+	require.NoError(t, c.CreateProjectRepo(pfs.DefaultProjectName, repo))
+	masterCommit := client.NewProjectCommit(pfs.DefaultProjectName, repo, "master", "")
+	for i := 0; i < 100; i++ {
+		// require.NoError(t, c.PutFile(masterCommit, fmt.Sprintf("%02d", i), io.LimitReader(strings.NewReader("ooooooo"), 2)))
```
brycemcanally, line 858: I'm assuming this should be removed.

src/server/pps/server/s3g_sidecar_test.go
```diff
@@ -837,3 +839,57 @@ func TestS3SkippedDatums(t *testing.T) {
 		testS3SkippedDatums(t, c, userToken, ns, projectName, func(p, r string) string { return r + "." + p })
 	})
 }
+
+// TestDontDownloadData tests that when a pipeline sets `S3: true` on an input,
+// the worker binary doesn't download any datums from that input to the worker.
+// This test limits the size of the worker pod's disk to a smaller size than the
+// input data exposed via a `S3: true` input. When the worker binary attempts to
+// download all 10GB of input, the test times out.
+func TestDontDownloadData(t *testing.T) {
```
brycemcanally, line 848: I am a bit concerned about adding a test with 10GB of data. Would an existence check of smaller files not be sufficient?

src/server/worker/datum/datum.go
```diff
@@ -258,6 +258,9 @@ func (d *Datum) downloadData(downloader pfssync.Downloader) error {
 				return nil
 			}),
 		}
+		if input.S3 {
```
brycemcanally, line 261: nit, would you move this above the opts setup.

msteffen: Done.

src/server/pps/server/s3g_sidecar_test.go
msteffen: I reduced it to 2GB of data for now. Just duplicating a slack question here: would it make sense for the test to check for data in `.scratch` directly? Or would that be too implementation-dependent?

src/server/pps/server/s3g_sidecar_test.go
msteffen: Yes :grimacing:, done!

src/server/worker/datum/datum.go
msteffen: I just didn't quite understand the control flow here; I've removed the functional option.

src/server/pps/server/s3g_sidecar_test.go
jrockway: Yeah, in the case of this bug the /pfs/input symlinks never get created, but we do still download all of the data, so the test has to either download enough data to blow up the machine (perhaps with the help of k8s disk limits), or it has to know "too much" about how the worker works today.

msteffen: Bit of an update: I may actually be required to scan `/pfs/.scratch`, as applying a storage limit to the container doesn't appear to do what we want. I tried reducing the amount of data written from 10GB to 2GB (which should still exceed the limit) and the test starting passing. The issue is that even though the container's ephemeral storage is limited to 1GB, `/pfs` is an `EmptyDir` mount that is unlimited:
```
...
Containers:
  user:
    ...
    Limits:
      ephemeral-storage:  1G
    Requests:
      ephemeral-storage:  1G
    ...
    Mounts:
      ...
      /pfs from pachyderm-worker (rw)
  ...
Volumes:
  ...
  pachyderm-worker:
    Type:       EmptyDir (a temporary directory that shares a pod's lifetime)
    Medium:
    SizeLimit:  <unset>
```

Previously, the test was just timing out because of the data size. I don't think there's a way to cap the size of the `pachyderm-worker` volume, so the `.scratch` approach it is?

msteffen: Test updated

brycemcanally (approved):
lgtm

PR #8742, opened 2023-04-12
[Jupyter] Display current notebook

bbonenfant:
Design:
image
My attempt at recreation:
image


jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
```diff
@@ -111,6 +112,7 @@ export const usePipeline = (
     };
     saveNotebookMetaData(sameMetadata);
   };
+  const currentNotebook = ppsContext?.notebookModel?.name ?? 'None';
```
chainlink, line 115: This should maybe be in `useEffect` ?

chainlink (approved):
Pending change to `useEffect`

chainlink:
I'm not sure which way is best here, I'll need to test out the new code. The new code looks more in line with what we have, but might be overkill in this case.

chainlink:
Tested! Let's ship it @bbonenfant

PR #8743, opened 2023-04-12
add a helm template to release artifacts for ddesktop local deploys

molinamelendezj:


molinamelendezj:
cc: @tybritten

tybritten:
Do we definitely want to include the notebook part or just the base helm template?

molinamelendezj:
> Do we definitely want to include the notebook part or just the base helm template?

I assumed so. was cool to see notebook setup and running locally. we can include it an improve any of this on future PRs but at least this template would be shipping on every release.

etc/deploy/dd/dd_notebook_base.yaml
```diff
@@ -0,0 +1,45 @@
...
+    matchLabels:
+      app: jupyter-notebook
+  template:
+    metadata:
+      labels:
+        app: jupyter-notebook
+    spec:
+      containers:
+        - name: pach-notebook
+          image: pachyderm/notebooks-user:vPACH_VERSION
```
djanicekpach, line 19: should this change `vPACH_VERSION -> "v${PACH_VERSION}"`?

djanicekpach (approved):

etc/deploy/dd/dd_notebook_base.yaml
molinamelendezj: Naw, it's just a place holder, the `sed` replaces it at release time.

chainlink:
Hey guys, just seeing this now, we shouldn't include the notebook portion (as it doesn't include storage, so the user will lose their work everytime) and the `sed` portion can be set with helm options (which is much safer)

chainlink:
We also shouldn't call this the "docker desktop" option, since it doesn't work in docker desktop (because of the underlying PVC issue with DD)

PR #8718, opened 2023-04-05
[Jupyter] PPS Metadata Format + Project Support

bbonenfant:
Format:
```
version: (str) version ID of the metadata format
config:
  image: (str) docker image (with tag)
  pipeline:
    name: (str) name of the pipeline
    project:
      name: (str) name of the pipeline's project
  requirements: (str) relative path to the requirements.txt file
  input_spec: (dict) JSON of a valid input spec
}
```

See jupyter-extension/jupyterlab_pachyderm/tests/data/TestNotebook.ipynb for example

Tasks:
- [x] Remove all SAME references
- [x] Update METADATA_KEY
- [x] Update BE to read new metadata format
- [x] Update FE to read/write new metadata format
- [x] Breakout input spec parsing into helper method
- [x] Add project support 

Future Tasks:
- [ ] Validate input spec
- [ ] Properly define an InputSpec type

Closes: https://pachyderm.atlassian.net/browse/INT-888 and https://pachyderm.atlassian.net/browse/INT-893 also potentially closes https://pachyderm.atlassian.net/browse/INT-908

PR #8744, opened 2023-04-12
[Jupyter] Switch to autosave in the Jupyter FE

msteffen:
Fix [INT-904]

[INT-904]: https://pachyderm.atlassian.net/browse/INT-904?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/jupyter-extension/jupyterlab_pachyderm/pps_client.py
+++ b/jupyter-extension/jupyterlab_pachyderm/pps_client.py
@@ -1,4 +1,5 @@
 import json
+import yaml
 import os.path
 from dataclasses import dataclass, asdict
 from datetime import datetime
@@ -42,7 +43,13 @@ class PPSClient:
         except ValueError as err:
             raise HTTPError(reason=f"Bad Request: {err}")
 
-        pipeline_spec = create_pipeline_spec(config, '...')
+        try:
+            pipeline_spec = create_pipeline_spec(config, '...')
+        except Exception as e:
+            raise HTTPError(
+                status_code=400,
+                reason="couldn't parse pipeline input: {repr(e)}"
+            )
         return json.dumps(pipeline_spec)
 
     async def create(self, path: str, body: dict):
@@ -77,8 +84,19 @@ class PPSClient:
         if config.requirements and not os.path.exists(config.requirements):
             raise HTTPError(reason="requirements file does not exist")
 
-        script, _resources = self.nbconvert.from_filename(str(path))
+        # Common error: user input spec isn't valid. Check that *before*
+        # creating the companion repo/branch (which would remain in case of a
+        # syntax error in the spec). As this is just for validation, don't store
+        # the result.
+        try:
+            create_pipeline_spec(config, '...')
+        except Exception as e:
+            raise HTTPError(
+                status_code=400,
+                reason=f"couldn't parse pipeline input: {repr(e)}"
+            )
 
+        script, _resources = self.nbconvert.from_filename(str(path))
         client = python_pachyderm.Client()  # TODO: Auth?
         try:  # Verify connection to cluster.
             client.inspect_cluster()
@@ -93,10 +111,25 @@ class PPSClient:
             )
 
         companion_branch = upload_environment(client, companion_repo, config, script.encode('utf-8'))
-        pipeline_spec = create_pipeline_spec(config, companion_branch)
-        client.create_pipeline_from_request(
-            python_pachyderm.parse_dict_pipeline_spec(pipeline_spec)
-        )
+        try:
+            pipeline_spec = create_pipeline_spec(config, companion_branch)
+        except Exception as e:
+            raise HTTPError(
+                status_code=500,  # shouldn't error after validation above
+                reason=f"error creating valid pipeline spec: {repr(e)}"
+            )
+        try:
+            client.create_pipeline_from_request(
+                python_pachyderm.parse_dict_pipeline_spec(pipeline_spec)
+            )
+        except Exception as e:
+            if hasattr(e, "details"):
+                # Common case: pull error message out of Pachyderm RPC response
+                raise HTTPError(status_code=400, reason=e.details())
+            raise HTTPError(
+                status_code=500,
+                reason=f"error creating pipeline: {repr(e)}"
+            )
 
         return json.dumps(
             dict(message="Create pipeline request sent. You may monitor its status by running"
@@ -165,7 +198,7 @@ def create_pipeline_spec(config: PpsConfig, companion_branch: str) -> dict:
     companion_repo = f"{config.pipeline.name}__context"
     input_spec = dict(
         cross=[
-            config.input_spec,
+            yaml.safe_load(config.input_spec),
             dict(
                 pfs=dict(
                     project=config.pipeline.project.name,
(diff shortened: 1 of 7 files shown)


chainlink:
Looks like we'll need to have a deeper look into the tests

chainlink:
This changes the metadata storage format in the notebook. Let's circle back on this after https://github.com/pachyderm/pachyderm/pull/8718 is merged, as that changes the format as well. The yaml/json parsing here is actually something we needed to do with SAME to stuff the whole input spec into one field (and saving as json for clarity instead of a YAML which is whitespace sensitive)

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
```diff
@@ -71,18 +70,26 @@ export const usePipeline = (
...
+      version: PPS_VERSION,
+      config: {
+        pipeline: pipeline,
+        image: imageName,
+        requirements: requirements,
+        input_spec: inputSpec,
+      },
+    };
+    saveNotebookMetaData(ppsMetadata);
+  }, [pipeline, imageName, requirements, inputSpec]);
```
msteffen, line 91: this block is now where autosave is implemented (rather than in the various `onChange` blocks in the component). This fixes a bug where `onChange` would trigger both `setInputSpec` and `callSavePipeline` and, because the latter was async, the two would sometimes race, and the pipeline spec would be saved without the most recent edit.

Now, React manages the update propagation, and `callSavePipeline` (which has been moved into the body of this lambda) is only run after `inputSpec` (and other fields) have been updated, eliminating the race. I've tested this locally and the update actually seems faster than before. Previously, there would sometimes be lag between typing and seeing changes to the notebook metadata, and now it seems consistently close to instant.

chainlink: Nice use of `useEffect` here to manage side effects

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
```diff
@@ -101,7 +108,10 @@ export const usePipeline = (
         }
       } catch (e) {
         if (e instanceof ServerConnection.ResponseError) {
-          setErrorMessage('Error creating pipeline');
+          // statusText is the only place that the user will get yaml parsing
+          // errors (though it will also include Pachyderm errors, e.g.
+          // "missing input repo").
+          setErrorMessage('Error creating pipeline: ' + e.response.statusText);
```
chainlink, line 114: We've gone back and forth a couple times on this code. I believe @bbonenfant looked into this more and couldn't get the error returned from the server?

chainlink:
@msteffen Can you update this to latest master?

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
msteffen: Yeah I'm a little torn--if we're doing the YAML parsing in python, it seems important that those errors get shown, so I wanted to display *something*. I got the code here from playing around in the browser console, as far as that goes. This is what it looks like:

msteffen: There's also a funny UX sort-of-bug that I'm looking into now (which you'll likely find if you start playing with the error reporting), which is that `yaml.safe_load()` seems much more flexible than I expected, and accepts stuff that I actually wish it would reject. I'm trying to figure out if I can parse the YAML into an input spec rather than a generic dict for error reporting purposes.

That issue doesn't seem catastrophic to me because `python_pachyderm` will still reject those specs when we try to turn the dict into a `CreatePipelineRequest`, but the error message is a little opaque

jupyter-extension/setup.cfg
```diff
@@ -18,6 +18,7 @@ classifiers =
 [options]
 packages = find:
 install_requires =
+  pyyaml>=6.0
```
bbonenfant, line 21: I don't think we need to change this, but just to note: in my experience `ruamel.yaml` is a more complete yaml parser than `pyyaml`. Since out input specs use basic yaml I don't think its an issue. Here's a reference for the differences: https://yaml.readthedocs.io/en/latest/pyyaml.html

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
```diff
@@ -115,30 +125,6 @@ export const usePipeline = (
     };
   }
 
-  const callSavePipeline = async () => {
-    setErrorMessage('');
-
-    let inputSpecJson;
-    try {
-      inputSpecJson = parseInputSpec(inputSpec);
```
bbonenfant, line 123: I believe you changes make `parseInputSpec` unneeded so you can delete. (GH won't let me comment on that section of code)

jupyter-extension/jupyterlab_pachyderm/pps_client.py
```diff
@@ -165,7 +198,7 @@ def create_pipeline_spec(config: PpsConfig, companion_branch: str) -> dict:
     companion_repo = f"{config.pipeline.name}__context"
     input_spec = dict(
         cross=[
-            config.input_spec,
+            yaml.safe_load(config.input_spec),
```
bbonenfant, line 201: rather than doing it here, I would do it within the `PpsConfig.from_notebook` method. That way this method shouldn't reasonably error and you can catch an invalid yaml on parsing.

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
bbonenfant: The only downside I see to this approach is that if/when we add a new field to this pipeline view that we would want to save, we need to remember to add it here.

bbonenfant (approved):
I've tested locally and it appears to work. I couldn't manage to get it into any bad state. I don't see any issue with moving the yaml parsing to the backend. Looks good to me!

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
msteffen: Done! Good catch.

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
msteffen: Yeah, makes sense. I added a comment to `types.ts` indicating that `usePipeline.tsx` needs to be kept in sync. In practice, hopefully they won’t drift too much because, without editing this block, a new field would never be written (hopefully an obvious bug)

jupyter-extension/src/plugins/mount/components/Pipeline/hooks/usePipeline.tsx
msteffen: Note for posterity: our Jira ticket about surfacing the right error message is [INT-934]

PR #8745, opened 2023-04-13
Rename fields

chainlink:
Closes https://pachyderm.atlassian.net/browse/INT-914

image


jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
```diff
@@ -95,7 +95,7 @@ const Pipeline: React.FC<PipelineProps> = ({
           className="pachyderm-pipeline-input-label"
           htmlFor="pipelineName"
         >
-          *Name:{'  '}
```
chainlink, line 98: @bbonenfant What does `{' '}` refer to here? is that just for spacing? Also, is the `*` in the description to denote required?

bbonenfant: Yes to both!

PR #8749, opened 2023-04-13
Validate All Commits passed to PFS APIs

acohen4:


src/server/pfs/server/val_server.go
```diff
@@ -217,34 +191,68 @@ func (a *validatedAPIServer) DiffFile(request *pfs.DiffFileRequest, server pfs.A
...
 
+func (a *validatedAPIServer) AddFileSet(ctx context.Context, req *pfs.AddFileSetRequest) (_ *types.Empty, retErr error) {
+	if err := checkCommit(req.Commit); err != nil {
+		return nil, err
+	}
+	return a.apiServer.AddFileSet(ctx, req)
+}
+
+func (a *validatedAPIServer) SubscribeCommit(request *pfs.SubscribeCommitRequest, stream pfs.API_SubscribeCommitServer) (retErr error) {
+	if request.From != nil {
```
brendoncarroll, line 213: Do we need this?  There is already a nil check in `checkCommit`

acohen4: request.From is not required, so we don't want to error out on a nil commit here

src/server/pfs/server/driver.go
```diff
@@ -628,7 +628,7 @@ func (d *driver) findCommits(ctx context.Context, request *pfs.FindCommitsReques
 			}
 			commit = inspectCommitResp.ParentCommit
 			return nil
-		}, zap.String("commit", commit.ID), zap.String("branch", commit.Branch.String()), zap.String("target", request.FilePath)); err != nil {
+		}, zap.String("commit", commit.ID), zap.String("branch", commit.GetBranch().String()), zap.String("repo", commit.Repo.String()), zap.String("target", request.FilePath)); err != nil {
```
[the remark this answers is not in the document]
acohen4: we're not. This was just a log line I figure I'd preserve. I can remove it

PR #8750, opened 2023-04-13
Make DeletePipelines() transactional

acohen4:
Part of CORE-1611

### DIFF
--- a/src/server/pfs/server/driver.go
+++ b/src/server/pfs/server/driver.go
@@ -440,12 +440,7 @@ func (d *driver) deleteRepos(ctx context.Context, projects []*pfs.Project) ([]*p
 func (d *driver) deleteReposInTransaction(txnCtx *txncontext.TransactionContext, projects []*pfs.Project) ([]*pfs.Repo, error) {
 	var ris []*pfs.RepoInfo
 	if err := d.listRepoInTransaction(txnCtx, false, "", projects, func(ri *pfs.RepoInfo) error {
-		ok, err := d.canDeleteRepo(txnCtx, ri.Repo)
-		if err != nil {
-			return err
-		} else if ok {
-			ris = append(ris, ri)
-		}
+		ris = append(ris, ri)
 		return nil
 	}); err != nil {
 		return nil, errors.Wrap(err, "list repos")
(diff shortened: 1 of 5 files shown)


src/server/pfs/server/driver.go
```diff
@@ -438,35 +438,44 @@ func (d *driver) deleteRepos(ctx context.Context, projects []*pfs.Project) ([]*p
...
+		ris = append(ris, ri)
 		return nil
 	}); err != nil {
 		return nil, errors.Wrap(err, "list repos")
 	}
-	if len(repoInfos) == 0 {
+	return d.deleteReposHelper(txnCtx, ris, false)
+}
+
+func (d *driver) deleteReposHelper(txnCtx *txncontext.TransactionContext, ris []*pfs.RepoInfo, force bool) ([]*pfs.Repo, error) {
```
brendoncarroll, line 451: Is this recursive?  Why is this a separate method if it's only called once and doesn't have a descriptive name?

src/server/pps/server/api_server.go
```diff
@@ -3014,7 +2970,7 @@ func (a *apiServer) DeletePipelines(ctx context.Context, request *pps.DeletePipe
 			Force:    request.Force,
 			KeepRepo: request.KeepRepo,
 		}
-		pp []*pps.Pipeline
+		ps []*pps.Pipeline
```
brendoncarroll, line 2973: 👍  lol

src/server/pfs/server/driver.go
brendoncarroll: I see that this is also called from the other file.  It seems like `deleteReposInTransaction` and `DeleteReposInTransaction` are the same things?  When would we use one vs. the other?

acohen4: it's an internal method that multiple functions rely on. Does Helper imply recursive? Happy to call this deleteReposInternal()

brendoncarroll: >  Does Helper imply recursive
It doesn't really imply anything.  It's not really descriptive, which is why I was trying to figure out why it was split out.

Either it has to be a separate function because it's recursive, or has multiple callers, or it's split out to make the code clearer.  With a name like `*Helper` it isn't really the latter, so it should be one of the first two.  I didn't see the multiple call sites at first.

brendoncarroll (approved):

PR #8751, opened 2023-04-14
[Jupyter] Add default pipeline image

bbonenfant:
The mount extension now hooks into the builtin `ISettingRegistry`. I used [this repo](https://github.com/jupyterlab/extension-examples/tree/master/settings) as a guide for implementation.

Users can set the value of `defaultPipelineImage` in two ways:
- setting the value in `<sys-prefix>/share/jupyter/lab/settings/overrides.json` where `sys-prefix` is the directory of there virtual environment where the extension is installed (venv).
```json
{
  "jupyterlab-pachyderm:mount":  {
    "defaultPipelineImage": "org/image:tag"
  }
}
```

- updating the value within the jupyter UI under `Settings > Advanced Settings Editor`
image

## Notes
- We can set this value to whatever we want, I just chose `python:3.10` as a non-controversial option.
- I don't think we'd be able to "detect what image the notebook server is running" very easily, and even if we could I don't know if its useful for that to be the default value.

### DIFF
--- a/jupyter-extension/package.json
+++ b/jupyter-extension/package.json
@@ -18,6 +18,7 @@
   },
   "files": [
     "lib/**/*.{d.ts,eot,gif,html,jpg,js,js.map,json,png,svg,woff2,ttf}",
+    "schema/**/*.json",
     "style/**/*.{css,js,eot,gif,html,jpg,json,png,svg,woff2,ttf}"
   ],
   "main": "lib/index.js",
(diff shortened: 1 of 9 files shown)


chainlink:
Nice one!

chainlink:
Tested this branch locally. The default value appears correctly in the pipeline extension

chainlink (approved):

PR #8758, opened 2023-04-14
Add FetchCache data structure for branch DAG internals

acohen4:


brendoncarroll:
I would prefer that we use Hashicorp's LRU cache, which we use elsewhere.  The latest version is also generic.
https://github.com/hashicorp/golang-lru

The FetchCache in this PR doesn't do eviction, so it has unbounded memory usage.

brendoncarroll (changes requested):
I would prefer that we use Hashicorp's LRU cache, which we use elsewhere. The latest version is also generic.
https://github.com/hashicorp/golang-lru

The FetchCache in this PR doesn't do eviction, so it has unbounded memory usage.

PR #8759, opened 2023-04-17
Add an archive download server

jrockway:
This is CORE-1584.

Console wants to let users download files as a ZIP archive after checking a checkbox for each file in the UI.  This PR implements a service that handles those downloads.  Each download is encoded in a special way (see  https://www.notion.so/2023-04-03-HTTP-file-and-archive-downloads-cfb56fac16e54957b015070416b09e94 if you work at Pachyderm).  The server is a plain HTTP server on a new port, and this PR includes the necessary configuration to hook it into the proxy.  End-users will not know that it's a separate HTTP server; everything thru the proxy continues to work normally.

We resolve each branch reference at the start of the request, so changing branch heads mid-download doesn't generate weird sets of output.  As a side effect, everything ends up in /project/repo/commit/file instead of /project/repo/branch or commit/file.  This was previously a "nice to have" in the description.

I added some XSRF protection to this endpoint, though it's not necessary right now.  In the future, where a user is logged in via OAuth+cookie, we don't want websites to embed an image tag for http://pachyderm.example.com/pfs/default/images/master/foo.png and exfiltrate that data with javascript;  XSRF protection prevents this.  It doesn't matter right now because there are no ephemeral credentials that the browser has access to; if they know your auth-token, sure, let them iframe a download or whatever.

I also added extensive tests, including Fuzz tests that I ran for about an hour each.

The doc talks about implementing a `/pfs` route in addition to a `/download` route.  We don't do that here.  The doc talks about OAuth-ing `/download` requests; we don't do that, a header must be included.  Everything else works like the doc says, though.  If `proxy.host` is `localhost`, then something like `http://localhost/download/AAAAA.zip` should get you an archive of files.

If you want to test this without rebuilding Pachyderm, just run `go run ./src/server/cmd/local-download-server` and direct your requests to `http://localhost:1659`.

I made some slightly-unrelated changes in this PR:

* Add annotations to the proxy deployment so that it is restarted when we regenerate envoy.json (or the user switches from non-TLS to TLS or vice-versa).
* Add a log.Exit for scripts.  I added a small local server that implements the file-download HTTP server so you don't have to rebuild/reinstall Pachyderm to work on the code, since it just uses RPCs internally to actually get the files.  log.Exit is convenient for this style of app, which we should have a lot more of.
* Added an exception to log.HTTP to not log Envoy health checks, because customer-eng hates health check logging.
* Shortened the request URL to 16KiB in HTTP logs, because we expect download URLs that are up to 2MB in size and that would be annoying to see on every log line.

There are some defects that need to be addressed before merging:

* ~~We need to add a `pachyderm_test` test that downloads real files thru the proxy.  I tested this manually, of course.~~  (Added.)
* We need to test larger URLs, larger downloads and make sure Envoy's buffering settings are correct for this use case.

There are some nice-to-haves not implemented:
* I added the URL encoding code, but nothing in Pachyderm uses it.  We should hook it into pachctl, or maybe generate a wasm module with tinygo so console doesn't have to implement zstd in Javascript.  (I tried tinygo -> wasm, it's pretty unusable!  This uses a lot of memory and tinygo's GC runs terribly in wasm.)
* We might need CORS support if console goes the header instead of request param route.

@pleeko let the integration begin!

@armaanv I added you as a reviewer because JIRA lists you as the secondary on this project.  @acohen4 went to more meetings about this than you, though, so don't feel obligated to review quite yet!

Here's an example exchange that works if you ran the OpenCV example in the default project:

```
$ rm -f /tmp/test.zip; curl -v localhost/download/ASi1L_0EaHUBAEQCZGVmYXVsdC9pbWFnZXNAbWFzdGVyOi8AbW9udGFnZS5wbmcAAxQEBQPYsGPLbFDb.zip\?authn-token=V4Ptmxchcj04TY5vmngJAD0RiuJ3JYc6 -o /tmp/test.zip; unzip -l /tmp/test.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 127.0.0.1:80...
* Connected to localhost (127.0.0.1) port 80 (#0)
> GET /download/ASi1L_0EaHUBAEQCZGVmYXVsdC9pbWFnZXNAbWFzdGVyOi8AbW9udGFnZS5wbmcAAxQEBQPYsGPLbFDb.zip?authn-token=V4Ptmxchcj04TY5vmngJAD0RiuJ3JYc6 HTTP/1.1
> Host: localhost
> User-Agent: curl/8.0.1
> Accept: */*
>
< HTTP/1.1 200 OK
< content-disposition: attachment; filename=pachyderm-download-2023-04-25T23:35:12Z.zip
< content-type: application/zip
< date: Tue, 25 Apr 2023 23:35:12 GMT
< x-envoy-upstream-service-time: 71
< server: envoy
< transfer-encoding: chunked
<
{ [61476 bytes data]
100 1595k    0 1595k    0     0  17.2M      0 --:--:-- --:--:-- --:--:-- 17.3M
* Connection #0 to host localhost left intact
Archive:  /tmp/test.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
    58644  04-25-2023 23:35   default/images/0c49d0fd81804b9eb3c84977198c1186/46Q8nDz.jpg
    80588  04-25-2023 23:35   default/images/0c49d0fd81804b9eb3c84977198c1186/8MN9Kg0.jpg
   104836  04-25-2023 23:35   default/images/0c49d0fd81804b9eb3c84977198c1186/g2QnNqa.jpg
  1411573  04-25-2023 23:35   default/montage/db4e824391684e00aa136c5285ffad19/montage.png
---------                     -------
  1655641                     4 files
```


BOsterbuhr:
> Added an exception to log.HTTP to not log Envoy health checks, because customer-eng hates health check logging

Thank you!

etc/generate-envoy-config/pachyderm-services.libsonnet
```diff
@@ -123,6 +123,34 @@
...
+    internal_port: 1659,
+    service: 'pachd-proxy-backend',
+    routes: [
+      {
+        match: {
+          prefix: '/download',
+        },
+        route: {
+          cluster: 'pachd-download',
+          idle_timeout: '600s',
```
armaanv, line 136: why have such a long idle timeout here?

src/internal/log/http.go
```diff
@@ -28,7 +29,17 @@ func AddLoggerToHTTPServer(rctx context.Context, name string, s *http.Server) {
 				requestID = append(requestID, uuid(interactiveTrace).String())
 			}
 			id := zap.Strings("x-request-id", requestID)
-			Info(ctx, "incoming http request", zap.Stringer("url", r.URL), zap.String("method", r.Method), zap.String("host", r.Host), id)
+
+			if ua := r.Header.Get("user-agent"); ua != "Envoy/HC" {
+				// Print info about the request if this is not an Envoy health check.
```
armaanv, line 34: nice!

src/internal/archiveserver/encode.go
```diff
@@ -0,0 +1,166 @@
...
+	case refState:
+		return "ref"
+	case fileState:
+		return "file"
+	}
+	panic("unknown DecodeV1 state")
+}
+
+// DecodeV1Path decodes a V1 path.
+func DecodeV1Path(path string) (*pfs.File, error) {
```
armaanv, line 114: Is this going to be replaced before merging?

armaanv (approved):

etc/generate-envoy-config/pachyderm-services.libsonnet
jrockway: Just paranoia about reading taking a long time.  I applied the same settings as with pachd-grpc.

src/internal/archiveserver/encode.go
jrockway: I cleaned this up.

PR #8765, opened 2023-04-18
Adjust mount-server auth flow to authenticate pipeline extension

kevyang:
Updates the auth flow in the Jupyter extension to accommodate authentication when running in sidecar mode. Previously, the pipeline extension only works unauthenticated or with the mount-server running in subprocess mode because the Jupyter extension itself did not have access to the session token that the mount-server negotiates upon login when the processes do not share the same container. This change updates the auth flow so that the mount-server auth is done in two steps, returning a redirect to the login page in the first and returning the Pachyderm configuration along with the session token in the second step. This enables the Jupyter extension to write the same configuration into its own container and use those for creating pipelines in Pachyderm deployments with authentication enabled.

Jira: [INT-895]

[INT-895]: https://pachyderm.atlassian.net/browse/INT-895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jupyter-extension/jupyterlab_pachyderm/mount_server_client.py
```diff
@@ -179,6 +182,21 @@ async def auth_login(self):
         response = await self.client.fetch(
             f"{self.address}/auth/_login", method="PUT", body="{}"
         )
+
+        resp_json = json.loads(response.body.decode())
+        # may bubble exception up to handler if oidc_state not in response
+        oidc = resp_json['oidc_state']
+
+        asyncio.create_task(self.auth_login_token(oidc))
```
smalyala, line 190: Does this need to be awaited?

smalyala (approved):
Left a question but otw LGTM!

jupyter-extension/jupyterlab_pachyderm/mount_server_client.py
kevyang: no, we specifically do not want to await this since we want to redirect the user to the login page without waiting for their login to complete. i believe `auth_login_token` will only finish once the user completes the login. this is similar to what is done before within the mount server by creating the authenticate go routine and updating the token without waiting on it.

kevyang: perhaps i should capture this in a comment

jupyter-extension/jupyterlab_pachyderm/mount_server_client.py
```diff
@@ -12,6 +14,7 @@
 
 lock = locks.Lock()
 MOUNT_SERVER_PORT = 9002
+PACH_CONFIG_DIR = '/home/jovyan/.pachyderm/config.json'
```
chainlink, line 17: I believe the `$NB_USER` should have the current username https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#user-related-configurations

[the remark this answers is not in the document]
kevyang: ah good call. is this preferable to the suggestion that @chainlink made?

jupyter-extension/jupyterlab_pachyderm/mount_server_client.py
```diff
@@ -179,6 +182,21 @@ async def auth_login(self):
...
+
+        asyncio.create_task(self.auth_login_token(oidc))
+        return response.body
+
+    async def auth_login_token(self, oidc):
+        response = await self.client.fetch(
+            f"{self.address}/auth/_login_token", method="PUT", body=f'{oidc}'
+        )
+        os.makedirs(os.path.dirname(PACH_CONFIG_DIR), exist_ok=True)
+        with open(PACH_CONFIG_DIR, 'w') as f:
```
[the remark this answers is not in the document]
kevyang: i don't think so, but i could be wrong. any time we overwrite it, we would simply be updating it with the configuration that exists in the mount-server, and i can't think of a situation where we would want the configurations to be different (at least not yet).

jupyter-extension/jupyterlab_pachyderm/mount_server_client.py
```diff
@@ -179,6 +182,21 @@ async def auth_login(self):
...
+
+        resp_json = json.loads(response.body.decode())
+        # may bubble exception up to handler if oidc_state not in response
+        oidc = resp_json['oidc_state']
+
+        asyncio.create_task(self.auth_login_token(oidc))
+        return response.body
+
+    async def auth_login_token(self, oidc):
+        response = await self.client.fetch(
```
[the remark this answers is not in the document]
kevyang: good point. i am thinking we can just use response.rethrow() for this as that seems like what we would want to do after reading through the documentation. but i am not really an expert on python http so let me know if this is not correct.

brendoncarroll (approved):
I updated CODEOWNER so it won't auto-assign PFS in the future.
https://github.com/pachyderm/pachyderm/pull/8778

jupyter-extension/jupyterlab_pachyderm/mount_server_client.py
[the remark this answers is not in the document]
bbonenfant: Should we have a check at least that we are in sidecar mode? At the moment if I run this locally wont it overwrite my config.json?

jupyter-extension/jupyterlab_pachyderm/mount_server_client.py
```diff
@@ -187,18 +187,60 @@ async def auth_login(self):
...
+
         return response.body
 
+    # updates the pachyderm config with the session token of the mount server
+    # config. this will try to insert the token into the mount_server context
+    # if it already exists, or copies the mount server config if it doesn't.
+    # then, it switches the active context over to the mount_server context.
+    def write_token_to_config(self, pach_config_dir, mount_server_config_str):
+        get_logger().info("write_token_to_config")
+        config_file = open(pach_config_dir)
```
bbonenfant, line 226: I would use a `with` context here rather than manually closing the file

jupyter-extension/jupyterlab_pachyderm/mount_server_client.py
```diff
@@ -187,18 +187,60 @@ async def auth_login(self):
...
+            with open(pach_config_dir, 'w') as f:
+                f.write(response.body.decode())
+
         return response.body
 
+    # updates the pachyderm config with the session token of the mount server
+    # config. this will try to insert the token into the mount_server context
+    # if it already exists, or copies the mount server config if it doesn't.
+    # then, it switches the active context over to the mount_server context.
+    def write_token_to_config(self, pach_config_dir, mount_server_config_str):
```
bbonenfant, line 224: it appears based on its usage that `pach_config_dir` is a file? Should it be named differently? Nonetheless it would be helpful to have a docstring that documents these arguments. The above comment can be moved into the docstrnig as well

bbonenfant (commented):
It might be worth breaking this `write_token_to_config` method out of the class (it doesn't look like it uses the `self` argument) so that we can write some tests for it. Also we should probably get another review from @chainlink on this new change

chainlink (approved):
Testing and working in both sidecar and non sidecar modes!

PR #8766, opened 2023-04-18
pfs: record project creation date; display in pachctl

jrockway:
This does what it says on the tin.

If the project was created before project creation date tracking, then we return nil, which we display as "-" in pachctl.


src/server/pfs/server/driver.go
```diff
@@ -649,6 +649,7 @@ func (d *driver) createProjectInTransaction(txnCtx *txncontext.TransactionContex
 	if err := projects.Create(pfsdb.ProjectKey(req.Project), &pfs.ProjectInfo{
 		Project:     req.Project,
 		Description: req.Description,
+		CreatedAt:   types.TimestampNow(),
```
brendoncarroll, line 652: I don't see an import statement in the diff so I assume this `types` package is the right way to do this.

brendoncarroll (approved):
LGTM

loganripplinger (approved):
I think that this will work for Console. A nil value for no time is fine. We can just treat that as the smallest time when we sort projects. Thank you!

PR #8767, opened 2023-04-19
Switch from "view" buttons to tabs

msteffen:
Todo:
- [ ] Wrong shade of gray in the tab bar
- [ ] double-width border on the left side of the "Explore" tab
- [ ] number medalions look dumb -- remove them


msteffen:
We should not merge this, because it'll require significant changes to move behavior that's currently in the `useDatum` hook. But maybe we can make similar UI changes that suggest the same workflow

PR #8769, opened 2023-04-19
do not wrap "not found" around non-notfound errors

djanicekpach:
Because out errors.IsNotFound checks for string containment of "not found", and we wrap some low level errors with the text "not found" we are swallowing errors we should be handling differently. This removes the string "not found" from those wrappers. Locust testing suggests this fixes some errors with concurrent pipeline deletion.


acohen4 (approved):
LGTM

PR #8770, opened 2023-04-19
Rename top links

msteffen:
Notes:
- [x] The fonts are slightly different--I'll have to figure out how to fix that
- I also added an `overscroll: auto` style to the pipeline extension, to avoid the "error has caused the view to overflow" issue. I can remove this if it's controversial, but I didn't want to simply forget about it.

Jira: [INT-924]

[INT-924]: https://pachyderm.atlassian.net/browse/INT-924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
+++ b/jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
@@ -10,7 +10,8 @@ type PipelineProps = {
   saveNotebookMetadata: (metadata: PpsMetadata) => void;
 };
 
-const placeholderInputSpec = `pfs:
+const placeholderInputSpec = `# example:
+pfs:
   repo: images
   branch: dev
   glob: /*
@@ -182,7 +183,7 @@ ${inputSpec
           className="pachyderm-button"
           onClick={callCreatePipeline}
         >
-          Run
+          Publish Pipeline
         </button>
         &nbsp;
         <button
--- a/jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
+++ b/jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
@@ -41,10 +41,7 @@ const Pipeline: React.FC<PipelineProps> = ({
   } = usePipeline(ppsContext, settings, saveNotebookMetadata);
 
   return (
-    <div
-      className="pachyderm-mount-pipeline-base"
-      style={{overscrollY: 'auto'}}
-    >
+    <div className="pachyderm-mount-pipeline-base">
       <div className="pachyderm-mount-pipeline-back">
         <button
           data-testid="Pipeline__back"
@@ -61,7 +58,7 @@ const Pipeline: React.FC<PipelineProps> = ({
         </button>
       </div>
       <span className="pachyderm-mount-pipeline-subheading">
-        Publish Notebook as Pipeline
+        Publish as Pipeline
       </span>
 
       <div className="pachyderm-pipeline-current-notebook-wrapper">
--- a/jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
+++ b/jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
@@ -41,7 +41,10 @@ const Pipeline: React.FC<PipelineProps> = ({
   } = usePipeline(ppsContext, settings, saveNotebookMetadata);
 
   return (
-    <div className="pachyderm-mount-pipeline-base">
+    <div
+      className="pachyderm-mount-pipeline-base"
+      style={{overscrollY: 'auto'}}
+    >
       <div className="pachyderm-mount-pipeline-back">
         <button
           data-testid="Pipeline__back"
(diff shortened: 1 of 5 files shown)


bbonenfant (approved):
I tested this locally and it seems to work. Two things I noticed though:
- The "Explore" text isn't aligned with the text of the buttons (I don't think this is super important to get perfect).
- I don't think your changes caused this, but I think the config view is inescapable? I clicked on the gear and then couldn't get back to any other view without reloading the page

bbonenfant:
Tests are failing because of the linter check. This is the relevant portion:
```
/home/circleci/project/jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
  46:15  error  Replace `·overscrollY:·'auto'·` with `overscrollY:·'auto'`  prettier/prettier
```

msteffen:
> * I don't think your changes caused this, but I think the config view is inescapable? I clicked on the gear and then couldn't get back to any other view without reloading the page

I ran into this too —I thought it was my local setup, but I guess not; I’ll look into it. The back button is hidden if it’s not getting 200s from pachd (it assumes you’ve lost your connection to the cluster and wants you to reconfigure it so it “jumps” you to the settings screen and hides the other buttons). The issue must be related to that not working

bbonenfant:
> > * I don't think your changes caused this, but I think the config view is inescapable? I clicked on the gear and then couldn't get back to any other view without reloading the page
> 
> I ran into this too —I thought it was my local setup, but I guess not; I’ll look into it. It looks like the back button is hidden if it’s not getting 200s from pachd (it assumes you’ve lost your connection to the cluster and wants you to reconfigure it so it “jumps” you to the settings screen and hides the other buttons). The issue must be related to that not working.

Oh, my local setup is probably causing this issue as well then. I don't think the mount-server is happy when I run jupyter locally because it's trying to mount to `/pfs` and macos doesn't want it to. I've just never dove in deeper since I was doing pps work and not mount-server work. Might be worth creating a ticket to investigate this

jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
```diff
@@ -41,7 +41,7 @@ const Pipeline: React.FC<PipelineProps> = ({
   } = usePipeline(ppsContext, settings, saveNotebookMetadata);
 
   return (
-    <div className="pachyderm-mount-pipeline-base">
+    <div className="pachyderm-mount-pipeline-base" style={{overflow: 'auto'}}>
```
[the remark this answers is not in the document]
msteffen: Should I create a new class? Or do you think this should apply to the other views?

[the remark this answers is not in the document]
chainlink: Looks like the only other place that class is used is in the splash screen (displayed when you have the pipeline screen active, but then select a non notebook tab. So, I'd set it on the class, then check that screen to be sure

jupyter-extension/src/plugins/mount/mount.tsx
```diff
@@ -139,43 +139,43 @@ export class MountPlugin implements IMountPlugin {
...
+              >
+                Test{' '}
+              </button>
+              <button
+                className="pachyderm-button-link"
+                onClick={() => this.setShowPipeline(true)}
+              >
+                Publish{' '}
+                <sup
+                  style={{
```
chainlink, line 160: Can you make class for this?

jupyter-extension/src/plugins/mount/mount.tsx
```diff
@@ -139,43 +139,43 @@ export class MountPlugin implements IMountPlugin {
...
-                  className="pachyderm-button-link"
-                  onClick={() => this.setShowPipeline(true)}
+              <div className="pachyderm-mount-base-title">Explore</div>
+              <button
+                className="pachyderm-button-link"
+                data-testid="Datum__mode"
+                onClick={() => this.setShowDatum(true)}
+                style={{
+                  marginRight: '0.25rem',
+                  fontFamily: ['Montserrat', 'sans-serif'],
```
chainlink, line 149: The font family should be captured in the classes, also, we probably don't want to change the font family for consistency and cross browser availability.

chainlink (approved):
Pre-approving with proposed changes

jupyter-extension/src/plugins/mount/mount.tsx
msteffen: Done.

jupyter-extension/src/plugins/mount/mount.tsx
msteffen: I have moved it to a new css class, though I hesitate the remove it completely. The issue is that the "misalignment" @bbonenfant referred to (which you can also see in the screenshot) is actually a difference in font. Changing the font of this title removes the incongruence.

Let me know what you think. I'll attach "before" and "after" screenshots; I do think it looks better with the matching font. If you think there's a different overall solution that we should use here, though, let me know

jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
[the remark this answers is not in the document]
msteffen: Done.

jupyter-extension/src/plugins/mount/mount.tsx
msteffen: Done.

PR #8774, opened 2023-04-20
[Jupyter] Run button saves notebook to disk

bbonenfant:
I opted to use two separate methods since `saveNotebookMetaData` wasn't async but `saveNotbookToDisk` needs to be. I'm happy to make `saveNotebookMetaData` async and add a `toDisk` arg if we want to consolidate.

Additionally I extracted the `buildMetadata` method out of the `callSavePipeline`.


bbonenfant:
CI is failing from a flaky test. Nothing I've done should be causing an error like that when deleting the test project

chainlink (approved):

PR #8775, opened 2023-04-24
Add error codes and extra logging to the JupyterLab-Pachyderm extension backend

msteffen:


jupyter-extension/jupyterlab_pachyderm/handlers.py
```diff
@@ -274,17 +275,32 @@ class PPSCreateHandler(BaseHandler):
 
     @tornado.web.authenticated
     async def get(self, path):
-        """Get the pipeline spec for the specified notebook."""
-        response = await self.pps_client.generate(path)
-        await self.finish(response)
+        try:
+            """Get the pipeline spec for the specified notebook."""
```
bbonenfant, line 279: The docstring should be the first line of the function

jupyter-extension/jupyterlab_pachyderm/handlers.py
```diff
@@ -274,17 +275,32 @@ class PPSCreateHandler(BaseHandler):
...
+            response = await self.pps_client.generate(path)
+            get_logger().debug(f"GetPipelineSpec: {response}")
+            await self.finish(response)
+        except Exception as e:
+            if isinstance(e, tornado.web.HTTPError):
+                # Common case: only way to print the "reason" field of HTTPError
+                get_logger().error(f"couldn't create pipeline: {e.reason}")
+            get_logger().error("\n".join(traceback.format_exception(e)))
```
bbonenfant, line 287: Is this supposed to be in an else block?

jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import {closeIcon} from '@jupyterlab/ui-components';
 import {usePipeline} from './hooks/usePipeline';
-import {PpsContext, PpsMetadata, MountSettings} from '../../types';
+import {PpsContext, PpsMetadata, MountSettings} from 'plugins/mount/types';
```
bbonenfant, line 4: I don't think these changes are relevant to the PR?

bbonenfant (commented):
This seems overall seems fine to me. Could you pull master into this branch, it seems like tests you probably fixed in your other PR are failing here. I want to verify these error codes are being sent before approving. I seem to remember these codes not working as expected when I tried to set them. If they are being set correctly, then `test_integrations.py::test_pps_validation_errors` should fail

jupyter-extension/jupyterlab_pachyderm/handlers.py
msteffen: Oops, thanks! Done.

jupyter-extension/jupyterlab_pachyderm/handlers.py
msteffen: I deliberately put it outside the `if`, so that we'd log the trace no matter what (while I was figuring out how to pass errors through to the UI, that was my best tool for understanding what was going on). Happy to move it to an `else` or format it differently or something?

jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
msteffen: They're not; it's just a change a pulled out of autosave. I can pull this into its own PR

bbonenfant (approved):
Looks great!

PR #8779, opened 2023-04-24
[Helm] Give all users projectCreator role at the cluster level

albscui:


etc/helm/pachyderm/values.yaml
```diff
@@ -523,7 +523,9 @@ pachd:
   #   - repoReader
   #   robot:eve:
   #   - repoWriter
-  pachAuthClusterRoleBindings: {}
+  pachAuthClusterRoleBindings:
+    allClusterUsers:
+      - projectCreator # by default, everyone should be able to create projects
```
molinamelendezj, line 528: No need for projectOwner?

molinamelendezj (approved):

etc/helm/pachyderm/values.yaml
```diff
@@ -523,7 +523,9 @@ pachd:
   #   - repoReader
   #   robot:eve:
   #   - repoWriter
-  pachAuthClusterRoleBindings: {}
+  pachAuthClusterRoleBindings:
```
acohen4, line 526: I forgot that we currently ship empty role bindings. I think this is ok - but if a user is currently activating auth via git ops, but is setting cluster role bindings via API, we will be overwriting those. I think we need to make sure this is what we want

etc/helm/pachyderm/values.yaml
albscui: No need for projectOwner. We want users to create projects, and become owners, but don't want to automatically make everyone owner by default.

albscui:
Closed in favor of: https://github.com/pachyderm/pachyderm/pull/8791

PR #8780, opened 2023-04-25
[CORE-1597] Lock down the QueryLoki RPC

armaanv:
This PR adds a new permission, `CLUSTER_GET_LOKI_LOGS`, that allows the user to read logs through the QueryLoki RPC. This permission is added to the debugger role and a new `LokiLogReader` role.


jrockway (approved):
This looks great.

PR #8785, opened 2023-04-26
Upgrade envoy

jrockway:
This upgrades us to 1.25.5; we can't use 1.26.0 because it's broken on their end.  1.26.1 will fix that issue though.


acohen4 (approved):
LGTM but weird with the 2.5.5 thing. Where can I learn about the use of Stream ID?

jrockway:
@acohen4 Read about the stream ID here: https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage

It says it "should" replace x-request-id, but I kept x-request-id for now.

PR #8786, opened 2023-04-26
loki: upgrade to latest chart, and latest app version

jrockway:
For some reason their charts no longer set a version, so we now add an image.tag in values.yaml.

I tested this with the various pachctl commands (logs, loki), and verified that promtail still sends logs.  It all looks good.


acohen4 (approved):
LGTM

PR #8787, opened 2023-04-26
Implement datum dotenv file functionality

brycemcanally:
This PR implements functionality that will expose a dotenv file (https://pypi.org/project/python-dotenv/) through the local filesystem (at `/pfs/.env`) for user code during datum processing. This functionality was designed with the goal of providing datum environment variables when datum batching is enabled, but it is going to be available for pipelines without datum batching enabled as well. When datum batching isn't enabled, the environment variables are still going to be set appropriately, but the dotenv file will be available for use by user code and be stored in the meta commit for provenance tracking.

### DIFF
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -6584,6 +6584,7 @@ func TestMetaRepoContents(t *testing.T) {
 		}))
 		expectedFiles := map[string]struct{}{
 			fmt.Sprintf("/meta/%v/meta", datumID):                      {},
+			fmt.Sprintf("/pfs/%v/.env", datumID):                       {},
 			fmt.Sprintf("/pfs/%v/%v/%v", datumID, dataRepo, inputFile): {},
 			fmt.Sprintf("/pfs/%v/out/%v", datumID, inputFile):          {},
 		}
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -9877,7 +9877,7 @@ func TestMalformedPipeline(t *testing.T) {
 		Input:     client.NewProjectPFSInput(pfs.DefaultProjectName, "out", "/*"),
 	})
 	require.YesError(t, err)
-	require.Matches(t, "input cannot be named \"out\"", err.Error())
+	require.Matches(t, "input cannot be named out", err.Error())
 
 	_, err = c.PpsAPIClient.CreatePipeline(c.Ctx(), &pps.CreatePipelineRequest{
 		Pipeline:  client.NewProjectPipeline(pfs.DefaultProjectName, pipelineName),
@@ -9885,7 +9885,7 @@ func TestMalformedPipeline(t *testing.T) {
 		Input:     &pps.Input{Pfs: &pps.PFSInput{Name: "out", Repo: dataRepo, Glob: "/*"}},
 	})
 	require.YesError(t, err)
-	require.Matches(t, "input cannot be named \"out\"", err.Error())
+	require.Matches(t, "input cannot be named out", err.Error())
 
 	_, err = c.PpsAPIClient.CreatePipeline(c.Ctx(), &pps.CreatePipelineRequest{
 		Pipeline:  client.NewProjectPipeline(pfs.DefaultProjectName, pipelineName),
--- a/src/server/pachyderm_test.go
+++ b/src/server/pachyderm_test.go
@@ -11641,4 +11641,23 @@ func TestDatumBatching(t *testing.T) {
 		request := createPipelineRequest(pipeline, script)
 		checkState(request, pps.JobState_JOB_FAILURE)
 	})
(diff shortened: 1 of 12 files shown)


src/server/pps/server/api_server.go
```diff
@@ -173,6 +179,19 @@ func validateNames(names map[string]bool, input *pps.Input) error {
 	return nil
 }
 
+func validateName(name string) error {
+	if name == "" {
+		return errors.Errorf("input must specify a name")
+	}
+	switch name {
+	case common.OutputPrefix:
```
jrockway, line 187: I think this can be `case common.OutputPrefix, common.EnvFileName:` and avoid the fallthrough.

src/server/worker/datum/datum.go
```diff
@@ -235,6 +240,24 @@ func (d *Datum) withData(cb func() error) (retErr error) {
 	})
 }
 
+func (d *Datum) createEnvFile() (retErr error) {
+	var envStr string
+	for _, e := range d.env {
+		envStr += e + "\n"
+	}
+	f, err := os.Create(path.Join(d.PFSStorageRoot(), common.EnvFileName))
```
jrockway, line 248: You can golf this down to `os.WriteFile(filepath.Join(...), []byte(envStr), <permissions>)`.  Also, `path.Join` is for URLs, and `filepath.Join` is for the OS that the code is running on.  Obviously we don't run on an OS that can't take slash-separated filenames, but I think `filepath` is technically more correct.

src/worker/worker.proto
jrockway, line 1: I think "make proto" is required, so the comments get copied into the generated files.  That way the next person that edits these files isn't confused by an unexpectedly-large diff.  (I guess we should have a CI test that tests this.)

src/server/worker/driver/driver_windows.go
```diff
@@ -12,6 +13,7 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/log"
 	"github.com/pachyderm/pachyderm/v2/src/server/worker/common"
+	"github.com/pachyderm/pachyderm/v2/src/server/worker/logs"
```
jrockway, line 16: I guess none of us run on Windows and so goimports didn't add this.  Oops!  We should probably test Windows in CI if people are actually using it.

src/server/pachyderm_test.go
```diff
@@ -11641,4 +11641,23 @@ func TestDatumBatching(t *testing.T) {
 		request := createPipelineRequest(pipeline, script)
 		checkState(request, pps.JobState_JOB_FAILURE)
 	})
+	t.Run("Env", func(t *testing.T) {
+		// This test will error if the datum id environment variable isn't set and distinct for each datum.
+		script := fmt.Sprintf(`
+			set -a
```
jrockway, line 11647: Not a required change; but I like to start shell scripts with "set -euo pipefail" just to find unexpected weirdness as quickly as possible.  (-e makes the script exit as soon as there is a non-zero command, -u makes unset variable references fatal, pipefail makes pipelines return the "worst" exit status in the pipeline; no pipes here so not relevant).  I don't actually know what "a" does (I guess it makes "source" create readable variables even though the text of the file is FOO=bar and not export FOO=bar?)

src/server/pps/server/api_server.go
```diff
@@ -173,6 +179,19 @@ func validateNames(names map[string]bool, input *pps.Input) error {
 	return nil
 }
 
+func validateName(name string) error {
```
jrockway, line 182: We should probably have a unit test for the calling code; I looked for one but didn't find it.  (validateNames calls validateName, validateInput calls validateNames, validatePipeline calls validateInput; we should at least test validatePipeline or perhaps validateInput).

jrockway (approved):
This is a great feature.  Code looks good to me; I have a couple minor comments that shouldn't block the release.

src/server/worker/datum/datum.go
brycemcanally: I'm going to leave it as `path.Join` for now to be consistent with the rest of the code, but I added a TODO for this.

src/server/pachyderm_test.go
brycemcanally: yeah, it exports created variables and functions.

src/server/pps/server/api_server.go
brycemcanally: Yeah, I think a test that at least ensures the validation is run probably makes sense. I'll add a TODO. I don't know if it really makes sense to verify all of the simpler validations though, like the names and whatnot.

jrockway: It's a shame that we wrote an integration test instead of a unit test, but no need to fix in this PR.

PR #8790, opened 2023-04-26
Add proxy support + support for zero config deploys on localhost

chainlink:


jrockway (approved):
I tested this with a non-localhost setting for proxy.host, and all is well.  I tested this with no proxy.host, and that also works now!  Very nice.

At some point, we need to make the explicit configuration "localhost" work.  I will follow up with @zmajeed on that.

PR #8792, opened 2023-04-26
Djanicek/core 1550/integration test cover direct

djanicekpach:
This change uploads code coverage from minikube during integration tests to codecov so we can get more complete coverage numbers.  
The steps to achieve this are:

1. Build new docker images with code coverage binaries instead of the normal binaries.
2. Use those images in the CI when running tests
3. For `Cleanup` on minikube tests, we use pachctl debug to download the coverage for all of the containers.
4. In the pipeline un-tar the coverage files, then merge/format them with `go tool covdata textfmt`
5. upload each `.txt` file to codecov [which merges the files.](https://docs.codecov.com/docs/merging-reports)

The rootless tests still use the non-coverage image to serve as a smoke test for the normal build. I don't think we miss much coverage from that.(_maybe_ like .05% in my testing)

[Here's the branch coverage on codecov](https://app.codecov.io/github/pachyderm/pachyderm/tree/djanicek%2Fcore-1550%2Fintegration-test-cover-direct?displayType=list)

This got us about 9-10% increase in coverage, which is less than I expected but I've gone through a bunch of files spot checking and comparing to my local coverage results and I haven't found any inconsistencies. 

I think a large part of the low number is that out `.pb` files are counted and they rarely get above 50% coverage. Since protobufs are often huge files, it seems like our code coverage tracks pretty closely with our protobuf coverage. I'm not sure if that's good or not. We could ignore generate code in codecov if we wanted to I think, but for now I think I'm ok with leaving it.


.circleci/main.yml
```diff
@@ -592,6 +637,41 @@ jobs:
       - run:
           name: Push docker
           command: make docker-push VERSION=${CIRCLE_SHA1}
+  build-docker-images-with-coverage:
```
molinamelendezj, line 640: any way we could simplify this instead of using a near duplicate of same job. maybe a conditional parameter. you also might want to rebase the branch. i updated goreleaser, and it not 1.10.3 anymore.

etc/testing/circle/wait_for_docker_images.sh
molinamelendezj, line 1: can you make sure this shell is not being called from else where, TEST_IMAGE_SHA will have a cover- prefix on the tag. this will break the release. because its used there to. 

the test that run on release will fail because they will be timeout waiting for that image that was never built in its workflow.

molinamelendezj (changes requested):

goreleaser/docker-cover.yml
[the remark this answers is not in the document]
djanicekpach: I couldn't find a way. I looked at this before and there are ways to [template things](https://goreleaser.com/customization/templates/), but not in the fields I need. I'm also concerned about `image_templates` field with normal tags like just `pachyderm:pachd` getting released with the coverage binary.

etc/testing/circle/wait_for_docker_images.sh
djanicekpach: Ok, I think I have a solution, let me know what you think.

.circleci/main.yml
djanicekpach: Done. I haven't rebased yet, but I thought I did pretty recently, I'll double check before merge.

molinamelendezj (approved):
check that regex one last time, think you need a prefix v for tags

goreleaser/docker-cover.yml
jrockway, line 1: At this point we really need to consider generating these files, kind of how etc/generate-envoy-config works.

jrockway (approved):
This is very cool.

goreleaser/docker-cover.yml
djanicekpach: I'll ticket this. IDK which epic, but I like the idea a lot.

PR #8794, opened 2023-04-26
dex: upgrade to v2.36.0

jrockway:
The version of dex in pachyderm/dex is v2.36.0 with only one of our changes; the localhost issuer -supporting stuff.  The rest of the stuff in there was only for hub and was becoming a bit of a burden, so it's removed.  For that reason, I think this will be a 2.7-targeting change, so there is plenty of burn time.

We also had to upgrade a lot of modules; we will see what that breaks.


acohen4 (approved):
LGTM! Thanks

chainlink:
Nice upgrade! Does this live on a branch in pachyderm/dex?

chainlink:
Also, I haven't been able to totally confirm this, but I'm pretty sure we can eliminate `insecureSkipIssuerCallbackDomainCheck` with the move to proxy everywhere. 

Minor nit: We should eliminate references to `login_hint` in the helm chart example configs if that's gone.

jrockway:
@chainlink I think we have to keep the insecureIssuerCallbackDomainCheck stuff... we pretty much always set the Issuer to `http://pachd:30658/` because console reads that and uses it to make requests to pachd; if it's `proxy.host` then that depends on a networking setup where the domain is accessible from inside the cluster, which is not always guaranteed (for example, if the proxy is behind a LoadBalancer that uses the PROXY protocol; and also when the issuer is "localhost"; console's view of "localhost" is different than the user's workstation's view of "localhost", so it won't be able to talk to the identity server in that case).  For that reason, I think we have to keep it; though we can get rid of it if we make console manually configure OIDC instead of having it auto-configure itself.  I sent them a change to do that a while back, but they and customer-eng preferred the "localhostIssuer" hacks.

PR #8810, opened 2023-05-02
[PFS-85] Disable inSidecars by Default

FahadBSyed:
The goal of the PFS team is to disable `inSidecars` for the next pachyderm minor release. This change should be made in coordination with the customer team so that customers are aware of the change in behavior and can adjust their helm configuration accordingly.


jrockway (approved):
This looks fine but you're committing to the 2.7 branch.  I assume that's what you meant by 2.8 :D

FahadBSyed:
I have to actually double check with Brendon, but I think we were intending 2.8 not 2.7 which would be the next release after this.

PR #8813, opened 2023-05-02
Fix erroneous semi-colon in auth proto

loganripplinger:
The auth proto had a double colon. This causes an error when importing the protos into Postman.


jrockway (approved):

albscui:
Thanks for the cleanup! On this note, I think we should look into proto formatters like `buf format` https://buf.build/docs/format/style/

PR #8815, opened 2023-05-02
[PFS-86] Fix get file url output paths

brycemcanally:
This PR makes a few changes to the get file url functionality:
- Removes the leading slash from the output paths. Directories as a concept don't exist in object storage and are typically simulated by a non-empty path with a trailing slash, so a root directory and paths with leading slashes aren't expected. This expectation seems to be baked into certain object storage providers (i.e. gcs), which is resulting with paths with double slashes due to bucket + object path concatenations. These seem to be removed automatically by the gcs client that was used prior to our migration to go cdk, so we now just remove the leading slash in our code.
- Fixes support for output path prefixes. Prior to the 2.5.0 get file url changes, an object path in the url would be used as a prefix for the exported files. This was unintentionally removed in the 2.5.0 changes, so it was re-implemented in terms of the new code and a test was added.
- The put / get file url tests were updated to use go cdk rather than the old object storage clients.

### DIFF
--- a/src/internal/dockertestenv/minio.go
+++ b/src/internal/dockertestenv/minio.go
@@ -9,12 +9,17 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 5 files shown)


FahadBSyed (approved):
Looks good to me. Thanks for updating the tests.

PR #8816, opened 2023-05-02
Djanicek/core 1553/collect test results

djanicekpach:
This adds a script to our test pipelines that pushes data from each test out to pachyderm.pachops. We are not really processing this data in the PR. The [next ticket ](https://pachyderm.atlassian.net/browse/CORE-1554) will be processing the data in an egress pipeline so we can visualize results in Grafana.

This ticket roughly has 2 parts:

1. Output test data to a file if we aren't already. We have slightly different was of doing this depending on if we are using gotestsum or go test, but the end result should be a jsonl file in the /tmp/test-results folder for each test job. (I'd like to reign some of these different test commands in when [I look at splitting the tests more efficiently](https://pachyderm.atlassian.net/browse/CORE-861).)
2. Run the go collector script at `etc/testing/circle/workloads/go-test-results/collector/main.go` after the tests finish in a separate job. 

Example data from this branch is here: https://pachyderm.pachops.com/lineage/ci-metrics/repos/go-test-results-raw/branch/default

Interestingly, test-go verbose output is a little over 1gb of data. I think this is OK short term, but long term I think we probably don't want to store 1GB per branch indefinitely. We could just clean the raw data out of pachyderm when the branch is no longer in use. Since the next step will be putting the important data into the DB for visualization, we shouldn't really need the raw data for an extended period of time. Follow up ticket for that if/when we think it is necessary: https://pachyderm.atlassian.net/browse/CORE-1645


etc/testing/circle/workloads/go-test-results/collector/main.go
```diff
@@ -0,0 +1,101 @@
...
+	logger.Println("Beginning results upload.")
+	robotToken := os.Getenv("PACHOPS_PACHYDERM_ROBOT_TOKEN")
+	if len(robotToken) == 0 {
+		logger.Println("No pachyderm robot token found. Continuing with unauthenticated pach client.")
+	}
+	resultsFolder := os.Getenv("TEST_RESULTS")
+	if len(resultsFolder) == 0 {
+		logger.Fatal("TEST_RESULTS needs to be populated to find the test results folder.")
+	}
+	branch := sanitizeName(os.Getenv("CIRCLE_BRANCH"))
```
molinamelendezj, line 57: @djanicekpach INT i believe also run on releases. in that case there will be no CIRCLE_BRANCH var instead you will see a CIRCLE_TAG. you should not fail when CIRCLE_BRANCH is null. Will be also nice to query later across tag of the release.

molinamelendezj (approved):
few comments.

 but i think we should consider maybe not failing the jobs if it cant upload. how reliable we think the pachops cluster will be. would hate to see alot of CI failures/block in case pachop cluster is not taking in the data. but if its reliable enough then its okay with me

djanicekpach:
> few comments.
> 
> but i think we should consider maybe not failing the jobs if it cant upload. how reliable we think the pachops cluster will be. would hate to see alot of CI failures/block in case pachop cluster is not taking in the data. but if its reliable enough then its okay with me

I was already considering this, but never quite made up my mind. Since it came up, I'll do it before merge.

etc/testing/circle/workloads/go-test-results/collector/main.go
djanicekpach: Added handling for this

PR #8817, opened 2023-05-02
postgres: add the option to enable pgx query logs

jrockway:
This adds the option to turn on SQL query logging.  Set `pachd.sqlQueryLogs=true` when upgrading and they're turned on (note that logLevel also needs to be debug).

The query logging machinery makes an effort to handle nil contexts or contexts without a logger  silently, because that's the most common case.  We have a loggable context in a lot of queries, but not everywhere, and it's not worth the warnings in the log.

I also tweaked the gc.RunOnce span logging a little bit, so the RunOnce span propagates into future operations, associating GC-related queries with this span.  Makes filtering the logs a little easier when debugging GC.

This is part of an investigation that's part of CORE-1644.


src/internal/serviceenv/service_env.go
```diff
@@ -408,6 +410,8 @@ func (env *NonblockingServiceEnv) newDirectListener() col.PostgresListener {
 		dbutil.WithDBName(env.config.PostgresDBName),
 		dbutil.WithUserPassword(env.config.PostgresUser, env.config.PostgresPassword),
 		dbutil.WithSSLMode(env.config.PostgresSSL),
+		// Note, enabling query logs with WithQueryLog on the direct listener seems to break
```
acohen4, line 413: yikes

jrockway: I think this is because of how pgx generates a "fake" DSN to connect to, so we end up connecting to the wrong place when the DSN is shared.  I didn't have time to look into it, though, but the symptom is that isAuthEnabled returns the wrong answer, so I know it breaks listeners.

acohen4: should we have an action item to understand this behavior?

acohen4 (approved):
LGTM. left one note of concern

src/internal/serviceenv/service_env.go
jrockway: I adjusted the note here.  It's understood, we use two different postgres libraries 😂 and only pgx can do logging.

PR #8822, opened 2023-05-03
Rename /download route to /archive

jrockway:
We uh forgot that /download is a console-internal route for single-file downloads.  This renames the archiveserver route to /archive, which makes more sense anyway.

Some may argue that the `pachctl misc` commands should also be renamed, but whatever, nobody is going to run them except us.


pleeko:
Out of curiosity, if console updated its `/download` route to something different would you have to make a change to accept that as well?

armaanv (approved):

PR #8823, opened 2023-05-03
[CORE-1483] Switch to internal use of internal client

robert-uhl:
This PR creates an internal Go client for use by the server and other internal code. Using the old external client as a base, it deletes deprecated non-project-aware functions and methods  and renames `FooProjectBar` functions to `FooBar`. This PR does not delete the external Go client; that remains as a follow-on step.

### DIFF
--- a/src/internal/clusterstate/v2.6.0/pfsdb.go
+++ b/src/internal/clusterstate/v2.6.0/pfsdb.go
@@ -3,6 +3,7 @@ package v2_6_0
 import (
 	"bytes"
 	"context"
+	"database/sql"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


chainlink:
Python tests passing here: https://github.com/pachyderm/pachyderm/compare/rau/core-1483-remove-server-dependency-on-client-simple...k-test-1483?expand=1

chainlink (approved):
Looks like our test suite passes here

src/internal/client/limit/limit.go
```diff
@@ -0,0 +1,65 @@
+// Package limit provides primitives to limit concurrency.
```
brycemcanally, line 1: I don't think this needs to be copied over. This package should probably be separate from the client package anyways.

robert-uhl: I’ll lift it up out of the client package, since it's not used in there. I do think that it should still be internal, so that nothing under the server package refers to it.

acohen4 (approved):
LGTM too. The doc you shared about this helped me too :)

PR #8825, opened 2023-05-04
pipeline extension - move project into its own field

kevyang:
Previously, we specified projects for pipeline creation by denoting it with a forward slash in the pipeline name. This is not very intuitive and we don't really explain this to the user anywhere. This change updates the UI so that there is a separate field for the project name, which defaults to 'default'.

Jira: [INT-946]

[INT-946]: https://pachyderm.atlassian.net/browse/INT-946


kevyang:
Screen Shot 2023-05-04 at 11 04 59 AM
Screen Shot 2023-05-04 at 11 09 17 AM

jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
```diff
@@ -169,8 +187,8 @@ const Pipeline: React.FC<PipelineProps> = ({
           data-testid="Pipeline__specPreview"
           name="specPreview"
           value={`pipeline:
-  name: ${pipeline.name}
-  project: ${pipeline.project?.name ?? 'default'}
+  name: ${pipelineName}
+  project: ${pipelineProject ?? 'default'}
```
bbonenfant, line 191: I don't think this is working. When I have no project specified, this field is empty in the pipeline spec preview

bbonenfant (approved):
Aside from the one minor bug I mentioned, this looks good to me. I want to confirm with @chainlink though that this is a post GA feature, unless we feel strongly that we should try to get it included before the release.

jupyter-extension/src/plugins/mount/components/Pipeline/Pipeline.tsx
kevyang: good catch, patched this just now

chainlink:
Couple of design thoughts:
- I think project name should go before pipeline name 
- it's a little confusing to have an empty text box, but then have `default` set in the pipeline. I think `default` should maybe be actual text instead of a hint in the textfield

chainlink:
And yeah, to Ben's comment, I think this should probably ship in 2.6.1 since it's an enhancement vs a bug fix. Happy to discuss though

PR #8831, opened 2023-05-05
Catch zero since values

robert-uhl:


jrockway (approved):
Interesting edge case.  What happens if you use --follow and want only logs that happen after the command starts running?

robert-uhl:
I did test that `--follow` does indeed follow the logs as they are created, and `--follow --since 1s` leaves off the earlier logs, but `--follow --since 0s` returns everything because the zero since value is ignored. Do you think that the special handling should only apply to Kubernetes logs? K8s log retrieval fails with a zero `SinceSeconds`.

PR #8841, opened 2023-05-10
[CORE-1650] Exclude pagination marker from ListFIle and WalkFile results

armaanv:
This makes these APIs consistent with the other paginated APIs which exclude the marker.

### DIFF
--- a/src/server/pfs/server/driver_file.go
+++ b/src/server/pfs/server/driver_file.go
@@ -293,7 +293,7 @@ func (d *driver) listFile(ctx context.Context, file *pfs.File, paginationMarker 
 	if reverse {
 		fis := newCircularList(number)
 		if err := s.Iterate(ctx, func(fi *pfs.FileInfo, _ fileset.File) error {
-			if paginationMarker != nil && strings.Compare(pfsfile.CleanPath(fi.File.Path), pfsfile.CleanPath(paginationMarker.Path)) == 0 {
+			if isPaginationMarker(paginationMarker, fi) {
 				return nil
 			}
 			if pathIsChild(name, pfsfile.CleanPath(fi.File.Path)) {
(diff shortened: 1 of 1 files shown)


brendoncarroll:
Can we create a function for this comparison (which appears 4 times). `paginationMarker != nil && strings.Compare(pfsfile.CleanPath(fi.File.Path), pfsfile.CleanPath(paginationMarker.Path)) == 0 `.

Maybe something like `pageContains(paginationMarker, path string) bool` What do you think?

armaanv:
> Can we create a function for this comparison (which appears 4 times). `paginationMarker != nil && strings.Compare(pfsfile.CleanPath(fi.File.Path), pfsfile.CleanPath(paginationMarker.Path)) == 0 `.
> 
> Maybe something like `pageContains(paginationMarker, path string) bool` What do you think?

I'm thinking I can add a path comparison function in `paths.go` which will do `strings.Compare(pfsfile.CleanPath(fi.File.Path), pfsfile.CleanPath(paginationMarker.Path)) == 0 ` and output a bool. So the checks in this file will become `paginationMarker != nil && pathIsSame(fi.File.Path, paginationMarker.Path)`. I think this makes the comparison a bit shorter and keeps it readable. 

I'm happy to defer to your suggestion if you feel that's more readable though

src/server/pfs/server/driver_file.go
```diff
@@ -293,6 +293,9 @@ func (d *driver) listFile(ctx context.Context, file *pfs.File, paginationMarker
 	if reverse {
 		fis := newCircularList(number)
 		if err := s.Iterate(ctx, func(fi *pfs.FileInfo, _ fileset.File) error {
+			if paginationMarker != nil && strings.Compare(pfsfile.CleanPath(fi.File.Path), pfsfile.CleanPath(paginationMarker.Path)) == 0 {
```
albscui, line 296: Hmm why use `strings.Compare`, when Go syntax natively supports string comparison? i.e. just use `==` directly?

albscui (commented):
What's the reason skipping the pagination marker? Is it so that we don't show duplicates across successive pages? Sorry, I'm not too familiar with how our pagination works.

brendoncarroll:
I would like one place that handles the pagination logic.  So we can blindly pass in the pagination marker and paths and know whether or not we should emit a given entry.

Also strings can be compared using `==` directly.  No need for `strings.Compare` or even `strings.Equal`.  Cleaning the paths seems like a good idea though.

brendoncarroll (approved):
LGTM

PR #8849, opened 2023-05-12
[CORE-1658] Implement credentials.PerRPCCredentials

robert-uhl:


src/internal/client/client.go
```diff
@@ -829,6 +829,25 @@ func (c *APIClient) connect(rctx context.Context, timeout time.Duration, unaryIn
...
+		clientData[auth.ContextTokenKey] = c.authenticationToken
+	}
+	// metadata API downcases all the key names
+	if c.metricsUserID != "" {
+		clientData["userid"] = c.metricsUserID
+		clientData["prefix"] = c.metricsPrefix
+	}
+	if len(os.Args) > 1 && os.Args[0] != "/pachd" {
+		clientData["command"] = strings.Join(os.Args, " ")
+	}
```
albscui, line 844: nit: this logic seems to be duplicated in `APIClient.AddMetadata`, which adds the metadata to `ctx`.

albscui (approved):
LGTM

src/internal/client/client.go
robert-uhl: Yeah, and a future PR will remove `APIClient.AddMetadata`. That requires a bigger refactor first, though.

PR #8851, opened 2023-05-12
Populate empty projects in project selection within Jupyter extension

kevyang:
Currently, the way we obtain the list of projects is by taking all of the repos and getting the project info from within them. With this logic, however, empty projects will not appear in the drop-down list because there are no repositories in them. This change adds a /projects endpoint to the mount-server and uses that endpoint to obtain a true list of Pachyderm projects.

Jira: [INT-945]

[INT-945]: https://pachyderm.atlassian.net/browse/INT-945


kevyang:
didn't realize the tests were broken, going to figure it out first before putting this up for review

PR #8852, opened 2023-05-12
Refactor multierrors to use the standard library's errors.Join

jrockway:
I moved over the multierror functionality from hashicorp and uber's multierr library, and refactored our code to use them.  Instead of picking one winner from the two, we are now just using Go's built-in multierrors.  I also added a lint rule to ban future uses of those two third-party libraries.

The most notable impact is that the two OSS libraries call the operation "Append", but Go calls it "Join", so our wrappers are also called Join.

Close/Invoke/InvokeN are based on go.uber.org/multierr, but without the indirection through an interface.  Closing things is 90% of our use case, so I made that nicer and didn't bring over the Invoker/Closer interfaces.  Invoke is rarely needed but logically makes sense to have.  InvokeN is useful for locks.  Any more utilities than that didn't seem necessary during the refactor, but if people have favorite wrappers we can add them here.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,75 +1,82 @@
 run:
-  timeout: 5m
-  skip-dirs:
-    - cached-deps
-  build-tags:
-    - k8s
-    - unit_test
+    timeout: 5m
+    skip-dirs:
+        - cached-deps
+    build-tags:
+        - k8s
+        - unit_test
 linters:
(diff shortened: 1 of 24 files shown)


brendoncarroll (approved):
LGTM.

I'm a big fan of errors.Close, and the `&retErr` pattern in general.

PR #8853, opened 2023-05-15
Add a "tail loki" command

jrockway:
I often want to watch all the logs from the various Pachyderm components, even as I restart pachd, so I added an RPC endpoint that does this.  You can start this up once and it will keep reconnecting and streaming logs forever, this means you don't have to compose a new "kubectl logs" every time you upgrade pachd ;)

The server side of this uses the Loki websocket API to do the tailing, which is presumably more efficient than a bunch of range requests that are polled.  If Loki closes the websocket, we sleep and try again without breaking the RPC; this lets you survive Loki restarts.   The RPC has a "warnings" channel that will keep the client informed of reconnect attempts.

On the client side, we also aggressively restart after EOFs/network errors, so that if you restart pachd you should see all the logs from the new server starting up, as well as the old server dying; as soon as the old server dies, we reconnect to the new one, so there isn't much of a gap in the logs.  (The client should probably do the same thing the server does; keep track of the Time of the latest message, and ask for that much "Since" when reconnecting.  But I didn't bother.)

Finally, all the various log formats for our components are canonicalized into the same format (JSON schema) as pachd itself, on the client side.  This means you can pipe `pachctl tail-loki` into `jlog` and not get any parse errors.  There is a `--raw` option if you don't want this.

You can do any loki search, but we default to `{suite="pachyderm"}` to see everything.  Another useful one is to set x-request-id in your browser using a header-setting extension, and then filter on that request-id: `{suite="pachyderm"} |= "ce2edb68-fe22-9999-9999-1c4509502697"`.


src/server/pps/cmds/cmds_test.go
```diff
@@ -71,7 +71,7 @@ const (
 	secret   = "secret"
 
 	create  = "create"
-	delete  = "delete"
+	del     = "delete"
```
jrockway, line 74: This conflicted with my desire to use `builtin.delete` to delete something from a map ;)

src/server/pps/cmds/cmds.go
```diff
@@ -783,7 +935,7 @@ each datum.`,
 	getLogs.Flags().BoolVar(&raw, "raw", false, "Return log messages verbatim from server.")
 	getLogs.Flags().BoolVarP(&follow, "follow", "f", false, "Follow logs as more are created.")
 	getLogs.Flags().Int64VarP(&tail, "tail", "t", 0, "Lines of recent logs to display.")
-	getLogs.Flags().StringVar(&since, "since", "24h", "Return log messages more recent than \"since\".")
+	getLogs.Flags().DurationVar(&since, "since", 24*time.Hour, "Return log messages more recent than \"since\".")
```
acohen4, line 938: 👍

src/internal/pctx/context.go
```diff
@@ -14,6 +17,11 @@ func TODO() context.Context {
 	return log.AddLogger(context.TODO())
 }
 
+// Interactive returns a context for use in interactive processes.
+func Interactive() (context.Context, context.CancelFunc) {
```
acohen4, line 21: do we want to pass a name arg here?

src/server/pps/cmds/cmds.go
```diff
@@ -619,9 +615,169 @@ each datum.`,
...
+					}
+					log.Error(ctx, "attempt to start tail stream failed; retrying", zap.Error(err))
+					sleep()
+					continue
+				}
+				for {
+					msg, err := c.Recv()
+					if err != nil {
+						if errors.Is(err, io.EOF) {
+							time.Sleep(time.Second)
```
acohen4, line 674: super minor, but why not use your defined sleep() here?

src/server/pps/cmds/cmds.go
```diff
@@ -619,9 +615,169 @@ each datum.`,
...
+							return errors.Wrap(err, "recv")
+						}
+						log.Error(ctx, "failed to recv message; retrying", zap.Error(err))
+						sleep()
+						continue main
+					}
+					connected = true
+					switch x := msg.GetResponse().(type) {
+					case *pps.TailLokiResponse_Warning:
+						log.Info(ctx, "server warning: "+x.Warning)
```
acohen4, line 687: How useful is it for the user to be alerted here? Is it worth making this log.Error to be more visible to the user?

src/server/pps/cmds/cmds.go
```diff
@@ -619,9 +615,169 @@ each datum.`,
...
+								switch x.Stream.Metadata["app"] {
+								case "pachd": // We adapt other logs to pachd's format.
+								case "pachyderm-kube-event-tail", "etcd":
+									obj["time"] = obj["ts"]
+									obj["message"] = obj["msg"]
+									obj["severity"] = obj["level"]
+									delete(obj, "ts")
+									delete(obj, "msg")
+									delete(obj, "level")
+								case "pachyderm-proxy":
```
acohen4, line 720: why don't we include a "message" field for pachyderm-proxy?

src/server/pps/server/api_server.go
```diff
@@ -1367,6 +1373,155 @@ func (a *apiServer) QueryLoki(request *pps.LokiRequest, apiQueryLokiServer pps.A
...
+				}
+				time.Sleep(time.Second)
+				continue main
+			}
+			for _, s := range value.Streams {
+				lvs := make([]*pps.LokiValue, len(s.Values))
+				for i, v := range s.Values {
+					t, err := types.TimestampProto(v.Time)
+					if err != nil {
+						t = nil
```
acohen4, line 1498: worth logging here?

acohen4 (approved):
LGTM. Added some questions and comments

src/server/pps/cmds/cmds.go
jrockway: No message key in its logs.

jrockway:
@acohen4 I'm planning on revisiting this after GetLogsV2 works... we can do 99% of this without "admin privileges" which ordinary users will appreciate, I think.

PR #8855, opened 2023-05-15
Populate empty projects in project selection within Jupyter extension

kevyang:
Currently, the way that we populate the projects list in the Jupyter extension is by taking the repos list and getting all projects from each of the repos. The issue with this is that projects with no repos can exist, and it can cause confusion if a user creates a new project without repos since it won't show up in the extension. This change adds a /projects endpoint to the mount server and makes the extension query it to retrieve all projects, including those without any repos.

### DIFF
--- a/jupyter-extension/src/plugins/mount/mount.tsx
+++ b/jupyter-extension/src/plugins/mount/mount.tsx
@@ -175,6 +175,7 @@ export class MountPlugin implements IMountPlugin {
               updateData={this._poller.updateData}
               mountedItems={[]}
               type={'mounted'}
+              projects={[]}
             />
           </div>
         )}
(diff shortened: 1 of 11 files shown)


jupyter-extension/src/plugins/mount/pollMounts.ts
```diff
@@ -76,6 +85,18 @@ export class PollMounts {
...
+  get projects(): ProjectInfo[] {
+    return this._projects;
+  }
+
+  set projects(data: ProjectInfo[]) {
+    if (data === this._projects) {
+      return;
+    }
+    this._projects = data;
+    this._projectSignal.emit(data);
```
bbonenfant, line 97: Is anything listening to this signal?

bbonenfant (approved):
This looks good to me. Just had that one question

jupyter-extension/src/plugins/mount/mount.tsx
```diff
@@ -185,18 +186,23 @@ export class MountPlugin implements IMountPlugin {
...
-            </div>
-            <SortableList
-              open={this.open}
-              items={unmounted ? unmounted : this._poller.unmounted}
-              updateData={this._poller.updateData}
-              mountedItems={this._poller.mounted}
-              type={'unmounted'}
-            />
-          </div>
+          <UseSignal signal={this._poller.projectSignal}>
```
kevyang, line 189: here

jupyter-extension/src/plugins/mount/pollMounts.ts
kevyang: yes, check my other comment

chainlink (approved):

PR #8856, opened 2023-05-15
gogo removal part 1: Remove use of goproto_stringer

jrockway:
I'm incrementally moving towards getting rid of gogoproto.  Step 1 is to be able to write certain methods on our own; this is a feature that gogoproto has, but not one that the official compiler has.  So I've hacked in `gopatch` to edit the generated files and remove conflicting methods.

More steps to follow.


brycemcanally (approved):
lgtm

PR #8857, opened 2023-05-16
Show the creator and caller of cancelled contexts

jrockway:
This adds pctx.WithCancel, a drop-in replacement for context.WithCancel, which includes the context creator and context canceller's file:line information in the Cause.  This also refactors most uses of ctx.Err() with context.Cause(ctx), so that most functions will report our improved error instead of "context cancelled".

I used the following gopatch to perform the refactoring:
```patch
 @@
@@
- import "context"
- context.WithCancel
+ import "github.com/pachyderm/pachyderm/v2/src/internal/pctx"
+ pctx.WithCancel
@@
@@
- ctx.Err()
+ import "context"
+ context.Cause(ctx)
@@
@@
- workerCtx.Err()
+ import "context"
+ context.Cause(workerCtx)
@@
@@
- cancelCtx.Err()
+ import "context"
+ context.Cause(cancelCtx)
@@
var x expression
var ctx identifier
@@
- import "context"
- context.Cause(ctx) == x
+ ctx.Err() == x
```

### DIFF
--- a/etc/benchmark-uploads/main.go
+++ b/etc/benchmark-uploads/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/minio/minio-go/v7"
 	"github.com/minio/minio-go/v7/pkg/credentials"
 	"github.com/pachyderm/pachyderm/v2/src/client"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pctx"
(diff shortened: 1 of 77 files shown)


brendoncarroll (approved):
LGTM

PR #8861, opened 2023-05-17
Re-gen protos off latest

chainlink:


python-sdk/pachyderm_sdk/api/__init__.py
```diff
@@ -1 +0,0 @@
-from .pfs import _additions as __pfs_additions
```
bbonenfant, line 1: Could you move this to the top most `__init__.py`? I mistakenly put it here and now it just gets deleted whenever the protos get generated

bbonenfant (approved):

python-sdk/pachyderm_sdk/api/__init__.py
chainlink: Looks like it already is? https://github.com/pachyderm/pachyderm/blob/update-python-sdk-protos/python-sdk/pachyderm_sdk/__init__.py#L33

bbonenfant: Oh nice!

PR #8862, opened 2023-05-17
Pass information about the external address to pachd and console

jrockway:
The puts proxy.host and proxy.tls.enabled into environment variables for pachd and console.  InspectCluster will also return these now.

This is so that we can generate http/https links to the archive download endpoint in pachd and local console.  Real console doesn't need this, it will always use window.host which is fine.


acohen4 (approved):
LGTM

PR #8864, opened 2023-05-18
[CORE-1590] jq filter for ListJobSet

armaanv:
I added the proto field and some logic for applying the filter. 

Validation and unit testing not done yet


robert-uhl (changes requested):
There is a risk with `jq`-style filters: the language is Turing-complete, so it can consume unbounded memory and run in unbounded time. The latter issue is avoidable by using a context, but I believe that the former issue is currently unavoidable given the knobs exposed by Go and the gojq package.

Here is a filter which will consume all memory: `{ source: ., count: 128, output: "a"} | until(.count <= 0; .output = .output + .output | .count -= 1) | .output`.

Here is a filter which will run forever: `{ source: ., output: "" } | until(.source == "quux"; {"foo": "bar"}) | .output`.

There should at least be a test for the latter case. I don’t know if there is anything useful to be done for the former case. In retrospect, we should probably have used a more restrictive language for filtering, but changing that is a different project.

Note also that the language accepted by `gojq` is slightly different from that accepted by `jq`: `{ source: ., count: 128, output: "a"} | until(.count <= 0; .output *= 2 | .count -= 1) | .output` is a valid `jq` filter but not a valid `gojq` filter.

src/server/pps/server/api_server.go
```diff
@@ -651,6 +651,45 @@ func (a *apiServer) ListJobSet(request *pps.ListJobSetRequest, serv pps.API_List
 
 	// Track the jobsets we've already processed
 	seen := map[string]struct{}{}
+	// Prepare filter functions.
+	projectsFilter := make(map[string]bool, len(request.GetProjects()))
+	for _, project := range request.GetProjects() {
+		projectsFilter[project.GetName()] = true
+	}
+	jqFilter, err := jqFilterFunc(serv.Context(), request.GetJqFilter())
```
robert-uhl, line 659: The new variable `jqFilter` here isn’t really a _jq_ filter; it’s just a message filter, specifically a job info filter. Maybe call it `jobInfoFilter` or just `filter`?

src/server/pps/server/api_server.go
```diff
@@ -651,6 +651,45 @@ func (a *apiServer) ListJobSet(request *pps.ListJobSetRequest, serv pps.API_List
...
+	if err != nil {
+		return errors.Wrap(err, "error creating jq filter function")
+	}
+	getJobSet := func(jobInfo *pps.JobInfo) (*pps.JobSetInfo, error) {
```
robert-uhl, line 663: Why is this declared here, only to be used once, [60 lines later](https://github.com/pachyderm/pachyderm/pull/8864/files#diff-e400d3c661e4039b9f23301b99e8a22707cf3fdfa18865f569b499d3d708840dR726)? It would be more readable in context. I don’t believe that you’re saving any computation by declaring this outside of the closure passed at [line 710](https://github.com/pachyderm/pachyderm/pull/8864/files#diff-e400d3c661e4039b9f23301b99e8a22707cf3fdfa18865f569b499d3d708840dR710).

src/server/pps/server/api_server.go
```diff
@@ -690,33 +723,16 @@ func (a *apiServer) ListJobSet(request *pps.ListJobSetRequest, serv pps.API_List
 				return nil
 			}
 		}
-		jobInfos, err := pachClient.InspectJobSet(id, request.Details)
+		jobset, err := getJobSet(jobInfo)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "error filtering jobset")
```
robert-uhl, line 728: Is the error in _filtering_ or in _getting_ the job set?

As noted above, if `getJobSet` were expanded in place then each error return could give some more context.

src/server/pps/server/api_server.go
```diff
@@ -828,18 +844,30 @@ func (a *apiServer) getJobDetails(ctx context.Context, jobInfo *pps.JobInfo) err
...
+	// Prepare filter functions.
+	projectFilter := make(map[string]bool, len(request.GetProjects()))
 	for _, project := range request.GetProjects() {
-		filter[project.GetName()] = true
+		projectFilter[project.GetName()] = true
 	}
-	keep := func(j *pps.JobInfo) error {
-		if len(filter) == 0 || filter[j.Job.Pipeline.Project.GetName()] {
-			return errors.Wrap(resp.Send(j), "could not send filtered job")
+	ctx := resp.Context()
```
robert-uhl, line 852: Why declare a variable here but only use it once, on the following line?

src/server/pps/server/api_server.go
```diff
@@ -828,18 +844,30 @@ func (a *apiServer) getJobDetails(ctx context.Context, jobInfo *pps.JobInfo) err
...
+	jqFilter, err := jqFilterFunc(ctx, request.GetJqFilter())
```
robert-uhl, line 853: Once it’s built, this isn’t really a _jq_ filter; it’s a _job_ filter.

src/server/pps/server/api_server.go
```diff
@@ -828,18 +844,30 @@ func (a *apiServer) getJobDetails(ctx context.Context, jobInfo *pps.JobInfo) err
...
+	if err != nil {
+		return errors.Wrap(err, "error creating jq filter function")
+	}
+	keep := func(j *pps.JobInfo) (bool, error) {
```
robert-uhl, line 857: I think that this too should be expanded at the call site. It would make the logic clearer too, since both a false or an error lead to immediate returns.

src/server/pps/server/api_server.go
```diff
@@ -2749,20 +2743,9 @@ func (a *apiServer) getLatestJobState(ctx context.Context, info *pps.PipelineInf
...
-		if err != nil {
-			return errors.EnsureStack(err)
-		}
-		jqCode, err = gojq.Compile(jqQuery)
-		if err != nil {
-			return errors.EnsureStack(err)
-		}
-		// ensure field names and enum values match with --raw output
-		enc = serde.NewJSONEncoder(&jsonBuffer, serde.WithOrigName(true))
+	jqFilter, err := jqFilterFunc(ctx, request.GetJqFilter())
```
robert-uhl, line 2746: This is a project filter.

src/server/pps/server/api_server.go
```diff
@@ -3605,3 +3567,49 @@ func ensurePipelineProject(p *pps.Pipeline) {
 		p.Project = &pfs.Project{Name: pfs.DefaultProjectName}
 	}
 }
+
+func jqFilterFunc(ctx context.Context, filter string) (func(proto.Message) (bool, error), error) {
```
robert-uhl, line 3571: Closing over a context is a lot like stashing it in a struct — something to be avoided unless using an intermediate context-ignorant library. It is better to pass the context through the call chain, so that each function can attenuate the context properly.

Since `jqFilterFunc` doesn’t really do anything requiring a context and is unexported, it doesn’t need to be passed in as an argument; it may always be added later. It’s not _wrong_ to include it, just not necessary.

Concretely, this should be `func jqFilterFunc(string) (func(context.Context, proto.Message) (bool, error), error)` and the call sites should look like `jqFilter(ctx, msg)`.

I believe that the name here is not so great either: this isn’t really a “jq filter func”; it _compiles_ a jq  filter to a filter func. I would probably call this `newFilter` or `newFilterFromJQ`. Since this is all unexported, `newFilter` or `newMessageFilter` is probably fine for now, but it’s up to you.

It’s probably not necessary to declare a type for the return value, although it too wouldn’t be wrong.

robert-uhl (changes requested):

src/server/pps/server/api_server.go
albscui: This `ctx` variable is used several times below.

robert-uhl: Doh! You’re correct. I didn’t see that in the diff.

src/server/pps/server/api_server.go
```diff
@@ -714,9 +713,10 @@ func (a *apiServer) ListJobSet(request *pps.ListJobSetRequest, serv pps.API_List
 			JobSet: client.NewJobSet(id),
 			Jobs:   jobInfosFiltered,
 		}); err != nil {
-			return err
+			return errors.Wrap(err, "error sending jobset")
```
robert-uhl, line 716: “Jobset” or “job set” or `JobSet`?

src/server/pps/server/api_server.go
```diff
@@ -652,6 +652,12 @@ func (a *apiServer) ListJobSet(request *pps.ListJobSetRequest, serv pps.API_List
 	// Track the jobsets we've already processed
 	seen := map[string]struct{}{}
 
+	// Create a filter function based on the request
```
robert-uhl, line 655: This comment (and the similar one at l. 831) may not add a lot (a bit like `x = 2 * x // double x`), but neither is wrong.

robert-uhl (approved):
Two very minor nits that don't necessarily require action. Look good to me!

PR #8866, opened 2023-05-18
Draft: Djanicek/core 1554/process test results

djanicekpach:
oh boy!
image
image

There is still an issue with data being re-processed too often, so looking into that but this is almost ready.

This PR should get pipeline test results along with some CI metadata to grafana so we can track them overtime. The overal architecture is like this:

- There is a `collector` go script that uploads to pachyderm. This was created previously but needed to have some circle meta data added and also had a refactor as part of that.
- There is an `egress` script that gets built into a docker container and added as a pachyderm pipeline. This egresses to the google cloud sql Postgres where grafana can read it.
- There is a `schema` sql query that creates the tables we use in the aforementioned egress.
- The other sql files are dashboard for grafana that I thought were useful.
- There are some scripts and instructions if you want to test locally.

Some notes to think about for review:

- In the screenshots above half failed because a master pipeline is running that does not have the collector stats. These failures should stop once the updated collector merges and propagates to people's branches. 11 that worked were my branch so it does work, and the inserted row count matches my expectations from local testing.
- We are piggybacking on the pachyderm sql instance but using a different database called `ci_metrics`. This looks like what we do with dex, and I don't see any issue with it since we have plenty of space on that DB, but it seemed liek a good quirk to call out.
- Despite the name, we are not using the built in pachyderm egress functionality :/. It deletes the table contents after every run which would have meant we would need to sync everything all the time which would not have been good.  

First dashboard: https://grafana.pachops.com/d/hndau_U4z/test-results?orgId=1
Pachyderm cluster: https://pachyderm.pachops.com/


etc/testing/circle/workloads/go-test-results/README.md
```diff
@@ -4,5 +4,20 @@
 * adjust variables in `etc/testing/circle/workloads/ci-results/collector/test-collect.sh` as needed
 * create the repos and pipeline images: `eval $(minikube -p minikube docker-env) && echo $(cd etc/testing/circle/workloads/ci-results && ./build-docker.sh)`
 * ensure the results you want to test with are in the `/tmp/test-results` folder 
+* Apply the schema `etc/testing/circle/workloads/go-test-results/schema.sql` to the DB if the tables do not exist.
 * run the collector with `etc/testing/circle/workloads/ci-results/collector/test-collect.sh
 
+If you want to test with local grafan check this page out: https://grafana.com/docs/grafana/latest/setup-grafana/installation/kubernetes/
```
jrockway, line 10: missing a in grafana

etc/testing/circle/workloads/go-test-results/collector/main.go
jrockway, line 1: I'd like to start putting these scripts in our main go repo, not under etc/ but src/.  So src/testing/cmds/go-test-results/ for this file.  This makes it so we only have one go.mod to manage, and so that we can use internal libraries in testing tools.

etc/testing/circle/workloads/go-test-results/collector/main.go
```diff
@@ -54,52 +51,144 @@ func main() {
...
+		}
+	}()
+	// upload general job information
+	jobInfo, err := findJobInfoFromCI()
+	if err != nil {
+		logger.Fatalf("Collecting job info: %v", err)
+	}
+	basePath := findBasePath(jobInfo)
+	if err = uploadJobInfo(basePath, jobInfo, pachClient, commit); err != nil {
+		logger.Fatalf("Uploading job info: %v", err)
```
jrockway, line 89: For log messages, I'd err on the side of indicating a problem.  For errors, this is a good message, for log messages, I'd be explicit: "problem uploading job info".

etc/testing/circle/workloads/go-test-results/collector/main.go
```diff
@@ -54,52 +51,144 @@ func main() {
...
-		logger.Fatalf("Failed creating files: %v", err)
+		logger.Fatalf("Putting files: %v", err)
 	}
+
 	logger.Println("Successfully uploaded files.")
 }
+
+func sanitizeName(s string) string {
+	ret := []byte(strings.ReplaceAll(s, " ", "_"))
+	invalidCharacters := regexp.MustCompile("[^a-zA-Z0-9-_]+")
```
jrockway, line 113: I'd lift regexp.MustCompile out to the package scope; `var invalidNameCharacters = regexp.MustCompile(...)`.

etc/testing/circle/workloads/go-test-results/collector/main.go
```diff
@@ -54,52 +51,144 @@ func main() {
...
+	ret = invalidCharacters.ReplaceAll(ret, []byte("-"))
```
jrockway, line 114: I think you should be consistent about operating on bytes or strings.  Use bytes.ReplaceAll above and cast the argument to a byte slice, or use ReplaceAllString here.  Strings are unicode, bytes are binary; bytes make the most sense to me.

etc/testing/circle/workloads/go-test-results/egress/main.go
```diff
@@ -0,0 +1,194 @@
...
+	Elapsed  float32   `json:"Elapsed"`
+	Output   string    `json:"Output"`
+}
+
+// This is built and runs in a pachyderm pipeline to egress data uploaded to pachyderm to postgresql.
+func main() {
+	logger.Println("Beginning egress transform of data to sql DB")
+	inputFolder := os.Args[1]
+
+	db, err := sqlx.Open("pgx", fmt.Sprintf("user=%s password=%s host=%s port=5432 database=ci_metrics", postgresqlUser, postgresqlPassword, postgresqlHost))
```
jrockway, line 46: This is likely to be somewhat limiting if you ever want to set raw pgx options.  Check out the "fake" DSN we create in dbutil for how to interface sqlx and pgx: https://github.com/pachyderm/pachyderm/blob/master/src/internal/dbutil/db.go#L102

etc/testing/circle/workloads/go-test-results/egress/main.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		logger.Fatalf("Failed opening db connection %v", err)
+	}
+	// filpathWalkDir does not evaluate the PFS symlink
+	sym, err := filepath.EvalSymlinks(inputFolder)
+	if err != nil {
+		logger.Fatal(err)
+	}
+	jobInfoPaths := make(map[string]gotestresults.JobInfo)
+	testResultPaths := []string{}
```
jrockway, line 56: Prefer `var testResultPaths []string`.

etc/testing/circle/workloads/go-test-results/egress/main.go
```diff
@@ -0,0 +1,194 @@
...
+	err = filepath.WalkDir(sym, func(path string, d fs.DirEntry, err error) error {
+		if !d.IsDir() {
```
jrockway, line 58: This isn't strictly problematic, but I'd invert the logic here; `if d.IsDir() { return nil }` and then un-indent the bulk of the code.

I think you need to check the passed-in error before doing anything else.   (`if err != nil { return err }` is likely adequate.)

etc/testing/circle/workloads/go-test-results/egress/main.go
jrockway, line 1: Did you know that we have postgres export built into pachyderm, btw?  https://docs.pachyderm.com/2.5.x/how-tos/basic-data-operations/export-data-out-pachyderm/sql-egress/

etc/testing/circle/workloads/go-test-results/egress/pipeline-local.json
jrockway, line 1: We also have jsonnet templating so that you don't have to maintain two copies of the same file: https://docs.pachyderm.com/2.5.x/how-tos/pipeline-operations/jsonnet-pipeline-specs/

etc/testing/circle/workloads/go-test-results/grafana.yaml
```diff
@@ -0,0 +1,83 @@
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
```
jrockway, line 3: Can you use an emptyDir in the pod volume mounts?  I don't know how much retention you're expecting to have, but maybe a read-only FS + mounting in grafana.yaml is what you're interested in for your test setup?  (I think that's what this manifest is for anyway.)

etc/testing/circle/workloads/go-test-results/sql/schema.sql
jrockway, line 1: You will probably want database schema migrations someday.  I recommend the `tern` library.  https://github.com/jackc/tern

You can build the SQL/migration files into a binary that does the migration when necessary; see `go doc embed`.  Then people can just build a binary that calls migrate on the right place.  (You can also make the worker do this at the start of its work.  Totally fine to run every time.)

jrockway (approved):
This is looking amazing!

etc/testing/circle/workloads/go-test-results/egress/main.go
```diff
@@ -0,0 +1,194 @@
...
+	"io/fs"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgconn"
+	_ "github.com/jackc/pgx/v4/stdlib"
```
jrockway, line 17: We can start using /v5/ for new code, I think.  We should migrate our /v4 code to it eventually, but no need to start outdated on day 1.

etc/testing/circle/workloads/go-test-results/egress/main.go
```diff
@@ -0,0 +1,194 @@
...
+		// output is mostly logs and coverage and it's not that useful, so we don't store output-only actions
+		if result.Action != "output" {
+			splitPath := strings.Split(path, "/")
+			fileName := splitPath[len(splitPath)-1]
+			// get job info previously inserted in order to link foreign keys to jobs table
+			jobInfo, ok := jobInfoPaths[strings.TrimSuffix(path, fileName)]
+			if !ok {
+				return errors.WithStack(fmt.Errorf("Failed to find job info for %v - file name %v - job infos %v ", path, fileName, jobInfoPaths))
+			}
+			_, err = db.Exec(`INSERT INTO public.test_results (
```
jrockway, line 167: BTW, there is a NamedExec method where you can use :go_name: instead of $1 placeholders, and then just pass a struct instead of individual args.

jrockway (commented):
More random comments from a second pass.

etc/testing/circle/workloads/go-test-results/egress/main.go
```diff
@@ -0,0 +1,194 @@
...
+		jobInfo.Commit,
+		jobInfo.Branch,
+		jobInfo.Tag,
+		jobInfo.PullRequests,
+	)
+	if err != nil {
+		var pgErr *pgconn.PgError
+		if errors.As(err, &pgErr) && pgErr.Code == "23505" {
+			// Unique key constraint violation. In other words, the record already exists.
+			// Ignore insert to since we don't want to change the timestamp, but log it. This is expected with multiple executors.
```
djanicekpach, line 132: This comment is outdated but close - should be updated now that the executors are a folder.

jrockway: In your SQL you can write "ON CONFLICT DO ...".  The ... can be NOTHING if you don't care, or an update statement if you want to do an update instead of an insert.  It's not a big deal to do it in Go, but Postgres can save you some trouble if you want.

etc/testing/circle/workloads/go-test-results/egress/main.go
djanicekpach: I think this is actually a pretty neat use case difference. The use case the built-in egress seems to be catering to is a "mirroring" case. So like, if I want to mirror my pachyderm repository exactly into a SQL DB, then the egress probably works pretty well.(Although it still seems slow to me) However, to enforce this mirroring we delete all the data in every table we get data for from the egress before inserting it. Talking with Albert, this was as designed.
image

Since I'm working with over time data, I'm kind of using pachyderm as a de-duplication system for the raw results and then expanding that data into SQL for reporting, which requires appending to existing data. I considered adding an `append` flag to the egress pipeline spec, but without a real customer use case it did not seem worth it. I would be open to that change in the future though.

etc/testing/circle/workloads/go-test-results/egress/main.go
djanicekpach: good call

etc/testing/circle/workloads/go-test-results/egress/main.go
djanicekpach: I changed this for ci_jobs table, but since test-results pulls from 2 separate structs NamedExec didn't seem as helpful.

etc/testing/circle/workloads/go-test-results/sql/schema.sql
djanicekpach: added to the worker code since I think that makes a lot of sense to just auto-migrate.

etc/testing/circle/workloads/go-test-results/egress/main.go
djanicekpach: I tried this and I have to chose between v5 and using dbutil :/ 
I went with using the dbutil.GetDB and hopefully when that upgrades, this file should upgrade with an import change.

jrockway: that's great.  i'm glad we're using the same libraries/helper functions for everything.

src/testing/cmds/go-test-results/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.20.3-alpine
```
jrockway, line 1: 1.20.4 is current now.

PR #8867, opened 2023-05-18
Add alias flag to pachctl connect

BOsterbuhr:
adds the ability to set the context's name when using pachctl connect


tybritten (approved):
LGTM

PR #8870, opened 2023-05-19
pachyderm_sdk: separate config parsing into its own module

kevyang:
Moving config parsing out into its own module. This will make testing and extending its capabilities easier. Reserving further improvements on parsing for a future PR.

Jira: [INT-916]

[INT-916]: https://pachyderm.atlassian.net/browse/INT-902

### DIFF
--- a/python-sdk/pachyderm_sdk/client.py
+++ b/python-sdk/pachyderm_sdk/client.py
@@ -20,6 +20,7 @@ from .api.pfs.extension import ApiStub as _PfsStub
 from .api.pps.extension import ApiStub as _PpsStub
 from .api.transaction.extension import ApiStub as _TransactionStub
 from .api.version import ApiStub as _VersionStub, Version
+from .config import ConfigFile
 from .constants import (
     AUTH_TOKEN_ENV,
     GRPC_CHANNEL_OPTIONS,
@@ -229,7 +230,7 @@ class Client:
         """
         # TODO: Should config_file be nullable?
         #  If null should we search for the local config?
-        config = _ConfigFile(config_file)
+        config = ConfigFile(config_file)
         active_context = config.active_context
         client = cls.from_pachd_address(
             active_context.active_pachd_address,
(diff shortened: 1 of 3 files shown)


python-sdk/pachyderm_sdk/client.py
```diff
@@ -229,7 +230,7 @@ def from_config(cls, config_file: Union[Path, str]) -> "Client":
         """
         # TODO: Should config_file be nullable?
         #  If null should we search for the local config?
-        config = _ConfigFile(config_file)
+        config = ConfigFile(config_file)
```
bbonenfant, line 233: Can't comment on exact line, but it might be worth moving the `env_config`, `spout_config` and `local_config` class variables in this PR. Maybe move them to some combination of `constants.py` and the new `config.py`?

python-sdk/tests/test_config.py
```diff
@@ -0,0 +1,75 @@
+import unittest
```
bbonenfant, line 1: I think this unittest import is unused

python-sdk/tests/test_config.py
```diff
@@ -0,0 +1,75 @@
...
+      }
+    },
+    "metrics": true
+  }
+}
+""", 'utf-8')
+
+class TestConfig():
+    @staticmethod
+    def test_from_file():
```
bbonenfant, line 56: You can choose if you want to use this, but I opt to use tools that pytest provides. An example using this test would be:
```python
@staticmethod
def test_from_file(tmp_path):
    config_path = tmp_path / config.json
    config_path.write_bytes(TEST_CONFIG)
    c = ConfigFile(config_path)
    ...
```
(above code not tested)

The tmp_path gives you a `pathlib.Path` object to a temporary directory that gets cleaned-up after the test finishes. That way you don't need to use the `NamedTemporaryFile` and its somewhat weird api.

bbonenfant (approved):
This looks good pending the comments I left

PR #8873, opened 2023-05-22
update opencv example stock images

djanicekpach:
imgur has decided to rate limit image downloads. The tests need these examples for pipelines to work. 

We should switch to using images with free use licenses like creative commons on a  different host. 

To unblock pipelines I've grabbed some similar images to open cv and linked to those instead. 

Example:
https://free-images.com/display/video_game_gamescom_star.html

I think we should probably host these images ourselves, and we still have a LOT of imgur references so we will need to figure out what to do with this long term.


jrockway (approved):
LGTM

PR #8877, opened 2023-05-22
update img after rate limiting

djanicekpach:
Missed one test when committing stop-gap for imgur rate limiting. 
Previous PR https://github.com/pachyderm/pachyderm/pull/8873


djanicekpach:
> We could probably host these ourselves if this becomes an issue

That is the plan in CORE-1697! Assigned to me next sprint.

molinamelendezj:
@djanicekpach can you put up a back port PR to 2.6.x for this. it just failed the release tests.

PR #8878, opened 2023-05-23
Relax 2.6 Duplicate alias check

acohen4:


jrockway (approved):

src/internal/clusterstate/v2.6.0/pfsdb.go
```diff
@@ -84,33 +84,22 @@ func validateExistingDAGs(cis []*v2_5_0.CommitInfo) error {
 		duplicates[commitBranchlessKey(ci.Commit)][oldCommitKey(ci.Commit)] = ci
 	}
 	// the only duplicate commits we allow and handle in the migration are two commits with a parent/child relationship
```
brycemcanally, line 86: nit, can we change this comment to "commits connected through ancestry". Also, maybe add a note about this being for the trigger case.

brycemcanally (approved):
lgtm, just a nit about the comment.

PR #8882, opened 2023-05-23
Update helm values for console

pleeko:
We want to prefix `PACHYDERM_PUBLIC_HOST` and `PACHYDERM_PUBLIC_TLS` with `REACT_APP_RUNTIME` so that they can be injected into the react app.


jrockway (approved):
That makes sense!

PR #8883, opened 2023-05-23
fix list commit showing inconsistent sizes (#8876)

molinamelendezj:
* add SizeBytesUpperBound to list commit with branch


brendoncarroll (approved):
LGTM

PR #8884, opened 2023-05-23
[CORE-1947] Switch minuend and subtrahend

robert-uhl:
Subtract the lesser from the greater to get a positive integer.


jrockway (commented):
This looks good, but there's another case:

```
src/server/pps/server/api_server.go
1815:           return errors.Errorf("Kubernetes name %q is %d characters longer than the %d max", k8sName, dnsLabelLimit-len(k8sName), dnsLabelLimit)
```

jrockway (approved):
Very nice cleanup!

src/pfs/pfs_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestProject_ValidateName(t *testing.T) {
```
brendoncarroll, line 11: Is it idiomatic to have underscores in test names?

brendoncarroll (approved):

src/pfs/pfs_test.go
robert-uhl: I stole the pattern from [example names](https://pkg.go.dev/testing#hdr-Examples), which for methods are of the form `ExampleProject_ValidateName`. Granted, it is really weird to see an underscore in a name in Go.

PR #8889, opened 2023-05-25
fix error handling and username

djanicekpach:
The nightly exposed problems with the data upload when there's no user. To resolve them:

1. I added username to the sql schema. It was a late addition and missed getting threaded through to the DB.
2. I allowed username to be empty since nightly runs apparently do not always have a user.
3. I changed the error handling so we don't use exit in logic. This was preventing the finishCommit from running deferred and blocking future commits.


molinamelendezj (approved):
Looks good to me. Not sure about the create being deleted then having an alter. If we ever needed to recreate we'd have to find that create script. Something to think about with maybe migration in future should this table get a lot of updates.

PR #8900, opened 2023-05-30
Debug Dump V2

acohen4:


src/server/debug/cmds/cmds.go
```diff
@@ -87,9 +91,32 @@ func Cmds(mainCtx context.Context, pachctlCfg *pachctl.Config) []*cobra.Command
...
+			client, err := pachctlCfg.NewOnUserMachine(mainCtx, false)
+			if err != nil {
+				return err
+			}
+			defer client.Close()
+			r, err := client.GetDumpV2Template(mainCtx, &debug.GetDumpV2TemplateRequest{})
+			if err != nil {
+				return err
+			}
+			bytes, err := json.MarshalIndent(r.Request, "", "    ")
```
jrockway, line 108: This should be jsonpb.

src/server/debug/server/database_dump.go
```diff
@@ -1,38 +1,41 @@
...
+func (s *debugServer) collectDatabaseDump(ctx context.Context, dir string, rp reportProgressFunc) (retErr error) {
 	defer log.Span(ctx, "collectDatabaseDump")(log.Errorp(&retErr))
 	ctxWithTimeout, cancel := context.WithTimeout(ctx, 1*time.Minute)
 	defer cancel()
+	rp(ctx, 0, 100)
 	var errs error
-	if err := s.collectDatabaseStats(ctxWithTimeout, tw, prefix...); err != nil {
-		errors.JoinInto(&errs, errors.Wrap(err, "collectDatabaseStats"))
+	if err := s.collectDatabaseStats(ctxWithTimeout, dir); err != nil {
+		multierr.AppendInto(&errs, errors.Wrap(err, "collectDatabaseStats"))
```
jrockway, line 29: errors.JoinInto is pachyderm/pachyderm's version of multierr.AppendInto.  You'll notice that the linter gets mad about importing multierr now.

src/server/debug/server/server.go
```diff
@@ -82,123 +80,59 @@ func NewDebugServer(env serviceenv.ServiceEnv, name string, sidecarClient *clien
...
-				if err := s.collectDatabaseDump(ctx, tw, databasePrefix); err != nil {
-					errors.JoinInto(&errs, errors.Wrap(err, "collectDatabaseDump"))
-				}
-			}
-			return errs
-		})
-	})
+// TODO:
+// Is it good that DumpRequest contains specific pods? The benefit is explicitness.
+// The downside is that the same template may be less relevant if evaluated later.
```
jrockway, line 89: I think we should have some way to select classes of pods; probably by label.  The default templates should use selectors instead of explicitly-named pods.

src/server/debug/server/server.go
```diff
@@ -214,67 +148,364 @@ func (s *debugServer) appLogs(ctx context.Context, tw *tar.Writer) (retErr error
...
+			}
+			return errors.Wrap(eg.Wait(), "run dump tasks")
+		})
 	})
 }
 
+func (s *debugServer) lokiLogs(ctx context.Context, dir string, pod *debug.Pod) (retErr error) {
+	ctx, end := log.SpanContext(ctx, "collectLokiLogs")
+	defer end(log.Errorp(&retErr))
+	ctx, cf := context.WithTimeout(ctx, 10*time.Minute)
```
jrockway, line 484: These should be configurable.

src/server/debug/server/server.go
```diff
@@ -480,133 +640,34 @@ func writeProfile(ctx context.Context, w io.Writer, profile *debug.Profile) (ret
...
-		request.Filter,
-		s.collectPachdDumpFunc(request.Limit),    /* collectPachd */
-		s.collectPipelineDumpFunc(request.Limit), /* collectPipeline */
-		s.collectWorkerDump,                      /* collectWorker */
-		redirectDump,                             /* redirect */
-		collectDump,                              /* collect */
-		true,
-		true,
-	)
+	return errors.New("Dump gRPC is deprecated in favor of DumpV2")
```
jrockway, line 661: status.Error(codes.Unimplemented, ...

jrockway (commented):
Overall this is looking good.  I think that `dir` as the API surface that dumpers have to interact with the dumping machinery is probably not the best.  There should probably be some sort of abstraction there.  The API probably wants to be similar to fs.FS but with write functionality.  Open -> File, file.Write and file.Close.  (Open/Write writes to a tmpfile, Close flushes to the underlying tar writer.)  Progress probably wants to be in this API as well, dump.StartOperation -> Operation, operation.Update(), operation.Finish().  I don't think the prefix... stuff needs to live on, each dumper should have an absolute file path in mind.  (You can have duplicate files in a TAR, and we already do... this seems to work out OK when I uncompress in Windows, but dunno about Linux.)

We should have some progress info for copying disk files into the tar stream.

Overall I'd look to the places where the code is indented 100 spaces and probably factor those out into separate functions.  RunTask(tarWriter) is probably warranted.

src/server/debug/server/server.go
acohen4: let's talk about this 👍

src/server/debug/server/server.go
jrockway: Let's do this in a follow-up PR.

src/server/debug/server/server.go
jrockway: We forgot to talk about this.  I think we can leave this as-is for now, but address this when we address pod/container differences in the spec.

src/server/debug/server/server.go
```diff
@@ -1135,7 +1226,7 @@ func (s *debugServer) collectJobs(tw *tar.Writer, pachClient *client.APIClient,
 			StrokeColor: chart.GetDefaultColor(2).WithAlpha(255),
 		},
 	}
-	if err := collectDebugFile(tw, "jobs", "json", func(w io.Writer) error {
+	if err := dfs.Write(filepath.Join(pipeline.Project.Name, pipeline.Name, "jobs.json"), func(w io.Writer) error {
 		// TODO: The limiting should eventually be a feature of list job.
 		var count int64
 		return pachClient.ListJobF(pipeline.Project.Name, pipeline.Name, nil, -1, false, func(ji *pps.JobInfo) error {
```
jrockway, line 1232: probably some of these args can be configurable through the template

src/server/debug/server/server.go
```diff
@@ -208,73 +143,418 @@ func (s *debugServer) appLogs(ctx context.Context, tw *tar.Writer) (retErr error
...
+
+func makeProfilesTask(server debug.Debug_DumpV2Server, apps []*debug.App) taskFunc {
+	return func(ctx context.Context, dfs DumpFS) error {
+		var errs error
+		rp := recordProgress(server, "profiles", len(apps))
+		for _, app := range apps {
+			for _, pod := range app.Pods {
+				for _, c := range pod.Containers {
+					for _, profile := range []string{"heap", "goroutine"} {
+						if err := dfs.Write(filepath.Join(appDir(app), "pods", pod.Name, c, profile), func(w io.Writer) (retErr error) {
```
jrockway, line 352: this is where i'd split into a helper function; collectProfile(pod, container, profile) or whatever.

src/server/debug/server/server.go
```diff
@@ -208,73 +143,418 @@ func (s *debugServer) appLogs(ctx context.Context, tw *tar.Writer) (retErr error
...
+	server debug.Debug_DumpV2Server
+}
+
+func (dcs *dumpContentServer) Send(bytesValue *types.BytesValue) error {
+	return dcs.server.Send(&debug.DumpChunk{Chunk: &debug.DumpChunk_Content{Content: &debug.DumpContent{Content: bytesValue.Value}}})
+}
+
+type incProgressFunc func(context.Context)
+
+// TODO should this be sending in a goro?
```
jrockway, line 454: Only one goroutine can call Send at a time.  If we parallelize dumping this will need additional synchronization.

src/server/debug/server/server.go
```diff
@@ -208,73 +143,418 @@ func (s *debugServer) appLogs(ctx context.Context, tw *tar.Writer) (retErr error
...
+	timeout := 25 * time.Minute // If no client-side deadline set, use 25 minutes.
+	if deadline, ok := server.Context().Deadline(); ok {
+		d := time.Until(deadline)
+		// Time out our own operations at ~90% of the client-set deadline. (22 minutes of
+		// server-side processing for every 25 minutes of debug dumping.)
+		timeout = time.Duration(22) * d / time.Duration(25)
+		var cf context.CancelFunc
+		ctx, cf = context.WithTimeout(ctx, timeout)
+		defer cf()
+	}
```
jrockway, line 494: I think the ctx.WithTimeout needs to be dedented from the deadline ok block.  If the deadline isn't OK, then we want to timeout at 25 minutes.

src/server/debug/server/server.go
```diff
@@ -208,73 +143,418 @@ func (s *debugServer) appLogs(ctx context.Context, tw *tar.Writer) (retErr error
...
+					})
+				}(f)
+			}
+			return errors.Wrap(eg.Wait(), "run dump tasks")
+		})
 	})
 }
 
+func (s *debugServer) kubeLogs(ctx context.Context, dfs DumpFS, app *debug.App, pod *debug.Pod) (retErr error) {
+	ctx, end := log.SpanContext(ctx, "collectLokiLogs")
```
jrockway, line 545: this one is kubeLogs, not lokiLogs.

src/server/debug/server/server.go
```diff
@@ -689,13 +788,14 @@ func (s *debugServer) collectCommits(rctx context.Context, tw *tar.Writer, pachC
 			}
 			return errors.EnsureStack(s.marshaller.Marshal(w, ci))
 		})
-	}, prefix...); err != nil {
+	}); err != nil {
 		return err
 	}
 	// Reverse the x values since we collect them from newest to oldest.
 	// TODO: It would probably be better to support listing jobs in reverse order.
```
jrockway, line 795: This comment talks about listing jobs in reverse order, which we can do, but this function is listing commits, which we can't do in reverse order 😂

src/server/debug/server/server.go
```diff
@@ -759,23 +859,45 @@ func collectGoInfo(tw *tar.Writer, prefix ...string) error {
...
 		return nil
-	}, prefix...)
+	})
 }
 
-func (s *debugServer) collectPachdVersion(ctx context.Context, tw *tar.Writer, pachClient *client.APIClient, prefix ...string) error {
-	return collectDebugFile(tw, "version", "txt", func(w io.Writer) (retErr error) {
-		defer log.Span(ctx, "collectPachdVersion")(log.Errorp(&retErr))
+func (s *debugServer) collectPachdVersion(ctx context.Context, dfs DumpFS, pachClient *client.APIClient) error {
+	return dfs.Write("/pachd/version.txt", func(w io.Writer) error {
```
jrockway, line 866: Someday this probably wants to ask each component (other pachd replicas, pachw, workers) their version, so we can detect faulty upgrades.

jrockway (approved):
This looks pretty good to me.

I don't fully understand the mechanics of recursing to the workers for things like profiles; I would think they should get their own template to control how they are inspected.  Right now, I don't think we can do things like "take a cpu profile of pachw", which should be possible.

Some follow-ups that we'll want to do ASAP:

1. IP/pod hash-less templates by default.
2. Ability to recurse to the other pachds for profiling, version, go_info, etc.
3. Download progress

src/server/debug/server/server.go
acohen4: so dumping is currently being parallelized. It doesn't just block on a channel?

jrockway: go doc grpc.ServerStream says:

>         // It is safe to have a goroutine calling SendMsg and another goroutine
>        // calling RecvMsg on the same stream at the same time, but it is not safe
>        // to call SendMsg on the same stream in different goroutines.

PR #8905, opened 2023-05-31
pachyderm_sdk: default to local pachyderm config if no config file provided

kevyang:
default to local pachyderm config if no config file provided

### DIFF
--- a/python-sdk/pachyderm_sdk/client.py
+++ b/python-sdk/pachyderm_sdk/client.py
@@ -219,7 +219,7 @@ class Client:
         ----------
         config_file : Union[Path, str]
             The path to a config json file.
-            If config_file is None, from_config searches for the local config.
+            If config_file is None, from_config defaults to the local config.
 
         Returns
         -------
--- a/python-sdk/pachyderm_sdk/client.py
+++ b/python-sdk/pachyderm_sdk/client.py
@@ -219,7 +219,7 @@ class Client:
         ----------
         config_file : Union[Path, str]
             The path to a config json file.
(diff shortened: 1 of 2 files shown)


python-sdk/pachyderm_sdk/config.py
```diff
@@ -6,9 +6,10 @@
 from typing import Dict, Optional, Union
 
 from .errors import ConfigError
+from .constants import CONFIG_PATH_LOCAL
 
 class ConfigFile:
-    def __init__(self, config_file: Union[Path, str]):
+    def __init__(self, config_file: Union[Path, str]=CONFIG_PATH_LOCAL):
```
bbonenfant, line 12: I wouldn't add a default here. In practice I've found when args are passed through a chain of functions, only have a default arg when they first appear results in the least bugs. In this case I would default it at the Client and not in Config

python-sdk/pachyderm_sdk/client.py
```diff
@@ -212,21 +212,20 @@ def from_pachd_address(
...
     @classmethod
-    def from_config(cls, config_file: Union[Path, str]) -> "Client":
+    def from_config(cls, config_file: Union[Path, str]=CONFIG_PATH_LOCAL) -> "Client":
         """Creates a Pachyderm client from a config file.
 
         Parameters
         ----------
         config_file : Union[Path, str]
             The path to a config json file.
+            If config_file is None, from_config searches for the local config.
```
bbonenfant, line 222: I would change the working here to say that it defaults to the local config

PR #8911, opened 2023-06-05
Create, Get, Update, and Delete Functions for Projects

FahadBSyed:


src/internal/coredb/projects.go
```diff
@@ -0,0 +1,96 @@
...
+	project.CreatedAt, err = types.TimestampProto(createdAt)
+	if err != nil {
+		return nil, errors.EnsureStack(err)
+	}
+	return project, nil
+}
+
+// UpdateProject updates all fields of an existing project entry in the core.projects table by name. If 'upsert' is set to true, UpdateProject()
+// will attempt to call CreateProject() if the entry does not exist. If updateTime is not provided, the current time will be used.
+func UpdateProject(ctx context.Context, tx *pachsql.Tx, project *pfs.ProjectInfo, upsert bool, updateTime ...*types.Timestamp) error {
```
albscui, line 57: Hmm the `updateTime ...*types.Timestamp` is annoying. I'm guess you are using variadic arguments to support optional arguments. Ideally, `pfs.ProjectInfo` already has the `updated_at` field.

src/internal/coredb/projects_test.go
```diff
@@ -0,0 +1,87 @@
...
+	testProjDesc = "this is a test project"
+)
+
+func TestProjectCreateAndGet(t *testing.T) {
+	ctx := pctx.TestContext(t)
+	db := dockertestenv.NewTestDB(t)
+	tx, err := db.BeginTxx(ctx, nil)
+	require.NoError(t, err, "should be able to create a tx")
+	defer tx.Rollback()
+	require.NoError(t, v2_7_0.SetupCore(ctx, tx), "should be able to set up core tables")
```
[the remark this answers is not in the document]
FahadBSyed: I don't have any objections to that 👍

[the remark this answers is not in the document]
FahadBSyed: Will do

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
+	"github.com/pachyderm/pachyderm/v2/src/pfs"
+	"time"
+)
+
+// CreateProject creates an entry in the core.projects table.
+func CreateProject(ctx context.Context, tx *pachsql.Tx, project *pfs.ProjectInfo) error {
+	_, err := tx.ExecContext(ctx, "INSERT INTO core.projects (name, description) VALUES ($1, $2);", project.Project.Name, project.Description)
+	//todo: insert project.authInfo into auth table.
+	return errors.EnsureStack(err)
```
albscui, line 17: No more `EnsureStack()` always use `Wrap()` instead.

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,80 @@
...
+
+// DeleteProject deletes an entry in the core.projects table.
+func DeleteProject(ctx context.Context, tx *pachsql.Tx, projectName string) error {
+	_, err := tx.ExecContext(ctx, "DELETE FROM core.projects WHERE name = $1;", projectName)
+	//todo: delete corresponding project.authInfo auth table.
+	return errors.EnsureStack(err)
+}
+
+// GetProject retrieves an entry from the core.projects table by project name.
+func GetProject(ctx context.Context, tx *pachsql.Tx, projectName string) (*pfs.ProjectInfo, error) {
```
albscui, line 28: I wonder if  read operations that are transactional should have a `Tx` suffix, like `GetProjectTx`.

FahadBSyed: 🤔 

I would say no because we only support one Get() function today which takes a transaction as a parameter. So that should be implicit in the behavior of the function. If we had multiple Get() functions, then maybe we could argue that their names should include `Tx`

albscui: But I definitely see us potentially having a non-transactional Get() in the future.

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,88 @@
...
+	"fmt"
+	"github.com/gogo/protobuf/types"
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/pfs"
+	"time"
+)
+
+// Querier defines an interface for functions shared across sqlx.Tx and sqlx.DB types.
+// Functions that take a querier support both running inside a transaction and outside it.
+type Querier interface {
```
albscui, line 15: Can we just embed the sqlx interfaces? Like
```
type Querier interface {
    sqlx.ExecerContext
    sqlx.QueryerContext
}
```

FahadBSyed: sure, I can try that

albscui (approved):
This is a beautiful API!

PR #5549, opened 2020-12-21
Finish V2 / V1 merge

brycemcanally:
This PR contains the rest of the changes needed to merge the V2 / V1 code paths.

PR #8919, opened 2023-06-06
gogo removal part 2: avoid all gogo-specific features

jrockway:
This removes the dependency on anything special gogoproto can do.  Phase 3 actually uses the standard compiler instead of gogo.


etc/proto/protoc-gen-zap/main.go
```diff
@@ -178,15 +178,7 @@ func generatePrimitiveField(g *protogen.GeneratedFile, f *protogen.Field, opts *
...
-			if proto.GetExtension(opts, gogoproto.E_Stdtime).(bool) {
-				// If gogoproto.stdtime is set, then this is a pointer to a
-				// time.Time, not a types.Timestamp.
-				g.P(`if t := x.`, gname, `; t != nil {`)
-				g.P(`enc.AddTime("`, fname, `", *t)`)
-				g.P(`}`)
-			} else {
-				g.P(g.QualifiedGoIdent(extensionPkg.Ident("AddTimestamp")), `(enc, "`, fname, `", x.`, gname, `)`)
-			}
+			g.P(g.QualifiedGoIdent(extensionPkg.Ident("AddTimestamp")), `(enc, "`, fname, `", x.`, gname, `)`)
```
brendoncarroll, line 181: 👍

src/server/license/server/api_server.go
```diff
@@ -327,7 +354,7 @@ func (a *apiServer) UpdateCluster(ctx context.Context, req *lc.UpdateClusterRequ
 
 func (a *apiServer) ListUserClusters(ctx context.Context, req *lc.ListUserClustersRequest) (resp *lc.ListUserClustersResponse, retErr error) {
 	clusters := make([]*lc.UserClusterInfo, 0)
-	if err := a.env.DB.SelectContext(ctx, &clusters, `SELECT id, cluster_deployment_id, user_address, is_enterprise_server FROM license.clusters WHERE is_enterprise_server = false`); err != nil {
+	if err := a.env.DB.SelectContext(ctx, &clusters, `SELECT id, cluster_deployment_id AS ClusterDeploymentId, user_address AS Address, is_enterprise_server AS EnterpriseServer FROM license.clusters WHERE is_enterprise_server = false`); err != nil {
```
brendoncarroll, line 357: ✅ We should look into generating these as additional tags.  sqlx looks at `db` struct tags.

src/server/identity/server/api_server.go
```diff
@@ -134,7 +134,7 @@ func (a *apiServer) setIdentityServerConfig(ctx context.Context, req *identity.S
 
 func (a *apiServer) GetIdentityServerConfig(ctx context.Context, req *identity.GetIdentityServerConfigRequest) (resp *identity.GetIdentityServerConfigResponse, retErr error) {
 	var config []*identity.IdentityServerConfig
-	err := a.env.DB.SelectContext(ctx, &config, "SELECT issuer, id_token_expiry, rotation_token_expiry FROM identity.config WHERE id=$1;", configKey)
+	err := a.env.DB.SelectContext(ctx, &config, "SELECT issuer, id_token_expiry AS idtokenexpiry, rotation_token_expiry AS rotationtokenexpiry FROM identity.config WHERE id=$1;", configKey)
```
brendoncarroll, line 137: What's the deal with the lower case here?   The other one matched the struct field with the upperCamelCase.

src/server/license/server/api_server.go
jrockway: Yeah, that was a feature gogo had, but I don't have any good ideas for getting it back.  gopatch can't add struct tags.  We could write a generator to make these dbXXX structs separately (complete with ToProto or whatever), if we wanted.

We could make sqlx use the "json" tag instead of "db", but that breaks all the existing hand-written structs that use "db:...", and worse, silently changes the behavior of every database query in the system.  When doing big refactors like this reflection is super scary!

Finally, it's usually dumb luck that you can scan into these things anyway.  You cannot read a time in the DB into a timestamppb.Timestamp, for example.  As soon as you hit one of those non-standard types, you have to do this second struct thing.   (gogo let you use a time.Time instead of a types.Timestamp, avoiding this problem occasionally.)

etc/testing/circle/workloads/go-test-results/collector/main.go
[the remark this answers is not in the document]
jrockway: I don't know where this came from; must be from a rebase or something.  I'll just delete it, since it's obviously not in master.

src/internal/protoutil/protoutil.go
```diff
@@ -0,0 +1,48 @@
+// Package protoutil contains some utilities for interacting with protocol buffer objects.
```
brendoncarroll, line 1: I like this.

You might be able to implement `MustTimestampFromPointer` in terms of `MustTimestamp`.  That would make it more difficult for their (currently identical) error messages to get out of sync.

src/server/auth/cmds/cmds.go
```diff
@@ -394,7 +395,11 @@ func WhoamiCmd(ctx context.Context, pachctlCfg *pachctl.Config) *cobra.Command {
 			}
 			fmt.Printf("You are \"%s\"\n", resp.Username)
 			if resp.Expiration != nil {
-				fmt.Printf("session expires: %v\n", resp.Expiration.Format(time.RFC822))
+				if t, err := types.TimestampFromProto(resp.Expiration); err != nil {
+					fmt.Fprintf(os.Stderr, "session exipration time unparseable: %v\n", err)
```
brendoncarroll, line 399: Why does one of these go to stdout (implicitly) and one go to stderr?

I think both should be explicit, and probably go to the same place.  If you expect another program to consume it, it should go to stdout.  If it's helpful information, subject to change, it should go to stderr.

src/server/auth/server/api_server.go
```diff
@@ -1629,7 +1637,11 @@ func (a *apiServer) ExtractAuthTokens(ctx context.Context, req *auth.ExtractAuth
 func (a *apiServer) RestoreAuthToken(ctx context.Context, req *auth.RestoreAuthTokenRequest) (resp *auth.RestoreAuthTokenResponse, retErr error) {
 	var ttl int64
 	if req.Token.Expiration != nil {
-		ttl = int64(time.Until(*req.Token.Expiration).Seconds())
+		t, err := types.TimestampFromProto(req.Token.Expiration)
+		if err != nil {
+			return nil, errors.Wrap(err, "convert Token.Expiration to time.Time")
+		}
+		ttl = int64(time.Until(t).Seconds())
```
brendoncarroll, line 1644: ✅ Confirmed this is in seconds.

brendoncarroll (approved):
Overall LGTM.

Please double check what's going on with that SQL statement with the lowercase.  I don't see how that can Scan into the protobuf.  It might just not be populating anything.

src/server/identity/server/api_server.go
jrockway: I'll look into this in more detail.  I think I did this one first, and my reading of the code was that it defaults to strings.ToLower to map the struct member names to db fields.  Then I had to put this aside for a week or so, forgot that, and did the other ones... which also appear to work.  I'll make it consistent though.

src/internal/protoutil/protoutil.go
jrockway: So in part 3, neither of these can error anymore, so I'm fine leaving this as is for now.  With the new API, you get <timestamp>.AsTime() and timestamppb.New(<go time>), and neither can error.

src/server/worker/common/common.proto
```diff
@@ -16,7 +15,6 @@ message Input {
   string group_by = 6;
   bool lazy = 7;
   string branch = 8;
-  string git_url = 9 [(gogoproto.customname) = "GitURL"];
```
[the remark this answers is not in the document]
jrockway: Yeah, it has been unused since 2.0.  We deleted all things that read this field in 81144a5214e69492775d7fd52c7b48c359337130 Finish V2 / V1 merge (#5549).  (I did a git bisect with a regex.)  So consider this part of the pre-2.0 work that was un-done until now ;)

src/server/auth/cmds/cmds.go
jrockway: The parse error goes to stderr, the actual message goes to stdout.  I don't have a problem with using fmt.Printf for things that should go to stdout.  In the follow-up to this, timestamp -> time conversion can't error, so this code path goes away :/

src/server/license/server/api_server.go
```diff
@@ -269,15 +270,41 @@ func (a *apiServer) DeleteAll(ctx context.Context, req *lc.DeleteAllRequest) (re
 	return &lc.DeleteAllResponse{}, nil
 }
 
+type dbClusterStatus struct {
```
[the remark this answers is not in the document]
jrockway: I don't mind standardizing on that.  when ToProto() is only used in one place, you might as well implement it in the for loop that copies into the result slice.

chainlink (approved):
LGTM pending tests on python-sdk and jupyter-extension.

chainlink:
Just out of curiosity, will this break the go-client? (I know we've moved a lot to `internal` but just from the use of the go-client in tests, seems like a bunch of definitions have changed)

PR #8921, opened 2023-06-07
gogo: get rid of

jrockway:
This is mostly the final part of the "get rid of gogo" saga.   After this, we'll have no dependency on gogo anywhere.  This PR switches the generation to the standard generator (!) as well as the "userspace" things like github.com/gogo/protobuf/types to google.golang.org/protobuf/types.  Also /proto and jsonpb -> protojson.

I added some lint rules to ensure that nobody uses the wrong version of the proto libraries, which is horrifyingly easy to do.

I added a test for the protos copied over from 2.5.x as part of the 2.6 migration, and now generate them from a copied .proto file instead of by copying the generated code over.  Other options were to allow gogo inside that package (messy and unnecessary; the wire format doesn't change with a change in the generator), or regenerate the protos with protoc-gen-go in 2.5.x and copy over the generated file.  I tried both and didn't like either way.

serde changed a little bit.  protojson doesn't have a Decoder like gogo did, so I looked at how that was implemented and implemented it in our protoutil package.  (Basically, use a normal JSON.Decoder to unmarshal JSON chunks into `json.RawMessage`, then pass that to protojson.  Exactly the approach gogo used; the standard library doesn't want to add any features not in the protobuf spec, and files containing multiple JSON-marshaled protos aren't in the spec.  So now we explicitly own that part of our file format.)

While I was debugging the tests, I noticed:
* src/server/debug/server tests were running twice.  Fixed that.
* test-go was doing caching wrong (restoring cache AFTER the builds) and installing goreleaser for no reason.  Fixed those.
* the worker logs `workerId` twice for every message.  This is a mistake from the logging refactor, and I fixed it.  Gogo's JSON parser didn't care, but `protojson` does care.  So that users can read log messages from broken versions, I adjusted the parser to handle this case.  (Unmarshal to map[string]any, Marshal, Unmarshal to proto.  Slow!  But only done on parse errors, so relatively safe.)

There is a lot of cleanup that can probably be done after this PR:

* The whole mock pachd thing implements <apiserver>.UnimplementedAPIServer now, which means that a lot of methods like `func Foo(req ...) (res, err) { return nil, errors.Unimplemented(...)` already exist.
* We have a lot of utilities for working with gogo's very annoying well-known types.  The new API fixes those problems (errors returned where errors aren't necessary), but some code still goes through wrappers we wrote.
* `make proto` should probably run python_sdk's proto generator.


.circleci/main.yml
```diff
@@ -90,6 +89,7 @@ jobs:
             --junitfile ${TEST_RESULTS}/gotestsum-report.xml \
             --jsonfile "${TEST_RESULTS}/go-test-results.jsonl" \
             --rerun-fails \
+            --rerun-fails-max-failures=100 \
```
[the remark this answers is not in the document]
jrockway: Yeah.  It's < 10% of the tests so I think fair.

chainlink:
Looks good! I don't think the python stuff will be affected (doesn't look like any changes to the core protos, and there's no explicit mount server changes)

Just out of curiosity. Have we tried using a non gogo pachctl to connect to an old gogo cluster?

jrockway:
Yup, I've tried old clients against a new server.  The wire format of the RPCs are not changing, just how we deal with them in user space.  (Also some quirks around how empty JSON files are parsed, and how JSON files with duplicate keys are handled.  We work around those in the necessary place; config loading that works with "PACH_CONFIG=/dev/null pachctl ..." and pps.LogMessage parsing with duplicate keys in the PPS API server.)

chainlink:
Awesome! Figured it *should* work with the wire format staying the same, but always good to check 😅

.golangci.yml
```diff
@@ -77,6 +78,18 @@ linters-settings:
             - log: "use the internal/log package"
             - go.uber.org/multierr: "use the internal/errors package"
             - github.com/hashicorp/go-multierror: "use the internal/errors package"
+            - github.com/gogo/protobuf/proto:
```
brendoncarroll, line 81: 👍

src/server/worker/datum/iterator.go
```diff
@@ -118,11 +118,15 @@ func iterateMeta(pachClient *client.APIClient, commit *pfs.Commit, pathRange *pf
 			if pfsserver.IsFileNotFoundErr(err) || errors.Is(err, io.EOF) {
 				return nil
 			}
-			return errors.EnsureStack(err)
+			return errors.Wrap(err, "next")
 		}
 		meta := &Meta{}
-		if err := jsonpb.Unmarshal(tr, meta); err != nil {
```
brendoncarroll, line 124: Leaving a note to come back to this.  seems a little sus, like we went from unmarshaling line by line, to reading the whole thing into memory.

src/server/pfs/server/driver_file.go
```diff
@@ -173,9 +174,9 @@ func (d *driver) copyFile(ctx context.Context, uw *fileset.UnorderedWriter, dst
 		return idx.Path == srcPath || strings.HasPrefix(idx.Path, fileset.Clean(srcPath, true))
 	})
 	fs = fileset.NewIndexMapper(fs, func(idx *index.Index) *index.Index {
-		idx2 := *idx
+		idx2 := protoutil.Clone(idx)
```
brendoncarroll, line 177: Good catch 💯

src/internal/storage/chunk/transform.go
```diff
@@ -191,7 +192,7 @@ func verifyData(id ID, x []byte) error {
 
 // Key returns a unique key for the Ref suitable for use in hash tables
 func (r *Ref) Key() pachhash.Output {
-	data, err := r.Marshal()
+	data, err := proto.MarshalOptions{Deterministic: true}.Marshal(r)
```
brendoncarroll, line 195: 👍

src/internal/storage/fileset/postgres_cache_test.go
```diff
@@ -37,8 +37,8 @@ func TestPostgresCache(t *testing.T) {
 		valueProto := &TestCacheValue{FileSetId: id.HexString()}
 		data, err := proto.Marshal(valueProto)
 		require.NoError(t, err)
-		value := &types.Any{
-			TypeUrl: "/" + proto.MessageName(valueProto),
+		value := &anypb.Any{
+			TypeUrl: "/" + string(proto.MessageName(valueProto)),
```
brendoncarroll, line 41: Just curious if this was auto-inserted by any tooling?  Or if proto.MessageName actually returns a type distinct from `string`

jrockway: Yeah that's weird.  It returns a string.

src/server/worker/datum/iterator.go
brendoncarroll: Circling back:  The outer loop is a tar stream, and you're only reading a file at a time.  We should really have a max on this but, I'm sure we are using `io.ReadAll` elsewhere.

src/internal/storage/fileset/postgres_cache_test.go
jrockway: Nevermind.  In gogo it was a string, in google it is a protoreflect.FullName, which is `type FullName string`, not `type FullName = string`.

brendoncarroll (approved):
PFS and storage changes look good.

src/server/worker/datum/iterator.go
jrockway: `jsonpb.Unmarshal(io.Reader, ...)` is a weird function; it basically uses a json.Decoder to read the first message into memory, and then it unmarshals that, and then returns.  I'm not sure those were the semantics we wanted here, so I used `protojson.Unmarshal` instead of `protoutil.NewProtoJSONDecoder().UnmarshalNext()`.  Maybe that's a mistake?

PR #8923, opened 2023-06-08
[CORE-1755] Implement `pachctl validate pipeline`

robert-uhl:
This provides offline-only pipeline spec validation.

### DIFF
--- a/src/internal/ppsutil/decoder.go
+++ b/src/internal/ppsutil/decoder.go
@@ -1,7 +1,6 @@
 package ppsutil
 
 import (
-	"bytes"
 	"io"
 
 	"gopkg.in/yaml.v3"
@@ -20,10 +19,11 @@ type PipelineManifestReader struct {
 	next    func() (*ppsclient.CreatePipelineRequest, error)
 }
 
-// NewPipelineManifestReader creates a new manifest reader from a path.
-func NewPipelineManifestReader(pipelineBytes []byte) (result *PipelineManifestReader, retErr error) {
+// NewPipelineManifestReader creates a new manifest reader which reads manifests
+// from an io.Reader.
+func NewPipelineManifestReader(r io.Reader) (result *PipelineManifestReader, retErr error) {
 	return &PipelineManifestReader{
-		decoder: yaml.NewDecoder(bytes.NewReader(pipelineBytes)),
+		decoder: yaml.NewDecoder(r),
 	}, nil
 }
 
--- a/src/internal/ppsutil/decoder.go
+++ b/src/internal/ppsutil/decoder.go
@@ -1,7 +1,10 @@
 package ppsutil
 
 import (
+	"fmt"
 	"io"
+	"os"
+	"strings"
 
 	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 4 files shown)


robert-uhl:
Of note, this does *not* validate the objects which would be created by PPS in K8s as a result of the pipeline spec; that is left as future work.

src/internal/ppsutil/decoder.go
```diff
@@ -16,23 +18,78 @@ import (
...
+// TODO(INT-1006): this exists only because the implementation of the /_mount_datums
+// endpoint in the FUSE server parses its PUT body as a full pipeline spec.
+func (r *PipelineManifestReader) DisableValidation() *PipelineManifestReader {
+	r.noValidate = true
+	return r
+}
+
+type unvalidatedCreatePipelineRequest ppsclient.CreatePipelineRequest
+
+// Reset implements proto.Message.
```
zmajeed, line 45: typo

src/internal/ppsutil/decoder.go
```diff
@@ -16,23 +18,78 @@ import (
...
+func (ucpr *unvalidatedCreatePipelineRequest) Reset() {
+	(*ppsclient.CreatePipelineRequest)(ucpr).Reset()
+}
+
+// String implements proto.Message.
```
zmajeed, line 50: typo

src/server/pps/cmds/cmds.go
```diff
@@ -1442,53 +1442,72 @@ All jobs created by a pipeline will create commits in the pipeline's output repo
...
-// there.
+// fileIndicatorToReader returns an IO reader for a file based on an indicator
+// (which may be a local path, a remote URL or "-" for stdin).
 //
 // TODO(msteffen) This is very similar to readConfigBytes in
 // s/s/identity/cmds/cmds.go (which differs only in not supporting URLs),
 // so the two could perhaps be refactored.
-func readPipelineBytes(pipelinePath string) (pipelineBytes []byte, retErr error) {
-	if pipelinePath == "-" {
+func fileIndicatorToReader(indicator string) (io.Reader, error) {
```
zmajeed, line 1480: could this be called `filePathToReader`?

robert-uhl: I think that would be incorrect, because it accepts more than just paths: literal `"-"` is interpreted as STDIN; a string which parses as a URL is interpreted as a URL; only all other strings are interpreted as paths.

src/internal/ppsutil/decoder.go
```diff
@@ -16,23 +18,78 @@ import (
...
+// DisableValidation disables pipeline validation.
+//
+// TODO(INT-1006): this exists only because the implementation of the /_mount_datums
+// endpoint in the FUSE server parses its PUT body as a full pipeline spec.
+func (r *PipelineManifestReader) DisableValidation() *PipelineManifestReader {
+	r.noValidate = true
+	return r
+}
+
+type unvalidatedCreatePipelineRequest ppsclient.CreatePipelineRequest
```
msteffen, line 43: I haven't tried it, but you might be able to get rid of `Reset()`, `String()`, and `ProtoMessage()` below by declaring the type like:
```
type unvalidatedCreatePipelineRequest struct {
  *ppsclient.CreatePipelineRequest
}
```
(and then use raw `unvalidatedCreatePipelineRequest` throughout to avoid double-dereferencing).

I think the in-memory representation is the same, but it does look a little weirder; the wrapper struct just exists to pass the methods through. Your call if it seems better or worse; just wanted to suggest it.

msteffen (approved):
LGTM, with one style suggestion

src/internal/ppsutil/decoder.go
robert-uhl: Unfortunately, that appears to interact badly with `serde.RoundTrip`; preventing it from realizing that it’s a `proto.Message` when unmarshalling. I _could_ unmarshal the data as a `ppsclient.CreatePipelineRequest` and wrap it if it needs to be validated, or just get rid of `unvalidatedCreatePipelineRequest`.

This was an attempt to use type safety to prevent unvalidated data escaping the decoder, but perhaps that is a bit too “clever”?

src/server/pps/cmds/cmds.go
zmajeed: Not a big deal but "-" is just a conventional "path" standin for terminal I/O offered by commandline tools - does it really need to be supported in a path parameter in an internal API?

Could something like "file:///dev/stdin" or "file:///dev/fd/0" be more clear and consistent?

src/internal/ppsutil/decoder.go
msteffen: Ah, I mean framed that way, it does make sense. On reflection, it also makes sense (as a Go feature) that you wouldn’t necessarily want to pass the wrapped type’s methods through if the wrapper type carries safety properties. LGTM

src/server/pps/cmds/cmds.go
robert-uhl: It’s not supported in the internal API, just in the `pachctl` command-line tool. Our code organization is misleading: it looks like this is server code (because it’s under `src/server`), but really it is client code (because it’s under `src/server/pps/cmd`: yes, the client code for each subsystem is located beneath the server directory). IMHO, at some point we should extract all the stuff in `src/server/.../cmds/` into something under `src/cmd/...`, but that’s another story☺

PR #8928, opened 2023-06-08
storage: Adds Server type and constructor

brendoncarroll:
This is to make it easier to debug small configuration errors with the storage system.  I pulled it out of another PR I'm working on.

Needs #8929


src/server/pfs/server/driver.go
```diff
@@ -82,14 +79,13 @@ type driver struct {
 	branches col.PostgresCollection
 	projects col.PostgresCollection
 
-	storage     *fileset.Storage
+	storage     storage.Server
```
brycemcanally, line 82: this could just be a pointer, which is what the `New` function returns.

brycemcanally (approved):
lgtm, just some minor nits.

PR #8929, opened 2023-06-08
Adds Package pachconfig

brendoncarroll:
The package dockertestenv should be very close to the bottom of the dependency graph, and available for import in basically every package. However, it's not, because it generates ConfigOptions for the databases it manages. Those ConfigOptions live in serviceenv, which creates a cycle. By moving the configuration portion of serviceenv out into pachconfig, we remove it as a dependency of dockertestenv.


jrockway (approved):
Good idea.

FahadBSyed (approved):

PR #8933, opened 2023-06-12
Determined pipeline users

acohen4:


etc/helm/pachyderm/values.yaml
```diff
@@ -712,6 +712,12 @@ pachd:
   defaultSidecarCPURequest: ""
   defaultSidecarMemoryRequest: ""
   defaultSidecarStorageRequest: ""
+  determined:
+    # a determined's cluster API URL
+    apiEndpoint: ""
+    # the name of the kubernetes secret containing the credentials for the determined user representing pachyderm
```
jrockway, line 718: This might be a good place to document which keys this secret needs to have.

src/pps/pps.proto
```diff
@@ -56,6 +56,11 @@ message Egress {
   }
 }
 
+message Determined {
+  repeated string workspaces = 1;
+  string password = 2;
```
jrockway, line 61: Annotate with `[(log.mask) = true]`

src/server/pps/server/api_server.go
```diff
@@ -2380,12 +2383,19 @@ func (a *apiServer) CreatePipelineInTransaction(txnCtx *txncontext.TransactionCo
...
 		newPipelineInfo.AuthToken = token
-
 		return nil
 	}(); err != nil {
 		return err
 	}
-
+	// TODO:
+	// - should this be done before the start of the txn?
+	// - should this be done in a retry?
```
jrockway, line 2392: Yes to both.  What we should really do is make txnCtx some interface that only has `Value(key any) any`, and make the logger stuff take that interface instead of context.Context.  That way we can log in transactions, but not do I/O.

src/server/pps/server/determined.go
```diff
@@ -0,0 +1,320 @@
+package server
```
jrockway, line 1: I think there probably wants to be a src/internal/determined package; that way, you don't have to include "det" or "determined" in every struct and method name.

src/server/pps/server/determined.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+		return nil, errors.Errorf("requested workspaces not found: %v", errWorkspaces)
+	}
+	return res, nil
+}
+
+func (a *apiServer) mintDeterminedToken(ctx context.Context) (string, error) {
+	u, p, err := a.determinedCredentials(ctx)
+	if err != nil {
+		return "", err
```
jrockway, line 166: Wrap this one so the flow is clear.

src/server/pps/server/determined.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+	}
+	req, err := http.NewRequestWithContext(ctx, method, a.env.Config.DeterminedURL+endpoint, b)
+	if err != nil {
+		return nil, errors.Wrap(err, "create determined http request")
+	}
+	req.Header.Add("Accept", "application/json")
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("Authorization", "Bearer "+token)
+	client := &http.Client{Timeout: 10 * time.Second}
```
jrockway, line 302: I think client.Timeout is deprecated in favor of setting a timeout on the request context above.  That would be preferred.

Also set the Transport field to `promtuil.InstrumentRoundTripper("determined", http.DefaultTransport)` for monitoring these calls.

src/server/pps/server/determined.go
```diff
@@ -0,0 +1,320 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, errors.Wrap(err, "execute determined http request")
+	}
+	if resp.StatusCode >= 200 && resp.StatusCode <= 299 {
+		defer resp.Body.Close()
```
jrockway, line 308: I think the body always needs to be closed, even if it's a 500 internal server error or something.  Move this out of the if condition and always do it after Do returns a nil error.

src/server/pps/server/determined.go
```diff
@@ -0,0 +1,320 @@
...
+		data, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, errors.Wrap(err, "read determined http response body")
+		}
+		return data, nil
+	}
+	return nil, errors.New(resp.Status)
```
jrockway, line 315: And yeah, I'd at least log the body for errors.  There might be a function that logs httputil.DumpResponse  floating around somewhere. 

The error message is probably useful while debugging the integration.  If you get a non-success status code, you probably want to see what message it returns.  It's somewhat maddening when you can't see what the server is telling you.

src/server/pps/server/determined.go
jrockway: With all of that in mind, they appear to be using gRPC-Gateway to serve this REST API.  That means we can probably use GRPC instead of HTTP, which would be easier to deal with when things go wrong.  At the very least, gRPC-Gateway generates HTTP client stubs that we should use instead of a hand-written API.  (Now that you've written it... it doesn't matter, but if they change stuff, it might save us pain.)

Ref: https://docs.determined.ai/latest/reference/rest-api.html

src/server/pps/server/determined.go
```diff
@@ -0,0 +1,320 @@
...
+	var respObj detLoginResponse
+	if err := json.Unmarshal(data, &respObj); err != nil {
+		return "", errors.Wrap(err, "unmarshal determined token")
+	}
+	return respObj.Token, nil
+}
+
+func (a *apiServer) determinedCredentials(ctx context.Context) (string, string, error) {
+	secretName := a.env.Config.DeterminedCredentialsSecret
+	secret, err := a.env.KubeClient.CoreV1().Secrets(a.namespace).Get(ctx, secretName, metav1.GetOptions{})
```
jrockway, line 185: I think we should avoid the k8s API for this, and inject this secret as either environment variables or as a file (if we want changes to take effect at runtime without having to restart pachd).

src/internal/pachconfig/config.go
```diff
@@ -139,6 +139,9 @@ type PachdSpecificConfiguration struct {
 	// the proxy, and ProxyTLS for whether or not to use https:// for generated URLs.
 	ProxyHost string `env:"PACHYDERM_PUBLIC_HOST,default="`
 	ProxyTLS  bool   `env:"PACHYDERM_PUBLIC_TLS,default=false"`
+	// Determined integration configuration
```
jrockway, line 142: I talk about this below, but why not inject the fields from the secret with secretKeyRef in the manifest.  That way we don't need to talk to k8s to get the username/password.

etc/helm/pachyderm/values.yaml
acohen4: 1000%

src/server/pps/server/determined.go
acohen4: done 👍

src/server/pps/server/determined.go
```diff
@@ -0,0 +1,213 @@
...
+	return 0, errors.Errorf("workspaceEditor role not found")
+}
+
+func resolveDeterminedWorkspaces(ctx context.Context, dc det.DeterminedClient, workspaces []string) ([]*workspacev1.Workspace, error) {
+	workspacesResp, err := dc.GetWorkspaces(ctx, &det.GetWorkspacesRequest{})
+	if err != nil {
+		return nil, errors.Wrap(err, "list determined workspaces")
+	}
+	workspaceSet := make(map[string]struct{})
+	for _, w := range workspacesResp.Workspaces {
```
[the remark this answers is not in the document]
acohen4: 100%! good catch

jrockway (approved):
We discussed in person!

PR #8943, opened 2023-06-14
[PFS-102] New cron trigger functionality

brycemcanally:
This PR implements the cron trigger functionality described here: https://www.notion.so/2023-06-06-Cron-Triggers-1e566caf0b61425897ca54cb4e66e4eb


src/server/pfs/server/trigger.go
```diff
@@ -59,7 +60,7 @@ func (d *driver) triggerCommit(
 // isTriggered checks to see if a branch should be updated from oldHead to
 // newHead based on a trigger.
 func (d *driver) isTriggered(txnCtx *txncontext.TransactionContext, t *pfs.Trigger, oldHead, newHead *pfs.CommitInfo) (bool, error) {
-	if t == nil {
+	if t == nil || t.CronSpec != "" {
```
acohen4, line 63: the other trigger condition types can be used non-exclusively right? Just calling out that this diverges from the `All` behavior

src/server/pfs/server/master.go
```diff
@@ -85,7 +86,7 @@ func (d *driver) master(ctx context.Context) {
 	})
 }
 
-func (d *driver) finishCommits(ctx context.Context) (retErr error) {
+func (d *driver) manageRepos(ctx context.Context) error {
```
brendoncarroll, line 89: 👍

src/server/pfs/server/trigger.go
```diff
@@ -135,19 +136,22 @@ func (d *driver) validateTrigger(txnCtx *txncontext.TransactionContext, branch *
...
 		return errors.Errorf("can't trigger on a negative number of commits")
 	}
+	if _, err := cronutil.ParseCronExpression(trigger.CronSpec); trigger.CronSpec != "" && err != nil {
+		return errors.Wrapf(err, "invalid trigger cron spec")
+	}
 
 	biMaps := make(map[string]*pfs.BranchInfo)
 	if err := d.listBranchInTransaction(txnCtx, branch.Repo, false, func(bi *pfs.BranchInfo) error {
-		biMaps[bi.Branch.Name] = bi
+		biMaps[bi.Branch.Name] = proto.Clone(bi).(*pfs.BranchInfo)
```
brendoncarroll, line 154: 👍

src/server/pfs/server/master.go
```diff
@@ -139,6 +157,119 @@ func (d *driver) finishCommits(ctx context.Context) (retErr error) {
...
+				log.Error(ctx, "error running cron trigger", zap.String("branch", key), zap.Error(err), zap.Duration("retryAfter", d))
+				return nil
+			})
+		}()
+		return nil
+	})
+}
+
+func (d *driver) runCronTrigger(ctx context.Context, branch *pfs.Branch) error {
+	branchInfo, err := d.inspectBranch(ctx, branch)
```
brendoncarroll, line 218: Should we be concerned about doing a read here?  We actually have access to the branch info in the caller, I think we should just pass that in directly.

src/server/pfs/server/master.go
```diff
@@ -139,6 +157,119 @@ func (d *driver) finishCommits(ctx context.Context) (retErr error) {
...
+		}
+		return err
+	}
+	schedule, err := cronutil.ParseCronExpression(branchInfo.Trigger.CronSpec)
+	if err != nil {
+		return err
+	}
+	// Use the current head commit start time as the previous tick.
+	// This prevents the timer from restarting if the master restarts.
+	ci, err := d.inspectCommit(ctx, branchInfo.Head, pfs.CommitState_STARTED)
```
brendoncarroll, line 232: What happens on an empty branch?

brendoncarroll: Also should we take the minimum of this and the current time?

brendoncarroll (changes requested):
A few questions.

src/server/pfs/server/master.go
```diff
@@ -139,6 +157,119 @@ func (d *driver) finishCommits(ctx context.Context) (retErr error) {
...
+				ct.cancel()
+			}
+			delete(cronTriggers, key)
+			return nil
+		}
+		branchInfo, err := d.inspectBranch(ctx, pfsdb.ParseBranch(key))
+		if err != nil {
+			return err
+		}
+		if branchInfo.Trigger == nil || branchInfo.Trigger.CronSpec == "" {
```
acohen4, line 189: shouldn't we also `delete(cronTriggers, key)` here? If a user updates their trigger from `!= nil -> == nil` we would be keeping the goro alive

src/server/pfs/server/master.go
acohen4: good point on the empty branch edge case - the semantics of triggering a branch -> nil doesn't make sense to me. We probably should exit if `branchInfo.Head == nil`

src/server/pfs/server/trigger.go
brycemcanally: Yeah, that seems implicit in the description of the new field in the proto, but I can make it more explicit.

src/server/pfs/server/master.go
```diff
@@ -118,17 +119,34 @@ func (d *driver) finishCommits(ctx context.Context) (retErr error) {
...
 						if err != nil {
 							return errors.Wrap(err, "error locking repo lock")
 						}
 						defer errors.Invoke1(&retErr, ring.Unlock, lockPrefix, "unlocking repo lock")
-						return d.finishRepoCommits(lockCtx, key)
+						var eg errgroup.Group
+						eg.Go(func() error {
+							return backoff.RetryUntilCancel(ctx, func() error {
+								return d.manageBranches(lockCtx, key)
+							}, backoff.NewInfiniteBackOff(), func(err error, d time.Duration) error {
```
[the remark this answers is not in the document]
brycemcanally: Yeah, but the max interval is 15s. When you say hard max, do you mean for each interval or the total time spent retrying?

src/server/pfs/server/master.go
brycemcanally: The branch can be deleted at any time. If it is deleted while this method is running, the method will return an error and be retried by the caller. If we don't reload the branch info after a retry, we will get stuck in a retry loop until the delete event is received. When we detect that the branch doesn't exist, we just log it and return nil since there is nothing to do.

src/server/pfs/server/master.go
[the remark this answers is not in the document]
brendoncarroll: For a given interval.

src/server/pfs/server/master.go
brendoncarroll: Should we check for the not exist error in the Retry's error handler?  That way we catch it where ever it occurs, instead of just at the start of this function?

src/server/pfs/server/master.go
brycemcanally: Branch heads cannot be nil since they always start with an empty commit as head. I'm not sure I understand the minimum suggestion. Is the concern clock drift? If not, then the commit start time will always be less than the current time.

src/server/pfs/server/master.go
```diff
@@ -139,6 +157,119 @@ func (d *driver) finishCommits(ctx context.Context) (retErr error) {
...
+	if err != nil {
+		return errors.EnsureStack(err)
+	}
+	for {
+		next := schedule.Next(prev)
+		if next.IsZero() {
+			log.Debug(ctx, "no more scheduled ticks; exiting loop")
+			return nil
+		}
+		log.Info(ctx, "waiting for next cron tick", zap.Time("next", next), zap.Time("prev", prev))
```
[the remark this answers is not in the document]
brycemcanally: sure, but I would rather just be consistent with the cron pipeline functionality for now, which uses the same pattern.

src/server/pfs/server/master.go
```diff
@@ -139,6 +157,119 @@ func (d *driver) finishCommits(ctx context.Context) (retErr error) {
...
+			cancel: cancel,
+			spec:   branchInfo.Trigger.CronSpec,
+		}
+		go func() {
+			backoff.RetryUntilCancel(ctx, func() error { //nolint:errcheck
+				return d.runCronTrigger(ctx, branchInfo.Branch)
+			}, backoff.NewInfiniteBackOff(), func(err error, d time.Duration) error {
+				log.Error(ctx, "error running cron trigger", zap.String("branch", key), zap.Error(err), zap.Duration("retryAfter", d))
+				return nil
+			})
```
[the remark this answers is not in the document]
brycemcanally: No, the cron triggers map should only be managed by one goroutine. This goroutine should just retry until its context is canceled or the branch is deleted.

src/server/pfs/server/master.go
brycemcanally: ahh, good catch. Yes, I will change that.

src/server/pfs/server/master.go
brycemcanally: I considered putting it there, but I leaned towards the `runCronTrigger` method being responsible for loading the current branch info. It also seemed kind of nice for when it is eventually unit testable. But, I am reconsidering just putting it in the notify callback. That is the pattern we use in the worker registry and it may be better for `runCronTrigger` to operate on a stable branch info and always return an error if the branch is deleted.

acohen4 (approved):
LGTM

brendoncarroll (approved):

PR #8944, opened 2023-06-15
Pfs 91 api

FahadBSyed:


src/internal/coredb/projects.go
```diff
@@ -87,8 +106,15 @@ func getProject(ctx context.Context, queryExecer QueryExecer, where string, wher
 	return project, nil
 }
 
-func ListProject(ctx context.Context, db *pachsql.DB) (*ProjectIterator, error) {
-	rows, err := db.QueryxContext(ctx, "SELECT name, description, created_at FROM core.projects")
+func ListProject(ctx context.Context, db *pachsql.DB, option ...ListProjectOption) (*ProjectIterator, error) {
```
albscui, line 109: I don't think it make sense to combine pagination and iterators. If you don't have pagination, then the results could be huge, so iterators make sense, but with pagination, I think it's better to just return the results as a slice. Basically, I think `ProjectIterator` should be able to "transparently" iterate through all projects (with or without pagination).

So I propose we have two ways to iterate:
- `ListAllProjects() (*ProjectIterator, error)`, behind the scenes, this could use the paged ListProject, which means for now, you don't have to implement this.
- `ListProject() ([]*ProjectInfo, error)`

src/internal/coredb/projects.go
```diff
@@ -3,19 +3,29 @@ package coredb
...
-	"io"
-	"time"
 )
 
 type ProjectIterator struct {
 	*sqlx.Rows
 }
 
+// ListProjectOption contains optional parameters that support pagination.
+// PageSize and PageNum are converted as-is into an SQL LIMIT and OFFSET.
```
albscui, line 22: Hmm I interpreted "as-is" as directly converting PageNumber to OFFSET and PageSize to LIMIT, where  in fact `OFFSET = PageNum * PageSize`

src/server/pfs/server/testing/server_test.go
```diff
@@ -373,11 +374,13 @@ func TestPFS(suite *testing.T) {
 			{tu.UniqueString("project."), badFormatErr},
 		}
 		for _, tc := range testCases {
+			fmt.Println(tc.projectName)
```
albscui, line 377: remove

src/server/pfs/server/testing/server_test.go
```diff
@@ -341,6 +341,7 @@ func TestPFS(suite *testing.T) {
 
 		c := env.PachClient
 		badFormatErr := "only alphanumeric characters"
+		uniqueConstraintErr := "duplicate key value violates unique constraint \"projects_name_key\" (SQLSTATE 23505)"
```
albscui, line 344: `uniqueConstraintErr` is only used once, so no need to instantiate a variable here.

src/server/pfs/fuse/server_test.go
```diff
@@ -1026,8 +1026,9 @@ func TestProjects(t *testing.T) {
 
 		require.NoError(t, json.NewDecoder(resp.Body).Decode(&projectData))
 		require.Equal(t, len(projectData), 3)
-		require.Equal(t, projectData[0].Project.Name, emptyProjectName)
+		// todo(fahad): I had to reverse the order of these, is that okay?
```
albscui, line 1029: I think this is expected, because our new projects table automatically sorts the projects in the order they are created.

albscui (changes requested):

src/internal/coredb/projects.go
FahadBSyed: Gotcha, I think maybe its better to just stick with the iterator then since that's already implemented. I can supply the other paginated approach for transactions

src/server/pfs/server/testing/server_test.go
```diff
@@ -377,7 +377,8 @@ func TestPFS(suite *testing.T) {
 				err := c.CreateProject(tc.projectName)
 				if tc.errMatch != "" {
 					require.YesError(t, err)
-					require.Matches(t, tc.errMatch, err.Error())
+
```
albscui, line 380: nit: no need for empty newline

albscui (approved):

src/server/pfs/fuse/server_test.go
```diff
@@ -1026,8 +1026,8 @@ func TestProjects(t *testing.T) {
 
 		require.NoError(t, json.NewDecoder(resp.Body).Decode(&projectData))
 		require.Equal(t, len(projectData), 3)
-		require.Equal(t, projectData[0].Project.Name, emptyProjectName)
+		require.Equal(t, projectData[0].Project.Name, pfs.DefaultProjectName)
```
albscui, line 1029: I just realized the order is not deterministic here because your `SELECT name, description, created_at FROM core.projects` doesn't order by anything. I think you should expose the ordering to the user, or by default order by `id`.

src/internal/coredb/projects.go
```diff
@@ -3,13 +3,15 @@ package coredb
...
 	"github.com/jmoiron/sqlx"
+
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
-	"io"
-	"time"
 )
 
 type ProjectIterator struct {
```
brendoncarroll, line 17: This kind of iterator will block the connection while it iterates.  I can't see everywhere this is used from the current diff, but this isn't the kind of iterator we would want to use for streaming out to an RPC.  For that it should have a `*sqlx.DB` internally, and buffer a page of results at a time, making queries as needed.

Additionally, if you implement `stream.Iterator` (Next takes a context) then you can use stream.ForEach and stream.Collect whereever this is used.  This isn't in the diff, so not a blocker for this PR.

src/internal/coredb/projects.go
```diff
@@ -98,6 +109,9 @@ func ListProject(ctx context.Context, db *pachsql.DB) (*ProjectIterator, error)
 	return iter, nil
 }
 
+//func ListProjectInTransaction(ctx context.Context, tx *pachsql.Tx, option ...ListProjectOption) ([]*pfs.ProjectInfo, error) {
```
brendoncarroll, line 112: Remove this or implement.

brendoncarroll: I think we have the convention of suffixing things like this with `Tx`.  And while it's technically more expressive to take both a context, and a `*pachsql.Tx`, since they do different things.  We have tried to pick on or the other.  The hope is that by omitting contexts in transaction methods, we will discourage long running operations in a transaction.

src/server/pfs/server/driver.go
```diff
@@ -5,14 +5,15 @@ import (
 	"crypto/rand"
 	"database/sql"
 	"fmt"
-	"github.com/pachyderm/pachyderm/v2/src/internal/coredb"
```
brendoncarroll, line 8: lol why was that up here.

brendoncarroll:
I just realized this isn't against master.  You should ignore the code owner rules for merging into your staging branch.  If you can't force merge, let me know.  I'll try to figure out how to turn off branch protections for this branch.

PR #8950, opened 2023-06-15
update python-sdk debug test to use dump v2

kevyang:
this test was previously broken because it used the dump command, which is now deprecated in favor of v2. this change updates the test to use the dump v2 command instead.

Jira: [INT-1009]

[INT-1009]: https://pachyderm.atlassian.net/browse/INT-1009


bbonenfant (approved):
These changes look fine to me. There's an opportunity here to replace `dump` with `dump_v2` but I don't think it's used enough to warrant that sort of usability change.

PR #8951, opened 2023-06-16
Update List To Buffer A Page of Results

FahadBSyed:


src/internal/coredb/projects.go
```diff
@@ -14,35 +14,81 @@ import (
...
+	}
+	iter := &ProjectIterator{
+		page: page,
+		db:   db,
+	}
+	return iter, nil
 }
 
+func listProjectPage(ctx context.Context, db *pachsql.DB, pageNum int) ([]projectRow, error) {
+	rows, err := db.QueryxContext(ctx, "SELECT name,description,created_at FROM core.projects ORDER BY id ASC LIMIT $1 OFFSET $2", pageSize, pageSize*pageNum)
```
albscui, line 73: You can replace this with a simple `Select()`
```go
var page []projectRow
err := db.SelectContext(ctx, &page, "SELECT name,description,created_at FROM core.projects ORDER BY id ASC LIMIT $1 OFFSET $2", pageSize, pageSize*pageNum)
return page
```

src/internal/coredb/projects.go
```diff
@@ -14,35 +14,81 @@ import (
...
+		return nil, errors.Wrap(err, "failed to list projects")
+	}
+	iter := &ProjectIterator{
+		page: page,
+		db:   db,
+	}
+	return iter, nil
 }
 
+func listProjectPage(ctx context.Context, db *pachsql.DB, pageNum int) ([]projectRow, error) {
```
albscui, line 72:
```suggestion
func listProjectPage(ctx context.Context, db *pachsql.DB, pageSize int, pageNum int) ([]projectRow, error) {
```

src/internal/coredb/projects.go
```diff
@@ -14,35 +14,81 @@ import (
...
 
+const (
+	pageSize = 100
+)
+
+// ProjectIterator batches a page of projectRow entries. Entries can be retrieved using iter.Next().
 type ProjectIterator struct {
-	*sqlx.Rows
+	pageNum int
+	page    []projectRow
```
albscui, line 24:
```suggestion
	projects []projectRow
```

src/internal/coredb/projects.go
```diff
@@ -14,35 +14,81 @@ import (
...
 )
 
+const (
+	pageSize = 100
+)
+
+// ProjectIterator batches a page of projectRow entries. Entries can be retrieved using iter.Next().
 type ProjectIterator struct {
-	*sqlx.Rows
+	pageNum int
```
albscui, line 23: We probably want to guard this with a mutex. Imagine multiple goroutines iterating over this. However, maybe we should discourage that, but there's nothing we can do in code to enforce that without a lock.

src/internal/coredb/projects.go
```diff
@@ -14,35 +14,81 @@ import (
...
 	}
-	if iter.Rows.Err() != nil {
-		return nil, errors.Wrap(iter.Rows.Err(), "failed iterating")
+	row := iter.page[iter.index]
+	projectTimestamp, err := types.TimestampProto(row.CreatedAt)
+	if err != nil {
+		return errors.Wrap(err, "failed to convert time.Time to proto timestamp")
 	}
-	return nil, io.EOF
+	*project = pfs.ProjectInfo{
```
albscui, line 51: Potential `nil pointer dereference` bug?

src/internal/coredb/projects.go
```diff
@@ -14,35 +14,81 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
 )
 
+const (
+	pageSize = 100
```
albscui, line 18: Why make this a constant? We know pageSize can change.

src/server/pfs/server/driver.go
```diff
@@ -807,25 +807,25 @@ func (d *driver) listProject(ctx context.Context, cb func(*pfs.ProjectInfo) erro
 	if err != nil {
 		return errors.Wrap(err, "could not list projects")
 	}
-	defer projIter.Close()
-	for projectInfo, err := projIter.Next(); !errors.Is(err, io.EOF); projectInfo, err = projIter.Next() {
+	projInfo := &pfs.ProjectInfo{}
+	for err := projIter.Next(ctx, projInfo); !errors.Is(err, io.EOF); err = projIter.Next(ctx, projInfo) {
```
albscui, line 811: Is GitHub displaying this correctly?? I'm seeing two calls to `.Next()`
```Go
for err := projIter.Next(ctx, projInfo); !errors.Is(err, io.EOF); err = projIter.Next(ctx, projInfo)
```

src/internal/coredb/projects.go
```diff
@@ -14,35 +14,81 @@ import (
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list projects")
+	}
+	defer rows.Close()
+	page := make([]projectRow, 0)
+	if err := sqlx.StructScan(rows, &page); err != nil {
+		return nil, errors.Wrap(err, "failed to scan row results for projects")
+	}
+	if len(page) == 0 {
+		return nil, io.EOF
```
albscui, line 83: I don't think it should be the responsibility of `listProjectPage` to return `io.EOF`. An empty page is not an error. I think the consumer of `listProjectPage`, i.e. `ListProject` should return `io.EOF`.

src/internal/coredb/projects.go
FahadBSyed: I made it a constant since I'm not sure if we are going to expose this to the user, but I suppose it doesn't have to be. I think for ListProjectTx we'll end up just exposing the pages

src/server/pfs/server/driver.go
FahadBSyed: Yes, so the first call gets the first entry and the second call advances the iterator similar to
go ```
for int i := 0; i < 100; i++ {
}
````

Comment (albscui) [reply src/server/pfs/server/driver.go:811]: I find this pattern really weird... doesn't https://github.com/pachyderm/pachyderm/pull/8951/files#diff-fe7ebf0ab8e4af97b3c44edd1b2666f1420b0e08053d1042caf22f1af212dec1R60 automatically advance the iterator?

File: src/internal/coredb/projects.go [hunk 9]
```diff
@@ -14,33 +15,70 @@ import (
...
+	iter := &ProjectIterator{
+		projects: page,
+		db:       db,
+	}
+	return iter, nil
+}
+
+func listProjectPage(ctx context.Context, db *pachsql.DB, pageSize, pageNum int) ([]projectRow, error) {
+	var page []projectRow
+	err := db.SelectContext(ctx, &page, "SELECT name,description,created_at FROM core.projects ORDER BY id ASC LIMIT $1 OFFSET $2", pageSize, pageSize*pageNum)
```
albscui, line 80: nit: I wonder if in the event of an error, `page` would be partially filled? In which case, it would be safer to do something like

```Go
	if err := db.SelectContext(ctx, &page, "SELECT name,description,created_at FROM core.projects ORDER BY id ASC LIMIT $1 OFFSET $2", pageSize, pageSize*pageNum); err != nil {
	     return nil, errors.Wrap(err, "could not to get project page")
	}
	return page, nil
```

src/server/pfs/server/driver.go
albscui: can't you do the same with something like
```Go
for !errors.Is(projectIter.Next(ctx, projInfo), io.EOF) {}
```

src/internal/coredb/projects.go
```diff
@@ -14,33 +15,70 @@ import (
...
+	CreatedAt   time.Time `db:"created_at"`
+}
+
+// Next advances the iterator by one row. It returns an io.EOF when there are no more entries.
+func (iter *ProjectIterator) Next(ctx context.Context, project *pfs.ProjectInfo) error {
+	if project == nil {
+		return errors.Wrap(fmt.Errorf("project is nil"), "failed to get next project")
+	}
+	var err error
+	iter.lock.Lock()
```
albscui, line 39: maybe also add a `defer iter.lock.Unlock()` just in case?

src/internal/coredb/projects.go
```diff
@@ -73,6 +111,11 @@ func DeleteProject(ctx context.Context, queryExecer QueryExecer, projectName str
 	return nil
 }
 
+func DeleteAllProjects(ctx context.Context, queryExecer QueryExecer) error {
+	_, err := queryExecer.ExecContext(ctx, "TRUNCATE core.projects CASCADE;")
```
albscui, line 115: I think we should be very careful with CASCADE. In fact, I don't think we want to implement that right now. I think it's safer to delete everything in order in Go code, instead of trusting SQL cascade deletes.

src/server/pfs/server/driver.go
FahadBSyed: Maybe! I can try that out. I agree this `for` pattern is awkward. Its more of a holdover from when the project was returned by Next()

src/internal/coredb/projects.go
FahadBSyed: It seems like there is some disagreement but I think I would like to do as much of it in postgres as possible only if doing it in postgres makes the code a lot simpler.

FahadBSyed: For now I can omit cascade and we can come back to it later.

src/internal/coredb/projects_test.go
```diff
@@ -81,27 +81,45 @@ func TestListProject(t *testing.T) {
...
 		expectedInfos[i] = createInfo
 		require.NoError(t, CreateProject(ctx, db, createInfo), "should be able to create project")
 	}
 	iter, err := ListProject(ctx, db)
 	require.NoError(t, err, "should be able to list projects")
-	defer iter.Close()
 	i := 0
-	for proj, err := iter.Next(); !errors.Is(err, io.EOF); proj, err = iter.Next() {
+	projectInfo := &pfs.ProjectInfo{}
+	for err := iter.Next(ctx, projectInfo); !errors.Is(err, io.EOF); err = iter.Next(ctx, projectInfo) {
```
albscui, line 95: Nit: I know this is just test code, but lets be consistent
```suggestion
	for !errors.Is(iter.Next(ctx, projectInfo), err, io.EOF)  {
```

albscui (approved):

PR #8952, opened 2023-06-16
storage: remove kv.MemCache in favor of Hashicorp's LRU

brendoncarroll:
Pulled out of #8643.  Before Go had generics we were serializing everything and storing it in caches as bytes.  We no longer need to do that.


src/internal/obj/minio_client.go
```diff
@@ -48,8 +48,9 @@ func newMinioClientV2(endpoint, bucket, id, secret string, secure bool) (*minioC
 func (c *minioClient) Put(ctx context.Context, name string, r io.Reader) (retErr error) {
 	defer func() { retErr = c.transformError(retErr, name) }()
 	opts := minio.PutObjectOptions{
-		ContentType: "application/octet-stream",
-		PartSize:    uint64(8 * 1024 * 1024),
+		ContentType:      "application/octet-stream",
+		PartSize:         uint64(8 * 1024 * 1024),
+		DisableMultipart: true,
```
brendoncarroll, line 53: Multipart uploads are generally bad for performance when dealing with our typical chunk sizes.  We use Go CDK for importing and exporting large files anyways.  But this is used in tests.

src/internal/storage/chunk/transform.go
```diff
@@ -45,11 +44,13 @@ func Create(ctx context.Context, opts CreateOptions, ptext []byte, createFunc fu
 
 // Get calls client.Get to retrieve a chunk, then verifies, decrypts, and decompresses the data.
 // cb is called with the uncompressed plaintext
```
jrockway, line 46: cb is no longer an arg

src/internal/storage/fileset/postgres_store.go
```diff
@@ -15,23 +16,26 @@ import (
...
 	db      *pachsql.DB
-	cache   kv.GetPut
+	cache   *lru.Cache[ID, *Metadata]
 	deduper *miscutil.WorkDeduper[ID]
 }
 
 // NewPostgresStore returns a Store backed by db
 // TODO: Expose configuration for cache size?
 func NewPostgresStore(db *pachsql.DB) MetadataStore {
+	mc, err := lru.New[ID, *Metadata](100)
```
jrockway, line 32: I think the configuration option StorageMemoryCacheSize needs to go away, or be read here.

brendoncarroll: I believe StorageMemoryCacheSize refers to the size of the chunk cache.  We don't expose a way to configure this.

jrockway: You're deleting some code that reads it here, though: https://github.com/pachyderm/pachyderm/pull/8952/files#diff-f20d955c17bf7f6b9947ddec0131271bce16608c1f09b2d5dfb18092a66d37daL50

jrockway: Nevermind, I see your fix now :)

jrockway (approved):
LGTM.

src/internal/storage/fileset/postgres_store.go
brendoncarroll: Yeah good catch.  I just added it back.  We configure that in `storage/config.go` now.

PR #8954, opened 2023-06-16
Add an "applies_to" field to roles, which shows which resource types a role is relevant for

jrockway:
Design: https://www.notion.so/2023-06-15-Returning-roles-in-a-user-friendly-way-2ff2947291fe4c06ac83a322ac2c123c


src/server/auth/server/roles.go
```diff
@@ -174,51 +185,57 @@ func init() {
...
 		},
 	})
 
 	// Project related roles
 	projectViewerRole := registerRole(&auth.Role{
-		Name:          auth.ProjectViewerRole,
-		ResourceTypes: []auth.ResourceType{auth.ResourceType_CLUSTER, auth.ResourceType_PROJECT},
+		Name:    auth.ProjectViewerRole,
+		Scopes:  []auth.ResourceType{auth.ResourceType_CLUSTER, auth.ResourceType_PROJECT},
+		Affects: []auth.ResourceType{auth.ResourceType_CLUSTER, auth.ResourceType_PROJECT},
```
albscui, line 200: I don't think project level roles (project viewer, writer, or owner) affects CLUSTER resource type.

albscui (changes requested):

src/auth/auth.proto
```diff
@@ -256,7 +256,11 @@ message Groups {
 message Role {
   string name = 1;
   repeated Permission permissions = 2;
-  repeated ResourceType resource_types = 3;
+  // Resources this role can be applied to (for example, you can't apply ClusterAdmin to a repo.)
```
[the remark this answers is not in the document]
jrockway: I think "applies_to" is the field name we're both looking for.  What do you think?

src/server/auth/server/roles.go
jrockway: Good point.

albscui (commented):
Overall, I think this design makes sense. My issue is that I fear we are still not 100% concrete with the words we use: `applies to`, `affects`, `scopes`, `relevant`.

Essentially, we want the ability to answer:
- "what resources types can this role be attached to?", relevant at role binding creation/modification time.
    - this leverages the `scopes`  field
    - because scopes is a noun, we sometimes want to use verbs to describe the relationship between a role and a resource that captures the meaning of scope. This is where I disagree with the word `applies to`. I think `applies to` can be easily confused with `affects`. This is why I advocate for something like `attached to`.
- "what resources types does this role affect during authorization checks?"
    - this leverages the `affects` field, perfect 👍
- Overall, think we should not use the words `applies to` or `relevant`.

Lastly, instead of `scopes` and `affects`, which is a noun and verb respectively, we should be consistent, and use both nouns, or both verbs. I prefer using verbs, because they describe the relationships between two nouns, in this case `role -- verb --> resource`. I think instead of `scopes` we can use something like `can_attach_to`. However this is just a nitpick.

jrockway:
This has been dramatically reworked.  Well, not dramatically.  But the fields are "can be applied to" and "returned for".  "can be applied to" is what we thought ResourceType meant, but it was never actually used for that.  It's nice documentation to keep around, though, because eventually we will use it.  "returned for" is what we actually care about on this issue.

I did fix the project* roles to not be returned_for the cluster.

src/auth/auth.proto
```diff
@@ -256,7 +256,13 @@ message Groups {
 message Role {
   string name = 1;
   repeated Permission permissions = 2;
-  repeated ResourceType resource_types = 3;
+  // Resources this role can be bound to.  For example, you can't apply clusterAdmin to a repo, so
```
albscui, line 259: Nice comments!

albscui (commented):
LGTM!

albscui (approved):

PR #8955, opened 2023-06-20
Migrate Projects to New Data Model

FahadBSyed:


src/internal/coredb/projects.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/jmoiron/sqlx"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
+	"github.com/pachyderm/pachyderm/v2/src/pfs"
+)
+
+// ProjectIterator batches a page of projectRow entries. Entries can be retrieved using iter.Next().
+type ProjectIterator struct {
+	lock     sync.Mutex
```
brendoncarroll, line 20: Iterators are serial in nature.  Concurrent access would not preserve order.  It's probably safe to assume that all access will be single threaded.

src/server/pfs/server/api_server.go
```diff
@@ -71,23 +72,23 @@ func (a *apiServer) ActivateAuth(ctx context.Context, request *pfs.ActivateAuthR
...
 func (a *apiServer) ActivateAuthInTransaction(txnCtx *txncontext.TransactionContext, request *pfs.ActivateAuthRequest) (response *pfs.ActivateAuthResponse, retErr error) {
 	// Create role bindings for projects created before auth activation
-	var projectInfo pfs.ProjectInfo
-	if err := a.driver.projects.ReadWrite(txnCtx.SqlTx).List(&projectInfo, col.DefaultOptions(), func(string) error {
+	projIter, err := coredb.ListProject(txnCtx.Context(), a.env.DB)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list projects")
+	}
+	projInfo := &pfs.ProjectInfo{}
+	for !errors.Is(projIter.Next(txnCtx.Context(), projInfo), io.EOF) {
```
brendoncarroll, line 80: This should be `stream.EOS`  to conform to `stream.Iterator`

src/server/pfs/server/driver.go
```diff
@@ -793,30 +789,47 @@ func (d *driver) getCompactedDiffFileSet(ctx context.Context, commit *pfs.Commit
...
 func (d *driver) listProjectInTransaction(txnCtx *txncontext.TransactionContext, cb func(*pfs.ProjectInfo) error) error {
 	projectInfo := &pfs.ProjectInfo{}
-	return errors.EnsureStack(d.projects.ReadWrite(txnCtx.SqlTx).List(projectInfo, col.DefaultOptions(), func(string) error {
-		return cb(projectInfo)
-	}))
+	projIter, err := coredb.ListProject(txnCtx.Context(), d.env.DB)
+	if err != nil {
+		return errors.Wrap(err, "could not list project")
+	}
+	for !errors.Is(projIter.Next(txnCtx.Context(), projectInfo), io.EOF) {
```
brendoncarroll, line 827: You can probably replace this with `stream.ForEach`.  I think the callback type would match exactly?

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,176 @@
...
+	return errors.Wrap(err, "could not delete all project rows")
+}
+
+// GetProject retrieves an entry from the core.projects table by project name.
+func GetProject(ctx context.Context, queryExecer QueryExecer, projectName string) (*pfs.ProjectInfo, error) {
+	return getProject(ctx, queryExecer, "name", projectName)
+}
+
+// GetProjectByID is like GetProject, but retrieves an entry using the row id.
+func GetProjectByID(ctx context.Context, queryExecer QueryExecer, id int64) (*pfs.ProjectInfo, error) {
```
brendoncarroll, line 129: IMO: `Get_By_` suggests the use of an index, or something other than the identity and `Get_` suggests retrieval by the identity (whatever that is).

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,176 @@
...
+	defer iter.lock.Unlock()
+	if iter.index >= len(iter.projects) {
+		iter.index = 0
+		iter.pageNum++
+		iter.projects, err = listProjectPage(ctx, iter.db, 100, iter.pageNum)
+		if err != nil {
+			return errors.Wrap(err, "failed to list project page")
+		}
+		if len(iter.projects) == 0 {
+			return io.EOF
```
brendoncarroll, line 49: `streams.EOS`

src/server/pfs/server/driver.go
```diff
@@ -793,30 +789,47 @@ func (d *driver) getCompactedDiffFileSet(ctx context.Context, commit *pfs.Commit
...
-		return cb(projectInfo)
-	}), "could not list projects")
+		if err := cb(projInfo); err != nil {
+			return errors.Wrapf(err, "could not execute callback on project %s", projInfo.Project.Name)
+		}
+	}
+	return nil
 }
 
 // The ProjectInfo provided to the closure is repurposed on each invocation, so it's the client's responsibility to clone the ProjectInfo if desired
```
brendoncarroll, line 820: 👍

src/server/pfs/server/driver.go
brendoncarroll: I don't think this handles errors correctly, either.  Non-EOS errors would cause the loop to execute and call cb with bogus.  Switching to ForEach would handle those cases correctly.

src/internal/clusterstate/v2.7.0_test.go
```diff
@@ -0,0 +1,76 @@
...
+	type project struct {
+		ID          int       `db:"id"`
+		Name        string    `db:"name"`
+		Description string    `db:"description"`
+		CreatedAt   time.Time `db:"created_at"`
+		UpdatedAt   time.Time `db:"updated_at"`
+	}
+
+	if DesiredClusterState.Number() > state_2_7_0.Number() {
+		t.Skip("skipping test because desired state is newer than this migration")
```
jrockway, line 28: Why?  The 2.6 -> 2.7 migration still has to work when we release 2.8.

src/internal/clusterstate/v2.7.0_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Apply the migration of interest
+	require.NoError(t, migrations.ApplyMigrations(ctx, db, migrationEnv, state_2_7_0))
+	require.NoError(t, migrations.BlockUntil(ctx, db, state_2_7_0))
+
+	// Check whether all the data is migrated to core.gotProjects table
+	gotProjects := []project{}
+	require.NoError(t, db.SelectContext(ctx, &gotProjects, `SELECT id, name, description, created_at, updated_at FROM core.projects ORDER BY id`))
+	require.Equal(t, len(expectedProjects), len(gotProjects))
+	for i, p := range gotProjects {
```
jrockway, line 69: I'd prefer:

```go
want := []Project{
   { Name: "project1", ...},
   ...
}
if diff := cmp.Diff(got, want); diff != "" {
   t.Errorf("projects (-got +want):\n%s", diff)
}
```
The problem with `require.Equal` for each field is that you can't see the entire failing state without re-running the test.  Diff shows you everything wrong with every project and every field, and removes the need for the len() assertion (which ends up being really hard to debug if it doesn't match).

src/internal/clusterstate/v2.7.0_test.go
```diff
@@ -0,0 +1,76 @@
...
+		require.Equal(t, i+1, p.ID)
+		require.Equal(t, expectedProjects[p.Name].Name, p.Name)
+		require.Equal(t, expectedProjects[p.Name].Description, p.Description)
+		require.Equal(t, expectedProjects[p.Name].CreatedAt, p.CreatedAt)
+		require.Equal(t, expectedProjects[p.Name].UpdatedAt, p.UpdatedAt)
+	}
+}
```
jrockway, line 76: In general, I don't feel great about doing these migration tests in-situ like this.  I think we have to model these after our upgrade tests; install 2.6.0, add a bunch of projects, upgrade to 2.7.0, check that everything is still there.  These are going to be less easy to debug when they fail, but more likely to ensure that we release usable code.

src/internal/coredb/projects.go
jrockway: ID is an index by nature of being a primary key, right? 

Having said that, I'm not sure we will use this function?  I feel like the IDs are always going to be for joins, whereas getting things by name is how every RPC is going to work.  Remember, IDs never leave the system boundary; the API doesn't know them.

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,176 @@
...
+	project.CreatedAt, err = types.TimestampProto(createdAt)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed converting project proto timestamp")
+	}
+	return project, nil
+}
+
+// UpdateProject updates all fields of an existing project entry in the core.projects table by name. If 'upsert' is set to true, UpdateProject()
+// will attempt to call CreateProject() if the entry does not exist.
+func UpdateProject(ctx context.Context, queryExecer QueryExecer, project *pfs.ProjectInfo, upsert bool) error {
```
jrockway, line 150: I think we want to be careful about `upsert bool` becoming a pattern.  Random booleans hanging off the end of functions to control the behavior of the function is pretty confusing to the reader of this code.  It is clearer to call this "InsertOrUpdateProject".

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,176 @@
...
+		projects: page,
+		db:       db,
+	}
+	return iter, nil
+}
+
+func listProjectPage(ctx context.Context, db *pachsql.DB, pageSize, pageNum int) ([]projectRow, error) {
+	var page []projectRow
+	if err := db.SelectContext(ctx, &page, "SELECT name,description,created_at FROM core.projects ORDER BY id ASC LIMIT $1 OFFSET $2",
+		pageSize, pageSize*pageNum); err != nil {
```
jrockway, line 81: I'm not sure I like doing this multiplication here.  I'd call the args "offset" and "limit" and let the caller figure this out.  It's the most clear 1:1 mapping with the underlying data storage.

src/server/pfs/server/api_server.go
```diff
@@ -71,23 +72,23 @@ func (a *apiServer) ActivateAuth(ctx context.Context, request *pfs.ActivateAuthR
 
 func (a *apiServer) ActivateAuthInTransaction(txnCtx *txncontext.TransactionContext, request *pfs.ActivateAuthRequest) (response *pfs.ActivateAuthResponse, retErr error) {
 	// Create role bindings for projects created before auth activation
-	var projectInfo pfs.ProjectInfo
-	if err := a.driver.projects.ReadWrite(txnCtx.SqlTx).List(&projectInfo, col.DefaultOptions(), func(string) error {
+	projIter, err := coredb.ListProject(txnCtx.Context(), a.env.DB)
```
jrockway, line 75: `Context` is a method that was mistakenly added to txnCtx, and appears to have no uses in the code right now.  We should get rid of that ASAP and use a better transaction object; it should not expose the context, but should implement Select/Exec/Query/etc. and automatically pass the context in.

jrockway: Also, I don't see a ton of value in using an iterator other than sql.Rows here; there is no need to do this a page at a time as we have to read every project here.  Might as well not make multiple round-trips to the database.

src/server/pfs/server/api_server.go
jrockway: I think we want some sort of utility function to call a callback for each valid row in a Stream.

jrockway (commented):
I haven't looked at everything yet but want to get some feedback out.  We need to, like right now, figure out the API for transactions.  pachsql.Tx is too limited, as it does not allow logging information to be consistent with the transaction-starter.  txnenv.TransactionContext is too big of a structure; it retains the context for logging (but scarily exposes it for I/O; I'm going to fix that immediately), but also has a ton of other stuff in it, like the PFS commit ID, that doesn't make sense for transactions in general.

src/internal/clusterstate/v2.7.0/coredb.go
```diff
@@ -0,0 +1,89 @@
...
+	if err := createProjectsTable(ctx, tx); err != nil {
+		return errors.Wrap(err, "error creating projects table")
+	}
+	return nil
+}
+
+func createProjectsTable(ctx context.Context, tx *pachsql.Tx) error {
+	if _, err := tx.ExecContext(ctx, `
+		CREATE TABLE IF NOT EXISTS core.projects (
+			id bigserial PRIMARY KEY,
```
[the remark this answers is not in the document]
brendoncarroll: I don't think you can have null primary keys.

src/internal/coredb/projects.go
brendoncarroll: You can have 1 `Get_` method.  You can have as many `Get_By_` methods as you want.  You get 1 primary key, you can have as many indexes as you want.

src/server/pfs/server/api_server.go
brendoncarroll: `stream.ForEach`?

jrockway: Yup.  Thinking about this more, I think we need a way of converting sqlx.Rows into a Stream, and a way of converting an Stream of []sqlx.Rows into a Stream.  This will let us always handle SQL queries with our iterator API, and let us represent the likely-to-be-very-common "paged" approach in the same way; the consumer doesn't have to care which implementation we chose.

brendoncarroll: That's a good idea, but I don't think that should block this PR.

brendoncarroll: Let's move that conversation here.
https://github.com/pachyderm/pachyderm/pull/8959

If it's usefui, then there will be motivation to use it for the Commit and Repo iterators, and we can migrate the Projects Iterator then.

src/internal/clusterstate/v2.7.0/proto.go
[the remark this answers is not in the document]
brycemcanally: I'm not following what is being suggested here. Not copying the protos, or more generally migration dependencies, means that migration logic can be affected by changes to those dependencies over time. Not doing this lead to the bug that @acohen4 recently dealt with when going from 2.4 -> 2.6. The projects migration pulled from the current commit info protos, which had changed in 2.6. An integration test spanning one minor version isn't going to catch that, and it cannot be prevented without some sort of snapshotting of the dependencies unless we add constraints to the changes we can make to dependencies (i.e. no deleting proto fields). Moving away from protos in the database over time should address this for postgres migrations, but this challenge will still apply to the storage format in the long term.

src/internal/clusterstate/v2.7.0_test.go
albscui: I prefer this type of testing for while I'm implementing the migration code. It is much faster to run. We could remove these tests after I implement, or simply skip them.

src/internal/clusterstate/v2.7.0/proto.go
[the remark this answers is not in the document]
jrockway: The only way to capture the state in 2.6 is to get a 2.6 binary and run commands against it. Copying the protos doesn't do enough to achieve this goal, as evidenced by real-life bugs from previous attempts at copying this.  It doesn't work and it's messy, so I don't see why we should do it again.  Basically, the only way to write to the database like 2.6 "would" is to run 2.6 to do the writes.

Unlike our current very-basic "upgrade tests", this one would have to stick around forever.  We typically change the upgrade tests after release to be "latest release -> master", but this new one has to be 2.6 -> master forever.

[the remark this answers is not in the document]
jrockway: Oh, the underlying assumption is that we wouldn't be deleting fields in the proto until 3.0 (at which point we'll ship a snapshot of the 2.x protos to read data to migrate into).  If we want to delete fields but still want to read them during migrations, I'm not sure that copying the proto is worth the effort.  Mark the deleted field deprecated and let the linter complain when you use them outside migrations.

src/internal/clusterstate/v2.7.0_test.go
albscui: I can add the upgrade tests later.

src/internal/clusterstate/v2.7.0/coredb.go
```diff
@@ -0,0 +1,89 @@
...
+			name text UNIQUE NOT NULL,
```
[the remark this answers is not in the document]
jrockway: project + pipeline < 63 or something, so if pipeline is 0, then project has to be less than 63.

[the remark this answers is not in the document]
jrockway: also: ```shell
$ pachctl create project abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz
invalid project name: project name "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz" is 27 characters longer than the 51 max
```

File: src/internal/clusterstate/v2.7.0/proto.go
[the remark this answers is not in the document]
Comment (brycemcanally) [reply src/internal/clusterstate/v2.7.0/proto.go:1]: I'm still not following. It is preferable to do nothing instead of somewhat, and probably mostly, protect against this issue? I don't even think the issue @acohen4 ran into could have been addressed without copying. Having old migrations depend on current version protos is a footgun that has already caused a problem.  What was the bug caused by copying? I only know of the bug that was caused by not copying.

File: src/internal/coredb/projects.go [hunk 15]
```diff
@@ -0,0 +1,176 @@
...
+	iter := &ProjectIterator{
+		projects: page,
+		db:       db,
+	}
+	return iter, nil
+}
+
+func listProjectPage(ctx context.Context, db *pachsql.DB, pageSize, pageNum int) ([]projectRow, error) {
+	var page []projectRow
+	if err := db.SelectContext(ctx, &page, "SELECT name,description,created_at FROM core.projects ORDER BY id ASC LIMIT $1 OFFSET $2",
```
brycemcanally, line 80: I think we should establish a different pattern for pagination because offset / limit will not perform well at scale. Although it is probably fine for projects, this seems like it will be the template for the rest of the tables. We should probably change this to keyset pagination.

src/internal/clusterstate/v2.7.0/proto.go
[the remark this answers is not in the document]
brycemcanally: Okay, yeah, the followup post clears things up (it looks like you posted it before I refreshed). So you are thinking we just shouldn't delete proto fields. I lean the other way because I generally prefer complexity to be moved to the migration process rather than the current codebase. I would prefer developers having to copy the protos involved in their migration rather than developers always having to account for deprecated fields in a proto.

src/internal/clusterstate/v2.7.0/coredb.go
```diff
@@ -0,0 +1,92 @@
...
+
+// CollectionRecord is a record in a collections table.
+type CollectionRecord struct {
+	Key       string    `db:"key"`
+	Proto     []byte    `db:"proto"`
+	CreatedAt time.Time `db:"createdat"`
+	UpdatedAt time.Time `db:"updatedat"`
+}
+
+func createCoreSchema(ctx context.Context, tx *pachsql.Tx) error {
```
brycemcanally, line 22: Are the other tables going to be added to this function over time? It seems like the table creation should be in the apply function that is performing the migration, and the future tables should just be a part of additional apply functions.

src/internal/coredb/projects.go
brycemcanally: Yeah, it seems like the names are inverted. `GetProject` should take the id and `GetProjectByName` should be used for the name. Internal usage may generally lean towards `Get*` with an id, specifically for the background processes that are watching and will receive ids. I don't think we have any background processes for projects yet, but repos, branches, commits, etc. will be using ids.

src/server/pfs/server/api_server.go
brycemcanally: I agree with getting rid of the context in txnCtx, although I'm not sure how I feel about doing queries outside of the transaction while using a transaction object. It seems like the usage should either be clearly outside of a transaction with a context, or clearly inside a transaction with only a `sql.Tx`.

src/internal/coredb/projects.go
brycemcanally: An alternative would be to just have an `Upsert` function and keep the complexity of additional parameters for the unexported function.

src/internal/coredb/projects.go
FahadBSyed: Based on our conversation, we want to standardize on retrieval by ID in our codebase. GetProject will be based on ID and we will have a GetProjectByName for RPCs.

src/internal/coredb/projects.go
FahadBSyed: I'm going to have UpsertProject which will determine the target project based on project name. Update project will require an id which should imply that the project already exists.

src/internal/clusterstate/v2.7.0/coredb.go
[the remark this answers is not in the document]
albscui: That's true, good point. While this is currently being validated in code, it's more rigorous to enforce this at the DB level.

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,177 @@
...
+	offset   int
+	projects []projectRow
+	index    int
+	db       *pachsql.DB
+}
+
+// ID is the auto-incrementing primary key used for entries in postgres tables.
+type ID uint64
+
+type projectRow struct {
```
albscui, line 29:
```suggestion
type projectRow struct {
		ID          ID      `db:"id"`
		Name        string    `db:"name"`
		Description string    `db:"description"`
		CreatedAt   time.Time `db:"created_at"`
		UpdatedAt   time.Time `db:"updated_at"`
	}
```

src/internal/dbutil/tx.go
```diff
@@ -113,7 +114,7 @@ func WithBackOff(bo backoff.BackOff) WithTxOption {
 // WithTx calls cb with a transaction,
 // The transaction is committed IFF cb returns nil.
 // If cb returns an error the transaction is rolled back.
-func WithTx(ctx context.Context, db *pachsql.DB, cb func(tx *pachsql.Tx) error, opts ...WithTxOption) error {
+func WithTx(ctx context.Context, db *pachsql.DB, cb func(ctx context.Context, tx *pachsql.Tx) error, opts ...WithTxOption) error {
```
albscui, line 117: I would name the inner ctx something else just to reduce ambiguity
```suggestion
func WithTx(ctx context.Context, db *pachsql.DB, cb func(ctx2 context.Context, tx *pachsql.Tx) error, opts ...WithTxOption) error {
```

src/internal/ppsutil/util.go
```diff
@@ -170,7 +170,7 @@ func SetPipelineState(ctx context.Context, db *pachsql.DB, pipelinesCollection c
 	logSetPipelineState(ctx, pipeline, from, to, reason)
 	var resultMessage string
 	var warn bool
-	err := dbutil.WithTx(ctx, db, func(sqlTx *pachsql.Tx) error {
+	err := dbutil.WithTx(ctx, db, func(ctx context.Context, sqlTx *pachsql.Tx) error {
```
albscui, line 173: Same line of reasoning as above, reduce ambiguity
```suggestion
	err := dbutil.WithTx(ctx, db, func(ctx2 context.Context, sqlTx *pachsql.Tx) error {
```

src/internal/storage/chunk/client.go
```diff
@@ -79,7 +79,7 @@ func (c *trackedClient) beforeUpload(ctx context.Context, chunkID ID, md Metadat
 		pointsTo = append(pointsTo, cid.TrackerID())
 	}
 	chunkTID := chunkID.TrackerID()
-	if err := dbutil.WithTx(ctx, c.db, func(tx *pachsql.Tx) (retErr error) {
+	if err := dbutil.WithTx(ctx, c.db, func(ctx context.Context, tx *pachsql.Tx) (retErr error) {
```
albscui, line 82: So since we are passing in a new `ctx`, below we should use the Contexted version of the sqlx API, like `tx.SelectContext()` etc.

src/internal/clusterstate/v2.7.0_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/pfs"
+)
+
+// Create sample test data in collections.projects table
+func setupTestData(t *testing.T, ctx context.Context, db *sqlx.DB) {
+	t.Helper()
+	tx, err := db.BeginTxx(ctx, nil)
+	require.NoError(t, err)
+	defer tx.Rollback()
+	for i := 0; i < 12; i++ {
```
jrockway, line 28: This feels a little too coupled to implementation details.  It would be best to go through the API server and collections API to add "old" records.  Otherwise the test is just testing how you think the API server works, not how it actually works, which isn't ideal.

src/internal/coredb/projects_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/pfs"
+)
+
+const (
+	testProj     = "testproj"
+	testProjDesc = "this is a test project"
+)
+
+func TestCreateProject(t *testing.T) {
+	ctx := pctx.TestContext(t)
```
jrockway, line 27: I think these can all be `t.Parallel()`

src/internal/coredb/projects_test.go
```diff
@@ -0,0 +1,153 @@
...
+	require.NoError(t, err, "should be able to create a tx")
+	defer tx.Rollback()
+	migrationEnv := migrations.Env{EtcdClient: testetcd.NewEnv(ctx, t).EtcdClient}
+	require.NoError(t, migrations.ApplyMigrations(ctx, db, migrationEnv, clusterstate.DesiredClusterState), "should be able to set up tables")
+	createInfo := &pfs.ProjectInfo{Project: &pfs.Project{Name: testProj}, Description: testProjDesc}
+	require.NoError(t, CreateProject(ctx, tx, createInfo), "should be able to create project")
+	getInfo, err := GetProjectByName(ctx, tx, testProj)
+	require.NoError(t, err, "should be able to get a project")
+	require.Equal(t, createInfo.Project.Name, getInfo.Project.Name)
+	require.Equal(t, createInfo.Description, getInfo.Description)
```
jrockway, line 39: It would probably be good to try creating a project with the same name here and seeing it fail because of the unique constraint.

src/internal/coredb/projects_test.go
```diff
@@ -0,0 +1,153 @@
...
+	db := dockertestenv.NewTestDB(t)
+	tx, err := db.BeginTxx(ctx, nil)
+	require.NoError(t, err, "should be able to create a tx")
+	defer tx.Rollback()
+	migrationEnv := migrations.Env{EtcdClient: testetcd.NewEnv(ctx, t).EtcdClient}
+	require.NoError(t, migrations.ApplyMigrations(ctx, db, migrationEnv, clusterstate.DesiredClusterState), "should be able to set up tables")
+	size := 3
+	for i := 0; i < size; i++ {
+		createInfo := &pfs.ProjectInfo{Project: &pfs.Project{Name: fmt.Sprintf("%s%d", testProj, i)}, Description: testProjDesc, CreatedAt: types.TimestampNow()}
+		require.NoError(t, CreateProject(ctx, tx, createInfo), "should be able to create project")
```
jrockway, line 118: Putting `i` in the message is always helpful.  Does the first iteration fail, or the last?

jrockway (commented):
This is looking good.  The upgrade tests are essential and must be in this PR.

bbonenfant (commented):
I ran these changes over the python sdk tests and now calling InspectPipeline on a pipeline that doesn't exist causes the following error:
```
<_InactiveRpcError of RPC that terminated with:
           status = StatusCode.UNKNOWN
           details = "error getting project "proj240": failed scanning project row: sql: no rows in result set"
           debug_error_string = "UNKNOWN:Error received from peer ipv4:192.168.49.2:30080 {created_time:"2023-06-27T17:30:15.608021246+00:00", grpc_status:2, grpc_message:"error getting project \"proj240\": failed scanning project row: sql: no rows in result set"}"
>
```

previously the error was:
```
<_InactiveRpcError of RPC that terminated with:
	status = StatusCode.NOT_FOUND
	details = "projects proj746 not found"
	debug_error_string = "UNKNOWN:Error received from peer ipv4:192.168.7.17:30080 {created_time:"2023-06-27T11:05:39.741937-07:00", grpc_status:5, grpc_message:"projects proj746 not found"}"
>
```

I would think we would still want to keep the status as `NOT_FOUND` and I'm not sure if we want that sql error to make it back to the client

src/internal/clusterstate/v2.7.0_test.go
albscui: Valid point, and I'm a bit torn on this. I wanted to decouple my test code from collections package as much as possible, because eventually we will be removing it. However, going through the API server makes sense!

jrockway: As I read down a little farther in the PR, I decided I don't object to this existing, but the scenario I'm talking about does need to be covered.

It's going to be more complicated to have tests that start by running the old version, but we're kind of changing the wings of an airplane while it's flying, so it's going to be more painful than starting from scratch.

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,171 @@
...
+func GetProjectByName(ctx context.Context, tx *pachsql.Tx, projectName string) (*pfs.ProjectInfo, error) {
+	return getProject(ctx, tx, "name", projectName)
+}
+
+func getProject(ctx context.Context, tx *pachsql.Tx, where string, whereVal interface{}) (*pfs.ProjectInfo, error) {
+	row := tx.QueryRowxContext(ctx, fmt.Sprintf("SELECT name, description, created_at FROM core.projects WHERE %s = $1", where), whereVal)
+	project := &pfs.ProjectInfo{Project: &pfs.Project{}}
+	var createdAt time.Time
+	var err error
+	if err = row.Scan(&project.Project.Name, &project.Description, &createdAt); err != nil {
```
jrockway, line 133: :=

jrockway:
Good point @bbonenfant.  I don't like the ad-hoc errors in the coredb/projects package.  We should create error types and use them throughout:

```go
type ProjectNotFound{Name string, ID int}

func (err *ProjectNotFound) Error() string {
    if n := err.Name; n != "" {
       return fmt.Sprintf("project %q not found", n)
    }
    if id := err.ID; n != "" {
        return fmt.Sprintf("project id=%d not found", id)
    }
    return "project not found"
}

func (err *ProjectNotFound) Is(target error) bool {
   ...
}

func (err *ProjectNotFound) GRPCStatus() status.Status {
    return status.New(codes.NotFound, ...)
}
```


func getProject(...){
     ...
     if err := sqlx.QueryRow(...); err != nil {
         if errors.Is(err, sql.ErrNoRows) {
              return ProjectNotFound{Name: foo}
         }
         return errors.Wrap(...)
     }
     ...
```

While I'm here, I'll also suggest not including the word "failed" in error objects.  It's an error so you know it failed:

```patch
- return errors.Wrap(err, "failed to look up project by name")
+ return errors.Wrap(err, "look up project by name")
```

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,171 @@
...
+type projectRow struct {
+	ID          ID        `db:"id"`
+	Name        string    `db:"name"`
+	Description string    `db:"description"`
+	CreatedAt   time.Time `db:"created_at"`
+	UpdatedAt   time.Time `db:"updated_at"`
+}
+
+// Next advances the iterator by one row. It returns a stream.EOS when there are no more entries.
+func (iter *ProjectIterator) Next(ctx context.Context, dst **pfs.ProjectInfo) error {
```
robert-uhl, line 39: How come `Next` writes a pointer to a pointer pointer and returns an error rather than returning a pointer and an error?

If it were writing data into a pointer, that would make sense as a performance hack (a single allocation of the underlying data for the entire run), but this allocations each time through. In languages with only a single return value, you have to take out-args in order to return a value and an error, but in Go you can just return both a value and an error directly.

FahadBSyed: Its to satisfy the stream.ForNext() pattern which uses generics. So if we want to update a *pfs.ProjectInfo, we need a pointer to that pointer in order to update the address of *pfs.ProjectInfo. Perhaps @brendoncarroll can clarify some of the details

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,171 @@
...
+	result, err := tx.ExecContext(ctx, "DELETE FROM core.projects WHERE name = $1;", projectName)
+	if err != nil {
+		return errors.Wrap(err, "failed to delete project")
+	}
+	rowsAffected, err := result.RowsAffected()
+	if err != nil {
+		return errors.Wrap(err, "could not get affected rows")
+	}
+	if rowsAffected == 0 {
+		return errors.Wrap(fmt.Errorf("project %s does not exist", projectName), "failed to delete project")
```
robert-uhl, line 110: You may wish to return some sort of structured error, or just a wrapped `ErrNotExist`/`ErrProjectNotExist`/something.

src/internal/coredb/projects.go
jrockway: This is the `stream` API.  `Iterator[*pfs.ProjectInfo]` is going to have a `Next` function that is `**pfs.ProjectInfo`.  Iterator is over `*pfs.ProjectInfo` because it's the type that implements `ProtoMessage()`, not `pfs.ProjectInfo`.

src/server/pfs/server/commit_store.go
```diff
@@ -256,24 +256,3 @@ func commitDiffTrackerID(commit *pfs.Commit, fs fileset.ID) string {
 func commitTotalTrackerID(commit *pfs.Commit, fs fileset.ID) string {
 	return commitTrackerPrefix + pfsdb.CommitKey(commit) + "/total/" + fs.HexString()
 }
-
-// SetupPostgresCommitStoreV0 runs SQL to setup the commit store.
-// DO NOT MODIFY THIS FUNCTION
-// IT HAS BEEN USED IN A RELEASED MIGRATION
-func SetupPostgresCommitStoreV0(ctx context.Context, tx *pachsql.Tx) error {
```
brendoncarroll, line 263: Where did this go?

src/internal/clusterstate/v2.7.0_test.go
albscui: Hmm after thinking about it a bit more, I'm not sure how to achieve your scenario 😓.
The big issue is that we already changed the implementation of our API server to use the new tables, and I'm not sure how to swap this implementation out for the old collections style.

bbonenfant (approved):
:+1: from the integrations team

src/server/pfs/server/commit_store.go
FahadBSyed: I am not sure, I don't think this should be getting deleted

src/internal/clusterstate/v2.7.0_test.go
jrockway: Start up pachyderm 2.6, make the API calls, upgrade to 2.7, check the data.  The old API server is running at the start of the test.  If the client changes so significantly that this isn't possible, then we need to put the pachctl binary from 2.6 into the CI environment and use that to create the objects in 2.6.

src/internal/coredb/projects.go
```diff
@@ -0,0 +1,229 @@
...
+func (err ErrProjectNotFound) Is(other error) bool {
+	_, ok := other.(ErrProjectNotFound)
+	return ok
+}
+
+func (err ErrProjectNotFound) GRPCStatus() *status.Status {
+	return status.New(codes.NotFound, err.Error())
+}
+
+// ErrProjectDoesNotExist is returned by DeleteProject() when a project is not found in postgres.
```
jrockway, line 48: Honestly, I think DoesNotExist and NotFound can be the same error.

src/server/pfs/server/api_server.go
```diff
@@ -69,25 +72,27 @@ func (a *apiServer) ActivateAuth(ctx context.Context, request *pfs.ActivateAuthR
...
 			// Grant all users ProjectWriter role for default project.
 			principal = auth.AllClusterUsersSubject
 			roleSlice = []string{auth.ProjectWriterRole}
 		}
-
-		err := a.env.AuthServer.CreateRoleBindingInTransaction(txnCtx, principal, roleSlice, &auth.Resource{Type: auth.ResourceType_PROJECT, Name: projectInfo.Project.Name})
+		err := a.env.AuthServer.CreateRoleBindingInTransaction(txnCtx, principal, roleSlice, &auth.Resource{Type: auth.ResourceType_PROJECT, Name: proj.Project.Name})
 		if err != nil && !col.IsErrExists(err) {
-			return errors.EnsureStack(err)
+			return errors.Wrap(err, "failed to create role binding in transaction")
```
jrockway, line 91: "create role binding in transaction"  (and 1 below)

src/server/pfs/server/driver.go
```diff
@@ -656,16 +653,23 @@ func (d *driver) createProjectInTransaction(txnCtx *txncontext.TransactionContex
 }
 
 func (d *driver) inspectProject(ctx context.Context, project *pfs.Project) (*pfs.ProjectInfo, error) {
-	pi := &pfs.ProjectInfo{}
-	if err := d.projects.ReadOnly(ctx).Get(pfsdb.ProjectKey(project), pi); err != nil {
-		return nil, errors.EnsureStack(err)
+	var pi *pfs.ProjectInfo
+	var err error
+	if err = dbutil.WithTx(ctx, d.env.DB, func(ctx context.Context, tx *pachsql.Tx) error {
```
jrockway, line 658: I don't love leaking the error between scopes like this.  In general `if err = ...` is the source of many mistakes, and if the linter were more flexible I'd have banned it by now.  I think in this case, moving `var error` into the callback makes the most sense.

```go
if err := dbutil.WithTx(..., func(...){ 
   var err error
   pi, err = coredb.GetProjectByName()
   ...
}
[x]
```
The way there is no `err` defined at `[x]`, which makes it less likely to be accidentally misused later.

src/server/pfs/server/driver.go
```diff
@@ -656,16 +653,23 @@ func (d *driver) createProjectInTransaction(txnCtx *txncontext.TransactionContex
...
+		pi, err = coredb.GetProjectByName(ctx, tx, pfsdb.ProjectKey(project))
+		if err != nil {
+			return err
+		}
+		return nil
+	}); err != nil {
+		return nil, errors.Wrapf(err, "error getting project %q", project)
```
jrockway, line 665: "get project %q"

src/server/pfs/server/driver.go
jrockway, line 658: Also, something I worry about in general is what happens when something transactional wants to call inspectProject.  I don't think it ever happens, so there is nothing to change here today, but it's why we see `xxxInTransaction` methods throughout the codebase.

Especially below when we AuthServer.GetPermissions(), ideally that would be the same view of the database as we had when retrieving the project.  But, Auth will probably eventually be its own service and not be able to be transactional with projects, so I'm not super bothered.  Something to keep in mind though, as the refactor proceeds.

src/server/pfs/server/driver.go
```diff
@@ -793,34 +797,40 @@ func (d *driver) getCompactedDiffFileSet(ctx context.Context, commit *pfs.Commit
...
-	projectInfo := &pfs.ProjectInfo{}
-	return errors.Wrap(d.projects.ReadOnly(ctx).List(projectInfo, col.DefaultOptions(), func(string) error {
-		if authIsActive {
-			resp, err := d.env.AuthServer.GetPermissions(ctx, &auth.GetPermissionsRequest{Resource: projectInfo.GetProject().AuthResource()})
-			if err != nil {
-				if errors.Is(err, auth.ErrNotActivated) {
-					// Avoid unnecessary subsequent Auth API calls.
-					authIsActive = false
-					return cb(projectInfo)
+	return errors.Wrap(dbutil.WithTx(ctx, d.env.DB, func(ctx context.Context, tx *pachsql.Tx) error {
```
jrockway, line 800: I think it would be best if listProject directly called listProjectInTransaction, rather than making the coredb calls itself.  I think it runs the risk of diverging when someone adds a special case here or there.

I suppose the wrinkle is that "InTransaction" refers to a pachyderm transaction and not a database transaction.  We probably need a new convention to allow all 3 cases; list X, list X in a sql Tx, list X in a pach Tx.  I think the right helper function could support not duplicating code and letting the caller decide; write `func (me) thatFunction(tx) { list stuff }`, and call `WithTx(tx, me.thatFunction)`/`WithPachTx(ptx, me.thatFunction)`/etc, eliminating the need for someone to type in each of these wrappers.

src/server/pfs/server/driver.go
```diff
@@ -793,34 +797,40 @@ func (d *driver) getCompactedDiffFileSet(ctx context.Context, commit *pfs.Commit
...
+		if err != nil {
+			return err
+		}
+		return stream.ForEach[*pfs.ProjectInfo](ctx, projIter, func(proj *pfs.ProjectInfo) error {
+			if authIsActive {
+				resp, err := d.env.AuthServer.GetPermissions(ctx, &auth.GetPermissionsRequest{Resource: proj.GetProject().AuthResource()})
+				if err != nil {
+					if errors.Is(err, auth.ErrNotActivated) {
+						// Avoid unnecessary subsequent Auth API calls.
+						authIsActive = false
```
jrockway, line 811: Something that's weird to me is that Pachyderm transactions have some internal state that knows if auth is activated or not, it seems weird that we're skipping that here.  But I guess the old code did it.

src/server/pfs/server/testing/server_test.go
```diff
@@ -352,7 +352,7 @@ func TestPFS(suite *testing.T) {
 			{tu.UniqueString("lenny-123"), ""},
 			// {tu.UniqueString("_project"), badFormatErr}, // Require CORE-1343
 			// {tu.UniqueString("project-"), badFormatErr},
-			{pfs.DefaultProjectName, "already exists"},
+			{pfs.DefaultProjectName, "duplicate key value violates unique constraint \"projects_name_key\" (SQLSTATE 23505)"},
```
jrockway, line 355: I bet the end user wants "already exists" as the message.  Internally, `src/internal/errutil/errutil.go` contains a utility that looks for that exact string 😓.

src/server/worker/driver/driver.go
```diff
@@ -17,6 +16,8 @@ import (
 	"syscall"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
jrockway, line 19: I'm not sure your IDE should have moved `metav1` up into its own section.  This is a weird goland-ism; goimports and gopls won't touch it; it assumes it's where you want it, but goland always does weird things like this.  Not a big deal and you don't have to change it, but if it's some setting that can be configured, it would be good to configure it.

I think we *aim* for sectioning imports by stdlib, pachyderm, and third-pary, but most reliably mix everything except stdlib into one paragraph.

jrockway (approved):
I'm feeling very good about this.  I really want a commitment to write the upgrade test immediately after this; we have to have a smooth upgrade process for 2.7.  Put the ticket into this sprint.

robert-uhl:
> ```go
> func (err *ProjectNotFound) GRPCStatus() status.Status {
>     return status.New(codes.NotFound, ...)
> }
> ```

I suggest using the structured errors in package `google.golang.org/genproto/googleapis/rpc/errdetails`:

```go
func (err *ProjectNotFound) GRPCStatus() status.Status {
    var resourceName string
    description := "project not found"
    switch {
      case err.Name != "":
        resourceName = "name:" + err.Name
      case err.ID != "":
        resourceName = "id:" + err.ID
      default:
        resourceName = "unknown"
        description = "project not found; no name or ID indicated"
    }
    return status.New(codes.NotFound, ...).WithDetails(&errdetails.ResourceInfo{
      ResourceType: "pfs:project",
      ResourceName: resourceName,
      Description: description,
  })
}
```

gRPC clients can then parse information about which project doesn’t exist out of the error status. More details at https://cloud.google.com/apis/design/errors

We could define our own details messages, but honestly the Google-provided ones are pretty decent.

src/internal/coredb/projects.go
brendoncarroll: Yeah this is just what happens when you mix the `stream.Iterator` with the protobuf types.  It's not ideal, but there are functions like `stream.Next` to make it a little less weird.

src/internal/storage/chunk/client.go
```diff
@@ -87,13 +87,13 @@ func (c *trackedClient) beforeUpload(ctx context.Context, chunkID ID, md Metadat
 		pointsTo = append(pointsTo, cid.TrackerID())
 	}
 	chunkTID := chunkID.TrackerID()
-	if err := dbutil.WithTx(ctx, c.db, func(tx *pachsql.Tx) (retErr error) {
+	if err := dbutil.WithTx(ctx, c.db, func(cbCtx context.Context, tx *pachsql.Tx) (retErr error) {
```
brendoncarroll, line 90: Calling this `cbCtx` means that any transaction code that was using a context, is still probably using that context.  Most of that code will be referring to a context `ctx`, which we would shadow by calling this `ctx`.

I can see the reason for doing this to keep the change non-functional, but we should renaming these back to `ctx`, and ensuring that it is the only context used in the transaction.

brendoncarroll (approved):
PFS/Storage stuff looks good.

brendoncarroll:
https://github.com/pachyderm/pachyderm/pull/8955#issuecomment-1611412432

The `pacherr` package was our attempt at moving towards structured errors, and away from the string matching in `errutil`.  Long term all of our errors should implement `GRPCStatus` and have the error code set appropriately.  So people without access to our internal packages can make decisions based on the code alone e.g. `NotFound`, instead of having to understand our error types in detail.

src/server/pfs/server/driver.go
albscui: So there's a couple of different ways to determine whether auth is active. The current "best practice" is just to assume auth is active, make any auth call and see if you get back a `ErrNotActivated`. Alternatively, we can just look at the db and see if a role binding exists for the cluster resource. I think the prior is better.

However, as you can see here, this is an example of doing IO within a SQL transaction, which is the constraint we relaxed as part of our design discussion around transactions.

Final point is that this code is a bit scary, since `GetPermissions` is not transactional. We are basically hoping that permissions don't change that much during the life time of this transactional list projects. I think for now, this is a safe assumption.

src/server/pfs/server/driver.go
FahadBSyed: For the scope of this PR I can have listProject call listProjectInTransaction, but I won't add the helpers just yet.

PR #8959, opened 2023-06-20
stream: Adds Blueshift for unpacking slices

brendoncarroll:


jrockway (approved):
LGTM

src/internal/stream/shift.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func (it *blueshift[T]) Next(ctx context.Context, dst *T) error {
+	if it.i >= len(it.buf) {
+		if err := it.inner.Next(ctx, &it.buf); err != nil {
+			return err
+		}
+		it.i = 0
+	}
+	*dst = it.buf[it.i]
```
brendoncarroll, line 28: Should use `it.cp` to do the copy here.

PR #8962, opened 2023-06-21
sync all debug server Send()s

acohen4:


jrockway:
In general, mutexes are a synchronization tool of last resort and are only appropriate for guarding fast operations.  Since `Send` does I/O, you can run into the situation where the mutex is held for a long time, blocking all progress of other goroutines.  Now, the case where it takes a long time is probably one where the stream is broken, so it might not matter.  But a better practice would be:

```go
type Sender struct {
   s debug.Dump_APIServer
   ch chan *debug.DumpChunk
}

func (s *Sender) Run() {
   for _, chunk := range s.ch {
       s.s.Send(chunk)
   }
}

func (s *Sender) SendProgress(ctx context.Context, p *debug.DumpChunk_Progress) error {
    select {
        case <-ctx.Done():
           return pctx.Err(ctx)
        case s.ch <- &Chunk{ Chunk: p }
           return nil
     }
}
...
```
This way, you can give up waiting and return early.  Like I said, I'm not sure it absolutely matters in this case, but we have more control over how long we wait for the send.  SendProgress could, for example, be even more aggressive; if the channel isn't free, don't bother sending the progress update.  That way, no operation is ever blocked by reporting its progress (and it can just have a MarkComplete that does block at the very end).  (that would look like:

```go
func SendProgress() {
   select {
       case <-ctx.Done():
           return err
       case s.ch <- progress:
           return nil
        default: // would have blocked, discard
            return nil
     }
}
```
The pattern that especially scares me is copying references to mutexes; the way you've done it looks safe to me, but I think it's a dangerous pattern.

jrockway (approved):

PR #8971, opened 2023-06-22
[PFS-91] Integrate PR feedback -- stream interface + misc

FahadBSyed:


src/internal/coredb/projects.go
```diff
@@ -120,14 +118,14 @@ func DeleteAllProjects(ctx context.Context, queryExecer QueryExecer) error {
 	return errors.Wrap(err, "could not delete all project rows")
 }
 
-// GetProject retrieves an entry from the core.projects table by project name.
-func GetProject(ctx context.Context, queryExecer QueryExecer, projectName string) (*pfs.ProjectInfo, error) {
-	return getProject(ctx, queryExecer, "name", projectName)
+// GetProject is like GetProjectByName, but retrieves an entry using the row id.
+func GetProject(ctx context.Context, queryExecer QueryExecer, id int64) (*pfs.ProjectInfo, error) {
```
albscui, line 122: I think the id can be `uint64`. In fact I think we should define our own int type by something like
```Go
type ID uint64
```

albscui (approved):
LGTM!

PR #8972, opened 2023-06-22
stream: make EOS a function instead of var, include stack trace

brendoncarroll:
This will add stack traces, so if the end of stream is unexpected, it will be easier to diagnose the problem.

### DIFF
--- a/src/internal/stream/iterator.go
+++ b/src/internal/stream/iterator.go
@@ -6,11 +6,20 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 )
 
-// EOS signals the end of the stream
-var EOS = errors.New("end of stream")
+type ErrEndOfStream struct{}
 
+func (e ErrEndOfStream) Error() string {
+	return "end of stream"
+}
+
+// EOS returns a new end of stream error
+func EOS() error {
+	return errors.EnsureStack(ErrEndOfStream{})
+}
+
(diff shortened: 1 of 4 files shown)


src/internal/stream/iterator.go
```diff
@@ -6,11 +6,20 @@ import (
...
+
+// EOS returns a new end of stream error
+func EOS() error {
+	return errors.EnsureStack(ErrEndOfStream{})
+}
+
+// IsEOS returns true if the error is an end of stream error.
 func IsEOS(err error) bool {
-	return errors.Is(err, EOS)
+	return errors.As(err, &ErrEndOfStream{})
```
jrockway, line 22: `errors.Is(err, ErrEndOfStream{})` should work here.  See: https://go.dev/play/p/Lpg2OyM35II

src/internal/stream/iterator.go
```diff
@@ -6,11 +6,20 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 )
 
-// EOS signals the end of the stream
-var EOS = errors.New("end of stream")
+type ErrEndOfStream struct{}
```
jrockway, line 9: This probably doesn't need to be public, since we want people to use IsEOS and not a manual errors.Is.

jrockway (approved):
LGTM

PR #8973, opened 2023-06-22
storage/kv: add kv.BucketStore for Go CDK blob.Buckets

brendoncarroll:
- Adds adapter from Go CDK's `blob.Bucket` to `kv.Store`.
- Adds Bucket field to `storage.Env`
- Storage server setup now uses the `blob.Bucket` if it is available.  Change storage integration test to use this.

### DIFF
--- a/src/internal/storage/kv/obj_adapter2.go
+++ b/src/internal/storage/kv/obj_adapter2.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
 	"github.com/pachyderm/pachyderm/v2/src/internal/miscutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pacherr"
 	"github.com/pachyderm/pachyderm/v2/src/internal/stream"
 
 	"gocloud.dev/blob"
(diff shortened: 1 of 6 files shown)


src/internal/storage/kv/obj_adapter2.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/miscutil"
+	"github.com/pachyderm/pachyderm/v2/src/internal/stream"
+
+	"gocloud.dev/blob"
+	"gocloud.dev/gcerrors"
+)
+
+type ObjectStore2 struct {
```
FahadBSyed, line 15: The type in `obj_adapter.go` is `objAdapter`, so you could potentially just call this `ObjectStore` and drop the `2`

FahadBSyed:
I think you might have some linting errors where you `return err`. I believe the team has converged on using `errors.Wrap()` to wrap these errors instead of using `errors.EnsureStack()`

FahadBSyed (approved):
Comments are non-blocking. Changes are approved assuming the linting errors will get fixed in the next commit to this branch.

src/internal/storage/kv/obj_adapter2.go
brendoncarroll: I was thinking we would rename this to something that makes more sense after we delete the other one.

PR #8974, opened 2023-06-22
[CORE-1786] Fix create pipeline from URL

robert-uhl:


molinamelendezj (approved):
great! thanks for the quick fix.

PR #8976, opened 2023-06-23
Lbliii/pachctl doc updates

lbliii:
This PR updates the cobra version, modifies `pachctl-doc` to output within its directory (previously it was still building a path for mkdocs), and re-writes CLI descriptions/examples for all PFS commands.

## How to Test the Pachctl CLI's docs output

1. Navigate to `/src/server/cmd/pachctl` and run `go build`. 
2. Execute `./pachctl` to see the updated CLI and interact with it.
3. Explore other command outputs in the same way you normally would, for example: `./pachctl list`, `./pachctl put file`, etc.


src/server/pfs/cmds/cmds.go
```diff
@@ -1924,15 +2043,17 @@ $ {{alias}} foo@master:path1 bar@master:path2`,
...
 	}
 	deleteFile.Flags().BoolVarP(&recursive, "recursive", "r", false, "Recursively delete the files in a directory.")
-	deleteFile.Flags().StringVar(&project, "project", project, "Project in which repo is located.")
+	deleteFile.Flags().StringVar(&project, "project", project, "Specify the project (by name) where the file's repo is located.")
 	shell.RegisterCompletionFunc(deleteFile, shell.FileCompletion)
 	commands = append(commands, cmdutil.CreateAliases(deleteFile, "delete file", files))
 
+	// OBJECT COMMANDS
+
 	objectDocs := &cobra.Command{
```
brendoncarroll, line 2052: I didn't think we supported object APIs and more.

brendoncarroll (approved):
LGTM

src/server/pfs/cmds/cmds.go
lbliii: I think you are right, that was the only one there. I can look into removing it after we get these changes in!

PR #8978, opened 2023-06-23
[CORE-1788] Handle SIGTERM

robert-uhl:


src/internal/signals/signals_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	g.Go(func() error {
+		p, err := os.FindProcess(os.Getpid())
+		if err != nil {
+			return errors.Wrap(err, "could not find own process")
+		}
+		return errors.Wrap(p.Signal(s), "could not signal self")
+	})
+
+	g.Go(func() error {
```
acohen4, line 37: looks good. out of curiosity, is there any value in this being in a goroutine?

acohen4 (approved):
LGTM

src/internal/signals/signals_test.go
robert-uhl: I think that it helps to think about signalling in one goroutine and handling signals in another, and I do like the consolidated error handling with `errgroup.Group`, but it’s not technically required.

kevyang (approved):
lgtm. thanks!

PR #8979, opened 2023-06-23
Break PFS -> clusterstate dependency cycle

brendoncarroll:


src/internal/pfsdb/schema.go
```diff
@@ -0,0 +1,29 @@
...
+	"context"
+
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
+)
+
+// SchemaCommitStoreV0 runs SQL to setup the commit store.
+// DO NOT MODIFY THIS FUNCTION
+// IT HAS BEEN USED IN A RELEASED MIGRATION
+func SchemaCommitStoreV0(ctx context.Context, tx *pachsql.Tx) error {
```
albscui, line 13: Nit: I find it kinda weird to put these "versioned" code in our main code base. If this code is meant to be used only by a migration, I think it ought to be in the clusterstate package. As part of the postgres migration, I'm keeping all the schema related code in the clusterstate package.

This is a good time to achieve consensus on where to keep this code.

brendoncarroll: A pattern used in the storage layer, when testing, is to only run the storage migrations, so I'd like to keep `storagedb` separate.  I don't really have strong opinions about the commit store stuff.  That can go anywhere, just not in the PFS server package.

albscui: Personally, I prefer keeping this type of function in `clusterstate` package, i.e. where it is being called.

If we decide the other way, then I'll update my migration code to `coredb` and `pfsdb` packages and so on.

brendoncarroll: Does there need to be one way to do this?  I can see small things making sense in `clusterstate`, but large things being in their own package.  It looks like we already do that for some of the larger migrations.

>  i.e. where it is being called.

In the storage use case I described, the storage systems would also be callers, not just `clusterstate`.

The difference here sounds organizational, which I care less about.   This PR is not about organizing code, it's about changing the dependency graph, so we can import the `clusterstate` migration code, in the subsystems that need it to setup the database.

src/internal/clusterstate/v2.1.0.go
```diff
@@ -16,7 +16,7 @@ var state_2_1_0 migrations.State = state_2_0_0.
 		return enterpriseserver.DeleteEnterpriseConfigFromEtcd(ctx, env.EtcdClient)
 	}).
 	Apply("create pfs cache v1", func(ctx context.Context, env migrations.Env) error {
-		return fileset.CreatePostgresCacheV1(ctx, env.Tx)
+		return storagedb.SchemaFilesetCacheV1(ctx, env.Tx)
```
brycemcanally, line 19: I think all of these functions should include `Create` in the name since they are creating database objects. Also, they are all creating tables rather than schemas, so we should change that too.

src/internal/pfsdb/schema.go
brycemcanally: Here are a few options:
- Only include the schema modifications in clusterstate, which means all of the subsystems need to be migrated to test a particular subsystem.
- In clusterstate, reference schema modifications that are versioned in the relevant packages. This lets us potentially test a subsystem without migrating other subsystems, but the versions would need to be immutable and there cannot be ordering dependencies across the subsystem migrations. For example, if PPS and PFS migrations needed to be interleaved to be correct for some reason.
- Only include the schema modifications in clusterstate, but have a "compacted" version of the modifications in the relevant packages for testing. This is kind of the best of both worlds, but there is a risk that the compacted version of the schema does not match the schema that would exist after running all of the migrations that are created over time.

PR #8980, opened 2023-06-23
[CORE-1774] Add cluster defaults collection

robert-uhl:


src/internal/ppsdb/ppsdb.go
```diff
@@ -170,3 +183,14 @@ func CollectionsV0() []col.PostgresCollection {
 		col.NewPostgresCollection(jobsCollectionName, nil, nil, nil, jobsIndexes),
 	}
 }
+
+// CollectionsV2_7_0 returns a list of collections for postgres-initialization
+// purposes.  These collections are not usable for querying.
+//
+// DO NOT MODIFY THIS FUNCTION
+// IT HAS BEEN USED IN A RELEASED MIGRATION
+func CollectionsV2_7_0() []col.PostgresCollection {
```
albscui, line 192: I personally prefer to keep these in the clusterstate package, to keep them in lock-step with whatever migration that introduced it.

albscui (approved):

src/internal/ppsdb/ppsdb.go
robert-uhl: I believe that these are generally in the foodb packages in order to prevent needing to export names — this is the same pattern as with the other tables.

PR #8983, opened 2023-06-26
Configurable Pachd Probes

BOsterbuhr:
I have made changes to the Helm chart to expose the liveness, readiness, and startup probe settings. These settings are now configurable through the `values.yaml` file.

The changes include:
- Adding `readinessProbe`, `livenessProbe`, and `startupProbe` fields under the `pachd` section in the `values.yaml` file.
- Modifying the `deployment.yaml` file to use the values from the `values.yaml` file, or use the current default if a value is not specified.

These changes provide users with the flexibility to customize the probe settings based on their specific needs. This can be particularly useful in environments where the default probe settings may not be suitable, or during longer running upgrades.

### DIFF
--- a/etc/helm/pachyderm/templates/pachd/deployment.yaml
+++ b/etc/helm/pachyderm/templates/pachd/deployment.yaml
@@ -471,31 +471,31 @@ spec:
             command:
             - /pachd
             - --readiness
-          initialDelaySeconds: {{ .Values.pachd.readinessProbe.initialDelaySeconds | default 0 }}
-          timeoutSeconds: {{ .Values.pachd.readinessProbe.timeoutSeconds | default 1 }}
-          periodSeconds: {{ .Values.pachd.readinessProbe.periodSeconds | default 10 }}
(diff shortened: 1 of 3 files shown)


etc/helm/pachyderm/templates/pachd/deployment.yaml
```diff
@@ -471,20 +471,31 @@ spec:
             command:
             - /pachd
             - --readiness
+          initialDelaySeconds: {{ .Values.pachd.readinessProbe.initialDelaySeconds | default 0 }}
```
jrockway, line 474: I feel like we usually put these values in the default values.yaml file.  Helm's documentation says:

> In an actual chart, all static default values should live in the values.yaml, and should not be repeated using the default command (otherwise they would be redundant). However, the default command is perfect for computed values, which cannot be declared inside values.yaml.

These aren't computed, so should probably live in values.yaml and not in the code.

jrockway (changes requested):
One minor quibble

jrockway (approved):
LGTM

PR #8985, opened 2023-06-26
Unprivileged jupyter lab extension -- enable /pfs folder to link 

jerryharrow:
Another increment to simplify use of jupyter lab /pfs extension with HPC containers.
No change in behavior unless  `-e NONPRIV_CONTAINER=1` specified.

This increment allows the existing pachyderm/notebooks-user to be deployed in an HPC container (singularity/podman) without change.

* Add sudo configuration to enable the jupyter-extension to have write access to / to enable removal of /pfs folder when switching it to a link for non-privileged container support.
* When in NONPRIV_CONTAINER=1 mode, mount_server_client to remove /pfs directory so that it can replace it with the necessary softlink.

Direct use of pachyderm/notebooks-user can be run without `--privileged` or `--user=root`, but does still require `SYS_ADMIN` capability and   `--device /dev/fuse`.

```docker run -it -p 8888:8888  --cap-add SYS_ADMIN --device /dev/fuse   pachyderm/notebooks-user:dev jupyter lab ```

### CODE REVIEW

File: jupyter-extension/jupyterlab_pachyderm/mount_server_client.py [hunk 1]
```diff
@@ -114,7 +114,13 @@ async def _ensure_mount_server(self):
                     if os.path.exists(self.mount_dir) or os.path.islink(self.mount_dir):
                         if os.path.isdir(self.mount_dir): 
                             get_logger().debug(f"Removing dir {self.mount_dir}")
-                            os.rmdir(self.mount_dir)
+                            try: 
+                                os.rmdir(self.mount_dir)
+                            except OSError:
```
bbonenfant, line 119: Can we restrict this exception catch in any way since the code in this block is pretty heavy handed? Is the actual error being raised a [`PermissionError`](https://docs.python.org/3/library/exceptions.html#PermissionError)? Or if its a raw `OSError` can we we match the [errno](https://docs.python.org/3/library/errno.html) explicitly?

jerryharrow: True, but this only triggers a retry.   It will then repeat the removal and if it fails again it will bubble it up just like it would have originally.   So tightening up the check won't really change the resulting visible behavior.

jerryharrow: I verified that podman at least throws PermisisonError, so tightening up the except clause to that.

bbonenfant (approved):
Looks good to me. Sorry for the delay

jerryharrow:
Thanks @bbonenfant
I don't have privileges to submit, so if you are good with it, please submit.

PR #8987, opened 2023-06-26
[CORE-1790] Returned structured error details from FindCommits

robert-uhl:


src/internal/client/pfs.go
```diff
@@ -285,11 +285,11 @@ func (c APIClient) FindCommits(req *pfs.FindCommitsRequest) (*FindCommitsRespons
 	defer cf()
 	client, err := c.PfsAPIClient.FindCommits(ctx, req)
 	if err != nil {
-		return nil, grpcutil.ScrubGRPC(err)
```
brendoncarroll, line 288: 👍

src/server/pfs/pfs.go
```diff
@@ -270,8 +272,17 @@ func (e ErrCommitNotFinished) Error() string {
 	return fmt.Sprintf("commit %v not finished", e.Commit)
 }
 
-func (e ErrCommitNotFinished) GRPCStatus() *status.Status {
-	return status.New(codes.Unavailable, e.Error())
+func (err ErrCommitNotFinished) GRPCStatus() *status.Status {
+	s, sErr := status.New(codes.Unavailable, fmt.Sprintf("commit %v not finished", err.Commit)).WithDetails(&epb.ResourceInfo{
```
brendoncarroll, line 276: https://grpc.github.io/grpc/core/md_doc_statuscodes.html

It seems like we should return `FAILED_PRECONDITION` when reading from open commits, and `UNAVAILABLE` when reading from finishing commits.

I don't know if we disambiguate, so this is probably fine.

brendoncarroll (approved):

src/server/pfs/pfs.go
robert-uhl: We don’t currently. I’ve opened CORE-1799 to consider disambiguation.

PR #8988, opened 2023-06-26
Allow WithTx to Use Contexts

FahadBSyed:


albscui (approved):

src/internal/log/context.go
```diff
@@ -17,6 +17,8 @@ import (
 
 type pachydermLogger struct{}
 
+type Context context.Context
```
brendoncarroll, line 20: Remove

src/internal/ppsutil/util.go
```diff
@@ -170,7 +170,7 @@ func SetPipelineState(ctx context.Context, db *pachsql.DB, pipelinesCollection c
 	logSetPipelineState(ctx, pipeline, from, to, reason)
 	var resultMessage string
 	var warn bool
-	err := dbutil.WithTx(ctx, db, func(sqlTx *pachsql.Tx) error {
+	err := dbutil.WithTx(ctx, db, func(_ context.Context, sqlTx *pachsql.Tx) error {
```
brendoncarroll, line 173: 👍

src/server/pfs/iface.go
```diff
@@ -13,7 +15,7 @@ type APIServer interface {
 
 	NewPropagater(*txncontext.TransactionContext) txncontext.PfsPropagater
 
-	CreateRepoInTransaction(*txncontext.TransactionContext, *pfs_client.CreateRepoRequest) error
+	CreateRepoInTransaction(context.Context, *txncontext.TransactionContext, *pfs_client.CreateRepoRequest) error
```
brendoncarroll, line 18: Doesn't the `TransactionContext` have a `context.Context` in it already?

src/server/pfs/server/driver.go
```diff
@@ -122,7 +123,7 @@ func newDriver(env Env) (*driver, error) {
 	return d, nil
 }
 
-func (d *driver) createRepo(txnCtx *txncontext.TransactionContext, repo *pfs.Repo, description string, update bool) error {
+func (d *driver) createRepo(ctx context.Context, txnCtx *txncontext.TransactionContext, repo *pfs.Repo, description string, update bool) error {
```
brendoncarroll, line 126: See my previous comment.  I don't think we need to add another context to all of these.  We should keep using the one from the `TransactionContext`

src/server/pfs/iface.go
albscui: We recently removed it.

PR #8989, opened 2023-06-26
[CORE-1709] Implement GetClusterDefaults

robert-uhl:
Add GetClusterDefaults. If there are none, returns empty defaults. Implements merger of the empty pipeline spec (also defined in this PR) with the cluster defaults.


src/pps/pps.go
```diff
@@ -57,3 +60,46 @@ func DataField(data []*InputFile) zap.Field {
 	}
 	return zap.Objects("data", data)
 }
+
+type PipelineSpec struct {
```
jrockway, line 64: I would really like to avoid a third copy of this data structure in the codebase.  What are our options?

robert-uhl: The current design is to go through a default/deprecate/delete process, ultimately leaving two similar data structures in the codebase: the user-facing JSON and the Protobuf `PipelineInfo_Details`. After deprecation, `CreatePipelineRequest` will be a thin Protobuf wrapping `details_json`.

`details_json` can’t be a `PipelineInfo_Details`, because that message is missing certain fields such as `pipeline`; nor can it be a `PipelineInfo`, because that has nested structure which is not congruent with the current request.

An option would be to create a new wire-compatible `CreatePipelineRequest` and rename the old one to `PipelineSpec`. The new would wrap the old, and `PipelineInfo` could be updated to contain a `PipelineSpec`. The biggest drawback to this is that it limits the evolution of the user-facing JSON auto-generated from `PipelineSpec` to whatever `protoc-gen-go` allows. That limitation is why the spec currently indicates the pipeline as `{"pipeline": {"project": {"name": "myProject"}}, "name": "myPipeline"}` instead of `{"project": "myProject", "pipeline": "myPipeline"}` and why e.g. the `Input` message does not use `oneof` (if it did, the user-visible JSON would be unusably complex). OTOH, the big pro is that we would have a single version of the structure going forward.

src/server/pps/server/api_server.go
```diff
@@ -3559,3 +3560,80 @@ func newMessageFilterFunc(jqFilter string, projects []*pfs.Project) (func(contex
...
+	}
+	result, err := json.Marshal(mergePatch(targetObject, patchObject))
+	if err != nil {
+		return "", errors.Wrap(err, "could not marshal merge patch result")
+	}
+	return string(result), nil
+}
+
+// mergePatch implements the RFC 7396 algorithm.  To quote the RFC “If the patch
+// is anything other than an object, the result will always be to replace the
```
jrockway, line 3613: Why do we allow anything other than an object?  Why not unmarshal into `pps.PipelineSpec` instead of `any`?

robert-uhl: > Why do we allow anything other than an object?

It’s the recursive algorithm defined by the RFC, so it doesn’t only get called on objects, but also on their values (at `targetMap[name] = mergePatch(targetMap[name], value)`), which may be of any type.

>Why not unmarshal into `pps.PipelineSpec` instead of `any`?

It may be called on a JSON object which represents a pipeline spec, but it may also be called on a JSON object which represents a resource request, or a toleration … or it may be called on a string/bool/number/array/nil.

I could implement this using reflection instead, but I don’t know if that’s a huge advantage.

robert-uhl: E.g. with a JSON merge, `mergePatch({"resource_limits": {"cpu": "2m"}, "resource_requests": {"cpu": "1m"}, {"resource_request": nil}) → {"resource_limits": {"cpu": "2m"}}`, but with a Go-typed, reflection-based approach it would yield `{}` because `"resource_limits"}` was omitted from the patch spec.

jrockway: I think the embedded messages that need a custom type to detect presence then:

```go
type ResourceRequests struct {
   Present bool `json:"_"`
   CPU int `json:"cpu"`
}
```
func (req *ResourceRequests) UnmarshalJSON(buf []byte) error {
   if len(buf) == 0 {
      return nil
   }
   req.Present = true
   var result struct { CPU int `json:"cpu"` }
   if err := json.Unmarshal(buf, &result); err != nil {
       return errors.Wrap(err, "unmarshal resource requests")
   }
   req.CPU = result.CPU
   return nil
}
func (req *ResourceRequests) MarshalJSON() ([]byte, error) {
    if !req.Present { return nil, nil }
    json.Marshal(...)
    ...
}
```
This is a lot of boilerplate to write, but allows invalid requests to be aborted early with a descriptive message.

File: src/pps/pps.go (hunk 1 above)
Comment (robert-uhl) [reply src/pps/pps.go]: Are you advocating this:

```
message CreatePipelineRequest { /* old request */ }

message CreatePipelineV2Request {
  string pipeline_spec = 1;  // JSON-encoded spec
  bool update = 2;
  bool reprocess = 3;
  bool dry_run = 4;
}

message PipelineInfo {
  ⋮
  reserved 12;
  reserved "details";
  PipelineSpec spec = 14;
  string user_spec = 14; // JSON-encoded spec
}

message PipelineSpec { … }

service API {
  rpc CreatePipeline(CreatePipelineRequest) returns (google.protobuf.Empty) {}
  rpc CreatePipelineV2(CreatePipelineV2Request) returns (CreatePipelineV2Response) {}
}
```

File: src/server/pps/server/api_server.go (hunk 2 above)
Comment (robert-uhl) [reply src/server/pps/server/api_server.go:3613]: It turns out that your initial hunch over at https://github.com/pachyderm/pachyderm/pull/9010#discussion_r1247209379 was correct: this need to be `protojson.Marshal`able, which means that the struct must be a protobuf message, which means that it’s not possible to add `Present` fields to it.

File: src/pps/pps.go (hunk 1 above)
Comment (jrockway) [reply src/pps/pps.go]: CreatePipelineV2Request.pipeline_spec should be the actual object, I think.  I understand why we want to involve JSON for the transport layer, but I'd like to hold off on that if possible.  (Really, with templating involved, you aren't sending a filled-out proto message, but rather a program that controls how the spec is filled out. But if we can pretend to not have JSON inside protos for this use case, that would be ideal.)

Comment (robert-uhl) [reply src/pps/pps.go]: > CreatePipelineV2Request.pipeline_spec should be the actual object, I think.

With proto3’s addition of field presence, I think that it is indeed possible to store the user-provided spec (as you note, really a program) as a protobuf. I’ll work up a quick proof-of-concept to see how well it works, but I think it’ll be alright.

My proposal is this:

```
message CreatePipelineRequest {
  bool update = 6;
  bool reprocess = 15;
  string pipeline_spec = 36; // or PipelineSpec pipeline_spec — either way
  bool dry_run = 37;
  reserved 1 to 5, 7 to 14, 16 to 35;
  reserved "pipeline", "tf_job", "transform", …,  "tolerations";
}

message PipelineInfo {
  ⋮
  reserved 12;
  reserved "details";
  PipelineSpec spec = 14;
  string user_spec = 15; // or PipelineSpec user_spec — either way
}

message PipelineSpec { … }

service API {
  rpc CreatePipeline(CreatePipelineRequest) returns (CreatePipelineResponse) {}
}
```

That would be in v2.10. In v2.9, the old Create Pipeline Request fields (reserved above) would be defined as optional, and would result in a runtime error message when present. In v2.8 (now the first version with cluster defaults), the old fields remain, but are only used if the pipeline spec is omitted. Likewise, in v2.8 the old Pipeline Info details field would be replaced with the spec, either as part of CORE-1706 or as a separate story.

I think the end result is functionally the same as your proposal, but with fewer messages and RPCs. OTOH, my proposal requires change over time, and in v2.8 & v2.9 does involve multiple copies of roughly the same message. I think that we are pretty certain to actually make the future changes because there project defaults are already scheduled, but there is a risk of leaving the evolution half-implemented.

src/server/pps/server/api_server.go
jrockway: We talked about this in person and the any type seems fine.

src/pps/pps.go
robert-uhl: We discussed this in person and in chat. There will be a new CreatePipelineV2 RPC. It will take a JSON-marshalled CreatePipelineRequest, which will be merged into the cluster-default CreatePipelineRequest to create the effective pipeline request. This reuses the existing message and removes a third version.

Details are documented in the tech design.

jrockway (approved):

PR #8990, opened 2023-06-26
Run upgrades in a single transaction

acohen4:


src/internal/migrations/migrations.go
```diff
@@ -87,12 +88,16 @@ func ApplyMigrations(ctx context.Context, db *pachsql.DB, baseEnv Env, state Sta
...
+		return errors.EnsureStack(err)
+	}
 	for _, state := range CollectStates(make([]State, 0, state.n+1), state) {
-		if err := applyMigration(ctx, db, baseEnv, state); err != nil {
+		if err := applyMigration(ctx, tx, baseEnv, state); err != nil {
 			return err
 		}
 	}
-	return nil
+	return errors.EnsureStack(tx.Commit())
```
jrockway, line 100: All I would suggest is that we log when Commit fails or the tx rolls back, so the logs are like:

```
applying migration 21: already applied
applying migration 22: ok
applying migration 23: failed due to badness
migrations 22, 23 rolled back, exiting
```

brendoncarroll:
I don't think this is going to work.  It looks like we will try to get the lock on `migrations` multiple times.  There is that ApplyMigrationTx method down below, which seems to actually apply the migration?

brendoncarroll (approved):
Make `ApplyMigrationTx` private, then LGTM.

acohen4:
> Make `ApplyMigrationTx` private, then LGTM.

it's exposed due to `pachctl misc test-migration`. We may want to remove it as a follow up

PR #9006, opened 2023-06-28
[Python-SDK] Restrict file operations within a TXN

bbonenfant:
[INT-1004]

[INT-1004]: https://pachyderm.atlassian.net/browse/INT-1004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-sdk/tests/test_transaction.py
+++ b/python-sdk/tests/test_transaction.py
@@ -3,6 +3,7 @@ import grpc
 from tests.fixtures import *
 
 from pachyderm_sdk.api import pfs, transaction
+from pachyderm_sdk.errors import InvalidTransactionOperation
 
 
 class TestTransaction:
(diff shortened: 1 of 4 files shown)


python-sdk/tests/test_transaction.py
```diff
@@ -84,3 +85,19 @@ def test_delete_transaction(client: TestClient, default_project: bool):
...
+        """Ensure that file operations """
+        repo = client.new_repo(default_project)
+        branch = pfs.Branch(repo=repo, name="master")
+        test_file = pfs.File.from_uri(f"{repo.project.name}/{repo.name}@master:/file.dat")
+
+        with client.transaction.transaction() as _txn:
+            repo_txn = client.new_repo(default_project)
+            with client.pfs.commit(branch=branch) as commit:
+                with pytest.raises(InvalidTransactionOperation):
+                    commit.put_file_from_bytes(path=test_file.path, data=b"hello world")
```
kevyang, line 100: up to you whether or not you want to implement this, but coverage for the other write operations seems lacking. it's going to be tedious to get them all but maybe adding copy/delete as well as a put would be a good idea

kevyang (approved):

PR #9007, opened 2023-06-28
python-sdk - drop certifi package

kevyang:
now that the [root certificate issue](https://github.com/grpc/grpc/issues/27727) in grpc has been patched, we no longer need certifi. this change removes that package and uses the grpc default for root certs.

[INT-901](https://pachyderm.atlassian.net/browse/INT-901)

### DIFF
--- a/python-sdk/pachyderm_sdk/client.py
+++ b/python-sdk/pachyderm_sdk/client.py
@@ -85,17 +85,14 @@ class Client:
             auth_token = os.environ.get(AUTH_TOKEN_ENV)
 
         tls = tls or (root_certs is not None)
-        if tls and root_certs is None:
-            # load default certs if none are specified
-            import certifi
-
-            with open(certifi.where(), "rb") as f:
-                root_certs = f.read()
-
         self.address = "{}:{}".format(host, port)
         self.root_certs = root_certs
+        self.tls = tls
         channel = _create_channel(
-            self.address, self.root_certs, options=GRPC_CHANNEL_OPTIONS
+            address=self.address,
+            root_certs=self.root_certs,
+            options=GRPC_CHANNEL_OPTIONS,
+            tls=self.tls
         )
 
         self._auth_token = auth_token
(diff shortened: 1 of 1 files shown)


python-sdk/pachyderm_sdk/client.py
```diff
@@ -353,8 +351,10 @@ def _create_channel(
     address: str,
     root_certs: Optional[bytes],
     options: MetadataType,
+    tls: bool = False,
```
bbonenfant, line 354: I don't love having both `root_certs` and `tls` as separate arguments, but I'm not sure of a better way to do this at the moment. Presumable if `root_certs!=None and tls=False` we still want to use the root certs in a secure channel? That appears to be the default behavior in the client

python-sdk/pachyderm_sdk/client.py
```diff
@@ -85,17 +85,14 @@ def __init__(
             auth_token = os.environ.get(AUTH_TOKEN_ENV)
 
         tls = tls or (root_certs is not None)
-        if tls and root_certs is None:
```
bbonenfant, line 88: Can't comment on the line, but you'll want to update the class docstring to remove the reference to certifi

python-sdk/pachyderm_sdk/client.py
bbonenfant, line 1: You'll also want to remove certifi as a dependency within the `pyproject.toml` file.

python-sdk/pachyderm_sdk/client.py
kevyang: yeah, i was also trying to find a way around having both but couldn't come up with a good way to do that. we essentially have 3 options we want to accommodate: 1) is insecure_channel via tls == false, 2) is secure_channel with default root_certs where tls == true and root_certs == None, and 3) is secure channel with specified root certs.

kevyang:
Closing this since after further investigation it seems like grpc is still not able to load the default certs.

PR #9008, opened 2023-06-28
[Python-SDK] Fixup extension methods

bbonenfant:
[INT-980]

[INT-980]: https://pachyderm.atlassian.net/browse/INT-980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-sdk/pachyderm_sdk/api/pfs/extension.py
+++ b/python-sdk/pachyderm_sdk/api/pfs/extension.py
@@ -255,7 +255,7 @@ class OpenCommit(ClosedCommit):
         self._stub.put_file_from_file(commit=self, path=path, file=file, append=append)
         return File(commit=self._commit, path=path)
 
-    def copy_file(self, *, src: "File", dst: str, append: bool = True) -> "File":
+    def copy_file(self, *, src: "File", dst: str, append: bool = False) -> "File":
         """Copies a file within PFS
 
         Parameters
@@ -557,7 +557,7 @@ class ApiStub(_GeneratedApiStub):
 
     @transaction_incompatible
     def copy_file(
-        self, *, commit: "Commit", src: "File", dst: str, append: bool = True
+        self, *, commit: "Commit", src: "File", dst: str, append: bool = False
     ) -> Empty:
         """Copies a file within PFS
 
--- a/python-sdk/pachyderm_sdk/api/pfs/extension.py
+++ b/python-sdk/pachyderm_sdk/api/pfs/extension.py
@@ -81,7 +81,7 @@ class ClosedCommit(Commit):
         """
         return self._stub.wait_commit(self)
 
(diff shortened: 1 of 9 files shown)


python-sdk/pachyderm_sdk/api/worker/extension.py
```diff
@@ -49,5 +49,4 @@ def batch_datum(self) -> ContextManager:
             yield
         except Exception as error:
             self.__error = repr(error)
-            # TODO: Probably want better logging here than a print statement.
```
bbonenfant, line 52: Will be addressed by INT-989

python-sdk/tests/test_pps.py
```diff
@@ -108,7 +108,6 @@ class TestUnitDatums:
 
     @staticmethod
     def test_datums(client: TestClient, default_project: bool):
-        # TODO: Why is this test sometimes slow?
```
bbonenfant, line 111: Will be addressed by [INT-1040]

python-sdk/tests/fixtures.py
```diff
@@ -127,7 +124,7 @@ def _generate_name(self) -> str:
                 .replace("/", "-")
                 .replace(":", "-")
                 .replace(".py", "")
-        )[:40]  # TODO: Make this the maximum it can be.
+        )[:40]
```
chainlink, line 127: Nit: add a comment as to where this magic number comes from on the core pach side

python-sdk/pachyderm_sdk/api/pfs/extension.py
```diff
@@ -231,7 +255,7 @@ def put_file_from_file(
         self._stub.put_file_from_file(commit=self, path=path, file=file, append=append)
         return File(commit=self._commit, path=path)
 
-    def copy_file(self, *, src: "File", dst: str, append: bool = True) -> "File":
+    def copy_file(self, *, src: "File", dst: str, append: bool = False) -> "File":
```
chainlink, line 258: Strange that the default for copy was append

bbonenfant: This was a mistake when moving this from python_pachyderm. It's False there, I'm not sure how these got flipped.

python-sdk/tests/fixtures.py
chainlink: @bbonenfant I think I saw a thread about this in slack. Can you add the note here?

chainlink (approved):
Pre-approving pending the one last comment todo.
Next actions:
- PFS Review of byte level manipulation (+ adding tests to verify the bytes?)
- Follow on Jira Tickets

PR #9009, opened 2023-06-28
makes make proto mac friendly

molinamelendezj:


jrockway:
What problem does BSD tar have with our data?

molinamelendezj:
> What problem does BSD tar have with our data?

I get a lot of warnings from mac files `tar: Ignoring unknown extended header keyword 'LIBARCHIVE.xattr.com.apple.provenance'` and failures since it picks up apple files like ._pps.proto

for example: 

`pps/._pps.proto:1:9: Expected top-level statement (e.g. "message").` 


so using the GNU version on mac allows me to generate protos. the tar I send to the linux proto image is happy.

etc/proto/build.sh
```diff
@@ -1,7 +1,13 @@
 #!/bin/bash
 # This script build protos using the pachyderm_proto image.
 
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  TAR=gtar # brew install gnu-tar if on mac
+else
+  TAR=tar
+fi
+
 read -ra proto_files < <(find src -name "*.proto" -print0 | xargs -0)
```
jrockway, line 10: Does `find src -regex '.*/[^/.]+[.]proto$' -print0` fix the problem?

jrockway:
Also: https://stackoverflow.com/a/74373784

etc/proto/build.sh
molinamelendezj: @jrockway hmm no i get:

```
tar: no files or directories specified
+ tar -C /go/src/github.com/pachyderm/pachyderm -xf /dev/stdin
tar: This does not look like a tar archive
tar: Exiting with failure status due to previous errors
```

I also tried using `tar`'s `--exclude` flag to exclude those `._ ` files but that didn't work ether. if we don't want to have `gtar` in the code base I can close this. ill just use it locally when I need to regenerate. or is your concern a different one?

I think other mac devs haven't hit this because they use a linux vm to do dev. 🤣 maybe I should just use a linux vm. I've made it so far not needing to 🤣

molinamelendezj: nope tried that to still get with and without the `--exclude .*` 

```
tar: This does not look like a tar archive
tar: Exiting with failure status due to previous errors
```

jrockway: I think something else is going on then; does the `find | xargs` pipeline return anything when you run it alone?

Like this definitely works with GNU find:

```
$ find src -regex '.*/[^/.]+[.]proto$'
src/enterprise/enterprise.proto
src/protoextensions/log.proto
src/transaction/transaction.proto
src/server/worker/pipeline/transform/transform.proto
src/server/worker/common/common.proto
src/server/worker/datum/datum.proto
src/server/pfs/server/pfsserver.proto
src/license/license.proto
src/pps/pps.proto
src/auth/auth.proto
src/worker/worker.proto
src/proxy/proxy.proto
src/internal/pfsload/pfsload.proto
src/internal/tracing/extended/extended_trace.proto
src/internal/collection/test.proto
src/internal/ppsload/ppsload.proto
src/internal/storage/fileset/index/index.proto
src/internal/storage/fileset/fileset.proto
src/internal/storage/chunk/chunk.proto
src/internal/metrics/metrics.proto
src/internal/task/task.proto
src/internal/clusterstate/v2.5.0/commit_info.proto
src/internal/config/config.proto
src/version/versionpb/version.proto
src/debug/debug.proto
src/pfs/pfs.proto
src/admin/admin.proto
src/task/task.proto
src/identity/identity.proto
```

jrockway (approved):
Try and do this some other way, but if you can't, merge it.

etc/proto/build.sh
molinamelendezj: no, using that regex doesn't return anything on my mac. that explains the tar not being valid.

PR #9010, opened 2023-06-28
[CORE-1710] Add SetClusterDefaults

robert-uhl:
This adds `SetClusterDefaults`; it also merges in the latest `master` branch.

Of note: it does not propagate changes of defaults to pipelines; that work is reserved for CORE-1708.


src/server/pps/server/api_server.go
```diff
@@ -3575,3 +3577,105 @@ func newMessageFilterFunc(jqFilter string, projects []*pfs.Project) (func(contex
...
+	var d = json.NewDecoder(strings.NewReader(req.GetClusterDefaults().GetDetailsJson()))
+	d.DisallowUnknownFields()
+	if err := d.Decode(&pps.PipelineSpec{}); err != nil {
+		s, err := status.New(codes.InvalidArgument, "invalid pipeline details").WithDetails(&errdetails.BadRequest{
+			FieldViolations: []*errdetails.BadRequest_FieldViolation{
+				{Field: "cluster_defaults.details_json", Description: err.Error()},
+			},
+		})
+		if err != nil {
+			log.Error(ctx, "could not add bad-request details", zap.Error(err))
```
jrockway, line 3659: I don't like the pattern of error handling error handling.  We should make a StatusWithDetails function that doesn't error.  WithDetails errors if the status is codes.OK; we should just ignore that one.  It additionally errors because it uses the old proto library and calls ptypes.MarshalAny instead of anypb.New.  anypb.New errors when the message is nil or, after MANY levels of indirection, when it can't write to the provided buffer.  The provided buffer is a []byte that it appends to, so it seems like that can never happen.  I'd just restrict the type that can be passed to your WithDetails wrapper, and pass a concrete instance instead of a pointer.  Then no errors can ever occur, and you don't have to write error handling error handling.

src/server/pps/server/api_server_test.go
```diff
@@ -335,3 +338,65 @@ func TestUpdatePipelineInputBranch(t *testing.T) {
...
 	))
 }
+
+func TestGetClusterDefaults(t *testing.T) {
+	ctx := pctx.TestContext(t)
+	env := realenv.NewRealEnv(ctx, t, dockertestenv.NewTestDBConfig(t))
+	resp, err := env.PPSServer.GetClusterDefaults(ctx, &pps.GetClusterDefaultsRequest{})
+	require.NoError(t, err, "GetClusterDefaults failed")
+	require.NotNil(t, resp.ClusterDefaults)
+	require.NotEqual(t, "", resp.ClusterDefaults.DetailsJson, "details must not be empty")
```
jrockway, line 348: I'd test that these can be unmarshaled.

src/server/pps/server/merge_test.go
```diff
@@ -0,0 +1,69 @@
...
+	var testCases = []struct{ target, patch, result string }{
+		// first example in RFC 7396
+		{`{"a": "b", "c": {"d": "e", "f": "g"}}`, `{"a":"z", "c": {"f": null}}`, `{"a": "z", "c": {"d": "e"}}`},
+
+		// second example in RFC 7396
+		{
+			`{"title": "Goodbye!", "author": {"givenName": "John", "familyName": "Doe"}, "tags":["example", "sample"], "content": "This will be unchanged"}`, ` {"title": "Hello!", "phoneNumber": "+01-123-456-7890", "author": {"familyName": null}, "tags": ["example"]}`, `{"title": "Hello!", "author": {"givenName": "John"}, "tags": ["example"], "content": "This will be unchanged", "phoneNumber": "+01-123-456-7890"}`,
+		},
+
+		// test cases from Appendix A of RFC 7396
```
jrockway, line 19: 👍 👍 👍 👍

src/server/pps/server/api_server.go
```diff
@@ -3575,3 +3577,105 @@ func newMessageFilterFunc(jqFilter string, projects []*pfs.Project) (func(contex
...
+		return "", errors.Wrap(err, "could not unmarshal patch JSON")
+	}
+	result, err := json.Marshal(mergePatch(targetObject, patchObject))
+	if err != nil {
+		return "", errors.Wrap(err, "could not marshal merge patch result")
+	}
+	return string(result), nil
+}
+
+func mergePatch(target, patch any) any {
```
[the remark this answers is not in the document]
robert-uhl: > I think we want to be stricter about the types of the arguments. At the very least, map[string]any instead of any.

That doesn’t work because the RFC 7396 is recursive and defined for any type, so it will end up being called with e.g. strings, integers and booleans as both targets and patches.

src/pps/pps.go
```diff
@@ -57,3 +59,46 @@ func DataField(data []*InputFile) zap.Field {
 	}
 	return zap.Objects("data", data)
 }
+
```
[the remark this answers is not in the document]
robert-uhl: It is unfortunate that `CreatePipelineRequest` and `PipelineDetails` are almost but not quite congruent. [ppsutil.PipelineReqFromInfo](https://github.com/pachyderm/pachyderm/blob/rau/core-1710-add-set-cluster-defaults/src/internal/ppsutil/util.go#L253-L282) is a curious function.

src/server/pps/server/api_server.go
```diff
@@ -3575,3 +3577,105 @@ func newMessageFilterFunc(jqFilter string, projects []*pfs.Project) (func(contex
...
 	}, nil
 }
+
+// This exists in case we end up not ommitting any empty values when marshalling
+// a spec.
+var emptyPipelineSpecJSON string
+
+func init() {
+	var spec pps.PipelineSpec
+	b, err := json.Marshal(spec)
```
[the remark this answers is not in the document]
robert-uhl: It’s a non-protobuf Go structure with a JSON mapping which provides the schema for the user-submitted pipeline spec.

src/server/pps/server/api_server.go
robert-uhl: How do you feel about the new `badRequest` and `unknownError` functions, which always return an error? They do still catch the errors from `WithDetails`, but I think that avoids relying on undocumented internals of other libraries not to change.

src/server/pps/server/api_server.go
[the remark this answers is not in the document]
robert-uhl: But you’re correct: pps.PipelineSpec needs to be a protobuf for unmarshalling to work. I’ll get cracking on this in CORE-1709/PR #8989.

src/server/pps/server/api_server.go
jrockway: I'm just saying to abstract this away somehow; the RPC handler doesn't need to know that error-creating functions can return errors.  It just complicates the "meat" of the code with an irrelevant detail that essentially can't happen.  I think restricting the API to 2 types out of the 16 or so error codes is unnecessarily restrictive.  `errors.NewRPC(code, details, proto.Message)`?

zmajeed (approved):

PR #9012, opened 2023-06-29
adds job scoped session token exposed to user code as PACH_TOKEN

molinamelendezj:


src/server/worker/pipeline/transform/worker.go
```diff
@@ -389,7 +389,7 @@ func forEachDatum(ctx context.Context, driver driver.Driver, baseLogger logs.Tag
 			meta.ImageId = userImageID
 			inputs := meta.Inputs
 			logger := baseLogger.WithData(inputs)
-			env := driver.UserCodeEnv(logger.JobID(), task.OutputCommit, inputs, driver.PipelineInfo().GetAuthToken())
+			env := driver.UserCodeEnv(logger.JobID(), task.OutputCommit, inputs, driver.PipelineInfo().GetAuthToken(), driver)
```
molinamelendezj, line 392: @acohen4 having some trouble understanding how we issue a token, this is wrong but hoping you could point me in the right direction. jobs are created by the driver so would we need to reach the api and get a token issued?

acohen4: the token should be issued by PPS. When a Job gets created in propagateJobs(), we should call out to the Auth API Server to create another auth token for the pipeline and set it as the job token

molinamelendezj: @acohen4 thanks, okay i create the token now. so last question here, driver.UserCodeEnv here is called at forEachDatum() do we need a token here to?

src/internal/ppsutil/util.go
```diff
@@ -336,6 +338,13 @@ func FinishJob(pachClient *client.APIClient, jobInfo *pps.JobInfo, state pps.Job
 		}); err != nil {
 			return errors.EnsureStack(err)
 		}
+		if jobInfo.AuthToken != "" {
```
acohen4, line 341: should we set `jobInfo.AuthToken = ""` here instead of in UpdateJobState()?

src/server/worker/driver/driver.go
```diff
@@ -525,6 +535,7 @@ func (d *driver) UserCodeEnv(
 	outputCommit *pfs.Commit,
 	inputs []*common.Input,
 	authToken string,
```
acohen4, line 537: I think authToken & pachToken here is redundant. We should only use the Job's Auth Token now in the UserCodeEnv instead of both the Job's and the Pipeline's Token

src/server/pachyderm_test.go
```diff
@@ -1337,6 +1337,56 @@ func TestRunPipeline(t *testing.T) {
...
+
+	pipeline := tu.UniqueString("pipeline")
+	require.NoError(t, rc.CreatePipeline(pfs.DefaultProjectName,
+		pipeline,
+		"",
+		[]string{"bash"},
+		[]string{
+			"echo PACH_TOKEN=$PACH_TOKEN",
+			"echo $PACH_TOKEN | pachctl auth use-auth-token",
+			"pachctl config update context --pachd-address 'grpc://pachd.test-cluster-1.svc.cluster.local:30660'",
```
acohen4, line 1365: this won't always run on test-cluster-1

acohen4: you should be able to run this against localhost - it can talk to the storage container

src/server/worker/pipeline/transform/worker.go
```diff
@@ -374,6 +374,10 @@ type datumCallback = func(ctx context.Context, logger logs.TaggedLogger, env []s
 
 // TODO: There are way too many parameters here. Consider grouping them into a reasonable struct or storing some of these in the driver.
 func forEachDatum(ctx context.Context, driver driver.Driver, baseLogger logs.TaggedLogger, task *DatumSetTask, status *Status, cacheClient *pfssync.CacheClient, di datum.Iterator, setOpts []datum.SetOption, cb datumCallback) error {
+	jobInfo, err := driver.GetJobInfo(task.Job)
+	if err != nil {
+		return errors.Wrap(err, "could not get job info from datum set task job")
```
acohen4, line 379: include job in error - something like `errors.Wrapf(err, "load datum set's job info for job %q", task.Job.String())`

src/server/worker/driver/driver.go
```diff
@@ -581,6 +592,10 @@ func (d *driver) UserCodeEnv(
 		}
 	}
 
+	if pachToken != "" {
+		result = append(result, "PACH_TOKEN="+pachToken)
+	}
+
```
acohen4, line 598: nit - remove new lines in-function

src/server/pachyderm_test.go
```diff
@@ -1337,6 +1337,56 @@ func TestRunPipeline(t *testing.T) {
...
+	require.NoError(t, rc.PutFile(commit, "file", strings.NewReader("foo\n"), client.WithAppendPutFile()))
+	require.NoError(t, rc.FinishCommit(pfs.DefaultProjectName, dataRepo, commit.Branch.Name, commit.Id))
+
+	pipeline := tu.UniqueString("pipeline")
+	require.NoError(t, rc.CreatePipeline(pfs.DefaultProjectName,
+		pipeline,
+		"",
+		[]string{"bash"},
+		[]string{
+			"echo PACH_TOKEN=$PACH_TOKEN",
```
acohen4, line 1363: can remove this line?

src/server/pachyderm_test.go
```diff
@@ -1337,6 +1337,56 @@ func TestRunPipeline(t *testing.T) {
...
+			"echo $PACH_TOKEN | pachctl auth use-auth-token",
+			"pachctl config update context --pachd-address 'grpc://pachd.test-cluster-1.svc.cluster.local:30660'",
+			"pachctl auth whoami",
```
acohen4, line 1366: maybe pipe this result into `/pfs/out/result.txt` and assert on JOB_SUCCESS that the contents of the file are the `pipeline:default/{pipeline}`?

acohen4 (approved):
LGTM!

PR #9015, opened 2023-06-29
Migrate Repos from collections to relational model

albscui:


src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,399 @@
...
+
+// Error satisfies the error interface.
+func (err ErrRepoNotFound) Error() string {
+	if id := err.ID; id != 0 {
+		return fmt.Sprintf("repo id=%d not found", id)
+	}
+	return fmt.Sprintf("repo %s/%s.%s not found", err.Project, err.Name, err.Type)
+}
+
+func (err ErrRepoNotFound) Is(other error) bool {
```
albscui, line 51: I'm a  bit confused by the intent of this function. And how we check for errors here in general.

This `ErrRepoNotFound.Is` is supposed to customize the behaviour of `errors.Is`, but right now it's basically returning true for any error that has the type `ErrRepoNotFound`. However, the point of `errors.Is` is to be value aware.

For example
```
err1 := ErrRepoNotFound{Name: "repo1"}
err2 := ErrRepoNotFound{Name: "repo2"}

// without customization
errors.Is(err1, err2) // returns false
// however
err1.Is(err2) // returns true
```

Based how you use `ErrRepoNotFound.Is` in tests, it seems like you care about the value after all. I think there's a simpler way to do this with native Go library

```
err := ErrRepoNotFound{Name: "repo1"}
err2 := ErrRepoNotFound{Name: "repo1"}
err3 := ErrRepoNotFound{Name: "repo2"}

// to compare two errors by value
fmt.Println(errors.Is(err, err2))               // false
fmt.Println(errors.Is(err, err3))               // false
// to check an error's type
fmt.Println(errors.As(err, &ErrRepoNotFound{})) // true
```

Lastly, a further argument for using `errors.Is` and `errors.As` is to handle wrapped errors.
For example
```
err := fmt.Errorf("adding context: %w", ErrRepoNotFound{Name: "repo1"})
_, ok := err.(ErrRepoNotFound) // ok == false
// but
errors.As(err, &ErrRepoNotFound{}) // returns true
```

albscui (commented):
Still working on reviewing the rest, but lets start with some comments about error handling.

src/server/pfs/server/driver_fsck.go
```diff
@@ -365,22 +366,28 @@ func (d *driver) fsck(ctx context.Context, fix bool, cb func(*pfs.FsckResponse)
 	branchInfos := make(map[string]*pfs.BranchInfo)
 	commitInfos := make(map[string]*pfs.CommitInfo)
 	repoInfos := make(map[string]*pfs.RepoInfo)
-	repoInfo := &pfs.RepoInfo{}
-	if err := d.repos.ReadOnly(ctx).List(repoInfo, col.DefaultOptions(), func(string) error {
```
albscui, line 369: Can we have a brief explanation here for the original intent for read only, and why we are switching to transactional.

src/internal/pfsdb/repos.go
FahadBSyed: I think I got this from a suggestion on projects and I just copy and pasted the same structure

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,388 @@
...
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
+	"github.com/pachyderm/pachyderm/v2/src/internal/stream"
+	"github.com/pachyderm/pachyderm/v2/src/pfs"
+)
+
+const (
+	getRepoAndBranches = "SELECT repo.id, repo.name, repo.type, repo.project_id, " +
+		"repo.description, array_agg(branch.proto) AS branches, repo.created_at, repo.updated_at FROM pfs.repos repo " +
+		"LEFT JOIN core.projects project ON repo.project_id = project.id " +
```
albscui, line 30: I think we can safely just use a `JOIN` on projects because a repo is guaranteed to belong to a project.

src/server/pfs/server/driver_fsck.go
FahadBSyed: I think we discussed this at the team meeting on 8/8/2023

src/server/pfs/server/trigger.go
```diff
@@ -1,10 +1,11 @@
 package server
 
 import (
+	"context"
+	"google.golang.org/protobuf/proto"
```
albscui, line 5: This seems like a formatter issue? Usually the non stdlib imports are below.

src/internal/coredb/projects_test.go
```diff
@@ -122,25 +122,6 @@ func TestListProject(t *testing.T) {
 	}))
 }
 
-func TestDeleteAllProjects(t *testing.T) {
```
albscui, line 125: Why delete this test?

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,354 @@
...
+			return nil, errors.Wrap(err, "error unmarshalling")
+		}
+		branches = append(branches, branchInfo.Branch)
+	}
+	return branches, nil
+}
+
+// UpsertRepo updates all fields of an existing repo entry in the pfs.repos table by name. If 'upsert' is set to true, UpsertRepo()
+// will attempt to call CreateRepo() if the entry does not exist.
+func UpsertRepo(ctx context.Context, tx *pachsql.Tx, repo *pfs.RepoInfo) error {
```
albscui, line 322: In Postgres, the canonical way of doing upserts is via the `ON CONFLICT` clause.

For example, lets say you already inserted `repo1` previously, you can upsert with the following:

```sql
INSERT INTO pfs.repos (name, type, project_id, description)
VALUES ('repo1', 'user', (SELECT id from core.projects where name='default'), 'repo1 description')
ON CONFLICT (name, type, project_id)
DO UPDATE SET description='updated description';
```

Note that we only update the `description` column because we already know for sure that the conflict is on `name, type, and project_id` column.

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,354 @@
...
+		branchInfo := &pfs.BranchInfo{}
+		if err := proto.Unmarshal(decodedString, branchInfo); err != nil {
+			return nil, errors.Wrap(err, "error unmarshalling")
+		}
+		branches = append(branches, branchInfo.Branch)
+	}
+	return branches, nil
+}
+
+// UpsertRepo updates all fields of an existing repo entry in the pfs.repos table by name. If 'upsert' is set to true, UpsertRepo()
```
albscui, line 320: Also what do you mean `If 'upsert' is set to true` here?

src/internal/watch/postgres/impl.go
```diff
@@ -0,0 +1,144 @@
...
+
+const (
+	// EventInsert happens when an item is added
+	EventInsert EventType = iota
+	// EventUpdate happens when an item is updated
+	EventUpdate
+	// EventDelete happens when an item is removed
+	EventDelete
+	// EventError happens when an error occurred
+	EventError
```
albscui, line 24: Note to self, this is probably not needed since a non-nil error value in the event is enough.

src/internal/watch/postgres/impl.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+const (
+	DefaultBufferSize = 1000
+)
+
+type Event struct {
+	Id        uint64
```
albscui, line 32: Note to self, use `pachsql.ID`?

src/internal/coredb/projects_test.go
FahadBSyed: Delete All relies on `truncate` which isn't allowed when you have a foreign key dependency. We could write a delete all that just does a list and calls delete, but we don't currently have a use case for either of those functions. I figured in the meantime it was better to remove them.

src/internal/pfsdb/repos.go
FahadBSyed: Sounds good! I think the desired behavior we want is to allow the user to override everything except the ID if they are explicitly upserting. This would support future resource renames. What do you think?

src/server/pfs/server/trigger.go
FahadBSyed: Yeah this is probably my IDE. I can fix that.

src/internal/pfsdb/repos.go
albscui: Your current upsert implementation doesn't support renaming a repo. nor should it. You are getting the repo ID by project, name, and type, and if there's an existing repo, then you update given that repo's ID. However, you already fixed the project, name, and type, so the only thing left to change is description. For renaming, your update functionality is all you need.

The correct semantics of upsert is to update an existing row if the value already exists, otherwise insert. To know if a row exists, we must identify it with a primary key or the unique id. We can't use id because we don't want to insert a custom id out of band, instead we want the db to manage the id to be internally consistent. Therefore, the only thing you can rely on is the primary key, which means the only column left that's available for updating is the description.

src/internal/clusterstate/v2.6.0/pfsdb.go
```diff
@@ -19,6 +19,72 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
 )
 
+var schemaUpdate = `
```
brendoncarroll, line 22: This whole section is for release that already went out.  Why is this in the diff?

FahadBSyed: We don't want to keep around the functions that have been added here in the rest of our code base (I think because of dependency import cycles), so we moved them into the cluster state packages to ensure previous migrations still work.

src/internal/clusterstate/v2.8.0.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/pachyderm/pachyderm/v2/src/internal/migrations"
+
+	v2_8_0 "github.com/pachyderm/pachyderm/v2/src/internal/clusterstate/v2.8.0"
+)
+
+var state_2_8_0 migrations.State = v2_8_0.Migrate(state_2_7_0)
+
+// DO NOT MODIFY THIS STATE
```
brendoncarroll, line 11: This comment isn't really accurate.  It doesn't mean anything if we include it in the place where we can add migrations.

src/internal/pachsql/db.go
```diff
@@ -46,6 +46,9 @@ type Tx = sqlx.Tx
 // Stmt is an alias for sqlx.Stmt which is the standard prepared statement type used throught the project
 type Stmt = sqlx.Stmt
 
+// ID is the auto-incrementing primary key used for entries in postgres tables.
+type ID uint64
```
brendoncarroll, line 50: Let's not include this here.  This doesn't give us any more type safety than just a uint64, since everything would depend on this, and then there is no distinction between the identifiers for different objects.

It would be a good idea to create types `RepoID` `CommitID` all based on uint64, because then we would have type safety.  It would be harder to confuse the identifier for a commit for that of a repo.

src/internal/pfsdb/pfsdb.go
```diff
@@ -208,43 +165,3 @@ func Branches(db *pachsql.DB, listener col.PostgresListener) col.PostgresCollect
 		}),
 	)
 }
-
-func ProjectKey(project *pfs.Project) string {
```
brendoncarroll, line 212: Are these in `clusterstate` now?

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+
+// Error satisfies the error interface.
+func (err ErrRepoNotFound) Error() string {
+	if id := err.ID; id != 0 {
+		return fmt.Sprintf("repo id=%d not found", id)
+	}
+	return fmt.Sprintf("repo %s/%s.%s not found", err.Project, err.Name, err.Type)
+}
+
+func (err ErrRepoNotFound) GRPCStatus() *status.Status {
```
brendoncarroll, line 48: 👍

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+	return status.New(codes.NotFound, err.Error())
+}
+
+func IsErrRepoNotFound(err error) bool {
```
brendoncarroll, line 52: 👍

src/internal/pfsdb/pfsdb.go
FahadBSyed: Yes they are, IIRC I don't think there will be any callers for the ProjectKey in the code base anymore and just the name will be used

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+	return "repo already exists"
+}
+
+func (err ErrRepoAlreadyExists) GRPCStatus() *status.Status {
+	return status.New(codes.AlreadyExists, err.Error())
+}
+
+func IsErrRepoAlreadyExists(err error) bool {
+	targetErr := &pgconn.PgError{}
+	ok := errors.As(err, targetErr)
```
brendoncarroll, line 77: This is *very* surprising.  There is a type called `ErrRepoAlreadyExists` and this would return false for it.

Maybe an `IsDuplicateKeyErr` in `pachsql` would help deal with this other usecase.

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+	if !ok {
+		return false
+	}
+	return targetErr.Code == "23505" // duplicate key SQLSTATE
+}
+
```
brendoncarroll, line 83: Throw in a `var _ stream.Iterator[RepoPair] = &RepoIterator{}` so the compiler helps make sure this implements `stream.Iterator`

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+	RepoType    string     `db:"type"`
+	CreatedAt   time.Time  `db:"created_at"`
+	UpdatedAt   time.Time  `db:"updated_at"`
+	// Branches is a string that contains an array of hex-encoded branchInfos. The array is enclosed with curly braces.
+	// Each entry is prefixed with '//x' and entries are delimited by a ','
+	Branches string `db:"branches"`
+}
+
+// NextWithID is like Next, but populates the 'id' and 'dst' parameters each iteration where 'id' is a pointer to the row ID and dst is the repoInfo.
+func (iter *RepoIterator) NextWithID(ctx context.Context, id *pachsql.ID, dst **pfs.RepoInfo) error {
```
brendoncarroll, line 110: I think you should define a type something like

```go
type RepoPair struct {
  ID RepoID
  Repo RepoInfo
}
```

And change Next to return one of these. `Next(ctx context.Context, dst *RepoPair) error`

It's easier to get rid of information than add it.  And `NextWithID` can't be used with any of the utilities in `stream`.  The more useful method should be `Next`.

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+	return iter.NextWithID(ctx, nil, dst)
+}
+
+// ListRepo returns a RepoIterator that exposes a Next() function for retrieving *pfs.RepoInfo references.
+func ListRepo(ctx context.Context, tx *pachsql.Tx) (*RepoIterator, error) {
+	return listRepo(ctx, tx, "", nil)
+}
+
+// ListRepoByIdxType is like ListRepo but only iterates over repo.type = repoType.
+func ListRepoByIdxType(ctx context.Context, tx *pachsql.Tx, repoType string) (*RepoIterator, error) {
```
brendoncarroll, line 149: Is this ascending order by type?

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+	return listRepo(ctx, tx, "type", repoType)
+}
+
+// ListRepoByIdxName is like ListRepo but only iterates over repo.name = name.
+func ListRepoByIdxName(ctx context.Context, tx *pachsql.Tx, repoName string) (*RepoIterator, error) {
+	return listRepo(ctx, tx, "name", repoName)
+}
+
+// ListRepo returns a RepoIterator that exposes a Next() function for retrieving *pfs.RepoInfo references.
+func listRepo(ctx context.Context, tx *pachsql.Tx, where string, whereVal interface{}) (*RepoIterator, error) {
```
brendoncarroll, line 159: This `whereVal` is a little suspicious, but I think it's fine for now.

src/internal/coredb/projects.go
```diff
@@ -17,13 +17,10 @@ import (
 	"github.com/pachyderm/pachyderm/v2/src/pfs"
 )
 
-// ID is the auto-incrementing primary key used for entries in postgres tables.
-type ID uint64
```
brendoncarroll, line 21: I think you should bring back a `type ProjectID uint64` and use that to enforce typing of projects identifiers.

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+		tx:    tx,
+	}
+	if where != "" && whereVal != nil {
+		iter.where = where
+		iter.whereVal = whereVal
+	}
+	return iter, nil
+}
+
+func listRepoPage(ctx context.Context, tx *pachsql.Tx, limit, offset int, where string, whereVal interface{}) ([]repoRow, error) {
```
brendoncarroll, line 177: This looks good 👍

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	err := tx.SelectContext(ctx, &page, getRepoAndBranches+"GROUP BY repo.id, project.name ORDER BY repo.id ASC LIMIT $1 OFFSET $2 ;", limit, offset)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get repo page")
+	}
+	return page, nil
+}
+
+// CreateRepo creates an entry in the pfs.repos table.
+func CreateRepo(ctx context.Context, tx *pachsql.Tx, repo *pfs.RepoInfo) error {
```
brendoncarroll, line 195: Should this return the `RepoID`?

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil && IsErrRepoAlreadyExists(err) {
+		return ErrRepoAlreadyExists{Project: repo.Repo.Project.Name, Name: repo.Repo.Name}
+	}
+	return errors.Wrap(err, "create repo")
+}
+
+// DeleteRepo deletes an entry in the pfs.repos table.
+func DeleteRepo(ctx context.Context, tx *pachsql.Tx, repoProject, repoName, repoType string) error {
+	result, err := tx.ExecContext(ctx, "DELETE FROM pfs.repos "+
+		"WHERE project_id=(SELECT id FROM core.projects WHERE name=$1) AND name=$2 AND type=$3;", repoProject, repoName, repoType)
```
brendoncarroll, line 214: I'm partial to a `LIMIT 1` here, but that might be a little tin-foil-hat.

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+			return nil, ErrRepoNotFound{Project: repoProject, Name: repoName, Type: repoType}
+		}
+		return nil, errors.Wrap(err, "scanning repo row")
+	}
+	return row, nil
+}
+
+// todo(fahad): rewrite branch related code during the branches migration.
+// GetRepo retrieves an entry from the pfs.repos table by using the row id.
+func GetRepo(ctx context.Context, tx *pachsql.Tx, id pachsql.ID) (*pfs.RepoInfo, error) {
```
brendoncarroll, line 254: As mentioned previously having a distinct `RepoID` type would help ensure only the correct identifiers get passed to this function.

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+		if err == sql.ErrNoRows {
+			return nil, ErrRepoNotFound{ID: id}
+		}
+		return nil, errors.Wrap(err, "scanning repo row")
+	}
+	return getRepoFromRepoRow(row)
+}
+
+// GetRepoByName retrieves an entry from the pfs.repos table by project, repo name, and type.
+func GetRepoByName(ctx context.Context, tx *pachsql.Tx, repoProject, repoName, repoType string) (*pfs.RepoInfo, error) {
```
brendoncarroll, line 270: I like that using the index follows the  `GetXByIndex` pattern. 👍

src/internal/pfsdb/repos.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil {
+		return errors.Wrap(err, "upsert repo")
+	}
+	return nil
+}
+
+// UpdateRepo overwrites an existing repo entry by pachsql.ID.
+func UpdateRepo(ctx context.Context, tx *pachsql.Tx, id pachsql.ID, repo *pfs.RepoInfo) error {
+	if repo.Repo.Type == "" {
+		repo.Repo.Type = "unknown"
```
brendoncarroll, line 344: This is essentially a change to the data model.  Is there a design doc where I can read about repo types, or any more context for this?

FahadBSyed: That's unintentional on my part. I wanted to added a 'sane' default because leaving this empty was leading to a panic during development. We can omit these defaults if we think that's better.

src/internal/pfsdb/repos.go
FahadBSyed: Looks like this doesn't work: https://www.crunchydata.com/blog/simulating-update-or-delete-with-limit-in-postgres-ctes-to-the-rescue

If it helps, there is a unique constraint on this table so that (project.name, name, type) should be unique

src/internal/pfsdb/repos.go
FahadBSyed: CreateProject does not either, but maybe it should 🤔

src/server/pps/server/api_server.go
```diff
@@ -414,12 +414,12 @@ const (
 // to perform 'operation' on the pipeline in 'info'
 func (a *apiServer) authorizePipelineOp(ctx context.Context, operation pipelineOperation, input *pps.Input, projectName, outputName string) error {
 	return a.txnEnv.WithReadContext(ctx, func(txnCtx *txncontext.TransactionContext) error {
-		return a.authorizePipelineOpInTransaction(txnCtx, operation, input, projectName, outputName)
+		return a.authorizePipelineOpInTransaction(ctx, txnCtx, operation, input, projectName, outputName)
```
brendoncarroll, line 417: These context changes to the transactions seem like they could be pulled out into another PR and reviewed quickly.

src/internal/watch/postgres/impl.go
```diff
@@ -0,0 +1,144 @@
...
+	"strings"
+	"sync"
+
+	"github.com/lib/pq"
+	"github.com/pachyderm/pachyderm/v2/src/internal/collection"
+	"github.com/pachyderm/pachyderm/v2/src/internal/errors"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pachsql"
+)
+
+type EventType int
```
brendoncarroll, line 14: If this goes into the database it shouldn't be a platform dependent type.  Pick an integer type.

src/internal/pfsdb/repos.go
FahadBSyed: You're right, all the callers are string types, so it doesn't need to be an interface. I think for projects there was a whereVal pattern where one caller was an int type and the other was a string type, but that is not needed here. Fixed!

src/internal/pfsdb/repos.go
FahadBSyed: This is ascending order filtered by type, I think it needs a more clear name. Maybe `ListRepoMatchingType`?

src/server/pps/server/api_server.go
FahadBSyed: will do that now

FahadBSyed: A PR is up for that.

src/server/pfs/server/driver.go
```diff
@@ -78,7 +78,6 @@ type driver struct {
 	prefix     string
 
 	// collections
-	repos    col.PostgresCollection
```
brendoncarroll, line 81: 👍

src/server/pfs/server/api_server.go
```diff
@@ -163,7 +169,11 @@ func (a *apiServer) InspectRepo(ctx context.Context, request *pfs.InspectRepoReq
 
 // ListRepo implements the protobuf pfs.ListRepo RPC
 func (a *apiServer) ListRepo(request *pfs.ListRepoRequest, srv pfs.API_ListRepoServer) (retErr error) {
-	return a.driver.listRepo(srv.Context(), true /* includeAuth */, request.Type, request.Projects, srv.Send)
+	return errors.Wrap(dbutil.WithTx(srv.Context(), a.env.DB, func(ctx context.Context, tx *pachsql.Tx) error {
+		return a.driver.txnEnv.WithWriteContext(ctx, func(txnCxt *txncontext.TransactionContext) error {
+			return a.driver.listRepoInTransaction(srv.Context(), txnCxt, true, request.Type, request.Projects, srv.Send)
```
brendoncarroll, line 174: This is the situation we were trying to avoid with the iterator.  We are calling `srv.Send` inside a transaction, which means that the gRPC client can keep open the transaction by reading slowly.

We are also performing a side effect inside the callback of dbutil.WithTx.   That means we could potentially be returning the results multiple times.  The `WithReadOnly` below might help with that, I'm not quite sure.

For streaming out over gRPC there are two options.
- Have the iterator take a `sqlx.DB` and do a transaction for a page at a time, read out the results.
- Read everything into a slice in memory in a transaction.  Wait for the transaction to succeed, then emit the results over gRPC.  This approach would not be appropriate for commits, but there are typically few enough repos that buffering in memory wouldn't be an issue.

src/internal/clusterstate/v2.8.0/clusterstate.go
```diff
@@ -0,0 +1,30 @@
...
+				return errors.Wrap(err, "creating pfs schema")
+			}
+			return nil
+		}).
+		Apply("Create pfs.repos table", func(ctx context.Context, env migrations.Env) error {
+			if err := createReposTable(ctx, env.Tx); err != nil {
+				return errors.Wrap(err, "creating pfs.repos table")
+			}
+			return nil
+		}).
```
brendoncarroll, line 23: This should use the new `migrations.Squash` functionality so that it all happens at once.  Add a squash for every one except for the first one.  It's just like in git during a rebase.

src/server/pfs/server/driver_fsck.go
```diff
@@ -317,26 +318,21 @@ func fsckCommits(commitInfos map[string]*pfs.CommitInfo, onError func(error) err
...
-		if err := sqlTx.Select(&ids, `SELECT commit_id from  pfs.commit_diffs`); err != nil {
-			return errors.Wrap(err, "select commit ids from pfs.commit_diffs")
-		}
-		for _, id := range ids {
-			cs[id] = pfsdb.ParseCommit(id)
-		}
-		return nil
-	}); err != nil {
-		return nil, err
+	var ids []string
```
brendoncarroll, line 323: Does this need to be in a transaction now?  What was wrong with opening a new transaction, like the old code?

src/server/pfs/server/api_server.go
FahadBSyed: Okay, I think I'm going to implement a non-transactional list to handle this.

src/server/pfs/server/driver_fsck.go
FahadBSyed: Based on testing that @albscui and I did, it seemed to be leading to dangling commit scenarios, but we are not 100% sure why. Our theory was that transactions were being interleaved and causing a state mismatch.

FahadBSyed: update: I split this across 3 transactions (list repos, list matching commits & branches, and listReferencedCommits)

src/server/pfs/server/api_server.go
FahadBSyed: This should be incorporated now

src/internal/watch/postgres/impl.go
brendoncarroll: This can be rebased after the other branch goes in.

src/internal/clusterstate/v2.8.0/clusterstate.go
brendoncarroll: This will also be included in changes in the other branch

src/server/pfs/server/api_server.go
```diff
@@ -163,7 +169,25 @@ func (a *apiServer) InspectRepo(ctx context.Context, request *pfs.InspectRepoReq
 
 // ListRepo implements the protobuf pfs.ListRepo RPC
 func (a *apiServer) ListRepo(request *pfs.ListRepoRequest, srv pfs.API_ListRepoServer) (retErr error) {
-	return a.driver.listRepo(srv.Context(), true /* includeAuth */, request.Type, request.Projects, srv.Send)
+	var repos []*pfs.RepoInfo
+	var err error
+	if err := errors.Wrap(dbutil.WithTx(srv.Context(), a.env.DB, func(ctx context.Context, tx *pachsql.Tx) error {
+		return a.driver.txnEnv.WithWriteContext(ctx, func(txnCxt *txncontext.TransactionContext) error {
```
brendoncarroll, line 175: Can this be `WithReadContext`?

src/server/pfs/server/driver_fsck.go
```diff
@@ -317,26 +318,21 @@ func fsckCommits(commitInfos map[string]*pfs.CommitInfo, onError func(error) err
...
+	for _, id := range ids {
+		cs[id] = pfsdb.ParseCommit(id)
+	}
+	ids = make([]string, 0)
+	if err := tx.Select(&ids, `SELECT commit_id from  pfs.commit_diffs`); err != nil {
+		return nil, errors.Wrap(err, "select commit ids from pfs.commit_diffs")
+	}
+	for _, id := range ids {
+		cs[id] = pfsdb.ParseCommit(id)
 	}
```
brendoncarroll, line 336: After the Commit migration, we should try to avoid writing bespoke SQL outside `pfsdb`.

src/server/pfs/server/master.go
```diff
@@ -87,78 +93,134 @@ func (d *driver) master(ctx context.Context) {
 	})
 }
 
-func (d *driver) manageRepos(ctx context.Context) error {
+func (d *driver) watchRepos(ctx context.Context) error {
 	ctx, cancel := pctx.WithCancel(ctx)
 	defer cancel()
-	repos := make(map[string]context.CancelFunc)
+	repos := make(map[pfsdb.RepoID]context.CancelFunc)
```
brendoncarroll, line 99: 👍

src/server/pfs/server/master.go
```diff
@@ -87,78 +93,134 @@ func (d *driver) master(ctx context.Context) {
...
+	var err error
+	if err := dbutil.WithTx(ctx, d.env.DB, func(ctx context.Context, tx *pachsql.Tx) error {
+		repo, err = pfsdb.GetRepo(ctx, tx, eventID)
+		return errors.Wrap(err, "get repo from event id")
+	}, dbutil.WithReadOnly()); err != nil {
+		return errors.Wrap(err, "get repo")
+	}
+	key := pfsdb.RepoKey(repo.Repo)
+	ctx, cancel := pctx.WithCancel(ctx)
+	repos[eventID] = cancel
```
brendoncarroll, line 186: This is a concurrent write to the map.

src/server/pfs/server/master.go
```diff
@@ -87,78 +93,134 @@ func (d *driver) master(ctx context.Context) {
...
-						if err := ring.Unlock(lockPrefix); err != nil {
-							return err
+			// watch for new repo events.
+			watcher, err := postgres.NewWatcher(d.env.DB, d.env.Listener, masterLockPath, v2_8_0.ReposChannelName)
+			if err != nil {
+				return errors.Wrap(err, "new watcher")
+			}
+			defer watcher.Close()
+			var eg errgroup.Group
+			eg.Go(func() error {
```
brendoncarroll, line 114: I don't think we need this to be in a call to `eg.Go`.  This is the main flow, this "for select" loop should be what the master does until it errors.

src/server/pfs/server/master.go
FahadBSyed: Interesting! I guess we were doing this before, but maybe we never ran into errors because of WatchF? (maybe watchF gets a lock under the hood)

src/server/pfs/server/master.go
brendoncarroll: Move to the bottom, outside of errgroup.

src/server/pfs/server/master.go
```diff
@@ -87,78 +93,134 @@ func (d *driver) master(ctx context.Context) {
...
-						delete(repos, key)
+					case <-ctx.Done():
+						return nil
 					}
-					return nil
 				}
-				if _, ok := repos[key]; ok {
+			})
+			// Get existing entries.
+			existingRepos := make([]*postgres.Event, 0)
```
brendoncarroll, line 130: This is where we should instantiate the `map[RepoID]context.CancelFunc`

src/server/pfs/server/master.go
```diff
@@ -87,78 +93,134 @@ func (d *driver) master(ctx context.Context) {
...
-				return nil
-			})
-			return errors.EnsureStack(err)
+			}
+			return errors.Wrap(eg.Wait(), "waiting for manageRepos goroutines to finish")
 		},
 	)
 }
 
+func (d *driver) manageRepos(ctx context.Context, ring *consistenthashing.Ring, repos map[pfsdb.RepoID]context.CancelFunc, ev *postgres.Event) error {
```
brendoncarroll, line 157: This can manage a single repo, and avoid needing the map.  Will be canceled through context by the main loop, when a delete occurs.

src/server/pfs/server/master.go
```diff
@@ -87,78 +93,134 @@ func (d *driver) master(ctx context.Context) {
...
-					return ev.Err
-				}
-				key := string(ev.Key)
-				lockPrefix := path.Join("repos", key)
-				if ev.Type == watch.EventDelete {
-					if cancel, ok := repos[key]; ok {
-						if err := ring.Unlock(lockPrefix); err != nil {
-							return err
+			// watch for new repo events.
+			watcher, err := postgres.NewWatcher(d.env.DB, d.env.Listener, masterLockPath, v2_8_0.ReposChannelName)
```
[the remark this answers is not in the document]
albscui: Should we do something like the following in pfsdb?
```
const ReposChannelName = v2_8_0.ReposChannelName
```

src/server/pfs/server/master.go
```diff
@@ -87,76 +92,123 @@ func (d *driver) master(ctx context.Context) {
...
+			d.manageRepo(ctx, ring, repo, lockPrefix)
+		case <-ctx.Done():
+			return nil
+		}
+	}
+}
+
+func (d *driver) manageRepo(ctx context.Context, ring *consistenthashing.Ring, repo *pfs.RepoInfo, lockPrefix string) {
+	key := pfsdb.RepoKey(repo.Repo)
+	go func() {
```
brendoncarroll, line 179: Can this go statement move up to the caller.  So it's `go d.manageRepo(...)` ?  That we we don't have to indent all of this.

src/server/pfs/server/master.go
```diff
@@ -87,76 +92,123 @@ func (d *driver) master(ctx context.Context) {
...
+						return err
+					}
+					cancel()
+					delete(repos, repoID)
 				}
-				if _, ok := repos[key]; ok {
+				continue
+			}
+			if _, ok := repos[repoID]; ok {
+				continue // the master has already called manageRepo for this repo.
```
brendoncarroll, line 158: 👍

brendoncarroll (approved):
LGTM

src/internal/watch/postgres/impl_test.go
```diff
@@ -3,11 +3,11 @@ package postgres_test
 import (
 	"context"
 	"fmt"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pfsdb"
```
albscui, line 6: Move this down to the group below.

src/server/pfs/server/driver_fsck.go
```diff
@@ -4,10 +4,12 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	col "github.com/pachyderm/pachyderm/v2/src/internal/collection"
```
albscui, line 7: Same comment as above.

src/internal/clusterstate/v2.8.0/pfsdb.go
```diff
@@ -109,7 +105,7 @@ func createReposTable(ctx context.Context, tx *pachsql.Tx) error {
 	}
 	// Create a trigger that notifies on changes to pfs.repos
 	// This is used by the PPS API to watch for changes to repos
-	if _, err := tx.ExecContext(ctx, generateTriggerFunctionStatement("pfs", "repos", ReposChannelName)); err != nil {
+	if _, err := tx.ExecContext(ctx, generateTriggerFunctionStatement("pfs", "repos", pfsdb.ReposChannelName)); err != nil {
```
albscui, line 108: This shouldn't import from pfsdb. Should be immutable, so just use a string literal.

src/internal/clusterstate/v2.8.0/pfsdb.go
```diff
@@ -3,6 +3,7 @@ package v2_8_0
 import (
 	"context"
 	"fmt"
+	"github.com/pachyderm/pachyderm/v2/src/internal/pfsdb"
```
albscui, line 6: Remove this and use a string literal for repo channel name.

src/server/pfs/server/master.go
FahadBSyed: sure we can do that

src/internal/clusterstate/v2.8.0/pfsdb.go
FahadBSyed: You're right (because it can change over time.) We should just redefine it here for the scope of this package too.

PR #9020, opened 2023-07-03
[CORE-1708] Propagate cluster defaults

robert-uhl:
This PR implements propagation of cluster defaults to pipelines when the `regenerate` flag is true. When it is, the effective spec for each pipeline is calculated from the cluster defaults and the user spec; if the newly-calculated effective spec differs from the old effective spec, then the effective spec is replaced (as in an update pipeline operation). Furthermore, if the `reprocess` flag is set then the pipeline will reprocess datums (again, as in an update pipeline operation).

In order to implement this, create-pipeline transactions had to be reworked. They were previously just the create pipeline request message; they are now a special message which captures the user and effective specs in addition to the create pipeline request. Rather than implement this change in the old private `ppsBuilderClient.CreatePipeline` methods — uncalled in the codebase — the dead code is deleted.

The defaults-handling code is abstracted, in order to provide a single place for wrapping and unwrapping of JSON values and in hopes of improving legibility of the API server.

The former `details_json` name is updated to the more accurate `user_spec_json`, to pair with the new `effective_spec_json`.


src/transaction/transaction.proto
```diff
@@ -22,8 +22,10 @@ message TransactionRequest {
   pfs_v2.CreateBranchRequest create_branch = 6;
   pfs_v2.DeleteBranchRequest delete_branch = 7;
   pps_v2.UpdateJobStateRequest update_job_state = 8;
-  pps_v2.CreatePipelineRequest create_pipeline = 9;
+  reserved 9;
+  reserved "create_pipeline";
```
bbonenfant, line 26: Do we have any formal deprecation process for our APIs? I don't have much issue with this particular change, but it happened recently with the debug dumpv2 API as well so I think it's worth bringing up.

src/pps/pps.proto
```diff
@@ -464,7 +464,9 @@ message PipelineInfo {
     Determined determined = 36;
   }
   Details details = 12;
-  string details_json = 13;
+  reserved "details_json";
```
bbonenfant, line 467: Same question about deprecation here. This one actually seems like an issue since we are renaming the field after it was released in v2.7.0?

src/pps/pps.proto
```diff
@@ -464,7 +464,9 @@ message PipelineInfo {
...
+  string user_spec_json = 13;
```
bbonenfant, line 468: Could you either comment these fields like you did in `CreatePipelineTransaction` or have a `see: CreatePipelineTransaction` if you don't want to duplicate?

src/transaction/transaction.proto
robert-uhl: > Do we have any formal deprecation process for our APIs?

I think the formal process is getting team sign-offs on the proto changes that we started in the last 48 hours 😁 

In this particular case, the issue is that the transaction request message formerly just contained a create pipeline request, whereas now it needs to contain three items of data. Rather than encode that knowledge of PPS information within the transaction service, I think it makes sense for it to pull in a create pipeline transaction message from PPS itself.

src/pps/pps.proto
robert-uhl: Renaming is generally safe because proto fields are identified by numbers. It _is_ implicated JSON or other text-formatted messages are involved, but `PipelineInfo` is not used as a text-formatted message anywhere.

src/server/pps/cmds/cmds.go
```diff
@@ -1695,6 +1644,46 @@ func Cmds(mainCtx context.Context, pachCtx *config.Context, pachctlCfg *pachctl.
...
+	}
+
+	client, err := pachctlCfg.NewOnUserMachine(ctx, false)
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	var resp *pps.SetClusterDefaultsResponse
+	if resp, err = client.PpsAPIClient.SetClusterDefaults(ctx, req); err != nil {
```
FahadBSyed, line 1671: [style question] could you avoid a forward declaration and just do?
```go
resp, err := client.PpsAPIClient.SetClusterDefaults(ctx, req)
if err != nil { ... }
```
not sure which is idiomatic or preferrable here

robert-uhl: I could. I personally prefer declarations, because it means I can see the exact type, and it means I can use `if resp, err = f(); err != nil { … }` exactly like `if foo, bar.baz, err = f(); err != nil { … }`. I don’t know of a Go style guide which takes a strong stance one way or another.

bbonenfant (approved):

FahadBSyed (approved):

src/server/pps/server/api_server.go
```diff
@@ -3771,21 +3739,75 @@ func unknownError(ctx context.Context, msg string, err error) error {
...
+			spec := ppsutil.PipelineReqFromInfo(pi)
+			// if the old details are missing, synthesize them
+			if pi.UserSpecJson == "" {
+				b, err := protojson.Marshal(spec)
+				if err != nil {
+					return errors.Wrap(err, "could not marshal spec to JSON")
+				}
+				pi.UserSpecJson = string(b)
+			}
+			if pi.EffectiveSpecJson == "" {
```
zmajeed, line 3769: could an empty user spec or effective spec be checked earlier to avoid conversion and other operations?

zmajeed:
couple general comments - I see CORE-1913 https://pachyderm.atlassian.net/browse/CORE-1913 - it will be interesting to see the performance impact from these new json conversions particularly for large pipelines

if regenerate changes a pipeline, will it be possible to tell the reason and distinguish it from a user update? also generally is it possible to see the source/origin of any pipeline spec field - if it's coming from defaults or not - without doing a diff?

src/server/pps/server/api_server.go
robert-uhl: I’ve moved the spec assignment into the conditional; now the empty user spec check is the very first thing performed on each pipeline.

robert-uhl:
> if regenerate changes a pipeline, will it be possible to tell the reason and distinguish it from a user update?

No, not currently.

> also generally is it possible to see the source/origin of any pipeline spec field - if it's coming from defaults or not - without doing a diff?

No, not currently.

PR #9023, opened 2023-07-03
Record Restart when Worker Master is cancelled

acohen4:
In the Worker Master, we rely on `RetryUntilCancel` & its `Notify` callback to record job restarts when the job fails. Since `RetryUntilCancel` won't run `Notify` when the context is cancelled, the worker master will restart without recording so

### DIFF
--- a/src/internal/backoff/retry.go
+++ b/src/internal/backoff/retry.go
@@ -121,7 +121,7 @@ func RetryNotify(operation Operation, b BackOff, notify Notify) error {
 }
 
 // RetryUntilCancel is the same as RetryNotify, except that it will not retry if
-// the given context is canceled.
+// the given context is canceled. If the context is cancelled, then notify wll not run.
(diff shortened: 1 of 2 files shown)


brycemcanally:
I don't think this would work as is since the root context (pach client context) being canceled would prevent any of the subsequent RPCs from proceeding (load, update job info, etc.). Also, restarts wouldn't be correct even if we could account for the worker master losing its lock because the worker master could also just crash at any point without recording a restart. Recording the attempts somewhere then inferring the restarts from that would work.

jrockway:
reg.driver's context is the one holding the master lock, so I think bryce is right; even doing the pj.driver = reg.driver thing still has a cancelled context.  We could think about relaxing that, but it's probably not a good idea; only an elected master should be writing job status, and using that context enforces that invariant.  What this means is that we need some other way to inform the rest of pachyderm about job cancellation due to losing the master lock.

PR #9026, opened 2023-07-05
[PFS-112] Add additional logging to the provenance migration

brycemcanally:
This PR adds logging to all of the places in the provenance migration that we expect to be expensive. This primarily applies to operations that are per-commit and per-job. The expectation after this change is that users will generally see a steady supply of logs throughout the entire migration.

### DIFF
--- a/src/internal/clusterstate/v2.6.0/pfsdb.go
+++ b/src/internal/clusterstate/v2.6.0/pfsdb.go
@@ -583,17 +583,26 @@ func branchlessCommitsPFS(ctx context.Context, tx *pachsql.Tx) error {
 			return errors.Wrapf(err, "could not migrate branch %q", branchKey(bi.Branch))
 		}
 	}
-	// map <project>/<repo>@<branch>=<id> -> <project>/<repo>@<id>; basically replace '@[-a-zA-Z0-9_]+)=' -> '@'
-	log.Info(ctx, "removing branch from internal keys (this can be an expensive operation)")
+	return branchlessCommitKeysPFS(ctx, tx)
+}
+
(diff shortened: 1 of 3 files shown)


src/internal/clusterstate/v2.6.0/pfsdb.go
```diff
@@ -578,6 +584,7 @@ func branchlessCommitsPFS(ctx context.Context, tx *pachsql.Tx) error {
 		}
 	}
 	// map <project>/<repo>@<branch>=<id> -> <project>/<repo>@<id>; basically replace '@[-a-zA-Z0-9_]+)=' -> '@'
+	log.Info(ctx, "removing branch from internal keys (this can be an expensive operation)")
```
brendoncarroll, line 587: Can we have some kind of done line for this.  Maybe using the Span functionality?

brendoncarroll (approved):
We might benefit from putting spans around some of these sections.  The logging in the for loops seems good to me.

src/internal/clusterstate/v2.6.0/pfsdb.go
```diff
@@ -174,6 +175,7 @@ func removeAliasCommits(ctx context.Context, tx *pachsql.Tx) error {
 	realAncestors := make(map[string]*v2_5_0.CommitInfo)
 	childToNewParents := make(map[*pfs.Commit]*pfs.Commit)
 	for _, ci := range deleteCommits {
+		log.Info(ctx, "computing parent / children after deleted alias commit", zap.String("commit", oldCommitKey(ci.Commit)))
```
acohen4, line 178: WDYT about adding a total (+ counter) for the index of the item processed?

brycemcanally: I added a progress field for each log that occurs in a for loop. It contains the string "index / total".

acohen4 (approved):
LGTM

PR #9033, opened 2023-07-06
Fix fuse integration test race

brycemcanally:
This PR fixes a race condition in the fuse integration tests. The race was between the tests reading commits and the PFS master finishing commits (commits are finished asynchronously after finish commit is called or a one-off put file). Although we allow reading open commits that have a finished parent, it is generally not recommended for performance reasons. We don't allow reading open commits that have an unfinished parent for the same performance reasons. The fuse integration tests were getting into a situation where they created a chain of commits then attempted to read the most recent commit without waiting on it to be finished, which would often cause an error due to the most recent commit having an unfinished parent. The fix was just to wait on the commits to be finished before attempting to read them.


chainlink (approved):
LGTM! TIL `WaitCommitSetAll`

PR #9040, opened 2023-07-06
[CORE-1804] Build Debug Version of 2.6.x for Customer Issue Reported in CORE-1804

FahadBSyed:


src/internal/clusterstate/v2.6.0/pfsdb.go
```diff
@@ -175,10 +176,13 @@ func removeAliasCommits(ctx context.Context, tx *pachsql.Tx) error {
...
-			return errors.Errorf("commit %q is listed as ALIAS but has a different ID than it's first real ancestor.",
-				oldCommitKey(ci.Commit))
+			notSame = append(notSame, oldCommitKey(ci.Commit))
+		} else {
+			count++
 		}
-		count++
+	}
+	if len(notSame) > 0 {
+		return errors.Errorf("commits %q are listed as ALIAS but have a different ID than their first real ancestor.", notSame)
```
acohen4, line 185: I think it needs to be %v instead of %q

PR #9043, opened 2023-07-07
generate python protos when running make proto

kevyang:
[INT-1019](https://pachyderm.atlassian.net/browse/INT-1019)

[INT-1019]: https://pachyderm.atlassian.net/browse/INT-1019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


etc/proto/test_no_changes.sh
```diff
@@ -28,8 +37,22 @@ find src -regex ".*\.pb\.go" \
   | awk '{print $1}'
 )"
 
+# hash newly generated python code
```
chainlink, line 40: Did you try modifying a proto and making sure this would trip?

chainlink (approved):
LGTM

Makefile
```diff
@@ -229,6 +229,7 @@ test-proto-static:
 
 proto: docker-build-proto
 	./etc/proto/build.sh
+	./python-sdk/proto/generate-protos.sh
```
bbonenfant, line 232: I don't know much about Makefiles or if you can chain `make` calls, but should we try to hook into the command within python-sdk/Makefile and run `make -C python-sdk proto`?

etc/proto/test_no_changes.sh
kevyang: yes, this tripped when i tested it because the python protos did indeed need to be regenerated (see the pfs change included in this PR)

PR #9044, opened 2023-07-07
Remove delete_all from python-sdk

chainlink:


bbonenfant (approved):
I'm OK with this change. We should make a ticket for the backlog though about potentially reintroducing this functionality under the admin api like we spoke about.

Additionally the CI should be fixed by bring this branch up to date with master.

BOsterbuhr:
Delete all is pretty useful during the early stages of development. I am not sure we should remove this, especially because it is still available in pachctl.

PR #9045, opened 2023-07-07
storage/kv: Add kv.Metered for instrumenting kv.Stores

brendoncarroll:
As part of the changes to move our caching and limiting to the `storage/kv` package, some of the metering was lost in the transition.  This PR adds it back.

Previously we were using prometheus metrics in the `obj` package, but since we have the new `meters` package it didn't make sense to port the prometheus implementation directly.  This PR uses the `meters` package to instrument any `kv.Store`.

As we setup the storage layer, we add metering to two named stores `bucket` which is object storage, and `disk` which is the local on-disk cache.


jrockway (changes requested):
Somewhere, probably in storage/config, we want to aggregate put, putBytes, get, and getBytes with  `NewAggregatedCounter(ctx, <name>, 0, meters.WithFlushInterval(time.Minute))` or similar.  This way, we hear from the metrics once a minute (per key) rather than for every byte/op.

PR #9046, opened 2023-07-07
GoCDKEnabled configuration parameter

brendoncarroll:
GoCDKEnabled sets whether or not to use the GoCDK library for accessing object storage. The intent here is to gate all of the setup logic on this parameter, so that when GoCDKEnabled=false, no GoCDK code is run as part of setup.

### DIFF
--- a/src/internal/obj/factory.go
+++ b/src/internal/obj/factory.go
@@ -549,3 +549,21 @@ func NewClient(ctx context.Context, storageBackend string, storageRoot string) (
 		return nil, errors.Errorf("unrecognized storage backend: %s", storageBackend)
 	}
 }
+
+// NewBucket creates a Bucket using the given backend and storage root (for
+// local backends).
+// TODO: Not sure if we want to keep the storage root configuration for
+// non-local deployments. If so, we will need to connect it to the object path
+// prefix for chunks.
+func NewBucket(ctx context.Context, storageBackend string, storageRoot string) (*Bucket, error) {
+	var b *Bucket
+	var err error
+	switch storageBackend {
+	case Local:
+		b, err = NewLocalBucket(storageRoot)
+	// TODO: other backends
+	default:
+		return nil, errors.Errorf("unrecognized storage backend: %s", storageBackend)
+	}
+	return b, err
+}
--- a/src/server/pfs/server/env.go
+++ b/src/server/pfs/server/env.go
@@ -16,11 +16,13 @@ import (
 	authserver "github.com/pachyderm/pachyderm/v2/src/server/auth"
 	ppsserver "github.com/pachyderm/pachyderm/v2/src/server/pps"
 	etcd "go.etcd.io/etcd/client/v3"
+	"gocloud.dev/blob"
 )
 
 // Env is the dependencies needed to run the PFS API server
 type Env struct {
 	ObjectClient obj.Client
+	Bucket       *blob.Bucket
 	DB           *pachsql.DB
 	EtcdPrefix   string
 	EtcdClient   *etcd.Client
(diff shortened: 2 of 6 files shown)


src/internal/obj/factory.go
```diff
@@ -549,3 +549,21 @@ func NewClient(ctx context.Context, storageBackend string, storageRoot string) (
 		return nil, errors.Errorf("unrecognized storage backend: %s", storageBackend)
 	}
 }
+
+// NewBucket creates a Bucket using the given backend and storage root (for
+// local backends).
+// TODO: Not sure if we want to keep the storage root configuration for
+// non-local deployments. If so, we will need to connect it to the object path
+// prefix for chunks.
+func NewBucket(ctx context.Context, storageBackend string, storageRoot string) (*Bucket, error) {
```
brendoncarroll, line 558: This is the function called from PFS at setup time, so all the setup logic will go in here in a subsequent PR.

src/server/pfs/server/env.go
```diff
@@ -41,17 +43,33 @@ type Env struct {
...
-	if err != nil {
-		return nil, err
+	cfg := env.Config()
+
+	// Only setup the new GoCDK bucket if necessary.  The storage layer will prefer it, if it is non-nil.
+	// Disabling GoCDK will completely bypass the entire setup, and revert to the old obj.Client.
+	// This is designed to be a quick fix in the field, if issues are discovered with Go CDK.
+	var objClient obj.Client
+	var bucket *obj.Bucket
+	if cfg.GoCDKEnabled {
```
brendoncarroll, line 53: This is where setup of GoCDK is gated.

src/server/pfs/server/env.go
```diff
@@ -41,17 +43,33 @@ type Env struct {
...
+			return nil, err
+		}
 	}
 	etcdPrefix := path.Join(env.Config().EtcdPrefix, env.Config().PFSEtcdPrefix)
 	if env.AuthServer() == nil {
 		panic("auth server cannot be nil")
 	}
 	return &Env{
 		ObjectClient: objClient,
+		Bucket:       bucket,
```
brendoncarroll, line 72: We always pass bucket through, and the storage layer will try to use it, if it is available.

src/server/pfs/server/env.go
brycemcanally: Have you considered an obj.Client adapter for Go CDK? If that is feasible, it seems like it would be a cleaner way to introduce Go CDK into the storage layer. It wouldn't be exactly the same as calling the Go CDK methods directly, but it should test out the stuff we would be concerned about (i.e. setup and general usage).

src/internal/obj/factory.go
```diff
@@ -549,3 +549,21 @@ func NewClient(ctx context.Context, storageBackend string, storageRoot string) (
...
+	var b *Bucket
+	var err error
+	switch storageBackend {
+	case Local:
+		b, err = NewLocalBucket(storageRoot)
+	// TODO: other backends
```
brycemcanally, line 564: Will this be done in this PR?

brendoncarroll: No I was planning on adding backends one at a time in follow up PRs.

src/server/pfs/server/env.go
brendoncarroll: Adapting GoCDK to obj.Client? That wouldn't gain us anything, the work has already been done to depend on GoCDK directly (which is also a better solution long term).  That's already in master.

Part of the motivation for doing this was to have access to the Go CDK interface, so we can do things like Signed URLs, and get sizes of objects during lists.  Just using Go CDK for setup, is all of the costs, and none of the benefits.

All that's left to do at this point is turn our environment variables into Go CDK buckets.  I think Fahad already has some code that turns `obj.URLs` into Go CDK buckets.  I need to sync with him to see how he recommends adapting that to use the environment variables.

brycemcanally: Okay, I wasn't aware of the change to using the `kv.Store` interface internally. I thought we were still using the `obj.Client` interface and you were going to switch internally based on what was set.

src/internal/obj/factory.go
brycemcanally: Okay, I would prefer us to not introduce any of the backend specific code then. It is untested and dead as of now.

PR #9047, opened 2023-07-07
Wrap DB calls in retries to catch connection flakiness

acohen4:


src/internal/collection/postgres_collection.go
```diff
@@ -188,8 +195,20 @@ type postgresReadOnlyCollection struct {
...
-	if err := sqlx.GetContext(ctx, q, result, queryString, key); err != nil {
-		return nil, c.mapSQLError(err, key)
+	// wrap sqlx.GetContext() in a retry to mitigate database connection flakiness.
+	if err := backoff.RetryUntilCancel(ctx, func() error {
+		if err := sqlx.GetContext(ctx, q, result, queryString, key); err != nil {
+			return c.mapSQLError(err, key)
+		}
+		return nil
+	}, backoff.RetryEvery(time.Second), func(err error, d time.Duration) error {
+		if strings.Contains(err.Error(), "broken pipe") || strings.Contains(err.Error(), "unexpected EOF") {
```
jrockway, line 205: Let's extract this into an errutils.IsDatabaseDisconnect because I'm sure there's some third message we'll want to test for eventually.  We should also check to see if `errors.Is(err, io.ErrUnexpectedEOF)` catches the error we care about; just to make this condition a little more precise.

jrockway (approved):
I added a commit to adjust get `get()` retry logic (basically only retry on the DB connection errors, not things like sql.ErrNoRows).  That should fix the tests.

Overall, I haven't tested this as much as I want to yet, but it's looking really good so far.  Restarting pgbouncer while user code is running basically has no effect, the user code finishes and moves on to the next datum.  So that's already a huge reliability win.  I think this is going to be good.

PR #9049, opened 2023-07-07
[Python-SDK] Extension to DebugDump

bbonenfant:
Creates api/debug/extension.py that overrides the debug.dump method, routing it to the DumpV2 endpoint and providing a sensible default (matching pachctl).


kevyang (approved):
lgtm

PR #9051, opened 2023-07-10
Lock Required Tables in 2.6 migrations

acohen4:


src/internal/clusterstate/v2.6.0/clusterstate.go
```diff
@@ -50,6 +51,24 @@ func Migrate(state migrations.State) migrations.State {
...
+				"collections.commits",
+				"pfs.commit_diffs",
+				"pfs.commit_totals",
+				"storage.tracker_objects",
+				"pfs.commits",
+				"pfs.commit_provenance",
+				"collections.pipelines",
+				"collections.jobs",
+			} {
+				if _, err := env.Tx.ExecContext(ctx, fmt.Sprintf("LOCK TABLE %s IN EXCLUSIVE MODE", table)); err != nil {
```
jrockway, line 68: We should probably add a migrations.Env.LockTable() method, so that we can skip locks during `test-migrations`s.

jrockway (approved):
Looks good to me.  I suppose the signature `LockTables(ctx context.Context, tables ...string)` might be less work for users, but on the other hand, a for loop is not a big deal.

albscui:
Just wondering, how would existing users who applied the 2.6.0 migration get this update? Or is this purely for net new users?

PR #9053, opened 2023-07-10
Better upgrade tests to increase confidence about migrations

albcui:


src/testing/deploy/upgrade_test.go
```diff
@@ -261,6 +262,88 @@ func TestUpgradeOpenCVWithAuth(t *testing.T) {
 	)
 }
 
+func TestUpgradeMultiProjectJoins(t *testing.T) {
+	if skip {
+		t.Skip("Skipping upgrade test")
+	}
+	fromVersions := []string{"2.5.0", "2.6.0"}
```
jrockway, line 269: We should probably include 2.4 here as well.

jrockway (approved):
This looks good.

src/testing/deploy/upgrade_test.go
albscui: So creation of projects is a new feature introduced in 2.5, and our auth interceptor expects it to be auth gated. Our test creates projects, so it's fundamentally incompatible with 2.4.

jrockway: ahh, ok!

PR #9054, opened 2023-07-10
Run python-sdk tests on core changes

bbonenfant:
[INT-1030]

[INT-1030]: https://pachyderm.atlassian.net/browse/INT-1030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/main.yml
+++ b/.circleci/main.yml
@@ -1446,6 +1446,8 @@ jobs:
       image: ubuntu-2004:202107-02
     working_directory: ~/project/python-sdk
     steps:
+      - attach_workspace:
+          at: ./
       - checkout:
           path: ~/project
       - install-minikube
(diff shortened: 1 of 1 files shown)


.circleci/main.yml
```diff
@@ -1452,6 +1452,12 @@ jobs:
       - install-helm
       - install-kubectl
       - start-minikube
+      - attach_workspace:
+          at: ./
+      - run:
+          name: Install pachctl
+          command: |
+            sudo mv ./dist-pach/pachctl/pachctl_linux_amd64_v1/pachctl /usr/local/bin/pachctl
```
molinamelendezj, line 1460: sorry maybe im reading this wrong. wouldn't this need to have this job run after pachctl has built. to do so would need a `requires` on the job

molinamelendezj (approved):

.circleci/main.yml
bbonenfant: Yeah the requires already is in place, I just forgot to actually "install" pachctl for this job. I figured since now core-team members are more likely to touch these jobs it would be useful to have pachctl actually present if they want to ssh in.

PR #9059, opened 2023-07-10
python-sdk - bring opencv example up to par with python-pachyderm

kevyang:
this change fleshes out the opencv example to match what we have written for python-pachyderm

[INT-1036](https://pachyderm.atlassian.net/browse/INT-1036)

[INT-1036]: https://pachyderm.atlassian.net/browse/INT-1036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


python-sdk/examples/opencv/README.md
```diff
@@ -0,0 +1,17 @@
+# Distributed Image Processing with OpenCV and Python Pachyderm
+
+This is a reproduction of Pachyderm's opencv example in python. A walkthrough is available in the Pachyderm *[docs](https://docs.pachyderm.io/en/latest/getting_started/beginner_tutorial.html)*
+
+This example showcases the pachyderm-sdk analogs of common `pachctl` commands, such as creating repos and pipelines or getting a file.
+
+**Prerequisites:**
+- A workspace on [Pachyderm Hub](https://docs.pachyderm.com/latest/hub/hub_getting_started/) (recommended) or Pachyderm running [locally](https://docs.pachyderm.com/latest/getting_started/local_installation/)
```
bbonenfant, line 8: We don't have a hub/saas offering anymore, so you can delete this portion

python-sdk/examples/opencv/README.md
```diff
@@ -0,0 +1,17 @@
...
+- pachyderm-sdk installed. A pip package is coming soon, but for now, you can run from the python-sdk directory:```
```
bbonenfant, line 9: I think you can leave out the  "pip package coming soon" since it's likely to be forgotten when this gets release, and say "install the latest package or ..."

python-sdk/examples/opencv/opencv.py
bbonenfant, line 1: Can you change the `Client()` calls here to `Client.from_config()`? That's most likely to work for users than hoping they are running on `localhost:30650`. Also update the comment at the start of `main():` if you make this change

python-sdk/examples/opencv/opencv.py
```diff
@@ -72,12 +72,17 @@ def main():
...
-        source = None  # TODO: Add correct path
+        source = relpath("images")
         client.pfs.put_files(commit=commit, source=source, path="/")
 
     # Wait for the commit (and its downstream commits) to finish
     commit.wait_set()
 
+    job = pps.Job(pipeline=pps.Pipeline(name="montage"), id=commit.id)
+    if client.pps.inspect_job(job=job).state != pps.JobState.JOB_SUCCESS:
+        print("Montage job failed, aborting. Check the pipeline logs for more details.")
```
bbonenfant, line 83: Since we just made the call to retrieve the JobInfo, it would be nice to print out the exact pachctl command used to get the logs.

python-sdk/examples/opencv/opencv.py
```diff
@@ -72,12 +72,17 @@ def main():
         # Add some images, recursively inserting content from the images
         # directory. Alternatively, you could use `client.put_file_url` or
         # `client_put_file_bytes`.
-        source = None  # TODO: Add correct path
+        source = relpath("images")
         client.pfs.put_files(commit=commit, source=source, path="/")
```
bbonenfant, line 76: It appears (#9060) that we probably want to use the images from the docs site instead of committing these images to the repo. You should be able to do `client.pfs.put_file_from_url(commit, url="...", path="/")` for each image.

bbonenfant (commented):
We should move the Dockerfile from python-pachyderm over into this directory as well. We don't need to figure out building/publishing this image at this time, but we should put a ticket in the backlog for doing so.

kevyang:
> We should move the Dockerfile from python-pachyderm over into this directory as well. We don't need to figure out building/publishing this image at this time, but we should put a ticket in the backlog for doing so.

do you mean the `python-sdk` directory, where `dist` would be generated? or would we expect users to specify this dockerfile?

bbonenfant:
> > We should move the Dockerfile from python-pachyderm over into this directory as well. We don't need to figure out building/publishing this image at this time, but we should put a ticket in the backlog for doing so.

> do you mean the python-sdk directory, where dist would be generated? or would we expect users to specify this dockerfile?

The Dockerfile for generating the `edges` pipeline container. Looks like I was misremembering where it was, it's [here](https://github.com/pachyderm/pachyderm/blob/master/examples/opencv/Dockerfile) so maybe instead of moving it just reference that it's there. It looks like edges.py is there as well, so maybe we don't need the edges.py or requirements.txt here either

python-sdk/examples/opencv/opencv.py
```diff
@@ -86,13 +90,13 @@ def main():
...
+    client = Client.from_config()
     client.pps.delete_pipeline(pipeline=pps.Pipeline(name="montage"))
     client.pps.delete_pipeline(pipeline=pps.Pipeline(name="edges"))
 
     client.pfs.delete_repo(repo=pfs.Repo.from_uri("images"), force=True)
 
 
 if __name__ == "__main__":
-    # clean()
+    clean()
```
bbonenfant, line 101: If we are going to incorporate this clean function, we should probably name the resources something more unique? I'm envisioning us accidentally deleting someones "images" repo. maybe prepend `opencv-` to the resource names?

python-sdk/examples/opencv/opencv.py
```diff
@@ -86,13 +90,13 @@ def main():
 
 
 def clean():
-    client = Client()
+    client = Client.from_config()
```
bbonenfant, line 93: We should probably localize the Client creation in case someone edits the `main` Client but forgets this one. Maybe pass the client as an argument to these functions and move the client creation and that large comment to the `if __name__ == "__main__":` section? Or it could live as a global at the top of the file so its more visible

python-sdk/examples/opencv/README.md
```diff
@@ -0,0 +1,16 @@
+# Distributed Image Processing with OpenCV and Python Pachyderm
+
+This is a reproduction of Pachyderm's opencv example in python. A walkthrough is available in the Pachyderm *[docs](https://docs.pachyderm.io/en/latest/getting_started/beginner_tutorial.html)*
+
+This example showcases the pachyderm-sdk analogs of common `pachctl` commands, such as creating repos and pipelines or getting a file.
+
+The image being used to run the pipeline code is built from the Dockerfile located at pachyderm/examples/opencv. You will also find the edges.py script there.
```
bbonenfant, line 7: [nit] I'd maybe provide a relative link within the repo or a url link to the pachyderm/example/opencv directory

bbonenfant (approved):
Looks great!

python-sdk/examples/opencv/opencv.py
kevyang: unfortunately edges.py assumes that the input is `/pfs/images` so unless we change that too then the example won't work

PR #9060, opened 2023-07-10
[2.5.x] use images hosted in docs for examples (#8897)

jrockway:
Since we are backporting stuff and imgur still returns 429s, we need this in 2.5.x now.


molinamelendezj (approved):
Good catch. Hopefully nothing else is lerking that needs back porting. Going be hard to tell.

PR #9065, opened 2023-07-11
security: go 1.20.6

jrockway:
Fixes GO-2023-1878


molinamelendezj (approved):
LGTM

@lemontree-ca this is an example of what we were just talking about earlier.

PR #9066, opened 2023-07-12
Dump from many pachd replicas

acohen4:


src/server/debug/server/server.go
```diff
@@ -292,10 +292,10 @@ func (s *debugServer) makeTasks(ctx context.Context, request *debug.DumpV2Reques
 			}
 		}
 		if len(sys.Profiles) > 0 {
-			ts = append(ts, makeProfilesTask(server, sys.Profiles))
+			ts = append(ts, s.makeProfilesTask(server, sys.Profiles))
```
jrockway, line 295: Do profiles run in parallel?  There should be some way to do that, so doing a 30s CPU profile takes 30s regardless of the number of profiles requested.  (If you do them 1 by 1, you do that sleep 30 n times.  If you do them in parallel, then you sleep 30 in parallel.)

jrockway (approved):
I figured this out.  I guess we hard-code goroutine and heap.

I always imagined profile being (app, profile, args), not just app.  That way you can do a CPU profile, and set options for it, etc.

PR #9072, opened 2023-07-12
[CORE-1844] Support proto3 optional fields in pachgen/protoc-gen-pach

robert-uhl:
It appears that protoc-gen-pach (aka pachgen) does not need explicit support for optional fields, as they are internally oneof fields and it has no special handling for them.


brycemcanally (approved):
lgtm

PR #9073, opened 2023-07-12
Add Determined

chainlink:
Installs the determined python cli as part of our image build process.

[INT-1042]

[INT-1042]: https://pachyderm.atlassian.net/browse/INT-1042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/jupyter-extension/Dockerfile
+++ b/jupyter-extension/Dockerfile
@@ -31,3 +31,4 @@ USER $NB_UID
 COPY dist dist
 WORKDIR /home/jovyan
 RUN pip install `find /app/dist/ -name \*.whl` nbgitpuller
+RUN pip install determined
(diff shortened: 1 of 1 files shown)


jupyter-extension/Dockerfile
```diff
@@ -31,3 +31,4 @@ USER $NB_UID
 COPY dist dist
 WORKDIR /home/jovyan
 RUN pip install `find /app/dist/ -name \*.whl` nbgitpuller
+RUN pip install determined
```
bbonenfant, line 34: You probably want to pin the version so it doesn't change out from under you when you rebuild

chainlink: Normally I'd agree, but I actually was thinking always grabbing latest might be best here?

bbonenfant: If this is just so that the users don't need to pip install determined then its probably fine, but as a result we won't be able to know what version of the determined SDK is included with a specified release of the extension without spinning up the container ourselves

PR #9074, opened 2023-07-12
use xlarge instead of 2xlarge

djanicekpach:
we were using 2xlarge as runners after talking with circle-ci but it didn't really help. Reverting back to xlarge.


molinamelendezj (approved):
💸

PR #9079, opened 2023-07-13
Return Loadbalancer IP if service type load balancer

chainlink:
Initial findings:
- AWS/GCP report the Loadbalancer IP to the `status` of the Service. So we'll need to introspect the status to get the IP
- GCP reports this as an IP, AWS reports it as a hostname

AWS:
```
status:
  loadBalancer:
    ingress:
      - hostname: [a1c557bb8644d434f88807df8cdf006b-415633651.us-west-2.elb.amazonaws.com](http://a1c557bb8644d434f88807df8cdf006b-415633651.us-west-2.elb.amazonaws.com/)
```

GCP:
```
 status:
   loadBalancer:
      ingress:
        - ip: 35.197.80.240
```

MetalLB:
```
status:
  loadBalancer:
    ingress:
    - ip: 10.45.2.243
```

IP is now exposed in console 
image

Future work:
- [ ] On upgrade cluster - does the service get deleted?
- [ ] Explore The kubernetes REST API (It seems to disambiguate this into the `External-IP` field, see `kubectl get svc` output 
- [ ] Test on newly created svc (with LB pending) and check for errors in pachd logs
- [ ] Gather other envs (openshift?) and see how they handle LBs

[INT-1047]

[INT-1047]: https://pachyderm.atlassian.net/browse/INT-1047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/server/pps/server/api_server.go
```diff
@@ -2658,7 +2658,19 @@ func (a *apiServer) inspectPipeline(ctx context.Context, pipeline *pps.Pipeline,
 					return nil, errors.EnsureStack(err)
 				}
 			} else {
-				info.Details.Service.Ip = service.Spec.ClusterIP
+				if info.Details.Service.Type == "LoadBalancer" {
+					//GCP: service.Status.LoadBalancer.Ingress[0].IP
+					//AWS: service.Status.LoadBalancer.Ingress[0].Hostname
+					if len(service.Status.LoadBalancer.Ingress) == 1 {
```
robert-uhl, line 2664: I think that you expect `len(service.Status.LoadBalancer.Ingress)` to be either zero or one here; is it worth explicitly erroring if it is two or more?

robert-uhl (approved):

src/server/pps/server/api_server.go
chainlink: We chatted about this offline. While having two more more in an exceptional condition, it's not an error condition. ie. we don't want to blow up the `inspectPipeline` call if there's too many load balancer IPs. I'm also not aware of any kube implementation that returns more than one.

PR #9088, opened 2023-07-14
fix chomp miss-use in helm chart 

molinamelendezj:
fixes `Error: INSTALLATION FAILED: YAML parse error on pachyderm/templates/pachd/deployment.yaml: error converting YAML to JSON: yaml: line 159: mapping values are not allowed in this context`


molinamelendezj:
closed. fixed in #9103

PR #9090, opened 2023-07-17
python-sdk - add spout101 example

kevyang:
adds the spout101 example ported for python-sdk. also updates the default behavior of an empty spout to ensure that it is serialized post init.

[INT-1027](https://pachyderm.atlassian.net/browse/INT-1027)

[INT-1027]: https://pachyderm.atlassian.net/browse/INT-1027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python-sdk/pachyderm_sdk/api/pps/_additions.py
+++ b/python-sdk/pachyderm_sdk/api/pps/_additions.py
@@ -1,0 +1,10 @@
+from . import CreatePipelineRequest, Service, Spout
+
+def _Create_Pipeline_Request__post_init__(self: "CreatePipelineRequest") -> None:
+    """
+    Ensure that we serialize an empty Spout object
+    """
+    if self.spout and self.spout == Spout():
+        self.spout = Spout(service=Service())
+
+CreatePipelineRequest.__post_init__ = _Create_Pipeline_Request__post_init__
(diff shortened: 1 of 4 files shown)


python-sdk/pachyderm_sdk/api/pps/_additions.py
```diff
@@ -0,0 +1,10 @@
+from . import CreatePipelineRequest, Service, Spout
+
+def _Create_Pipeline_Request__post_init__(self: "CreatePipelineRequest") -> None:
```
bbonenfant, line 3: I think the simpler approach here is to have this __post_init__ method be implemented on the `Spout` object rather than the CreatePipelineRequest object. What was the reason you chose this object?

python-sdk/pachyderm_sdk/api/pps/_additions.py
```diff
@@ -0,0 +1,10 @@
+from . import CreatePipelineRequest, Service, Spout
```
bbonenfant, line 1: Could you copy the structure of file docstring from the `pfs/_additions.py` file?

bbonenfant (approved):
Looks great!

python-sdk/pachyderm_sdk/api/pps/_additions.py
kevyang: yeah, that was the direction i had wanted to go initially as well. but i was not able to get it working; this is what i tried:
```
def _Spout__post_init__(self: "Spout") -> None:
    if not self.service:
        self.service = Service()

Spout.__post_init__ = _Spout__post_init__
```
i feel like the spout object should get created which should cause this to kick in but maybe i'm wrong on that.

PR #9094, opened 2023-07-17
Jonathan/delete repo returns deleted repos2

lemontree-ca:


jrockway (changes requested):
Let's add some tests in `src/server/pfs/server/cmds/cmds_test.go` to check the new exit status code.  Otherwise, looks good.

src/pfs/pfs.proto
```diff
@@ -204,6 +204,12 @@ message DeleteRepoRequest {
   bool force = 2;
 }
 
+// DeleteRepoResponse returns the repos that were deleted by a DeleteRepo call.
+message DeleteRepoResponse {
+  // The repos that were deleted, perhaps none.
+  repeated Repo deleted_repos = 1;
```
brycemcanally, line 210: Is there an expectation that this will eventually include system repos? Otherwise, it seems like the single deleted repo would suffice.

brycemcanally:
Is the goal with this change just to know whether the delete was a no-op? Also, this is a breaking API change, so we might want to consider delivering this change differently.

src/pfs/pfs.proto
lemontree-ca: Hi Bryce, this is to add a response message when "pachctl delete repo XXX" is executed to delete a non-existing (not created) repo saying "No repos deleted". Not sure 

e.x.:
americaswangiri@Iriss-MacBook-Pro bin % ./pachctl delete repo iris_repo1 --project iris1
No repos deleted.

lemontree-ca: Hi Bryce, I'm not sure what you mean this comment: "Also, this is a breaking API change, so we might want to consider delivering this change differently.". I need more context for that.Thanks.

src/server/pachyderm_test.go
```diff
@@ -8841,6 +8843,26 @@ func TestDeleteSpecRepo(t *testing.T) {
 	require.YesError(t, err)
 }
 
+func TestDeleteRepo(t *testing.T) {
```
brendoncarroll, line 8846: Can we move this test to one of the files in `/src/server/pfs/server/testing`?

worker_init
brendoncarroll, line 1: Why is this file committed?

src/pfs/pfs.proto
brendoncarroll: DeleteRepo will only ever return 1 Repo.  This should return a boolean or a single Repo.

Let's change this to:
```protobuf
Repo deleted_repo = 1;
```

or 

```protobuf
bool deleted = 1;
```

brendoncarroll (changes requested):
A few things to fix, let's focus on the API first.

src/pfs/pfs.proto
lemontree-ca: updated as "Repo deleted_repo = 1"

src/server/pachyderm_test.go
lemontree-ca: moved to "server_test.go"

worker_init
lemontree-ca: removed.

src/pfs/pfs.proto
```diff
@@ -207,7 +207,7 @@ message DeleteRepoRequest {
 // DeleteRepoResponse returns the repos that were deleted by a DeleteRepo call.
 message DeleteRepoResponse {
   // The repos that were deleted, perhaps none.
-  repeated Repo deleted_repos = 1;
+  Repo deleted_repos = 1;
```
brendoncarroll: `deleted_repo` -> `deleted_repos`

brendoncarroll (approved):
Fix plural -> singular, then looks good.

